PR #6687, opened 2021-10-01
Fix plugin provides load order

kennytv:
Fixes dumb plugin load order when a plugin "provides" another plugin via the plugin.yml https://hub.spigotmc.org/jira/browse/SPIGOT-6740


Proximyst (approved):

MiniDigger (approved):
lgtm 
maybe mention that issue in the paper comment for context

NotMyFault (approved):
That does the trick.

PR #6690, opened 2021-10-01
Add missing team sidebar display slots

Machine-Maker:
There are 19 display slots, not just the 3 in the original enum


MiniDigger (approved):
TIL these exist

electronicboy (approved):

dawon:
Isn't it a problem that there are two server patches numbered 814?

Machine-Maker:
not really, should've been rebased and moved before merging, but itll get fixed on the next merge. Patches are just applied in the file system order, the name or number of the patch doesn't really matter.

dawon:
Ah, OK I was rebasing my PR and was wondering why I have there those weird renames... Thank you for clarification! :-)

MiniDigger:
but yes, generally we try to prevent that kinda stuff, but cat was asleep while committing :D

PR #6697, opened 2021-10-02
Properly handle damage inside EntityDamageByBlockEvent

### LINKED ISSUE #6696: Calling the Entity.damage() function in an EntityDamageEvent with block damage causes an error.
### Expected behavior

The block damage and the damage from the damage() call to both successfully damage their respective targets without causing errors.

### Observed/Actual behavior

The damage call did not go through and instead threw an error
Full Error: https://paste.gg/p/anonymous/5af0356e85824f8c875132996ab9f665

### Steps/models to reproduce

Creating a registering an EventHandler in the form of this:

@EventHandler
  public void damagedEntity(EntityDamageEvent event) {
    if (event.getCause() != EntityDamageEvent.DamageCause.CUSTOM)
      for (Player player : Bukkit.getOnlinePlayers())
        player.damage(1.0D);  
  }
  
  Will reproduce the error.
  I've done some testing, and as far as I can tell, all that you need to cause this error is to put some blockdamage on the stack, and then put a different type of damage on the stack as well.

### Plugin list

Plugins (1): BlockDamageTest

(The plugin I created to reproduce the error)

### Paper version

[22:22:17 INFO]: Checking version, please wait...
[22:22:17 INFO]: This server is running Paper version git-Paper-295 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: f905057)
Previous version: git-Paper-788 (MC: 1.16.5)

### Agreements

### Other

_No response_

Machine-Maker:
Fixes #6696.

If this isnt the best way to go about this, it at least illustrates the problem. When you have a bunch of entities being damaged at once by the same thing (an entity with passengers taking fall damage from dripstone pointy things), you cant clear the blockDamage right after because there's another entity right behind it taking damage (the passenger).

So I just changed it to a map that entities are added to, and then removed after.

### DIFF
--- a/patches/server/0818-Properly-handle-damage-inside-EntityDamageByBlockEve.patch
+++ b/patches/server/0818-Properly-handle-damage-inside-EntityDamageByBlockEve.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake <jake.m.potrebic@gmail.com>
+Date: Tue, 30 Nov 2021 21:29:35 -0800
+Subject: [PATCH] Properly handle damage inside EntityDamageByBlockEvent
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index aaf47d3f58819ead3553973c159c5c6eec4abe43..7d57d5283a12880cb18ee4d8192d0d8afbbc9b83 100644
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Ok, significantly simplified this fix. Just clearing and re-assigning after the event handler seems to fix the issue.

Machine-Maker:
Handled by [upstream](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/3a8161fe41ccfd5fb7ffdfe13bd3e08884520193)

Issue #6598, opened 2021-09-13
Spawn limit missing for underground water creature


Toseflo:
### Expected behavior

There should be a config option to change the spawn limit of underground water creatures (glow squids and axolotls).

### Observed/Actual behavior

In paper.yml, there is a spawn-limits config option, but underground water creatures are missing.

### Steps/models to reproduce

Change every spawn-limits value to 0 in paper.yml
Run the new /paper mobcaps command
Notice how underground_water_creatures still has the default spawn limit of 5, even though every other group is 0

### Plugin list

Multiverse-Core

### Paper version

This server is running Paper version git-Paper-259 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 810d200)
You are running the latest version
Previous version: git-Paper-119 (MC: 1.17.1)

### Agreements

- [X] I am running the latest version of Paper available from https://papermc.io/downloads.
- [X] I have searched for and ensured there isn't already an open issue regarding this.
- [X] My version of Minecraft is supported by Paper.

Doc94:
i see a PR here where tell this need to be make in upstream first..

then for tracing i add here the PR i make for bukkit/cb if its not problem.

https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/934/overview
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/pull-requests/662/overview

Doc94:
PR merged (really fast...) and upstream in a3dfe6d

PR #6698, opened 2021-10-02
Added underwater animals config

adrian-kong:
added feature #6598 
/paper mobcaps now updates the under-water config.


kennytv (changes requested):
Please don't auto format classes, and "underground water" != "underwater". This also has to be merged into the original patch (see contributing guidelines on how to edit patches)

lynxplay (changes requested):
Hi and welcome to paper! In regards to kennys request, you can find the detailed explanation on how to edit an existing patch [here](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#modifying-patches),

jpenilla (changes requested):
Should probably wait until upstream adds configs regarding `underground_water_creature` before we do this, it's inevitable that right after we merge this upstream will add their configs with slightly different names (as they don't use the official mojang ones), and it would be nice to not need to write migration code later/explain to users why the paper and bukkit configs use different names.

PR #6701, opened 2021-10-02
Fix CME in CraftPersistentDataTypeRegistry

Gerrygames:
When using PersistentDataContainers in an asynchronous task a CME can occur. Afaik creating/modifing ItemStacks asynchronously should be allowed. This PR fixes the CME.
http://paste.cytooxien.de/icetuzifeq.bash

### DIFF
--- a/patches/server/0814-Fix-CME-in-CraftPersistentDataTypeRegistry.patch
+++ b/patches/server/0814-Fix-CME-in-CraftPersistentDataTypeRegistry.patch
@@ -5,16 +5,15 @@ Subject: [PATCH] Fix CME in CraftPersistentDataTypeRegistry
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
-index 355c9f79fd3132848a00eacde951d1e1bfa92737..1078a4ad4d81ab0c967ac0c658c1eb04676a3262 100644
+index 355c9f79fd3132848a00eacde951d1e1bfa92737..6070f7d954e201e06efb6ef27ad0e7d9ff1ab9ff 100644
 --- a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
 +++ b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
-@@ -92,7 +92,8 @@ public final class CraftPersistentDataTypeRegistry {
+@@ -92,7 +92,7 @@ public final class CraftPersistentDataTypeRegistry {
          }
      }
  
 -    private final Map<Class, TagAdapter> adapters = new HashMap<>();
-+    // Paper - Replace HashMap with ConcurrentHashMap to avoid CME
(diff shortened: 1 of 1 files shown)


imDaniX:
Why not `ConcurrentHashMap`?

Gerrygames:
Yeah ConcurrentHashMap is better. I will change that.

patches/server/0814-Fix-CME-in-CraftPersistentDataTypeRegistry.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
+index 355c9f79fd3132848a00eacde951d1e1bfa92737..1078a4ad4d81ab0c967ac0c658c1eb04676a3262 100644
+--- a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
++++ b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataTypeRegistry.java
+@@ -92,7 +92,8 @@ public final class CraftPersistentDataTypeRegistry {
+         }
+     }
+ 
+-    private final Map<Class, TagAdapter> adapters = new HashMap<>();
++    // Paper - Replace HashMap with ConcurrentHashMap to avoid CME
```
me4502: Please put the Paper comment on the line of the change itself

kennytv (approved):

Machine-Maker:
I'm thinking it doesnt have to be a concurrent hash map for **every** instance of CraftPersistentDataTypeRegistry, just ones created by CraftMetaItem. Why needlessly make it a concurrent hash map for entities, chunks, tile entities if those arent supposed to be modified off the main thread?

Gerrygames:
CraftPersistentDataTypeRegistry is a singleton.

Proximyst (approved):

Machine-Maker:
Are you sure? I was pretty sure I saw it being instantiated in multiple places

Machine-Maker:
> CraftPersistentDataTypeRegistry is a singleton.

Yeah, its definitely **not** a singleton. It's instantiated 4 different times, as static final fields on `BlockEntity`, `LevelChunk`, `CraftEntity`, and `CraftMetaItem`? why shouldn't only the one created for CraftMetaItem use the ConcurrentHashMap?

Proximyst:
I don’t see an issue with it. It’s a per-bucket locked map, and registering adapters is not really a problem to do concurrently.

Machine-Maker:
> I don’t see an issue with it. It’s a per-bucket locked map, and registering adapters is not really a problem to do concurrently.

I don't know the performance different between ConcurrentHashMap, and HashMap, but that adapter map is used in PersistentDataContainer#get, has, and set, computing the missing adapter if its not already present in that registry. If there's any difference at all performance-wise between the two, I think the faster should be used in the situation where single-threadedness is already expected

Owen1212055:
> I think the faster should be used in the situation where single-threadedness is already expected

How about just making it so you can specify in the constructor or what not to be backed by a synchronized vs normal map? 
This way it would only affect item stacks or whatnot.

PR #6703, opened 2021-10-02
Re-readd root/admin user detection

NoahvdAa:
This PR re-adds the root reverted in e8830b27. The check was throwing false positives because of a OpenJDk bug (no issue # yet, will add here later). This PR uses a different method for getting the UID (executing the id command), which isn't affected by the openjdk issue.

### DIFF
--- a/patches/server/0814-Add-root-admin-user-detection.patch
+++ b/patches/server/0814-Add-root-admin-user-detection.patch
@@ -12,10 +12,10 @@ Co-authored-by: Noah van der Aa <ndvdaa@gmail.com>
 
 diff --git a/src/main/java/io/papermc/paper/util/ServerEnvironment.java b/src/main/java/io/papermc/paper/util/ServerEnvironment.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..9bc2cbf015998b98b6681a427596b151ed1806a0
+index 0000000000000000000000000000000000000000..18ea1dddf600920999d6b254bc55cf5d0d8a1cfb
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/util/ServerEnvironment.java
-@@ -0,0 +1,36 @@
+@@ -0,0 +1,39 @@
 +package io.papermc.paper.util;
 +
 +import com.sun.security.auth.module.NTSystem;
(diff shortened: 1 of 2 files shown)


clrxbl:
Why run `id -u` with the username specifically?
The command should already return the current user's ID.

Prof-Bloodstone:
Wondering if it'd be better to use available API as before, and if it returns root as user, then use the command to confirm it?
This way it could potentially be faster and doesn't require for the `id` binary to be on the system most of the time.

patches/server/0814-Add-root-admin-user-detection.patch
```diff
@@ -0,0 +1,75 @@
...
++
++    static {
++        if (SystemUtils.IS_OS_WINDOWS) {
++            RUNNING_AS_ROOT_OR_ADMIN = Set.of(new NTSystem().getGroupIDs()).contains(WINDOWS_HIGH_INTEGRITY_LEVEL);
++        } else {
++            boolean isRunningAsRoot = false;
++            if (new UnixSystem().getUid() == 0) {
++                // Due to an OpenJDK bug, UnixSystem#getUid incorrectly returns 0 when the user doesn't have a username.
++                // Because of this, we'll have to double-check if the user ID is actually 0 by running the id command.
++                try (final InputStream inputStream = Runtime.getRuntime().exec("id -u").getInputStream()) {
```
Proximyst: Should use processes.

Proximyst (changes requested):

patches/server/0814-Add-root-admin-user-detection.patch
NoahvdAa: Is there an advantage to spawning a process with something like a ProcessBuilder (assuming that's what you mean here) vs spawning one via `Runtime#exec`?

me4502:
How common is it for servers to not contain the id binary?

I'd question whether it's better to have it default to warning when it's missing - but guess it depends how often it's missing 

Logic being a host can easily fix it themselves if it shows up due to missing id binary

patches/server/0814-Add-root-admin-user-detection.patch
```diff
@@ -0,0 +1,78 @@
...
++    private static final boolean RUNNING_AS_ROOT_OR_ADMIN;
++    private static final String WINDOWS_HIGH_INTEGRITY_LEVEL = "S-1-16-12288";
++
++    static {
++        if (SystemUtils.IS_OS_WINDOWS) {
++            RUNNING_AS_ROOT_OR_ADMIN = Set.of(new NTSystem().getGroupIDs()).contains(WINDOWS_HIGH_INTEGRITY_LEVEL);
++        } else {
++            boolean isRunningAsRoot = false;
++            if (new UnixSystem().getUid() == 0) {
++                // Due to an OpenJDK bug, UnixSystem#getUid incorrectly returns 0 when the user doesn't have a username.
```
lynxplay: Should probably now also reference the [openjdk issue](https://bugs.openjdk.java.net/browse/JDK-8274721) to ease traceability/link to more information as to why this double check had to be implemented.

lynxplay (changes requested):

me4502 (approved):

Proximyst (approved):

PR #6706, opened 2021-10-02
Fix placeable/destory keys deserialization from spigot

### LINKED ISSUE #6689: CanDestroy and CanPlaceOn tags from serialized item stacks are lost when switching between Spigot and Paper.
### Expected behavior

Server owners being able to freely switch between Spigot and Paper back and forth without risking serialized item data to be lost.

### Observed/Actual behavior

A user of one of my plugins, which uses the Bukkit ItemStack config serialization API to save and load item stacks to and from config files, complained about the `CanDestroy` and `CanPlaceOn` item tags being lost when they switched from using Spigot to using Paper. I assume the same issue applies in the opposite direction as well, when switching from Paper back to Spigot.

### Steps/models to reproduce

* Create an item stack with a `CanDestroy` tag. For example:
`/minecraft:give @s minecraft:diamond_pickaxe{CanDestroy:["minecraft:iron_ore"]}`
* Use the Bukkit config serialization API to save this item stack on Spigot. Example output:
```
==: org.bukkit.inventory.ItemStack
v: 2730
type: DIAMOND_PICKAXE
meta:
  ==: ItemMeta
  meta-type: UNSPECIFIC
  internal: H4sIAAAAAAAA/+NiYOBk4HJOzHNJLS4pyq/kYGBgYGQQys3MS00uSkwrscosys+Lzy9KZQAAMhDaySoAAAA=
```
* Switch over to using Paper, or copy this data to a server that runs Paper.
* Load this item stack using the Bukkit config serialization API.
* Observation: The item's `CanDestroy` tag is not loaded.
* Saving the just loaded item stack back to the config yields this output (no traces of the item's previous `CanDestroy` data remains):
```
==: org.bukkit.inventory.ItemStack
v: 2730
type: DIAMOND_PICKAXE
```

As per below explanation for the cause of this issue, I assume the same issue also applies when serializing an item stack with this `CanDestroy` tag on Paper, switching from Paper to Spigot, and attempting to load the item stack there. The serialized item stack data on Paper, which Spigot will not know and therefore ignore, will look like this:
```
==: org.bukkit.inventory.ItemStack
v: 2730
type: DIAMOND_PICKAXE
meta:
  ==: ItemMeta
  meta-type: UNSPECIFIC
  CanDestroy:
  - minecraft:iron_ore
```

### Plugin list

Shopkeepers

The underlying cause of the issue should be apparent, but if wanted, I can prepare a more minimal reproduction plugin that simply saves and loads an item stack to and from the plugin's config file.

### Paper version

This server is running Paper version git-Paper-295 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: f905057)
Previous version: git-Paper-280 (MC: 1.17.1)

### Agreements

### Other

My assumption is that the issue is caused by the following patch: https://github.com/PaperMC/Paper/blob/master/patches/server/0267-Implement-an-API-for-CanPlaceOn-and-CanDestroy-NBT-v.patch

On Spigot, the `CanDestroy` and `CanPlaceOn` item tags are serialized as part of the `unhandledTags` (under key `internal`).
On Paper, these tags have been removed from Spigot's `unhandledTags` and treated separately during serialization and deserialization: They are serialized under the keys `CanDestroy` and `CanPlaceOn` inside configs.

The problem with this is that Paper ignores any already serialized data for these tags that may previously have been saved by Spigot as part of the `internal` / `unhandledTags`. And the other way around, Spigot ignores any data that Paper may have previously saved under the `CanDestroy` and `CanPlaceOn` keys.

I see two possible solutions:
* Paper could additionally import the data for these tags from Spigot's serialized `internal` data. However, this would only resolve one side of the issue: When the serialized item data is transferred from a Spigot server over to a Paper server. Copying the serialized item data from a Paper server over to a Spigot server will not work with this solution.
* Paper could stick to reading and writing the data for these tags from/to Spigot's `unhandledTags` (under key `internal`). The Paper-specific API could still be implemented on top of that. The benefit of this would be that any items saved on the Paper server can also be transferred over to a Spigot server without these tags being lost.  
Since Paper servers will already have serialized items with these tags, Paper would additionally need to import any tags that have previously been saved under the `CanDestroy` and `CanPlaceOn` keys.

Machine-Maker:
Fixes #6689

### DIFF
--- a/patches/server/0258-Implement-an-API-for-CanPlaceOn-and-CanDestroy-NBT-v.patch
+++ b/patches/server/0258-Implement-an-API-for-CanPlaceOn-and-CanDestroy-NBT-v.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Implement an API for CanPlaceOn and CanDestroy NBT values
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java
-index d5df6fc244ab82b94196be9c436ba77020716df2..ac779f3cedb7ddd74f39a18f08afbcdad8cd13b1 100644
+index d5df6fc244ab82b94196be9c436ba77020716df2..d08ab9cd7eede4e7a072c5567d9ad0de2abb00c9 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java
 @@ -83,6 +83,12 @@ import org.bukkit.persistence.PersistentDataContainer;
@@ -37,8 +37,8 @@ index d5df6fc244ab82b94196be9c436ba77020716df2..ac779f3cedb7ddd74f39a18f08afbcda
      private boolean unbreakable;
      private int damage;
 +    // Paper start - Implement an API for CanPlaceOn and CanDestroy NBT values
-+    private Set<Namespaced> placeableKeys = Sets.newHashSet();
-+    private Set<Namespaced> destroyableKeys = Sets.newHashSet();
++    private List<Namespaced> placeableKeys = Lists.newArrayList();
++    private List<Namespaced> destroyableKeys = Lists.newArrayList();
 +    // Paper end
  
      private static final Set<String> HANDLED_TAGS = Sets.newHashSet();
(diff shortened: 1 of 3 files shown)


blablubbabc:
Somethings else to consider regarding the representation of those tags in Bukkit: Minecraft stores these tags in a list (duplicates are not filtered, order is preserved), whereas this CraftMetaItem implementation seems to put them into an unordered HashSet. So a back and forth conversion from Minecraft -> Bukkit -> Minecraft might result in those tags being reordered, potentially affecting the comparisons of Minecraft items. Maybe it makes sense to store these tags in a List in Bukkit as well. Or at least store them in a LinkedHashSet. Or enforce a particular order, similar to how enchantments are reordered in Paper.

Machine-Maker:
> Somethings else to consider regarding the representation of those tags in Bukkit: Minecraft stores these tags in a list (duplicates are not filtered, order is preserved), whereas this CraftMetaItem implementation seems to put them into an unordered HashSet. So a back and forth conversion from Minecraft -> Bukkit -> Minecraft might result in those tags being reordered, potentially affecting the comparisons of Minecraft items. Maybe it makes sense to store these tags in a List in Bukkit as well. Or at least store them in a LinkedHashSet. Or enforce a particular order, similar to how enchantments are reordered in Paper.

Yeah, I just handled that by changing the fields on CraftMetaItem to lists, but the API still returns a set.

patches/server/0267-Implement-an-API-for-CanPlaceOn-and-CanDestroy-NBT-v.patch
```diff
@@ -5,10 +5,18 @@ Subject: [PATCH] Implement an API for CanPlaceOn and CanDestroy NBT values
...
-index 71320d9484842be3a694117de25159f3581bd2a3..0aa7854119ab876f0e46bb7667282d79d2a57e3d 100644
+index 71320d9484842be3a694117de25159f3581bd2a3..979e7626efcf114646fe48f9a0a2b9768606b993 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItem.java
-@@ -83,6 +83,12 @@ import org.bukkit.persistence.PersistentDataContainer;
+@@ -26,6 +26,7 @@ import java.util.Collection;
+ import java.util.Comparator; // Paper
+ import java.util.EnumSet;
+ import java.util.HashMap;
++import java.util.HashSet;
```
jpenilla: this import doesn't look used anymore?

Machine-Maker: Yeah, removed it

Machine-Maker:
Rebased for 1.18.1

electronicboy (approved):

Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Not applicable anymore

PR #6628, opened 2021-09-20
Allow explosions to affect projectiles

coredoesdev:
Pretty much just copying my BandagedPaper patch over here. This has been an issue in Spigot and it's forks for a while, and since it's a pretty easy patch I decided to make a PR.

From my understanding of this bug, as a lagfix Spigot introduced a check where if entities did not take damage from an explosion and/or are not in a specific whitelist, the explosion is not processed for that entity. Due to projectiles having no health, they cannot take damage from an explosion, and Spigot forgot to add them to the whitelist. This patch simply adds it.

PR #6707, opened 2021-10-03
Force explosion knockbacks when appropriate

### LINKED ISSUE #5415: Ender dragon takes no knockback from exploding beds
### Expected behavior

When the ender dragon is perching in a dragon fight, it should go a few blocks up when a bed is exploded below it. That's how it is in singleplayer and in vanilla host

### Observed/Actual behavior

Dragon does not mind the beds at all, keeps going down and breaks the next placed bed.

### Steps/models to reproduce

How it should work (vanilla server): https://drive.google.com/file/d/1eWu46BHeAf3efbf9hWYzCEbF-FrUASqL/view?usp=sharing

Paper server: https://drive.google.com/file/d/1NGUl6SnVSo1C6h2wPl7Ry-6zW4Dl5ewS/view?usp=sharing

### Plugin list

None while testing

### Paper version

[10:26:03 INFO]: This server is running Paper version git-Paper-570 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)

### Agreements

- [x] My version of Minecraft is supported by Paper.

### LINKED ISSUE #5429: tnt not affecting projectiles
### Expected behavior

I expected tnt to launch projectiles.

### Observed/Actual behavior

tnt did not launch projectiles, either in or out of water.

### Steps/models to reproduce

place tnt, spawn arrows / any projectile, light tnt

### Plugin list

None

### Paper version

This server is running Paper version git-Paper-571 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)

### Agreements

Machine-Maker:
Fixes #5415
Fixes #5429

Probably supercedes #6628 (I think this is a better way to fix it probably)

The dragon stuff was the most annoying, since you have to delay setting the movement until all the parts have had their events called. Its also a little hard to test, since Craftbukkit just flat out doesnt let dragons move at all when in the hover phase (see L252 of EnderDragon.java). Commenting the added conditional check out shows the effects better.

### DIFF
--- a/patches/server/0830-Force-explosion-knockbacks-when-appropriate.patch
+++ b/patches/server/0830-Force-explosion-knockbacks-when-appropriate.patch
@@ -1,0 +1,161 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 2 Oct 2021 17:20:01 -0700
+Subject: [PATCH] Force explosion knockbacks when appropriate
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index e17a3afa41fd628d2c4a3637ae19418e258a99b8..e9989ee5b77967fa072184eb4a8d79a66fe7f76d 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
++++ b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -572,7 +572,11 @@ public class EnderDragon extends Mob implements Enemy {
+                 if (source.getEntity() instanceof Player || source.isExplosion()) {
+                     float f1 = this.getHealth();
+ 
+-                    this.reallyHurt(source, amount);
++                    // Paper start
++                    if (!this.reallyHurt(source, amount)) {
++                        return false;
++                    }
++                    // Paper end
+                     if (this.isDeadOrDying() && !this.phaseManager.getCurrentPhase().isSitting()) {
+                         this.setHealth(1.0F);
+                         this.phaseManager.setPhase(EnderDragonPhase.DYING);
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/AbstractHurtingProjectile.java b/src/main/java/net/minecraft/world/entity/projectile/AbstractHurtingProjectile.java
+index 0dc335b3003ae3cf11828cc849763e271a3b365b..a5b01c43633204b29009126e4a96acf81061688f 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/AbstractHurtingProjectile.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/AbstractHurtingProjectile.java
+@@ -177,6 +177,7 @@ public abstract class AbstractHurtingProjectile extends Projectile {
+ 
+     @Override
+     public boolean hurt(DamageSource source, float amount) {
++        this.forceExplosionKnockback = true; // Paper
+         if (this.isInvulnerableTo(source)) {
+             return false;
+         } else {
+@@ -186,6 +187,7 @@ public abstract class AbstractHurtingProjectile extends Projectile {
+             if (entity != null) {
+                 // CraftBukkit start
+                 if (CraftEventFactory.handleNonLivingEntityDamageEvent(this, source, amount, false)) {
++                    this.forceExplosionKnockback = false; // Paper
+                     return false;
+                 }
+                 // CraftBukkit end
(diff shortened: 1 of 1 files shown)


coredoesdev:
This is likely better than my PR. Closing previous as this is a better solution.

embeddedt:
Any chance this (or a similar fix) will get looked at for 1.18? One-cycling has been a vanilla feature since 1.14 and it'd be great if it would be supported by Paper in this version.

mexiscool:
I made a thing that launches the dragon by spawning 267 tnt with a 0 tick fuse next to the dragon right when it gets summoned (to be exact this is where the TNT is spawned 1 134.76125 0). This works in vanilla mc, tested in 1.17.1 and 1.18.1. However, with this patch applied and the craftBukkit check commented out for good measure the dragon does not get launched. When commenting out this
```
if (entity instanceof net.minecraft.world.entity.boss.EnderDragonPart part && !wasDamaged) {
    cancelledDragonKnockbacks.add(part.parentMob.getId());
}
```
the dragon does get launched.

Machine-Maker:
> the dragon does get launched.

See that's odd. `wasDamaged` should return true there, since I imagine the explosions are damaging the dragon part. so it shouldn't be cancelling the knockback.

mexiscool:
I changed it to
```
if (entity instanceof net.minecraft.world.entity.boss.EnderDragonPart part && !wasDamaged) {
                            System.out.println(part.name);
                            cancelledDragonKnockbacks.add(part.parentMob.getId());
                        }
```

and with some python analyzed how often each dragon part name was printed for 267 TNT exploding found on first try {'head': 21, 'neck': 77, 'body': 265, 'tail': 417, 'wing': 519}, and on second try {'head': 2, 'neck': 65, 'body': 259, 'tail': 489, 'wing': 466}. Which is kinda a surprising result. In single-player with carpet mod, I have confirmed that all explosions happen at the exact same location in the same tick.

Machine-Maker:
Yeah ok, so I'm altering it a bit, where if even 1 part of the dragon has that `wasDamaged` variable set to true, it will apply the knockback to the entire dragon. That seems to work and still lets plugins cancel it by making sure to cancel all EntityDamageEvents relating to it.

Machine-Maker:
@mexiscool Try the patch now, remembering to comment out that HOVERING check on L252 of EnderDragon.java.

idk why that's there, it seems ancient and afaik, HOVERING is only the initial default state, not a state the dragon goes into at any point during an actual encounter.

mexiscool:
nope doesn't work, I also see that L261 in Explosion.java only gets executed 3 (4 times the second try) times is that correct?
Might that be the hurt method in EnderDragonPart.java being a sort of cooldown to limit wasDamaged being true to once per tick per part?

Machine-Maker:
> nope doesn't work, I also see that L261 in Explosion.java only gets executed 3 (4 times the second try) times is that correct?
> Might that be the hurt method in EnderDragonPart.java being a sort of cooldown to limit wasDamaged being true to once per tick per part?

Hmm. What I was testing def works. I just had a command block spawn a dragon, and then while it was circling around that point, I placed, and lit a tnt under it. And it def looked like it was pushed up. I’ll make a video of it in a bit.

Machine-Maker:
https://user-images.githubusercontent.com/15055071/145701285-81219f2d-eb14-40b1-af90-e3c2372863c5.mp4

mexiscool:
I also made some videos, so with the original change I did by removing this

 ```
 if (entity instanceof net.minecraft.world.entity.boss.EnderDragonPart part && !wasDamaged) {
     cancelledDragonKnockbacks.add(part.parentMob.getId());
 }
 ```

 it looks like this (same as in vanilla, the dragon gets shot very far away into an non-entity ticking chunk):



https://user-images.githubusercontent.com/27775925/145730100-6e6b28cd-2fe2-4417-8c7f-d0039b4b4e85.mp4


Now this is the current implementation it looks this


https://user-images.githubusercontent.com/27775925/145733489-dc9a3393-d949-4466-843e-6a2d7caf266f.mp4

Machine-Maker:
ok, so it is applying some knockback, just too little. ill look into this more. Can you share the way you are testing that so I can try and reproduce those exact results? the command block setup, etc. or even just a world download.

mexiscool:
[world_the_end.zip](https://github.com/PaperMC/Paper/files/7699992/world_the_end.zip)
All you need to do is place the 4 end crystals and the command blocks spawn the TNT each time the dragon spawns
in the spigot.yml i have changed the max-tnt-per-tick to 1000 and all the entity-activation-range's to at least 32 to make sure I'm also within range of the dragon and TNT when the explosion happens (I'm not sure if it matters for this but it does matter with arrow and pearl cannons).

mexiscool:
I edited my comment with the videos because I think I managed to mix up some video files so I retested it.

Machine-Maker:
is that test with the current implementation with or without that extra HOVERING check removed? cause that's gonna prevent it from moving at all until its phase changes the first time.

mexiscool:
hovering check is removed

Machine-Maker:
Well, [upstream will prob fix this](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/988/overview)

PR #6710, opened 2021-10-03
Added the ability to override YAML configurations

### LINKED ISSUE #6332: Add an environment variable for velocity secret

When deploying Minecraft servers at scale, you would want to automate as much as you can. What I would like to achieve is an immutable Docker image to deploy my Minecraft server. I would like to store Bukkit, Spigot, Paper, etc config inside the Docker image. However, it is not a good idea to store sensitive information like forwarding secret inside a Docker image. 

What would be ideal is to have an environment variable that can be defined at runtime, e.g. `VELOCITY_FORWARDING_SECRET`. This makes it easy to set in an environment like Kubernetes or Docker.

I have considered templating paper.yml as a ConfigMap with a templating tool like Helm, but this creates an enormous ConfigMap. 

### Agreements

### Other

_No response_

Cubxity:
## Why?

This patch allows users to override their Paper, Spigot, Bukkit, and Minecraft (server.properties) config at runtime using environment variables and system properties. This is very useful for containerized deployment of Minecraft servers, such as on Kubernetes and Docker.

## Example

```bash
PAPER_SETTINGS_VELOCITY_SUPPORT_SECRET=xxxx java -jar server.jar ...
```
This will override `settings.velocity-support.secret` to `xxxx` at runtime.

Resolves #6332.


NoahvdAa:
I feel like this should probably be part of the [paper config patch](https://github.com/PaperMC/Paper/blob/e39159134f3f52142a06f37a389a00c9ce3e724e/patches/server/0004-Paper-config-files.patch).

MiniDigger:
I love this. is it feasible to add it to bukkit.yml, spigot.yml and the server properties too? 
if not, this is prolly good enough as it already and can be expanded on later.

I would appreciate a PR to the docs that documents this new feature https://github.com/PaperMC/PaperDocs/blob/master/source/server/configuration.rst
maybe directly at the top or something?

Cubxity:
> I love this. is it feasible to add it to bukkit.yml, spigot.yml and the server properties too? if not, this is prolly good enough as it already and can be expanded on later.

Will look into it, since I believe it's going to be quite useful. Should the patch be renamed then, since it applies to other configs too?

Cubxity:
I left `bukkit.yml` alone since it is slightly more complicated to implement an override for it.

Cubxity:
@MiniDigger the requested changes have been implemented and the PR is ready to be reviewed. The documentation PR can be found here: https://github.com/PaperMC/PaperDocs/pull/118.

lynxplay:
We could probably safe a) a lot of repetition and b) support for the bukkit configuration using a layout similar to https://github.com/lynxplay/paper/commit/328f4f07d9f7e71f92d405130016eb9ed234bcc1.

I haven't particularly tested this yet, but shifting the override logic into the config itself would centralize it too, meaning future changes would be easier to implement.

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,330 @@
...
++/**
++ * An implementation of {@link OverrideSource}, which fetches values from system's environment.
++ * The requested path will be transformed in the following manner:
++ * - dot (.) will be replaced with an underscore (_),
++ * - dash (-) will be replaced with an underscore (_),
++ * - and it will be capitalized.
++ * <p>
++ * The group name will be prepended, followed by an underscore (_).
++ */
++public final class EnvironmentOverrideSource implements OverrideSource {
```
lynxplay: Is there actually any reason to implement these in their own classes ?
Looking at the `OverrideSource` interface, it seems to be a pretty straight forward candidate for a `FunctionalInterface`.
E.g. instead of having to mess around with private constructors and instances, 

You could move these as constants into the `OverrideSource` interface, e.g.
```java
/**
 * An implementation of {@link OverrideSource}, which fetches values from system's environment.
 * The requested path will be transformed in the following manner:
 * - dot (.) will be replaced with an underscore (_),
 * - dash (-) will be replaced with an underscore (_),
 * - and it will be capitalized.
 * <p>
 * The group name will be prepended, followed by an underscore (_).
 */
OverrideSource ENVIRONMENT_SOURCE = (path, group) -> {
    final var environmentKey = group + "_" + path.replace(".", "_").replace("-","_");
    return System.getenv(environmentKey.toUpperCase(Locale.ROOT));
};
```

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,330 @@
...
++            try {
++                return Long.parseLong(override);
++            } catch (NumberFormatException ignored) {
++
++            }
++        }
++        return super.getLong(path, def);
++    }
++
++    private @Nullable String getOverride(@NotNull String path) {
```
lynxplay, line 133: I don't know if there is a better way to do this, but this method basically matches the logic found in the `static` version in `OverrideSource`. Idk if we could expand that method to also take the sources array (making it purely functional), allowing basically any implementation to just call the static method.

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,330 @@
...
++        Validate.notNull(file, "File cannot be null");
++
++        OverridableYamlConfiguration config = new OverridableYamlConfiguration(group);
++
++        try {
++            config.load(file);
++        } catch (FileNotFoundException ex) {
++        } catch (IOException ex) {
++            Bukkit.getLogger().log(Level.SEVERE, "Cannot load " + file, ex);
++        } catch (InvalidConfigurationException ex) {
```
lynxplay: Could really just collapse these exceptions, no need to stick to the spigot double logic.

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,330 @@
...
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServerProperties.java
+@@ -144,4 +144,16 @@ public class DedicatedServerProperties extends Settings<DedicatedServerPropertie
+ 
+         return this.worldGenSettings;
+     }
++
++    // Paper start
++    @org.jetbrains.annotations.Nullable
++    @Override
++    public String getStringRaw(String key) {
```
lynxplay, line 241: Is this actually the best place to implement this ?
The parent of the `DedicatedServerProperties` class (`Settings`) already has the method `Settings#getOverride(String,String);String` that is called by the `Settings#getStringRaw` method.

Cubxity: `getOverride` is private in Settings. Currently, the only subclass of it is `DedicatedServerProperties`. However, we don't know whether something else is going to subclass it in the future.

lynxplay: Hmmm yes that is problematic tho with this implementation, environment vars would override even the command line flags respected by Settings#getStringRaw and the getOverride method respectively, wouldn't they ?

Cubxity: Good point, however I think having environment variables override command line flags is a good idea. Let's say you have the command line flags specified in Dockerfile and you would like to override it at runtime.

lynxplay: I think we will need more eyes on this. I'd personally expect command line flags to basically take priority over anything else.
Let's hope we get some maintainer/developer input 😅

MiniDigger: I would expect CLI > system prop > env var > file > default as a order, cause that's what the systems I encountered before do (for example, spring)
For your docker case, you can still make it work by having your flags as an env var in the start script and then you can override that.

lynxplay: With the `Settings#getOverride(String,String)` method public, wouldn't it now be feasible to implement this behaviour in an override of the `Settings#getOverride(String,String)` method in the `DedicatedServerProperties` class ?
Seems like it might save you the hassle of manually fetching the property (and might just in general be a better place for the logic).

Tho this is really just a comment, this (at least from my end) is also a valid solution.

Cubxity: I'm honestly not sure, Settings is an abstract class and we don't know whether something else will subclass it later. I personally wouldn't make another constructor or field there.

lynxplay: No no, like override the 'getOverride' method in the dedicated server properties class. Sorry if I wasn't clear on that.

Cubxity: Ah, although that would require transforming the `options` field afaik.

lynxplay: You are right sir, I stand corrected. For some reason my brain still believed the `Settings#getOverride(String,String)` would just return null and we could easily check :+1:

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,285 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cubxity <cubxity@protonmail.com>
```
Proximyst, line 2: Coauthor lynx as you took a major jab from his patch :p 

```
Co-authored-by: Bjarne Koll <lynxplay101@gmail.com>
```

Proximyst (changes requested):
Would this maybe be possible to work with world-specific stuff, too? 🤔

patches/server/0814-Overridable-YAML-configuration.patch
Cubxity: My bad, I coauthored them now.

Cubxity:
> Would this maybe be possible to work with world-specific stuff, too? thinking

That may be possible, however I personally don't see a use for it.

lynxplay:
> > Would this maybe be possible to work with world-specific stuff, too? thinking
> 
> That may be possible, however I personally don't see a use for it.

I would also personally support world configuration. Environment variables could also override the default "world" settings, allowing a single variable to affect basically every world.

//edit[0]: this should potentially already work ? World configurations use their parent yaml instance which was now replaced.

Cubxity:
> > > Would this maybe be possible to work with world-specific stuff, too? thinking
> > 
> > 
> > That may be possible, however I personally don't see a use for it.
> 
> I would also personally support world configuration. Environment variables could also override the default "world" settings, allowing a single variable to affect basically every world.

That's a great idea actually.

Edit: I believe this would already work as PaperWorldConfig uses PaperConfig#config.

Cubxity:
@Proximyst @lynxplay I've conducted some testing and concluded that the current implementation does work with world settings.

Tested with:
```bash
PAPER_VERBOSE=true PAPER_WORLD_SETTINGS_DEFAULT_OPTIMIZE_EXPLOSIONS=true java -jar server.jar nogui
```

image
image

Proximyst (approved):
Good work

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,286 @@
...
++
++    /**
++     * @param group   the group name of this configuration. This will be used as a prefix in sources.
++     *                The group name must only contain a-z, A-Z, and 0-9. It may contain underscore (_),
++     *                however, it must not end with underscore. Underscore and dots will be appended to
++     *                the group name in the respective sources.
++     * @param sources array of {@link OverrideSource} to fetch overridden values from.
++     */
++    public OverridableYamlConfiguration(@NotNull String group, @NotNull OverrideSource[] sources) {
++        Preconditions.checkArgument(group.matches("[a-zA-Z0-9_]*[a-zA-Z0-9]+"));
```
MiniDigger: can we precompile this regex in a static constant? its only called 3 times but still would look nicer.

e-im:
For ease of support if hosts happen to set some of these (I'm guessing things like `settings.timings.url`, rcon/query related things or spigot's restart script may be used by some), it would be nice to log when options are set this way when debug logging is enabled.

Cubxity:
> For ease of support if hosts happen to set some of these (I'm guessing things like `settings.timings.url`, rcon/query related things or spigot's restart script may be used by some), it would be nice to log when options are set this way when debug logging is enabled.

I'm not sure if that would be possible, since the overriding is done in OverridableYamlConfiguration. Unless, we want to ignore the "verbose" flag.

patches/server/0814-Overridable-YAML-configuration.patch
```diff
@@ -0,0 +1,289 @@
...
+
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index f421e6a2e43e0a673dbb8a9a2b4331387e523e02..702477b94b28f0463fa51ee2aa999515457eaa68 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+@@ -18,6 +18,7 @@ import java.util.logging.Level;
+ import java.util.regex.Pattern;
+ 
+ import com.google.common.collect.Lists;
++import io.papermc.paper.configuration.OverridableYamlConfiguration;
```
Machine-Maker: Remove this import since you are only using it once below

lynxplay:
I would personally really appreciate seeing this pulled :+1:

Machine-Maker (changes requested):
There should be a nice way to abstract and reduce the duplication of the overrideable YAML config. Lots of those “get” methods have very similar method bodies

electronicboy:
I forgot this existed, one thing, would this even work or be doable with configurate?
I can maaaybe understand some properties for some stuff but if this is defo not something I wanna maintain long run knowing that I'm going to be ripping out the entire current config system at some point

lynxplay:
> I forgot this existed, one thing, would this even work or be doable with configurate? I can maaaybe understand some properties for some stuff but if this is defo not something I wanna maintain long run knowing that I'm going to be ripping out the entire current config system at some point

Does configurate support an out of box environment overwrite ? I have been looking through the documentation and source but was unable to find logic that indicates support for these. In either case tho, we might be able to merge the configuration node parsed from the configuration files with a node constructed from the environment ?

electronicboy:
> Does configurate support an out of box environment overwrite ?

my understanding was no which was why I asked this question; I also have no idea exactly how you'd sanely produce a node from the environment or how merging even works given many of the plans we have to clean up the configuration mess that's been created over the years

zml2008:
In the case of configurate, it would be possible to construct a node from the environment variables *only* if you have a known set of keys unfortunately, and of course it would be read-only. You'd have to be careful to keep data sources separate, to still be able to update the on-disk file as settings change.

lynxplay:
> In the case of configurate, it would be possible to construct a node from the environment variables _only_ if you have a known set of keys unfortunately, and of course it would be read-only. You'd have to be careful to keep data sources separate, to still be able to update the on-disk file as settings change.

Hmm, we could potentially achieve this by keeping the configurations defaults in its own node couldn't we ? E.g. construct the entire defaults of the configuration as a node at runtime so we can then `envNode.mergeFrom(diskNode.mergeFrom(defaultNodes))` while we construct the `envNode` based on the nodes (and child nodes) found in `defaultNodes`.

Cubxity:
I would personally appreciate this getting merged 😅

electronicboy:
I honestly do not feel that this is a maintainable solution especially in the long term given our plans redo all the configuration stuff

Cubxity:
> I honestly do not feel that this is a maintainable solution especially in the long term given our plans redo all the configuration stuff

What do you suggest would be a solution that can work before the refactor then?

electronicboy:
I honestly have no idea but I'm not exactly in favour of something which would be slated for removal on day 1, and have no idea how this would be translated into configurate as-is, the way the environment variables are formed at least screws us over on the potential of tryna mock some configuration nodes up to attempt to pull this together given that you need to know how to parse that stuff beforehand into a set of config nodes which you can work with

PR #6712, opened 2021-10-03
Add methods to find targets for lightning strikes

dawon:
There was no way to find lightning rods around a location except for reimplementing methods already present in NMS...

Thanks to @lynxplay for investigating a problem with ATs :)

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -241,3 +241,6 @@ public net.minecraft.world.entity.projectile.Projectile ownerUUID
 public org.bukkit.craftbukkit.scoreboard.CraftScoreboardTranslations
 public org.bukkit.craftbukkit.scoreboard.CraftScoreboardTranslations toBukkitSlot(I)Lorg/bukkit/scoreboard/DisplaySlot;
 public org.bukkit.craftbukkit.scoreboard.CraftScoreboardTranslations fromBukkitSlot(Lorg/bukkit/scoreboard/DisplaySlot;)I
+
+# Add methods to find targets for lightning strikes
+public net.minecraft.server.level.ServerLevel findLightningRod(Lnet/minecraft/core/BlockPos;)Ljava/util/Optional;
--- a/patches/api/0337-Add-methods-to-find-targets-for-lightning-strikes.patch
+++ b/patches/api/0337-Add-methods-to-find-targets-for-lightning-strikes.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jakub Zacek <dawon@dawon.eu>
+Date: Mon, 4 Oct 2021 08:29:36 +0200
+Subject: [PATCH] Add methods to find targets for lightning strikes
+
+
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index 77577de113b1413ed1474850b4ef8e7fd0c07dd6..4f673e9123145dc78564dc3eef0edf75795dafc2 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -757,6 +757,37 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+     @NotNull
+     public LightningStrike strikeLightningEffect(@NotNull Location loc);
+ 
++    // Paper start
++    /**
++     * Finds the location of the nearest unobstructed Lightning Rod in a 128-block
++     * radius around the given location. Returns {@code null} if no Lightning Rod is found.
++     *
++     * <p>Note: To activate a Lightning Rod, the position one block above it must be struck by lightning.</p>
++     *
++     * @param location {@link Location} to search for Lightning Rod around
++     * @return {@link Location} of Lightning Rod or {@code null}
++     */
++    @Nullable
++    public Location findLightningRod(@NotNull Location location);
++
++    /**
++     * Finds a target {@link Location} for lightning to strike.
++     * <p>It selects from (in the following order):</p>
++     * <ol>
++     *  <li>the block above the nearest Lightning Rod, found using {@link World#findLightningRod(Location)}</li>
++     *  <li>a random {@link LivingEntity} that can see the sky in a 6x6 cuboid
++     *      around input X/Z coordinates. Y ranges from <i>the highest motion-blocking
++     *      block at the input X/Z - 3</i> to <i>the height limit + 3</i></li>
++     * </ol>
++     * <p>Returns {@code null} if no target is found.</p>
++     *
++     * @param location {@link Location} to search for target around
++     * @return lightning target or {@code null}
++     */
++    @Nullable
++    public Location findLightningTarget(@NotNull Location location);
++    // Paper end
++
+     /**
+      * Get a list of all entities in this World
+      *
--- a/patches/server/0821-Add-methods-to-find-targets-for-lightning-strikes.patch
+++ b/patches/server/0821-Add-methods-to-find-targets-for-lightning-strikes.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jakub Zacek <dawon@dawon.eu>
+Date: Mon, 4 Oct 2021 10:16:44 +0200
+Subject: [PATCH] Add methods to find targets for lightning strikes
+
+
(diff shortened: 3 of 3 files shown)


kennytv (approved):

patches/server/0814-Add-methods-to-find-targets-for-lightning-strikes.patch
```diff
@@ -0,0 +1,34 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+index 89f2cd2edfc42ea4642828ea44404006449ac5df..2469c11ae43964079a146107fdb460e42c3c388c 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -699,6 +699,23 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+         return (LightningStrike) lightning.getBukkitEntity();
+     }
+ 
++    // Paper start
```
jpenilla: Make use of `MCUtil.toBlockPosition` and `MCUtil.toLocation`

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
```diff
@@ -0,0 +1,47 @@
...
++    public Location findLightningRod(@NotNull Location loc);
++
++    /**
++     * Finds a target {@link Location} for lightning to strike to.
++     * <p>
++     * It looks for:
++     * <ol>
++     * <li>Lightning Rod using {@link World#findLightningRod(Location)}</li>
++     * <li>Random alive {@link LivingEntity} that can see sky
++     * in 10x10x10 cube around input {@link Location} </li>
```
jpenilla: This is incorrect. The AABB used for selecting entities is from the highest motion-blocking block at the given pos to the height limit, then inflated by 3, so is effectively a 6x6x(max height - highest block + 6)

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
```diff
@@ -0,0 +1,47 @@
...
+index dc497606ebf9903c2d32276ecddda3231457346a..0221aa063a048e511821d1e6c985eef0fc61b943 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -757,6 +757,36 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+     @NotNull
+     public LightningStrike strikeLightningEffect(@NotNull Location loc);
+ 
++    // Paper start
++    /**
++     * Finds closest Lightning Rod in 128 block radius around input location.
```
jpenilla: Closest unobstructed Lightning Rod (must be highest non-air block)

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
```diff
@@ -0,0 +1,47 @@
...
++     * Finds a target {@link Location} for lightning to strike to.
++     * <p>
++     * It looks for:
++     * <ol>
++     * <li>Lightning Rod using {@link World#findLightningRod(Location)}</li>
++     * <li>Random alive {@link LivingEntity} that can see sky
++     * in 10x10x10 cube around input {@link Location} </li>
++     * </ol>
++     *
++     * If neither target is found, input {@link Location} will be returned
```
jpenilla: It will return a location with the same x and z as the input, but with the highest motion blocking pos as the y. Does it maybe make sense to return null instead, as we didn't find a "target" (either entity or lightning rod)?

build-data/paper.at
```diff
@@ -236,3 +236,7 @@ public-f com.mojang.brigadier.tree.CommandNode requirement
 # Block Enderpearl Travel Exploit
 public net.minecraft.world.entity.projectile.Projectile cachedOwner
 public net.minecraft.world.entity.projectile.Projectile ownerUUID
+
+# Add methods for lightning target lookups
```
jpenilla: Use the name of the server patch here, ie `Add methods to find targets for lightning strikes`

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
```diff
@@ -0,0 +1,47 @@
...
++     * @param loc Location to search Lightning Rod around
++     * @return Location of Lightning Rod or null if not found
++     */
++    @Nullable
++    public Location findLightningRod(@NotNull Location loc);
++
++    /**
++     * Finds a target {@link Location} for lightning to strike to.
++     * <p>
++     * It looks for:
```
jpenilla: Reword this so it's more clear that it checks these in order

jpenilla (changes requested):

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
dawon: I totally agree with you on this one, but: How should I detect such a case? I see three possible imperfect solutions:

1a) comparing the resulting location's x&z against the source location. This solution risks the case when the target entity/rod was found on these x&z coords -> false negative. 
1b) I could check x&y&z (with finding highest motion blocking y) - I guess same as 1) applies but with less probability? (Motion blocking block == living entities can't go through it? - that would be also a lightning rod, right?
2) I could do the same as 1b) and then check that the resulting location is not a lightning rod. This still is risking an entity on these coordinates.
3) Modifying the original method (splitting it into 2 methods, where one returns null on non found). This should be fine, but is that acceptable by Paper standards? The smallest possible changes are encouraged and this would be quite a big one. Future maintenance of this would be hell...

Or am I overthinking it too much?

jpenilla: Third way would be preferred, and the diff wouldn't really be big or difficult to maintain. (also removes the need for one of the access transformers)

```diff
diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
index 9e4ad810dd..cb697f382f 100644
--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
+++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
@@ -956,6 +956,11 @@ public class ServerLevel extends Level implements WorldGenLevel {
     }
 
     protected BlockPos findLightningTargetAround(BlockPos pos) {
+        // Paper start
+        return this.findLightningTargetAround(pos, false);
+    }
+    public BlockPos findLightningTargetAround(BlockPos pos, boolean returnNullWhenNoTarget) {
+        // Paper end
         BlockPos blockposition1 = this.getHeightmapPos(Heightmap.Types.MOTION_BLOCKING, pos);
         Optional<BlockPos> optional = this.findLightningRod(blockposition1);
 
@@ -970,6 +975,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
             if (!list.isEmpty()) {
                 return ((LivingEntity) list.get(this.random.nextInt(list.size()))).blockPosition();
             } else {
+                if (returnNullWhenNoTarget) return null; // Paper
                 if (blockposition1.getY() == this.getMinBuildHeight() - 1) {
                     blockposition1 = blockposition1.above(2);
                 }
```

dawon: Ah ok, then. Will do, thank you!

patches/api/0335-Add-methods-to-find-targets-for-lightning-strikes.patch
dawon: Is this description sufficient and understandable? Really not sure about the wording: _a random alive {@link LivingEntity} that can see the sky in 6x6 cuboid around input X/Z coordinates. Y is in the range from input Y - 3 to maximum build height + 3_

jpenilla: I think the detail about the entity needing to be alive can be excluded for brevity, as most people would assume dead entities won't be selected anyways. The Y range is from the highest motion-blocking block at the input X/Z - 3 to the height limit + 3, not from input Y - 3 to height limit + 3, see the call to getHeightmapPos.

dawon:
Remarks corrected and rebased...

Machine-Maker (approved):

jpenilla (approved):

PR #6713, opened 2021-10-03
Add event context api

FrankHeijden:
Sometimes, plugins call "fake" bukkit events to determine their outcome, for example to detect whether an area is accessible to player's block placements. Currently, calling such events may negatively affect a plugins state.

This PR aims to provide a way to add context to an event, as a key/value map located inside the `org.bukkit.Event` class. Using the `ignoreContextKey` field added in `@EventHandler`, a plugin may ignore an event altogether if a certain context key is present.

### DIFF
--- a/patches/api/0338-Add-event-context-api.patch
+++ b/patches/api/0338-Add-event-context-api.patch
@@ -1,0 +1,142 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Frank van der Heijden <frank.boekanier@gmail.com>
+Date: Sat, 9 Oct 2021 22:17:17 +0200
+Subject: [PATCH] Add event context api
+
+
+diff --git a/src/main/java/org/bukkit/event/Event.java b/src/main/java/org/bukkit/event/Event.java
+index 8ec56cd6b8e0f5c5dd8c7c88b4671e18dcf109d0..c491564cbde0874923b771d3392422d2f7da7ff6 100644
+--- a/src/main/java/org/bukkit/event/Event.java
++++ b/src/main/java/org/bukkit/event/Event.java
+@@ -3,6 +3,8 @@ package org.bukkit.event;
+ import org.bukkit.plugin.Plugin;
+ import org.bukkit.plugin.PluginManager;
+ import org.jetbrains.annotations.NotNull;
++import java.util.HashMap;
++import java.util.Map;
+ 
+ /**
+  * Represents an event.
+@@ -15,6 +17,7 @@ import org.jetbrains.annotations.NotNull;
+ public abstract class Event {
+     private String name;
+     private final boolean async;
++    private final Map<String, Object> eventContext = new HashMap<>(); // Paper - event context
+ 
+     /**
+      * The default constructor is defined for cleaner code. This constructor
+@@ -66,6 +69,18 @@ public abstract class Event {
+         return name;
+     }
+ 
++    // Paper start - event context
++    /**
++     * Provides the context of this event.
++     *
++     * @return The event context map.
++     */
++    @NotNull
++    public Map<String, Object> getEventContext() {
++        return eventContext;
++    }
++    // Paper end - event context
++
+     @NotNull
+     public abstract HandlerList getHandlers();
+ 
+diff --git a/src/main/java/org/bukkit/event/EventHandler.java b/src/main/java/org/bukkit/event/EventHandler.java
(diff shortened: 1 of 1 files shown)


blablubbabc:
I am trying to understand how exactly this API would resolve the issues around calling fake events.

> Currently, calling such events may negatively affect a plugins state.

Do you have a concrete example for this, and how this API would resolve that?

Usually, if a plugin listens for an event and does _something_ as reaction, I would expect that it is natural for the caller of the fake event to assume that its fake event actually triggers this _something_ in its entirety. Because even with this fake event API, there is currently no way for the caller of the fake event to declare (or even know about) which fine-grained aspects it wants to execute in other plugins, and which aspects it wants to skip.
In the mentioned example of checking region access: Maybe the caller of the fake event wants to not only execute the region access check, but also trigger things like feedback messages for the affected player. In another situation, eg. a different plugin, the plugin might not want to trigger this feedback message sending. But even with this proposed fake event API there is still no way for the caller of the fake event to declare these things (nor am I sure that there is a meaningful way to 'standardize' all possible effects plugins might trigger in reaction to receiving some event inside the Bukkit API).

Same situation on the event receiver side: How would a plugin receiving a fake event know what the purpose of the caller of the fake event is? I.e. which aspects of its event handler it is supposed to execute, and which it should skip in this case? This selection of what to skip seems arbitrary, and the purpose of the fake event caller might change from plugin to plugin.

Even more, what is the purpose for an event handler to completely ignore all fake events from even reaching it, i.e. skipping all of its actions? The only use case I can currently think of might be for some plugin to ignore its own fake events, because it is specifically interested in the effects other plugins have on the called event.
But what if two or more plugins all call fake events, and each wants to ignore its own fake events? The new `ignoreFake` EventHandler option might make it easy for the developers of these plugins to think that they are using this feature correctly, i.e. to ignore their fake events, whereas in fact they might accidentally break compatibility with those other fake event calling plugins without noticing it.

Instead of this simple 'is-fake' API, what about adding something similar to the 'MetaData API' but for events?
I.e. each event could have a reference to an optionally set 'EventContext', which would basically be a Set of String keys, or a Map of String-Object entries, similar to the entity MetaData API. These keys / entries would be provided by the caller of the event. And plugins handling events could define certain String keys to have specific meaning. In the example above, a region protection plugin could define a context key `"silent"`, and a caller of the fake event who knows about this key can specify it in the EventContext of the fake event.
This would allow plugins (callers and receivers of events) to pass additional contextual information between each other, without forming a strong plugin dependency.

FrankHeijden:
@blablubbabc What you mentioned there is indeed a very valid concern. I think the solution you provided is interesting as well, and could definitely work out if we also introduce "standardised" EventContext keys (such as for example `protection_check`). I am interested in what kind of standardised keys we can provide by default, you mentioned `silent`, would you prefer that key name over `protection_check`? I will try to update my PR later today / tomorrow.

blablubbabc:
> if we also introduce "standardised" EventContext keys

I don't think the API can anticipate all possible context keys that would be required to cover all possible needs plugins can come up with. Even if we could come up with some general useful context keys, it would then probably make more sense to just introduce a dedicated event for those cases (eg. a `CheckRegionAccess` event that is explicitly meant for plugin communication). But I doubt that something like this is likely to be added:
* This could also be added by a common plugin that everybody would need to agree to use (similar to Vault). However, even if such an event were to be added directly to the API, one would still need to convience everyone to actually use it. So there is not much gain over a plugin defining this event.
* This would open the door for similar requests by plugin developers, asking for special-purpose events like this, with no suitable rule being available on which events should be added and which should not.

FrankHeijden:
Instead of the fake event api, as proposed by @blablubbabc, a simple event context was added. Updated PR title/description as well.

patches/api/0337-Add-event-context-api.patch
```diff
@@ -0,0 +1,133 @@
...
++    /**
++     * Define if the handler ignores the event if the specified context key is present.
++     * <p>
++     * If ignoreContextKey is set to the empty string, this method is always called.
++     * Otherwise, this method is called if and only if
++     * the specified context key is NOT present in the event context.
++     *
++     * @return whether events containing the specified context key should be ignored
++     */
++    String ignoreContextKey() default "";
```
me4502: If the context is a map, would it make sense for this to be an array so multiple could be ignored?

FrankHeijden: That's fair, then multiple keys may be used in order to ignore. And as a bonus it resolves the "ugly" default as empty string :)

PR #6724, opened 2021-10-05
Saveable Scoreboards

Machine-Maker:
~~Couple problems with this that I don't know the best way to solve. If plugins get/join the returned completable future at the wrong time (like during plugin onEnable) it locks up cause the server is waiting for the plugin to load so the main thread (which at the moment is required for scoreboard creating/registration) can't register the scoreboard.~~

I'm not 100% sure that scoreboard registration needs to be done on the main thread. Would wrapping the WeakCollection of scoreboards in CraftScoreboardManager as a syncronized collection really solve that issue of scoreboard creation requiring the main thread.

EDIT: So I did go ahead and make the scoreboard collection a synchronous collection and removed the async catchers for scoreboard creation.

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -311,3 +311,8 @@ public-f net.minecraft.world.level.chunk.ChunkGenerator strongholdSeed
 
 # More Sculk Sensor API
 public-f net.minecraft.world.level.gameevent.vibrations.VibrationListener listenerRange
+
+# Saveable Scoreboards
+public-f org.bukkit.craftbukkit.scoreboard.CraftScoreboard
+public org.bukkit.craftbukkit.scoreboard.CraftScoreboard <init>(Lnet/minecraft/world/scores/Scoreboard;)V
+public net.minecraft.server.ServerScoreboard server
--- a/patches/api/0371-Saveable-Scoreboards.patch
+++ b/patches/api/0371-Saveable-Scoreboards.patch
@@ -1,0 +1,59 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Mon, 4 Oct 2021 22:20:44 -0700
+Subject: [PATCH] Saveable Scoreboards
+
+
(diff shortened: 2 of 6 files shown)


patches/api/0336-Saveable-Scoreboards.patch
```diff
@@ -0,0 +1,59 @@
...
++++ b/src/main/java/org/bukkit/scoreboard/ScoreboardManager.java
+@@ -29,4 +29,16 @@ public interface ScoreboardManager {
+      */
+     @NotNull
+     Scoreboard getNewScoreboard();
++
++    // Paper start
++
++    /**
++     * Gets a new Scoreboard to be tracked by the server. This
```
Chew: Incomplete?

Machine-Maker:
Rebased for 1.18.1

electronicboy (approved):

patches/server/0830-Saveable-Scoreboards.patch
```diff
@@ -0,0 +1,130 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftScoreboardManager.java b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftScoreboardManager.java
+index 1a3b1eb7b70b9a668aa33ea943c13890eaa23a05..afea73e764b9f12997a8ceabc7933bbcd2a7a212 100644
+--- a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftScoreboardManager.java
++++ b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftScoreboardManager.java
+@@ -25,7 +25,7 @@ import org.bukkit.scoreboard.ScoreboardManager;
+ public final class CraftScoreboardManager implements ScoreboardManager {
+     private final CraftScoreboard mainScoreboard;
+     private final MinecraftServer server;
+-    private final Collection<CraftScoreboard> scoreboards = new WeakCollection<CraftScoreboard>();
++    private final Collection<CraftScoreboard> scoreboards = java.util.Collections.synchronizedCollection(new WeakCollection<CraftScoreboard>()); // Paper
```
electronicboy: if we wanna use a synchronised collection you'd need to wrap the getScores method iteration in a sync block over this

Issue #6718, opened 2021-10-04
updating 1.12.2 region creates endless loop / fills up HD with .backup files


terbin:
### Stack trace

```
[World upgrader thread for world world_the_end #4/WARN]: Backing up regionfile "xxx\world_the_end\DIM1\region\r.-3.-1.mca" to xxx\world_the_end\DIM1\region\r.-3.-1.mca.4295739377041195202.backup
[World upgrader thread for world world_the_end #4/WARN]: Backed up the regionfile to xxx\world_the_end\DIM1\region\r.-3.-1.mca.4295739377041195202.backup
[Server thread/INFO]: [io.papermc.paper.world.ThreadedWorldUpgrader] 98,98% completed (92237 / 93184 chunks)...
[World upgrader thread for world world_the_end #4/INFO]: Starting summary of changes for regionfile xxx\world_the_end\DIM1\region\r.-3.-1.mca
[World upgrader thread for world world_the_end #4/INFO]: End of change summary for regionfile xxx\world_the_end\DIM1\region\r.-3.-1.mca
[World upgrader thread for world world_the_end #4/INFO]: Successfully wrote new header to disk for regionfile xxx\world_the_end\DIM1\region\r.-3.-1.mca
[World upgrader thread for world world_the_end #4/ERROR]: Attempting to read chunk data at [-83, -30] but got chunk data for [45, 34] instead! Attempting regionfile recalculation for regionfile xxx\world_the_end\DIM1\region\r.-3.-1.mca
[World upgrader thread for world world_the_end #4/WARN]: Corrupt regionfile header detected! Attempting to re-calculate header offsets for regionfile xxx\world_the_end\DIM1\region\r.-3.-1.mca
java.lang.Throwable: null
	at net.minecraft.world.level.chunk.storage.RegionFile.recalculateHeader(RegionFile.java:137) ~[patched_1.17.1.jar:git-Paper-302]
	at net.minecraft.world.level.chunk.storage.RegionFileStorage.read(RegionFileStorage.java:219) ~[app:?]
	at net.minecraft.world.level.chunk.storage.RegionFileStorage.read(RegionFileStorage.java:221) ~[app:?]
	at net.minecraft.world.level.chunk.storage.RegionFileStorage.read(RegionFileStorage.java:221) ~[app:?]
	at net.minecraft.world.level.chunk.storage.RegionFileStorage.read(RegionFileStorage.java:221) ~[app:?]
[...]

PR #6725, opened 2021-10-05
Use file hash in region backup name

Prof-Bloodstone:
In case there's an error when recalculating region header, backups won't fill the FS with endless, duplicated files

Ref: #6718


kennytv (approved):

electronicboy (approved):

Machine-Maker:
Is this still needed after 3391ccf988b51fe0e97be82a0844c16a5c2e1da5?

Prof-Bloodstone:
Is it necessary - no. It's just a safety fallback that if another bug pops up in the future (after update?), this won't fill the disk and potentially bring down other services running on the same machine, but rather simply error out without spamming the FS.
I haven't looked at the logic when it decides to not re-read the regionfile if there was an issue, perhaps it'd minimize the chance of this happening well enough.

PR #6726, opened 2021-10-05
Allow components in conversation prompts

### LINKED ISSUE #6644: Possibility to use adventure component in Prompt#getPromptText()

When using the conversation api and dealing with prompts, it would be nice to be able to use the adventure component system to style the prompt text.

Being able to alternative provide a component instead of a string when setting the prompt text.

Using LegacyComponentSerializer#legacySection() to serialize the component into a string.

### Agreements

### Other

_No response_

kumpelblase2:
This allows prompts to alternatively return a text component, implementing #6644.

I'm not entirely happy with how it is right now, but the alternative I've explored also didn't sit well with me. Specifically, the `Prompt` now doesn't force a user to implement a method of getting something to display. Both methods are default, so there's no hard requirement and it will just compile fine without overwriting either of them and eventually crash at runtime. Unfortunately, making either of them non-default would be unsatisfactory in my eyes.
- Making `getPromptMessage` (the new API) non-default: Breaks every prompt in existence, so not an option.
- Making `getPromptText` (the existing API) non-default: Any prompts that want to return a component directly now have to deal with a dead method.

The alternative that I explored was creating a new prompt interface, say `ComponentPrompt` that would include the new method and provide a default for the `getPromptText` method.
<details>
  <summary>ComponentPrompt</summary>

  ```java
public interface ComponentPrompt extends Prompt {

    @Override
    @NotNull
    default String getPromptText(@NotNull ConversationContext context) {
        throw new UnsupportedOperationException("Use getPromptMessage() for component prompts.");
    }

    /**
     * Gets the message component to display to the user when this prompt is first
     * presented.
     *
     * @param context Context information about the conversation.
     * @return The chat component to display.
     */
    @NotNull Component getPromptMessage(ConversationContext context);
}
  ```
</details>

This would ensure old prompts won't break as they wouldn't be forced to implement this new interface. It would also force users of both prompt types (the old and new) to implement the correct method for getting the prompt text. I think this was more satisfactory from a user perspective, but this would now require instance checking internally when using the prompts. As a result I found this solution also a little unsatisfactory. But maybe the upside of having a clear guide on what to implement is more important?

Unfortunately, adding `sendMessage(Component)` to the `Conversible` causes a clash with the equivalent method in `Audience` - to some degree on purpose. Since it was defined as default there, it had to be overwritten in both `Player` and `ConsoleCommandSender`. Not particularly happy about this either, but creating yet another new method for this kind of thing felt awful. Overriding with the default implementation felt like a lesser evil.

I was also debating allowing the prefix to be a component too while I was at it. It would effectively have the same implementation challenge as the prompt so decided to leave that aside for now and see where we're going with the prompt implementation.

I acknowledge that this is a small and low priority change but I'd still be appreciative of feedback.

### DIFF
--- a/patches/api/0336-Allow-components-in-conversation-prompts.patch
+++ b/patches/api/0336-Allow-components-in-conversation-prompts.patch
@@ -1,0 +1,107 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tim Hagemann <tim@eternalwings.de>
+Date: Tue, 5 Oct 2021 20:49:08 +0200
+Subject: [PATCH] Allow components in conversation prompts
+
+
+diff --git a/src/main/java/org/bukkit/command/ConsoleCommandSender.java b/src/main/java/org/bukkit/command/ConsoleCommandSender.java
+index f309c2ed39bcc700e5b491ff549bf0608a07aab8..704ecebef291a9e8477eab5e0d2b30aeff463f63 100644
+--- a/src/main/java/org/bukkit/command/ConsoleCommandSender.java
++++ b/src/main/java/org/bukkit/command/ConsoleCommandSender.java
+@@ -3,4 +3,10 @@ package org.bukkit.command;
+ import org.bukkit.conversations.Conversable;
+ 
+ public interface ConsoleCommandSender extends CommandSender, Conversable {
++    // Paper start - Allow components in conversation prompts
++    @Override
++    default void sendMessage(@org.jetbrains.annotations.NotNull net.kyori.adventure.text.Component message) {
++        CommandSender.super.sendMessage(message);
++    }
++    // Paper end
+ }
+diff --git a/src/main/java/org/bukkit/conversations/Conversable.java b/src/main/java/org/bukkit/conversations/Conversable.java
+index b7d8dd30360a38dbdc7bbce40c8e6ced7261f833..14e9e75b622de0e3ee8ca92a75a923714789b09f 100644
+--- a/src/main/java/org/bukkit/conversations/Conversable.java
++++ b/src/main/java/org/bukkit/conversations/Conversable.java
+@@ -57,6 +57,8 @@ public interface Conversable {
+      */
+     public void sendRawMessage(@NotNull String message);
+ 
++    public void sendMessage(@NotNull net.kyori.adventure.text.Component component); // Paper - Allow components in conversation prompts
++
+     /**
+      * Sends this sender a message raw
+      *
+diff --git a/src/main/java/org/bukkit/conversations/Conversation.java b/src/main/java/org/bukkit/conversations/Conversation.java
+index bf2407c838bc20197802687c150d513f4e86ed2b..d909b14fd2ccf6704e0611e13f619c1331424a2d 100644
+--- a/src/main/java/org/bukkit/conversations/Conversation.java
++++ b/src/main/java/org/bukkit/conversations/Conversation.java
+@@ -287,7 +287,11 @@ public class Conversation {
+         if (currentPrompt == null) {
+             abandon(new ConversationAbandonedEvent(this));
+         } else {
+-            context.getForWhom().sendRawMessage(prefix.getPrefix(context) + currentPrompt.getPromptText(context));
++            // Paper start - Allow components in conversation prompts
++            final net.kyori.adventure.text.Component prefixComp = net.kyori.adventure.text.Component.text(prefix.getPrefix(context));
++            final net.kyori.adventure.text.Component message = currentPrompt.getPromptMessage(context);
++            context.getForWhom().sendMessage(prefixComp.append(message));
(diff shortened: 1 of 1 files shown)


patches/api/0336-Allow-components-in-conversation-prompts.patch
```diff
@@ -0,0 +1,107 @@
...
+--- a/src/test/java/org/bukkit/conversations/FakeConversable.java
++++ b/src/test/java/org/bukkit/conversations/FakeConversable.java
+@@ -51,6 +51,13 @@ public class FakeConversable implements Conversable {
+         lastSentMessage = message;
+     }
+ 
++    // Paper end - Allow components in conversation prompts
++    @Override
++    public void sendMessage(@NotNull net.kyori.adventure.text.Component component) {
++        lastSentMessage = ((net.kyori.adventure.text.TextComponent)component).content();
```
MiniDigger, line 101: I guess this is good enough for a test but really, a serialisier should be used in situations like this (plain or legacy I guess)

MiniDigger:
I don't have an IDE right now, nor do I have the whole class hierarchy in my mind, but couldn't you avoid the issue with sendMessage by making conversable an audience, just like command sender? then the clash goes avoid, right? 

generally, for adventure stuff, we omitted the get prefix from the getters, I guess this should apply here too?

MiniDigger:
retried the pipeline because it got stuck, now you got failed tests
https://github.com/PaperMC/Paper/pull/6726/checks?check_run_id=3814340093#step:5:1729

patches/api/0336-Allow-components-in-conversation-prompts.patch
kumpelblase2: Fair enough, I just wanted the most simple thing here given that that no test actually gets to this spot. Would have to resort to referencing that directly though, since calling `plainSerializer()` would fail here. When I do that, I'll expand the tests while I'm at it.

kumpelblase2:
I considered that option, but I did not go that way because I didn't think it would make much sense to have functionality like `showBossBar`, `showTitle`, etc. be defined for a `Conversable`. It would work around that issue, that is correct. I guess it's not much of an issue given that everything has a default implementation and both console command sender & player already implement audience anyway.

> generally, for adventure stuff, we omitted the get prefix from the getters, I guess this should apply here too?

Alright, will update accordingly.

MiniDigger:
oh right, those are defined on audience too, yeah, fair enough then

patches/api/0336-Allow-components-in-conversation-prompts.patch
```diff
@@ -0,0 +1,107 @@
...
+index bf2407c838bc20197802687c150d513f4e86ed2b..d909b14fd2ccf6704e0611e13f619c1331424a2d 100644
+--- a/src/main/java/org/bukkit/conversations/Conversation.java
++++ b/src/main/java/org/bukkit/conversations/Conversation.java
+@@ -287,7 +287,11 @@ public class Conversation {
+         if (currentPrompt == null) {
+             abandon(new ConversationAbandonedEvent(this));
+         } else {
+-            context.getForWhom().sendRawMessage(prefix.getPrefix(context) + currentPrompt.getPromptText(context));
++            // Paper start - Allow components in conversation prompts
++            final net.kyori.adventure.text.Component prefixComp = net.kyori.adventure.text.Component.text(prefix.getPrefix(context));
```
jpenilla, line 45: should the prefix be legacy deserialized? also, as far as a solution for component prefixes, you can extend ConversationPrefix and provide a default that just throws for the legacy method, adding a new Component returning method, then in the implementation check whether the ConversationPrefix is a ComponentConversationPrefix

jpenilla:
> I considered that option, but I did not go that way because I didn't think it would make much sense to have functionality like `showBossBar`, `showTitle`, etc. be defined for a `Conversable`. It would work around that issue, that is correct. I guess it's not much of an issue given that everything has a default implementation and both console command sender & player already implement audience anyway.
> 
> > generally, for adventure stuff, we omitted the get prefix from the getters, I guess this should apply here too?
> 
> Alright, will update accordingly.

I think extending Audience is a better solution, all Conversables are also Audiences as is, and I don't see that changing really.

Owen1212055:
What do we even want to do with the conversation api? I’m honestly surprised people even use it. 

Is this something we want to add onto and upkeep? Or do we just want to push this api to the side here?

olijeffers0n:
I used it once, it was honestly kind of a pain to use and would probably be easier to implement myself just using chat events. I would be happy to just deprecate and move on from it.

Owen1212055:
Seeing as though there hasn't been really much push for this, and seeing that this API generally hasn't aged well and is most likely better being phased out this will be closed.

This API looks to be full of legacy string usage in general, and although this is kind of a workaround in order to allow modern components it's probably best to just encourage people to move away from this entirely. 

Thank you for your contribution, regardless. 😄

PR #6733, opened 2021-10-06
Remove old reduce debug info patch

Owen1212055:
This would resend the value when the client joined. It seemed that this was fixed a while ago, as removing this doesn't seem to change anything


Machine-Maker (approved):
Looking into how the client handles this now, it looks to be set in the ClientboundLoginPacket and the same thing would happen when sending the EntityEvent packet right after. So I think it is def redundant and can be removed.

Owen1212055:
Rebased!

Owen1212055:
Can someone remove this patch when dealing with 1.18?

kennytv:
Done, thanks :p

PR #6734, opened 2021-10-06
Add some player gamerule overrides

Owen1212055:
Allows you to override some of the gamerules per player. Ideally, this patch can be expanded on in the future to include more overrides for certain values. Changes have been tested and seem to work just fine.

This PR is heavily inspired by the fix/persistence branch.
Requires #6733 (as my PR doesn't cover this condition)


Owen1212055:
Rebased, additionally, a lot of the ugly conditions can be removed when adventure is updated.

https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/util/TriState.java
```java
public boolean toBooleanOrElse(final boolean other) {
    switch (this) {
      case TRUE: return true;
      case FALSE: return false;
      default: return other;
    }
  }
```

Owen1212055:
Any feedback on this would be appreciated :)

Some of these paper comments would have to be cleaned up, this is an older PR and I have to cleanup the style.

Owen1212055:
Cleaned up the diff significantly thanks to utilities introduced in the new adventure update.

+ rebased

NoahvdAa (changes requested):
This would cause a lot of methods to be added to `Player` as this API expands. Maybe these individual methods could be replaced with a getter/setter that takes a gamerule. Whether a gamerule is override-able could be determined via some kind of override-able parameter on the GameRule class, or maybe a new interface (OverrideableGameRule)?

Machine-Maker:
I like what Noah said about having 1 method take a game rule to override.

Owen1212055:
I've decided not to go through with this and instead just opt for using the new player effect api.

Owen1212055:
Nvm

PR #6736, opened 2021-10-07
Collision API

Owen1212055:
This gives you the ability to check for collisions for an entity or a bounding box in a world. 
This allows you to see for example if an entity can fit at a certain location.

Feedback appreciated.


patches/api/0337-Collision-API.patch
```diff
@@ -0,0 +1,67 @@
...
++     *
++     * Checks for any collisions in the given boundingbox.
++     * This will check for any colliding entities / worldborder / blocks.
++     *
++     * @param boundingBox the box to check collisions in
++     * @param loadChunks should load chunks within bounding box
++     * @return collides or not
++     */
++    boolean collidesAt(@NotNull BoundingBox boundingBox, boolean loadChunks);
++    // Paper End - Collision API
```
NoahvdAa: Put this in the existing paper block instead (line 37 of this patch)

patches/server/0819-Collision-API.patch
```diff
@@ -0,0 +1,41 @@
...
+     }
+     // Paper end
++    // Paper Start - Collision API
++    @Override
++    public boolean collidesAt(@org.jetbrains.annotations.NotNull BoundingBox boundingBox, boolean loadChunks) {
++        net.minecraft.world.phys.AABB aabb = new AABB(boundingBox.getMinX(), boundingBox.getMinY(), boundingBox.getMinZ(), boundingBox.getMaxX(), boundingBox.getMaxY(), boundingBox.getMaxZ(), false);
++
++        return !getHandle().level.noCollision(getHandle(), aabb);
++    }
++    // Paper End - Collision API
```
NoahvdAa: Same here, put this in the existing paper block.

patches/server/0819-Collision-API.patch
```diff
@@ -0,0 +1,41 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
+@@ -894,4 +894,13 @@ public abstract class CraftRegionAccessor implements RegionAccessor {
+ 
+         throw new IllegalArgumentException("Cannot spawn an entity for " + clazz.getName());
+     }
++
++    // Paper Start - Collision API
++    @Override
++    public boolean collidesWith(@org.jetbrains.annotations.NotNull org.bukkit.util.BoundingBox boundingBox) {
++        net.minecraft.world.phys.AABB aabb = new AABB(boundingBox.getMinX(), boundingBox.getMinY(), boundingBox.getMinZ(), boundingBox.getMaxX(), boundingBox.getMaxY(), boundingBox.getMaxZ(), false);
```
NoahvdAa: Maybe null check the bounding box? Something like
```java
Preconditions.checkArgument(boundingBox != null, "boundingBox cannot be null");
```

patches/server/0819-Collision-API.patch
```diff
@@ -0,0 +1,41 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
+@@ -1264,4 +1264,12 @@ public abstract class CraftEntity implements org.bukkit.entity.Entity {
+         return getHandle().isTicking();
+     }
+     // Paper end
++    // Paper Start - Collision API
++    @Override
++    public boolean collidesAt(@org.jetbrains.annotations.NotNull BoundingBox boundingBox, boolean loadChunks) {
++        net.minecraft.world.phys.AABB aabb = new AABB(boundingBox.getMinX(), boundingBox.getMinY(), boundingBox.getMinZ(), boundingBox.getMaxX(), boundingBox.getMaxY(), boundingBox.getMaxZ(), false);
```
NoahvdAa: Same null check here.

Owen1212055:
Removed the chunk loading param and moved some more of the implementation into the craft classes. This should resolve the error with the check collisions at the location method. (The location was offset)

patches/server/0844-Collision-API.patch
```diff
@@ -0,0 +1,48 @@
...
++    @Override
++    public boolean collidesAt(@org.jetbrains.annotations.NotNull Location location) {
++        net.minecraft.world.phys.AABB aabb = this.getHandle().getBoundingBoxAt(location.getX(), location.getY(), location.getZ());
++
++        return !this.getHandle().level.noCollision(this.getHandle(), aabb);
++    }
++
++    @Override
++    public boolean collidesAt(@org.jetbrains.annotations.NotNull BoundingBox boundingBox) {
++        net.minecraft.world.phys.AABB aabb = new AABB(boundingBox.getMinX(), boundingBox.getMinY(), boundingBox.getMinZ(), boundingBox.getMaxX(), boundingBox.getMaxY(), boundingBox.getMaxZ(), false);
```
lynxplay: In regards to repeated conversion between AABB and BoundingBox, the MCUtil utility class does not yet seem to have a neat conversion method for this.

Might be worth to add one [into this patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0006-MC-Utils.patch) and use that to convert from/to an AABB.

Incredibly minor nitpick tho, so really up to you.

patches/api/0347-Collision-API.patch
```diff
@@ -0,0 +1,62 @@
...
+      */
+     @NotNull
+     public <T extends Entity> T spawn(@NotNull Location location, @NotNull Class<T> clazz, boolean randomizeData, @Nullable Consumer<T> function) throws IllegalArgumentException;
++
++    // Paper Start - Collision API
++
++    /**
++     *
++     * Checks for any collisions in the given boundingbox.
++     * This will check for any colliding hard entities / worldborder / blocks.
```
lynxplay: What exactly are "hard entities" ? Might need clarification that this mainly wraps boats, shulkers and minecarts (as of right now).

patches/server/0844-Collision-API.patch
Owen1212055: Decided against this in the end, as really this constructor is only needed here once anyways... So IDK, and the constructor used in question is added in a later patch.

patches/api/0347-Collision-API.patch
Owen1212055: Gave some examples!

Owen1212055:
Rebased

Owen1212055:
Rebased

kennytv (approved):

Owen1212055:
Rebased and fixed documentation, I appreciate any last feedback but functionality seems to be fine.

Owen1212055:
Rebased

kennytv (approved):

Owen1212055:
Rebased

Owen1212055:
Rebased

electronicboy (approved):

Owen1212055:
image
A useful diagram that may or may not help people.

PR #6737, opened 2021-10-07
Deprecate API methods added by 'Close Plugin Class Loaders on Disable'

jpenilla:
Upstream always closes the PluginClassLoader now


electronicboy (approved):

MiniDigger:
for the future, is there a usecase to have like one giant "deprecated apis go here" patch?

PR #6748, opened 2021-10-08
Filter ip address on join if setting is true

### LINKED ISSUE #6746: Disabling IP address logging still logs it in certain messages
### Expected behavior

The player's IP address should be censored in the "logged in with entity id" message.

### Observed/Actual behavior

Even with ``settings.log-player-ip-addresses`` set to ``false``, the player's IP address is still logged in the "logged in with entity id" message when they log into the server.

### Steps/models to reproduce

1): Change ``settings.log-player-ip-addresses`` to ``false``.
2): Join the server and observe the "logged in with entity id" message.

### Plugin and Datapack List

None.

### Paper version

This server is running Paper version git-Paper-318 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 0d1e187)

### Other

_No response_

Machine-Maker:
Fixes #6746


electronicboy (approved):

NoahvdAa (approved):
Good catch, managed to miss that one 😅

PR #6751, opened 2021-10-08
Fix cancelled snow bucket placement

### LINKED ISSUE #6630: Cancelling BlockPlaceEvent for powdered snow keeps grass snowy
### Expected behavior

The grass to not turn snowy.

### Observed/Actual behavior

The grass turned snowy, persists after relogging/restarting the server.

### Steps/models to reproduce

Reproducable using the following code:
```java
@EventHandler
public void onBlockPlace(BlockPlaceEvent event) {
	if (event.getBlock().getType().equals(Material.POWDER_SNOW)) {
		event.setBuild(false);
		event.setCancelled(true);
	}
}
```
After placing powdered snow, the grass will still be snowy (snowy: true in f3). I've confirmed this isn't just some visual/client sided thing by relogging/restarting the server.

### Plugin list

Reproducable without any plugins (except one to cancel the BlockPlaceEvent for powdered snow)

### Paper version

This server is running Paper version git-Paper-266 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 0d19cd7)
Previous version: git-Paper-"4d9b14e" (MC: 1.17.1)

### Agreements

### Other

_No response_

Machine-Maker:
SolidBucketItems need to use the existing capturedBlockStates, but have their block place event handled elsewhere, so the event called in ItemStack needs to be skipped.

Fixes #6630

### DIFF
--- a/patches/server/0865-Fix-cancelled-powdered-snow-bucket-placement.patch
+++ b/patches/server/0865-Fix-cancelled-powdered-snow-bucket-placement.patch
@@ -1,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Fri, 8 Oct 2021 13:12:58 -0700
+Subject: [PATCH] Fix cancelled powdered snow bucket placement
+
+Cancelling the placement of powdered snow from the powdered
+snow bucket didn't revert grass that became snowy because of the
+placement.
+
+diff --git a/src/main/java/net/minecraft/world/item/BlockItem.java b/src/main/java/net/minecraft/world/item/BlockItem.java
+index 893d5bf448ddbccb30db0ee751c7f4a4e83634b9..8f3b9b8784f0d7b137a1ad87655ee8bad801b59d 100644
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Rebased for 1.18

lerokko:
I am unfamiliar with the merging approval procedure, forgive me if this is the wrong place to ask. But this bug is one of 2 that keeps me from updating. Is there any eta on when or if this is merged at all?

Owen1212055 (changes requested):
It seems like your fix causes an unintended desync, where ghost blocks will appear but the server will be fine. I'm guessing it works fine if you are targetting the affected block since it's resent to the client late.

https://user-images.githubusercontent.com/23108066/147372065-fcd05d08-9e72-4a92-b2b4-10cdea1be65c.mp4

Machine-Maker:
Rebased.

And @Owen1212055 I think I fixed that issue. I remember when I originally made my PR thinking that I must be missing smth, but I did only test with direct placing, and so the client is updated to the clicked on block not changing.

Owen1212055 (approved):
Tested fix, works!

electronicboy (approved):

NoahvdAa (approved):
Works!

Rollczi (approved):

PR #6753, opened 2021-10-09
Updated Upstream (CraftBukkit/Spigot)

Machine-Maker:

CraftBukkit Changes:
960f3109 PR-937: Fixes related to unplaced BlockStates and the recent world generation changes.
4faf479e SPIGOT-6754: We ignore any still present TileEntity now when we create a BlockState for a block of type AIR.
a72d5404 SPIGOT-6754: Temporarily restore previous behaviour for tile entities with removed blocks

Spigot Changes:
dc75aca8 Remove redundant 'Log null TileEntity Owner' patch


Machine-Maker:
Done in 42760138335574ae20c549a6ecb13b29d20b6d20

See https://github.com/PaperMC/Paper/pull/6780

PR #6761, opened 2021-10-11
Fix HumanEntity#drop not updating the client inventory

### LINKED ISSUE #6760: Player::dropItem not work like expected
### Expected behavior

When use the method `player.dropItem(true)` the player drop the item (all stack) in hand.

### Observed/Actual behavior

When use the method the item drop but still show in hotbar, but if you click the item in hotbar then disapear.

### Steps/models to reproduce

Using player.dropItem in any cases.

### Plugin and Datapack List

Plugins: ConterCraft, CoreProtect, FastAsyncWorldEdit (WorldEdit), HackedCore*, HackedServer*, LibsDisguises, LuckPerms, PlugMan, ProtocolLib, SentryIntegration, spark

DataPacks: [vanilla (built-in)], [file/bukkit (world)], [file/VoidDimension-v1.2.zip (world)], [file/InfectionAdvancements-v.1.zip (world)], [file/FarmingLandDimension-v1.0.zip (world)], [file/BunnyBurrowDimension-v1.0.zip (world)]

### Paper version

Paper version git-Paper-325 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 4a0d935)

### Other

This working fine in 1.16.5, in the last version need use updateInventory for avoid glitch.

Machine-Maker:
Fixes #6760

### DIFF
--- a/patches/server/0965-Fix-HumanEntity-drop-not-updating-the-client-inv.patch
+++ b/patches/server/0965-Fix-HumanEntity-drop-not-updating-the-client-inv.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 10 Oct 2021 18:18:01 -0700
+Subject: [PATCH] Fix HumanEntity#drop not updating the client inv
+
+== AT ==
+public net.minecraft.server.level.ServerPlayer containerSynchronizer
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
+index 11a06cd0a9cee01107941a2294fbcb493a11ab63..e54574c60c5cbc4e083c4ca11e0b7dd49bd03478 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
+@@ -749,8 +749,15 @@ public class CraftHumanEntity extends CraftLivingEntity implements HumanEntity {
+     // Paper end
+     @Override
+     public boolean dropItem(boolean dropAll) {
+-        if (!(this.getHandle() instanceof ServerPlayer)) return false;
+-        return ((ServerPlayer) this.getHandle()).drop(dropAll);
++        // Paper start - notify client of remote slot change
++        if (!(this.getHandle() instanceof ServerPlayer player)) return false;
++        boolean success = player.drop(dropAll);
++        if (!success) return false;
++        final net.minecraft.world.entity.player.Inventory inv = player.getInventory();
++        final java.util.OptionalInt optionalSlot = player.containerMenu.findSlot(inv, inv.selected);
++        optionalSlot.ifPresent(slot -> player.containerSynchronizer.sendSlotChange(player.containerMenu, slot, inv.getSelected()));
++        return true;
++        // Paper end
+     }
+ 
+     @Override


patches/server/0826-Fix-HumanEntity-drop-not-updating-the-client-inv.patch
```diff
@@ -0,0 +1,18 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 11f46c1b8f4c8414e0667d1873542c17d6e01f2a..93eb8f8d1d535c1d64a4580eea3ca42a1d0c7673 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -2273,6 +2273,7 @@ public class ServerPlayer extends Player {
+ 
+         this.containerMenu.findSlot(playerinventory, playerinventory.selected).ifPresent((i) -> {
+             this.containerMenu.setRemoteSlot(i, playerinventory.getSelected());
++            this.containerSynchronizer.sendSlotChange(this.containerMenu, i, playerinventory.getSelected()); // Paper - notify client of remote slot change
```
kennytv: This results in unnecessary updates in case of even cancellation (and for any call that isn't from the HumanEntity#dropItem method), you should move it up to the actual cb method

kennytv (dismissed):

Machine-Maker:
Rebased for 1.18.1

Owen1212055 (approved):
Fixes the issue, tested...

NoahvdAa (approved):

lynxplay (commented):
The AT should be moved into the patch before merge, presumably our new maintainer can do that while merging tho :)

patches/server/0878-Fix-HumanEntity-drop-not-updating-the-client-inv.patch
```diff
@@ -0,0 +1,27 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
+@@ -715,8 +715,14 @@ public class CraftHumanEntity extends CraftLivingEntity implements HumanEntity {
+     // Paper end
+     @Override
+     public boolean dropItem(boolean dropAll) {
+-        if (!(this.getHandle() instanceof ServerPlayer)) return false;
+-        return ((ServerPlayer) this.getHandle()).drop(dropAll);
++        // Paper start - notify client of remote slot change
++        if (!(this.getHandle() instanceof ServerPlayer player)) return false;
++        boolean success = player.drop(dropAll);
```
lynxplay: What is preventing us from technically returning early here ?
`success` should be false only if no item was dropped whatsoever, so we could skip some logic here unless I am missing something.

Machine-Maker: That's the purpose of the PR. To fix the desync. The server can cancel the drop when the client expects it to drop, so we need to resync.

Machine-Maker:
Rebased for 1.19.3

patches/server/0878-Fix-HumanEntity-drop-not-updating-the-client-inv.patch
lynxplay: Was it ? At least in my testing the issue was that the client specifically does *not* expect an item drop and hence
does not update the inventory.

> When use the method the item drop but still show in hotbar, but if you click the item in hotbar then disapear.
The issue seems to match that.

Machine-Maker: Yeah, that sounds right. I guess I totally forgot what this was about. The cancellation logic for PlayerDropItemEvent handles updating the inventory.

lynxplay (approved):

PR #6275, opened 2021-07-26
Add critical damage API

MagicCheese1:
Previously, if a plugin listened to EntityDamageEvent for player hits and wanted to figure out if the hit was critical, they had to check for stuff like if the player was falling etc.
I have added EntityDamageEvent#IsCritical() that can be used instead. For regular player hits, it uses the flag in the player's attack code. If the damager was an AbstractArrow it returns whatever AbstractArrow#isCritical() returns.
Closes #6752

PR #6762, opened 2021-10-11
Add isCritical to EntityDamageByEntityEvent

### LINKED ISSUE #6752: Add a PlayerCriticalDamageEvent

There's no built-in method to check if a player's damage is critical.
 
Here's an example of what will the event look like
```
@EventHandler
public void onCritical(PlayerCriticalDamageEvent event) {
     if(event.isCritical) {
          event.getPlayer().sendMessage("Critical Damage!!!");
     }
}
```
The event should be cancellable and have ```event.getPlayer()``` ,  ```event.getDamage()``` , ```event.setCritical()``` , and ```event.setDamage()```.

I tried making a method like this 
```
public boolean isCritical(Player player) {
     return !player.isOnGround
      && !player.hasPotionEffect(PotionType.BLINDNESS);
}
```
But sometimes it doesn't work or it's not very accurate.

### Other
This event will certainly be handy in some plugins like RPG-Based plugins when you do critical damage it will say "Crit!"

Machine-Maker:
Closes #6752

### DIFF
--- a/patches/api/0338-Add-isCritical-to-EntityDamageByEntityEvent.patch
+++ b/patches/api/0338-Add-isCritical-to-EntityDamageByEntityEvent.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 10 Oct 2021 18:55:04 -0700
+Subject: [PATCH] Add isCritical to EntityDamageByEntityEvent
+
+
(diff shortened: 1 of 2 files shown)


imDaniX:
What about #6275?

Machine-Maker:
TBH, I didn’t realize that existed. There are a couple problems with it. I think the isCritical should only be in the EntityDamageByEntity event, not the EntityDamageByBlock event or the EntityDamageEvent base. 

But idrc, whatever people think is better.

kennytv:
Put a "conflict on change" comment on the critical/flag2 boolean

Machine-Maker:
Gonna close this cause the other PR is close enough. Ill comment my suggestions over there, not looking to take credit for smth someone else pretty much already did

PR #6765, opened 2021-10-11
Strip raytracing for entity line of sight

PaulBGD:
The IBlockAccess#rayTrace method is very wasteful in both allocations,
and in logic. While EntityLiving#hasLineOfSight provides static
parameters for collisions with blocks and fluids, the method still does
a lot of dynamic checks for both of these, which result in extra work.
As well, since the fluid collision option is set to NONE, the entire
fluid collision system is completely unneeded, yet used anyways.

This is also used for fast line of sight merging in ItemEntity, as it
follows the same rules as hasLineOfSight.

### DIFF
--- a/patches/server/0826-Strip-raytracing-for-entity-line-of-sight.patch
+++ b/patches/server/0826-Strip-raytracing-for-entity-line-of-sight.patch
@@ -1,0 +1,162 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Paul Sauve <paul@technove.co>
+Date: Sat, 31 Oct 2020 18:43:02 -0500
+Subject: [PATCH] Strip raytracing for entity line of sight
+
+The IBlockAccess#rayTrace method is very wasteful in both allocations,
+and in logic. While EntityLiving#hasLineOfSight provides static
+parameters for collisions with blocks and fluids, the method still does
+a lot of dynamic checks for both of these, which result in extra work.
+As well, since the fluid collision option is set to NONE, the entire
+fluid collision system is completely unneeded, yet used anyways.
+
+This is also used for fast line of sight merging in ItemEntity, as it
+follows the same rules as hasLineOfSight.
+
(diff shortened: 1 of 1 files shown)


patches/server/0826-Strip-raytracing-for-entity-line-of-sight.patch
```diff
@@ -0,0 +1,162 @@
...
+index 6200a8ab4f7b2c40e7139cfb90a62f42c5828de2..3af837135a2b0f43992067d360640a125fdd4b31 100644
+--- a/src/main/java/net/minecraft/world/level/BlockGetter.java
++++ b/src/main/java/net/minecraft/world/level/BlockGetter.java
+@@ -73,6 +73,16 @@ public interface BlockGetter extends LevelHeightAccessor {
+         });
+     }
+ 
++    // Paper start - broken down variant of below rayTraceBlock, used by World#rayTraceDirect
++    default net.minecraft.world.phys.BlockHitResult.Type rayTraceBlockDirect(Vec3 vec3d, Vec3 vec3d1, BlockPos blockposition, BlockState iblockdata, net.minecraft.world.phys.shapes.CollisionContext voxelshapecoll) {
++        if (iblockdata.isAir()) return null; // Tuinity - optimise air cases
```
lynxplay: The `// Tunity` comment probably slipped through :+1: should be `// Paper` or, as all this is in a paper comment block, just a comment maybe ?

lynxplay (changes requested):

patches/server/0826-Strip-raytracing-for-entity-line-of-sight.patch
PaulBGD: Whoops, thought I got that removed. It was copied from an opt added to the original `rayTraceBlockDirect` by leaf, I'll just remove the comment now that this is all squished down.

PaulBGD:
Bump

Spottedleaf:
This looks fine

PaulBGD:
Closing as it'd have to be rebased/redone for 1.18 anyways.

PR #6766, opened 2021-10-11
Optimize random calls in chunk ticking

PaulBGD:
Especially at over 30,000 chunks these random calls are fairly heavy. We
use a different method here for checking lightning, and for checking
ice.

Lightning: Each chunk now keeps an int of how many ticks until the
lightning should strike. This int is a random number from 0 to 100000 * 2,
the multiplication is required to keep the probability the same.

Ice and snow: We just generate a single random number 0-16 and increment
it, while checking if it's 0 for the current chunk.

Depending on configuration for things that tick in a chunk, this is a
5-10% improvement.

### DIFF
--- a/patches/server/0826-Optimize-random-calls-in-chunk-ticking.patch
+++ b/patches/server/0826-Optimize-random-calls-in-chunk-ticking.patch
@@ -1,0 +1,92 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Paul Sauve <paul@technove.co>
+Date: Thu, 7 Jan 2021 11:49:36 -0600
+Subject: [PATCH] Optimize random calls in chunk ticking
+
+Especially at over 30,000 chunks these random calls are fairly heavy. We
+use a different method here for checking lightning, and for checking
+ice.
+
+Lightning: Each chunk now keeps an int of how many ticks until the
+lightning should strike. This int is a random number from 0 to 100000 * 2,
+the multiplication is required to keep the probability the same.
+
+Ice and snow: We just generate a single random number 0-16 and increment
+it, while checking if it's 0 for the current chunk.
+
+Depending on configuration for things that tick in a chunk, this is a
+5-10% improvement.
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerChunkCache.java b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
+index 7470f3ba66c2e894b5a5b0ba392ecabf8b04aff9..7dfdd945df176b8e559b116b67954f4b3432c4e2 100644
+--- a/src/main/java/net/minecraft/server/level/ServerChunkCache.java
++++ b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
+@@ -983,6 +983,7 @@ public class ServerChunkCache extends ChunkSource {
+             }
+             // Paper end - optimize isOutisdeRange
+             this.level.getProfiler().push("pollingChunks");
++            this.level.resetIceAndSnowTick(); // Paper - reset ice & snow tick random
+             int k = this.level.getGameRules().getInt(GameRules.RULE_RANDOMTICKING);
+             boolean flag2 = level.ticksPerAnimalSpawns != 0L && worlddata.getGameTime() % level.ticksPerAnimalSpawns == 0L; // CraftBukkit
+ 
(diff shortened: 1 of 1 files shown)


MiniDigger (approved):
lgtm

PaulBGD:
Bump

Spottedleaf:
The random tick call is very complicated in nature to optimise, mostly because of how hard it is to even benchmark. 
From my previous experience with benchmarking, the random tick call is approximately 1us. So when I looked at this patch, I have significant doubt that the random.nextInt call was expensive. But I'm not interested in declining PRs unless I can actually prove the difference is negligible.

So I wrote a very simple patch to benchmark the entire call, seen here:
(excuse the typos for the time, it should be us, not ms)
https://gist.github.com/Spottedleaf/88e72a1f79bd117b260b3733529ee5ff

You'll notice the results vary from run-to-run, and they vary significantly depending on whether timings is on or not. So at that point I needed to figure out the run-to-run discrepancy, notably the initial run which generated the chunks.

To start I turned off timings in paper.yml. I also added extra info - I recorded the total number of setBlock calls that occurred during the entire random tick process to get an estimate of the number of block updates occurring - perhaps after gen there are more updates?

And this one made even less sense:
https://gist.github.com/Spottedleaf/8576cc63913c856fd5d88ab7155d7d58

The numbers are all over the place. I don't think using this method as a benchmark is a good idea.

This is the benchmark patch I'm using:
https://gist.github.com/Spottedleaf/7bb887f5861b82906203917fbdd5b3b3
There are some changes in there that don't matter, as I occasionally use the current dev workspace
to test other changes.

Rather than looking at random ticking as a whole for minor performance changes (as there doesn't appear to be reliable ways to measure mostly random functions), let's benchmark the code that's been changed.

I wrote a comparison test for Random#nextInt, ThreadUnsafeRandom#nextInt, and the field incrementing random you wrote here: https://gist.github.com/Spottedleaf/6ab12875a422461a8c8315a9eb309e6d

The results obviously show the field incrementor as the winner. But the important part is _by how much_.
Per call each function took:
threadUnsafeRandomBenchmark -> 0.8ns
incrementRandomBenchmark -> 0.2ns
threadSafeRandomBenchmark -> 3.28ns

One of the problems with this benchmark is that the CPU is very smart, and even though I tried to force it to write back to the fields, it's not going to - except in the case of the Random#nextInt, as it uses CAS. So the CAS should be pretty accurate - the rest are smaller than they really will be. Even assuming threadUnsafeRandomBenchmark is as slow as its CAS counterpart, you're only going to be making 2*3.28ns = 6.56ns max per random call. That's around 1.3% - 0.7% improvement at best, assuming the implementation you wrote is 0ns and threadUnsafeRandomBenchmark is as slow as the thread-safe variant. That's the _absolute best case_ (which we _can't verify_) it looks like - and it's barely 1% over the random tick method. 


So that's my problem with this patch. It doesn't appear to be doing anything. Looking at it again it makes sense - it's just eliminating some add/multiply operations? The most expensive part of those calculations are the field read/writes (see instruction_tables - add/mul/and instructions are all 1 cycle) - the function does a lot more than that. Like a lot more.

There is gain to made from this function, but eliminating simple operations isn't the way to go. It's far more complicated than that. Just benchmarking the thing is a process on its own. 

If you can show your benchmarking setup you got to get that 10%, _reliably_, I will look at this again. But as far as I can tell, the error range for measuring this function is well over 10% per run. 

I know you're trying to move _more_ patches from Airplane to Paper, so I need to make sure the _process_ of benchmarking and creating these kinds of patches is correct. I would suggest making the environment you are testing these patches with to be more transparent to me, as I can't replicate/verify any numbers you produce without knowing the environment you are testing them in.

Spottedleaf (changes requested):
See my comment

PaulBGD:
I appreciate you taking the time to look at this in-depth, so let me explain a little bit more about how I benchmarked this. I alluded to it in the comment, but essentially it is just loading 30,000 chunks with and without the patch to see how performance changes. Testing this over a longer period of time (such as 20 minutes) averages out most of the random stuff to give an accurate measurement of how it performs. This is how I do a lot of testing of my patches.

As for this patch, I reran the benchmark and came to the same conclusion you did. It seems that the difference is fairly negligible, giving only around a 2% improvement. Possibly in 1.16 when the patch was created, this area was relatively slower compared to the rest of the server. Either way, I think it's not significantly important now and we could see similar gains just replacing the random with the thread unsafe random if we really wanted to.

PR #6767, opened 2021-10-11
Cache palette array

PaulBGD:
Instead of allocating the 4KB for every chunk section, cache it locally and
reuse it for other chunk sections to save on allocations. These allocations add
up very quickly when saving chunks frequently.

For example, with 30,000 chunks and 16 chunk sections per chunk (which will
actually get larger in 1.18) it allocates nearly 2GB, which this patch
eliminates entirely.

### DIFF
--- a/patches/server/0826-Cache-palette-array.patch
+++ b/patches/server/0826-Cache-palette-array.patch
@@ -1,0 +1,61 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Paul Sauve <paul@technove.co>
+Date: Thu, 4 Feb 2021 23:28:46 -0600
+Subject: [PATCH] Cache palette array
+
+Instead of allocating the 4KB for every chunk section, cache it locally and
+reuse it for other chunk sections to save on allocations. These allocations add
+up very quickly when saving chunks frequently.
+
+diff --git a/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java b/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java
+index c9e942669458668a184aaec3bc0a5509dd6ab5f0..ac84d49d7819666a89cacbe9ef1199cf22ac2ac3 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java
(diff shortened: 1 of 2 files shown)


MiniDigger (approved):
lgtm

patches/server/0826-Cache-palette-array.patch
```diff
@@ -0,0 +1,61 @@
...
+                 T object2 = this.get(j);
+diff --git a/src/main/java/net/minecraft/world/level/chunk/storage/ChunkSerializer.java b/src/main/java/net/minecraft/world/level/chunk/storage/ChunkSerializer.java
+index 7921ee2786d0d6a60d43786b20efc03a0f9178e3..1eb085e1ecf438677b6784becb3f3e7f5a35ea65 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/storage/ChunkSerializer.java
++++ b/src/main/java/net/minecraft/world/level/chunk/storage/ChunkSerializer.java
+@@ -500,6 +500,7 @@ public class ChunkSerializer {
+         return new AsyncSaveData(blockLight, skyLight, blockTickListSerialized, fluidTickListSerialized, blockEntitiesSerialized, world.getGameTime());
+     }
+ 
++    private static final ThreadLocal<int[]> paletteArray = ThreadLocal.withInitial(() -> new int[4096]); // Paper
```
Proximyst: Naming conventions

me4502 (approved):

PaulBGD:
Bump

Spottedleaf:
Looks fine to me

patches/server/0826-Cache-palette-array.patch
PaulBGD: Sorry, I assume you're talking about leaving a descriptive comment but is there anything wrong with the variable name?

Proximyst: Constants should be `UPPER_CAMEL_CASE` is all. Merger can handle it 😃

kennytv (approved):

Machine-Maker:
@PaulBGD This patch needs to be re-worked for 1.18. There's a Codec in the mix now for PalettedContainer and multiple sizes of int arrays, (64 & 4096). It's probably still a viable improvement, it just needs work.

PR #4989, opened 2021-01-04
Add BlockBreakBlockEvent

Machine-Maker:
So not really sure what to call this event, but there wasn't a way to affect the drops when water or a piston breaks a block. So this adds that. Kind of an odd name I know, but if you've got a better one, suggest away please.

PR #6771, opened 2021-10-14
Add BlockDropResourcesEvent and EntityBreakBlockEvent

qixils:
Expands upon #4989 (cc @Machine-Maker) by:

- Creating a general-purpose BlockDropResourcesEvent for all resources (items) dropped by a block (which, unlike BlockDropItemEvent, works for more than just players)
- Making the previous BlockBreakBlockEvent extend this new event
- Creating a new EntityBreakBlockEvent for item drops caused by an entity

Use cases:

- adding new item drops to blocks (making all leaves drop apples, for instance)
- modifying the drops of crops from farming villagers
- a lot more, probably

~~note: blocks broken by players seem to not exist at time of event calling (i blame all the weirdly player-specific block-related events in bukkit), will investigate soon~~  
i identified the logic that leads to this behavior but it unfortunately doesn't seem workaround-able so i opted to just include a snapshot of the blockstate inside the event instead which is Good Enough


qixils:
Rebased (and fixed up) for 1.18.2

patches/api/0323-Add-BlockDropResourcesEvent-and-subclasses.patch
```diff
@@ -0,0 +1,196 @@
...
++     * If a snapshot is not provided then one will be created from {@link Block#getState()}.
++     *
++     * @param block      the block dropping resources
++     * @param blockState a snapshot of the block created before removal from the world
++     * @param drops      the resources being dropped
++     */
++    public BlockDropResourcesEvent(@NotNull Block block, @Nullable BlockState blockState, @NotNull List<ItemStack> drops) {
++        super(block);
++        this.blockState = blockState != null ? blockState : block.getState(true);
++        this.drops = Preconditions.checkNotNull(drops, "drops");
```
Machine-Maker: since you guarantee mutability it'd be good to just do new ArrayList() around the passed in argument.

Machine-Maker (changes requested):

patches/api/0323-Add-BlockDropResourcesEvent-and-subclasses.patch
qixils: good catch; fixed

PR #6773, opened 2021-10-14
Add more Llama API

BillyGalbreath:
This adds some useful API to Llamas, including two new events

### DIFF
--- a/patches/api/0338-Add-more-Llama-API.patch
+++ b/patches/api/0338-Add-more-Llama-API.patch
@@ -1,0 +1,190 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: William Blake Galbreath <blake.galbreath@gmail.com>
+Date: Thu, 14 Oct 2021 13:15:07 -0500
+Subject: [PATCH] Add more Llama API
+
+
+diff --git a/src/main/java/io/papermc/paper/event/entity/LlamaJoinCaravanEvent.java b/src/main/java/io/papermc/paper/event/entity/LlamaJoinCaravanEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..646ba4a585d61950bd89dec1e896b68eef0672bc
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/entity/LlamaJoinCaravanEvent.java
+@@ -0,0 +1,61 @@
++package io.papermc.paper.event.entity;
++
++import org.bukkit.entity.Llama;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.entity.EntityEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a Llama tries to join a caravan.
++ * <p>
++ * Cancelling the event will not let the Llama join. To prevent future attempts
++ * at joining a caravan use {@link Llama#setShouldJoinCaravan(boolean)}.
++ */
++public class LlamaJoinCaravanEvent extends EntityEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
++    private boolean canceled;
++    private final Llama head;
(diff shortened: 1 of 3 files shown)


patches/server/0827-Add-more-Llama-API.patch
```diff
@@ -0,0 +1,141 @@
...
+         this.caravanHead = llama;
+         this.caravanHead.caravanTail = this;
+     }
+@@ -450,6 +457,13 @@ public class Llama extends AbstractChestedHorse implements RangedAttackMob {
+         return this.caravanHead;
+     }
+ 
++    // Paper start
++    @Nullable
++    public Llama getCaravanTail() {
```
Machine-Maker: Probably better to just access transform the field instead of adding a getter for it

Machine-Maker (changes requested):

patches/api/0338-Add-more-Llama-API.patch
```diff
@@ -0,0 +1,190 @@
...
++}
+diff --git a/src/main/java/org/bukkit/entity/Llama.java b/src/main/java/org/bukkit/entity/Llama.java
+index d23226ccb0f6c25028f000ce31346cd0a8898e6a..d0057be88f7838449839e51d0a0e93dea6284bb7 100644
+--- a/src/main/java/org/bukkit/entity/Llama.java
++++ b/src/main/java/org/bukkit/entity/Llama.java
+@@ -3,6 +3,7 @@ package org.bukkit.entity;
+ import com.destroystokyo.paper.entity.RangedEntity;
+ import org.bukkit.inventory.LlamaInventory;
+ import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
```
lynxplay: Should probably be replaced either with its fully qualified name in the respective usecases (tho the ranged entity api imports a few lines above seems to disagree with the guidelines :joy:)

patches/api/0338-Add-more-Llama-API.patch
```diff
@@ -0,0 +1,190 @@
...
++     * Leave current caravan if in one
++     */
++    void leaveCaravan();
++
++    /**
++     * Check if another Llama is following this Llama
++     *
++     * @return True if being followed in the caravan
++     */
++    boolean hasCaravanTail();
```
lynxplay, line 162: Is this method actually useful ? I know it maps an internal method directly to the API but it seems to be just a somewhat useful shortcut for `getCaravanTail() != null`.

patches/server/0827-Add-more-Llama-API.patch
```diff
@@ -0,0 +1,127 @@
...
+     public Llama caravanTail;
++    public boolean shouldJoinCaravan = true; // Paper
+ 
+     public Llama(EntityType<? extends Llama> type, Level world) {
+         super(type, world);
+@@ -95,7 +96,7 @@ public class Llama extends AbstractChestedHorse implements RangedAttackMob {
+         if (!this.inventory.getItem(1).isEmpty()) {
+             nbt.put("DecorItem", this.inventory.getItem(1).save(new CompoundTag()));
+         }
+-
```
Proximyst: Don't drop this line, please. Same applies for other places where such is done.

Proximyst (approved):
Looks OK to me.

patches/api/0338-Add-more-Llama-API.patch
Proximyst: It's common practice to have this method in the API for some reason. I don't find it useful personally, but it's not a big deal to just have it.

Note also that it's not even useful in terms of IDE (and Kotlin) support for it, as an IDE will still yell at:

```java
if (!llama.hasCaravanTail()) {
  return;
}

llama.getCaravanTail().isOp(); // IntelliJ will say it's potentially throwing an NPE.
```

Machine-Maker (approved):

MiniDigger (approved):

PR #6776, opened 2021-10-14
Add LivingEntity#safeFallDistance API

BillyGalbreath:
This adds API to set the amount of safe fall distance for living entities


patches/api/0338-Add-LivingEntity-safeFallDistance-API.patch
```diff
@@ -0,0 +1,30 @@
...
++
++    /**
++     * Gets the distance this entity can safely fall without taking damage
++     *
++     * @return Safe fall distance
++     */
++    float getSafeFallDistance();
++
++    /**
++      * Set the distance this entity can safely fall without taking damage
```
Machine-Maker, line 27: I would make a note here, that this won't affect mobs that take no fall damage already. Like Bats, bees, basically all the mobs that override that checkFallDamage method in LivingEntity

Machine-Maker (commented):
Is it intentional that this isn't persisted across entity loads?

Machine-Maker (approved):

PR #6782, opened 2021-10-15
Expose the potential player cause of a lightning

### LINKED ISSUE #6781: Add a way to get Player that called LightningStrikeEvent with Cause.TRIDENT

See https://github.com/EngineHub/WorldGuard/issues/1845

1. Add method to LightningStrikeEvent#getSource() or similar that returns Player or null if this is a weather-related strike
2. Create subevent named like ChannelingStrikeEvent that called in that case and contains getter for the Player

Listen to trident launch event, store the Player and use it later in LightningStrikeEvent..?

### Other

I also think that currently there are no way to know who involved trident-related lighting strike, but if I'm wrong, please tell me. Thanks!

lynxplay:
As the internal lightning bolt representation already stores the player
instance that may have caused the lightning in the first place, through
tridents, this commit expands the existing additions to the lightning
api to expose the previously mentioned player instance that caused the
lightning.

Resolves: #6781

### DIFF
--- a/patches/api/0240-More-lightning-API.patch
+++ b/patches/api/0240-More-lightning-API.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] More lightning API
 
 
 diff --git a/src/main/java/org/bukkit/entity/LightningStrike.java b/src/main/java/org/bukkit/entity/LightningStrike.java
-index be347c3d0291f44036bae29a4e7e4645d6a4cdf6..2c81a3f685588431a3c7675c84b35a28975232af 100644
+index be347c3d0291f44036bae29a4e7e4645d6a4cdf6..6f5b6901032eb03606c4566b24459a03baac0c73 100644
 --- a/src/main/java/org/bukkit/entity/LightningStrike.java
 +++ b/src/main/java/org/bukkit/entity/LightningStrike.java
-@@ -31,4 +31,38 @@ public interface LightningStrike extends Entity {
+@@ -31,4 +31,72 @@ public interface LightningStrike extends Entity {
      @Override
      Spigot spigot();
      // Spigot end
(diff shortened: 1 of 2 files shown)


lynxplay:
Actually, might also be worth to have a setter for this? While I don't *really* see the use-case it could be nice for tracking damage from custom lightnings.

electronicboy (approved):

lynxplay:
A few edits to the PR: 

1. added the setter to the `LightningStrike` and
2. renamed the getters/setters. While mojang maps the player as plain `cause`, I am unsure if that is a good name in terms of API as the `Cause` term is usually coined to the different cause enums (which may be added in a future PR to the lightning as they are not fully represented through papers `getEntitySpawnReason` yet).

Machine-Maker:
I’d almost keep the api cause as an Entity, makes it more flexible for the future if some mob can cause lightning strikes.

lynxplay:
I mean, yea the getter *could* return an entity, but wouldn't this also be possible with some somewhat sane deprecation policies ?

The method name (at least right now) is specific enough to allow for an easy depreciation and replacement with something like `getCausingEntity` if mojang throws a mob at us that can summon lightning.
Current method could hence be kept (in case of depreciation) with a plain `instanceOf` check on the result of `getCausingEntity`.

Machine-Maker:
Right, but is there a downside to having it just return entity now? Can just put in the javadoc comment that an example is from a player using a trident

molor:
> I’d almost keep the api cause as an Entity, makes it more flexible for the future if some mob can cause lightning strikes.

I didn't really delve into how it works, but it's became interesting: if drowned throws a trident enchanted with channeling into target player, lightning bolt will not use the drowned as causer?

lynxplay:
> Right, but is there a downside to having it just return entity now? Can just put in the javadoc comment that an example is from a player using a trident

Well, nothing specifically in terms of usage. I personally would just find API confusing that exposes a setter that only takes an entity and exposes a respective getter that throws back an entity. But I digress, I'll update `getCausingPlayer` to `getCausingEntity` respectively :+1: 

//edit[0]: Is there any merit in just replacing the mojang cause with an entity ? From looking at it, it is only ever used internally once (to increase the statistics of lightning channeling) which could be easily wrapped with an `instanceOf` check.

lynxplay:
> drowned as

No, right now only players can be linked as the internal cause of a lightning bolt (from mojangs side of things).

Machine-Maker (approved):

kennytv (approved):

PR #6790, opened 2021-10-17
Add missing block data mins and maxes

Machine-Maker:
I noticed some inconsistency in methods for BlockDatas. For any integer property, there should be a getMinimum*Value* method so long as the minimum value isn't 0, and always a getMaximum*Value* method. This was missing in a couple places.

### DIFF
--- a/patches/api/0370-Add-missing-block-data-mins-and-maxes.patch
+++ b/patches/api/0370-Add-missing-block-data-mins-and-maxes.patch
@@ -1,0 +1,67 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 16 Oct 2021 22:57:10 -0700
+Subject: [PATCH] Add missing block data mins and maxes
+
+
+diff --git a/src/main/java/org/bukkit/block/data/Levelled.java b/src/main/java/org/bukkit/block/data/Levelled.java
+index 5255538fecae6da413546be3adacd2a99f6c74e9..860f072dee391b300cb1629058a3f9c23dfd95e2 100644
+--- a/src/main/java/org/bukkit/block/data/Levelled.java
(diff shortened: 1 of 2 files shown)


patches/server/0827-Add-missing-block-data-mins-and-maxes.patch
```diff
@@ -0,0 +1,57 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/block/data/type/CraftLeaves.java b/src/main/java/org/bukkit/craftbukkit/block/data/type/CraftLeaves.java
+index 5b0da54982fc0879005fd1db104284eb3318bee8..3c5a2de56724bc784f619f3087140c72a42dc57b 100644
+--- a/src/main/java/org/bukkit/craftbukkit/block/data/type/CraftLeaves.java
++++ b/src/main/java/org/bukkit/craftbukkit/block/data/type/CraftLeaves.java
+@@ -3,7 +3,7 @@ package org.bukkit.craftbukkit.block.data.type;
+ import org.bukkit.block.data.type.Leaves;
+ import org.bukkit.craftbukkit.block.data.CraftBlockData;
+ 
+-public class CraftLeaves extends CraftBlockData implements Leaves {
++public abstract class CraftLeaves extends CraftBlockData implements Leaves { // Paper - make abstract (not used anyways)
```
kennytv: unnecessary

kennytv (approved):

patches/server/0827-Add-missing-block-data-mins-and-maxes.patch
Machine-Maker: Wouldn't I have to implement the methods I just added in this Class if I didnt make it abstract? I think that's why plenty of these are abstract so not all methods from the interfaces have to be implemented cause they are templates.

Machine-Maker:
Rebased for 1.18.1.

**Needs a re-review, changes made**

electronicboy (approved):

PR #6794, opened 2021-10-17
Bonemeal API

Owen1212055:
This adds some bonemeal api with a blockstate predicate. I am aware there is already api to simulate bonemeal, but this offers the advantage of being able to chose which blocks actually get modified.


patches/server/0827-BoneMeal-API.patch
```diff
@@ -0,0 +1,70 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sun, 17 Oct 2021 14:53:35 -0400
+Subject: [PATCH] BoneMeal API
+
+
+diff --git a/src/main/java/net/minecraft/world/item/BoneMealItem.java b/src/main/java/net/minecraft/world/item/BoneMealItem.java
+index c5026e617c3f71ba79532de135f571c27b7dd4a7..e7daa1be4212de9ec3942f8053baa6fc1d5d12cf 100644
+--- a/src/main/java/net/minecraft/world/item/BoneMealItem.java
++++ b/src/main/java/net/minecraft/world/item/BoneMealItem.java
```
NoahvdAa: Missing paper comments for this file

patches/server/0827-BoneMeal-API.patch
```diff
@@ -0,0 +1,70 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -2160,5 +2160,28 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+ 
+         return this.adventure$pointers;
+     }
++
++    @Override
++    public boolean applyBoneMeal(@org.jetbrains.annotations.NotNull Location location, org.bukkit.block.BlockFace face, boolean showParticles, @org.jetbrains.annotations.Nullable Predicate<BlockState> predicate) {
++        BlockPos pos = new BlockPos(location.getBlockX(), location.getBlockY(), location.getBlockZ());
++        var context = new net.minecraft.world.item.context.UseOnContext(this.getHandle(), null, net.minecraft.world.InteractionHand.MAIN_HAND, net.minecraft.world.item.Items.BONE_MEAL.getDefaultInstance(), new net.minecraft.world.phys.BlockHitResult(Vec3.ZERO, CraftBlock.blockFaceToNotch(blockFace), pos, false));
```
lynxplay: The variable `blockFace` does not exist here. I am guessing this should have been `face`.

lynxplay (changes requested):

patches/server/0827-BoneMeal-API.patch
Owen1212055: Yea was just onto fixing that, thanks tho!

patches/server/0827-BoneMeal-API.patch
```diff
@@ -0,0 +1,73 @@
...
++        BlockPos pos = new BlockPos(location.getBlockX(), location.getBlockY(), location.getBlockZ());
++        var context = new net.minecraft.world.item.context.UseOnContext(this.getHandle(), null, net.minecraft.world.InteractionHand.MAIN_HAND, net.minecraft.world.item.Items.BONE_MEAL.getDefaultInstance(), new net.minecraft.world.phys.BlockHitResult(Vec3.ZERO, CraftBlock.blockFaceToNotch(face), pos, false));
++
++        world.captureTreeGeneration = true;
++        world.captureBlockStates = true;
++
++        var result = net.minecraft.world.item.BoneMealItem.applyBonemeal(context, showParticles);
++
++        world.captureTreeGeneration = false;
++        world.captureBlockStates = false;
```
Machine-Maker: I think this might be a situation where you want to set the capture variables back to what they were before,

patches/server/0827-BoneMeal-API.patch
```diff
@@ -0,0 +1,73 @@
...
++
++        for (BlockState blockState : world.capturedBlockStates.values()) {
++            if (predicate != null && predicate.test(blockState)) {
++                blockState.update(true);
++            }
++        }
```
Machine-Maker: I'm concerned the capturedBlockStates collection will sometimes have blocks in it already when a plugin calls this method. Say during a block place event handler or something. So you really need to only check blockstates that have been added directly as a result of the applyBonemeal call, not ones that were already captured before.

Machine-Maker (changes requested):

Owen1212055:
I'd like your opinion on this, I now store the previous collections + if it is capturing before and reset it after. But i'd like to see if I should also capture the blocks that the plugin updates via the predicate or just ignore it (like I currently ignore it).

Owen1212055:
Rebased

Owen1212055:
Feedback still appreciate on this, and I still think that this API would be useful.

Owen1212055:
Still relevant and looking for feedback. :)

patches/api/0351-BoneMeal-API.patch
```diff
@@ -0,0 +1,30 @@
...
++++ b/src/main/java/org/bukkit/World.java
+@@ -3831,4 +3831,19 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+             }
+         }
+     }
++    // Paper start
++    /**
++     * Applies bonemeal at a specific location
++     * <p>
++     * Modification done in the predicate are respected
```
NoahvdAa:
```suggestion
+     * Modifications done in the predicate are respected
```

patches/server/0839-BoneMeal-API.patch
```diff
@@ -0,0 +1,84 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -2154,5 +2154,39 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+ 
+         return this.adventure$pointers;
+     }
++
++    @Override
++    public boolean applyBoneMeal(@org.jetbrains.annotations.NotNull Location location, org.bukkit.block.BlockFace face, boolean showParticles, @org.jetbrains.annotations.Nullable Predicate<BlockState> predicate) {
++        BlockPos pos = new BlockPos(location.getBlockX(), location.getBlockY(), location.getBlockZ());
++        var context = new net.minecraft.world.item.context.UseOnContext(this.getHandle(), null, net.minecraft.world.InteractionHand.MAIN_HAND, net.minecraft.world.item.Items.BONE_MEAL.getDefaultInstance(), new net.minecraft.world.phys.BlockHitResult(Vec3.ZERO, CraftBlock.blockFaceToNotch(face), pos, false));
```
NoahvdAa: var 😭

Owen1212055: I have a complicated relationship with var.

PR #6795, opened 2021-10-17
Undeprecate Delegate Tree method

Owen1212055:
This was wrongly deprecated because it was thought it did not include tile entities, however with some basic testing it certainly does support them.
The issue here is that the treetypes themselves don't use any generators that contain beehives, so you will never be able to encounter them via the api.
For looking at the logic, see AbstractTreeGrower. This part is not used in the API at all, but as you can see here it specifies the special tree type if there are nearby flowers.

Tree without beehive (used by api):
```java
 public static final ConfiguredFeature<TreeConfiguration, ?> OAK = register("oak", Feature.TREE.configured((new TreeConfiguration.TreeConfigurationBuilder(new SimpleStateProvider(Features.States.OAK_LOG), new StraightTrunkPlacer(4, 2, 0), new SimpleStateProvider(Features.States.OAK_LEAVES), new SimpleStateProvider(Features.States.OAK_SAPLING), new BlobFoliagePlacer(ConstantInt.of(2), ConstantInt.of(0), 3), new TwoLayersFeatureSize(1, 0, 1))).ignoreVines().build()));
```

Tree WITH beehive (unused):
```java
 public static final ConfiguredFeature<TreeConfiguration, ?> OAK_BEES_0002 = register("oak_bees_0002", Feature.TREE.configured(OAK.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_0002))));
    public static final ConfiguredFeature<TreeConfiguration, ?> OAK_BEES_002 = register("oak_bees_002", Feature.TREE.configured(OAK.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_002))));
    public static final ConfiguredFeature<TreeConfiguration, ?> OAK_BEES_005 = register("oak_bees_005", Feature.TREE.configured(OAK.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_005))));
    public static final ConfiguredFeature<TreeConfiguration, ?> BIRCH_BEES_0002 = register("birch_bees_0002", Feature.TREE.configured(BIRCH.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_0002))));
    public static final ConfiguredFeature<TreeConfiguration, ?> BIRCH_BEES_002 = register("birch_bees_002", Feature.TREE.configured(BIRCH.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_002))));
    public static final ConfiguredFeature<TreeConfiguration, ?> BIRCH_BEES_005 = register("birch_bees_005", Feature.TREE.configured(BIRCH.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_005))));
    public static final ConfiguredFeature<TreeConfiguration, ?> FANCY_OAK_BEES_0002 = register("fancy_oak_bees_0002", Feature.TREE.configured(FANCY_OAK.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_0002))));
    public static final ConfiguredFeature<TreeConfiguration, ?> FANCY_OAK_BEES_002 = register("fancy_oak_bees_002", Feature.TREE.configured(FANCY_OAK.config().withDecorators(ImmutableList.of(Features.Decorators.BEEHIVE_002))));
```

Tested by making the api use OAK_BEES instead of OAK.


Machine-Maker:
Maybe im not understanding, but the api TALL_BIRCH TreeType does use Features.SUPER_BIRCH_BEES_0002 so that does deal with tile entities.

Owen1212055:
> Maybe im not understanding, but the api TALL_BIRCH TreeType does use Features.SUPER_BIRCH_BEES_0002 so that does deal with tile entities. 

Ahh, I was only looking at the oak tree type. I'll look into this further, because at least with my testing using generate tree I was able to make beehives just fine.

Machine-Maker:
What the deprecation message means, is that the BlockChangeDelegate doesn't handle tile entities. So I think what happens, is that the beehive block is created, but the tile entity is not. I think if you reload the world, you'd get an exception thrown somewhere cause its expecting a tile entity at the same place as the beehive but none were created.

Owen1212055:
I'll need to play around with it, since the current method doesn't seem to allow you to prevent the placement of a block... unlike this one.

Owen1212055:
> What the deprecation message means, is that the BlockChangeDelegate doesn't handle tile entities. So I think what happens, is that the beehive block is created, but the tile entity is not. I think if you reload the world, you'd get an exception thrown somewhere cause its expecting a tile entity at the same place as the beehive but none were created.

The thing is it calls the method that is located in the BlockChangeDelgate
```java
if (grownTree) { // Copy block data to delegate
            for (BlockState blockstate : world.capturedBlockStates.values()) {
                BlockPos position = ((CraftBlockState) blockstate).getPosition();
                net.minecraft.world.level.block.state.BlockState oldBlock = this.world.getBlockState(position);
                int flag = ((CraftBlockState) blockstate).getFlag();
                delegate.setBlockData(blockstate.getX(), blockstate.getY(), blockstate.getZ(), blockstate.getBlockData());
                net.minecraft.world.level.block.state.BlockState newBlock = this.world.getBlockState(position);
                this.world.notifyAndUpdatePhysics(position, null, oldBlock, newBlock, newBlock, flag, 512);
            }
            world.capturedBlockStates.clear();
            return true;
        } else {
            world.capturedBlockStates.clear();
            return false;
        }
```

It's up to you to place it, it doesn't really mess with tile entities at all here. And when testing, all beehives seem to have had tile entities just fine.

Owen1212055:
Rebased and included in a pre-existing patch.

Machine-Maker (changes requested):
Ok, I finally figured out what was wrong with the BlockChangeDelegate. Tree's with beehives are supposed to generate with between 2 and 4 bees in the hive. these trees generate with none. The beehive block and block entity are created just fine, but are NOT populated with the correct nbt data, because when BeehiveDecorator goes to place it, its not placed in the world, but captured. then it checks if the block entity exists and if it does, populates it. but it doesnt exist cause the placed blockState was captured.

Owen1212055:
> Ok, I finally figured out what was wrong with the BlockChangeDelegate. Tree's with beehives are supposed to generate with between 2 and 4 bees in the hive. these trees generate with none. The beehive block and block entity are created just fine, but are NOT populated with the correct nbt data, because when BeehiveDecorator goes to place it, its not placed in the world, but captured. then it checks if the block entity exists and if it does, populates it. but it doesnt exist cause the placed blockState was captured.

So what should be done about that? Honestly I'd like to see if we can fix this method instead of deprecate it... as there isn't any api like this currently. (Unless we just make a new one, but with a predicate or something)

Machine-Maker:
I mean there kinda is no? Isn’t there a method that takes a consumer of block states right above it?

wizjany:
any reason
```java
interface BlockChangeDelegate {
    @Deprecated
    boolean setBlockData(int x, int y, int z, BlockData blockData);
    default boolean setBlockState(int x, int y, int z, BlockState blockState) {
        return setBlockData(x, y, z, blockState.getBlockData());
    }
}
```
doesn't work? can shove the blockentities with bees through via blockstate, and bees will only disappear for un-updated/legacy delegates. (tbh idk who uses this besides WE)

Owen1212055:
I mean the only behavior I use here is the ability to just totally cancel the block being placed there. The other one doesn’t allow you to cancel the block totally, only replace it. (It’s not a predicate)

And wiz, thanks for that suggestion I’ll try to play with that. In my testing I really never noticed anything wrong, so I’ll play around with these solutions.

Machine-Maker:
> The other one doesn’t allow you to cancel the block totally, only replace it. (It’s not a predicate)

you should just be able to set the blockstate to air no?, or do whatever the blockstate currently is cause it hasn't changed. That effectively cancels it.

Owen1212055:
Upstream is going to somewhat resolve this with a predicate instead. https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/987

PR #6796, opened 2021-10-17
Add Shearable API

Owen1212055:
Not sure if the source is needed in the method, but I thought it might be neat to include.


patches/api/0337-Add-Shearable-API.patch
```diff
@@ -0,0 +1,61 @@
...
+@@ -0,0 +1,10 @@
++package io.papermc.paper.entity;
++
++import net.kyori.adventure.sound.Sound;
++
++public interface Shearable {
++
++    void shear(Sound.Source source);
++
++    boolean canBeShared();
```
kezz: Typo here :p

Owen1212055: _oopsies_

patches/api/0337-Add-Shearable-API.patch
```diff
@@ -0,0 +1,71 @@
...
++import net.kyori.adventure.sound.Sound;
++
++public interface Shearable {
++
++    /**
++     * Causes the entity to play the effect of being shared by a player.
++     *
++     * @param source Sound source to play any sound effects on
++     */
++    void shear(Sound.Source source);
```
imDaniX: `Nullable`/`NonNull` annotation missing?

lynxplay:
Ignore the references :sweat_smile: I typo-ed the "Resolves" link from `6797` to `6796`.

patches/api/0337-Add-Shearable-API.patch
```diff
@@ -0,0 +1,71 @@
...
++import net.kyori.adventure.sound.Sound;
++
++public interface Shearable {
++
++    /**
++     * Causes the entity to play the effect of being shared by a player.
++     *
++     * @param source Sound source to play any sound effects on
++     */
++    void shear(@org.jetbrains.annotations.NotNull Sound.Source source);
```
Machine-Maker: You are creating a new interface here, can import this annotation.

Machine-Maker (changes requested):
I think all of the server modifications can be condensed into a CraftShearable interface with default methods for shear and canBeSheared and then a interface method for getHandle that returns the nms Shearable interface. That way you arent writing essentially 3 copies of the same code.

patches/server/0819-Add-Shearable-API.patch
```diff
@@ -0,0 +1,71 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/entity/CraftShearable.java b/src/main/java/io/papermc/paper/entity/CraftShearable.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..07c9c251e832742aa4b3052b7e61e61897a9662f
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/CraftShearable.java
+@@ -0,0 +1,20 @@
++package io.papermc.paper.entity;
++
++import io.papermc.paper.adventure.*;
```
lynxplay: These wildcard imports seem pretty useless, same goes for the one found in the `io.papermc.paper.entity.Shearable` interface. Are they there on purpose ?

electronicboy: http://www.javadude.com/posts/20040522-import-on-demand-is-evil/

Owen1212055: Should be resolved here.

patches/server/0819-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
++++ b/src/main/java/io/papermc/paper/entity/CraftShearable.java
+@@ -0,0 +1,21 @@
++package io.papermc.paper.entity;
++
++import io.papermc.paper.adventure.PaperAdventure;
++import net.kyori.adventure.sound.Sound;
++import net.minecraft.world.entity.Shearable;
++import org.jetbrains.annotations.NotNull;
++
++public interface CraftShearable extends io.papermc.paper.entity.Shearable {
```
Machine-Maker: I actually may have been incorrect in suggesting this be called CraftShearable, sometimes we use Paper as the prefix, not Craft, but sometimes not. I'm not quite sure it matters at all, so this isnt a requested change, just a comment for people to respond to

Machine-Maker (approved):

patches/server/0819-Add-Shearable-API.patch
Owen1212055: Changed regardless.

Owen1212055:
Also rebased

patches/server/0835-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
++import net.minecraft.world.entity.Shearable;
++import org.jetbrains.annotations.NotNull;
++
++public interface PaperShearable extends io.papermc.paper.entity.Shearable {
++
++    Shearable getHandle();
++
++    @Override
++    default boolean canBeSheared() {
++        return getHandle().readyForShearing();
```
jpenilla: Use `this.` for accessing instance methods

patches/server/0835-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
++    }
++
++    @Override
++    default void shear(@NotNull Sound.Source source) {
++        getHandle().shear(PaperAdventure.asVanilla(source));
```
jpenilla: Use `this.` for accessing instance methods

patches/api/0343-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
++     * @param source Sound source to play any sound effects on
++     */
++    void shear(@NotNull Sound.Source source);
++
++    /**
++     * Gets if the entity is able to be sheared or not.
++     *
++     * @return if the entity can be sheared
++     */
++    boolean canBeSheared();
```
jpenilla: I think the Mojang name `readyForShearing` better communicates what the return value means. In addition to renaming the Javadoc needs improvement as well, implementing `Shearable` implies it is able to be sheared, this method is checking if it's ready to be sheared.

patches/api/0343-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
++import org.jetbrains.annotations.NotNull;
++
++public interface Shearable {
++
++    /**
++     * Causes the entity to play the effect of being shared by a player.
++     *
++     * @param source Sound source to play any sound effects on
++     */
++    void shear(@NotNull Sound.Source source);
```
jpenilla: Maybe include a default overload which uses the PLAYER source, which is what vanilla uses?

patches/api/0343-Add-Shearable-API.patch
```diff
@@ -0,0 +1,72 @@
...
+@@ -0,0 +1,21 @@
++package io.papermc.paper.entity;
++
++import net.kyori.adventure.sound.Sound;
++import org.jetbrains.annotations.NotNull;
++
++public interface Shearable {
++
++    /**
++     * Causes the entity to play the effect of being shared by a player.
```
jpenilla: If `readyFoShearing` is true, then calling this method will actually shear the Shearable in addition to playing the sound. This should be noted in the Javadoc.

jpenilla (dismissed):

Owen1212055:
Clarified the javadocs a bit, hopefully, it makes more sense now.

Owen1212055:
Some entities also implement an interface called ``Bucketable``, which could allow us to add some methods such as being able to get the item produced when bucketing it, or forcing it to be bucketed too.
image
image

This is somewhat similar to this PR, in that it just exposes an interface that certain entities can implement. Should I make this PR a bit more "generic" to "Expose Vanilla Entity Interfaces" or something?

So I am proposing I add the ``Bucketable`` interface to this PR as well.

patches/api/0347-Add-Shearable-API.patch
```diff
@@ -0,0 +1,81 @@
...
+index 0000000000000000000000000000000000000000..dd2d00b3dfeca5400a86d8d6b78e1e1d74c71d8d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/Shearable.java
+@@ -0,0 +1,30 @@
++package io.papermc.paper.entity;
++
++import net.kyori.adventure.sound.Sound;
++import org.jetbrains.annotations.NotNull;
++
++public interface Shearable {
```
Machine-Maker: This can probably extend the base Entity interface

patches/api/0347-Add-Shearable-API.patch
```diff
@@ -0,0 +1,81 @@
...
++
++    /**
++     * Forces the entity to play the effect of being sheared by a player.
```
Machine-Maker: I feel that its wrong to put "play the effect of being sheared" before it *actually* being sheared. The method is `shear`, the javadoc should first state, this will shear the entity, and add that it plays associated effects

Owen1212055:
Rebased and made provided changes.

Owen1212055:
Still around

jpenilla:
Something that should at least be documented to some extent is that the entity may no longer be valid after shearing, such as in the case of a Mooshroom, it will be replaced by a new entity.

Also, do we care about calling the shear events and returning a boolean? Interested what others think on both these concerns.

Owen1212055:
> Something that should at least be documented to some extent is that the entity may no longer be valid after shearing, such as in the case of a Mooshroom, it will be replaced by a new entity.
> 
> Also, do we care about calling the shear events and returning a boolean? Interested what others think on both these concerns.

Personally I think that the event here can still be followed, as it's supposed to simulate __as if it were sheared by a player/source__. But yea, i'd wonder what other swould think to.

And yep, documented & rebased!

PR #6804, opened 2021-10-19
Properly remove overriding potion effect modifiers

lynxplay:
As of minecraft 1.15.2's first pre-release potion effects no longer
completely replace weaker once of the same potion effect type but rather
build a list of overriding potion effects to preserve and eventually
re-apply weaker potion effects that last longer than more powerful but
more shortlived overriding potion effects of the same type.

This replacement of such overriding potion effects simply removes the
attribute modifier added by the potion effect based on a single uuid
defined by each potion effect type.
After the removal, a new attribute modified is added that holds the
updated modifications based on the lower amplifier of the weaker potion
effect.
As the removal of the stronger potion effect is purely based on the
single uuid of each individual potion effect type the amplifier of the
old effect is not needed to properly remove the effect and is sadly not
passed to the removal logic in the first place.

This however fails for the absorption potion effect type, as it cannot
make use of the attribute modifier system and the single uuid concept
used by other potion effect types.
Hence, prior to this commit, it simply calculates the amount of
absorption hearts granted by the amplifier of the potion effect and
removes these.
As the server however only provides the amplifier (more specifically the
new potion effect) to the update method once an overriding absorption
effect is replaced, the removal only removes the amount of absorption
hearts.

This patch combats this faulty behaviour by now properly passing the old
and new mob effect (and their respective amplifier) through the update
logic of the living entity and mob effects.
While this prevents the absorption effect from leaving the player with
absorption hearts after the effect completely expires, it did allow the
effect to re-generate absorption hearts that were lost due to damage
when replacing the overriding effect with the lower potion effect.

The patch introduces MobEffect#updateAttributeModifiers which is
responsible for performing the previously described logic of removing
and applying attribute modifiers.
This method is then leveraged by the absorption mob effect to correctly
calculate the difference between the old and new amplifiers.

Resolves: https://bugs.mojang.com/browse/MC-182497

-------------------------------------------------------------------

Note, the consumer call in `MobEffect#tick` was moved up as the mob effect class does not offer an easy `copy` method to safely store the old mob effect before updating the effect values.

### DIFF
--- a/patches/server/0910-Properly-remove-overriding-potion-effect-modifiers.patch
+++ b/patches/server/0910-Properly-remove-overriding-potion-effect-modifiers.patch
@@ -1,0 +1,193 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bjarne Koll <lynxplay101@gmail.com>
+Date: Tue, 19 Oct 2021 17:24:47 +0200
+Subject: [PATCH] Properly remove overriding potion effect modifiers
+
+As of minecraft 1.15.2's first pre-release potion effects no longer
+completely replace weaker once of the same potion effect type but rather
+build a list of overriding potion effects to preserve and eventually
+re-apply weaker potion effects that last longer than more powerful but
+more shortlived overriding potion effects of the same type.
+
+This replacement of such overriding potion effects simply removes the
+attribute modifier added by the potion effect based on a single uuid
+defined by each potion effect type.
+After the removal, a new attribute modified is added that holds the
+updated modifications based on the lower amplifier of the weaker potion
+effect.
+As the removal of the stronger potion effect is purely based on the
+single uuid of each individual potion effect type the amplifier of the
+old effect is not needed to properly remove the effect and is sadly not
+passed to the removal logic in the first place.
+
+This however fails for the absorption potion effect type, as it cannot
+make use of the attribute modifier system and the single uuid concept
+used by other potion effect types.
+Hence, prior to this commit, it simply calculates the amount of
+absorption hearts granted by the amplifier of the potion effect and
+removes these.
+As the server however only provides the amplifier (more specifically the
+new potion effect) to the update method once an overriding absorption
+effect is replaced, the removal only removes the amount of absorption
+hearts.
+
+This patch combats this faulty behaviour by now properly passing the old
+and new mob effect (and their respective amplifier) through the update
+logic of the living entity and mob effects.
+While this prevents the absorption effect from leaving the player with
+absorption hearts after the effect completely expires, it did allow the
+effect to re-generate absorption hearts that were lost due to damage
+when replacing the overriding effect with the lower potion effect.
+
+The patch introduces MobEffect#updateAttributeModifiers which is
+responsible for performing the previously described logic of removing
+and applying attribute modifiers.
+This method is then leveraged by the absorption mob effect to correctly
+calculate the difference between the old and new amplifiers.
+
+Resolves: https://bugs.mojang.com/browse/MC-182497
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index c8057f98e16ba6e19640e0b250e5201e0f4f57db..ad6c37168a54d77a569a5f8a5a5e07820d10e771 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1774,6 +1774,13 @@ public class ServerPlayer extends Player {
+     @Override
+     protected void onEffectUpdated(MobEffectInstance effect, boolean reapplyEffect, @Nullable Entity source) {
+         super.onEffectUpdated(effect, reapplyEffect, source);
++        // Paper start - properly remove old overridden potion effect
++    }
++
++    @Override
++    protected void onEffectUpdated(MobEffectInstance effect, @Nullable MobEffectInstance oldEffect, boolean reapplyEffect, @Nullable Entity source) {
++        super.onEffectUpdated(effect, oldEffect, reapplyEffect, source);
++        // Paper end - properly remove old overridden potion effect
+         this.connection.send(new ClientboundUpdateMobEffectPacket(this.getId(), effect));
+         CriteriaTriggers.EFFECTS_CHANGED.trigger(this, source);
+     }
+diff --git a/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java b/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
+index 8378ec06edd0a62b5e8c26ea1be9d8813f7c4c0a..e95ecdc30ac2da435e445507fc82d73dae494e1a 100644
+--- a/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
++++ b/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
+@@ -10,13 +10,35 @@ public class AbsoptionMobEffect extends MobEffect {
+ 
+     @Override
+     public void removeAttributeModifiers(LivingEntity entity, AttributeMap attributes, int amplifier) {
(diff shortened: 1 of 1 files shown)


patches/server/0828-Properly-remove-overriding-potion-effect-modifiers.patch
```diff
@@ -0,0 +1,188 @@
...
+     }
+diff --git a/src/main/java/net/minecraft/world/effect/MobEffectInstance.java b/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
+index 544b3711bc7538609f64ad30995ef927f562a36f..676126535c3723a4232dae6157f359b860486432 100644
+--- a/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
++++ b/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
+@@ -132,6 +132,11 @@ public class MobEffectInstance implements Comparable<MobEffectInstance> {
+     }
+ 
+     public boolean tick(LivingEntity entity, Runnable overwriteCallback) {
++        // Paper end - properly remove old overridden potion effect
```
NoahvdAa:
```suggestion
+        // Paper start - properly remove old overridden potion effect
```

patches/server/0828-Properly-remove-overriding-potion-effect-modifiers.patch
```diff
@@ -0,0 +1,188 @@
...
+diff --git a/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java b/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
+index c4352471cbbfb68e4b46704bbf68bb2e1404dbb0..ee70940cdd198ad2ace3ec4120a5064a9cad3fb6 100644
+--- a/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
++++ b/src/main/java/net/minecraft/world/effect/AbsoptionMobEffect.java
+@@ -10,13 +10,31 @@ public class AbsoptionMobEffect extends MobEffect {
+ 
+     @Override
+     public void removeAttributeModifiers(LivingEntity entity, AttributeMap attributes, int amplifier) {
+-        entity.setAbsorptionAmount(entity.getAbsorptionAmount() - (float)(4 * (amplifier + 1)));
++        entity.setAbsorptionAmount(entity.getAbsorptionAmount() - calculateGrantedAbsorptionHearts(amplifier)); // Paper start - properly remove old overridden potion effect
```
NoahvdAa:
```suggestion
+        entity.setAbsorptionAmount(entity.getAbsorptionAmount() - calculateGrantedAbsorptionHearts(amplifier)); // Paper - properly remove old overridden potion effect
```
Same for line 84

NoahvdAa (approved):

patches/server/0828-Properly-remove-overriding-potion-effect-modifiers.patch
lynxplay: :pray: Thank you, the good old copy/paste for these comments is not always the most fun

lynxplay:
Rebased and tested :+1:

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
```diff
@@ -0,0 +1,190 @@
...
+     @Override
+     protected void onEffectUpdated(MobEffectInstance effect, boolean reapplyEffect, @Nullable Entity source) {
+         super.onEffectUpdated(effect, reapplyEffect, source);
++        // Paper start - properly remove old overridden potion effect
++    }
++
++    @Override
++    protected void onEffectUpdated(MobEffectInstance effect, @Nullable MobEffectInstance oldEffect, boolean reapplyEffect, @Nullable Entity source) {
++        super.onEffectUpdated(effect, oldEffect, reapplyEffect, source);
++        // Paper end - properly remove old overridden potion effect
```
Machine-Maker: I think this will break anywhere that calls the "legacy" onEffectUpdated method because its no longer gonna start tracking the levitation position. You should make the "legacy" method call the new one with a null extra argument. (just like you did when you added the new method in LivingEntity)

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
```diff
@@ -0,0 +1,190 @@
...
+diff --git a/src/main/java/net/minecraft/world/effect/MobEffect.java b/src/main/java/net/minecraft/world/effect/MobEffect.java
+index 8bbb9bdcf95989f1737714655f6f6a269d46d7f2..28a0809a043640d1bd912106d0bf367090252360 100644
+--- a/src/main/java/net/minecraft/world/effect/MobEffect.java
++++ b/src/main/java/net/minecraft/world/effect/MobEffect.java
+@@ -192,6 +192,13 @@ public class MobEffect {
+ 
+     }
+ 
++    // Paper start - properly remove old overridden potion effect - moved here from LivingEntity#onEffectUpdated
++    public void updateAttributeModifiers(final LivingEntity entity, final AttributeMap attributes, final int oldAmplifier, final int newAmplifier) {
```
Machine-Maker: Could you remove the `attributes` parameter? and just get the attributes from the LivingEntity

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
lynxplay: I am unsure if I understand exactly what you mean here.

The `levitationStartPos` is maintained by `ServerPlayer#onEffectAdded` which previously was also only called once (when the first levitation effect was added, updates did not re-trigger it). None of the changes in this patch touch the points at which `onEffectAdded` is called.

Additionally, the "legacy" method `ServerPlayer#onEffectUpdated` simply calls its super method `LivingEntity#onEffectUpdated`: 
```java
protected void onEffectUpdated(MobEffectInstance effect, boolean reapplyEffect, @Nullable Entity source) {
    // Paper start - properly remove old overridden potion effect
    this.onEffectUpdated(effect, null, reapplyEffect, source);
}
```
which will properly call the new `ServerPlayer#onEffectUpdated` with a null argument.

I might have completely misunderstood your point tho :sweat_smile: so if you could elaborate a bit on it, that would be great <3

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
lynxplay: Sure, this method signature mainly followed the existing `removeAttributeModifiers` and `addAttributeModifiers` methods, which both take the `LivingEntity` and its `AttributeMap` as parameters, hence I was inclined to follow in those footsteps (to comply with whatever the reason for the "doubled" passing of the attribute map was on mojangs side).

I'll remove it, thanks for the feedback. Sorry it took so long I just got back from vacation and did not have a development ready machine with me.

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
Machine-Maker: oh, I think I mis-typed, thinking I was looking at the effectsAdded method. What I should've said, was anything that called the legacy `ServerPlayer#onEffectUpdated` method would no longer trigger the advancement criteria or send a mob effect update packet, since that legacy method doesn't delegate to the overloaded method in ServerPlayer, rather the super method. 

EDIT: But yeah, even that will be called in the super method, so you can ignore this.

lynxplay: Okay, sweet!

patches/server/0817-Properly-remove-overriding-potion-effect-modifiers.patch
lynxplay: Removed the passed attribute map.

If this is really that much nicer in terms of code, idk.
E.g. we now need to grab the attribute map explicitly when overwriting the method in the absorption mob effect.

lynxplay:
Rebased for 1.18.2 :+1:

lynxplay:
Rebased for 1.19.

I move some stuff around too to maintain compatibility with original vanilla call logic order. Specifically, the overwriteCallback is now again called after the original effect was updated.

PR #6806, opened 2021-10-20
Fix kelp modifier having reverse effect

### LINKED ISSUE #6805: Kelp growth modifier behaving incorrectly
### Expected behavior

Increasing the growth modifier will increase the growth rate of a crop.

### Observed/Actual behavior

Increasing the growth rate modifier of kelp slows down the growth rate.

### Steps/models to reproduce

Lower the kelp growth rate to 20, plant some kelp and observe how long it takes for it to grow. Set the kelp growth rate to 200 and observe again. 

### Plugin and Datapack List

None

### Paper version

Master branch

### Other

It looks like the formula used to modify crop growth rate on kelp works differently from CropBlock growth rate. Where increasing the growth modifier on crops that qualify as a CropBlock increases the chance of a crop growing, **decreasing** the crop growth modifier increases the chance of a crop growing.

I copied over the formulas used currently (line 44 of GrowingPlantHeadBlock.java and line 91 of CropBlock) to an online java IDE and ran each one a hundred times to see how many times the crop would grow (how often those lines would return true). After doing that, as previously stated, increasing the kelp growth modifier would work the opposite as increasing other crop growth modifiers.

Machine-Maker:
Fixes #6805

### DIFF
--- a/patches/server/0819-Fix-kelp-modifier-changing-growth-for-other-crops.patch
+++ b/patches/server/0819-Fix-kelp-modifier-changing-growth-for-other-crops.patch
@@ -6,6 +6,10 @@ Subject: [PATCH] Fix kelp modifier changing growth for other crops
 Also add growth modifiers for twisting vines, weeping vines, cave vines,
 and glow berries
 
+Also fix above-mentioned modifiers from having the reverse effect
+
+Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
+
 diff --git a/src/main/java/net/minecraft/world/level/block/CaveVinesBlock.java b/src/main/java/net/minecraft/world/level/block/CaveVinesBlock.java
 index 87dabe3c80b48bff52f2e3dbbaceb37a1a21e431..effee89e308c9a663938ac5b00a8c6512ce407c2 100644
 --- a/src/main/java/net/minecraft/world/level/block/CaveVinesBlock.java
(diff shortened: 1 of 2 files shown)


patches/server/0819-Fix-kelp-modifier-changing-growth-for-other-crops.patch
```diff
@@ -51,7 +55,7 @@ index def3b62feada5cebae4049883fa967b12f6f32b4..d3dde8ce3f55bda0b7f55491aef3bc9a
 +        } else {
 +            modifier = 100; // Above cases are exhaustive as of 1.18
 +        }
-+        if ((Integer) state.getValue(GrowingPlantHeadBlock.AGE) < 25 && random.nextDouble() < (100.0D / modifier) * this.growPerTickProbability) { // Spigot
++        if ((Integer) state.getValue(GrowingPlantHeadBlock.AGE) < 25 && random.nextDouble() < (modifier / 100.0D) * this.growPerTickProbability) { // Spigot // Paper - fix plant growth modifier
```
jpenilla:
```suggestion
+        if ((Integer) state.getValue(GrowingPlantHeadBlock.AGE) < 25 && random.nextDouble() < (modifier / 100.0D) * this.growPerTickProbability) { // Spigot // Paper - fix growth modifier having reverse effect
```

jpenilla (changes requested):

jpenilla (approved):

PR #6810, opened 2021-10-21
Fixed a bunch of missing method mappings

Machine-Maker:
Just found with a simple `\s\w\(` regex search, probably more out there.

### DIFF
--- a/build-data/mappings-patch.tiny
+++ b/build-data/mappings-patch.tiny
@@ -28,18 +28,16 @@ c	net/minecraft/server/level/WorldServer	net/minecraft/server/level/ServerLevel
 c	net/minecraft/world/level/SpawnerCreature$d	net/minecraft/world/level/NaturalSpawner$SpawnState
 	m	(Lnet/minecraft/world/entity/EnumCreatureType;I)Z	a	canSpawnForCategory
 
-# missed mapping?
-c	net/minecraft/world/level/block/MultifaceBlock	net/minecraft/world/level/block/MultifaceBlock
-	m	(Lnet/minecraft/world/level/block/state/IBlockData;Lnet/minecraft/world/level/IBlockAccess;Lnet/minecraft/core/BlockPosition;Lnet/minecraft/core/EnumDirection;)Lnet/minecraft/world/level/block/state/IBlockData;	c	getStateForPlacement
-
-# another missed one
-c	net/minecraft/server/players/UserCache	net/minecraft/server/players/GameProfileCache
-	m	(Ljava/lang/String;)Ljava/util/Optional;	getProfile	get
-		p	0		name
-
-# change dimension in ServerPlayer
+# CraftBukkit adds a TeleportCause to methods in ServerPLayer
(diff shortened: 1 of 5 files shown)


MiniDigger (approved):

build-data/mappings-patch.tiny
```diff
@@ -63,3 +70,78 @@ c	net/minecraft/world/entity/EntityInsentient	net/minecraft/world/entity/Mob
 # Craftbukkit overloads convertTo in Mob adding various API reasons
 c	net/minecraft/world/entity/EntityInsentient	net/minecraft/world/entity/Mob
 	m	(Lnet/minecraft/world/entity/EntityTypes;ZLorg/bukkit/event/entity/EntityTransformEvent$TransformReason;Lorg/bukkit/event/entity/CreatureSpawnEvent$SpawnReason;)Lnet/minecraft/world/entity/EntityInsentient;	a	convertTo
+
+# Craftbukkit changes the return type of DefaultDispenseItemBehavior#spawnItem
+c	net/minecraft/core/dispenser/DispenseBehaviorItem	net/minecraft/core/dispenser/DefaultDispenseItemBehavior
+	m	(Lnet/minecraft/world/level/World;Lnet/minecraft/world/item/ItemStack;ILnet/minecraft/core/EnumDirection;Lnet/minecraft/core/ISourceBlock;)Z	a	spawnItem
+
+# Missed mappings
+c	net/minecraft/core/dispenser/IDispenseBehavior$24	net/minecraft/core/dispenser/DispenserItemBehavior$24
```
jpenilla: `DispenserItemBehavior` doesn't exist, it's `DispenseItemBehavior`, additionally this mapping is already present in the generated mappings, the issue seems to maybe be with mercury not remapping it properly? That is unless the inner classes somehow got mismatched in the generated mappings. Either way this mapping is wrong because the inner class name is not 24 in both mappings.

jpenilla (changes requested):
I don't have time to review every single mapping here, and I think that actually outlines the problem with this PR, is the maintenance burden for having this many mapping patches which need to be manually reviewed and updated for each version.

I think rather than just grepping the source for missed mappings, we should focus on mapping things which we actually notice and use which would be nice to give a proper name to.

The methods where CraftBukkit added parameters are probably all fine, but they aren't excluded from needing checking and updating with even a minor revision bump. And stuff like the `DispenseItemBehavior` mapping I commented on should questionably be in the mappings patch at all.

build-data/mappings-patch.tiny
Machine-Maker: Huh, why does it actually work then? it does successfully remap the method name inside `DispenseItemBehavior`, either way ill change it. 

As for it not being 24, I just used intellij's structure view tab to see what number it was, and both have it as 24.

jpenilla: https://github.com/PaperMC/paperweight/pull/99

jpenilla (approved):

MiniDigger (approved):

PR #6813, opened 2021-10-21
Fix method name for Critical damage

Doc94:
Only a method name fix, based in comment from the commit where is changes are implemented.
https://github.com/PaperMC/Paper/commit/bb397ba74c034ffe5ae164d774a043cb7bca0472#r58405431


Proximyst (approved):

NoahvdAa (commented):
Patch probably needs to be rebuilt, was this edited by hand?

ghost (approved):

Doc94:
> Patch probably needs to be rebuilt, was this edited by hand?

hand.. but its only rename then in teory this not cause a issue in the patch not (? (based in build check in GitHub CI)

electronicboy (approved):

PR #6817, opened 2021-10-22
Custom biome API

Seggan:
This is probably a most controversial change, due to the scope of the change and due to the fact that I heard upstream was taking a look into this. Currently this is a draft PR so we can discuss about whether this PR is even needed or critique the API (I probably could have done this better). I also have not fully replaced the old `Biome` enum yet or tested this. This might be a job for the maintainers, not for 2nd time contributors like me, so I won't be surprised if this does get closed. TL;DR this is my own idea of a custom biome API

### Todo
- [ ] Test this
- [ ] Replace the `Biome` enum
- [ ] I really want to do away with `BiomeManager` somehow

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -246,3 +246,8 @@ public net.minecraft.world.entity.ai.attributes.AttributeSupplier instances
 
 # Add ItemFactory#getSpawnEgg API
 public net.minecraft.world.item.SpawnEggItem BY_ID
+
+# Biome API
+public net.minecraft.world.level.biome.Biome climateSettings
+public net.minecraft.world.level.biome.Biome$ClimateSettings
+public net.minecraft.world.level.biome.Biome$ClimateSettings temperatureModifier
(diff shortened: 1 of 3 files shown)


Machine-Maker:
So just a quick note on replacing the Biome enum, It appears upstream is going to do that, and already has a [PR replacing quite a few of the enums](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/931/overview) (except Material). So I'd hold of on trying to do that yourself.

Depending on whether we want to probably end up deprecating a lot of this once upstream's removal of the Biome enum takes place, this could be blocked until that happens.

Seggan:
I completely understand, in fact I did mention upstream's interest in this as a potential block in the description.

Owen1212055:
Thank you for your contribution, however, we will most likely be implementing things like this under our registry manipulation API with paper plugins.

We also most likely will be using spigot’s enum removal PR to be able to implement custom types.

If you’re interested in adopting this PR to use our new system, feel free to leave a comment.

PR #6474, opened 2021-08-23
ci: update setup-java action to v2.3.0

sksat:
Updates the setup-java action to v2.3.0 for cache option.

We can cache gradle dependencies by setup-java action now!
https://github.com/actions/setup-java/releases/tag/v2.3.0

PR #6819, opened 2021-10-23
ci: Update gh actions with caching [Supersedes 6474]

NotMyFault:
cc @MiniDigger https://github.com/PaperMC/Paper/pull/6474#issuecomment-904419334

Built in caching reduces the build duration by ~87%, see https://github.com/NotMyFault/Paper/actions/runs/1375290878

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,22 +16,13 @@ jobs:
                 java: [16]
             fail-fast: true
         steps:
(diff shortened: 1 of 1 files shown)


Proximyst (approved):
Hashing is done in the action, so I’m happy.
https://github.com/actions/setup-java/blob/ae26cabe43aaa84f8539613cdb171b33610fa704/src/cache.ts#L52

PR #6559, opened 2021-09-04
Prevent crashes or kicks from invalid enchantment keys

NoahvdAa:
Fixes #6558. I felt like this is worth patching since it can be abused on servers that give access to creative mode (or servers with some really scuffed command blocks). Swallowing the exception feels like the smallest and easiest solution as there doesn't seem to be a convenient way of checking if a namespacedkey is valid.

PR #6825, opened 2021-10-24
Prevent CraftNamespacedKey induced server crash

### LINKED ISSUE #6558: Server crash when player euqpiting armor with enchant "minecraft:"
### Stack trace

```
https://paste.gg/p/anonymous/d7ab86091ec14d20922fbbe2ef431617/files/76da774522f8406e9309a42f18c43be5/raw
```

### Plugin list

None

/give yourself a armor peice* with enchant "minecraft:" then euqpit it

### Paper version

This server is running Paper version git-Paper-249 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 6fa1983)
Previous version: git-Paper-1618 (MC: 1.12.2)

### Agreements

### Other

_No response_

kennytv:
Bukkit is Bukkit and does not accept empty paths, while Vanilla does. This doesn't actually fix the underlying parsing problem, but properly returns null for (what Bukkit thinks) are invalid locations, instead of throwing an error
Fixes #6558
Closes #6559

### DIFF
--- a/patches/server/0837-Fix-Bukkit-NamespacedKey-shenanigans.patch
+++ b/patches/server/0837-Fix-Bukkit-NamespacedKey-shenanigans.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Nassim Jahnke <jahnke.nassim@gmail.com>
+Date: Sun, 24 Oct 2021 15:49:35 +0200
+Subject: [PATCH] Fix Bukkit NamespacedKey shenanigans
+
+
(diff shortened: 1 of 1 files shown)


NoahvdAa (approved):

Machine-Maker:
This will probably result in items not stacking together right? when an item goes through any bukkit conversion, that invalid resource location is going to be stripped from the item. So it wont stack with other items that havent. Probably a non-issue, but thought I'd mention it.

kennytv:
Yes, tho this will outright remove those empty named enchantments from being saved so the server doesn't just crash, but *actually* fixing this (not just for enchantments) would require the NamespacedKey key regex to be changed, which has one too many implications on all of Bukkit for my taste, and would be a *~~never~~* post-hardfork thing

Proximyst (approved):

PR #3628, opened 2020-06-26
Add Raw Byte Entity Serialization (1.16)

Proximyst:
Adds an unsafe API (feel free to add a "safe" one like last time, Aikar) to serialize & deserialize entity data.

PR #6826, opened 2021-10-24
Add Raw Byte Entity Serialization

JRoy:
Updated version of #3628

https://user-images.githubusercontent.com/10731363/138611513-228d5683-0ff3-4632-8edf-bc0680b09dba.mp4

hacky test code 
```java
  private byte[] data = null;

  public void on(EntityDamageByEntityEvent event) {
    if (event.getDamager() instanceof Player) {
      data = Bukkit.getUnsafe().serializeEntity(event.getEntity());
    }
  }

  @EventHandler()
  public void on(ChatEvent event) {
    if (data != null) {
      Bukkit.getUnsafe().deserializeEntity(data, event.getPlayer().getWorld()).spawnAt(event.getPlayer().getLocation());
    }
  }
```

### DIFF
--- a/patches/server/0399-Add-Raw-Byte-ItemStack-Serialization.patch
+++ b/patches/server/0399-Add-Raw-Byte-ItemStack-Serialization.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add Raw Byte ItemStack Serialization
 Serializes using NBT which is safer for server data migrations than bukkits format.
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java b/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
-index f8cb210390958ddba9f9685f2ec1f8bb91690162..6b79b1ea43b713105f37de517019c033477f9835 100644
+index f8cb210390958ddba9f9685f2ec1f8bb91690162..3b727b96c4166e88ec23df3e3e1e98d6aed58637 100644
 --- a/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
 +++ b/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
-@@ -378,6 +378,46 @@ public final class CraftMagicNumbers implements UnsafeValues {
+@@ -378,6 +378,53 @@ public final class CraftMagicNumbers implements UnsafeValues {
      public boolean isSupportedApiVersion(String apiVersion) {
          return apiVersion != null && SUPPORTED_API.contains(apiVersion);
      }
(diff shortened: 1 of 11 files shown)


patches/server/0399-Add-Raw-Byte-ItemStack-Serialization.patch
```diff
@@ -19,9 +19,23 @@ index f8cb210390958ddba9f9685f2ec1f8bb91690162..6b79b1ea43b713105f37de517019c033
...
++    }
++
++    @Override
++    public ItemStack deserializeItem(byte[] data) {
++        Preconditions.checkNotNull(data, "null cannot be deserialized");
++        Preconditions.checkArgument(data.length > 0, "cannot deserialize nothing");
++
++        CompoundTag compound = deserializeNbtFromBytes(data);
++        int dataVersion = compound.getInt("DataVersion");
++        Dynamic<Tag> converted = DataFixers.getDataFixer().update(References.ITEM_STACK, new Dynamic<Tag>(NbtOps.INSTANCE, compound), dataVersion, getDataVersion());
```
Proximyst, line 32: Should probably use DataConverter now.

electronicboy:
biggest concern I had with this kinda thing was the UUID, like, that should ideally be treated as like, something controlled via a boolean rather than just straight up enforced that it yeets UUIDs

lynxplay:
Should this potentially document the event related behaviour of the `spawnAt` methods a bit more in detail ?

- The full range of events will be called and respected.
- No spawn attempt will be made / no events called if the entity is already spawned somewhere.

JRoy:
> Should this potentially document the event related behaviour of the `spawnAt` methods a bit more in detail ?
> 
> * The full range of events will be called and respected.
> * No spawn attempt will be made / no events called if the entity is already spawned somewhere.

ok

patches/api/0343-Add-Raw-Byte-Entity-Serialization.patch
```diff
@@ -0,0 +1,62 @@
...
++     * @return Whether the entity was successfully spawned.
++     */
++    public default boolean spawnAt(@NotNull Location location) {
++        return spawnAt(location, org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason.DEFAULT);
++    }
++
++    /**
++     * Spawns the entity in the world at the given {@link Location} with the reason given.
++     * <p>
++     * This will not spawn the entity (and still return true) if the entity is already spawned or has previously been despawned.
```
lynxplay: Is this actually the case ?
Looking at 
```java
return !entity.valid && entity.level.addEntity(entity, reason);
```
a) This will not return true if the entity is already being tracked (`valid == true`)
b) Afaik entities that are despawned from the world are actually "invalidated" ([see here](https://github.com/PaperMC/Paper/blob/master/patches/server/0072-Entity-AddTo-RemoveFrom-World-Events.patch#L22))  and *should* hence *could* be spawned with this ?

Proximyst (approved):

PR #6829, opened 2021-10-24
Update jline and TCA

electronicboy:
updates jline to 3.21.0, and TCA to 1.3.0

There is some considerations here given that now the ASM library is bundled with the mojang server, we could maybe go back to using the JNA version of jline, which, IIRC, worked better in some areas vs jline, I forget if there was another reason for switching to jline, given the years of terminals being terminals it's hard to remember what were the implications of all solutions available, windows being a common breaking ground here...

maybe want to perform the typical "has it broken an environment we care about" tests...

jar available here in the next few minutes: https://keybase.pub/electronicboy/builds/paper/versions/1.17.1/Paper-1.17.1-R0.1-SNAPSHOT-update-jline.jar

### DIFF
--- a/patches/server/0137-Use-TerminalConsoleAppender-for-console-improvements.patch
+++ b/patches/server/0137-Use-TerminalConsoleAppender-for-console-improvements.patch
@@ -19,7 +19,7 @@ Other changes:
     configuration
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 1f4c0f2156c588fcbb3b0329a416dc20a4b355a8..9b70376813531718c02082633e9f8105f4879a63 100644
+index 1f4c0f2156c588fcbb3b0329a416dc20a4b355a8..e2ec23e597a3519550251b9f150bad8795b29a48 100644
 --- a/build.gradle.kts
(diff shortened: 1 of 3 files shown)


MiniDigger:
where taco man with his stack of test VMs of obscure OSs when you need him :/

zml2008:
Yeah, last time I looked at the options of terminals i went with jansi just cuz it was smaller

we've been using this version of jline in Sponge for quite a while. However, looking at the newer releases, it looks like 3.21.0 updates jansi to a version that fixes an issue with apple silicon and adds actual natives for aarch64, you'll probably want that

electronicboy:
I apparently brain farted and went to 3.20 instead of 3.21, thanks for pointing that out zml; Just updated, jar is uploading now (same link), will be there in a few (currently uploading, keybase like 10s, 9, 8, 7, 6, 10....)

electronicboy:
We've had a few people test various setups and no complaints thus far nor do I really expect any issues from this, just kinda like having a few quick tests of common environments, think we'll just merge this as-is, am sure somebody will show up in a week or twos time about some OS environment we've never heard of before

PR #6831, opened 2021-10-25
Fix issues with mob conversion

Machine-Maker:
* PiglinAbstract docs were incorrect for setConversionTime
* converting skeletons to strays would spam EntityTransformEvent if the event was cancelled
* converting piglins in the overworld had the same spam-event result
* upstream used the wrong method when setting a skeletons conversion time


kennytv:
Not sure I like resetting the timer since that's a bit of a behavioral break

Machine-Maker:
> Not sure I like resetting the timer since that's a bit of a behavioral break

That’s what the event does for other conversions. Upstream just didn’t add that for the newer things. I think it’s consistent.

kennytv (approved):

Issue #6812, opened 2021-10-21
setFrozen or freezetick similar to firetick


Boy0000:
### Is your feature request related to a problem?

Unlike with firetick, the freeze effect is only applied when inside powder snow for x ticks. Simply repeatedly setting the freezetick to max for a player doesnt solve it.

### Describe the solution you'd like.

Some way to specify that an entity is frozen, as if they were inside a powder snowblock with max freezetick.
There is already entity.isFrozen, but it only returns if it is frozen or not.

This would ofcourse not be removed when not in a powder snow block, and be a constant effect until it is "false"

### Describe alternatives you've considered.

Constantly setting the entities freezetick to max (doesnt work)
Sending a blockchange to the location of the entity

### Other

![136946145-73542c0f-9f03-4a20-b234-ffed604c8142](https://user-images.githubusercontent.com/62521371/138312993-b207a45f-c888-4aaf-a4d6-1ae71d4c17e3.gif)
When constantly setting freezeticks, this is the result.

Machine-Maker:
For someone who looks into this, I think a TriState would be applicable here. To essentially override the default isFullyFrozen check with true, false, or not set, which falls back on the vanilla mechanics.

PR #6833, opened 2021-10-25
Entity powdered snow API

Machine-Maker:


Boy0000:
Would this sort out #6812 by letting u say that `isInPowderedSnow = true` or is it only for returning the state?
Freeze effect is very cool but it isn't easy to apply to the player

Machine-Maker:
> Would this sort out #6812 by letting u say that `isInPowderedSnow = true` or is it only for returning the state? Freeze effect is very cool but it isn't easy to apply to the player

That flag is reset every tick. So setting it for an entity wouldn't really only have an effect for part of that entities tick. What you are looking for would require some more work, that's out of the scope of that PR.

kennytv (approved):

PR #6837, opened 2021-10-26
Allow spawners to be disabled

HexedHero:
Allows mob spawners to be disabled without adding a really high tick rate.
Useful for:
- Hubs/Lobby servers and creative servers that use/allow mob spawners to be used for decoration.
- Plugins that take complete control over mob spawners and just cancel any and all mob spawner events, adding their own spawn system on top.

I found that with plugins that disable all related events, spawners were taking upwards of 4-6% on a spawner heavy server even with a very high configured tick rate on timings reports and further inspection with Spark showed the cost was from the spawner server tick doing AABB checks and EntityTypes#create costing performance.

I'll make a PaperDocs PR to mention this change if this gets approved/merged.


Proximyst (approved):

MiniDigger (approved):
pending docs PR, else looks good

Strahilchu:
This is nice, always good to have such options

HexedHero:
Made https://github.com/PaperMC/PaperDocs/pull/124 for this

PR #6841, opened 2021-10-28
Deduplicate strings in ObfHelper

jpenilla:
Also use mapping-io instead of tiny-mappings-parser, and replace Guava immutable collections with JDK native ones.


MiniDigger:
whats the difference between mapping-io and tiny-mappings-parser?

edit: answered myself https://github.com/FabricMC/mapping-io/issues/1

MiniDigger (approved):

PR #6842, opened 2021-10-28
[ci skip] Fix incorrect javadoc for Mob pathfinding API

NoahvdAa:
The javadoc mentions the method returning null if the pathfinding is done, but `int`'s can obviously never be null. A different solution would be to update the method to return an `Integer`, and null if the pathfinding is done, although that would break backwards compatibility.

Thanks to @andrew121410 for spotting this [on Discord](https://canary.discord.com/channels/289587909051416579/555462289851940864/903357214415290419).


Machine-Maker (commented):
Maybe instead of `a number equal to or higher than to the amount of points` which has a typo I think (or higher than to the), I think mentioning something about the size of getPoints, like getNextPointIndex may be greater than the max index of getPoints.

Machine-Maker (approved):

NoahvdAa:
Rebased for 1.18

Machine-Maker (approved):

NoahvdAa:
Rebased.

electronicboy (approved):

PR #6850, opened 2021-10-31
Fix Lure infinite loop

HexedHero:
Fixes an issue where the Lure Enchantment reduction amount in the fishing calculations would make the timeUntilLured value go to 0 or into the negatives causing an infinite loop fixing an issue where if a server owner changes the max fishing wait time in their paper.yml to a value of 300/200/100 or less will break Lure 3/2/1+. This patch also fixes Lure 6+ with the default values. This will now simply put the time until the next catch to 1 tick instead of the infinite loop happening.

From looking at the patch you may get confused about why aren't we checking the random generated timeUntilLured value just above, it's a little confusing but Minecraft in vanilla will sometimes get negative numbers from the lure reduction and will simply do another loop until it has a positive one, this patch only fixes the cases where a positive number can **never** show up causing the loop while keeping normal behaviour.


Machine-Maker (approved):
I think this should go into the paper patch that adds the configuration options for the min/max wait times as that increases the visibility of the issue in the first place

HexedHero:
Moved to configurable fishing time patch

electronicboy (approved):

Machine-Maker (approved):

PR #6854, opened 2021-10-31
Call onRemove logic for breakNaturally

### LINKED ISSUE #6874: breakNaturally() with Containers
### Expected behavior

Hey, I recently found a Bug when using the Block#breakNaturally() function in combination with containers!
Normally the container should drop all his content when using this function, but with 1.17.1 only the chest gets dropped!

### Observed/Actual behavior

Now, when you are using the Block#breakNaturally() function, only the chest drops and all stuff in there is gone ;D

### Steps/models to reproduce

1. Place a Chest and put stuff in there!
2. Use the Block#breakNaturally() function to break it!
3. Only the Chest will drop!

### Plugin and Datapack List

Any Plugin which uses this function in combination with Containers :D

### Paper version

This server is running Paper version git-Paper-370 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 29bb5a9)

### Other

Thanks for looking into the Problem ;D

Machine-Maker:
Addresses https://github.com/EngineHub/WorldEdit/issues/1924

breakNaturally previously wasn't calling the block's onRemove behavior, which includes dropping items in containers and more. I didn't put the logic inside CraftBlock#setTypeAndData because all the other calls to setTypeAndData shouldn't be calling this anyways, so I think having it as part of breakNaturally is fine.

### DIFF
--- a/patches/server/0843-Call-onRemove-logic-for-breakNaturally.patch
+++ b/patches/server/0843-Call-onRemove-logic-for-breakNaturally.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 31 Oct 2021 14:49:43 -0700
+Subject: [PATCH] Call onRemove logic for breakNaturally
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
+index 5154cfffc414e1f6039e55f1a256bbaacb56bc55..6f01af8cc3f9ed4d2eaa3304990ca33f8692a453 100644
+--- a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
++++ b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
+@@ -513,8 +513,14 @@ public class CraftBlock implements Block {
+             if (triggerEffect) world.levelEvent(org.bukkit.Effect.STEP_SOUND.getId(), position, net.minecraft.world.level.block.Block.getId(block.defaultBlockState())); // Paper
+             result = true;
+         }
++        // Paper start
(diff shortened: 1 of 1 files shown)


patches/server/0835-Call-onRemove-logic-for-breakNaturally.patch
```diff
@@ -0,0 +1,26 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
++++ b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
+@@ -508,8 +508,14 @@ public class CraftBlock implements Block {
+             if (triggerEffect) world.levelEvent(org.bukkit.Effect.STEP_SOUND.getId(), position, net.minecraft.world.level.block.Block.getId(block.defaultBlockState())); // Paper
+             result = true;
+         }
++        // Paper start
++        net.minecraft.world.level.block.state.BlockState newBlockState = Blocks.AIR.defaultBlockState();
++        if (this.world instanceof net.minecraft.world.level.Level level) {
++            this.getNMS().onRemove(level, position, newBlockState, false);
```
Owen1212055: I am curious, is there any reason why you are getting the block state again here? It seems a bit higher it's stored in a variable already.

```java
  // Paper end
        // Order matters here, need to drop before setting to air so skulls can get their data
        net.minecraft.world.level.block.state.BlockState iblockdata = this.getNMS();
        net.minecraft.world.level.block.Block block = iblockdata.getBlock();
        net.minecraft.world.item.ItemStack nmsItem = CraftItemStack.asNMSCopy(item);
        boolean result = false;
```

Doc94:
Maybe i dont understand.. but how to work breakNaturally with this PR?

Machine-Maker:
> but how to work breakNaturally with this PR?  

I'm not sure I'm following what you are asking here, this just adds more functionality to breakNaturally making it behave more like an actual block break.

lynxplay:
Resolves: #6874

cksmith51:
Any chance we could get reviews for this? Not trying to nag or anything but this particular bug has been causing issues for me longer than I realized. Thanks in advance.

MiniDigger (approved):

PR #6864, opened 2021-11-03
Configuration option for showing sleep messages

### LINKED ISSUE #6863: Add option to disable "x/y players sleeping" message.

Minecraft 1.17 added new feature - it sends a message to all players when player goes to the bed. That function ignores vanished players and that ones in creative mode which leaks the number of online administrations on the server.

obraz

Add option in config or gamerule to disable that message.

Welp, there is no option built-in in game server to disable that.

### Other

_No response_

lynxplay:
Configuration option for showing sleep messages

The minecraft server now announces sleep progress, e.g. players already
sleeping or the night being skipped, to all players on the server
through a hotbar messages.
While this vanilla behaviour works well in vanilla itself it is not
always appreciated, especially when hiding players.

This commit introduces a new sleep configuration node in the player
world configuration that allows server administrators to prevent the
server from sending either one of the messages to the clients connected.

Resolves: #6863

---------------------------------------------------------------------------------------------------

A <ins>**PR to the PaperDocs**</ins> will follow once this PR has been approved/finalized to the point of approval.

Additionally a few questions I ran into while making this:

- Should events exist to properly remove individual players from the player count ? 
- Right now a parent node `sleep` was added as well as its only child node `messages`. This could be reduced to `sleep-messages`.

### DIFF
--- a/patches/server/0836-Configuration-option-for-showing-sleep-messages.patch
+++ b/patches/server/0836-Configuration-option-for-showing-sleep-messages.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bjarne Koll <lynxplay101@gmail.com>
+Date: Wed, 3 Nov 2021 18:55:54 +0100
+Subject: [PATCH] Configuration option for showing sleep messages
+
+The minecraft server now announces sleep progress, e.g. players already
+sleeping or the night being skipped, to all players on the server
+through a hotbar messages.
+While this vanilla behaviour works well in vanilla itself it is not
+always appreciated, especially when hiding players.
+
+This commit introduces a new sleep configuration node in the player
+world configuration that allows server administrators to prevent the
+server from sending either one of the messages to the clients connected.
+
+Resolves: #6863
+
+diff --git a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
+index c2d8294ffeff017a6ec9e7725b50eaef8eb75dfd..cdad9e1447ea2ac781b2539b09938eea2875fba1 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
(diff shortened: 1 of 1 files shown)


Doc94:
About `Should events exist to properly remove individual players from the player count ?` maybe can be useful a parameter in Player like ignoreInCountForSleep (best name in the world) and modify sleepStatus when generate the count.

now im thinking.. in API not exists a method for get players need for sleep?

MiniDigger:
really not a fan of disabling a vanilla feature like this for a bandaid fix. much rather would have an event for the announcing, which exposes the counts and and maybe even the message. then the vanish plugin can hook into that.

lynxplay:
> really not a fan of disabling a vanilla feature like this for a bandaid fix. much rather would have an event for the announcing, which exposes the counts and and maybe even the message. then the vanish plugin can hook into that. 

Yeah I figured as much, thanks for the feedback I'll update the PR shortly.

Prof-Bloodstone:
I think people who use sleep datapacks/plugins could still find such option useful.
Though proper fix would be to expose API for canceling the message and update your sleep plugin to support it.

lynxplay:
> I think people who use sleep datapacks/plugins could still find such option useful.
> Though proper fix would be to expose API for canceling the message and update your sleep plugin to support it. 

My main reason to at least attempt a config option was the fact that events will just generate a bunch of different plugins to disable the feature in the end anyway, spreading out something that could be standardized in the server software.

But I guess it is a way cleaner fix for us to leave this functionality to the community to implement as we can solve both issues (disable completely and modify) with events.

MiniDigger:
well, maybe we should have both then. but I for sure want the event so vanish plugins can handle this properly.
its also a nice "dont wanna expose vanished players when sleeping? use paper to this plugin can hide them from there"

Prof-Bloodstone:
Fair. We could have the config option cause the event to fire in already canceled state - then we can solve both of these use cases at once.

lynxplay:
I absolutely agree :+1: I'll close the PR for now as I believe the event layout needs a bit of thinking (e.g. single event ? multiple events ? how will the status state be represented ?) that I do not have the time for to finish before sunday, so if others want to give it a try, go for it :+1: 

I'll be opening a respective PR with the above suggestions next week if no alternative solution has been proposed.

PR #6870, opened 2021-11-04
Add isCollidable methods to various places

Machine-Maker:
isSolid isn't a good representation of what a entity can stand on, this should be better. For example, an Oak Sign block returns true for isSolid, which it doesnt collide with an entity at all.

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -249,3 +249,6 @@ public net.minecraft.world.item.SpawnEggItem BY_ID
 
 # Zombie API - breaking doors
 public net.minecraft.world.entity.monster.Zombie supportsBreakDoorGoal()Z
+
+# Add Material#hasCollision
+public net.minecraft.world.level.block.state.BlockBehaviour hasCollision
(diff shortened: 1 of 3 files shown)


electronicboy (approved):

MWHunter:
+1

I use the isSolid method and I have to manually set plates, signs, and corals as not solid.  And potted plants, carpet, diodes, candles, and heads as solid.  And also slime blocks, repeaters, snow, anvils, lily pads, sea pickles, turtlee eggs, chorus flower, chorus plants, ladders, end rods, scaffolding, cocoa, sculk sensors, powder snow, big dripleaf, azalea, flowering azalea, and pointed dripstone as solid.  All these places are where isSolid isn't isCollidable.

It gets annoying manually diffing isSolid() and the NMS collidable every update and this PR would help clean up my code - if I were to drop spigot support.

me4502 (approved):

me4502:
pls rebase 🥺

PR #6876, opened 2021-11-06
Make BlockData creation thread safe again

PierreSchwang:
As the caching of already created `CraftBlockData`-instances makes `CraftBlockData#newData` overall faster and more efficient, it breaks the thread safety of `CraftBlockData#newData`.

This change for the patch aims to re-establish the thread safety for plugins calling `Bukkit#createBlockData` in an asynchronous context. As `CraftBlockData#newData` does not modify anything server-related the method should stay thread-safe. Calling the named method asynchronous is a well-known pattern for plugins like world generators, where they put the load for BlockData creation onto other threads to ensure a smooth experience. 

Also relates to https://hub.spigotmc.org/jira/projects/SPIGOT/issues/SPIGOT-6784

### DIFF
--- a/patches/server/0502-Cache-block-data-strings.patch
+++ b/patches/server/0502-Cache-block-data-strings.patch
@@ -17,29 +17,15 @@ index 95842327aa08d4717f86e9dcc0519ab24c41ca14..135b3e44fb6054d360327a0ce46decc4
  
          if (this.isSameThread()) {
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java b/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java
-index 3594f432a25b580173e8577bf324be954f5eddd1..dc51181dde6e49d2cbed9f21c007b8256a8910c6 100644
+index 3594f432a25b580173e8577bf324be954f5eddd1..d62ba8d02228b2a00202190ff16a0efc342c7b83 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java
 +++ b/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java
-@@ -8,11 +8,13 @@ import com.mojang.brigadier.exceptions.CommandSyntaxException;
- import java.util.HashMap;
- import java.util.Map;
- import java.util.Set;
-+import java.util.concurrent.ConcurrentHashMap;
- import java.util.stream.Collectors;
- import net.minecraft.commands.arguments.blocks.BlockStateParser;
- import net.minecraft.core.Direction;
- import net.minecraft.core.Registry;
- import net.minecraft.nbt.CompoundTag;
-+import net.minecraft.resources.ResourceLocation;
- import net.minecraft.util.StringRepresentable;
- import net.minecraft.world.level.block.Block;
- import net.minecraft.world.level.block.state.BlockState;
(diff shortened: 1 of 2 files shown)


patches/server/0521-Cache-block-data-strings.patch
```diff
@@ -17,23 +17,29 @@ index 4af56ca699d456eb23794673861323ab06cf6de8..bb18848f2b91b80ab5099c00b423738b
...
-index 3594f432a25b580173e8577bf324be954f5eddd1..6dc8f9f269db6971b8b46819e017357899ccd118 100644
+index 3594f432a25b580173e8577bf324be954f5eddd1..dc51181dde6e49d2cbed9f21c007b8256a8910c6 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java
 +++ b/src/main/java/org/bukkit/craftbukkit/block/data/CraftBlockData.java
-@@ -13,6 +13,7 @@ import net.minecraft.commands.arguments.blocks.BlockStateParser;
+@@ -8,11 +8,13 @@ import com.mojang.brigadier.exceptions.CommandSyntaxException;
+ import java.util.HashMap;
+ import java.util.Map;
+ import java.util.Set;
++import java.util.concurrent.ConcurrentHashMap;
```
lynxplay: While the patch suggests otherwise, this should be a fully qualified import instead.
Maybe while fixing this patch, you could just remove the other import as well to clean up the patch.

A simple 
```patch
- ResourceLocation key = Registry.BLOCK.getKey(block);
+ var key = Registry.BLOCK.getKey(block);
``` 
would suffice already to also removed the resource location import.

lynxplay (changes requested):

patches/server/0521-Cache-block-data-strings.patch
PierreSchwang: Good point, makes merging with upstream easier

Machine-Maker (approved):

PaulBGD (approved):

svdgoor:
Could someone here confirm the issue our plugin is having, is caused by this?
I can run the tests after this patch is merged, as well.

PierreSchwang:
> Could someone here confirm the issue our plugin is having, is caused by this?
> I can run the tests after this patch is merged, as well.

Based on the linked stacktrace
`java.util.ConcurrentModificationException`, yeah, that's the issue and my reason for this fix

svdgoor:
Perfect, thank you!

MiniDigger (approved):

patches/server/0521-Cache-block-data-strings.patch
```diff
@@ -54,7 +46,7 @@ index 3594f432a25b580173e8577bf324be954f5eddd1..6dc8f9f269db6971b8b46819e0173578
 +        if (material != null) {
 +            Block block = CraftMagicNumbers.getBlock(material);
 +            if (block != null) {
-+                ResourceLocation key = Registry.BLOCK.getKey(block);
++                var key = Registry.BLOCK.getKey(block);
```
kennytv: No reason to use var here, please use a fully qualified import. just by reading the patch it's not obvious this is a resourcelocation, and it's not like it saves all that much space here

kennytv (dismissed):

PierreSchwang:
I _think_ something is f* up rn - but idk why that happens tho 👀 - Maybe this can be fixed during merge? yikes

kennytv (approved):

PR #6881, opened 2021-11-10
cleanup: Add hand to bucket events patch

Machine-Maker:
I noticed a lot of diff in this patch that could easily be fixed by just changing some method param types to ServerLevel instead of Level. 

The whole getSharedSpawnPos method was moved up to Level and didnt need to be at all.

If its an issue, I can add back a method with the same signature to Level that then gets overriden by the one in ServerLevel. who knows if some plugin calls it for some reason.


kennytv (approved):

kennytv:
... just needs to remove the mapping in `reobf-mappings-patch.tiny`

PR #6882, opened 2021-11-11
rework bstats to use metrics base

MiniDigger:
* removes a ton of our custom code and replaces it with the metrics base lib
* if we create the bstats config, we now send a bit of info to inform users that metrics are opt-out
* I think this fixes the "outdated" versions chart on the bstats global page, as we never did send bukkitVersion and bukkitName

### DIFF
--- a/patches/server/0008-Paper-Metrics.patch
+++ b/patches/server/0008-Paper-Metrics.patch
@@ -13,39 +13,53 @@ tangible effect on server performance. The data is used to allow the authors of
 PaperMC to track version and platform usage so that we can make better management
 decisions on behalf of the project.
 
+diff --git a/build.gradle.kts b/build.gradle.kts
+index 4db0cc3f8505747e77d314320545eb71904b4eac..6baabc30a363d132ea3d8a7da54a40aaf918f15b 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -16,6 +16,7 @@ dependencies {
+     implementation("org.apache.logging.log4j:log4j-iostreams:2.14.1") // Paper
+     implementation("org.ow2.asm:asm:9.2")
+     implementation("org.ow2.asm:asm-commons:9.2") // Paper - ASM event executor generation
++    implementation("org.bstats:bstats-base:2.2.1") // Paper
+     runtimeOnly("org.xerial:sqlite-jdbc:3.36.0.3")
+     runtimeOnly("mysql:mysql-connector-java:8.0.27")
+ 
+@@ -65,6 +66,7 @@ relocation {
+     relocate("org.bukkit.craftbukkit" to "org.bukkit.craftbukkit.v$packageVersion") {
+         exclude("org.bukkit.craftbukkit.Main*")
+     }
++    relocate("org.bstats:bstats-base" to "org.bstats") // Paper
+ }
+ 
+ tasks.shadowJar {
 diff --git a/src/main/java/com/destroystokyo/paper/Metrics.java b/src/main/java/com/destroystokyo/paper/Metrics.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..e3b74dbdf8e14219a56fab939f3174e0c2f66de6
+index 0000000000000000000000000000000000000000..6402516b8b1a04489184dc2adde83d6cbc5e83d5
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/Metrics.java
-@@ -0,0 +1,670 @@
+@@ -0,0 +1,236 @@
 +package com.destroystokyo.paper;
 +
++import java.io.File;
++import java.io.IOException;
++import java.util.HashMap;
++import java.util.Map;
++import java.util.UUID;
++import java.util.regex.Matcher;
++import java.util.regex.Pattern;
 +import net.minecraft.server.MinecraftServer;
++import org.bstats.MetricsBase;
++import org.bstats.charts.DrilldownPie;
++import org.bstats.charts.SimplePie;
++import org.bstats.charts.SingleLineChart;
++import org.bstats.json.JsonObjectBuilder;
 +import org.bukkit.Bukkit;
 +import org.bukkit.configuration.file.YamlConfiguration;
 +import org.bukkit.craftbukkit.util.CraftMagicNumbers;
 +import org.bukkit.plugin.Plugin;
-+
-+import org.json.simple.JSONArray;
-+import org.json.simple.JSONObject;
-+
-+import javax.net.ssl.HttpsURLConnection;
-+import java.io.ByteArrayOutputStream;
-+import java.io.DataOutputStream;
-+import java.io.File;
-+import java.io.IOException;
-+import java.net.URL;
-+import java.util.*;
-+import java.util.concurrent.Callable;
-+import java.util.concurrent.Executors;
-+import java.util.concurrent.ScheduledExecutorService;
-+import java.util.concurrent.TimeUnit;
-+import java.util.logging.Level;
-+import java.util.logging.Logger;
-+import java.util.regex.Matcher;
-+import java.util.regex.Pattern;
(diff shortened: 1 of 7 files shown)


patches/server/0008-Paper-Metrics.patch
```diff
@@ -54,643 +59,144 @@ index 0000000000000000000000000000000000000000..e3b74dbdf8e14219a56fab939f3174e0
...
++            config.isLogSentDataEnabled(),
++            config.isLogResponseStatusTextEnabled()
++        );
++        if (old != null) {
++            System.setProperty("bstats.relocatecheck", old);
++        }
++
++        if (!config.didExistBefore()) {
++            // Send an info message when the bStats config file gets created for the first time
++            logger.info("Velocity and some of its plugins collect metrics"
```
NoahvdAa:
```suggestion
+            logger.info("Paper and some of its plugins collect metrics"
```

MiniDigger: saw before you saw it :P
thanks

MiniDigger:
```json
{
    "playerAmount": 0,
    "onlineMode": 1,
    "bukkitVersion": "git-Paper-\"93f8fc5\" (MC: 1.17.1)",
    "bukkitName": "Paper",
    "javaVersion": "16",
    "osName": "Windows 10",
    "osArch": "amd64",
    "osVersion": "10.0",
    "coreCount": 24,
    "service": {
        "id": 580,
        "customCharts": [
            {
                "chartId": "minecraft_version",
                "data": {
                    "value": "1.17.1"
                }
            },
            {
                "chartId": "legacy_plugins",
                "data": {
                    "values": {
                        "0 ?": {
                            "0": 1
                        }
                    }
                }
            },
            {
                "chartId": "online_mode",
                "data": {
                    "value": "online"
                }
            },
            {
                "chartId": "paper_version",
                "data": {
                    "value": "git-Paper-\"93f8fc5\""
                }
            },
            {
                "chartId": "java_version",
                "data": {
                    "values": {
                        "Java 16": {
                            "16": 1
                        }
                    }
                }
            }
        ]
    },
    "serverUUID": "c6aa11bc-2f20-46de-8808-4cb98dce6518",
    "metricsVersion": "2.2.1"
}
```

NoahvdAa:
```json
                "chartId": "legacy_plugins",
                "data": {
                    "values": {
                        "0 ?": {
                            "0": 1
                        }
                    }
                }
```
is :sunglasses: being properly encoded here, since it's showing up as a question mark?

MiniDigger:
mmh, idk, I wanna blame the logging (edit: yeah, internally its fine)

also gotta relocate stuff, idk why I did to choose to ignore the relocate check, lmao

MiniDigger:
ok, should be good to go now, ideally @Bastian can take a look, that would be very much appreciated, especially if this is right, because we didn't do that before https://github.com/PaperMC/Paper/blob/rework-bstats/patches/server/0008-Paper-Metrics.patch#L190-L193

Bastian:
Sending this extra data has zero effect and are just ignored by bStats.
The only fields parsed by bStats are "coreCount", "osArch", "osVersion", and "os" . Everything else is a default chart.

patches/server/0008-Paper-Metrics.patch
```diff
@@ -54,643 +67,140 @@ index 0000000000000000000000000000000000000000..e3b74dbdf8e14219a56fab939f3174e0
...
-+
-+        JSONArray pluginData = new JSONArray();
-+        pluginData.add(getPluginData());
-+        data.put("plugins", pluginData);
++    private MetricsBase metricsBase;
 +
++    public Metrics() {
++        File bStatsFolder = new File((File) MinecraftServer.getServer().options.valueOf("plugins"), "bStats");
++        File configFile = new File(bStatsFolder, "config.yml");
++        MetricsConfig config;
```
Bastian: Unfortunately, you cannot use the MetricsConfig. You still have to create and parse the config file yourself.
The `MetricsConfig` is not a YAML file but uses its own format that looks like this and is not compatible with the existing config files:
```
# bStats (https://bStats.org) collects some basic information for plugin authors, like
# how many people use their plugin and their total player count. It's recommended to keep
# bStats enabled, but if you're not comfortable with this, you can turn this setting off.
# There is no performance penalty associated with having metrics enabled, and data sent to
# bStats is fully anonymous.");
enabled=true
server-uuid=<UUID>
log-errors=false
log-sent-data=false
log-response-status-text=false
```
(see source: [MetricsConfig.java](https://github.com/Bastian/bStats-Metrics/blob/master/base/src/main/java/org/bstats/config/MetricsConfig.java))

This config is only intended for new platforms and cannot be used by Paper for legacy reasons.
For the config, you should use the Bukkit Metrics class as a reference: [Metrics.java#L39-L64](https://github.com/Bastian/bStats-Metrics/blob/master/bukkit/src/main/java/org/bstats/bukkit/Metrics.java#L39-L64)

MiniDigger: ahh, seeing that bukkit didn't use it confused me, that makes sense.
alternatively we could migrate existing files? ah no, cause other plugins, meh. ok, will fix later

Strahilchu:
Wait you guys keep bstats and metrics enabled?

Bastian:
> Wait you guys keep bstats and metrics enabled?

bStats is opt-out on any platform except Sponge.

MiniDigger:
> Wait you guys keep bstats and metrics enabled?

bstats always has been and always will be enabled by default on paper, but you can easily disable it before it first sends data. This PR will even make it send a headsup message on first start

metrics on bstats are **__extremly__** valuable to us, it's how we decide which java/mc versions to support for example, or see how quickly new (paper/mc/java) versions are addapted. It's also valuable to get an estimate on market share.

They are also fully anonymous, don't disclose any personal data, and will not create any lag (obviously, lol), so there is no reason to disable them, ever, altho we do respect that choice if you wanna make that (altho we wouldn't understand it).

(side-note, if you are concerned about this, go audit all your plugins, the plugins that don't use bstats for usage stats often collect intrusive information like server name, server IP, plugin names etc, those we would never allow happening to our users)

Bastian (approved):
Looks good to me.

If you didn't already, I would recommend you to build it once and enable the `logSentData` and `logResponseStatusText` just to be sure that the correct data is sent and the bStats backend accepts it.

MiniDigger:
> Looks good to me.
> 
> If you didn't already, I would recommend you to build it once and enable the `logSentData` and `logResponseStatusText` just to be sure that the correct data is sent and the bStats backend accepts it.

thanks for taking a look!
did that already :)

MiniDigger:
somebody with more knowledge about bundler and stuff needs to figure out relocation and shadow here please

patches/server/0008-Paper-Metrics.patch
```diff
@@ -13,39 +13,53 @@ tangible effect on server performance. The data is used to allow the authors of
...
+     implementation("org.ow2.asm:asm-commons:9.2") // Paper - ASM event executor generation
++    implementation("org.bstats:bstats-base:2.2.1") // Paper
+     runtimeOnly("org.xerial:sqlite-jdbc:3.36.0.3")
+     runtimeOnly("mysql:mysql-connector-java:8.0.27")
+ 
+@@ -65,6 +66,7 @@ relocation {
+     relocate("org.bukkit.craftbukkit" to "org.bukkit.craftbukkit.v$packageVersion") {
+         exclude("org.bukkit.craftbukkit.Main*")
+     }
++    relocate("org.bstats:bstats-base" to "org.bstats") // Paper
```
Machine-Maker, line 32: we don't relocate libs anymore right?

NoahvdAa: [bStats requires you to relocate it](https://github.com/Bastian/bStats-Metrics/blob/master/base/src/main/java/org/bstats/MetricsBase.java#L211)

PR #6892, opened 2021-11-14
Remove redundant getRemoteAddress added by Paper that causes an error

PureGero:
The original code for getting a player's ip address is as follows:
```java
String s1 = "local";

if (connection.getRemoteAddress() != null) {
    s1 = connection.getRemoteAddress().toString();
}
```

However, this s1 variable became unused and the following was used instead:
```java
postChunkLoadJoin( ... connection.getRemoteAddress().toString() ... )
```

This new code does not check if the `connection.getRemoteAddress()` is null, and throws a NullPointerException when it is null.

I have fixed this by replacing the new code with the original s1 variable. However, postChunkLoadJoin is run inside a lambda function, thus an effectively final copy of s1 was needed to be created.
```java
String finalS1 = s1;
postChunkLoadJoin( ... finalS1 ... )
```

### DIFF
--- a/patches/server/0168-API-to-get-a-BlockState-without-a-snapshot.patch
+++ b/patches/server/0168-API-to-get-a-BlockState-without-a-snapshot.patch
@@ -51,40 +51,28 @@ index 1e416b23a38458f16add472cea09b0ac5ac91869..6f61fd8224fb4094f38a851300ab55f9
          return null;
      }
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
-index d8b9a8276204ad8a2f045af9fe94185dbf150bc6..3250559589eb4db7b3df828144cee81a19651842 100644
+index c0af4b05b5d77aad99487fa7f8d840e797b6554c..40fc3268e7b662cdd45b569224ccd2abfad0a595 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
 +++ b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
-@@ -317,7 +317,20 @@ public class CraftBlock implements Block {
- 
-     @Override
-     public BlockState getState() {
-+        // Paper start
-+        return this.getState(true);
-+    }
-+
-+    @Override
-+    public BlockState getState(boolean useSnapshot) {
-+        boolean prev = CraftBlockEntityState.DISABLE_SNAPSHOT;
-+        CraftBlockEntityState.DISABLE_SNAPSHOT = !useSnapshot;
-+        try {
+@@ -320,6 +320,13 @@ public class CraftBlock implements Block {
          return CraftBlockStates.getBlockState(this);
-+        } finally {
-+            CraftBlockEntityState.DISABLE_SNAPSHOT = prev;
-+        }
-+        // Paper end
      }
  
++    // Paper start
++    @Override
++    public BlockState getState(boolean useSnapshot) {
++        return CraftBlockStates.getBlockState(this, useSnapshot);
++    }
++    // Paper end
++
      @Override
+     public Biome getBiome() {
+         return this.getWorld().getBiome(this.getX(), this.getY(), this.getZ());
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/CraftBlockEntityState.java b/src/main/java/org/bukkit/craftbukkit/block/CraftBlockEntityState.java
-index 7629a51ec284cab0db7e9238027d6acfa4f3083c..27f38af1eb71dda5dd175dcabc7a467ed772757a 100644
+index 7629a51ec284cab0db7e9238027d6acfa4f3083c..c2c0b4bfab64394e55f2832e37e49bccb7e955ab 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/CraftBlockEntityState.java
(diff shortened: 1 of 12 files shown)


MiniDigger:
can't you do smth like
```
final String s1;

if (connection.getRemoteAddress() != null) {
    s1 = connection.getRemoteAddress().toString();
} else {
   s1 = "local";
}
```
instead?

patches/server/0819-Add-config-option-for-logging-player-ip-addresses.patch
```diff
@@ -93,12 +93,3 @@ index 99aa8f2ba2f10578f37de621d1a5a8e222cd70b1..eaa005c1c9b4386bcdbe1d6eb28c3eca
          }
  
          // Spigot start - spawn location event
```
Proximyst: Am I being blind or did you just entirely break another patch?

Proximyst (dismissed):

PureGero:
> can't you do smth like
> 
> ```
> final String s1;
> 
> if (connection.getRemoteAddress() != null) {
>     s1 = connection.getRemoteAddress().toString();
> } else {
>    s1 = "local";
> }
> ```
> 
> instead?

That is possible, but that'd mean modifying upstream code which isn't recommended right? Minimal modifications are preferred

patches/server/0819-Add-config-option-for-logging-player-ip-addresses.patch
PureGero: s1 uses the exact same code, that patch duplicated the same code twice

PureGero: That patch also includes:
```java
         if (connection.getRemoteAddress() != null) {
-            s1 = connection.getRemoteAddress().toString();
+            s1 = com.destroystokyo.paper.PaperConfig.logPlayerIpAddresses ? connection.getRemoteAddress().toString() : "<ip address withheld>"; // Paper
         }
```

Proximyst:
> Aren't minimal
> 
> > can't you do smth like
> > ```
> > final String s1;
> > 
> > if (connection.getRemoteAddress() != null) {
> >     s1 = connection.getRemoteAddress().toString();
> > } else {
> >    s1 = "local";
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead?
> 
> That is possible, but that'd mean modifying upstream code which isn't recommended right? Minimal modifications are preferred

It’d place a diff on change on it, which is also a good idea in places such as this where the variable name is not descriptive and probably just generated. I’d say change upstream :)

PureGero:
> > > can't you do smth like
> > > ```
> > > final String s1;
> > > 
> > > if (connection.getRemoteAddress() != null) {
> > >     s1 = connection.getRemoteAddress().toString();
> > > } else {
> > >    s1 = "local";
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > instead?
> > 
> > 
> > That is possible, but that'd mean modifying upstream code which isn't recommended right? Minimal modifications are preferred
> 
> It’d place a diff on change on it, which is also a good idea in places such as this where the variable name is not descriptive and probably just generated. I’d say change upstream :)

I've changed the upstream, should I also rename s1 to something more descriptive? That'll involve modifying a patch before it and a line of code that hasn't had a diff on it yet if I rename all references in the file.

Proximyst:
> > > > can't you do smth like
> > > > ```
> > > > final String s1;
> > > > 
> > > > if (connection.getRemoteAddress() != null) {
> > > >     s1 = connection.getRemoteAddress().toString();
> > > > } else {
> > > >    s1 = "local";
> > > > }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > instead?
> > > 
> > > 
> > > That is possible, but that'd mean modifying upstream code which isn't recommended right? Minimal modifications are preferred
> > 
> > 
> > It’d place a diff on change on it, which is also a good idea in places such as this where the variable name is not descriptive and probably just generated. I’d say change upstream :)
> 
> I've changed the upstream, should I also rename s1 to something more descriptive? That'll involve modifying a patch before it and a line of code that hasn't had a diff on it yet if I rename all references in the file.

No, changing the name isn’t necessary. :)

patches/server/0416-Load-Chunks-for-Login-Asynchronously.patch
```diff
@@ -136,7 +136,24 @@ index 9609579e6d61c3975740d07a394680aeef952516..6b23b9389ff92ae8016d4adb289ac2a0
...
  
          if (nbttagcompound != null) {
-@@ -260,6 +267,52 @@ public abstract class PlayerList {
+@@ -213,11 +220,15 @@ public abstract class PlayerList {
+         if (nbttagcompound == null) player.fudgeSpawnLocation(worldserver1); // Paper - only move to spawn on first login, otherwise, stay where you are....
+ 
+         player.setLevel(worldserver1);
+-        String s1 = "local";
++        // Paper start - make s1 effectively final
++        String s1;
```
Proximyst: Explicitly make it final.

Proximyst (changes requested):

patches/server/0416-Load-Chunks-for-Login-Asynchronously.patch
PureGero: Done, thank you

Machine-Maker (approved):
Needs a rebase, but changes look good

PR #3818, opened 2020-07-06
Fix anvil not working when level req is 0

JanTuck:
When an anvil has a level req 0 of 0 on any operation you won't be able to take out the result slot, this will fix this issue while also updating the cursor since that seemed to not be updated without specifically updating it.

Code used to test this fix is:

```java
    @EventHandler
    public void anvilEvent(PrepareAnvilEvent event){
        AnvilInventory itemStacks = event.getInventory();

        if (itemStacks.getFirstItem() == null || itemStacks.getFirstItem().getType() != Material.DIRT)
            return;
        if (itemStacks.getSecondItem() == null || itemStacks.getSecondItem().getType() != Material.DIRT)
            return;

        event.setResult(new ItemStack(Material.DIAMOND));
    }
```

PR #6895, opened 2021-11-15
Fix AnvilPrepareEvent not working with zero xp

JRoy:
Ports #3818 to 1.17

On the 1.17 client, the hack to update the player's cursor is no longer needed.

Confirmed to work using same test code on OP.


Machine-Maker (approved):
I do wonder if this might break existing vanilla behavior but I can't think of a situation where it might.

kennytv (approved):

PR #6897, opened 2021-11-15
Add PlayerItemFrameChangeEvent

### LINKED ISSUE #6783: PlayerItemFrameInsertItemEvent and PlayerItemFrameRemoveItemEvent

I'm trying to modify the item being placed inside an item frame. When using the PlayerInteractEntityEvent, I can check the used item, but that corresponds to the whole stack the player is holding, not just the single item which ends up being placed inside the frame, so modifying it changes the remaining items the player is holding.

A relatively simple event, containing the item frame, the player who inserts the item, and the modifiable itemstack of the item which ends up being placed inside the frame. An event for removing an item from an item frame isn't needed in my use case, but would probably be nice to have for the sake of completeness.

My solution was to schedule a task a few ticks delayed after the interact event, checking and modifying the itemstack, which in my opinion isn't very efficient for such a simple problem.

### Other

_No response_

SamB440:
This adds a PlayerItemFrameChangeEvent for checking the adding, removing, and rotating items in item frames. Can be cancelled and also allows the modification of the item that will be placed. Closes #6783 

Tested with the following code, and with a redstone comparator connected to make sure redstone works as it should:
```java
@EventHandler
public void onChange(PlayerItemFrameChangeEvent event) {
	this.getLogger().info(String.format("Action was: %s", event.getAction()));
	if (event.getAction() == PlayerItemFrameChangeEvent.ItemFrameChangeAction.ROTATE) {
		event.setItemStack(new ItemStack(Material.DIAMOND));
		event.setCancelled(true); // Tested with both cancelled and not cancelled
		return;
	}

	final ItemStack itemStack = event.getItemStack();
	itemStack.setType(Material.DIRT);
}
```

Sorry for all the commits, kept finding tiny things to fix...

### DIFF
--- a/patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
+++ b/patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
@@ -6,12 +6,13 @@ Subject: [PATCH] Add PlayerItemFrameChangeEvent
 
 diff --git a/src/main/java/io/papermc/paper/event/player/PlayerItemFrameChangeEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerItemFrameChangeEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..425cab00fe162588706c91385a5aa3d2de8f47f5
+index 0000000000000000000000000000000000000000..1fd8511d82b222a4b3545203b035a9b658196536
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/player/PlayerItemFrameChangeEvent.java
-@@ -0,0 +1,93 @@
+@@ -0,0 +1,96 @@
 +package io.papermc.paper.event.player;
 +
++import org.bukkit.Material;
 +import org.bukkit.entity.ItemFrame;
 +import org.bukkit.entity.Player;
 +import org.bukkit.event.Cancellable;
@@ -19,6 +20,7 @@ index 0000000000000000000000000000000000000000..425cab00fe162588706c91385a5aa3d2
 +import org.bukkit.event.player.PlayerEvent;
 +import org.bukkit.inventory.ItemStack;
 +import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
 +
 +/**
 + * Called when an {@link ItemFrame} is having an item rotated, added, or removed from it.
@@ -50,7 +52,7 @@ index 0000000000000000000000000000000000000000..425cab00fe162588706c91385a5aa3d2
 +
 +    /**
 +     * Gets the {@link ItemStack} involved in this event.
-+     * This is the item being added or removed from the {@link ItemFrame}.
++     * This is the item being added, rotated, or removed from the {@link ItemFrame}.
 +     * @return the {@link ItemStack} being added or removed
 +     */
 +    @NotNull
@@ -60,10 +62,11 @@ index 0000000000000000000000000000000000000000..425cab00fe162588706c91385a5aa3d2
 +
 +    /**
 +     * Sets the {@link ItemStack} that this {@link ItemFrame} holds.
++     * If null is provided, the resulting ItemStack will be empty.
 +     * @param itemStack {@link ItemFrame} item
 +     */
-+    public void setItemStack(@NotNull ItemStack itemStack) {
-+        this.itemStack = itemStack;
++    public void setItemStack(@Nullable ItemStack itemStack) {
++        this.itemStack = itemStack == null ? new ItemStack(Material.AIR) : itemStack;
 +    }
 +
 +    /**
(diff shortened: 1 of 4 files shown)


lynxplay:
As a heads up on this PR, upstream is also proposing an addition that would solve the issue in https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/pull-requests/672/overview (if you don't have access to the PRs, the gist of it is the addition of `ItemFrameChangeItemEvent` as well as `ItemFrameChangeRotationEvent` which both extend `ItemFrameChangeEvent`, tho just because upstream has an open PR for this, this PR is certainly not invalid :+1:

patches/server/0844-Add-PlayerItemFrameChangeEvent.patch
```diff
@@ -0,0 +1,60 @@
...
+ import net.minecraft.core.BlockPos;
+ import net.minecraft.core.Direction;
+ import net.minecraft.nbt.CompoundTag;
+@@ -185,6 +186,14 @@ public class ItemFrame extends HangingEntity {
+                     return true;
+                 }
+                 // CraftBukkit end
++                // Paper start - call PlayerItemFrameChangeEvent
++                if (source.getEntity() instanceof Player player) {
++                    var event = new PlayerItemFrameChangeEvent((org.bukkit.entity.Player) player.getBukkitEntity(), (org.bukkit.entity.ItemFrame) this.getBukkitEntity(), this.getItem().asBukkitCopy(), PlayerItemFrameChangeEvent.ItemFrameChangeAction.REMOVE);
```
lynxplay: Any particular reason this isn't a fully qualified import similar to the `org.bukkit.entity.Player` in the same line ?

patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
```diff
@@ -0,0 +1,105 @@
...
++    public ItemStack getItemStack() {
++        return itemStack;
++    }
++
++    /**
++     * Sets the {@link ItemStack} that this {@link ItemFrame} holds.
++     * @param itemStack {@link ItemFrame} item
++     */
++    public void setItemStack(@NotNull ItemStack itemStack) {
++        this.itemStack = itemStack;
```
kezz: Should probably have a null check here

Doc94: Not sure but if null is passed (ignoring the NotNull) the item resulting in NMS when convert is EMPTY.

kezz: Then that should be documented and the nullability annotation changed

patches/server/0844-Add-PlayerItemFrameChangeEvent.patch
SamB440: The `org.bukkit.entity.Player` is mainly there because the nms player was already imported (which I use in the instanceof), so it needs that. 

As for the PlayerItemFrameChangeEvent and var, that was mostly to make it prettier and reduce the length of the line - I could change it but it looks really messy to me otherwise.

patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
SamB440: For the sake of usability, I've made it nullable and then if its null just set the itemstack to air - this mimics the default behaviour if it were to be passed as null whilst making sure the `getItemStack` method never returns null.

kennytv (changes requested):
Cancelling the event should entail no item being set

SamB440:
Okay, should be fixed now! Wasn't sure originally what to do when it was cancelled because I thought someone might have wanted to set it and cancel it, but I guess they can get the ItemFrame instead and set it there.

kennytv (approved):

patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
```diff
@@ -0,0 +1,108 @@
...
++     * @return the {@link ItemStack} being added or removed
++     */
++    @NotNull
++    public ItemStack getItemStack() {
++        return itemStack;
++    }
++
++    /**
++     * Sets the {@link ItemStack} that this {@link ItemFrame} holds.
++     * If null is provided, the resulting ItemStack will be empty.
```
kezz: This is super confusing - you should explicitly state that it is set to AIR both here and in the javadoc for the getter, noting that if it returns air then the item stack will be empty?

patches/server/0844-Add-PlayerItemFrameChangeEvent.patch
Machine-Maker: yeah, I think its fine. looks like you are using it 6 times in this class o

patches/api/0347-Add-PlayerItemFrameChangeEvent.patch
SamB440: I'm not sure what you mean by "set to both"?
Edit: oops didn't see edit

kashike (approved):

Issue #4793, opened 2020-11-21
Visual glitch with custom hearts container


Doc94:
### What behaviour is expected:
If you have custom heart containers (example 40 not the default 20) when you join you can see all containers filled.

### What behaviour is observed:
When join (you have full life) the client only show 20 hearts filled and the another 20 not filled, if you take damaage then the client reload the rest of containers full.

### Steps/models to reproduce:
1. Join to the server
2. Run this command `/attribute @s minecraft:generic.max_health base set 40`
3. Get all the containers filled
4. Disconnect from server
5. Reconnect to Server
6. Take damage

### Plugin list:
AsyncWorldEdit, ConterCraft, CoreProtect, LibsDisguises, LuckPerms, PlugMan, ProtocolLib, WorldEdit

### Paper version:
git-Paper-288 (MC: 1.16.4) (Implementing API version 1.16.4-R0.1-SNAPSHOT)

### Anything else:
Gif of issue: https://imgur.com/rp8sWqA

Doc94:
Its a Spigot bug but they cant find the cause of issue: https://hub.spigotmc.org/jira/browse/SPIGOT-6250

HexedHero:
Thought I'd mention that I've been using a workaround fix in my plugin for a while now, simply just doing
`Common.sendPacketToPlayer(event.getPlayer(), new PacketPlayOutUpdateAttributes(((CraftPlayer) event.getPlayer()).getHandle().getId(), ((CraftPlayer) event.getPlayer()).getHandle().getAttributeMap().b()));`
on PlayerJoinEvent

Owen1212055:
Can reproduce as of 1.19.1, looking into this myself.

Owen1212055:
Is this able to be reproduced as of 1.19.2 using the chunk system recode? 

Some testing was done and it seems that it resolves this issue, but my PR might serve a purpose as there is no reason to be resending the player tracker data like that anyways.

Doc94:
Finally this issue looks fixed then i prefer close because the main issue is handled 🍡

PR #6904, opened 2021-11-17
Add Player#sendHealthUpdate

SamB440:
This simply allows the CraftPlayer sendHealthUpdate method to be accessible via API in the Player interface. The method is useful if you have used setHealth and want to show the heart flash animation to a player to make it appear more natural. (if you don't know what I mean check out https://www.spigotmc.org/threads/make-players-hearts-flash-like-when-they-naturally-heal.484328/)

I already use the CraftPlayer method in some plugins but it would be nice (and better) for it to be accessible through API. Example:
```java
player.setHealth(player.getHealth() + addition);
player.sendHealthUpdate();
```

### DIFF
--- a/patches/api/0348-Add-player-health-update-API.patch
+++ b/patches/api/0348-Add-player-health-update-API.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SamB440 <sam@islandearth.net>
+Date: Wed, 17 Nov 2021 12:30:36 +0000
+Subject: [PATCH] Add player health update API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index 9a6e410206852029f1fea0c4409352d5743dcf64..566b47da282ed0c8bc87b0f1757019f5c7e2acfb 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -1558,6 +1558,15 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      */
+     public double getHealthScale();
+ 
++    // Paper start
++    /**
++     * Forcefully sends a health update to the player.
++     * This includes the current health, saturation, and food level.
++     * This may show the heart animation of gaining or losing health.
++     */
++    public void sendHealthUpdate();
++    // Paper end
++    
+     /**
+      * Gets the entity which is followed by the camera when in
+      * {@link GameMode#SPECTATOR}.
--- a/patches/api/0348-Add-player-health-update-API.patch
+++ b/patches/api/0348-Add-player-health-update-API.patch
@@ -5,18 +5,36 @@ Subject: [PATCH] Add player health update API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 9a6e410206852029f1fea0c4409352d5743dcf64..566b47da282ed0c8bc87b0f1757019f5c7e2acfb 100644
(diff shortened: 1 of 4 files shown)


kennytv (approved):

patches/server/0845-Add-player-health-update-API.patch
```diff
@@ -0,0 +1,18 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+index 4ff4143f3a7cd89ef92f4b8882fa3e5addfe0f06..01f18c29a4107a11b3e6d607658868f302f557d7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -2009,6 +2009,7 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+         this.getHandle().maxHealthCache = getMaxHealth();
+     }
+ 
++    @Override // Paper
```
kennytv: Comment for clarity's sake; the server patch is not needed, but it's fine to stay as a forced diff on method move/change

patches/api/0348-Add-player-health-update-API.patch
```diff
@@ -0,0 +1,26 @@
...
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -1558,6 +1558,15 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      */
+     public double getHealthScale();
+ 
++    // Paper start
++    /**
++     * Forcefully sends a health update to the player.
++     * This includes the current health, saturation, and food level.
++     * This may show the heart animation of gaining or losing health.
```
kezz: In what situations will it/won't it show the animation? Might be worth documenting?

SamB440: I believe it will only show if there is a heart change, so afaik it will only work after using setHealth

SamB440:
I wonder if it would be an idea to expand on this and add parameters to the method, allowing control of the health food and saturation level sent to the client? This would be much more useful imo and make the server patch not useless...

I've already just tested such a method and it works, just wondering if this would be a bad idea and/or anyone would be against this. Example being:
```java
    @EventHandler
    public void onJoin(PlayerJoinEvent event) {
            event.getPlayer().sendHealthUpdate(15, 15, 0); // Health, food, saturation
        }, 20L);
    }
```

Doc94:
This can help like a.. alternative for [#4793](https://github.com/PaperMC/Paper/issues/4793) (?

SamB440:
> 
> 
> This can help like a.. alternative for [#4793](https://github.com/PaperMC/Paper/issues/4793) (?

That issue is indeed fixed if you call sendHealthUpdate on join

SamB440:
I think this is ready again - not sure about the paper comments in the server patch, though.
Edit: no idea why gh did that review thing

patches/api/0348-Add-player-health-update-API.patch
```diff
@@ -0,0 +1,44 @@
...
+index 9a6e410206852029f1fea0c4409352d5743dcf64..c5aaebeb2a1ea61b526c3763d6ad687d2746c35f 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -1558,6 +1558,33 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      */
+     public double getHealthScale();
+ 
++    // Paper start
++    /**
++     * Forcefully sends a health update to the player's client.
```
kennytv: "to the player.", and remove the empty line below this. same for the other method

kennytv (approved):

PR #6922, opened 2021-11-21
Don't get block light level if the sky level is 15

NoahvdAa:
A *very* small light engine optimization. If the sky light level is 15, the block light level will never be higher,
since Minecraft doesn't support light levels above 15. This will prevent an extra call to `getLightValue` for
areas with sky access during the day. After some quick profiling on a flat world, this seems to reduce the time spent
on the `getBrightness` method by ~23%. The method is called every time a grass block randomly ticks
(along with crop growth), so this will cause a very slight reduction in the tick duration for most servers.


Proximyst (approved):

NoahvdAa:
Rebased for 1.18:tm:

Machine-Maker (approved):

MiniDigger (approved):

jpenilla:
This PR needs to target Starlight, as it is this will get lost/reverted during the 1.19 update.

PR #6924, opened 2021-11-22
Create check to discard ThrownEnderpearls upon owner death even when loaded lazily/in another chunk

### LINKED ISSUE #6920: Ender pearls not despawning upon owner death if lazy loaded
### Expected behavior

If an ender pearl is in a non entity processing chunk, it should either forget the owner as if it's unloaded, or despawn upon the owner dying as if it's loaded.

### Observed/Actual behavior

When an ender pearl is in a non entity processing chunk and the owner dies, it doesn't despawn. It can still be used to teleport a player after the owner has died.

### Steps/models to reproduce

Build portal loader, place stasis chamber in the lazy loaded border around it, go far away to unload the chunks and die, go back and load the chunks, stasis chamber still works.

### Plugin and Datapack List

FastAsyncWorldEdit (WorldEdit)

### Paper version

This server is running Paper version git-Paper-388 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 22aaf91)
Previous version: git-Paper-386 (MC: 1.17.1)

### Other

_No response_

reflqct:
Fixes #6920
Adds ArrayList to remember all thrown enderpearls (every time a pearl is thrown it is added to the list), every tick check whether to discard.


vytskalt:
Maybe use ObjectArrayList from fastutil instead of ArrayList?

electronicboy:
thanks for the PR however this was not the ideal way to do this, and a better mechanism taking advantage of changes to how the game works has been created

PR #6933, opened 2021-11-24
Patch/optimize collision to not load chunks

Machine-Maker:
the CollisionSpliterator was renamed (and reworked a little) to BlockCollisions which was the big thing preventing this patch from being merged.


Machine-Maker:
Done manually in 62c41173db773997be04396c3d98503ad95114d4

PR #6944, opened 2021-11-28
Port flat bedrock generator settings to 1.18

NoahvdAa:
1.18 changed bedrock generation around a bunch, so creating our own ConditionSource seemed like the cleanest solution.


Machine-Maker (commented):
Are you making any changes to anything in the SurfaceRules class besides the dev fixes? I don't think you have to do that at all then, just let it not be imported, and you dont have to fix anything.

NoahvdAa:
> Are you making any changes to anything in the SurfaceRules class besides the dev fixes? I don't think you have to do that at all then, just let it not be imported, and you dont have to fix anything.

I could've sworn the access transformers didn't work without the decompile fix, looks like I was wrong.

EDIT: Looks like the build passes, must've confused it with something else.

patches/server/0344-Generator-Settings.patch
```diff
@@ -0,0 +1,176 @@
...
+index 514493f20bcd0a697d6787e11ec7042e601e1de2..5e4da6882b9ba35775c110822ba3a8dd1a6fb5ba 100644
+--- a/src/main/java/net/minecraft/data/worldgen/SurfaceRuleData.java
++++ b/src/main/java/net/minecraft/data/worldgen/SurfaceRuleData.java
+@@ -54,6 +54,50 @@ public class SurfaceRuleData {
+         return overworldLike(true, false, true);
+     }
+ 
++    // Paper start
++    // Taken from SurfaceRules$VerticalGradientRuleSource
++    private final record PaperBedrockGradientRule(net.minecraft.resources.ResourceLocation randomName, VerticalAnchor trueAtAndBelow, VerticalAnchor falseAtAndAbove, boolean invert) implements SurfaceRules.ConditionSource {
```
Machine-Maker: 1. This isn't a rule, its a ConditionSource
2. I think you mean `Taken from SurfaceRules$VerticalGradientConditionSource`
3. (optional) make randomName a String just to remove the need for the fully qualified import in the record ctor and in usages of this condition source.

patches/server/0344-Generator-Settings.patch
```diff
@@ -0,0 +1,176 @@
...
++        this.level = level; // Paper
+     }
+ 
+     public int getMinGenY() {
+@@ -19,4 +22,6 @@ public class WorldGenerationContext {
+     public int getGenDepth() {
+         return this.height;
+     }
++
++    public net.minecraft.world.level.Level getWorld() { return this.level; } // Paper
```
Machine-Maker: Throw an exception that's not an NPE if this is called when level is null I think. I suppose its possible if some plugin creates a WorldGenerationContext

Machine-Maker (changes requested):

kennytv:
Reopen to master plz

PR #6945, opened 2021-11-28
Implement VarInt optimizations from Velocity

Gerrygames:
I did some crude testing and this did result in faster initial chunk sending. The creation of ClientboundLevelChunkWithLightPacket uses getVarIntSize quite a lot.


patches/server/0811-Optimize-VarInts.patch
```diff
@@ -0,0 +1,90 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gero <gecam59@gmail.com>
```
kennytv: Tux as an author

kennytv:
Reopen to master plz

PR #6246, opened 2021-07-22
[ci skip] adjust upstreamCommit script to not mention paper issues

Machine-Maker:
Prevents those nasty mentions of old issues from commit msgs from upstream repos

PR #6956, opened 2021-11-30
[ci skip] adjust upstreamCommit script to not mention paper issues

Machine-Maker:
was auto closed because dev/1.18 branch was deleted
See #6246


Machine-Maker:
Updated to include longer commit refs, I didn't realize they could be over 8 chars in length sometimes.

electronicboy (approved):

PR #6957, opened 2021-11-30
Implement Velocity VarInt optimizations

Gerrygames:


Euphillya:
Hello,

I don't understand your optimization, I did a little benchmark and the code is 6 times slower than the base one if the value is high.

SirYwell:
> I did a little benchmark and the code is 6 times slower than the base one if the value is high.

It would make sense to share that benchmark then so they can be compared to existing benchmarks.

patches/server/0815-Optimize-VarInts.patch
```diff
@@ -0,0 +1,90 @@
...
++            writeShort(w);
++        } else {
++            writeVartFull(value);
+         }
+-
+-        this.writeByte(value);
+         return this;
+     }
+ 
++    private void writeVartFull(int value) {
```
astei: Nit: `writeVarIntFull`. Or just peel the common one and two byte cases (`kotlinx.serialization` does this for example) and fall back to the original Mojang code, which works just as well. This function mainly exists to show that every case can be unwrapped for maximum performance for more obscure cases.

astei (changes requested):

astei:
> Hello,
> 
> I don't understand your optimization, I did a little benchmark and the code is 6 times slower than the base one if the value is high.

Alright, I'd really like to see your benchmark, as one of the people who helped develop this technique.

My benchmark is here: https://github.com/astei/varint-writing-showdown

It generates a fixed sequence of 2,048 numbers, all of varying bit lengths. Note that the sequence of numbers used is very unrealistic since VarInt size distributions in Minecraft are strongly skewed towards the numbers that require the fewest bytes to be represented, so the test itself is not particularly realistic. I fixed one bug (in favor of slowing down the entire test, it was representing only up to 30-bit numbers, this was fixed and made up to 31-bit numbers).

The result?
```
VarIntWriterBenchmark.blendedVarintWrite      thrpt    5  218785.175 ±  4147.073  ops/s
VarIntWriterBenchmark.oldVelocityVarintWrite  thrpt    5  109601.596 ±  7728.028  ops/s
```
It's about twice as fast, and this is with a ludicrously random sequence of numbers.

How about reducing the size of the numbers to up to just 14-bit numbers? That seems to me a more realistic expectation, although it is still strongly off-skew since 1 byte VarInts (up to 7 bits) are by far the most commonly appearing in the protocol:
```
VarIntWriterBenchmark.blendedVarintWrite      thrpt    5  344570.042 ± 22517.372  ops/s
VarIntWriterBenchmark.oldVelocityVarintWrite  thrpt    5  150831.645 ±  1901.416  ops/s
```
More than twice as fast.

electronicboy (approved):

PR #6961, opened 2021-11-30
Use LOSING_JOB change reason for villager career change event

roro1506HD:
This PR fixes an issue with VillagerCareerChangeEvent. This event has two possible ChangeReason (EMPLOYED, LOSING_JOB) but the second one is never fired.

This PR also have an open [spigot JIRA ticket](https://hub.spigotmc.org/jira/browse/SPIGOT-6820) so it is very likely to be deleted soon.


andrew121410:
Fixed in https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/a46e7f18e838abf2bb7dc7f1d5f4b98036146191

Chew:
Yep, Spigot beat us. Fixed in https://github.com/PaperMC/Paper/pull/6966. Thank you anyway!

PR #6968, opened 2021-12-01
WIP brain API

Owen1212055:
This API is supposed to completely expose the "Brain" API that are used by certain entities such as goats, axolotls and villagers. However this api ALSO allows **manually ticking brains of all entities**. 

This API is HEAVILY in WIP. I have created this pull request in order to collect as much feedback as I can here, so there is a lot of stuff like imports, access transformers, and paper comments I have to still do.

Some things I need help with:
- Registering Custom Types
   - As of 1.18.2 you can no longer register stuff like this during runtime, how should I handle the registry freezing for registering custom memories/sensors/etc? 
   - Memories
       - Memories that are serialized don't get loaded correctly, basically see the big comment in the Brain diff.
- Is this level of exposure necessary? 
    - Although there is a lot you have access to here, I worry that this may become quite a huge burden due to the amount of hacks that this requires.

I pushed a rough example in the test plugin, the pickled brain! See https://canary.discord.com/channels/289587909051416579/925530366192779286/989010753371652106 for the behavior.
ALL FEEDBACK WELCOME!

image
If you would like to use the brain debugger, use this fabric mod. This will allow the debugger to render.
[enable-brain-debug-1.0-SNAPSHOT.zip](https://github.com/PaperMC/Paper/files/8982474/enable-brain-debug-1.0-SNAPSHOT.zip)


patches/api/0343-WIP-Brain-API.patch
```diff
@@ -0,0 +1,523 @@
...
++All entities have a brain, but not all entities actually tick their brain.
++
++
++Behaviors are just goals combined with eachother, and are defined by "Activities".
++
++Sensors populate memories, and memories are used to trigger/read by activities.
++
++
++ */
++public interface BrainHolder<T extends LivingEntity> {
```
kashike: I think it would be better to expose a `Brain` object instead of an interface like this that entities extend.

kashike (changes requested):

patches/api/0343-WIP-Brain-API.patch
Owen1212055: How would I handle the generic type then? Additionally, not all entities use their brain so it may be confusing to some people.

Should I make it so that this interface has a method to RETURN a brain instance? I'd love a pointer here.

kashike: All `LivingEntity` have a `Brain` in vanilla, for what it is worth.

Machine-Maker: yeah, but not all, in fact most make no use of it iirc. So I do kinda think its best to only add it to mobs that make use of it.

Owen1212055: That is true, but only some of them actually tick their brain. So if we were to expose this api to all entities, it may be optimal to force tick the brain for these types of entities? I am not sure, additionally, how would I handle the generics here?

(edit: whoops MM was faster!)

Owen1212055:
```java
private final MemoryKey<Boolean> IS_SCARED = Bukkit.createMemoryKey(NamespacedKey.fromString("scared", this), Boolean.class);
    private final SensorKey SCARIES_SENSOR =  Bukkit.createSensorKey(NamespacedKey.fromString("scary_mobs_finder", this));
    private final ActivityKey SCARED_ACTIVITY = Bukkit.createActivityKey(NamespacedKey.fromString("scared", this));
```

I updated how custom keys are created. This allowed me to get rid of the maps, and instead just do some instance checking in the utility classes (which have now been merged as well).

Owen1212055:
I have included a new example that further shows what this system can use and have documented most of the api. Opinions appreciated. :)

maestro-denery:
This will be a really nice API, I have a team which makes a "Vanilla with additions" server, so we develop our own APIs and libraries for "mod-like" development with plugins, we do many NMS hacks to reach "modded" experience for players, obviously making hacks is bad, I tried to make something like "Brain API" via VarHandles and reflection, but it looked **very** dirty, and probably the performance in this case would be very poor. so as a plugin developer I need fast and minimally hacky API for interacting with entity AI. I plan contribute in this WIP API, because I think that it will be useful for other people like me.

maestro-denery:
Hello everyone, I did some [refactor](https://github.com/ExtraExtremeERA/Paper/commit/5b9cd80a75f9c3f1269830eda86a605f2efd98bc) of this draft API. (Owen knows about it) This refactor is about making better backward compatibility with Bukkit MemoryKey API, cleaner implementation and better maintainability of Brain API. I tested it, but anyway it can contain some bugs. Feel free to give a feedback about this refactor!
### API changes:
1. Created `MemoryModuleType<U>` API interface which has `NamespacedKey` getter and `Optional<MemoryKey<U>>` getter, this getter returns `Optional` having some value **only if** this `MemoryModuleType` is vanilla and represented in MemoryKey API.
2. Created `MemoryManager` interface (singletone like `BrainManager`) where I transferred methods manipulating with memory (sensor, activity, behaviour methods aren't transferred) from `BrainManager`, edited them with `MemoryModuleType` and added additional `setMemory()` and `getMemory()` methods for manipulating with `MemoryModuleType` instead of `MemoryKey`.
3. Changed usage of `MemoryKey` in `BrainManager`, `Sensor` and etc. to `MemoryModuleType`.
4. Deleted methods for creation of custom MemoryKeys in Bukkit and Server and replaced them with creation of API `MemoryModuleType` both from Bukkit `MemoryKey` and `NamespacedKey`.
### Impl details:
1. `MemoryManager` has `PaperMemoryManager` implementation which has memory methods implemented and also contains two `HashBiMap`s, first `BiMap` contains registered custom types of API `MemoryModuleType`s and second one is a cache for `MemoryModuleType`s representing vanilla memories.
2. API interface `MemoryModuleType` has `PaperMemoryModuleType` implementation containing NMS `MemoryModuleType`, its own unique namespaced key and optionally `MemoryKey`. This impl has two ways of creation, first one creates from `NamespacedKey` and second one creates from `MemoryKey`, first makes new NMS `MemoryModuleType` puts `NamespacedKey` and itself in registry defined in `PaperMemoryManager` and also puts new NMS `MemoryModuleType` into NMS `Registry`. I'll show later why NMS `Registry` is needed. Second gets value from the cache defined in `PaperMemoryManager` if it exists, if not, it gets NMS `MemoryModuleType` through `CraftMemoryKey.fromMemoryKey(key)` and inserts its `NamespacedKey` into API `MemoryModuleType`'s one. And puts this instance to the cache and returns.
3. I added in `PaperBrainUtil` three methods, first is `getHandle()` for getting NMS `MemoryModuleType` for API one, and `toPaper()` which gets API `MemoryModuleTypes` from NMS one. How? It is the place where NMS `Registry` helps. This method looks like:
```java
public static PaperMemoryModuleType<?> toPaper(net.minecraft.world.entity.ai.memory.MemoryModuleType<?> nms) {
        // custom types actually register in NMS Registry, that's why we are getting them from it.
        final PaperMemoryModuleType<?> customMemoryType = (PaperMemoryModuleType<?>) PaperMemoryManager.INSTANCE.getTypeByKey(CraftNamespacedKey.fromMinecraft(Registry.MEMORY_MODULE_TYPE.getKey(nms)));
        // Here we are getting Vanilla type (from cache) if custom type doesn't exist.
        return customMemoryType == null ? PaperMemoryModuleType.of(CraftMemoryKey.toMemoryKey(nms)) : customMemoryType;
}
``` 
It is the main method which makes possible refactoring of everything else. And the third is `handleIfVanilla()`, this method operates with memory value and maps this value to NMS if `Optional<MemoryKey<U>>` presents in API `MemoryModuleType`, it is needed for proper manipulation with memory values of Vanilla `MemoryModuleType`s in `MemoryManager`.
### Code examples.
This is how this API `MemoryModuleType` creates:
```java
public static final MemoryModuleType<List<Squid>> SQUID_CANDIDATES = Bukkit.createMemoryModuleType(NamespacedKey.fromString("squid_candidates"));
public static final MemoryModuleType<List<Parrot>> NEARBY_PARROTS = Bukkit.createMemoryModuleType(NamespacedKey.fromString("nearby_parrots"));
public static final MemoryModuleType<Location> MEETING_POINT = Bukkit.createMemoryModuleType(MemoryKey.MEETING_POINT);
```
This is how to manipulate with memory in this refactor:
```java
    // Obtaining MemoryManager from BrainManager
    MemoryManager memoryManager = Bukkit.getBrainManager().getMemoryManager();
    // Getting memory value of given entity.
    List<Squid> squids = memoryManager.getMemory(entity, TestPlugin.SQUID_CANDIDATES).orElse(new ArrayList<>());
        if (target == null || target.isDead()) {
            if (squids.isEmpty()) {
                // Erasing it.
                memoryManager.eraseMemory(entity, TestPlugin.SQUID_CANDIDATES);
                memoryManager.eraseMemory(entity, TestPlugin.SQUID_RAGE);
                return;
            }

            target = squids.remove(0);
        } else {
            entity.getPathfinder().moveTo(target);
        }
```
### Afterword.
As Owen said: "ALL FEEDBACK WELCOME!", I wanna see Opinions about this refactor, especially about its maintainability, and impl.

Machine-Maker:
Responding to @ExtraExtremeERA 
Couple things here...
1. That new interface `MemoryModuleType<U>`, I don't think it fits well unless you are just replacing all of MemoryKey with it, cause now there are two MemoryKey APIs essentially instead of just one. I prefer Owen's PaperMemoryKey (but that has its own problems too which ill comment on later)

2. I like having concept of separating the Memory stuff into a separate interface, but I haven't really been a fan of the whole Manager interface. Why not just make that interface one that Entities extend directly and then we implement it in some `PaperMemoryHolder` interface that we make CraftLivingEntity extend?

patches/api/0353-WIP-Brain-API.patch
```diff
@@ -0,0 +1,802 @@
...
++++ b/src/main/java/io/papermc/paper/entity/brain/memory/MemoryPair.java
+@@ -0,0 +1,9 @@
++package io.papermc.paper.entity.brain.memory;
++
++import org.bukkit.entity.memory.MemoryKey;
++
++/**
++ * A memory pair represents a memory and what key status it is required to have.
++ */
++public record MemoryPair(MemoryKeyStatus requiredStatus, MemoryKey<?> key) {
```
Machine-Maker: any benefit to this having a generic param?

patches/api/0353-WIP-Brain-API.patch
```diff
@@ -0,0 +1,802 @@
...
+-public final class MemoryKey<T> implements Keyed {
++public class MemoryKey<T> implements Keyed {
+ 
+     private final NamespacedKey namespacedKey;
+     private final Class<T> tClass;
+ 
++    protected MemoryKey(NamespacedKey namespacedKey, Class<T> tClass, boolean dummy) {
++        this.namespacedKey = namespacedKey;
++        this.tClass = tClass;
++    }
```
Machine-Maker: Creating custom memory keys should be a thing, but they should both be registered with the NMS registry and the bukkit one. this ctor doesn't do that.

Owen1212055:
> 2. I like having concept of separating the Memory stuff into a separate interface, but I haven't really been a fan of the whole Manager interface. Why not just make that interface one that Entities extend directly and then we implement it in some `PaperMemoryHolder` interface that we make CraftLivingEntity extend?

Regarding this point, I would defo like your opinion on this. The issue is that the type of the entity is required in some cases, and in the future when most likely more entities are moved over to use the brain system it will cause the generic to be tossed everywhere. See: https://canary.discord.com/channels/289587909051416579/925530366192779286/925667741984243723

Ping me in contrib tho and we can discuss more, cause i'd like more feedback.

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++        brain.memories.put(PaperBrainUtil.getHandle(memoryKey), Optional.empty());
++    }
++
++    @SuppressWarnings("unchecked")
++    @Override
++    public <V> void setMemory(@NotNull LivingEntity livingEntity, @NotNull MemoryKey<V> memoryKey, @Nullable V value, long expireIn) {
++        Brain<?> brain = getBrain(livingEntity);
++
++        if (memoryKey instanceof PaperMemory<V>) {
++            brain.setMemoryWithExpiry(PaperBrainUtil.getHandle(memoryKey), value, expireIn);
```
Machine-Maker: setMemoryWithExpiry expects a non null value

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++
++import net.minecraft.world.entity.schedule.Activity;
++import org.bukkit.NamespacedKey;
++import org.jetbrains.annotations.NotNull;
++
++public class PaperActivity extends Activity implements ActivityKey {
++
++    private final NamespacedKey key;
++
++    public PaperActivity(NamespacedKey key) {
```
Machine-Maker: Custom nms types with registries should be registered with those registries

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++++ b/src/main/java/io/papermc/paper/entity/brain/memory/PaperCustomMemory.java
+@@ -0,0 +1,20 @@
++package io.papermc.paper.entity.brain.memory;
++
++import net.minecraft.world.entity.ai.memory.MemoryModuleType;
++import org.bukkit.entity.memory.MemoryKey;
++
++import java.util.Optional;
++
++public class PaperCustomMemory<T> extends MemoryModuleType<T> {
```
Machine-Maker: will throw an NPE on toString

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++        this.memoryModuleType = new PaperCustomMemory<>(this); // No serializer
++    }
++
++    public MemoryModuleType<T> getMemoryModuleType() {
++        return memoryModuleType;
++    }
++
++    @Override
++    public @NotNull Class<T> getMemoryClass() {
++        throw new UnsupportedOperationException("Custom paper memories don't have their type stored.");
```
Machine-Maker: Why not? just make the PaperMemory ctor take a type token

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++
++import net.minecraft.world.entity.ai.sensing.SensorType;
++import org.bukkit.NamespacedKey;
++import org.jetbrains.annotations.NotNull;
++
++public class PaperSensor extends SensorType implements SensorKey {
++
++    private final NamespacedKey key;
++
++    public PaperSensor(NamespacedKey key) {
```
Machine-Maker: same think about registering this stuff in the nms registries

patches/server/0842-WIP-Brain-API.patch
```diff
@@ -0,0 +1,1020 @@
...
++        return new io.papermc.paper.entity.brain.activity.PaperActivity(key);
++    }
++
++    @Override
++    public io.papermc.paper.entity.brain.sensor.SensorKey createSensorKey(NamespacedKey key) {
++        return new io.papermc.paper.entity.brain.sensor.PaperSensor(key);
++    }
++
++    @Override
++    public org.bukkit.entity.memory.MemoryKey<?> createMemoryKey(NamespacedKey key) {
```
Machine-Maker: missing type param, and also, this should take a type token to store to make it as much like the vanilla representation of the MemoryKey

Machine-Maker (changes requested):
also a bunch of unneeded imports, line changes, stuff like that but those don't really matter right now.

patches/api/0353-WIP-Brain-API.patch
Owen1212055: I didn't think it was needed because the type of the key isn't really relevant when it's in a memory pair. Typically this is just used with vanilla condition statements. Such as only allowing something to activate when a memory key has reached a certain status.

patches/server/0842-WIP-Brain-API.patch
Owen1212055: Hmm should I make a new expiry method instead that is not null then?

Owen1212055:
Rebased to 1.19, redid the example, asking for feedback, you can now tick all entity brains, etc.

patches/server/0842-WIP-Brain-API.patch
yannicklamprecht: Redis has a concept we could follow.

You can 1. set a non-null value and 2. set its expiry. When 0 it will directly expire. If possible I would not use a long in API instead a duration to make conversion easier and calculation a non issue.
Defining two constant durations is recommended as well.
Like EXPIRE_NOW (0) and REMOVE_EXPIRY (-1). Both as duration for sure.

patches/api/0387-WIP-Brain-API.patch
```diff
@@ -0,0 +1,865 @@
...
++
++import java.util.HashMap;
++import java.util.Map;
++
++/**
++ * Represents an activity used by minecraft.
++ */
++public final class VanillaActivityKey implements ActivityKey {
++
++    public static final Map<NamespacedKey, ActivityKey> ACTIVITY_KEYS = new HashMap<>();
```
Lulu13022002, line 405: This field allow mutation for plugins which we don't really want here

fantahund:
Hey, would it be possible to add the BrainAPI for the camels? For example, that I can prevent the camels from lying down.

maestro-denery:
> Hey, would it be possible to add the BrainAPI for the camels? For example, that I can prevent the camels from lying down.

iirc in this draft API there is no modification of already existing entities' brains for now, but I think it would be cool, tho it would require a lot of code generation which is bad, because paper api has no good code generation tool or whatever for now.

maestro-denery:
oh, nvm, I deleted my previous comment, I think this API needs a rebase, but code gen for it is still an issue

Owen1212055:
For it to be more properly integrated, this PR will need to be redone with all the new registry-related APIs.
Additionally, Mojang has redone its brain system's format to be significantly different than what this API proposes.

I am closing this as in general this will need to be rewritten, and don't have the time for that atm.

PR #6973, opened 2021-12-01
Skip stripping metadata from empty stacks

Machine-Maker:
I couldn't figure out how exactly to reproduce the error, somehow an Empty itemstack makes it way into the equipment change handler. Best to skip them entirely I think.

Fixes https://paste.gg/p/anonymous/10bf03f791284577924a792e72bf0c8b


electronicboy (approved):

PaulBGD:
🙏

I'm not sure the repro either, but on a server with just mcmmo it would happen sometimes with stuff like dying/placing crafting tables..

PR #6976, opened 2021-12-01
Don't populate a useless spigot.yml field

Machine-Maker:
Spigot removed the `max-entity-collisions` from their yml file, but paper adds it back and tries to migrate from the spigot.yml file. This, previous to this PR, caused a default to be set in the spigot yml which caused the value to be saved to the file. This prevents the adding of that default so as to not cause confusion with the same `max-entity-collisions` appearing in both `spigot.yml` and `paper.yml`.

### DIFF
--- a/patches/server/0130-Cap-Entity-Collisions.patch
+++ b/patches/server/0130-Cap-Entity-Collisions.patch
@@ -12,7 +12,7 @@ just as it does in Vanilla, but entity pushing logic will be capped.
 You can set this to 0 to disable collisions.
 
 diff --git a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
-index f383f30b9dd1a7c6cf69d342f99118beec70b206..7cd6a36e000d9e1958d260739e6fcfb064402b4d 100644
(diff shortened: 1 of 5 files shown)


kennytv (approved):

patches/server/0130-Cap-Entity-Collisions.patch
```diff
@@ -55,3 +55,25 @@ index 36aef3dbfe407ca6fa206bc3fc4390e5c2770aa3..e0fe53045da81fc9b2eac3b215d641cf
...
+     }
+ 
+     public int getInt(String path, int def)
++    // Paper start - get int without setting default
+     {
+-        this.config.addDefault( "world-settings.default." + path, def );
+-        return this.config.getInt( "world-settings." + this.worldName + "." + path, this.config.getInt( "world-settings.default." + path ) );
++        return this.getInt(path, def, true);
++    }
++    public int getInt(String path, int def, boolean setDef)
```
MiniDigger: not sure if this should be added to the general config patch? guess we can move it later if we need more of these

Machine-Maker: yeah... I guess I'll move it now. There may be cases in the future where we want to do something similar.

Issue #6981, opened 2021-12-01
Update `falling-block-height-nerf` and `tnt-entity-height-nerf`


molor:
### Expected behavior

Values < 0 is supported and works correctly

### Observed/Actual behavior

Values < 0 will disable these features, but the minimum world height since 1.17 can be -63 or even -2031 if this is a custom dimension with custom `min_y`.

### Steps/models to reproduce

Set `falling-block-height-nerf` or `tnt-entity-height-nerf` in paper.yml to -63

### Plugin and Datapack List

This is a reminder from Paper's discord channel, so I think that this is not needed

### Paper version

Experimental 1.18

### Other

Don't know how to add ability to disable these features because of negative heights instead of values < 0, maybe just use boolean `false` instead of a number

kennytv:
As far as I can see that assumption is wrong? It's only disabled if the value is exactly 0

lynxplay:
Validated this on a local setup :+1:  0 is the only value that disables this feature. The documentation however is still mentioning that `Values less than 1 will disable this feature.` I'll open a respective issue :+1:

molor:
I don't understand how it's works. The description says `The height at which falling blocks will be removed from the server`. I set it to `0`  — seems that it was disabled. I'm set it to `-63` — falling blocks (in my case gravel) just drops as item every time at height >= (higher or equals, not lower!) -63. I'm set it to `8` — and behavior is the same like with -63, but slightly higher..

I think that server should remove falling block if it is _below_ the specified height, and not higher, as it works now. Or am I doing something wrong?

lynxplay:
You are correct in the usage, the server removes falling blocks/tnt when they are above the threshold specified.
It was initially implemented to limit flying tnt/falling blocks from going too high in the world if I recall correctly ([in this commit](https://github.com/PaperMC/Paper/commit/a26fef50efb43a1b75d532c88b12a18d74c696f4)).

PR #6986, opened 2021-12-01
Fix configs that relied on outdated min/max y levels

### LINKED ISSUE #7208: nether-ceiling-void-damage-height doesn't alter portal search
### Expected behavior

Portals made above the nether roof should not be considered as valid targets as they'll instant kill any player who travel to them when nether-ceiling-void-damage-height is set to the roof - it should try for a different portal instead

### Observed/Actual behavior

Player exits above roof, takes void damage, dies, and so do their items.

### Steps/models to reproduce

Build portal on roof, then set nether-ceiling-void-damage-height=128

### Plugin and Datapack List

NoCheatPlus
IllegalStack
BanItem
Pistonchat

### Paper version

[01:27:22 INFO]: This server is running Paper version git-Paper-400 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 4ac995a on ver/1.17.1)
You are 2 version(s) behind
Previous version: git-Paper-398 (MC: 1.17.1)

### Other

_No response_

Machine-Maker:
Fixes #7208
Related to #6981

These 3 paper yml settings relied on outdated information about min/max y values which leads to valid values meaning "disabled". This refactors and adds migrations to have a separate boolean for toggling its enable status and a level for configuring the level.

I tested the migrations both with values in the default world-settings section and the specific world settings.

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -923,10 +923,10 @@ index 0000000000000000000000000000000000000000..69add4a7f1147015806bc9b63a8340d1
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/PaperConfigurations.java b/src/main/java/io/papermc/paper/configuration/PaperConfigurations.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..18c64590b35a26a27a884494d6905989fc45af03
+index 0000000000000000000000000000000000000000..1055a079a061d8761be13d3b9efbbd90d193caa2
 --- /dev/null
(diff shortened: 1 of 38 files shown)


Machine-Maker:
Rebased for 1.18.1

electronicboy (approved):

Machine-Maker:
Rebased for 1.18.2. Also fixes teleporting into void damage.

electronicboy (approved):
merge reminder: config version needs bumping

kennytv (approved):

PR #6989, opened 2021-12-02
Fix NPE when getting destroy speed of air

SoSeDiK:
Prevents `java.lang.NullPointerException: Cannot invoke "net.minecraft.world.item.ItemStack.getItem()" because "nmsItemStack" is null` when calling `Block#getDestroySpeed(ItemStack itemStack)` for air.

### DIFF
--- a/patches/server/0538-Add-Destroy-Speed-API.patch
+++ b/patches/server/0538-Add-Destroy-Speed-API.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add Destroy Speed API
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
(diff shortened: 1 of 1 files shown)


electronicboy:
That method is explictly has a NotNull param, really should just have a preconditions check to warn about such a violation rather than just tryna silently ignore it

SoSeDiK:
> That method is explictly has a NotNull param, really should just have a preconditions check to warn about such a violation rather than just tryna silently ignore it

It's for `new ItemStack(Material.AIR)` and such items, not `null`

Edit: without this patch `block.getDestroySpeed( player.getInventory().getItemInMainHand() )` will throw NPE when holding air/nothing. Now it'll return `1.0` (default value for air/empty hand).

electronicboy:
Oh, in that case, I'm not fond of the fact that this patch basically reimplements asNMSCopy but without the copy, maybe that should be turned down into a CraftItemStack method which does the same semantics as Copy but, without the actual copy if it's an NMS stack, not fond of the cloned logic here more or less

electronicboy (approved):

PR #6994, opened 2021-12-02
Linking nether portals 1.18

GroGaloraGran:
Hello, i noticed that in 1.18 previously created portals lost their link. For example, if from a portal created earlier in Nether, which was normally linked in the upper world in 1.16.5 and 1.17.1, in 1.18 it leads many levels below. This does not work correctly in the reverse order, if you go from the upper world to the nether - it teleports to the wrong portal, which is much higher in Nether and was linked to another portal in the upper world. I’m not sure if this is a bug or a feature of the update.
my version with test:
This server is running Paper version git-Paper-19 (MC: 1.18) (Implementing API version 1.18-R0.1-SNAPSHOT) (Git: 892c292)
sorry if wrong category was selected


electronicboy:
issues tab for reporting issues

PR #6997, opened 2021-12-02
Use mojang item ids for alternative item despawn rate

### LINKED ISSUE #6996: Migrate "alternative item-despawn-rate" to use item ids instead of Material enum names
https://github.com/PaperMC/Paper/blob/master/patches/server/0364-Implement-alternative-item-despawn-rate.patch

NoahvdAa:
Fixes #6996. This will migrate old Material values over to Mojang item ids (or keys, I suppose)

This also makes worlds use the values from the default config when there is no world-specific config.


lynxplay:
As a general question to whoever wants to review this, why are we not caching/calculating the despawn rate once per item instance?

With this change we are now performing two hashmap lookups per tick, per item which could be drastically reduced to simply returning an integer that is calculated once (either lazy or just during object construction).

The only reason I could think of is the ability to reload both spigot and paper configs on the fly, however that seems, at least to me, like a rather poor argument to justify not optimizing this.

patches/server/0364-Implement-alternative-item-despawn-rate.patch
```diff
@@ -4,66 +4,97 @@ Date: Mon, 3 Jun 2019 02:02:39 -0400
...
-@@ -565,5 +565,52 @@ public class PaperWorldConfig {
+@@ -565,5 +565,68 @@ public class PaperWorldConfig {
              Bukkit.getLogger().warning("You have enabled permission-based Anti-Xray checking - depending on your permission plugin, this may cause performance issues");
          }
      }
 -}
  
 +    public boolean altItemDespawnRateEnabled;
-+    public java.util.Map<org.bukkit.Material, Integer> altItemDespawnRateMap;
++    public java.util.Map<String, Integer> altItemDespawnRateMap = new HashMap<>();
```
kashike: Should use `ResourceLocation` as the key instead.

kashike (changes requested):

NoahvdAa:
Rebased and updated to use ResourceLocation. I've also added caching to the despawn timeout (fetched when the itemstack is set) and merged the `Respect despawn rate in item merge check` patch into this one, since I feel like it doesn't need to be its own patch.

kashike (approved):

NoahvdAa:
Rebased

patches/server/0361-Implement-alternative-item-despawn-rate.patch
```diff
@@ -4,101 +4,149 @@ Date: Mon, 3 Jun 2019 02:02:39 -0400
...
+index e683e5bf47abe7bd3d2f7e9811a377549308ded4..a3c3656ba4e8bb85bfb3186d6284f02f09975b13 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+@@ -70,8 +70,8 @@ public class PaperConfig {
+         commands = new HashMap<String, Command>();
+         commands.put("paper", new PaperCommand("paper"));
+ 
+-        version = getInt("config-version", 24);
+-        set("config-version", 24);
++        version = getInt("config-version", 25);
```
electronicboy: changes to this should be in the OG patch

electronicboy (changes requested):

patches/server/0361-Implement-alternative-item-despawn-rate.patch
NoahvdAa: Thanks, fixed.

Machine-Maker (approved):
Tested with 
```yml
  default:
    alt-item-despawn-rate:
      enabled: false
      items:
        COBBLESTONE: 300
        STONE: 400
  world_nether:
    alt-item-despawn-rate:
      enabled: true
      items:
        NETHERRACK: 100
```

kashike (approved):

electronicboy (approved):

NoahvdAa:
Rebased for 1.18.2

NoahvdAa:
bad bot. Rebased.

patches/server/0356-Implement-alternative-item-despawn-rate.patch
```diff
@@ -3,102 +3,136 @@ From: kickash32 <kickash32@gmail.com>
...
++            mapSection = config.getConfigurationSection("world-settings.default." + path + ".items");
 +        }
-+        if(altItemDespawnRateEnabled) {
-+            for(org.bukkit.Material key : altItemDespawnRateMap.keySet()) {
-+                log("Alternative item despawn rate of " + key + ": " + altItemDespawnRateMap.get(key));
++        if (mapSection != null) {
++            for (String key : mapSection.getKeys(false)) {
++                int val = mapSection.getInt(key);
++                net.minecraft.resources.ResourceLocation keyLocation = new net.minecraft.resources.ResourceLocation(key);
++                if (net.minecraft.core.Registry.ITEM.getOptional(keyLocation).isPresent()) {
```
electronicboy: No longer logs invalid entries

electronicboy (approved):

PR #7010, opened 2021-12-03
Sanitize Sent BlockEntity NBT

Owen1212055:
The method BlockEntity#saveWithoutMetadata applies the PersistentDataContainer tag to the given tag.
A lot of block entities use this method in the getUpdateTag() method in order to receive certain data (like sign lines) on the client.

This removes this data sent to the client (this applies to lots of tile entities).
I also am wondering, should I move this method into the patch that adds
``` if (blockEntity instanceof net.minecraft.world.level.block.entity.SkullBlockEntity) { net.minecraft.world.level.block.entity.SkullBlockEntity.sanitizeTileEntityUUID(compoundTag); } // Paper```?

Although this issue was fixed, I was told it was supposed to be kept. Not sure what to do here about that.


Machine-Maker (approved):

Owen1212055:
Still relevant

electronicboy (approved):

Owen1212055:
Rebased

PR #7019, opened 2021-12-04
Add Moving Piston API

### LINKED ISSUE #6479: Expose underlying block in a moving piston

If accessing the world and finding a moving piston block, there is no way to determine what block is being moved. The only thing the API exposes, from what I tell, is that the block is moving. Wow.

An API feature to call `getMovingBlock` and return a Block instance of what is being moved.

Listening to the piston extend/retract event before the block is moved. 

### Agreements

### Other

_No response_

Owen1212055:
Resolves: https://github.com/PaperMC/Paper/issues/6479

This exposes the MovingPiston block entity into the api, which is used whenever a block is "moving" by a piston.

### DIFF
--- a/patches/server/0819-Add-Moving-Piston-API.patch
+++ b/patches/server/0819-Add-Moving-Piston-API.patch
@@ -1,0 +1,42 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sat, 4 Dec 2021 13:29:36 -0500
+Subject: [PATCH] Add Moving Piston API
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/block/CraftMovingPiston.java b/src/main/java/org/bukkit/craftbukkit/block/CraftMovingPiston.java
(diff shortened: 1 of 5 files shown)


patches/server/0819-Add-Moving-Piston-API.patch
```diff
@@ -0,0 +1,42 @@
...
+ import net.minecraft.world.level.block.piston.PistonMovingBlockEntity;
+ import org.bukkit.World;
+ 
+-public class CraftMovingPiston extends CraftBlockEntityState<PistonMovingBlockEntity> {
++public class CraftMovingPiston extends CraftBlockEntityState<PistonMovingBlockEntity> implements io.papermc.paper.block.MovingPiston { // Paper - Add Moving Piston API
+ 
+     public CraftMovingPiston(World world, PistonMovingBlockEntity tileEntity) {
+         super(world, tileEntity);
+     }
++    // Paper - Add Moving Piston API
```
NoahvdAa:
```suggestion
+    // Paper start - Add Moving Piston API
```

Owen1212055:
don't kill

Machine-Maker (commented):
Is there a reason this tile entity wasn't exposed in the API before? I know its kind of a special tile entity that doesn't exist for very long, but wondering if anyone else had some more info.

electronicboy (approved):

PR #5687, opened 2021-05-19
Fix sending CreatureSpawnEvent for portals

Machine-Maker:
Fixes [SPIGOT-6138](https://hub.spigotmc.org/jira/browse/SPIGOT-6138)

PR #7020, opened 2021-12-04
Prevent unnecessary creation of CraftEntity instances

Machine-Maker:
I don't really know if this is important or not, I just noticed that during Entity#load when an entity is being teleported between worlds it creates a new instance of whatever CraftEntity in at least these 2 places only to have it overwritten with the entity being copied's bukkit entity a few lines later in Entity#teleportTo

### DIFF
--- a/patches/server/0824-Prevent-unnecessary-creation-of-CraftEntity-instance.patch
+++ b/patches/server/0824-Prevent-unnecessary-creation-of-CraftEntity-instance.patch
@@ -1,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 4 Dec 2021 11:37:21 -0800
+Subject: [PATCH] Prevent unnecessary creation of CraftEntity instances
+
+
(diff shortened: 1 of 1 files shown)


patches/server/0824-Prevent-unnecessary-creation-of-CraftEntity-instance.patch
```diff
@@ -0,0 +1,47 @@
...
+@@ -2188,7 +2188,11 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+ 
+                 ((ServerPlayer) this).setLevel(bworld == null ? null : ((CraftWorld) bworld).getHandle());
+             }
+-            this.getBukkitEntity().readBukkitValues(nbt);
++            // Paper start - don't create a bukkit entity needlessly
++            if (this.bukkitEntity != null) {
++                this.bukkitEntity.readBukkitValues(nbt);
++            }
++            // Paper end
```
Machine-Maker, line 20: Might have to be a boolean field of whether to load this or not set before restoring an entity and reset after.

Machine-Maker:
nah, I changed my mind pretty quickly on this one. It shouldnt really matter after #5687 is merged.

PR #7022, opened 2021-12-04

Machine-Maker:

Bukkit Changes:
e4358b82 PR-686: Fix contains for default section generating real sections
0912ace4 SPIGOT-6830: Fix addDefaults with Configuration overrides child Sections in the defaults
62144c8b Add Player#openSign API to edit a placed sign

CraftBukkit Changes:
f49e9d19 PR-971: Remove strong chunk reference in PDC
27a27cdb Add Player#openSign API to edit a placed sign

---
Unfortunately, we can't do away with our two patches added the openSign functionality since we added it to HumanEntity and upstream added it to Player. Also, we didn't require the sign tile entity to be placed, and I'm not sure why it would have to be?


Doc94:
only can thing in maybe add a argument for "check if is placed" if is true then return the player method if not then the currently in patch and by default make openSign use the false argument for keep the logic here?
(maybe i explain very bad in this xd)

Machine-Maker:
Well what do we preserve? The functionality that's existed in Paper for years that doesn't care if its placed, or change it suddenly? I think the former. Its been that way in Paper for a while. **IF** upstream had used a different method signature, then that makes more sense. But since its the exact same method, I think we should keep Paper's behavior around "placed" or not.

electronicboy (approved):

jpenilla (approved):

PR #7023, opened 2021-12-05
Properly pass SplashPotion entity to damage source

lynxplay:
This commit modifies the potion splash event fixes added by paper, as
they passed the affected entity as the entity to the damage source,
meaning the entity would technically attempt to hurt itself.

This change fixes this by now properly passing the splash potion to the
damage source, allowing custom damage logic like the enderman water
damage logic, to properly detect the entity that is attempting to hurt
the affected target.


Machine-Maker (approved):

lynxplay:
Closing in favour of #7024

PR #7025, opened 2021-12-05
Forward CraftEntity in teleport command

Machine-Maker:
Just like in Entity#teleportTo the CraftEntity should be forwarded to the new nms Entity

### DIFF
--- a/patches/server/0070-Use-a-Shared-Random-for-Entities.patch
+++ b/patches/server/0070-Use-a-Shared-Random-for-Entities.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Use a Shared Random for Entities
 Reduces memory usage and provides ensures more randomness, Especially since a lot of garbage entity objects get created.
 
 diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
-index ddbbef22813831267c407d9864ce8decb5706df5..664d7199e21f0c617ba41bc7596ead99a0325792 100644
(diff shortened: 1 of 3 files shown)


electronicboy (approved):

patches/server/0824-Forward-CraftEntity-in-teleport-command.patch
```diff
@@ -0,0 +1,25 @@
...
+@@ -195,7 +195,14 @@ public class TeleportCommand {
+                     target.unRide();
+                     Entity entity1 = target;
+ 
++                    // Paper start - transfer bukkit entity
++                    org.bukkit.craftbukkit.entity.CraftEntity bukkitEntity = target.bukkitEntity;
+                     target = target.getType().create(world);
++                    if (bukkitEntity != null) {
++                        bukkitEntity.setHandle(target);
++                        target.bukkitEntity = bukkitEntity;
```
jpenilla: Should we maybe move this logic into Entity#restoreFrom instead? It's only called in TeleportCommand and teleportTo, and seems to be a better place for this

jpenilla (changes requested):

patches/server/0824-Forward-CraftEntity-in-teleport-command.patch
Machine-Maker: Yeah, and has the benefit of removing the need for the AT for the bukkitEntity field.

done.

Machine-Maker:
Rebased for 1.18.1

jpenilla (approved):

PR #7034, opened 2021-12-05
Add dropped hunk from mid-tick tasks

Machine-Maker:
I had to re-order it, cause the hunk depended on being after the optimise chunk iteration patch.


patches/server/0756-Execute-chunk-tasks-mid-tick.patch
```diff
@@ -107,8 +107,28 @@ index 088334869cb62797a1e1d1bbb6187f03189d852d..c245c1f4611f7273c8da629f774e0c64
...
+ 
++            int chunksTicked = 0; // Paper
+             try {
+             while (iterator1.hasNext()) {
+                 LevelChunk chunk1 = iterator1.next();
+@@ -1046,6 +1047,7 @@ public class ServerChunkCache extends ChunkSource {
+                         this.level.tickChunk(chunk1, k);
+                     }
+                 }
++                if ((chunksTicked++ & 1) == 0) net.minecraft.server.MinecraftServer.getServer().executeMidTickTasks(); // Paper
```
jpenilla: Not sure when this changed (as in 1.17 it's as you have here), but at one point this call was made directly after the call to tickChunk (see https://github.com/Tuinity/Tuinity/blob/ver/1.16.5/patches/server/0015-Execute-chunk-tasks-mid-tick.patch#L234), so I think that should be fixed

jpenilla (changes requested):

jpenilla (approved):

PR #7037, opened 2021-12-05
FallingBlock auto expire setting

Owen1212055:
This gives the ability to disable falling blocks from being able to expire automatically.
This also saves with nbt under ``DisableAutoExpire``, although I am not sure this behavior is needed.


imDaniX:
Maybe a couple of options to actually set the expire timers instead? Also, what's the use-case for this nbt?

Owen1212055:
> Maybe a couple of options to actually set the expire timers instead? Also, what's the use-case for this nbt?

What do you propose? So like modifying it without changing the entity age? (Currently set age modifies this timer and the entity age) And I’m not sure, since I don’t know if new fields like this should be serialized into the entity or not.

imDaniX:
> What do you propose? So like modifying it without changing the entity age? (Currently set age modifies this timer and the entity age) And I’m not sure, since I don’t know if new fields like this should be serialized into the entity or not.

Sorry, actually did read an entire PR wrong.

Owen1212055:
~~Simplified by removing the tag, this is now a simple boolean.
Current variables for falling blocks by spigot aren't retained either... so I felt perhaps it wasn't needed to keep it.~~

This is a rather simple API addition and is ready.

Owen1212055:
Rebased

patches/server/0877-FallingBlock-auto-expire-setting.patch
```diff
@@ -0,0 +1,68 @@
...
+ 
+     }
+ 
+@@ -360,6 +362,10 @@ public class FallingBlockEntity extends Entity {
+             int srcZ = nbt.getInt("SourceLoc_z");
+             this.setOrigin(new org.bukkit.Location(level.getWorld(), srcX, srcY, srcZ));
+         }
++
++        if (nbt.contains("Paper.AutoExpire")) {
++            this.autoExpire = nbt.contains("Paper.AutoExpire");
```
kennytv: get the boolean value

kennytv (changes requested):

patches/server/0877-FallingBlock-auto-expire-setting.patch
Owen1212055: LOL nice, will do

Owen1212055:
Rebased and fixed

kennytv (approved):

PR #6613, opened 2021-09-15
Rework filtering spawn egg and tile entity nbt config

Machine-Maker:

PR #7038, opened 2021-12-05
Improve behavior of filter-nbt-data-from-spawn-eggs-and-related config

Owen1212055:
This patch seemed to no longer work exactly anymore, as at least the behavior of filtering is no longer in this patch.

This readds it, and also instead of completely making blocks like command blocks turn to stone instead just wipe all tile entity data (opinion wanted on this).

Other nbt tags to remove can be suggested!

### DIFF
--- a/patches/server/0110-Filter-bad-data-from-ArmorStand-and-SpawnEgg-items.patch
+++ b/patches/server/0110-Filter-bad-data-from-ArmorStand-and-SpawnEgg-items.patch
@@ -5,7 +5,6 @@ Subject: [PATCH] Filter bad data from ArmorStand and SpawnEgg items
 
 Co-authored-by: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
 
-
 diff --git a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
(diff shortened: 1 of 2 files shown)


Machine-Maker:
This kinda sorta duplicates an already existing PR https://github.com/PaperMC/Paper/pull/6613

Owen1212055:
> This kinda sorta duplicates an already existing PR #6613

I mean this is just adding onto something that already exists and wasn’t updated. The original patch would also remove tags, but it got removed sometime. Does it hurt to have both? I don’t know, because this is just supposed to remove tags that we consider “dangerous”. You can have the decision here however, tbh.

Machine-Maker:
I this the configurability of #6613 is better, leaves it open to server owners with sane defaults.

However I do think it's better to wipe tile entity data instead of setting to stone which is the other thing your PR does.

Owen1212055:
> I this the configurability of #6613 is better, leaves it open to server owners with sane defaults.
> 
> 
> 
> However I do think it's better to wipe tile entity data instead of setting to stone which is the other thing your PR does.

What do I do with the patch then? This adds a config option that only half works, and only prevents special falling sand blocks. Should I just rename it?

Owen1212055:
@Machine-Maker Do you have any update on how you want this handled? This PR is a bit different, and tbh, this option seems a bit pointless to me anyways with ur PR.

But it leaves the question if we were to go with ur PR instead should the config option just be removed then?

Machine-Maker:
Handled in https://github.com/PaperMC/Paper/pull/6613 (Owen said to)

PR #7041, opened 2021-12-06
Properly resend entities

Owen1212055:
In some cases, when an entity is resent due to people being able to do funky things in the API it doesn't resend the entire entity properly.

Fixes
- ~~Axlotls disappearing when trying to cancel interact event and clicking on them with water bucket (they don't extend AbstractFish, but they do implement Bucketable)~~
- Bucketed entities not sending all their data when pickup is cancelled (such as passengers)

I also renamed the base patch to handle a variety of different scenarios related to entities not being sent correctly, but set myself as co author (I am not sure if I am suppoed to, but I don't want to "steal" the patch).


Owen1212055:
Rebased

Owen1212055:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/979/overview

There is now a PR upstream for fixing the axolotl part of this issue.

Owen1212055:
Rebased

Machine-Maker (approved):

Owen1212055:
Rebased, please remove tag.

HexedHero:
Using https://www.spigotmc.org/resources/unbreakableleashes.73020/ if you teleport outside of the render distance with a leashed entity like a horse their position isn't updated and they are invisible until they get hurt etc. with a visual effect of you leashing an entity from 100's of blocks away. The plugin is open-source and is simply using API. The only way to get around this is to put this section of code https://gitlab.com/ranull/minecraft/unbreakableleashes/-/blob/master/src/main/java/com/ranull/unbreakableleashes/manager/LeashManager.java#L33-40 in a 2 tick or longer delay task. I had a similar issue wth my own code when updating to 1.18, The UnbreakableLeashes plugin is just a easy way to reproduce it. Would this be something considered to be fixed in this PR?

Owen1212055:
It does not, explanation can be seen https://canary.discord.com/channels/289587909051416579/555462289851940864/972574161417424906)

electronicboy (approved):

PR #7042, opened 2021-12-06
[ci skip] Add nullable annotation to field

Owen1212055:
Not sure if this change needs to go through, but it's just a one-line change. 

Intellij always marks this field as not able to be null, and although it's not null when the entity is alive it can be null when it's dead. Generally just a bit misleading.

### DIFF
--- a/patches/server/0242-Add-Debug-Entities-option-to-debug-dupe-uuid-issues.patch
+++ b/patches/server/0242-Add-Debug-Entities-option-to-debug-dupe-uuid-issues.patch
@@ -79,7 +79,7 @@ index 2a11514554b6aea819046282cfcaeeb43d1ed920..2981a29b011cb1a08a776abc5ec6a942
      public com.destroystokyo.paper.loottable.PaperLootableInventoryData lootableData; // Paper
      private CraftEntity bukkitEntity;
  
-+    public @javax.annotation.Nullable net.minecraft.server.level.ChunkMap.TrackedEntity tracker; // Paper
++    public @org.jetbrains.annotations.Nullable net.minecraft.server.level.ChunkMap.TrackedEntity tracker; // Paper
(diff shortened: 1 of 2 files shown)


patches/server/0242-Add-Debug-Entities-option-to-debug-dupe-uuid-issues.patch
```diff
@@ -72,20 +72,20 @@ index 9a51aa9f40ceaaa3500595bf08b2b5515704b91a..9a9ae2cac9f44e38f58bd58c279fc648
...
-index 5df955eea09b5abd32c12abf5d13030bf9a3a9ae..d442725c1c9ae274b11f3a11ea4cf15f51d62def 100644
+index 2a11514554b6aea819046282cfcaeeb43d1ed920..780bdbefad1edbeec1f5218514efde36d8390cc3 100644
 --- a/src/main/java/net/minecraft/world/entity/Entity.java
 +++ b/src/main/java/net/minecraft/world/entity/Entity.java
 @@ -170,6 +170,8 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
      public com.destroystokyo.paper.loottable.PaperLootableInventoryData lootableData; // Paper
      private CraftEntity bukkitEntity;
  
-+    public net.minecraft.server.level.ChunkMap.TrackedEntity tracker; // Paper
++    public @javax.annotation.Nullable net.minecraft.server.level.ChunkMap.TrackedEntity tracker; // Paper
```
kashike:
```suggestion
+    public @org.jetbrains.annotations.Nullable net.minecraft.server.level.ChunkMap.TrackedEntity tracker; // Paper
```

kashike (changes requested):

kashike (approved):

Owen1212055:
Rebased and should be ready

Machine-Maker (approved):

PR #7047, opened 2021-12-06
Expose internal ids for entity types

Yive:
Entity types IDs in the Bukkit API rarely match the internal entity type IDs in the NMS. This will allow for plugins that make use of packets to no longer need reflection or Burger to get the proper entity type IDs for the server's version.


kashike:
👎 I am not a fan of this - packets are not part of the API.

jpenilla:
I'm not sure why you would need to use reflection, paperweight userdev would allow you to call this method (and is the supported way to use internals)

PR #7048, opened 2021-12-06
Cache GameRule.Key's id's hashCode

rafaelsms:
It is used every random tick and mob spawn for map retrieval. Over a long period of time, it starts to add up on spark. Memory impact is small as there is only a few game rules.


kennytv:
Can you provide the spark profile? Strings already cash hash codes with nothing more than an extra `== 0` check...

rafaelsms:
I made this a while ago, didn't submit it to git though. So I don't know which spark profile I was looking.

Didn't know Strings already cached hash codes! I guess we can close it now.
Perhaps the spark's `GameRule.map.get(...)` made me believe the reason was recalculating hashCode() every time on HashMap#get.

PR #7049, opened 2021-12-06
Optimize despawn when no player nearby

rafaelsms:
I don't believe there is a reason to calculate distance here. We should discard the entity since there is no valid player on the level close enough to keep the entity alive.


Titaniumtown (approved):
This patch has been personally working well for me. Maybe worth looking at?

electronicboy:
This should go into the existing patch there

rafaelsms:
Will update the PR fixing this. Sorry for not closing/updating this earlier, I was busy.
Edit: the city I am currently in has a congested internet connection, this is taking way longer than what I expected

```
<=========----> 73% EXECUTING [16m 37s]
> :downloadSpigotDependencies
```

Edit: fixed it :)

mexiscool:
I don't have time to test it but I have a question. Does this make it if there are not players for example in the nether and hostile mobs get sent from the overworld to the nether trough a portal. Do they then instantly despawn in the nether or not? In vanilla if there are no players in a dimension and hostile mobs are loaded they do not despawn until a player goes into the overworld and is further away than 128 distance. Just asking to be sure.

electronicboy:
anilla doesn't tick empty worlds, craftbukkit and forks do; This patch is100% irrelevant here and just optimises existing behavior

mexiscool:
Nether portal chunkloaders can keep chunks loaded indefinitely without players

electronicboy:
which is irrelevant here

PR #7050, opened 2021-12-06
Fix simulation distance

rafaelsms:
Previously some block ticking was occurring outside of simulation distance and tickChunks was called only for entity ticking chunks, which is 1 chunk radius lower than block ticking chunks.

Also, I renamed viewDistance to distance on AreaMap, but now I don't think it is necessary. I am not very good with patches and git, so I don't know how to merge this and upstream/master since this PR breaks some patches (I used rebuild patches).

The git commit message about low TPS was fixed by allocating more ram (who knew, Xmx=1GB is not enough for 32 view distance even for debugging!)


rafaelsms:
I'll just redo the patches... I tried merging and I am sure I'm doing something wrong

MiniDigger:
if you need help be sure to ask in #paper-dev, we are happy to assist!

PR #7053, opened 2021-12-06
Fix ticking distances (mob spawn range, entity, chunk and block ticking distances)

### LINKED ISSUE #7054: [1.18] Simulation distance behaves differently than vanilla
### Expected behavior

The various forms of ticking (block, fluid, entity, etc) to be respected when simulation distance is set (and behave closer to vanilla)

### Observed/Actual behavior

Some forms of ticking (such as fluids and redstone) happen regardless of simulation distance with Paper, which is not the same behavior as Vanilla.

### Steps/models to reproduce

I found it easier to demonstrate and explain the issue with voice over and a video

Vanilla behavior: https://www.youtube.com/watch?v=XY8AQFoaT0w

Paper behavior: https://www.youtube.com/watch?v=kABeLz6tcf8

### Plugin and Datapack List

No plugins or datapacks - all testing was done with a fresh instance of Vanilla and Paper.

### Paper version

This server is running Paper version git-Paper-55 (MC: 1.18) (Implementing API version 1.18-R0.1-SNAPSHOT) (Git: 0e25db2)

### Other

It looks like #7053 might address this issue? (edit: maybe not) I have not tested aforementioned PR.

It's also worth noting that it's quite odd that vanilla has the entity ticking and other forms of ticking be offset by a chunk from each other... I would expect simulation distance to apply to all forms of ticking equally just like leafs no-tick. I question if maybe Paper should make simulation distance behave better in this regard? (although if no-tick is coming back, might be pointless)

rafaelsms:
* Mob spawn range: trackers have been updated to use Spigot's mob spawn range instead of MOB_SPAWN_RANGE (static value of 8)
* Mob spawn range is still limited to 8 chunks max, but now we either limit it by the view distance or the simulation distance, the greatest of both
* Trackers that used view distance have been replaced with simulation distance (such as ticking blocks and entities) as we would only iterate through more chunks without any benefit
* Chunk ticking now uses the maximum range between mob spawn and chunk ticking (which uses simulation distance), it'll check only one condition and assume the other one
  - for example, if mob spawn > simulation distance, we will iterate every chunk in mob spawn range, but we will only check if we should tick on those chunks
  - if mob spawn < simulation distance, we will iterate every chunk in simulation distance, but we will only check if we should spawn mobs on those chunks
* Entity activation is already updated by Spigot's ActivationRange, which is limited by simulation distance
* Entity ticking already checks chunk's ticking tracker


patches/server/0829-Use-simulation-distance-on-block-and-chunk-ticking-f.patch
```diff
@@ -0,0 +1,225 @@
...
+-
+     private CompletableFuture<Either<List<ChunkAccess>, ChunkHolder.ChunkLoadingFailure>> getChunkRangeFuture(ChunkPos centerChunk, int margin, IntFunction<ChunkStatus> distanceToStatus) {
+         List<CompletableFuture<Either<ChunkAccess, ChunkHolder.ChunkLoadingFailure>>> list = new ArrayList();
+         List<ChunkHolder> list1 = new ArrayList();
+@@ -1791,7 +1783,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+                 }
+             }
+         } else {
+-            final double range = (DistanceManager.MOB_SPAWN_RANGE * 16) * (DistanceManager.MOB_SPAWN_RANGE * 16);
++            final double range = (level.spigotConfig.simulationDistance * 16.0) * (level.spigotConfig.simulationDistance * 16.0);
```
rafaelsms, line 75: this is executed when reducedRange = false, which is removed from the code now (after this patch, reducedRange = true always)

rafaelsms:
#7054 will be fixed by this PR (edit: I don't know how to link issues)

electronicboy:
modifications to existing patches should be made in those patches rather than creating an additional one

rafaelsms:
> modifications to existing patches should be made in those patches rather than creating an additional one


Ok! I'll try to do it later. The last time I tried, I couldn't get it to rebuild patches correctly from the commits and some changes were left out in some patches. I'll try to do it again more cleanly this time. Any tip would be appreciated.

rafaelsms:
Well, the block ticking issue was fixed by Spottedleaf (758e2a7), so I will close this as my changes are a little bit hash on vanilla working.

PR #7055, opened 2021-12-07
Fix multiple ItemStack array nullability mistakes

MrPowerGamerBR:
`setContents` requires non null ItemStacks, so, if you want to restore an inventory's content, it is kinda messy:
```kotlin
val inventoryContents = player.inventory.contents // Array<ItemStack?>
player.inventory.setContents(inventoryContents) // doesn't work because it requires a Array<ItemStack>
player.inventory.setContents(inventoryContents.map { it ?: ItemStack(Material.AIR) }.toTypedArray())
```

This PR changes `setContents(...)` to match the `getContents()` return type.

Here's the discussion about this on Paper's Discord server: https://canary.discord.com/channels/289587909051416579/555462289851940864/917552139650351144

### DIFF
--- a/patches/api/0175-Fix-Spigot-annotation-mistakes.patch
+++ b/patches/api/0175-Fix-Spigot-annotation-mistakes.patch
@@ -105,8 +105,31 @@ index 2ff1b1308571d8f8056d3359e8a8ba4a589c3726..e669ad8ecd182c6899c7820414e6ee1f
      @NotNull
      public int[] getExpLevelCostsOffered() {
          int[] levelOffers = new int[offers.length];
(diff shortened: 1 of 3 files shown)


Machine-Maker:
You seem to have accidentally moved an api method out of one patch and into another.

MrPowerGamerBR:
> You seem to have accidentally moved an api method out of one patch and into another.

Yeah noticed that after making the PR, ended up editing a newer version of the Inventory class and that ended up happening, fixing that asap

kennytv (approved):

electronicboy (approved):

PR #7058, opened 2021-12-07
Add EntityPushedByEntityEvent (#5886)

### LINKED ISSUE #5886: Hitting a TNT Minecart does not trigger EntityKnockbackByEntityEvent
### Expected behavior

I melee a TNT Minecart with a knockback 2 sword and expect my message listening on EntityKnockbackByEntityEvent to print out the EntityTypes of both #.getEntityType() and #.getHitBy().getEntityType(). (The event should trigger properly.)

### Observed/Actual behavior

No message is printed out because hitting a TNT Minecart (or regular minecart) does not trigger EntityKnockbackByEntityEvent. Also tested with a punch bow on mobs (does trigger) but does not trigger on the Minecart.

### Steps/models to reproduce

Listen on EntityKnockbackByEntityEvent and print a message any time event is fired with the entitytype of the hit entity. Hit a TNT Minecart or regular minecart.

### Plugin list

ApolloUHC (my custom plugin)

### Paper version

git-Paper-"3d43c5e"

### Agreements

### Other

Using a knockback sword better shows the minecart getting knocked back. Spam click the minecart so that it does not break immediately.

aerulion:
*Edit:
Since the EntityKnockbackByEntityEvent can't be reused on non-living entities due to being implemented with the use of the LivingEntity interface, this PR adds a new EntityPushedByEntityEvent that does get called on collision pushes as well as pushes caused by an entity attack. These two cases can be distinguished by the PushReason enum.  Closes issue #5886.

This is my first time working with patches, so I have no idea if I did it correctly. Please tell me if I did something wrong, and I will try to fix it as fast as I can.


patches/api/0114-Add-EntityKnockbackByEntityEvent.patch
```diff
@@ -65,8 +64,8 @@ index 0000000000000000000000000000000000000000..9efecabab813f575bb447a356e5e7e95
 +     */
 +    @NotNull
 +    @Override
-+    public LivingEntity getEntity() {
-+        return (LivingEntity) super.getEntity();
++    public Entity getEntity() {
++        return super.getEntity();
```
Machine-Maker: You can't do this to the API. what if a plugin is using this event and is relying on this method returning a LivingEntity, as the contract currently states. but you change it to an entity (which I think is an ABI break), and suddenly the plugin is gonna have all sorts of problems.

Machine-Maker (changes requested):
I don't think you can re-use this event for non-LivingEntitys. The event probably should've been created without using the LivingEntity interface, but that didnt happen.

aerulion:
Thank you for the review. I understand that changing existing API in that way is something that shouln't be done. Sorry for the inconvinience. In this case, would it be reasonable to implement new event like a "EntityPushEntityEvent"? That would be a bit misleading in my opinion since it is only fired when pushed by attacking, and only covers entities that aren't covered by the existing one.

aerulion:
I hope implementing this new event provides a better solution to the initial problem. I implemented the event not only for pushes by attack, but also for collisions, since I think it less misleading and generally more useful in this way. Furthermore, I hope this doesn't have a negative impact on performance, since the event gets called quite often.

PR #7061, opened 2021-12-08
Expand Panda API

Owen1212055:
Allows you to play with some of the animations that pandas have.

Such as...
Sitting
image
On back
image
Rolling
https://user-images.githubusercontent.com/23108066/145124804-de6c8fb9-b9b2-4da8-818a-950e9699a110.mp4
and more!

Currently sneezing has a tick and boolean statement, as I am not sure this is supported, but negative numbers work fine when "sneezing"... IDK if this behavior should be kept or not but it's funny.
https://user-images.githubusercontent.com/23108066/145125032-048fe4e4-a4a5-4f3e-b536-58b92bdd1e93.mp4

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -221,6 +221,8 @@ public net.minecraft.world.entity.animal.Fox isDefending()Z
 public net.minecraft.world.entity.animal.Fox setDefending(Z)V
 public net.minecraft.world.entity.animal.Fox isFaceplanted()Z
 public net.minecraft.world.entity.animal.Fox setFaceplanted(Z)V
+public net.minecraft.world.entity.animal.Panda getEatCounter()I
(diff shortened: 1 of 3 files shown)


MiniDigger:
I am not sure about the documentation, while its very detailed, it might be too implementation specific? if mojang changes any of these numbers we will be provided outdated and invalid info :/
what do others think?
else looks good

electronicboy (approved):
I agree. with Mini, I'm not fond of vanilla impl detail being in the docs outside of urgent need, given that this documentation is rarely ever updated when it's added

provisionally approved

Owen1212055:
Thanks, point understood. I removed that documented behavior.

Machine-Maker (approved):
Tested all the various new methods, and they work as expected.

PR #7069, opened 2021-12-10
Update Log4J

darbyjack:


electronicboy (approved):

patches/server/0831-Bump-log4j-to-snapshp.patch
```diff
@@ -0,0 +1,34 @@
...
+
+diff --git a/build.gradle.kts b/build.gradle.kts
+index 5678a9f8d2aa8a7afbc5729570ec297b12acf75d..e6b4d8ca94f670dd969d28d3274f08223e140c36 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -14,6 +14,9 @@ repositories {
+     maven("https://maven.fabricmc.net/") {
+         mavenContent { includeModule("net.fabricmc", "mapping-io") }
+     }
++    maven("https://repository.apache.org/content/groups/snapshots/") {
```
jpenilla: This needs to go in libraryRepositories in the root build file as well.

jpenilla (changes requested):
Should check whether `com.lmax:disruptor` should be bumped as well.

Also need to look into whether https://github.com/PaperMC/Paper/blob/master/patches/server/0717-Set-AsyncAppender-dispatch-thread-to-be-a-daemon-thr.patch should be dropped

jpenilla (approved):

clrxbl:
Log4J 2.15.0 is releasing, shouldn't we move to that?

electronicboy:
is it out yet?

clrxbl:
https://github.com/apache/logging-log4j2/pull/608#issuecomment-990474429

Seems like it might take a few hours, but it might be worth waiting for the full release instead of moving to snapshots right now. (especially with the fix already in Paper)

electronicboy:
fix is missing a few expects, hence snapshot; Will take them a few hours to release and then a few hours to propagate often, easy to just go for it and update post that

ken-kentan:
Do you have a plan to backport this update to 1.17.1?

darbyjack:
Working on that right now!

Luca-Guettinger:
Is there a plan to backport this update to 1.15.2?

ghost:
if you need to mitigate the log4j exploit below 1.18, you can follow the instructions over [here](https://www.minecraft.net/en-us/article/important-message--security-vulnerability-java-edition)

kennytv:
The legacy builds have already been updated. See the Minecraft post for further information or just use Vanilla clients/recently updated mod loaders.

PR #7081, opened 2021-12-10
Add translation keys to GameMode enum

Machine-Maker:
GameMode's really have two associated translation keys, a longer form (one that includes the "Mode" word at the end) and a shorter form. The short form is only used on the client, so I opted for the long format to be default via the `Translatable` interface, but added two methods for clarity.

### DIFF
--- a/patches/api/0220-Add-methods-to-get-translation-keys.patch
+++ b/patches/api/0220-Add-methods-to-get-translation-keys.patch
@@ -86,6 +86,38 @@ index 4a97e73ce59c0eee77661967f1d3ac23508aae3e..d543b2b6aa131efec9b978d0b71a228f
      }
  
      /**
+diff --git a/src/main/java/org/bukkit/GameMode.java b/src/main/java/org/bukkit/GameMode.java
+index 938c3217f92e6d3ef9a637269c469f8359af6347..ef49495909a37d718a87d5dfbcd644d46e27fa88 100644
+--- a/src/main/java/org/bukkit/GameMode.java
++++ b/src/main/java/org/bukkit/GameMode.java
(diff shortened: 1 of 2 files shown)


patches/api/0222-Add-methods-to-get-translation-keys.patch
```diff
@@ -85,6 +85,63 @@ index 4a97e73ce59c0eee77661967f1d3ac23508aae3e..d543b2b6aa131efec9b978d0b71a228f
...
++    }
++
++    /**
++     * Gets the shorter format translation key.
++     *
++     * @return the short form key
++     */
++    public @org.jetbrains.annotations.NotNull String shortTranslationKey() {
++        return "selectWorld.gameMode." + this.name;
++    }
```
kashike: Not a fan of this method - this is using a very specific translation key (for a client-side screen).

Machine-Maker: Why can’t it have a method? It’s a valid translation key. I’m pretty sure there are other translationckey methods that return things only used on the client.

kashike: It's not the fact that it's something only used by the client, but the fact that it's a very specific screen's translation keys.

Machine-Maker: I can change the method name or javadoc to mention that. But since it *is* a key based on the game mode name, I think it should be exposed.

Prob need some more people to chime in.

Owen1212055: I don't think it hurts to have.
However, I think that you should add an English example in the javadocs of what differentiates the longer vs shorter version.

Spectator / Spectator Mode

electronicboy (approved):

PR #7083, opened 2021-12-11
Update Cache palette array to 1.18

NoahvdAa:
1.18 allows palette arrays to have different lengths than 4096, so I figured it'd make sense to move the cached palette array into the strategy/provider.

CCing @PaulBGD since he made this patch and knows a lot more about chunk palettes than I do.


kennytv (changes requested):
Nope, that is very wrong

Spottedleaf:
I have very significant doubts about this patch being remotely effective on this version anymore. Unless you have shown you have put in the effort to prove that it is effective, this will remain closed

NoahvdAa:
I assumed it would still be necessary because of the comment on the patch, I should've done more testing, my bad

PR #7085, opened 2021-12-11
Add Movement Direction API

Owen1212055:
This API is a bit niche... but it does have use cases especially for trying to implement mounting with full WASD controls.
This basically exposes the raw movement directions that entities use when moving in a direction.

Certain entities can move sideways, for example, skeletons.
However, when these values are not updated they are still ticked and it causes them to get smaller and smaller
image

I would like opinions on how to document this.


Owen1212055:
Still gonna let this hang around, unless it's too much of a niche api.

ryantheleach:
Is this effectively a 'velocity' of sorts? e.g. it represents the mobs move intention, but decays/smooths due to a 'friction' until reaching 0?

Edit: I'm not expert, but after some googling I found the term desiredVelocity being used elsewhere in game development.

https://docs.unity3d.com/ScriptReference/AI.NavMeshAgent-desiredVelocity.html

https://forums.unrealengine.com/t/set-velocity-on-character-movement/28143/2

If it's modifying the velocity by an offset, that would be a force or impulse of some kind.

Owen1212055:
> Is this effectively a 'velocity' of sorts? e.g. it represents the mobs move intention, but decays/smooths due to a 'friction' until reaching 0?
> 
> Edit: I'm not expert, but after some googling I found the term desiredVelocity being used elsewhere in game development.
> 
> https://docs.unity3d.com/ScriptReference/AI.NavMeshAgent-desiredVelocity.html
> 
> https://forums.unrealengine.com/t/set-velocity-on-character-movement/28143/2
> 
> If it's modifying the velocity by an offset, that would be a force or impulse of some kind.

I apologize for not responding sooner, I missed this. Yes, it represents a sort of velocity, and can basically be represented as "how much does an entity want to move foward/sideways".

This is usually used for things like player mounts as well, for determining how much forward/sideways the entity should move.

kennytv:
Should maybe use a vector instead?

Machine-Maker (approved):

lynxplay (approved):

PR #7089, opened 2021-12-11
Schoolable Fish API

Owen1212055:
image

This isn't exposed in any form currently, as a result, this adds a new ``SchoolingFish`` class which a couple of fish implement.


patches/api/0346-Schooling-Fish-API.patch
```diff
@@ -0,0 +1,92 @@
...
++     */
++    int getSchoolSize();
++
++    /**
++     * Gets the fish that this entity is currently following.
++     *
++     * @return following fish
++     */
++    @Nullable
++    SchoolingFish getCurrentFollowedFish();
```
MiniDigger:
```suggestion
+    SchoolingFish getCurrentlyFollowedFish();
```
english isn't my main lang but this sounds more correct to me?

MiniDigger:
SchoolingFish also sounds wrong to me, but don't have a better name at hand

Owen1212055:
> SchoolingFish also sounds wrong to me, but don't have a better name at hand

What about something like how Minecraft names it, “SchoolableFish”. I’m not sure of any other name.

patches/api/0346-Schooling-Fish-API.patch
yannicklamprecht: Would 
getSchoolFollowTarget()
fit?

Owen1212055:
I renamed the interface and method.

build-data/paper.at
```diff
@@ -289,3 +289,7 @@ public net.minecraft.world.level.levelgen.SurfaceSystem getOrCreateRandomFactory
 
 # Fix removing recipes
 public net.minecraft.world.item.crafting.RecipeManager byName
+
+# Add Schooling Fish API
```
darbyjack: Shouldn't this be changed to Schoolable too? Also, the patch name in general could be renamed to Schoolable.

Owen1212055: Agreed, will do.

Owen1212055: Done

patches/api/0347-Schoolable-Fish-API.patch
```diff
@@ -0,0 +1,92 @@
...
++     */
++    int getSchoolSize();
++
++    /**
++     * Gets the fish that this entity is currently following.
++     *
++     * @return following fish
++     */
++    @Nullable
++    SchoolableFish getCurrentlyFollowedFish();
```
kashike: Why not `getSchoolLeader`?

patches/api/0347-Schoolable-Fish-API.patch
```diff
@@ -0,0 +1,92 @@
...
++     * leader.
++     */
++    void stopFollowing();
++
++    /**
++     * Returns if this fish under normal circumstances can be followed.
++     *
++     * @return can be followed or not
++     */
++    boolean canBeFollowed();
```
jpenilla: seems like the implementation for this will return false when a fish has no followers? also might want to expose the max school size

jpenilla (changes requested):

patches/api/0347-Schoolable-Fish-API.patch
Owen1212055: Correct, it's like this because it determines if other fish are able to follow that fish's school. 
image
It's used in this task, where if the fish doesn't have a leader or can be followed it will try to follow it's leader.

 Also exposed max school size!

patches/server/0840-Schoolable-Fish-API.patch
```diff
@@ -0,0 +1,110 @@
...
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/PaperSchoolingFish.java
+@@ -0,0 +1,59 @@
++package io.papermc.paper.entity;
++
++import net.minecraft.world.entity.animal.AbstractSchoolingFish;
++import org.bukkit.craftbukkit.CraftServer;
++import org.bukkit.craftbukkit.entity.CraftFish;
++
++public class PaperSchoolingFish extends CraftFish implements SchoolableFish {
```
Machine-Maker: Since a new non-interface has been added here to the whole CraftEntity hierarchy, I think it should be added to the massive method in CraftEntity for the creation of new entities.
image

Machine-Maker (changes requested):
I believe this has mob goals API implications because right now the `follow_flock_leader` goal is incorrectly typed to the bukkit `Fish` entity, which includes pufferfish to with said goal does not apply. This change has to be handled in the MobGoalHelper entity class map and the list of goal keys in VanillaGoal

patches/server/0840-Schoolable-Fish-API.patch
Owen1212055: Do I need to update 
image
as well?

Machine-Maker: yeah, I suppose it should be changed there too.

jpenilla (approved):
Not tested, but code looks fine

Owen1212055:
rebased

Machine-Maker (approved):

Lulu13022002 (approved):

Owen1212055:
Rebased and merged into patch

PR #7096, opened 2021-12-12
Remove java 8 and prior reflection code from CustomTimingsHandler

chickeneer:
For some unknown reason this wrong code got triggered for the first time, reported over at
https://github.com/aikar/commands/issues/353

As far as I can tell, this java version detection code has been broken forever. Anyway. Simple explanation on the issue is that it would detect java 17.0.1 as java 0 since the split would have 2 or more. It looks like this was written with the expectation that java.version wouldn't return these minor java versions.

Anyway, here is my fix. Let me know if anyone thinks it needs tweaking.

### DIFF
--- a/patches/api/0006-Timings-v2.patch
+++ b/patches/api/0006-Timings-v2.patch
@@ -3536,10 +3536,10 @@ index 5ca863b3692b2e1b58e7fb4d82f554a92cc4f01e..612958a331575d1da2715531ebdf6b11
 +
 +}
 diff --git a/src/main/java/org/spigotmc/CustomTimingsHandler.java b/src/main/java/org/spigotmc/CustomTimingsHandler.java
-index 44badfedcc3fdc26bdc293b85d8c781d6f659faa..3cbe5c2bb55dead7968a6f165ef267e3e2931061 100644
+index 44badfedcc3fdc26bdc293b85d8c781d6f659faa..123647bb10fc89508437d7a0bd3fd31d58ee7c82 100644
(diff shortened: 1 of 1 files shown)


jpenilla (changes requested):
This looks like it would work, but I can imagine it breaking with some obscure version strings still. My main question though, is that on 1.18, we require Java 17 or newer (and Java 16 on 1.17), so is this check even needed still?

chickeneer:
I think you bring a valid point. It is a paper patch completely here, so I am on board with dropping altogether the <=8 java version check since it should always be false.

jpenilla:
Yeah, I think that would be best.

chickeneer:
The patch diff is absolutely horrible to read. But...
all of the sunReflectAvailable stuff has been removed which was used for version <= 8 compatability. Due to java 17 requirement, the removed code would be constant condition never run by that requirement.

Everything below and including `new AuthorNagException("Deprecated use of CustomTimingsHandler. Please Switch to Timings.of` is unchanged and retained.

The remaining class is quite small now.

jpenilla:
Checking out the code locally I'm noticing that we could possibly use StackWalker to determine the calling plugin, but I think that would be fine as a separate PR, given this PR as is fixes a bug and doesn't change behavior regarding whether the calling plugin will be resolved. (edit: we also print the stacktrace anyways so not sure it's needed with our change to include jar names in stack traces)

jpenilla (approved):

PR #7097, opened 2021-12-12
Fix WorldGenRegion leak when converting pre-1.18 chunks

### LINKED ISSUE #7094: Increased memory usage when loading old chunks
### Timings or Profile link

N/A

### Description of issue

For chunks generated in the latest version of Paper-1.17.1-399 and subsequently loaded using Paper-1.18.1-69, memory usage rapidly increases until the server crashes. This only happens the first time these chunks are loaded.

It is possible to reproduce this issue easily without any plugins installed, but it occurs much more quickly when using a plugin to load the chunks.

To reproduce (without any plugins):
- Create a world in 1.17.1 and generate some chunks. For my testing I loaded between -1000, -1000 and 1000, 1000.
- Load the world in 1.18.1 and get in-game. Using spectator mode, fly through all of the chunks.
- Memory usage increases until the server crashes. This will happen after just a few minutes with a 4GB heap allocated.

It's also convenient to have the server GUI open to view the memory graph as this is occurring:

40a3bc6a-3c94-4cef-9d13-8f2e96b63706

### Plugin and Datapack List

```
> plugins
[18:22:25 INFO]: Plugins (0):
> datapack list
[18:22:33 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
[18:22:33 INFO]: There are no more data packs available
```

### Server config files

Default configuration for everything, except for level-seed (mentioned below).

### Paper version

```
> version
[18:26:18 INFO]: Checking version, please wait...
[18:26:19 INFO]: This server is running Paper version git-Paper-69 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 29bd57b)
Previous version: git-Paper-68 (MC: 1.18.1)
```

### Other

Used `level-seed=8190275241236263855` for consistent results while testing.

jpenilla:
The Blender passed in here holds a WorldGenRegion which contains a list of surrounding chunks

closes #7094

### DIFF
--- a/patches/server/0831-Fix-WorldGenRegion-leak-when-converting-pre-1.18-chu.patch
+++ b/patches/server/0831-Fix-WorldGenRegion-leak-when-converting-pre-1.18-chu.patch
@@ -1,0 +1,23 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jason Penilla <11360596+jpenilla@users.noreply.github.com>
+Date: Sun, 12 Dec 2021 04:43:30 -0800
+Subject: [PATCH] Fix WorldGenRegion leak when converting pre-1.18 chunks
+
+The Blender passed in here holds a WorldGenRegion which contains a list of surrounding chunks
+
+diff --git a/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java b/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
+index 96cb3e8cad9e7a5edd2a448ea88f2447104fbb5a..5aeaaae6f15050a2da271fe196d0a234ecafc8a1 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
++++ b/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
+@@ -410,6 +410,11 @@ public abstract class ChunkAccess implements BlockGetter, BiomeManager.NoiseBiom
+     }
+ 
+     public NoiseChunk getOrCreateNoiseChunk(NoiseSampler noiseColumnSampler, Supplier<NoiseChunk.NoiseFiller> columnSampler, NoiseGeneratorSettings chunkGeneratorSettings, Aquifer.FluidPicker fluidLevelSampler, Blender blender) {
++        // Paper start - create a new one each time to avoid leaking
++        if (blender != Blender.empty()) {
++            return NoiseChunk.forChunk(this, noiseColumnSampler, columnSampler, chunkGeneratorSettings, fluidLevelSampler, blender);
++        }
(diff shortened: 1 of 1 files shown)


jpenilla:
Tested to fix #7094, a better solution is probably to null out that reference somewhere when the noise chunk is "done" being used, not sure where that is however

kennytv:
works:tm:, but else no idea where that could be unreferenced

electronicboy:
I ain't tested against vanilla, no sane means to test there, but, this is at least an issue upstream

electronicboy:
image
Well, that explains that, the issue here is that everything basically ends up just creating these huge cache chains seemingly, this is defo an issue elsewhere, it's just things like chunky running full pelt appears to basically be worst-case scenario for this type of thing as it just loads a huge contig section of chunks

jpenilla:
Tested with fabric loader + fabric api + chunky, and the issue is present there as well. This is not a "vanilla" testing setup, but it's pretty close and enough for me to say this isn't a Paper-exclusive issue.

chickeneer:
I have tested the branch and agree that it appears to mitigate the leak. I don't have any real feedback on improving the PR. I would note that systematically loading tons of pre-conversion chunks with `World#getChunkAtAsync()` and then unloading with `Chunk#unload` did not appear to cause a leak (or at least not a noticeable one).

To give credence to the original the issue. I verified that the memory leak does in fact occur in vanilla single player when converting a world by just flying around. This is an issue that I had observed on several occasions in the past week on our fork which caused out of memory crashes. Didn't immediately report due to uncertainty on whether it was occurring upstream and personal time constraints.

Reproduction was easy and as mentioned I tested the fix to solve the issue I was personally experiencing. Haven't the foggiest idea how to improve the PR to have some amount of cache, as the fix strikes me as potentially having a large performance hit with zero caching.

electronicboy:
>  I would note that systematically loading tons of pre-conversion chunks with World#getChunkAtAsync() and then unloading with Chunk#unload did not appear to cause a leak (or at least not a noticeable one).

Yea, that's purely the cache which is leaking, if you have players moving around all over you're basically just gonna end up with chains of cached chunks, we could maybe turn this into a Cache of some form, but, I forget the implications of that given that those have cleaning logic to run after some time; I was gonna suggest to pop that they could probs take advantage of this knowledge other day when I spotted it as a test, kinda forgot

Common area this is seen is with stuff like people running things like chunky which just load up chunks basically like a snake and so just create an endless cache list going all the way down, chunky is pretty much literal worse case for this; Benchmarks would be interesting to see if this has a massive impact, but, unless mojang gives us word soon™ i think going without the cache is better than committing servers to guaranteed death if they're not new maps

electronicboy (approved):

PR #7102, opened 2021-12-12
Dolphin Api

Owen1212055:
Exposes some dolphin things.
Dolphins have the ability to dry out and drown. It wasn't possible to retrieve the value for ``drying`` out, so this exposes that and some small other little values.

If a different name for the "dryout" behavior could be suggested that would be great, as "moistness" is a word I am not sure I should use here.


patches/server/0686-Missing-Entity-Behavior-API.patch
```diff
@@ -106,6 +106,47 @@ index a4f909123de26d911aea7cd767d2315ed1f697c9..0eee53c068bca070a86645d0ba54fb1a
...
++    }
++
++    @Override
++    public boolean hasFish() {
++        return this.getHandle().gotFish();
++    }
++
++    @Override
++    public org.bukkit.Location getTreasureLocation() {
++        return net.minecraft.server.MCUtil.toLocation(this.getHandle().level, this.getHandle().getTreasurePos());
```
Machine-Maker: Not sure about this... As ugly as the bukkit Location class is, this might have to be a null world. The entity doesn't track the world, its only a BlockPos. adding a world here, might incorrectly have plugins make that assumption that the treasure location the dolphin was actually referring to was in the world associated with the returned location.

Owen1212055: Although I agree, I took reference from turtle api here. Is it worth the inconsistency? The explanation makes sense but idk, does it matter?

Owen1212055: image
This is currently the turtle api, which uses a block position as well which is wrapped in the bukkit location. I'm not sure it's really worth the inconvenience it brings + inconsistency here, especially since really this represents a turtle that has an active world. If the entity isn't in a world, it'll just be null.

patches/api/0347-Dolphin-API.patch
```diff
@@ -0,0 +1,62 @@
...
++    Location getTreasureLocation();
++
++    /**
++     * Sets the treasure location that this dolphin will try to lead the player to.
++     * This only has an effect if the dolphin is currently leading a player, as this value is recalculated next time it leads a player.
++     *
++     *  @param location location to guide to
++     */
++    void setTreasureLocation(@org.jetbrains.annotations.NotNull Location location);
++} // Paper end - Dolphin API
```
Machine-Maker: I would make a note in the javadocs what the "world" value for these locations means, for get it will always be the world the entity is in, and for set it has literally 0 effect.

Machine-Maker (approved):

kennytv (approved):
pufferfish

PR #7104, opened 2021-12-13
Add config option for logging player login location

Titaniumtown:
This PR adds a new config option in `paper.yml` (`settings.log-player-login-location`) that toggles coordinates being logged when a player logs onto the server.  This config could be useful to server owners who wish to protect the locations of player logins or in the case where the login location of players don't matter and only waste space in log files (like in the case of lobby servers).

Here's an example of what that would look like in practice in the log files when a player logs in:

Config option being enabled (default):
`username[/xxx.xxx.xxx.xxx:xxxxx] logged in with entity id 14494 at ([world]100, 63.0, -100)`

Config option being disabled:
`username[/xxx.xxx.xxx.xxx:xxxxx] logged in with entity id 14494`

### DIFF
--- a/patches/server/0830-Add-config-option-for-logging-player-login-location.patch
+++ b/patches/server/0830-Add-config-option-for-logging-player-login-location.patch
@@ -1,0 +1,41 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Simon Gardling <titaniumtown@gmail.com>
+Date: Sun, 12 Dec 2021 19:56:35 -0500
+Subject: [PATCH] Add config option for logging player login location
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index cd918cec00d8202252af0d20b1a8891371c538e3..f6dab8134ca835a7593c73a50e06f46ad250f89e 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+@@ -96,6 +96,11 @@ public class PaperConfig {
+         logPlayerIpAddresses = getBoolean("settings.log-player-ip-addresses", logPlayerIpAddresses);
+     }
+ 
++    public static boolean logPlayerLoginLocation = true;
++    private static void playerLoginLocation() {
++        logPlayerLoginLocation = getBoolean("settings.log-player-login-location", logPlayerLoginLocation);
++    }
++
+     public static int maxJoinsPerTick;
+     private static void maxJoinsPerTick() {
+         maxJoinsPerTick = getInt("settings.max-joins-per-tick", 3);
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
+index 042be2cf60a9d01698808d84f2e537a5eb952079..bbdc799a1e89aa5932652896b3bd9f35c34fbc8e 100644
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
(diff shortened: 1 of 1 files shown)


kashike:
Sorry, but... what? This makes zero sense. Anyone who has access to a server console can do a lot more than just see the location of somebody who is logging in.

Titaniumtown:
> Sorry, but... what? This makes zero sense. Anyone who has access to a server console can do a lot more than just see the location of somebody who is logging in.

I was more referring to log files and not the server console itself (because of course someone with access to the server console can view much more than player login location). But it allows server owners to be more transparent with their log files. It may have a very niche use-case, but I thought it would be helpful to someone else as well, and I didn't see the harm in PRing it. Sorry if I didn't explain that well previously.

molor:
> only waste space in log files

In that case maybe you can add option to disable "Disconnecting ..." and/or "... lost connection" messages too? If a player is banned and tries to connect to the server, then both of these messages are shown in log and says essentially the same info. And given that not all players read the reason on the first attempt to connect and, as a rule, try to connect 5 times in a row — just this ability can help to not waste space in logs..

Update — example:
```
...
[15:04:40] Disconnecting com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665): Test
[15:04:40] com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665) lost connection: Test
[15:04:41] Disconnecting com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665): Test
[15:04:41] com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665) lost connection: Test
[15:04:43] Disconnecting com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665): Test
[15:04:43] com.mojang.authlib.GameProfile@ae567c5[...] (/127.0.0.1:53665) lost connection: Test
...
```

Titaniumtown:
Uhhh what?

oskarkk:
I think that generally this kind of thing can be done by log4j configuration. Here you can see the default file that comes with paper (it now may be outdated, I don't know): https://gist.github.com/RoyCurtis/7b5a930221215c58c5184c4b04606bc9
You can see that you could for example filter "player moved wrongly/too quickly" messages:

```
<RegexFilter regex=".* moved wrongly!" onMatch="DENY" onMismatch="NEUTRAL"/>
<RegexFilter regex=".* moved too quickly!.*" onMatch="DENY" onMismatch="NEUTRAL"/>
```

With these lines these messages would stop appearing in logs. I'm not sure, but maybe you could use it to remove just that part with player coords from the message. 

Here's a nice guide to log4j: https://www.reddit.com/r/admincraft/comments/69271l/guide_controlling_console_and_log_output_with/

Titaniumtown:
> I think that generally this kind of thing can be done by log4j configuration. Here you can see the default file that comes with paper (it now may be outdated, I don't know): https://gist.github.com/RoyCurtis/7b5a930221215c58c5184c4b04606bc9 You can see that you could for example filter "player moved wrongly/too quickly" messages:
> 
> ```
> <RegexFilter regex=".* moved wrongly!" onMatch="DENY" onMismatch="NEUTRAL"/>
> <RegexFilter regex=".* moved too quickly!.*" onMatch="DENY" onMismatch="NEUTRAL"/>
> ```
> 
> With these lines these messages would stop appearing in logs. I'm not sure, but maybe you could use it to remove just that part with player coords from the message.
> 
> Here's a nice guide to log4j: https://www.reddit.com/r/admincraft/comments/69271l/guide_controlling_console_and_log_output_with/

So does that completely invalidate this PR? Just checking in if I should close this in favor of using a log4j config.

JRoy:
This is more in the scope of a plugin anyway, so yes I would say

Titaniumtown:
Got it!

PR #7106, opened 2021-12-13
Fully remove spigot's tick limiters

Titaniumtown:
It seems some remnants of spigot's tick limiter (more specifically the config options and initializing them) were still in-place even though the actual usage of these tick limiters when ticking tile entities and entities was previously disabled. This PR simply fully removes that unnecessary code as it does nothing in Paper and the config option could cause some confusion with server owners as it no longer does anything.

### DIFF
--- a/patches/server/0044-Disable-spigot-tick-limiters.patch
+++ b/patches/server/0044-Disable-spigot-tick-limiters.patch
@@ -5,10 +5,36 @@ Subject: [PATCH] Disable spigot tick limiters
 
 
 diff --git a/src/main/java/net/minecraft/world/level/Level.java b/src/main/java/net/minecraft/world/level/Level.java
-index 809f1ac0aca97d484e2f92ebf38a0303499f08ae..db226fe9515e904b8520a063b6dcde62b315a9b1 100644
+index 16f212b4005469dc99fdd83ed6e5810a5b6f8abf..7ec372adae2ea59b71af9c503c12ac1e0786dd47 100644
 --- a/src/main/java/net/minecraft/world/level/Level.java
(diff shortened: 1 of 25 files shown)


Machine-Maker:
Do we need the extra diff of removing the class and the unused fields? I agree if a config setting is unused, like the ones in spigot.yml are not used, they shouldn't be created in new spigot.ymls, but idk about removing everything else.

Titaniumtown:
Well if you remove the config options in `spigot.yml`, then spigot's `TickLimiter` class can't be initialized in `Level`, so you'd comment that out too. Then the only reference of spigot's tick limiter would the the `TickLimiter` class in itself. Just seemed pointless to keep the class file at that point  🤷🏻

Titaniumtown:
> Do we need the extra diff of removing the class and the unused fields? I agree if a config setting is unused, like the ones in spigot.yml are not used, they shouldn't be created in new spigot.ymls, but idk about removing everything else.

@Machine-Maker Should I remove it then? I don't see a need of keeping it.

Titaniumtown:
Rebased. Is there anything I can do on my end to get this merged? I still haven't gotten word from Machine-Maker on if I should not remove the `TickLimiter` class or not. Would like to hear further input as to what I can do!

Machine-Maker (approved):
eh whatever, prob doesnt matter if its removed or not.

Titaniumtown:
rebased.

Owen1212055:
Might just be a nicer diff to deprecate them for removal and hide the config options. This would just make it easy to cleanup when we hard fork instead of needing to comment out fields and such.

Titaniumtown:
not interested in minecraft anymore, forgot about this PR

PR #7107, opened 2021-12-13
[ci skip] Fix typo in getNearbyLivingEntities documentation

darbyjack:


Machine-Maker:
~~There is an existing patch for fixing api javadoc issues, this should be moved into that. 0054-fix-upstream-javadocs~~

Machine-Maker (changes requested):
These are paper-added methods it seems, so should fixed in the patch that added them

(just requesting changed so its marked correctly)

Machine-Maker (approved):

PR #7108, opened 2021-12-13
Save value for setCanMobPickup

### LINKED ISSUE #6929: Item#setCanMobPickup isn't saved between chunk loads

We are using a plugin that adds money as items, and it works fine. These items may exist for months and even years, unlike normal items. On item creating, we call `Item#setCanMobPickup` with `false` on it, for example, to prevent zombies from picking it up. This solution works fine... until chunk was reloaded. After this, any mob can pick it up it and destroy with himself.

I think that `Item#setCanMobPickup` need to be saved persistently inside item entity NBT data, to solve any problems like this one

Currently I'm listening to EntitiesLoadEvent, iterate over entity list and use setCanMobPickup(false) on needed entities again

### Other

_No response_

Machine-Maker:
Closes #6929 

If there is a expectation that this isn't saved, which isn't made clear anywhere in the javadocs, then this might be too big a break with existing behavior. It was always my expectation that methods like this WERE persisted to loads/unloads of the entity, but that might just be me.

### DIFF
--- a/patches/server/0139-Item-canEntityPickup.patch
+++ b/patches/server/0139-Item-canEntityPickup.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Item#canEntityPickup
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/Mob.java b/src/main/java/net/minecraft/world/entity/Mob.java
-index a2310dc7b045d4e4c3924d656a4dcf0a1d66c095..3eda6821d12bf616037f6c00815602428afbcdaf 100644
+index 405ed6034c894bdff889307d9bbf8900f7085a94..bd00b33e4e8e4f8177ad6c8056d2e693effa9965 100644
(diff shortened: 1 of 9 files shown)


kennytv (approved):
Considering this is a patch from 4 years ago, it possibly breaks expected behavior to some people?, buut I'm not fully against changing that here

molor:
Maybe add something like `#setCanMobPickup(boolean value, boolean persistent)`, to not break existing behavior?

Owen1212055:
Honestly yea this is quite old, it may just be better to suggest people use PDC and apply it themselves when the entity is loaded? And the fact this has only been suggested after 4 years perhaps suggests that this behavior isn't needed anyways.

electronicboy (approved):

electronicboy:
Did we ever decide on what to do with this?

PR #7115, opened 2021-12-14
Optimize Util#sequence

IzzelAliz:
There's a huge increase on startup memory footprint after updating to 1.18. Investigation shows that `net.minecraft.Util#sequence` allocates O(n^2) memory on n-size list.

After some rough testing, my optimization can reduce at least 500mb startup RAM usage, as well as a slightly increased chunk loading speed.

Allocation record:
allocate

Original implementation:
orig

Optimization:
opti

### DIFF
--- a/patches/server/0831-Optimize-Util-sequence.patch
+++ b/patches/server/0831-Optimize-Util-sequence.patch
@@ -6,18 +6,10 @@ Subject: [PATCH] Optimize Util#sequence
 Original method allocates O(n^2) memory on n-size list.
 
 diff --git a/src/main/java/net/minecraft/Util.java b/src/main/java/net/minecraft/Util.java
-index f6561599391583ba7d669af42b5716cda0df2d68..d7a167245c0ec72d435f0eef72c7801fe8905e2f 100644
+index f6561599391583ba7d669af42b5716cda0df2d68..15b6e36242fd91c66bc15f7407c22b6d78a24457 100644
 --- a/src/main/java/net/minecraft/Util.java
 +++ b/src/main/java/net/minecraft/Util.java
-@@ -27,6 +27,7 @@ import java.security.PrivilegedActionException;
- import java.security.PrivilegedExceptionAction;
- import java.time.Duration;
- import java.time.Instant;
-+import java.util.ArrayList;
- import java.util.Arrays;
- import java.util.Iterator;
- import java.util.List;
-@@ -391,21 +392,16 @@ public class Util {
+@@ -391,21 +391,16 @@ public class Util {
      }
  
      public static <V> CompletableFuture<List<V>> sequence(List<? extends CompletableFuture<? extends V>> futures) {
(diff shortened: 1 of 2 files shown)


patches/server/0831-Optimize-Util-sequence.patch
```diff
@@ -0,0 +1,50 @@
...
+
+diff --git a/src/main/java/net/minecraft/Util.java b/src/main/java/net/minecraft/Util.java
+index f6561599391583ba7d669af42b5716cda0df2d68..d7a167245c0ec72d435f0eef72c7801fe8905e2f 100644
+--- a/src/main/java/net/minecraft/Util.java
++++ b/src/main/java/net/minecraft/Util.java
+@@ -27,6 +27,7 @@ import java.security.PrivilegedActionException;
+ import java.security.PrivilegedExceptionAction;
+ import java.time.Duration;
+ import java.time.Instant;
++import java.util.ArrayList;
```
lynxplay: This should most likely be a fully qualified name e.g.
```java
final var list = new java.util.ArrayList<V>(future.size());
```

patches/server/0831-Optimize-Util-sequence.patch
```diff
@@ -0,0 +1,42 @@
...
+-        }, (completableFuture, completableFuture2) -> {
+-            return completableFuture.thenCombine(completableFuture2, (list, list2) -> {
+-                List<V> list3 = Lists.newArrayListWithCapacity(list.size() + list2.size());
+-                list3.addAll(list);
+-                list3.addAll(list2);
+-                return list3;
++        // Paper start - optimize
++        return CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
++            .thenApply(v -> {
++                var list = Lists.<V>newArrayListWithCapacity(futures.size());
```
kennytv: Var makes this more awkward to read, just do `List<V> list = new ArrayList<>(futures.size());`

patches/server/0831-Optimize-Util-sequence.patch
```diff
@@ -0,0 +1,42 @@
...
+-                return list2;
+-            });
+-        }, (completableFuture, completableFuture2) -> {
+-            return completableFuture.thenCombine(completableFuture2, (list, list2) -> {
+-                List<V> list3 = Lists.newArrayListWithCapacity(list.size() + list2.size());
+-                list3.addAll(list);
+-                list3.addAll(list2);
+-                return list3;
++        // Paper start - optimize
++        return CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
```
kennytv: Put `new CompletableFuture[0]` into a private static final field to reuse

kennytv (changes requested):

kennytv (approved):

kennytv:
Well looks like basically the same code will make it into Vanilla 1.18.2

PR #7127, opened 2021-12-15
Use team display name for quit message

### LINKED ISSUE #7123: Team suffixes and prefixes are not displayed in player quit messages
### Expected behavior

Using the `/team` command (vanilla Minecraft command) you can create teams and set an individual prefix and suffix for each team. Prefixes and suffixes shows up in the tab list, in player names above their heads and in chat. In a vanilla server this also affect the join and quit messages. For example, a player `Max` with team prefix `[Admin] ` and team suffix ` (Lvl 19)` would show up as `[Admin] Max (Lvl 19)`.

### Observed/Actual behavior

The prefix/suffix show up everywhere expected, _but not_ in the quit message. 

### Steps/models to reproduce

- Create a new Paper server
- Log into the server as an operator and run the following commands
  - `/team add test`
  - `/team modify test suffix "TestSuffix"`
  - `/team join test`
- Then leave and re-enter the server. The suffix shows up on the join message, but not on the quit message. This can be seen in the console or using a second account.

### Plugin and Datapack List

No plugins and no datapacks except the default ones.

### Paper version

This server is running Paper version git-Paper-76 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 95d881f)

### Other

I set both `use-display-name-in-quit-message` and `use-vanilla-world-scoreboard-name-coloring` to `true` in `paper.yml` and I also searched the other configuration files for options which might solve this problem. Neither my internet research nor asking on Discord could provide me with a solution or a hint onto what might be going on.

Machine-Maker:
Fixes #7123

This almost certainly goes in an earlier patch, maybe Adventure, not sure.

But its my understanding that the game profile name has been used in quit msgs for ever, hence the paper.yml setting to use display names, but that is only the "bukkit" display name, not the team display name formatted with team prefixes, colors, etc. So I think if we want to preserve the original behavior, the `use-display-name-in-quit-message` setting should be changed from a boolean to a multi-option setting, with options, NAME, BUKKIT_DISPLAY_NAME, and TEAM_DISPLAY_NAME or something like that.

### DIFF
--- a/patches/server/0009-Adventure.patch
+++ b/patches/server/0009-Adventure.patch
@@ -2349,7 +2349,7 @@ index be097f13dba5d408f58d6fada893bed2638d4219..3d7d1ba148dbc3591d8c76b99a2ee7d9
  
                  @Override
 diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
-index c9497bead11bf2b3b859f1c91c9692ce6007e5ed..5481a029a4a710eb20a4689633d5a4ec509d49d8 100644
+index c9497bead11bf2b3b859f1c91c9692ce6007e5ed..046560b409c73c73f2c8e645625697bde3e564c2 100644
 --- a/src/main/java/net/minecraft/server/players/PlayerList.java
 +++ b/src/main/java/net/minecraft/server/players/PlayerList.java
 @@ -8,6 +8,7 @@ import com.mojang.logging.LogUtils;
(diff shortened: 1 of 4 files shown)


eliaspr:
Why has this pull request been silently resting for 16 days? I'm pretty sure you have a good reason, so I'd appreaciate If you could tell me. Thanks!

JRoy:
> Why has this pull request been silently resting for 16 days? I'm pretty sure you have a good reason, so I'd appreaciate If you could tell me. Thanks!

I'm sorry we will not be able to provide you a good reason at this time. Please try again later.

electronicboy:
We sincerely apologize that in the rundown towards the holidays where people are working on stuff last minute or spending time with family, etc, that we aren't slaving away for free to work on stuff which is oh so important to you.

eliaspr:
I sincerely respect all the effort put in by the Paper community members. I was just wondering if there is a technical issue or concern associated with this change which prevented the PR from being approved. Also, the PR was created just a few days after I posted the original issue so I thought it was going to be merged right away.

electronicboy (approved):

Owen1212055 (approved):
I agree that this should just go in the adventure patch, but lgtm.

PR #7129, opened 2021-12-16
Configurable max block light for monster spawning

kennytv:
So people can revert this to how it was before 1.18
https://github.com/PaperMC/PaperDocs/pull/138

Not using Math.max so it's obvious 0 is the Vanilla default when looking at code, used when the config is set to -1


lynxplay (commented):
Shouldn't this go into the `game-mechanics` section of the world config ? Would fit pretty well there imo

kennytv:
Uh not really, it'd be the only mob spawning option in there then

lynxplay:
> Uh not really, it'd be the only mob spawning option in there then

Pillager spawning is also in there isn't it ? But I digress, would fit just as well into the enormous amount of other random options that modify game mechanics but are not in the game mechanics section.

Machine-Maker (approved):

jpenilla (approved):
ok

PR #7134, opened 2021-12-17
Fix breakNaturally for fluid-logged blocks

Machine-Maker:
When a player breaks a fluid-logged block, the fluid remains. This is not currently the case for breakNaturally. I think this isn't a breaking change cause the method is **supposed** to simulate to the best it can, a block being broken by a player.


kennytv (approved):

Owen1212055 (approved):
Can confirm works just fine when tested

PR #7141, opened 2021-12-18
Add material tags for copper blocks

Machine-Maker:


jacobsandersen:
After this is reviewed I'll make a new PR to add tags for the other metal blocks (iron, gold, diamond, perhaps others that I didn't notice on first glance) that I didn't see

kennytv (approved):

PR #7162, opened 2021-12-20
Fix cancelling PlayerInteractEntityEvent on bucketable entities

Machine-Maker:
One possible issue with this, is the client seems to send multiple interaction packets which cause the AddMob and EntityMetadata packets to be sent twice. I didn't see any issues on the client, so I'm not sure if that's a problem or not.

Basically, the root issue is that the PlayerInteractEntityEvent can be cancelled but the client will still remove the entity. The 56parts of the if statement are meant to be all the checks needed to see the client will think the entity should be removed.

This certainly seems very much of a *meh* solution, but I'm not sure how else to do it.


Owen1212055:
I sorta already covered this in: https://github.com/PaperMC/Paper/pull/7041

However, it seems that you also extended the checking... so I am not sure.

Owen1212055:
Oh yea, no looking at this again it should be noted that a condition for this already exists and was added upstream. The issue is that it doesn't cover axlotls currently, so I am not sure this is needed.
image

Machine-Maker:
Yeah, probably.

PR #4924, opened 2020-12-21
Add per-player block reach

Camotoy:
This commit implements the outline I presented in #4908. Most importantly, I have confirmed that all vanilla values remain correct without modification from the API, and when resetting block range back to their vanilla values. Please do let me know if there's something I can improve (especially documentation).

I'm also willing to release these patches under MIT, if there is a process for that. 

Fixes #4908

PR #7163, opened 2021-12-20
Add ability to modify range checks on players

### LINKED ISSUE #4908: Allow customizable block placing and destroying range
(I want to preface this by saying that I am no anticheat expert, nor have I extensively worked with Minecraft internals. I am quite familiar with packets at the Java code level.)

I'm one of the developers at Geyser (a Bedrock-client-to-Java-server proxy). While a lot of things translate over nicely from Bedrock to Java (and vice versa) without server interference, one of the difficulties between the platforms is the block range differences. Whereas Java range is capped at 6/7 blocks (depending on gamemode), mobile Bedrock players can go as far as 12 blocks in creative mode. 

I propose adding new methods to the API to allow for changing a specific player's block reach. I've loosely tested this functionality (by replacing the hardcoded numbers), and I am proposing a feature request before a PR to make sure this is something fit for Paper (I would PR this myself). There are two individual but related checks: block placing and block destroying. 

The general structure I propose involves two new methods added with getters/setters: `get/setCreativeBlockReach` and `get/setSurvivalBlockReach` that ultimately control player reach. (Alternatively: two getters and one setter?: `setBlockReach(int survivalReach, int creativeReach)`). This should probably have a bounds check of the vanilla survival/creative reach numbers; otherwise, with lower numbers, ghost blocks will occur on vanilla clients. 

For block placing (in `PlayerConnection`, under the listener for "PacketPlayInUseItem"), there are two checks put in place: a vanilla reach check that simply ensures a block is within a distance squared of 8 * 8, and a check added by CraftBukkit that measures eye location and a more strict reach check of either 7 (creative) or 6 (survival). `SURVIVAL_PLACE_DISTANCE_SQUARED` and `CREATIVE_PLACE_DISTANCE_SQUARED` (two static constants that control this check) can be replaced with the player's reach getters. The vanilla distance check - currently hardcoded to 64D - can probably be something like `player.getMaxReachDistance` (the greater range number between the two reaches) + 1. 

For block destroying (in `PlayerInteractManager#a(BlockPosition, PacketPlayInBlockDig.EnumPlayerDigType, EnumDirection, int`), there is a hardcoded reach check of 36D (6 * 6). This can be set to the previously described `player.getMaxReachDistance` minus 1 (Since we aren't placing a block but interacting with an existing block).

Alternatively, with direct server access, Geyser doesn't need to send the block place packet, and can just manually place the block on the server and call the player interact event ourselves. This has a couple of downsides:

- From what I can tell, it requires NMS, making it not immediately forwards-compatible and more difficult to implement for future versions while maintaining backwards compatibility.
- Knowing plugins, attempting to replace the entire block placing/breaking system will likely break something obscure. While such a feature request described here would certainly break certain anticheats, having a getter for block reach would enable other plugins to verify block placement.

Camotoy:
Fixes #4908; supersedes #4924.

The approach I've taken this time around is finding any "magic number" used in range checking and exposing the value, allowing plugins to modify each per-player. Feedback is appreciated; the "max block reach" here isn't necessarily good API choice but working around existing design (and honestly the entire API could be shoved as unsafe and I'd be OK with it).

### DIFF
--- a/patches/api/0353-Add-ability-to-modify-range-checks-on-players.patch
+++ b/patches/api/0353-Add-ability-to-modify-range-checks-on-players.patch
@@ -1,0 +1,110 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Camotoy <20743703+Camotoy@users.noreply.github.com>
+Date: Mon, 20 Dec 2021 15:21:18 -0500
+Subject: [PATCH] Add ability to modify range checks on players
+
+
+diff --git a/src/main/java/io/papermc/paper/world/PlayerReachSettings.java b/src/main/java/io/papermc/paper/world/PlayerReachSettings.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..69dfb75c4806ef434fecd482bebaef34d6807a89
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/world/PlayerReachSettings.java
+@@ -0,0 +1,65 @@
++package io.papermc.paper.world;
++
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.Range;
++
++public interface PlayerReachSettings {
++    /**
++     * Used when placing or interacting with blocks in creative mode. Cannot be lower than 49.
++     */
++    @Contract(pure = true)
++    @Range(from = 49, to = Integer.MAX_VALUE)
++    int creativeBlockReachSquared();
++
++    /**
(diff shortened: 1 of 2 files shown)


Camotoy:
No.

Camotoy:
Needs a rebase.

Camotoy:
OK, fine, I'll close it. I want to come back to this though.

nitrog0d:
Don't close it tf this is kinda good

Astralchroma:
> Don't close it tf this is kinda good

It's 7 months behind main, there is no way it could be merged in its current state.

Issue #7062, opened 2021-12-08
Paper 1.18 Experimental Crash due to net.minecraft.world.level.block.CaveVinesBlock.getGrowIntoState(CaveVinesBlock.java:56)


ilektron:
### Stack trace

```
---- Minecraft Crash Report ----
// Why did you do that?

Time: 12/7/21, 11:48 PM
Description: Exception ticking world

java.lang.NoSuchMethodError: 'float java.util.Random.nextFloat(float)'
	at net.minecraft.world.level.block.CaveVinesBlock.getGrowIntoState(CaveVinesBlock.java:56)
	at net.minecraft.world.level.block.GrowingPlantHeadBlock.randomTick(GrowingPlantHeadBlock.java:61)
	at net.minecraft.world.level.block.state.BlockBehaviour$BlockStateBase.randomTick(BlockBehaviour.java:962)
	at net.minecraft.server.level.ServerLevel.tickChunk(ServerLevel.java:831)
	at net.minecraft.server.level.ServerChunkCache.tickChunks(ServerChunkCache.java:1049)
	at net.minecraft.server.level.ServerChunkCache.tick(ServerChunkCache.java:923)
	at net.minecraft.server.level.ServerLevel.tick(ServerLevel.java:622)
	at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1607)
	at net.minecraft.server.dedicated.DedicatedServer.tickChildren(DedicatedServer.java:480)
	at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1470)
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1264)
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:317)
	at java.base/java.lang.Thread.run(Thread.java:831)


A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Thread: Server thread
Stacktrace:
	at net.minecraft.world.level.block.CaveVinesBlock.getGrowIntoState(CaveVinesBlock.java:56)
	at net.minecraft.world.level.block.GrowingPlantHeadBlock.randomTick(GrowingPlantHeadBlock.java:61)
	at net.minecraft.world.level.block.state.BlockBehaviour$BlockStateBase.randomTick(BlockBehaviour.java:962)
	at net.minecraft.server.level.ServerLevel.tickChunk(ServerLevel.java:831)
[...]

PR #7165, opened 2021-12-21
Remove invalid reference to Random.nextFloat(float) from CaveVinesBlock

oasis9:
This was crashing my server. I'm not sure if it's correct to have two renamed files included with the patch file in my commit but this fixes the issue for me. It was caused by net.minecraft.world.level.block.CaveVinesBlock in a section modified by Paperspigot patch "Fix kelp modifier changing growth for other crops".


jpenilla:
There is nothing invalid here, duplicate of https://github.com/PaperMC/Paper/issues/7062, your java installation is outdated.

PR #7167, opened 2021-12-21
Fix remapping issue with RangedAttackMob and RangedEntity

Machine-Maker:
This is one possible fix, another would be to change the generic in CraftRangedEntity from `T extends RangedAttackMob` to `T extends Mob & RangedAttackMob`, which would remove the need to add a method in RangedAttackMob. I think that would break any plugin that maybe did smth odd with CraftRangedEntity.


electronicboy:
Not fond of maintaining manual mappings if we can avoid it, ofc
I kinda err on changing the generic param, I don't think this would induce any ABI breaks to my knowledge, as for source, can't really think of much there either

Machine-Maker:
Ok, redid this to change the generic.

electronicboy (approved):

jpenilla (approved):

PR #7170, opened 2021-12-21
Add api for getting direct entity cause of knockback

### LINKED ISSUE #7168: EntityKnockbackByEntityEvent called with wrong entities
### Expected behavior

The EntityKnockbackByEntityEvent#hitBy() entity should be called with the entity that did the knockback

### Observed/Actual behavior

Instead, if the entity that did the knockback is a projectile or a TNT, the #hitBy entity is a player (shooter of the projectile or the one that lit the TNT) instead of the entity that actually did the knockback

### Steps/models to reproduce

Example 1:
1. Throw a snowball
2. EntityKnockbackByEntityEvent will be called with #hitBy set to the shooter of the snowball instead of the actual snowball entity

Example 2:
1. Get a TNT and flint and steel
2. Prime the TNT
3. EntityKnockbackByEntityEvent will be called with #hitBy set to the player that primed the TNT instead of the TNT entity

Example 3:
1. /summon tnt
2. EntityKnockbackByEntityEvent will not be called because there was no entity that primed the TNT

### Plugin and Datapack List

Only a plugin that was printing the #hitBy entity into the console. No datapacks.

### Paper version

This server is running Paper version git-Paper-85 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: df8d28a)

### Other

_No response_

Machine-Maker:
Closes https://github.com/PaperMC/Paper/issues/7168

### DIFF
--- a/patches/api/0114-Add-EntityKnockbackByEntityEvent.patch
+++ b/patches/api/0114-Add-EntityKnockbackByEntityEvent.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add EntityKnockbackByEntityEvent
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/entity/EntityKnockbackByEntityEvent.java b/src/main/java/com/destroystokyo/paper/event/entity/EntityKnockbackByEntityEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..9efecabab813f575bb447a356e5e7e952d110f30
+index 0000000000000000000000000000000000000000..cff7f42e84798be2789c7ebb8f2b60b9dd40a711
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/entity/EntityKnockbackByEntityEvent.java
-@@ -0,0 +1,82 @@
+@@ -0,0 +1,106 @@
 +package com.destroystokyo.paper.event.entity;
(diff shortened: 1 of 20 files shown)


vytskalt:
When is this and #7171 getting merged?

Machine-Maker:
I'm not entierly happy that `EntityKnockbackByEntityEvent#getDirectHitBy()` is nullable when it'll only be null if the legacy ctor is used. What if that return value was wrapped in a requireNonNull call? Cause it really should be notnull.

vytskalt:
> I'm not entierly happy that `EntityKnockbackByEntityEvent#getDirectHitBy()` is nullable when it'll only be null if the legacy ctor is used. What if that return value was wrapped in a requireNonNull call? Cause it really should be notnull.

Yes, I think its better to do that instead of keeping it nullable

vytskalt:
Still need this.

vytskalt:
no

Lulu13022002:
Is it worth to wait for https://github.com/PaperMC/Paper/pull/8058 and pass directly the damage source here and also for the PushEvent (another PR) ?

lynxplay:
I think we may be better of deprecating the existing method and choosing some *more descriptive* names for this 👍

PR #7171, opened 2021-12-21
Fire entity knockback event for ownerless tnt

### LINKED ISSUE #7169: EntityKnockbackByEntityEvent not called if there's no source entity
### Expected behavior

The event should be called even if there is no source entity for a TNT or a projectile.

### Observed/Actual behavior

Its not called if there's no source entity.

### Steps/models to reproduce

1. /summon tnt
2. The event wasn't called

However, if player primes the TNT with a flint and steel, then it is called.

### Plugin and Datapack List

Only test plugin which debugs the event.

### Paper version

This server is running Paper version git-Paper-85 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: df8d28a)

### Other

Related: #7168

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7169

### DIFF
--- a/patches/server/0202-Add-EntityKnockbackByEntityEvent-and-EntityPushedByE.patch
+++ b/patches/server/0202-Add-EntityKnockbackByEntityEvent-and-EntityPushedByE.patch
@@ -37,7 +37,7 @@ index 6fbde57320a58600f8c4b9ce598fa93bd2772e8b..cf907e0c1a89639639b6453032f8f6fe
  
      protected void markHurt() {
 diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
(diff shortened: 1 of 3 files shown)


electronicboy (approved):

Machine-Maker:
Rebased for 1.18.2

Warriorrrr (approved):

Machine-Maker:
Rebased for 1.20.1

YHDiamond:
Any idea when this will be merged? I'm relying on it for a project so if there is anything I could do to make it go faster please let me know.

vLuckyyy:
> Any idea when this will be merged? I'm relying on it for a project so if there is anything I could do to make it go faster please let me know.

I think it would be great if we could get this merged. I have same problem with.

lynxplay (approved):

PR #7175, opened 2021-12-22
Load effect amplifiers greater than 127 correctly

Machine-Maker:
effect amplifier is stored as a signed byte -128 to 127, but then read into an int so an effect amplifier of 128 (stored as -128) is read as int -128 which is reset back to 0.

Amplifier 0-127 stored as 0-127 (byte) and 127-255 stored as -128 to -1


electronicboy (approved):

jaylawl:
Kudos @Machine-Maker .... you really are working like a machine with the amount of PRs!

kashike:
If this is an issue in Minecraft itself, please open a [Mojira](https://bugs.mojang.com/) issue.

NoahvdAa:
[MC-118857](https://bugs.mojang.com/browse/MC-118857)

MiniDigger:
ideally we add that mojira to the patch description and/or the comment

Machine-Maker:
> ideally we add that mojira to the patch description and/or the comment

Added to commit descriptions

PR #7181, opened 2021-12-24
Expose isFuel and canSmelt methods to FurnaceInventory

BillyGalbreath:
I'm adding this to solve https://github.com/EngineHub/CraftBook/issues/221#issuecomment-597388101 for CraftBook, but these methods can be valuable to many other plugins.

Example use-case is `canSmelt` would replace all [this](https://github.com/EngineHub/CraftBook/blob/f3c9a57d3fca4054c166fb0104a8a671bfca1c8b/src/main/java/com/sk89q/craftbook/util/ItemUtil.java#L462-L636), and `isFuel` would replace all [this](https://github.com/EngineHub/CraftBook/blob/f3c9a57d3fca4054c166fb0104a8a671bfca1c8b/src/main/java/com/sk89q/craftbook/util/ItemUtil.java#L692-L763) and would add the ability for recipes added via API and data packs to work correctly.

CC: @me4502 ;)

### DIFF
--- a/patches/api/0351-Expose-isFuel-and-canSmelt-methods-to-FurnaceInvento.patch
+++ b/patches/api/0351-Expose-isFuel-and-canSmelt-methods-to-FurnaceInvento.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Expose isFuel and canSmelt methods to FurnaceInventory
 
 
 diff --git a/src/main/java/org/bukkit/inventory/FurnaceInventory.java b/src/main/java/org/bukkit/inventory/FurnaceInventory.java
-index 3f46259c1e0f82941ffc3038d1b51be199114abd..f469f1718c8649f9e5ff2444fbbb9c6d8eae62be 100644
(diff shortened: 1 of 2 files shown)


MiniDigger:
I don't fully remember the details of the itemstack api, but passing `new ItemStack(Material.WHATEVER)` would throw a class cast exception here, right? so need to handle those cases (didn't we have util methods for shit like this around somewhere?)

BillyGalbreath:
> I don't fully remember the details of the itemstack api, but passing `new ItemStack(Material.WHATEVER)` would throw a class cast exception here, right? so need to handle those cases (didn't we have util methods for shit like this around somewhere?)

Oh wow, I didn't even test for that case. Updated to use `CraftItemStack#asNMSCopy` to prevent CCEs as mentioned.

MiniDigger (approved):

kashike (approved):

jpenilla (approved):

PR #7183, opened 2021-12-24
Fix desync on teleporting entity on first tick

### LINKED ISSUE #5939: ArmorStands teleporting at an offset to their teleportation location
### Expected behavior

Armorstands to teleport to a location in relation to the player

### Observed/Actual behavior

image

Armorstands appear to be at an offset location to the location they are being teleported to. In the case of this image the armorstand (circled in red) is meant to be under the name tag armorstand (circled in blue). The name tag armorstand is teleporting to the main armorstand meaning it should always be above it (meaning this could potentially just be a visual issue that all players can see). This was all functioning perfectly in previous versions but in 1.17 has started causing this issue. (NOTE: The error does not always occur and seems to be relatively unpredictable as to when it will start being out of line with the  player.)

### Steps/models to reproduce

The two plugins which I have seen effected by this are TAB and ESFollowers (a private plugin of mine) both which teleport armorstands in relation to the player. TAB which uses armorstand packets above the player's head to allow for multiple lines of information under their name. There are never any errors produced by this and it seems to be relatively random as to when this will occur so I don't quite know what else to offer to help reproduce this.

### Plugin list

ActivityRewarder, AncientGates, ArcherPraise, ArmorStandTools, BetterRTP, BossShopPro, Citizens, CitizensHologram*, ClientFPS, DeluxeChat, DeluxeTags, EnchantedBosses, EpicRename, ESFollowers, ESMusicRegions, ESParticles, Essentials, EssentialsSpawn, ESStaffChat-Satellite, ESTrading, ESWelcome, GSit, HolographicDisplays, IllegalStack, LootArrows, LuckPerms, Multiverse-Core, MVdWPlaceholderAPI*, NoteBlockAPI, Notes, PlaceholderAPI, PlayerPoints, PlayerVaults, ProtocolLib, ReportBook, SkillsLibrary, SkyblockNPCs, StructureVoidViewer, Superheroes2, SuperheroesPlusUltra, SuperheroesWorldGuard, TAB, Towny, TownySkillsLibraryAddon, Vault, VoidSpawn, VoidWorld, VotingPlugin, WorldEdit, WorldGuard, WorldGuard-Towny*, WorldGuardEvents

### Paper version

[21:30:40 INFO]: This server is running Paper version git-Paper-"22280c5" (MC: 1.17) (Implementing API version 1.17-R0.1-SNAPSHOT)
Previous version: git-Paper-"33d6b6f" (MC: 1.17)

### Agreements

### Other

_No response_

Owen1212055:
If the entity's location was changed between them being spawned and them being ticked for the 2nd time, it could cause a desync to occur.  This is because the entity's location would change on the server but would not be sent to the client since it was its first time ticking (tick counter is updated lower).

If the entity was moved within a 8x8x8 block radius it would send the location relatively, which since it was desynced would cause issues.

Reproduction Code:
```java
        Creeper creeper = player.getWorld().spawn(player.getLocation(), Creeper.class);
        creeper.setAI(false);
        creeper.setInvulnerable(true);

        new BukkitRunnable(){

            @Override
            public void run() {
                Location location = player.getEyeLocation();
                location.add(5, 0, 3);

                player.spawnParticle(Particle.BLOCK_MARKER, location, 1, Bukkit.createBlockData(Material.BAMBOO));
                creeper.teleport(location);
            }
        }.runTaskTimer(this, 1, 1);
```
Demonstration:
https://user-images.githubusercontent.com/23108066/147318951-ba26be60-af5f-4e0a-9fa8-fa91005e3b29.mp4
- In this example, I used the teleport command to force the creeper's location to be resent exactly.

Fix:
https://user-images.githubusercontent.com/23108066/147318803-ac837eb3-c5f1-448b-9b1f-ae1b6c8283bd.mp4

This bug affects latest paper + spigot.

Fixes: https://github.com/PaperMC/Paper/issues/5939


kashike:
Does this affect vanilla? If so, please find (or open) the associated Mojira issue and include it in the patch.

Owen1212055:
> Does this affect vanilla? If so, please find (or open) the associated Mojira issue and include it in the patch.

I only see two issues, one is resolved and one says only players are affected by this issue.
https://bugs.mojang.com/browse/MC-228577
https://bugs.mojang.com/browse/MC-90686

Should I open an issue, or update the open one saying this applies to all entities?

electronicboy (approved):

Machine-Maker (approved):
I can't see an immediate issues with that, but I think that the condition should just be `if (true ||...)` since `tickCount` is always going to be non-negative. But I do wonder why that check was there in the first place, so that's why I'm unsure if there's some side-effect that I just don't see.

Owen1212055:
Done and rebased!

Owen1212055:
This fix caused an issue with paintings to be discovered, where paintings don't handle relative movement correctly because they don't call the #setPacketCoordinates method when being spawned. This would cause them to be teleported to a relative location to the origin of the world.

Fixed that!

Alvinn8 (approved):
I tested this and it fixed issues I was having with a plugin of mine. Did very little other testing so there might still be side effects, because as mentioned here before it's kinda odd that the check exists in the first place.

PR #7187, opened 2021-12-25
[WIP] Poi API

Owen1212055:
Allows methods for accessing poi blocks
An example is provided in the test plugin, where all nearby POI will be marked with a redstone block on top of it. 

Documentation is not yet finished, **feedback is welcome**!!

### DIFF
--- a/patches/api/0351-Poi-API.patch
+++ b/patches/api/0351-Poi-API.patch
@@ -1,0 +1,190 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Fri, 24 Dec 2021 23:24:03 -0500
+Subject: [PATCH] Poi API
+
+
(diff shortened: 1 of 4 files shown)


Owen1212055:
None of this is really exposed in the current API, feedback here would be appreciated.

Owen1212055:
I'll finish this up and just let it sit here for any additional feedback.

Owen1212055:
K

Owen1212055:
Darn phone decided to scroll right to the close button.

patches/api/0351-Poi-API.patch
```diff
@@ -0,0 +1,190 @@
...
++    }
++
++    @NotNull
++    Collection<RegisteredPoi> getPoisInChunk(int x, int z);
++
++    @NotNull
++    Collection<RegisteredPoi> getPoisInRange(@NotNull Location location, int range);
++
++    @Nullable
++    RegisteredPoi getClosestPoi(@NotNull Location location, int range, @NotNull Predicate<PoiType> poiTypePredicate, @NotNull Predicate<Location> locationPredicate);
```
Lulu13022002, line 54: Maybe create the mirror function (locateNearestPOI) in bukkit world since there already locateNearestStructure and locateNearestBiome. ~Also why not the name PoiBlock instead of RegisteredPoi ?~

PR #7188, opened 2021-12-25
Dont register multiple ASK_SERVER suggestions under one parent node

### LINKED ISSUE #7185: /tp and /teleport Tab autocomplete don't work for the first argument
### Expected behavior

typing /tp or /teleport and pressing Tab should give you a list of possible arguments

### Observed/Actual behavior

nothing appears, even after typing the first letter from an online username and spamming Tab it doesn't respond. But only on the first command argument. The second argument works perfectly.

### Steps/models to reproduce

I investigated and this issue started appearing in paper builds 93-101. Just type /tp and you will see

### Plugin and Datapack List

tried with a clean brand new server with no plugins

### Paper version

tried in versions 90-101, 90-92 worked, broke in 93 and still hasn't been fixed

### Other

from what I've seen so far commands other than /tp and /teleport are not affected, needs more triage though

### LINKED ISSUE #7194: UUID's no longer autofill with certain vanilla commands following Paper 1.18.1 Patch #93
### Expected behavior

When targeting an entity, their UUID should normally autofill when trying to autocomplete certain commands such as /execute, /data, and /teleport. Following Patch #93, the UUID's would no longer autofill as they should, even when targeting an entity.

### Observed/Actual behavior

Following Patch #93, entity UUIDs would no longer autofill as they should, even when targeting an entity.
image
image

### Steps/models to reproduce

Simply run any paper server Version #93 and up, target an entity such as an armor stand, and attempt to fill out a command that targets an entity such as /teleport or /execute as. Their UUID will not show up in the autofill list when it normally should.

### Plugin and Datapack List

This was tested and confirmed to occur on a Paper server with no active plugins or datapacks.

### Paper version

Paper Version that works normally:
This server is running Paper version git-Paper-92 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 264b11d)

Paper Versions with mentioned issue:
This server is running Paper version git-Paper-93 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 2cf6a57)
This server is running Paper version git-Paper-101 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 464b171)\nYou are running the latest version

Also occurs in all versions in between

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7185 and
fixes #7194

The client doesn't handle multiple sibling nodes asking the server for suggestions correctly, so make sure to only register it once per parent.

### DIFF
--- a/patches/server/0834-Fix-entity-type-tags-suggestions-in-selectors.patch
+++ b/patches/server/0834-Fix-entity-type-tags-suggestions-in-selectors.patch
@@ -25,21 +25,55 @@ index cd918cec00d8202252af0d20b1a8891371c538e3..6d02910a903f2c6352202c49149172e3
      public static final class PacketLimit {
          public final double packetLimitInterval;
          public final double maxPacketRate;
+diff --git a/src/main/java/net/minecraft/commands/CommandSourceStack.java b/src/main/java/net/minecraft/commands/CommandSourceStack.java
+index cb0045fc4ddd738c45dee89d57b213a633b9a136..098182d2426a25cef0bc285356bc346db0af8172 100644
+--- a/src/main/java/net/minecraft/commands/CommandSourceStack.java
++++ b/src/main/java/net/minecraft/commands/CommandSourceStack.java
+@@ -322,4 +322,20 @@ public class CommandSourceStack implements SharedSuggestionProvider, com.destroy
+         return this.source.getBukkitSender(this);
+     }
+     // CraftBukkit end
++    // Paper start - override getSelectedEntities
++    @Override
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Also fixes #7194 now. But this maybe should be turned off by default now. I don't know how often a client on a server asks for suggestions for the entity selector, but with it now doing a targeted entity lookup...

electronicboy (approved):

jpenilla (approved):

PR #7195, opened 2021-12-26
Expose saturation in FoodLevelChangeEvent

### LINKED ISSUE #7193: Add saturationModifier to FoodLevelChangeEvent

I'm experimenting with custom food items and trying to implement new food types. When listening to the FoodLevelChangeEvent I'm able to modify the resulting food level, but there is no simple way to modify the amount of saturation the player receives.

Adding these two methods to the FoodLevelChangeEvent:

- FoodLevelChangeEvent#getSaturationModifier()
- FoodLevelChangeEvent#setSaturationModifier(float saturationModifier)

Listening to the event and doing the saturation calculation on my own, then setting the resulting value after the event has been processed.

### Other

_No response_

Owen1212055:
Resolves https://github.com/PaperMC/Paper/issues/7193

It may just be better to add an entirely new event.. but I would like opinions on this. Although this mostly works fine, it gets a little wonky when it comes to canceling natural hunger decay. 
```
((ServerPlayer) this.entityhuman).getBukkitEntity().sendHealthUpdate();
@@ -66,10 +66,11 @@ public class FoodData {
                 this.saturationLevel = Math.max(this.saturationLevel - 1.0F, 0.0F);
             } else if (enumdifficulty != Difficulty.PEACEFUL) {
                 // CraftBukkit start
-                org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, Math.max(this.foodLevel - 1, 0));
+                org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, Math.max(this.foodLevel - 1, 0), this.saturationLevel); // Paper - Expose Saturation

                 if (!event.isCancelled()) {
                     this.foodLevel = event.getFoodLevel();
+                    this.saturationLevel = event.getSaturationModifier(); // Paper - Expose Saturation
                 }
```

Specifically here, I can't really "cancel" saturation modification since this would be a behavior change. 
So, it's either we can live with this slight issue, or we can just make an entirely new event that would be executed alongside FoodLevelChangeEvent, this would allow us to add an event for natural saturation decay as well. However, canceling the FoodLevelChangeEvent will also cancel modification of the saturation in some situations (like eating food). So I would like some pointers here.

### DIFF
--- a/patches/api/0351-FoodLevelChangeEvent-Expose-Saturation.patch
+++ b/patches/api/0351-FoodLevelChangeEvent-Expose-Saturation.patch
@@ -1,0 +1,67 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sat, 25 Dec 2021 23:04:06 -0500
+Subject: [PATCH] FoodLevelChangeEvent Expose Saturation
+
+
+diff --git a/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java b/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
+index 7f4c5ea8607b08f8676528cc71b0b312575809ed..b1f5ed6656d9d80a3dcca2a5bacf9fc8079ab321 100644
+--- a/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
++++ b/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
+@@ -15,16 +15,27 @@ public class FoodLevelChangeEvent extends EntityEvent implements Cancellable {
+     private boolean cancel = false;
+     private int level;
+     private final ItemStack item;
++    private float saturationModifier; // Paper - Expose Saturation
+ 
+     public FoodLevelChangeEvent(@NotNull final HumanEntity what, final int level) {
+         this(what, level, null);
+     }
+ 
++    @Deprecated // Paper - Expose Saturation
+     public FoodLevelChangeEvent(@NotNull final HumanEntity what, final int level, @Nullable final ItemStack item) {
+         super(what);
+         this.level = level;
+         this.item = item;
++        this.saturationModifier = what.getSaturation(); // Just incase, make sure the player's saturation doesn't change. // Paper - Expose Saturation
+     }
++    // Paper Start - Expose Saturation
++    public FoodLevelChangeEvent(@NotNull final HumanEntity what, final int level, @Nullable final ItemStack item, final float saturationModifier) {
++        super(what);
++        this.level = level;
++        this.item = item;
++        this.saturationModifier = saturationModifier;
++    }
++    // Paper End - Expose Saturation
+ 
+     @NotNull
+     @Override
+@@ -67,6 +78,28 @@ public class FoodLevelChangeEvent extends EntityEvent implements Cancellable {
+         this.level = level;
+     }
+ 
++    // Paper Start - Expose Saturation
++    /**
++     * Gets the saturation modifier that the entity
++     * involved in this event should be set to.
++     *
++     * @return saturation modifier
++     */
++    public float getSaturationModifier() {
++        return saturationModifier;
++    }
++
++    /**
++     * Sets the satuartion modifier that the entity
++     * involved in this event will be set to.
++     *
++     * @param saturationModifier new saturation modifier
++     */
++    public void setSaturationModifier(float saturationModifier) {
++        this.saturationModifier = saturationModifier;
++    }
++    // Paper End - Expose Saturation
++
+     @Override
+     public boolean isCancelled() {
+         return cancel;
--- a/patches/api/0351-FoodLevelChangeEvent-Expose-Saturation.patch
+++ b/patches/api/0351-FoodLevelChangeEvent-Expose-Saturation.patch
@@ -5,14 +5,14 @@ Subject: [PATCH] FoodLevelChangeEvent Expose Saturation
 
 
 diff --git a/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java b/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
-index 7f4c5ea8607b08f8676528cc71b0b312575809ed..b1f5ed6656d9d80a3dcca2a5bacf9fc8079ab321 100644
+index 7f4c5ea8607b08f8676528cc71b0b312575809ed..963396f5f70a1b896faae00c6d6b565f9f25ab68 100644
 --- a/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
 +++ b/src/main/java/org/bukkit/event/entity/FoodLevelChangeEvent.java
 @@ -15,16 +15,27 @@ public class FoodLevelChangeEvent extends EntityEvent implements Cancellable {
      private boolean cancel = false;
      private int level;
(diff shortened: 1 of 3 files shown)


patches/server/0839-FoodLevelChangeEvent-Expose-Saturation.patch
```diff
@@ -0,0 +1,89 @@
...
+@@ -66,10 +66,11 @@ public class FoodData {
+                 this.saturationLevel = Math.max(this.saturationLevel - 1.0F, 0.0F);
+             } else if (enumdifficulty != Difficulty.PEACEFUL) {
+                 // CraftBukkit start
+-                org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, Math.max(this.foodLevel - 1, 0));
++                org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, Math.max(this.foodLevel - 1, 0), this.saturationLevel); // Paper - Expose Saturation
+ 
+                 if (!event.isCancelled()) {
+                     this.foodLevel = event.getFoodLevel();
++                    this.saturationLevel = event.getSaturationModifier(); // Paper - Expose Saturation
```
kennytv: This shouldn't be used as an absolute value here

kennytv (changes requested):

patches/server/0839-FoodLevelChangeEvent-Expose-Saturation.patch
Owen1212055: Can you explain what you mean by "absolute value"?

Machine-Maker: You are using the new field in the event for two different things. Sometimes its just the amount the saturation level will change, and here its what the player's final saturation level. That's not a good thing, it should only have one meaning, either how much the saturation level will change (the saturation delta) or a before and after of the player's saturation level.

Owen1212055: Ahh yea, I see what you mean here. Yea so this was the case that I wasn't really sure what to do about here, since the saturation level is changed higher up.  Should I set it at -1? make it a nullable primitive wrapper? I'm unsure.

Machine-Maker: You should be able to do a refactor of that code, probably duplicating the bit of code that lowers the sat level cause that happens regardless of difficulty, and adding a new fire of FoodLevelChangeEvent, with the level not changing, but the modifier changing (when the difficulty is peaceful). 

But this does give more merit to the idea of a separate event entirely.

patches/server/0839-FoodLevelChangeEvent-Expose-Saturation.patch
```diff
@@ -0,0 +1,89 @@
...
++++ b/src/main/java/net/minecraft/world/level/block/CakeBlock.java
+@@ -89,7 +89,7 @@ public class CakeBlock extends Block {
+             // entityhuman.getFoodData().eat(2, 0.1F);
+             int oldFoodLevel = player.getFoodData().foodLevel;
+ 
+-            org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, 2 + oldFoodLevel);
++            org.bukkit.event.entity.FoodLevelChangeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callFoodLevelChangeEvent(player, 2 + oldFoodLevel, 0.1F); // Paper - Expose Saturation
+ 
+             if (!event.isCancelled()) {
+                 player.getFoodData().eat(event.getFoodLevel() - oldFoodLevel, 0.1F);
```
Machine-Maker: You aren't using the sat level from the event here.

Machine-Maker (changes requested):

patches/server/0839-FoodLevelChangeEvent-Expose-Saturation.patch
Owen1212055: Yea, the issue is that sometimes the editing of saturation is also affected by this food level change event. So it's a bit hard to actually separate it properly without having to do some weirdness with this current event.

patches/api/0351-FoodLevelChangeEvent-Expose-Saturation.patch
```diff
@@ -0,0 +1,67 @@
...
+     public FoodLevelChangeEvent(@NotNull final HumanEntity what, final int level) {
+         this(what, level, null);
+     }
+ 
++    @Deprecated // Paper - Expose Saturation
+     public FoodLevelChangeEvent(@NotNull final HumanEntity what, final int level, @Nullable final ItemStack item) {
+         super(what);
+         this.level = level;
+         this.item = item;
++        this.saturationModifier = what.getSaturation(); // Just incase, make sure the player's saturation doesn't change. // Paper - Expose Saturation
```
Machine-Maker: This doesn't support what I would think most people take `saturationModifier` to mean, that its the change in saturation. I would call it `saturationLevel` if you want it to be what the player's saturation is set to after, not changed to. **But** based on what most of the calls to this event do, I think `saturationDelta` or `saturationChange` might be better, and if thats the case setting it to the entities sat level is incorrect

Owen1212055: Ah, yea... I think I defo misunderstood what this value actually meant. Thanks so much for this, I'll make the given changes.

Owen1212055:
Alright, I now capture the result and store it instead of trying to manually do the logic (except for the tick methods).
This also causes a fix as a result, where the food result would appear greater than 20 since there was no logic for making sure it was less than 20 in the event, but there was in the eat method.

I also made saturation represent the new changes saturation to fit food level.

aerulion:
I guess that it makes more sense to represent the "new saturation level" instead of the modifier, since the food level also provides the final value and not the delta. The only "problem" I have with it, is that in my opinion it kind of misses the point of my initial issue, at least regarding what I was trying to achieve. Having the final value might be useful in some cases, but it is less useful for modifying food, since one has to do the saturation calculation manually according to the modified food level and still has to figure out the correct formula. The only issue that having access to the value would solve, is somehow setting the new saturation after the event has processed. But maybe this event isn't ideal for implementing custom food, and a new event for eating would be the better way to go (That could also provide the food delta and the modifier, instead of final values).

Owen1212055:
> I guess that it makes more sense to represent the "new saturation level" instead of the modifier, since the food level also provides the final value and not the delta. The only "problem" I have with it, is that in my opinion it kind of misses the point of my initial issue, at least regarding what I was trying to achieve. Having the final value might be useful in some cases, but it is less useful for modifying food, since one has to do the saturation calculation manually according to the modified food level and still has to figure out the correct formula. The only issue that having access to the value would solve, is somehow setting the new saturation after the event has processed. But maybe this event isn't ideal for implementing custom food, and a new event for eating would be the better way to go (That could also provide the food delta and the modifier, instead of final values).

Agreed! Added a separate event for this, https://github.com/PaperMC/Paper/pull/7213

nathanfranke:
This shouldn't be closed as I desire this feature

electronicboy (approved):

nathanfranke:
bruh

PR #7196, opened 2021-12-26
Clear Titles On Join

### LINKED ISSUE #7161: Reset the players title message on join

Currently, you can send a title (maybe action bar too) to a player with a very long time basically infinite, then the player can leave the server and join another or even go into single-player and the title persists with no way to fix it unless you restart the game or send the player a new title which is annoying and frustrating. Many players go idle on Hypixel for example and get the "Your idle" title and then leave to join another server and they end up having to restart the game since it persists over. Resetting the title or sending a blank title fixes this.

Reset the players title/send a empty title on player join

A plugin? But it'll mess with any other plugin that wants to generally send a title on join.

### Other

_No response_

Owen1212055:
Resolves https://github.com/PaperMC/Paper/issues/7161


kennytv (changes requested):
Hard no from me, this breaks proxy networks

molor:
I think that this fix needs an option in `paper.yml` too

Owen1212055:
> Hard no from me, this breaks proxy networks

Yea so what if this was under a config option and perhaps only ran if bungeecord was set to false? Would like your opinion here, else we can just leave this to fix for Mc?

Machine-Maker:
If there's a way to know 100% that a proxy is being used, that's probably OK, otherwise I think the setting should default to off.

Owen1212055:
Added configuration setting disabled by default

Owen1212055:
Closing, as in general this is such a trivial issue and can easily be fixed yourself if you really want.

PR #7197, opened 2021-12-26
Sleep Message API + Config

### LINKED ISSUE #6863: Add option to disable "x/y players sleeping" message.

Owen1212055:
Resolves: https://github.com/PaperMC/Paper/issues/6863

Called per-player, allows for modifying the message and getting what type it represents. 
This also adds a configuration option, which will disable/enable these messages. If they are disabled the event will be called canceled by default.

```java
@EventHandler
    public void event(PlayerSleepMessageEvent event) {
        
        player.sendMessage(event.getMessageType().toString());
        player.sendMessage(event.getMessage());
        event.setMessage(Component.text("haha night go skippy"));
    }
```

image
image

### DIFF
--- a/patches/api/0351-Sleep-Message-API.patch
+++ b/patches/api/0351-Sleep-Message-API.patch
@@ -1,0 +1,114 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sun, 26 Dec 2021 00:22:29 -0500
+Subject: [PATCH] Sleep Message API
+
+
+diff --git a/src/main/java/io/papermc/paper/event/player/PlayerSleepMessageEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerSleepMessageEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..bbc8be31d9054f9300c45bd5bfba002e12470f37
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/player/PlayerSleepMessageEvent.java
+@@ -0,0 +1,102 @@
++package io.papermc.paper.event.player;
++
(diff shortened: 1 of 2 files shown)


kennytv:
This seems like an xy situation. If people want their own handling of this, they can set the `playersSleepingPercentage` gamerule to a value higher than 100. Changing the message only without how many players are sleeping doesn't actually hide that there is vanished players, which otherwise needs https://papermc.io/javadocs/paper/1.17/org/bukkit/entity/Player.html#setSleepingIgnored(boolean) (already affecting that number)

Owen1212055:
> This seems like an xy situation. If people want their own handling of this, they can set the `playersSleepingPercentage` gamerule to a value higher than 100. Changing the message only without how many players are sleeping doesn't actually hide that there is vanished players, which otherwise needs https://papermc.io/javadocs/paper/1.17/org/bukkit/entity/Player.html#setSleepingIgnored(boolean) (already affecting that number)

The issue is that even if you change the number of players you can't change the message still. It seems that people wanted the ability to cancel the message itself, in which case an event + config was easiest as it solved both issues.
I at least referenced the conversation in https://github.com/PaperMC/Paper/pull/6864.

Should I keep the config? The event? Or should this entire suggestion be reconsidered perhaps? I'm open to feedback about this.

pascalhanimann:
> This also adds a configuration option, which will disable/enable these messages. If they are disabled the event will be called canceled by default.

Assuming you still want feedback: For our server this would be the best way to go. I would love to have an option to disable the message entirely.

Badbird5907:
Any status on this? Since there is no current API to modify these messages.

PR #7202, opened 2021-12-26
Block Ticking API

Owen1212055:
This adds the ability to force tick blocks, useful if you want to for example simulate plant growing on certain blocks.


patches/server/0839-Block-Ticking-API.patch
```diff
@@ -0,0 +1,41 @@
...
++        blockData.getBlock().randomTick(blockData, level, position, level.random);
++    }
++
++    @Override
++    public boolean isRandomlyTicked() {
++        net.minecraft.world.level.block.state.BlockState blockData = this.getNMS();
++
++        return blockData.getBlock().isRandomlyTicking(blockData);
++    }
++    // Paper start - Block tick API
```
me4502: Should be Paper end, and doesn't need the comment about why

patches/api/0351-Block-Ticking-API.patch
```diff
@@ -0,0 +1,40 @@
...
++     */
++    void tick();
++
++    /**
++     * Causes the block to be ticked randomly.
++     * This has a chance to execute naturally if {@link Block#isRandomlyTicked()} is true.
++     * <p>
++     * For certain blocks, this behavior may be the same as {@link Block#tick()}.
++     */
++    void tickRandomly();
```
jpenilla: I think randomTick would be a better name than tickRandomly, as this is executing the random tick logic for the block, not causing it to tick randomly or something

patches/api/0351-Block-Ticking-API.patch
```diff
@@ -0,0 +1,40 @@
...
++
++    /**
++     * Gets if this block is ticked randomly in the world.
++     * The blocks current state may change this value.
++     *
++     * @return is ticked randomly
++     */
++    boolean isRandomlyTicked();
```
jpenilla: Might this method make more sense on BlockData?

jpenilla (dismissed):

Owen1212055:
Agreed with the provided comments, changed & rebased!

patches/server/0840-Block-Ticking-API.patch
```diff
@@ -0,0 +1,50 @@
...
++    }
++
++    @Override
++    public void randomTick() {
++        net.minecraft.world.level.block.state.BlockState blockData = this.getNMS();
++        var level = this.world.getMinecraftWorld();
++
++        blockData.getBlock().randomTick(blockData, level, position, level.random);
++    }
++    // Paper start - Block tick API
```
MeFisto94: nit: That's probably supposed to be Paper end instead

Owen1212055: I thought I removed that! Whoops!

electronicboy (approved):

Owen1212055:
Rebased

patches/server/0910-Block-Ticking-API.patch
```diff
@@ -0,0 +1,50 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/block/CraftBlock.java
+@@ -536,6 +536,23 @@ public class CraftBlock implements Block {
+         // SPIGOT-6778: Directly call setBlock instead of setTypeAndData, so that the tile entiy is not removed and custom remove logic is run.
+         return this.world.setBlock(position, fluidState.createLegacyBlock(), 3) && result; // Paper - leave liquid if waterlogged
+     }
++    // Paper start - Block tick API
++    @Override
++    public void tick() {
++        net.minecraft.world.level.block.state.BlockState blockData = this.getNMS();
++        var level = this.world.getMinecraftWorld();
```
Lulu13022002: var ?? :smiling_face_with_tear: (also see below)

Owen1212055:
Rebased

kennytv (approved):

PR #7204, opened 2021-12-26
Bucketable API

Owen1212055:
This exposes the bucket-able interface used in Minecraft, which are used by entities like axolotls, fishes, and the upcoming tadpole mob.

  This was decided to be put into a separate patch. [see relevant conversation](https://canary.discord.com/channels/289587909051416579/555462289851940864/923108395656020019)


patches/server/0839-Bucketable-API.patch
```diff
@@ -0,0 +1,69 @@
...
++    }
++
++    @Override
++    default void setFromBucket(boolean fromBucket) {
++        this.getHandle().setFromBucket(fromBucket);
++    }
++
++    @Override
++    default ItemStack getBaseBucketItem() {
++        return CraftItemStack.asCraftMirror(this.getHandle().getBucketItemStack());
```
Machine-Maker: This should be asBukkitCopy so it doesn't give any wrong ideas about modifying the returned itemstack having an effect on the entities ability to reflect changes (which it doesn't have)

Machine-Maker (changes requested):

jpenilla (approved):
Looks good, might want to look into also exposing `Bucketable.bucketMobPickup`, possibly a method to actually bucket the Bucketable, and another to get it's bucket item with nbt. (either in this PR or another)

Owen1212055:
> Looks good, might want to look into also exposing `Bucketable.bucketMobPickup`, possibly a method to actually bucket the Bucketable, and another to get it's bucket item with nbt. (either in this PR or another)

Yea thing is it basically uses raw interaction code, and fires an event. So I guess if it's really wanted it can be pr'd later, but this should be a good base if anyone would want to.

Machine-Maker (approved):

PR #7207, opened 2021-12-27
Lock Frozen Ticks API

### LINKED ISSUE #6812: setFrozen or freezetick similar to firetick

Owen1212055:
Resolves https://github.com/PaperMC/Paper/issues/6812

Allows you to lock freeze ticks to a certain number, allowing you to prevent it from being updated nicely.


kashike (approved):
Not tested, but the code looks good.

patches/server/0839-Freeze-Override-API.patch
```diff
@@ -0,0 +1,112 @@
...
++                // Paper end - Frozen Override API
+ 
+                 attributemodifiable.addTransientModifier(new AttributeModifier(LivingEntity.SPEED_MODIFIER_POWDER_SNOW_UUID, "Powder snow slow", (double) f, AttributeModifier.Operation.ADDITION));
+             }
+@@ -4276,6 +4280,9 @@ public abstract class LivingEntity extends Entity {
+ 
+     @Override
+     public boolean canFreeze() {
++        if (this.frozenOverride == net.kyori.adventure.util.TriState.TRUE) {return true;} // Paper - Frozen Override API
++        else if (this.frozenOverride == net.kyori.adventure.util.TriState.FALSE) {return false;} // Paper - Frozen Override API
```
Machine-Maker: I'm not sure about this. I think the `canFreeze` method should be left to its vanilla logic, but rather in places where canFreeze is used, check the override. Or maybe make another method that checks it and call that one in place of canFreeze

patches/server/0839-Freeze-Override-API.patch
```diff
@@ -0,0 +1,112 @@
...
++        } else {
++            this.entityData.set(Entity.DATA_TICKS_FROZEN, realFrozenTicks);
++        }
++    }
++    // Paper end - Frozen Override API
+ 
+     public void setTicksFrozen(int frozenTicks) {
+-        this.entityData.set(Entity.DATA_TICKS_FROZEN, frozenTicks);
++        updateFrozenEffect(); // Paper - Frozen Override API
++        this.realFrozenTicks = frozenTicks; // Paper - Frozen Override API
```
Owen1212055: blocker, need to fix this. The field needs to be set first.

Owen1212055:
Reworked pr to instead lock the frozen ticks, also made it persistent.

Machine-Maker (approved):

kennytv (approved):

PR #7212, opened 2021-12-27
Deprecate localized name API

Owen1212055:
This has been unused since 1.13 (see V1458 datafixer).
image

Because of how long this has been broken, it's not even worth fixing it IMO, just point towards adventure stuff.

### DIFF
--- a/patches/api/0177-Fix-Spigot-annotation-mistakes.patch
+++ b/patches/api/0177-Fix-Spigot-annotation-mistakes.patch
@@ -169,6 +169,42 @@ index 686e2a0b9fe061816b41435ef2337870dbdca8e5..aacf8ea85909299355d16cad0386072e
      public ItemMeta getItemMeta() {
          return this.meta == null ? Bukkit.getItemFactory().getItemMeta(this.type) : this.meta.clone();
      }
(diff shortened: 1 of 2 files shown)


Machine-Maker:
If it's unused, why is it still part of the CraftMetaItem thing then? it's still applied to the nbt tag of the itemstack, shouldn't that be removed if it doesn't do anything? and if that shouldn't be removed, then why? and depending on that reason, its possible that it shouldn't be deprecated.

Owen1212055:
> If it's unused, why is it still part of the CraftMetaItem thing then? it's still applied to the nbt tag of the itemstack, shouldn't that be removed if it doesn't do anything? and if that shouldn't be removed, then why? and depending on that reason, its possible that it shouldn't be deprecated.

My idea on the deprecation was because this has no actual effect on the client anymore, it simply does nothing. However fair point, I will prevent it from being written to and read from the item stack.

Machine-Maker:
Well... actually I'm not sure now. Probably some plugin that still depends on it. It is stored on the itemstack, and persisted through saving/loading. so that might break behavior, in which case just deprecation would be the way to go.

kennytv (approved):

Machine-Maker (approved):

PR #7213, opened 2021-12-28
PlayerEatItemEvent

Owen1212055:
Requires https://github.com/PaperMC/Paper/pull/7195

This event is meant to instead represent the food item being consumed only, which is useful when making custom food items. This is executed before the FoodLevelChangeEvent, only executing if the provided item stack is not null.


u9g:
Hey, I'm interested in this, would be great to have it reopened.

Cazcez:
I'm also interested with this

Owen1212055:
Will be basically resolved with 1.20.5

PR #7214, opened 2021-12-28
Respect x-ray permission in World#refreshChunk

### LINKED ISSUE #5551: World#refreshChunk​(int x, int z) is broken since 1.16
### Expected behavior

[World#refreshChunk​(int x, int z)](https://papermc.io/javadocs/paper/1.16/org/bukkit/World.html#refreshChunk-int-int-) should resend the entire chunk to nearby players. For example if Anti-Xray is enabled and a player gets the permission to bypass Anti-Xray, [World#refreshChunk​(int x, int z)](https://papermc.io/javadocs/paper/1.16/org/bukkit/World.html#refreshChunk-int-int-) should remove the Anti-Xray fake blocks from that chunk.

### Observed/Actual behavior

Since 1.16 this isn't the case anymore. The source code of [World#refreshChunk​(int x, int z)](https://papermc.io/javadocs/paper/1.16/org/bukkit/World.html#refreshChunk-int-int-) contains the following comments:
```java
// If there are more than 64 updates to a chunk at once, it will update all 'touched' sections within the chunk
// And will include biome data if all sections have been 'touched'
// This flags 65 blocks distributed across all the sections of the chunk, so that everything is sent, including biomes
```
However, this assumption isn't true anymore in 1.16. Only `PacketPlayOutMultiBlockChange` are used for those updates.

### Steps/models to reproduce

- Enable Anti-Xray in `engine-mode: 2` with `use-permission: true` and add `- air` to `hidden-blocks` in paper.yml.
- Join the server as spectator and non op (no permissions). (You should see the fake ores in the underground.)
- Op yourself.
- Invoke [World#refreshChunk​(int x, int z)](https://papermc.io/javadocs/paper/1.16/org/bukkit/World.html#refreshChunk-int-int-) for a visible chunk.

### Plugin list

A plugin to invoke [World#refreshChunk​(int x, int z)](https://papermc.io/javadocs/paper/1.16/org/bukkit/World.html#refreshChunk-int-int-).

### Paper version

> version
[01:33:01 INFO]: Checking version, please wait...
[01:33:01 INFO]: This server is running Paper version git-Paper-629 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)

### Agreements

Owen1212055:
Fixes: https://github.com/PaperMC/Paper/issues/5551

I guess the old behavior of this method was that the anti-xray was not run. However, around a month ago this method was fixed which caused it to use the x-ray behavior by default. This disables this, however, it is possible a new API method could be added to exclude Xray, but I am not sure. Opinions appreciated.

### DIFF
--- a/patches/server/0355-Anti-Xray.patch
+++ b/patches/server/0355-Anti-Xray.patch
@@ -1616,26 +1616,25 @@ index 669e4b41cf0751afb67d94b6a511bcfd18ce7ef4..3ff85d9d8518db712ca1d2977a5865da
  
      @Override
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
-index 0edb08a391f806e56ed1bd4812eb9c9d2b966bd7..f3a5f08541cb792aed21d3820aade1adde670252 100644
+index 0edb08a391f806e56ed1bd4812eb9c9d2b966bd7..6bdf3679ea9461475b139a0622857b1dc56ea5b2 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
 +++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
-@@ -401,11 +401,17 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+@@ -401,11 +401,16 @@ public class CraftWorld extends CraftRegionAccessor implements World {
                  List<ServerPlayer> playersInRange = playerChunk.playerProvider.getPlayers(playerChunk.getPos(), false);
                  if (playersInRange.isEmpty()) return;
  
 -                ClientboundLevelChunkWithLightPacket refreshPacket = new ClientboundLevelChunkWithLightPacket(chunk, this.world.getLightEngine(), null, null, true);
-+                Map<Object, ClientboundLevelChunkWithLightPacket> cacheMap = new HashMap<>(); // Paper - Xray refreshChunk
++                // Paper start - Anti-Xray - Bypass
++                Map<Object, ClientboundLevelChunkWithLightPacket> refreshPackets = new HashMap<>();
                  for (ServerPlayer player : playersInRange) {
                      if (player.connection == null) continue;
  
 -                    player.connection.send(refreshPacket);
-+                    // Paper start - Xray refreshChunk
 +                    Boolean shouldModify = chunk.getLevel().chunkPacketBlockController.shouldModify(player, chunk);
-+
-+                    player.connection.send(cacheMap.computeIfAbsent(shouldModify, (s) -> { // Use  connection to prevent creating firing event
++                    player.connection.send(refreshPackets.computeIfAbsent(shouldModify, s -> { // Use connection to prevent creating firing event
 +                        return new ClientboundLevelChunkWithLightPacket(chunk, this.world.getLightEngine(), null, null, true, (Boolean) s);
 +                    }));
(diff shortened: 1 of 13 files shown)


stonar96:
I think the correct solution is to use `chunk.getLevel().chunkPacketBlockController.shouldModify(player, chunk);` and create 2 different chunk packets if necessary, similar to how it's done [here](https://github.com/PaperMC/Paper/blob/master/patches/server/0362-Anti-Xray.patch#L1159-L1180). Imho a chunk resend shouldn't be treated differently than any other chunk send.

Owen1212055:
> I think the correct solution is to use `chunk.getLevel().chunkPacketBlockController.shouldModify(player, chunk);` and create 2 different chunk packets if necessary, similar to how it's done [here](https://github.com/PaperMC/Paper/blob/master/patches/server/0362-Anti-Xray.patch#L1159-L1180). Imho a chunk resend shouldn't be treated differently than any other chunk send.

So you think that the x-ray should be applied in this situation? A bit confused, as this PR removes x-ray. However, if it's instead preferred to have x-ray applied I can use that method instead for the implementation.

stonar96:
I think the chunk should be modified if the `#shouldModify(player, chunk)` method says so. (This method returns `true` if Anti-Xray is enabled and the player doesn't have the `paper.antixray.bypass` permission, `false` otherwise.)

Owen1212055:
> I think the chunk should be modified if the `#shouldModify(player, chunk)` method says so. (This method returns `true` if Anti-Xray is enabled and the player doesn't have the `paper.antixray.bypass` permission, `false` otherwise.)

Ahhh thank you so much, I misunderstood the issue. Thanks for the clarification!

Owen1212055:
@stonar96 Reimplemented to use the specified method, caching packets when possible.

stonar96:
LGTM, thanks for fixing this!

patches/server/0362-Anti-Xray.patch
```diff
@@ -1608,6 +1608,30 @@ index 91ef1b0e06c30668fe4bfb18ecdf2fe499f72fee..36b7de78fa69f652079d74252286bb6d
...
+-                ClientboundLevelChunkWithLightPacket refreshPacket = new ClientboundLevelChunkWithLightPacket(chunk, this.world.getLightEngine(), null, null, true);
++                Map<Object, ClientboundLevelChunkWithLightPacket> cacheMap = new HashMap<>(); // Paper - Xray refreshChunk
+                 for (ServerPlayer player : playersInRange) {
+                     if (player.connection == null) continue;
+ 
+-                    player.connection.send(refreshPacket);
++                    // Paper start - Xray refreshChunk
++                    Boolean shouldModify = chunk.getLevel().chunkPacketBlockController.shouldModify(player, chunk);
++
++                    player.connection.send(cacheMap.computeIfAbsent(shouldModify, (s) -> { // Use  connection to prevent creating firing event
```
stonar96: Just noticed that there are two spaces in `// Use  connection ...`. Not sure how nitpicky we are. :D

Owen1212055:
It should be noted that when upstream fixed this method they forgot to mark it as undeprecated, https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/pull-requests/682 will resolve this.

Owen1212055:
Rebased

electronicboy (approved):

electronicboy:
apparently awaiting some changes from stonar, leaving approval as this seems fine so far

Owen1212055:
Rebased and those changes have been rebased in ^
Should be all ready

PR #7223, opened 2021-12-30
BlockProperty API

Machine-Maker:
My take on how to add a "get block data properties individually" on top of bukkit's existing method-based system.

### Structure
3 types of data properties: `Boolean`, `Integer`, and `Enum`. This pr creates a base `BlockProperty` class with 3 (+ special cases) implementations of it for each type.

#### Special Cases
So far, I've noticed 2 API types that are used to wrap integers, so this also adds handling for those, not exposing them as integers.
* `rotation` as BlockFace
* `note` as Note

The `BlockPropertyHolder` interface contains all the methods to interact with a set of properties. I left it as a standalone interface in case other uses in the future popup where a separate impl might be warranted. I added a `Mutable` subinterface to that to separate out the "set" methods, some implementations of it might not want to support mutating properties (such as future FluidState API https://github.com/PaperMC/Paper/pull/8609).

### Tests
I wrote some significant tests to ensure all nms `Property`s from `BlockStateProperties` have a counterpart, and they are equal to each other.

### DIFF
--- a/patches/api/0501-BlockProperty-API.patch
+++ b/patches/api/0501-BlockProperty-API.patch
@@ -1,0 +1,1063 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 8 Dec 2021 16:49:33 -0800
+Subject: [PATCH] BlockProperty API
+
+
+diff --git a/src/main/java/io/papermc/paper/block/fluid/FluidData.java b/src/main/java/io/papermc/paper/block/fluid/FluidData.java
+index 0750219fc68261e5c396636967e0b633ae17b72e..09dd83cd595e7d4be18809e1ba0aec3a72d0232d 100644
+--- a/src/main/java/io/papermc/paper/block/fluid/FluidData.java
++++ b/src/main/java/io/papermc/paper/block/fluid/FluidData.java
+@@ -1,5 +1,6 @@
+ package io.papermc.paper.block.fluid;
+ 
++import io.papermc.paper.block.property.BlockPropertyHolder;
+ import org.bukkit.Fluid;
+ import org.bukkit.Location;
+ import org.bukkit.util.Vector;
+@@ -11,7 +12,7 @@ import org.jspecify.annotations.NullMarked;
+  * This type is not linked to a specific location and hence mostly resembles a {@link org.bukkit.block.data.BlockData}.
+  */
+ @NullMarked
+-public interface FluidData extends Cloneable {
++public interface FluidData extends Cloneable, BlockPropertyHolder {
+ 
+     /**
+      * Gets the fluid type of this fluid data.
+diff --git a/src/main/java/io/papermc/paper/block/property/AsIntegerProperty.java b/src/main/java/io/papermc/paper/block/property/AsIntegerProperty.java
+new file mode 100644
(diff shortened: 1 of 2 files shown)


kashike:
I do not like the name `DataProperty`. What is wrong with just `Property`? These are not limited to just blocks, as the game also uses them for fluids (and potentially more in the future?).

montlikadani:
> I do not like the name `DataProperty`. What is wrong with just `Property`?

I guess because there is already `org.bukkit.inventory.InventoryView.Property` enum in `InventoryView` class, to do not confuse other's about these 2 Property class or smth.

kashike:
That's nice, but not a reasonable reason.

yannicklamprecht:
> > I do not like the name `DataProperty`. What is wrong with just `Property`?
> 
> I guess because there is already `org.bukkit.inventory.InventoryView.Property` enum in `InventoryView` class, to do not confuse other's about these 2 Property class or smth.

No reason as the Damagable exists two times as well. For ItemMeta and the Entities.


I'm happy that such API get introduced into Paper-API. First steps to be able to handle custom block data in future.

Machine-Maker:
> I do not like the name `DataProperty`. What is wrong with just `Property`? These are not limited to just blocks, as the game also uses them for fluids (and potentially more in the future?).

What does the name “DataProperty” have to do with limiting it to blocks? Or limiting its expansion to other things. Nms calls them Property, and creates them all in a class called BlockStateProperties. They are properties for block/fluid states. And the closest Bukkit representation of that, is BlockData, which is used for both blocks and fluids. I think DataProperty makes sense in that context. 

On a much smaller improvement scale, it’s also different from the nms name for it, which makes imports just nicer.

Machine-Maker:
Ok, renamed to `Property` and added better handling for int properties that have non-int representations in the api
So far those special cases are
* BlockFace ("rotation")
* Note ("note")

Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased for 1.19.1

electronicboy (approved):
Looking over this, this looks fine

patches/api/0391-Property-API.patch
```diff
@@ -0,0 +1,949 @@
...
++import java.util.Set;
++
++final class NoteProperty extends Property<Note> implements IntegerProperty.IntRepresented<Note> {
++
++    private final BiMap<Integer, Note> cache;
++
++    public NoteProperty(@NotNull String name) {
++        super(name, Note.class);
++        this.cache = HashBiMap.create(24);
++        for (int i = 0; i <= 24; i++) {
```
Lulu13022002, line 311: You should use BlockStateProperties.NOTE.max to avoid magic value.

yannicklamprecht (approved):

Owen1212055 (approved):
As specified before, we probably want to rename this to BlockProperty or something. 

The propertyholder/etc classes however I'm not sure about, because that may be useful API to have shared with items/blocks/etc. Unsure tho. The item property PR however creates a very similar property holder class and such.

I'd also maybe mark the block properties for stuff like chiseled bookshelf as experimental.
Looks good tho, really nice powerful api! 😄

kashike:
> As specified before, we probably want to rename this to BlockProperty or something.

I disagree, that would limit its use for other things in the future.

Owen1212055:
> > As specified before, we probably want to rename this to BlockProperty or something.
> 
> I disagree, that would limit its use for other things in the future.

Then we may need to see about aligning this property API in a way that makes it work with the item property api too then?
Because at least, it was discussed for them to be separate concepts, and if we are wanting to inline them there's gonna need more talk with that.

PR #7229, opened 2021-12-30
fix powder snow cauldrons not turning to water

Machine-Maker:
Powder snow cauldrons should turn to water when
extinguishing an entity

This diff could be a lot cleaner if we broke the signature of LayeredCauldronBlock#handleEntityOnFireInside to add an entity param and return a bool.


Owen1212055 (approved):
Tested and works

kennytv (approved):

PR #7232, opened 2021-12-31
Add method for getting a sound's subtitle key

Machine-Maker:
Soooooo since the client uses one of the many resources it downloads for associated a sound with a subtitle key, the test makes a request to mojangs resources api. idk how good that is, maybe the test should be disabled most of the time or smth. But, it works. I also wrote a function that generates the big switch statement thats commented out of the test by default cause most of the time it doesnt need to be run.


Owen1212055 (approved):
Tested, looks good.

Machine-Maker:
Rebased for 1.19.2.

Currently, this uses the `taskCache` folder inside `Paper-Server` instead of the root one. Idk if that matters, it just threw me for a loop when I went looking for the downloaded json files earlier today.

jpenilla:
It shouldn't be a big deal since it's gitignored. Maybe try building and then changing a patch, reapplying patches and building again with --console=verbose to see if it reruns the task

PR #7242, opened 2022-01-01
Update the rewriteForIde for new cb package version

Machine-Maker:
I'm pretty sure this is still a useful thing. For example, taking the [paperweight-test-plugin](https://github.com/PaperMC/paperweight-test-plugin) and loading the non-reobf'd plugin jar on a mojmapped server without relocations will currently fail because its not rewriting the relocations. But with this (and running the server with `-Ddebug.rewriteForIde` it now works.

Not sure on the best way to automatically know the cb package, so I just threw it in the manifest, prob some better way to do it with gradle that I just don't know.

### DIFF
--- a/patches/server/0003-Build-system-changes.patch
+++ b/patches/server/0003-Build-system-changes.patch
@@ -21,7 +21,7 @@ index 75c858870a55f7282a0e3500f803cb330b462f37..d39ed9a8a32913193aec1bba2a59a352
      runtimeOnly("org.xerial:sqlite-jdbc:3.36.0.3")
      runtimeOnly("mysql:mysql-connector-java:8.0.27")
  
-@@ -32,6 +31,7 @@ tasks.jar {
+@@ -33,6 +32,7 @@ tasks.jar {
          val gitHash = git("rev-parse", "--short=7", "HEAD").getText().trim()
          val implementationVersion = System.getenv("BUILD_NUMBER") ?: "\"$gitHash\""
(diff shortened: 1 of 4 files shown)


patches/server/0003-Build-system-changes.patch
```diff
@@ -21,20 +21,21 @@ index 9c4382ef7373110b6e72163779316ff40234a992..bba267ecca0121df22fdfa6096101786
...
              "Main-Class" to "org.bukkit.craftbukkit.Main",
              "Implementation-Title" to "CraftBukkit",
-@@ -44,6 +44,8 @@ tasks.jar {
+@@ -45,6 +45,9 @@ tasks.jar {
              "Specification-Title" to "Bukkit",
              "Specification-Version" to project.version,
              "Specification-Vendor" to "Bukkit Team",
 +            "Git-Branch" to gitBranch, // Paper
 +            "Git-Commit" to gitHash, // Paper
++            "Craftbukkit-Package-Version" to craftbukkitPackageVersion, // Paper
```
kashike:
```suggestion
+            "CraftBukkit-Package-Version" to craftbukkitPackageVersion, // Paper
```

kashike (dismissed):

patches/server/0003-Build-system-changes.patch
Machine-Maker: done

kennytv (approved):

Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

PR #7244, opened 2022-01-01
Improve GameRule API

Machine-Maker:
Getting a gamerule's default value shouldn't be tied to needing a world instance since it's the same without a world. Also, the methods to get the gamerule's value for a world and default world should be notnull, so the methods I added to GameRule to get/set the world have an Objects.requireNonNull wrapped around them.

### DIFF
--- a/patches/api/0490-Improve-GameRule-API.patch
+++ b/patches/api/0490-Improve-GameRule-API.patch
@@ -1,0 +1,103 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 1 Jan 2022 15:54:09 -0800
+Subject: [PATCH] Improve GameRule API
+
+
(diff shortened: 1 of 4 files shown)


kennytv:
I don't like how this duplicates all the methods for no real reason, the nullability change is just a confusing hack with the world method still being marked as nullable. Imo the getDefault in Server and the world methods in GameRule should be removed

Machine-Maker:
> I don't like how this duplicates all the methods for no real reason, the nullability change is just a confusing hack with the world method still being marked as nullable. Imo the getDefault in Server and the world methods in GameRule should be removed

I need the getDefault in server to have a way of getting the default rule value without a reference to the world. currently, the only way to get the default rule value requires a rule, which is totally unneeded.

kennytv:
Ah I didn't see GameRule is not an interface implemented in CraftBukkit

Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

Machine-Maker:
Rebased for 1.19.1

PR #7245, opened 2022-01-02
Add ClientGameEvents

tahmid-23:
I realize this is a bit ugly; I wasn't sure whether to expose the ClientGameEvent's ID as internal API or to use a massive else-if, and there's no namespaced key to use either
The value param is also a magic value, but as far as I see I don't really know how to use some of these game events without a magic value param other than creating a method for each case

### DIFF
--- a/patches/api/0353-Add-ClientGameEvents.patch
+++ b/patches/api/0353-Add-ClientGameEvents.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: thamid-gamer <60953955+thamid-gamer@users.noreply.github.com>
+Date: Sat, 1 Jan 2022 13:07:08 -0500
+Subject: [PATCH] Add ClientGameEvents
+
+
+diff --git a/src/main/java/io/papermc/paper/ClientGameEvent.java b/src/main/java/io/papermc/paper/ClientGameEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..df0727ec2f30f959caf792aed50fa793184903e0
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/ClientGameEvent.java
+@@ -0,0 +1,25 @@
++package io.papermc.paper;
++
++/**
++ * An event which changes part of a client's status.
++ */
++public final class ClientGameEvent {
++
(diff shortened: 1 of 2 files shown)


patches/api/0353-Add-ClientGameEvents.patch
```diff
@@ -0,0 +1,58 @@
...
++    public final static ClientGameEvent WIN_GAME = new ClientGameEvent();
++    public final static ClientGameEvent DEMO_EVENT = new ClientGameEvent();
++    public final static ClientGameEvent ARROW_HIT_PLAYER = new ClientGameEvent();
++    public final static ClientGameEvent RAIN_LEVEL_CHANGE = new ClientGameEvent();
++    public final static ClientGameEvent THUNDER_LEVEL_CHANGE = new ClientGameEvent();
++    public final static ClientGameEvent PUFFER_FISH_STING = new ClientGameEvent();
++    public final static ClientGameEvent GUARDIAN_ELDER_EFFECT = new ClientGameEvent();
++    public final static ClientGameEvent IMMEDIATE_RESPAWN = new ClientGameEvent();
++
++}
```
ghost, line 37: Shouldn't this be an enum?

patches/server/0844-Add-ClientGameEvents.patch
```diff
@@ -0,0 +1,66 @@
...
++        }
++        else if (io.papermc.paper.ClientGameEvent.GUARDIAN_ELDER_EFFECT.equals(event)) {
++            type = ClientboundGameEventPacket.GUARDIAN_ELDER_EFFECT;
++        }
++        else if (io.papermc.paper.ClientGameEvent.IMMEDIATE_RESPAWN.equals(event)) {
++            type = ClientboundGameEventPacket.IMMEDIATE_RESPAWN;
++        }
++        else {
++            throw new IllegalArgumentException("Unknown ClientGameEvent");
++        }
```
ghost, line 58: This should be replaced with a switch statement.

Machine-Maker (changes requested):
I don't like this at all. That float value is just a magic value, with no indication of what a different value will do. Some of these already have implementations, like you can already do an immediate respawn, I'm not sure of the best way to handle all of these, or even if any of them should be directly exposed like this, and not have specific methods for each.

PR #7249, opened 2022-01-02
Fix FurnaceInventory for smokers and blast furnaces

Machine-Maker:
creating an inventory of type smoker or blast furnace should return an instance of CraftInventoryFurnace, but I noticed when I tried to cast it, it was a CraftInventory. This is not correct, because literally everything that applies to Furnace furnaces applies to the other 2 as well. The ctor for CraftInventoryFurnace even takes the AbstractFurnaceBlockEntity so all existing methods work as they should.


Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

PR #7251, opened 2022-01-02
Better menu property API

Machine-Maker:
~~Needs a bunch of javadoc comments still~~

The old property system had a few big issues (`InventoryView$Property`), 
1.  it was incomplete, missing 2 properties, one for stonecutter, and one for loom
2. it only allowed for one inventory type per property which is not true anymore, the 4 furnace ones apply to Furnaces, Blast Furnaces, and Smokers
3. It relied on someone using magic values to set potion effect types or enchantments which is bad practice for the API

I changed it into a class format, with a generic for the value type that's converted to an integer in the server impl.

### DIFF
--- a/patches/api/0449-Better-menu-property-API.patch
+++ b/patches/api/0449-Better-menu-property-API.patch
@@ -1,0 +1,464 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 2 Jan 2022 00:39:48 -0800
+Subject: [PATCH] Better menu property API
+
+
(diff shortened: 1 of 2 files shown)


kashike:
https://github.com/PaperMC/Paper/pull/7223#issuecomment-1002915683

Machine-Maker:
Literally every class I ever add now, is gonna be called `DataProperty` :smiling_imp:

Machine-Maker:
Should be good, finished adding javadocs

Owen1212055 (approved):
image

Tested with some inventories, looks good!
(Behavior with sharing properties for stuff like smokers and furnace works fine as well)

Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased and updated for 1.20.4.

Renamed type to MenuProperty
Added crafter options

Machine-Maker:
No longer needed because there are subinterfaces for each view type (like EnchantmentView, BrewingStandView, etc.).

PR #7252, opened 2022-01-02
Implement Translatable on InventoryType

Machine-Maker:
All but 2 of the bukkit InventoryType's can easily implement Translatable. The two that cant, Composter (which doesn't have a UI to begin with, no clue why its part of the bukkit enum) and Merchant which relies on the villager name. So can either to a nonnull method that returns an empty string for its translation key, or a nullable method and NOT implement Translatable, or have it throw an exception of there isn't one and someone tries to use the method.

### DIFF
--- a/patches/api/0211-Add-methods-to-get-translation-keys.patch
+++ b/patches/api/0211-Add-methods-to-get-translation-keys.patch
@@ -398,6 +398,217 @@ index d841d94d46462e0ceb7c6b04cc8fc36792bd9201..8c8176121cafed0ed09239b6a7b392dc
      }
  
      // Paper start - Add villager reputation API
+diff --git a/src/main/java/org/bukkit/event/inventory/InventoryType.java b/src/main/java/org/bukkit/event/inventory/InventoryType.java
+index 19ba2d948ad83baab2a14ae6f7b3ce43c3d4971f..293711e1ca784d1fdff8da4f8bfc88c3204e39fd 100644
+--- a/src/main/java/org/bukkit/event/inventory/InventoryType.java
++++ b/src/main/java/org/bukkit/event/inventory/InventoryType.java
+@@ -19,136 +19,136 @@ import org.jetbrains.annotations.NotNull;
+  *
+  * @see org.bukkit.Bukkit#createInventory(InventoryHolder, InventoryType)
+  */
+-public enum InventoryType {
++public enum InventoryType implements net.kyori.adventure.translation.Translatable { // Paper - translations
+ 
+     /**
+      * A chest inventory, with 0, 9, 18, 27, 36, 45, or 54 slots of type
+      * CONTAINER.
+      */
+-    CHEST(27, "Chest"),
++    CHEST(27, "Chest", "container.chest"), // Paper
+     /**
+      * A dispenser inventory, with 9 slots of type CONTAINER.
+      */
+-    DISPENSER(9, "Dispenser"),
++    DISPENSER(9, "Dispenser", "container.dispenser"), // Paper
+     /**
+      * A dropper inventory, with 9 slots of type CONTAINER.
+      */
+-    DROPPER(9, "Dropper"),
++    DROPPER(9, "Dropper", "container.dropper"), // Paper
+     /**
+      * A furnace inventory, with a RESULT slot, a CRAFTING slot, and a FUEL
+      * slot.
+      */
+-    FURNACE(3, "Furnace"),
++    FURNACE(3, "Furnace", "container.furnace"), // Paper
+     /**
+      * A workbench inventory, with 9 CRAFTING slots and a RESULT slot.
+      */
+-    WORKBENCH(10, "Crafting"),
++    WORKBENCH(10, "Crafting", "container.crafting"), // Paper
+     /**
+      * A player's crafting inventory, with 4 CRAFTING slots and a RESULT slot.
+      * Also implies that the 4 ARMOR slots are accessible.
+      */
+-    CRAFTING(5, "Crafting", false),
++    CRAFTING(5, "Crafting", false, "container.crafting"), // Paper
+     /**
+      * An enchantment table inventory, with two CRAFTING slots and three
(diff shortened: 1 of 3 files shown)


Owen1212055 (approved):
Looks good, although I really wonder WHY the composter has this special inventory type to begin with. 

Probably out of the scope of this PR however..

Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased for 1.19.2

patches/api/0220-Add-methods-to-get-translation-keys.patch
```diff
@@ -393,6 +393,198 @@ index c799ac46dbe257d470d3b236cf55b41240f5fda8..d61855b83da0776fe910dee8cde184c7
...
+      */
+-    BREWING(5, "Brewing"),
++    BREWING(5, "Brewing", "container.brewing"), // Paper
+     /**
+      * A player's inventory, with 9 QUICKBAR slots, 27 CONTAINER slots, 4 ARMOR
+      * slots and 1 offhand slot. The ARMOR and offhand slots may not be visible
+      * to the player, though.
+      */
+-    PLAYER(41, "Player"),
++    PLAYER(41, "Player", "container.crafting"), // Paper
```
Lulu13022002: I think this should be null, the survival crafting slots is not even mentioned in the javadoc, and the whole inventory doesn't have a name. Additionally those 4 crafting slots gets the right CRAFTING type and not PLAYER.

patches/api/0220-Add-methods-to-get-translation-keys.patch
```diff
@@ -393,6 +393,198 @@ index c799ac46dbe257d470d3b236cf55b41240f5fda8..d61855b83da0776fe910dee8cde184c7
...
+-    private InventoryType(int defaultSize, /*@NotNull*/ String defaultTitle) {
+-        this(defaultSize, defaultTitle, true);
+     }
+ 
+-    private InventoryType(int defaultSize, /*@NotNull*/ String defaultTitle, boolean isCreatable) {
++    private InventoryType(int defaultSize, /*@NotNull*/ String defaultTitle, boolean isCreatable, @org.jetbrains.annotations.Nullable String translationKey) { // Paper
+         size = defaultSize;
+         title = defaultTitle;
+         this.isCreatable = isCreatable;
+         this.defaultTitleComponent = net.kyori.adventure.text.Component.text(defaultTitle); // Paper - Adventure
```
Lulu13022002: Maybe this should be updated, but ideally the title would be nullable if you don't want to break the api you could fallback to the hardcoded value for inventory that doesn't have a translation key

patches/server/0477-Add-methods-to-get-translation-keys.patch
```diff
@@ -174,5 +174,16 @@ index fab3063fffa959ac7f0eb5937f2fae94d11b6591..9292a0119499d14c9ed170999ac3b8df
...
++
++    @Test
++    public void testInventoryType() {
++        for (org.bukkit.event.inventory.InventoryType type : org.bukkit.event.inventory.InventoryType.values()) {
++            if (type == org.bukkit.event.inventory.InventoryType.MERCHANT || type == org.bukkit.event.inventory.InventoryType.COMPOSTER) {
++                continue;
++            }
++            Assert.assertNotEquals(type + " has an unresolved translation key",type.translationKey(), net.kyori.adventure.text.serializer.plain.PlainTextComponentSerializer.plainText().serialize(net.kyori.adventure.text.Component.translatable(type)));
++        }
++        Assert.assertThrows(IllegalStateException.class, org.bukkit.event.inventory.InventoryType.MERCHANT::translationKey);
```
Lulu13022002: This would be more usefull in the loop before the continue i think

patches/server/0477-Add-methods-to-get-translation-keys.patch
```diff
@@ -174,5 +174,16 @@ index fab3063fffa959ac7f0eb5937f2fae94d11b6591..9292a0119499d14c9ed170999ac3b8df
...
 +        }
++    }
++
++    @Test
++    public void testInventoryType() {
++        for (org.bukkit.event.inventory.InventoryType type : org.bukkit.event.inventory.InventoryType.values()) {
++            if (type == org.bukkit.event.inventory.InventoryType.MERCHANT || type == org.bukkit.event.inventory.InventoryType.COMPOSTER) {
++                continue;
++            }
++            Assert.assertNotEquals(type + " has an unresolved translation key",type.translationKey(), net.kyori.adventure.text.serializer.plain.PlainTextComponentSerializer.plainText().serialize(net.kyori.adventure.text.Component.translatable(type)));
```
Lulu13022002: Missing space between the first and second argument. I think another test here that would compare and check that the hardcoded values are equals to their en_us translation might be convenient (this is especially useful if the defaultTitle is not updated see above)

patches/api/0220-Add-methods-to-get-translation-keys.patch
Machine-Maker: I use that, cause that's what is used on the client for the player's inventory screen.

patches/api/0220-Add-methods-to-get-translation-keys.patch
Machine-Maker: idk if it would even be breaking the API. it returns a component, no guarantees of what type of component.

patches/server/0477-Add-methods-to-get-translation-keys.patch
Machine-Maker: Not a good way of dynamically getting the en_us values, so it'd just be more hardcoding I think.

Warriorrrr (approved):

lynxplay:
Closing as not really applicable to inventory type (translation key for  GENERIC3x9 is different for  chest vs barrel.
Translatable also isn't as great anymore with more and more things potentially defining their display components as non-translatables once data driven.

PR #7254, opened 2022-01-02
Provide option to disable username validation (Closes #7250)

### LINKED ISSUE #7250: Request add `enable-validate-username` option in `paper.yml`

Yes.

Without this option, players using UTF-8 name cannot join my server. The last version of paper-1.17.1 we can use is `paper-1.17.1-404.jar`, since the 405 version (Commit: [2a827e3](https://github.com/PaperMC/Paper/commit/2a827e31abec3602dd1ed381d207a6cbb16a8afc)) has added a validate username part which will force validate player's usernames, and kick players using UTF-8 names, and we cannot disable this function.

To give an example, there is a Minecraft version is `China Edtion` (https://minecraft.fandom.com/wiki/China_Edition), it's jointly operated by Mojang and NetEase, this is a special Minecraft version for Chinese players, and they can use China words in game. Which looks like this:
China Edtion 1
China Edtion 2

I'm running servers on China Edtion, and i'm also using Paper, the `validate username` update has completely prevented me from using the latest version. I'll be very appreciate if you could add an option that let us enable/disable the validate username function.

Add an option which could control whether to enable/disable the validate username function.

There is no alternatives I think, except I revert the `validate username` commit, and build Paper from source myself, but if I do like this it will be very troublesome, and all future paper versions need to be built by me, and I cannot receive some plugins' support since they will not provide support for self-built paper.

### Other

Thank you for your amazing work, and happy new year!

electronicboy:
I hate this, but, I don't see any other sane option especially for oddball setups which blatantly violate mojangs apis. This is an unsupported setting for various reasons

### DIFF
--- a/patches/server/0842-Validate-usernames.patch
+++ b/patches/server/0842-Validate-usernames.patch
@@ -4,8 +4,25 @@ Date: Sat, 1 Jan 2022 05:19:37 -0800
 Subject: [PATCH] Validate usernames
 
 
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index d5aa95846e7f52108a03e3731023527281b21d73..1d3cc8836d2ccbec4a8660f86501be35c76e8b0b 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+@@ -493,6 +493,12 @@ public class PaperConfig {
+         set("settings.unsupported-settings.allow-tnt-duplication", null);
+     }
+ 
++    public static boolean performUsernameValidation;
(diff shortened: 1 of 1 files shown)


Spottedleaf (approved):

kashike (approved):

BillyGalbreath:
Why not just [expose it as regex](https://github.com/PurpurMC/Purpur/blob/ver/1.18.1/patches/server/0252-Configurable-valid-characters-for-usernames.patch) so the server admin has control over it instead of an all or nothing kinda deal?

electronicboy:
Was very iffy on that because I really do not wanna be providing support for how to write a regex especially when things like china edition come into play

Bedlessdeus (commented):
Confirmed, why not in 1.8
(Ik 1.8 isnt supported, its Kinda a joke cus i need this option, so i now have to add it myself...)

Issue #7259, opened 2022-01-02
Cannot disable tripwire strings with shears since 1.18.1/83


Auzathoth:
### Expected behavior

Left-clicking with some shears on a string attached to a tripwire hook should disarm the string so any entity standing on the block does not trigger the tripwire. The "disarmed" state of the minecrat:tripwire block should be set to "true".

### Observed/Actual behavior

The string/tripwire block is broken, and a string item is dropped

### Steps/models to reproduce

Start a new Paper server, login and in creative, obtain some shears, tripwire hooks and string. Set up a tripwire three or more blocks wide (eg solid block, tripwire hook, string[, string], tripwire hook, solid block,

Switch to survival mode and left click with the shears on any string between the two tripwire hooks.

### Plugin and Datapack List

> plugins
[14:43:16 INFO]: Plugins (0):

> datapack list
[14:43:28 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
[14:43:28 INFO]: There are no more data packs available

### Paper version

> version
[14:43:40 INFO]: This server is running Paper version git-Paper-126 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: d3f3429)
You are running the latest version

### Other

This behaviour does not happen in Paper-82, and starts happening in Paper-83 after this [commit](https://github.com/PaperMC/Paper/commit/6178609e1d7e9fda3d59786b69425a3b00d73298).

PR #7261, opened 2022-01-02
Fixed Tripwires being destroyed by shears #7259

zJanny:
Fixed Tripwires beeing destroyed by sheers instead of the Tripwire beeing disarmed


Owen1212055 (changes requested):
Instead of making a new patch, edit the one that broke this behavior. (https://github.com/PaperMC/Paper/commit/6178609e1d7e9fda3d59786b69425a3b00d73298)

zJanny:
It should now be included in the .patch file that broke the feature.

(Sorry for these messy commits i am not very good with github)

zJanny:
> Instead of making a new patch, edit the one that broke this behavior. ([6178609](https://github.com/PaperMC/Paper/commit/6178609e1d7e9fda3d59786b69425a3b00d73298))

I have included my changes in the patch file you referenced.

PR #7263, opened 2022-01-02
Add method to set noclip on arrows

Machine-Maker:
Tested by firing one of each of the 3 arrows, (regular, spectral, and tipped) and a trident. The arrows work as expected, the trident just immediately returns to the player without going anywhere. Should that be documented?


kennytv (approved):

Owen1212055 (commented):
Tested and looks good, I would say to document the behavior with tridents.

Additionally, with arrows it causes them to be able to be immediately picked up by players.
Tested with EntityShootBowEvent

electronicboy (approved):

Issue #7266, opened 2022-01-03
FilteredText not being set via sign API


AoElite:
### Expected behavior

Whenever you set a sign's text via the API it should set the sign's FilteredText1-4 nbt as well as Text1-4.

### Observed/Actual behavior

Whenever you set a sign's text via the API it does not set the sign's FilteredText1-4 nbt but does Text1-4.

### Steps/models to reproduce

```java
Block block = //get sign
Sign sign = (Sign) block.getState();
sign.line(1, Component.text().append(Component.text("hello!")).color(NamedTextColor.BLACK).build());
sign.update();
```
You can examine the sign's NBT data and see that it's only changing the `Text1` field in the nbt json but the `FilteredText1` is not being changed.

This is a issue for players who have the filtering text option enabled as they will not be able to see the sign's text. I do not know how players have this filtered text feature enabled and neither do they. I've seen it happen for players using any type of client, vanilla, fabric, and lunar randomly. However they all have one thing in common; they are sending a settings packet to the server that says they have text filtering enabled. https://wiki.vg/Protocol#Client_Settings

### Plugin and Datapack List

[02:42:56 INFO]: Plugins (0):

[02:43:20 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]

### Paper version

[02:43:47 INFO]: This server is running Paper version git-Paper-407 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: bf23668 on ver/1.17.1)
You are running the latest version

MiniDigger:
filtering can be enabled on child accounts, people should check their privacy settings in the xbox account settings.

have you checked if this is only a problem with the adventure method or does it also affect the normal bukkit method?

AoElite:
Bukkit method has this issue as well.

MiniDigger:
has or had, after that commit?

kennytv:
had, because the update isn't done in the setLine/line methods but in one update method

Flugelo:
save bro.

Do you intend to put the entities in async?

I realized that the only limitation of 1.18.1 is on entities that run on MainThread...

if they managed to put the entities in Async, I'm sure the server would run 10x better.

TheShermanTanker:
> 

Blindly dumping things onto another thread is never going to help, if that was all it took to optimize logic flow Paper (and quite possibly Mojang) would've done it by now

PR #7268, opened 2022-01-03
Add text filtering and allow server listing client options

NoahvdAa:
Might be useful for servers that implement custom chat filters and want to detect if they should filter chat by default for a particular person.

I don't have a Microsoft child account, so I tested this with the following pakkit script:
```java
exports.upstreamHandler = function (meta, data, server, client) {
  if (meta.name === 'settings') {
    data.enableTextFiltering = true
  }
  server.sendPacket(meta, data)
}

exports.downstreamHandler = function (meta, data, server, client) {
  client.sendPacket(meta, data)
}
```

### DIFF
--- a/patches/api/0194-Add-Player-Client-Options-API.patch
+++ b/patches/api/0194-Add-Player-Client-Options-API.patch
@@ -90,10 +90,10 @@ index 0000000000000000000000000000000000000000..4a0c39405d4fbed457787e3c6ded4cc6
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerClientOptionsChangeEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerClientOptionsChangeEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..d4509813aa523bccce3c58cf55ed1144007d5d64
+index 0000000000000000000000000000000000000000..9521092cdcd954249e11833e91cf64befc6ca2e7
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerClientOptionsChangeEvent.java
-@@ -0,0 +1,124 @@
+@@ -0,0 +1,125 @@
 +package com.destroystokyo.paper.event.player;
 +
 +import com.destroystokyo.paper.ClientOption;
@@ -123,6 +123,7 @@ index 0000000000000000000000000000000000000000..d4509813aa523bccce3c58cf55ed1144
 +    private final boolean allowServerListings;
 +    private final boolean textFilteringEnabled;
 +
++    @Deprecated
 +    public PlayerClientOptionsChangeEvent(@NotNull Player player, @NotNull String locale, int viewDistance, @NotNull ChatVisibility chatVisibility, boolean chatColors, @NotNull SkinParts skinParts, @NotNull MainHand mainHand) {
 +        this(player, locale, viewDistance, chatVisibility, chatColors, skinParts, mainHand, false, false);
 +    }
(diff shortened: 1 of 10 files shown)


kennytv (dismissed):
I wouldn't add API for an unfinished system

NoahvdAa:
I assumed it was at least somewhat finished because of [Mini's comment](https://github.com/PaperMC/Paper/issues/7266#issuecomment-1003948243), but haven't been able to check for myself (because I don't have a child account)

> filtering can be enabled on child accounts, people should check their privacy settings in the xbox account settings.

MiniDigger:
since this is part of the settings packet, why not expand the client settings api? https://papermc.io/javadocs/paper/1.18/com/destroystokyo/paper/ClientOption.html
https://github.com/PaperMC/Paper/blob/master/patches/server/0400-Implement-Player-Client-Options-API.patch
def one of my nicer apis :D

Owen1212055:
I didn't think the filtered text functionality was currently functional? At least the implementation in Minecraft is current no-op. Not to mention I don't think the tags are even used yet, only are currently being stored (not even filtered).

NoahvdAa:
Updated to add a ClientOption instead. Also added a `ALLOW_SERVER_LISTINGS` option since I noticed it was missing.

patches/api/0194-Add-Player-Client-Options-API.patch
```diff
@@ -118,15 +120,23 @@ index 0000000000000000000000000000000000000000..f7f171c4ee0b8339b2f8fbe82442d65f
 +    private final boolean chatColors;
 +    private final SkinParts skinparts;
 +    private final MainHand mainHand;
++    private final boolean allowServerListings;
++    private final boolean textFilteringEnabled;
 +
 +    public PlayerClientOptionsChangeEvent(@NotNull Player player, @NotNull String locale, int viewDistance, @NotNull ChatVisibility chatVisibility, boolean chatColors, @NotNull SkinParts skinParts, @NotNull MainHand mainHand) {
```
MiniDigger, line 127: maybe deprecate the old ctor?

MiniDigger (approved):

Doc94:
im sure the text filter its disable in NMS.. based in a spigot ticket (this ticket because the filter not work with chat in spigot) https://hub.spigotmc.org/jira/browse/SPIGOT-6805

And in Mojand JIRA has a issue for this https://bugs.mojang.com/browse/MC-233228

i hope im not confuse with another feature xd

patches/api/0194-Add-Player-Client-Options-API.patch
```diff
@@ -118,15 +120,24 @@ index 0000000000000000000000000000000000000000..f7f171c4ee0b8339b2f8fbe82442d65f
...
 +    private final MainHand mainHand;
++    private final boolean allowServerListings;
++    private final boolean textFilteringEnabled;
 +
++    @Deprecated
 +    public PlayerClientOptionsChangeEvent(@NotNull Player player, @NotNull String locale, int viewDistance, @NotNull ChatVisibility chatVisibility, boolean chatColors, @NotNull SkinParts skinParts, @NotNull MainHand mainHand) {
++        this(player, locale, viewDistance, chatVisibility, chatColors, skinParts, mainHand, false, false);
++    }
++
++    public PlayerClientOptionsChangeEvent(@NotNull Player player, @NotNull String locale, int viewDistance, @NotNull ChatVisibility chatVisibility, boolean chatColors, @NotNull SkinParts skinParts, @NotNull MainHand mainHand, boolean allowServerListings, boolean textFilteringEnabled) {
```
Machine-Maker, line 131: I feel like this ctor should take some other form, a set of wrapper objects that then get spread back out into individual fields, or a map of client options to values.

 otherwise deprecated ctors are just gonna stack up

patches/server/0400-Implement-Player-Client-Options-API.patch
```diff
@@ -119,6 +119,10 @@ index b04aaab4f7cb7367d0fbc6268b0db269b55b2d17..3f25e799ba13d8280c644a943ca0d191
 +            return type.getType().cast(getMainHand());
 +        } else if(com.destroystokyo.paper.ClientOption.VIEW_DISTANCE.equals(type)) {
```
Machine-Maker, line 120: might as well fix the missing spaces on the other if statements here. 
And idk if it makes literally any difference, but is there a reason not to use reference equality here?

PR #7273, opened 2022-01-03
Added EntityClimbEvent

zJanny:
This is the first time i have added an event to the API, feedback would be appreciated.

### DIFF
--- a/patches/api/0352-Added-EntityClimbEvent.patch
+++ b/patches/api/0352-Added-EntityClimbEvent.patch
@@ -1,15 +1,15 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Jan Trummer <jan.trummer@hotmail.com>
-Date: Mon, 3 Jan 2022 18:06:18 +0100
+Date: Mon, 3 Jan 2022 18:48:36 +0100
 Subject: [PATCH] Added EntityClimbEvent
 
 
 diff --git a/src/main/java/io/papermc/paper/event/entity/EntityClimbEvent.java b/src/main/java/io/papermc/paper/event/entity/EntityClimbEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..7aa1ee22171d6eb049e80e41cc88d36caa36bd4f
+index 0000000000000000000000000000000000000000..e71e6bc1ff376dc9c575f7b4ea46912d0419a186
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/entity/EntityClimbEvent.java
-@@ -0,0 +1,36 @@
+@@ -0,0 +1,42 @@
 +package io.papermc.paper.event.entity;
 +
 +import org.bukkit.block.Block;
@@ -19,6 +19,9 @@ index 0000000000000000000000000000000000000000..7aa1ee22171d6eb049e80e41cc88d36c
 +import org.bukkit.event.entity.EntityEvent;
 +import org.jetbrains.annotations.NotNull;
 +
++/**
++ * Is called when a Entity goes up a Climbable block
++ */
 +public class EntityClimbEvent extends EntityEvent{
 +
 +    private static final HandlerList HANDLER_LIST = new HandlerList();
@@ -40,9 +43,12 @@ index 0000000000000000000000000000000000000000..7aa1ee22171d6eb049e80e41cc88d36c
 +        return HANDLER_LIST;
 +    }
 +
++    /**
++     *
++     * @return The Block the Entity is climbing on
++     */
 +    @NotNull
-+    public Block getClimbableBlock()
-+    {
++    public Block getClimbableBlock() {
 +        return climbableBlock;
 +    }
 +}
--- a/patches/api/0352-Added-EntityClimbEvent.patch
+++ b/patches/api/0352-Added-EntityClimbEvent.patch
@@ -1,12 +1,12 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Jan Trummer <jan.trummer@hotmail.com>
-Date: Mon, 3 Jan 2022 18:48:36 +0100
+Date: Mon, 3 Jan 2022 21:02:33 +0100
 Subject: [PATCH] Added EntityClimbEvent
 
 
 diff --git a/src/main/java/io/papermc/paper/event/entity/EntityClimbEvent.java b/src/main/java/io/papermc/paper/event/entity/EntityClimbEvent.java
 new file mode 100644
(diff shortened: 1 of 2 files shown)


patches/api/0352-Added-EntityClimbEvent.patch
```diff
@@ -0,0 +1,48 @@
...
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return HANDLER_LIST;
++    }
++
++    @NotNull
++    public Block getClimbableBlock()
++    {
++        return climbableBlock;
++    }
```
Owen1212055: Please stick to the same style, the opening bracket shouldn't be on a separate line.

patches/server/0842-Added-EntityClimbEvent.patch
```diff
@@ -0,0 +1,63 @@
...
+ import org.bukkit.craftbukkit.attribute.CraftAttributeMap;
++import org.bukkit.craftbukkit.block.CraftBlock;
+ import org.bukkit.craftbukkit.event.CraftEventFactory;
+ import org.bukkit.craftbukkit.inventory.CraftItemStack;
+ import org.bukkit.entity.Player;
+@@ -139,7 +142,6 @@ import org.bukkit.event.entity.EntityTeleportEvent;
+ import org.bukkit.event.player.PlayerItemConsumeEvent;
+ // CraftBukkit end
+ 
+-import co.aikar.timings.MinecraftTimings; // Paper
```
Owen1212055: I recommend undoing these import changes and simply using fully qualified imports. It cleans up the diff.

patches/server/0842-Added-EntityClimbEvent.patch
```diff
@@ -0,0 +1,63 @@
...
++                event.callEvent();
++                //Paper end
+                 this.lastClimbablePos = Optional.of(blockposition);
+                 return true;
+             } else if (iblockdata.getBlock() instanceof TrapDoorBlock && this.trapdoorUsableAsLadder(blockposition, iblockdata)) {
+@@ -2838,7 +2846,6 @@ public abstract class LivingEntity extends Entity {
+             if (d2 < 0.0D && !this.getFeetBlockState().is(Blocks.SCAFFOLDING) && this.isSuppressingSlidingDownLadder() && this instanceof net.minecraft.world.entity.player.Player) {
+                 d2 = 0.0D;
+             }
+-
```
Owen1212055, line 15: Unneeded line change

patches/api/0352-Added-EntityClimbEvent.patch
```diff
@@ -0,0 +1,48 @@
...
++package io.papermc.paper.event.entity;
++
++import org.bukkit.block.Block;
++import org.bukkit.entity.LivingEntity;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.entity.EntityEvent;
++import org.jetbrains.annotations.NotNull;
++
++public class EntityClimbEvent extends EntityEvent{
```
Owen1212055, line 25: Additionally, you need to include javadocs. Look at other events for reference.

Owen1212055 (changes requested):

Machine-Maker:
Can you lay out some use cases for this? What comes to my mind as a replacement is using the LivingEntity#isClimbing method in a move event.

zJanny:
> Can you lay out some use cases for this? What comes to my mind as a replacement is using the LivingEntity#isClimbing method in a move event.

If you just want to know if the player is climbing and what he is climbing on, the move event would trigger everytime the player moves instead only when he climbs like here, so it would probably be a bit faster.

electronicboy:
The move event fires anyways, so that's pretty much a none issue at all

I'm all for the climbing state being exposed in the move event and extra info exposed there if needed, events like this are sorely detached from everything else that it provides practically 0 useful context to what is going on so I don't really see much of a use case for this which isn't massively limited or nuanced

zJanny:
I fixed the things now.

A use case for this would be if you made a Plugin that modifies the block the Player is climbing on. The Player move event does not **directly** expose that information.  You could create Visual Effects by changing the block, or maybe create one time usable ladders as a Item in a PvP gamemode. 

(Why does it say "removed the request for review from now "? I clicked "re request review")

patches/server/0842-Added-EntityClimbEvent.patch
```diff
@@ -0,0 +1,44 @@
...
+@@ -1849,6 +1849,12 @@ public abstract class LivingEntity extends Entity {
+             BlockState iblockdata = this.getFeetBlockState();
+ 
+             if (iblockdata.is((Tag) BlockTags.CLIMBABLE)) {
++                //Paper start
++                io.papermc.paper.event.entity.EntityClimbEvent event = null;
++                org.bukkit.craftbukkit.block.CraftBlock craftBlock = org.bukkit.craftbukkit.block.CraftBlock.at(this.level, blockposition);
++                event = new io.papermc.paper.event.entity.EntityClimbEvent(this.getBukkitLivingEntity(), craftBlock);
++                event.callEvent();
++                //Paper end
```
Machine-Maker: If this is going to be a thing, the LivingEntity#onClimbable method is not the right place. CraftLivingEntity calls that when you call LivingEntity#isClimbing, and its called all over the place, in damage areas, collision checks, spigot's activation range. I feel like this event is going to fire seemingly randomly for plugins

Machine-Maker (changes requested):
But you *can* get the block the player is climbing on. Look at what the onClimbable method does, it gets the block at the feed position and checks if its on the CLIMBABLE block tag.

patches/server/0842-Added-EntityClimbEvent.patch
zJanny: It checks if the block the player is on has a climbable tag, so it shouldnt matter what calls the method. Or do i miss something?

Machine-Maker: right, but lets say the event fires, and in the event, something calls LivingEntity#isClimbing, its going to fire again, and then you've got an infinite loop right there.

Or lets say within the span of 1 tick multiple places call that method, and since its the same tick, the player is on the climbable block, so the event has fired twice, or 4 times in the same tick.

Machine-Maker: That method, I think, is supposed to a check *if* the entity is on a climbable thing. If this event is going to be implemented, it should be where some action is taken based on the return value of onClimbable being true

zJanny: Okay thanks, i didnt think about that.

zJanny:
It seems like such  an Event is not needed, so i am gonna close this for now.

PR #7274, opened 2022-01-03
Player steer vehicle event

rgnter:
Event called when packet [`PacketPlayInSteerVehicle`](https://wiki.vg/Protocol#Steer_Vehicle) is sent by player.

### DIFF
--- a/patches/api/0352-steer-vehicle-event.patch
+++ b/patches/api/0352-steer-vehicle-event.patch
@@ -1,0 +1,78 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: rgnter <maros.prejsa1@gmail.com>
+Date: Mon, 3 Jan 2022 17:56:28 +0100
+Subject: [PATCH] steer vehicle event
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerSteerVehicleEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerSteerVehicleEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..dbe465517e3c75d2baa99a34db30d7b4d3604534
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerSteerVehicleEvent.java
+@@ -0,0 +1,66 @@
++package com.destroystokyo.paper.event.player;
++
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when player is steering any vehicle.
++ */
++public class PlayerSteerVehicleEvent extends Event {
++
++    private static final HandlerList handlers
++        = new HandlerList();
++
++    private final float sideways;
++    private final float forward;
++
++    private final boolean isJumping;
++    private final boolean isSneaking;
++
++    public PlayerSteerVehicleEvent(float sideways, float forward, boolean isJumping, boolean isSneaking) {
++        this.sideways = sideways;
++        this.forward = forward;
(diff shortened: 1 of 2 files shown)


electronicboy:
Wrong package, events generally fire before mutation, i also don't see the use of this as it's going to miss a lot of context, etc

rgnter:
Nothing gets mutated, and what do you mean by wrong package?

Owen1212055:
The ``io.papermc.paper`` package should be used instead.

lynxplay:
Your event should live in the io.papernc package. Additionally even without mutation (which idk why this event isn't allowing mutation in the first place) events should still be called before the internal logic happens.

patches/api/0352-steer-vehicle-event.patch
```diff
@@ -0,0 +1,78 @@
...
++package com.destroystokyo.paper.event.player;
++
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when player is steering any vehicle.
++ */
++public class PlayerSteerVehicleEvent extends Event {
```
Owen1212055, line 22: This should extend PlayerEvent, we have no way to actually get the player who steered this.

electronicboy:
> Nothing gets mutated

So, what you're telling me, is that nothing on the server happens at all from this packet?

rgnter:
Alright sorry about that, can fix that. Anyways, if you do not approve of this API extension i am not going to bother. I believe the event should be rather a monitor of action.

patches/server/0842-steer-vehicle-event.patch
```diff
@@ -0,0 +1,29 @@
...
+ import com.mojang.brigadier.ParseResults;
+@@ -494,6 +495,11 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Ser
+     public void handlePlayerInput(ServerboundPlayerInputPacket packet) {
+         PacketUtils.ensureRunningOnSameThread(packet, this, this.player.getLevel());
+         this.player.setPlayerInput(packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown());
++        // paper start - steer vehicle event
++        this.cserver.getPluginManager().callEvent(new PlayerSteerVehicleEvent(
++            packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown()
++        ));
++        // paper end
```
Owen1212055, line 26: Use new PlayerSteerVehicleEvent(...).callEvent();
Use a fully qualified import as well.

rgnter:
The values get used by the server, thats right. But since the event is immutable I dont think its really necessary to call it before server logic(which does not mutate these data)

Owen1212055 (changes requested):
Here is my advice here on some minor nit-picks.

rgnter:
Alright, totally forgot about player in this event. I will modify the patch.

electronicboy:
The values get used by the server, which is thus mutating the state of the server

I don't see the need for random events slapped around the server for knowing that a packet was sent by the client, that's more protocol lib territory; I fail to see a useful application of this which wouldn't be better suited for proper API over different mechanisms exposed in the game

rgnter:
Alright, i won't bother modifying the patch if you do not see use of this event.

MiniDigger:
useful application of this would be riding entities cat, I definitely do see a value here, I had to deal with this stuff myself in the past.
I would name the event in a way that its clear that this is the input from the client only.

yannicklamprecht:
> useful application of this would be riding entities cat, I definitely do see a value here, I had to deal with this stuff myself in the past.
> I would name the event in a way that its clear that this is the input from the client only.

PlayerMovementInputEvent?

PR #7275, opened 2022-01-03
Fix some nullability things

Machine-Maker:
* Removes a no-longer-needed `@UndefinedNullability` annotation since upstream changed behavior.
* getLootTable now honors its method contract and returns null if there isn't a loot table with that key.


Machine-Maker:
Updated & rebased for 1.18.2

electronicboy (approved):

PR #7278, opened 2022-01-04
Add new overload to PersistentDataContainer#has

u9g:
Adds the new overload: `PersistentDataContainer#has(NamespacedKey key)`


patches/api/0352-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
```diff
@@ -0,0 +1,30 @@
...
+      */
+     @NotNull
+     PersistentDataAdapterContext getAdapterContext();
++
++    // Paper start
++    /**
++     * Returns if the persistent metadata provider has metadata registered
++     * matching the provided key.
++     * 
++     * @param key the key to look up in the custom tag map
```
lynxplay: The concept of a custom tag map is pretty unknown to the API.
This should probably be rephrased to something along the lines of
```
@param key the key  for which existence should be checked.
```

patches/server/0842-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
```diff
@@ -0,0 +1,27 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataContainer.java b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataContainer.java
+index 2c59f09a9261a1690951161fd856a5848d9885b7..987c8144738c6e016f5b60b41806bb2a9afbd1a2 100644
+--- a/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataContainer.java
++++ b/src/main/java/org/bukkit/craftbukkit/persistence/CraftPersistentDataContainer.java
+@@ -161,5 +161,15 @@ public class CraftPersistentDataContainer implements PersistentDataContainer {
+     public void clear() {
+         this.customDataTags.clear();
+     }
++
++    public boolean has(@NotNull NamespacedKey key) {
```
lynxplay: I believe we do infact want the `@Override` annotation here.

patches/server/0842-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
```diff
@@ -0,0 +1,27 @@
...
++        Validate.notNull(key, "The provided key for the custom value was null");
++
++        Tag value = this.customDataTags.get(key.toString());
```
lynxplay: This entire section can be simplified to
```java
return this.customDataTags.has(key.toString());
```

lynxplay (changes requested):

patches/server/0842-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
u9g: Done :^)

patches/api/0352-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
u9g: Done :D

patches/server/0842-Patch-Add-new-overload-to-PersistentDataContainer-ha.patch
u9g: Done ;D

Machine-Maker (approved):
Looks good. Is there an existing patch this might fit into? Kinda small for its own patch, but if there isn’t, it’s fine.

lynxplay:
Was looking through it, no patch yet that touches the PDC API, so this might be a good first other changes can append into later.

lynxplay (approved):

kennytv (approved):

JRoy (approved):

PR #7280, opened 2022-01-04
Added fire-tick-delay config option

### LINKED ISSUE #7277: Fire tick speed

Griefing

A config option to set fire tick speed.

Looking for another plugin without success.

### Other

I noticed fires on my server spread very quickly and would like to slow them down. I however do **not** want to disable fire spread entirely.

zJanny:
As requested in [this](https://github.com/PaperMC/Paper/issues/7277) issue

This makes the speed at which fire should spread configurable, the default value is 30. Increasing the value makes the fire spread slower, decreasing it speeds it up.

### DIFF
--- a/patches/server/0842-Added-fire-tick-delay-config-option.patch
+++ b/patches/server/0842-Added-fire-tick-delay-config-option.patch
@@ -19,7 +19,7 @@ index 5a5db15493cd9b83815c36487c2f38cb8ac76f3a..3cabdb76060314bbcbe8d95f685a5ea9
 +    }
  }
 diff --git a/src/main/java/net/minecraft/world/level/block/FireBlock.java b/src/main/java/net/minecraft/world/level/block/FireBlock.java
-index d8e4fda2d501545e5f891bca317e2aa5f9368f47..f1888fdbc34176ddd4cc26f98d85f16012696b5f 100644
+index d8e4fda2d501545e5f891bca317e2aa5f9368f47..c2f0c9f13a8f407f667751f3f30b8a6500e1224c 100644
 --- a/src/main/java/net/minecraft/world/level/block/FireBlock.java
 +++ b/src/main/java/net/minecraft/world/level/block/FireBlock.java
 @@ -164,7 +164,7 @@ public class FireBlock extends BaseFireBlock {
@@ -31,7 +31,7 @@ index d8e4fda2d501545e5f891bca317e2aa5f9368f47..f1888fdbc34176ddd4cc26f98d85f160
          if (world.getGameRules().getBoolean(GameRules.RULE_DOFIRETICK)) {
              if (!state.canSurvive(world, pos)) {
                  fireExtinguished(world, pos); // CraftBukkit - invalid place location
-@@ -362,14 +362,20 @@ public class FireBlock extends BaseFireBlock {
+@@ -362,14 +362,21 @@ public class FireBlock extends BaseFireBlock {
      // Paper start - ItemActionContext param
      public void onPlace(BlockState iblockdata, Level world, BlockPos blockposition, BlockState iblockdata1, boolean flag, UseOnContext itemActionContext) {
          super.onPlace(iblockdata, world, blockposition, iblockdata1, flag, itemActionContext);
@@ -40,6 +40,7 @@ index d8e4fda2d501545e5f891bca317e2aa5f9368f47..f1888fdbc34176ddd4cc26f98d85f160
 -        world.scheduleTick(blockposition, this, getFireTickDelay(world.random));
      }
  
++    @Deprecated
      private static int getFireTickDelay(Random random) {
          return 30 + random.nextInt(10);
      }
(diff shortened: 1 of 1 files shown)


molor:
(add space between `//` and `Paper` in comments)

patches/server/0842-Added-fire-tick-delay-config-option.patch
```diff
@@ -0,0 +1,54 @@
...
+     // Paper start - ItemActionContext param
+     public void onPlace(BlockState iblockdata, Level world, BlockPos blockposition, BlockState iblockdata1, boolean flag, UseOnContext itemActionContext) {
+         super.onPlace(iblockdata, world, blockposition, iblockdata1, flag, itemActionContext);
++        world.scheduleTick(blockposition, this, getFireTickDelay(world.random, world.paperConfig.fireTickDelay));
+         // Paper end
+-        world.scheduleTick(blockposition, this, getFireTickDelay(world.random));
+     }
+ 
+-    private static int getFireTickDelay(Random random) {
+-        return 30 + random.nextInt(10);
```
tahmid-23: it might be better to not just outright delete this method and instead call `getFireTickDelay(30);`

zJanny: I did not delete the method, i just added the "int tickDelayBase" argument to it.
`
    private static int getFireTickDelay(Random random, int tickDelayBase) {
        return tickDelayBase + random.nextInt(10);
     }
`
Or do you mean that i should remove the random from the function and just use the value from the config without adding the random value?

patches/server/0842-Added-fire-tick-delay-config-option.patch
```diff
@@ -0,0 +1,54 @@
...
++++ b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
+@@ -1019,4 +1019,10 @@ public class PaperWorldConfig {
+     private void minBlockLightForMobSpawning() {
+         this.maxBlockLightForMonsterSpawning = getInt("monster-spawn-max-light-level", maxBlockLightForMonsterSpawning);
+     }
++
++
++    public int fireTickDelay = 30;
++    private void fireTickDelay() {
++        fireTickDelay = getInt("fire-tick-delay", 30);
```
Machine-Maker:
```suggestion
+        fireTickDelay = getInt("fire-tick-delay", this.fireTickDelay);
```

patches/server/0842-Added-fire-tick-delay-config-option.patch
```diff
@@ -0,0 +1,54 @@
...
+@@ -362,13 +362,15 @@ public class FireBlock extends BaseFireBlock {
+     // Paper start - ItemActionContext param
+     public void onPlace(BlockState iblockdata, Level world, BlockPos blockposition, BlockState iblockdata1, boolean flag, UseOnContext itemActionContext) {
+         super.onPlace(iblockdata, world, blockposition, iblockdata1, flag, itemActionContext);
++        world.scheduleTick(blockposition, this, getFireTickDelay(world.random, world.paperConfig.fireTickDelay));
+         // Paper end
+-        world.scheduleTick(blockposition, this, getFireTickDelay(world.random));
+     }
+ 
+-    private static int getFireTickDelay(Random random) {
```
Machine-Maker: Instead of changing this method, I would just add a totally separate one that just takes a Level parameter and then do the logic for calculating tick delay in there.

Machine-Maker (changes requested):

patches/server/0842-Added-fire-tick-delay-config-option.patch
```diff
@@ -0,0 +1,55 @@
...
+@@ -362,14 +362,20 @@ public class FireBlock extends BaseFireBlock {
+     // Paper start - ItemActionContext param
+     public void onPlace(BlockState iblockdata, Level world, BlockPos blockposition, BlockState iblockdata1, boolean flag, UseOnContext itemActionContext) {
+         super.onPlace(iblockdata, world, blockposition, iblockdata1, flag, itemActionContext);
++        world.scheduleTick(blockposition, this, getFireUpdateDelay(world));
+         // Paper end
+-        world.scheduleTick(blockposition, this, getFireTickDelay(world.random));
+     }
+ 
+     private static int getFireTickDelay(Random random) {
```
Machine-Maker, line 44: Throw an `@Deprecated` annotation on this, to mark it as shouldnt be used, but to also cause a conflict if it ever changes and the new method needs to reflect changes.

zJanny: This should now be in the newest commit.

(I unresolved the conversation because i am not sure if i was supposed to resolve it)

patches/server/0842-Added-fire-tick-delay-config-option.patch
```diff
@@ -0,0 +1,56 @@
...
+--- a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
+@@ -1019,4 +1019,9 @@ public class PaperWorldConfig {
+     private void minBlockLightForMobSpawning() {
+         this.maxBlockLightForMonsterSpawning = getInt("monster-spawn-max-light-level", maxBlockLightForMonsterSpawning);
+     }
++
++    public int fireTickDelay = 30;
++    private void fireTickDelay() {
++        fireTickDelay = getInt("fire-tick-delay", this.fireTickDelay);
```
electronicboy, line 18: potentially for merger to look into, this should maybe go under game-mechanics rather than the root of the world settings

electronicboy (approved):

Machine-Maker (approved):
I agree it should go under `game-mechanics` (hopefully it won't matter for too long anyways)

PR #7282, opened 2022-01-04
Unregister the enchantments registered by plugins on reload

ShiinaSekiu:
I've meet bugs on slimefun and LiteXpansion with reload command. The LiteXpansion could not register its new class because there is already one old instance in the enchantment list. So it would use the old classes and cause some suspicious errors and memory leak. This commit would unregister the enchantments which is registered by plugins, so it can fix wrong things and avoid confusing errors.
By the way we have repaired the tab completion of the reload command.

### DIFF
--- a/patches/api/0352-Unregister-the-enchantment-registered-by-plugin-when.patch
+++ b/patches/api/0352-Unregister-the-enchantment-registered-by-plugin-when.patch
@@ -1,0 +1,63 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Hai_tun <1346653001@qq.com>
+Date: Wed, 5 Jan 2022 04:12:52 +0800
+Subject: [PATCH] Unregister the enchantment registered by plugin when reload.
+ And optimize the tab completion of the reload command.
+
+
+diff --git a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+index 0c7ba0718de2b93d013968ca0fec34ffd423990f..46d6f62589498b43743f1bdcba4e66e013d331dd 100644
+--- a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
++++ b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+@@ -8,6 +8,7 @@ import org.bukkit.ChatColor;
+ import org.bukkit.command.Command;
+ import org.bukkit.command.CommandSender;
+ import org.jetbrains.annotations.NotNull;
++import com.google.common.collect.Lists; // Paper
+ 
+ public class ReloadCommand extends BukkitCommand {
+     public ReloadCommand(@NotNull String name) {
+@@ -60,6 +61,6 @@ public class ReloadCommand extends BukkitCommand {
+     @NotNull
+     @Override
+     public List<String> tabComplete(@NotNull CommandSender sender, @NotNull String alias, @NotNull String[] args) throws IllegalArgumentException {
+-        return com.google.common.collect.Lists.newArrayList("permissions", "commands"); // Paper
++        return args.length == 1 ? Lists.newArrayList("permissions", "commands") : Collections.emptyList(); // Paper
+     }
+ }
+diff --git a/src/main/java/org/bukkit/enchantments/Enchantment.java b/src/main/java/org/bukkit/enchantments/Enchantment.java
+index 1e1c5a9d9a769018c4604e6e44fc5ed2312981e9..6ca77ad15d077a8e2c8a1e7b5d1cb0ea35e29588 100644
+--- a/src/main/java/org/bukkit/enchantments/Enchantment.java
++++ b/src/main/java/org/bukkit/enchantments/Enchantment.java
+@@ -342,6 +342,17 @@ public abstract class Enchantment implements Keyed, net.kyori.adventure.translat
+      */
+     @NotNull
+     public abstract java.util.Set<org.bukkit.inventory.EquipmentSlot> getActiveSlots();
++
++    /**
++     * Gets the enchantment is registered by server.
++     * If the enchantment is registered by plugin, it should return false and can be unregistered when reload.
++     *
++     * @return true if the enchantment is registered by server.
++     */
++    public boolean isServerEnchantment()
++    {
++        return false;
++    }
+     // Paper end
+ 
+     @Override
+@@ -387,6 +398,13 @@ public abstract class Enchantment implements Keyed, net.kyori.adventure.translat
+         byName.put(enchantment.getName(), enchantment);
+     }
+ 
++    // Paper start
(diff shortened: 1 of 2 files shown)


patches/api/0352-Unregister-the-enchantment-registered-by-plugin-when.patch
```diff
@@ -0,0 +1,63 @@
...
+
+diff --git a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+index 0c7ba0718de2b93d013968ca0fec34ffd423990f..46d6f62589498b43743f1bdcba4e66e013d331dd 100644
+--- a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
++++ b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+@@ -8,6 +8,7 @@ import org.bukkit.ChatColor;
+ import org.bukkit.command.Command;
+ import org.bukkit.command.CommandSender;
+ import org.jetbrains.annotations.NotNull;
++import com.google.common.collect.Lists; // Paper
```
electronicboy, line 16: no

patches/api/0352-Unregister-the-enchantment-registered-by-plugin-when.patch
```diff
@@ -0,0 +1,63 @@
...
+ 
+ public class ReloadCommand extends BukkitCommand {
+     public ReloadCommand(@NotNull String name) {
+@@ -60,6 +61,6 @@ public class ReloadCommand extends BukkitCommand {
+     @NotNull
+     @Override
+     public List<String> tabComplete(@NotNull CommandSender sender, @NotNull String alias, @NotNull String[] args) throws IllegalArgumentException {
+-        return com.google.common.collect.Lists.newArrayList("permissions", "commands"); // Paper
++        return args.length == 1 ? Lists.newArrayList("permissions", "commands") : Collections.emptyList(); // Paper
```
electronicboy, line 25: This belongs in the patch that did this, not a new one

patches/server/0842-Unregister-the-enchantment-registered-by-plugin-when.patch
```diff
@@ -0,0 +1,44 @@
...
+index 11c1eb0e0bc326b28dc0cab16f67c413cc52e98c..7687f86b1c4f9dd8ff441aefccb57ae40c4350a9 100644
+--- a/src/main/java/org/bukkit/craftbukkit/enchantments/CraftEnchantment.java
++++ b/src/main/java/org/bukkit/craftbukkit/enchantments/CraftEnchantment.java
+@@ -223,6 +223,12 @@ public class CraftEnchantment extends Enchantment {
+         return java.util.stream.Stream.of(target.slots).map(org.bukkit.craftbukkit.CraftEquipmentSlot::getSlot).collect(java.util.stream.Collectors.toSet());
+     }
+ 
++    @Override
++    public boolean isServerEnchantment()
++    {
```
electronicboy, line 38: bad code style

electronicboy (changes requested):
I err on stuff like this, hacking into internals is already not supported and when you do unsupported stuff, it's 100% on you to ensure that you clean up after yourself, so, am not exactly sure if this is something I desire, making CraftEnchantment forcefully return true also completely means that this is basically useless if a plugin actually registers stuff properly into the server?

Machine-Maker (changes requested):
I don't think this is something that needs to be in Paper. I agree with cat that this is internals, and registering custom enchants is not something that is supported (aka its not API). The reload command, (I assume you mean bukkit:reload, not minecraft:reload) is also not super well supported, but it does exist and its on plugins to correctly handle it.

astro-angelfish:
> I don't think this is something that needs to be in Paper. I agree with cat that this is internals, and registering custom enchants is not something that is supported (aka its not API). The reload command, (I assume you mean bukkit:reload, not minecraft:reload) is also not super well supported, but it does exist and its on plugins to correctly handle it.

So why don't you just remove ```org.bukkit.enchantments.Enchantments#registerEnchantment``` since it is not supported? And you marked it ```Generally not to be used from within a plugin```. but you didn't say it would cause undefined behavior or deprecate this method or move it to craft bukkit module. I suggest you may need to remove that API or move it to the craft bukkit since it should be used internally.
And for performance, you may need to use a local variable to store the tab complete results instead of creating a new list which would consume a lot of CPU.
And I think you should remove the /reload command since it is no longer supported and can avoid a lot of issues in the future.
Maybe you need to add a developer option to reload all the plugins dynamically in order to reduce time wasted on waiting the server restarts.

Machine-Maker:
> So why don't you just remove org.bukkit.enchantments.Enchantments#registerEnchantment since it is not supported? And you marked it Generally not to be used from within a plugin. but you didn't say it would cause undefined behavior or deprecate this method or move it to craft bukkit module. I suggest you may need to remove that API or move it to the craft bukkit since it should be used internally.


It's used in internals to match up nms enchantments with bukkit ones. That's why the boolean guarding it exists to prevent further registrations. It has to be on the api Enchantment abstract class, because the api class needs to have a byName/byKey to CraftEnchantment instances, which is what is passed as an argument to registerEnchantment from the nms Enchantments class with a new CraftEnchantment instance that wraps the nms Enchantment. This has to happen because the EnchantmentWrapper class has to have some way to get the *actual* CraftEnchantment implementation because the wrapper only redirects to that, it doesn't contain any information about the enchantment.

Machine-Maker:
> And I think you should remove the /reload command since it is no longer supported and can avoid a lot of issues in the future.
Maybe you need to add a developer option to reload all the plugins dynamically in order to reduce time wasted on waiting the server restarts.

There shouldn't be a need for *reloading* the server unless a plugin hasn't provided an option to only reload things that specific plugin needs, like a config file, or lang file, or something like that. As for `reducing time wasted` on server restarts, the paper server, with one plugin for "development" takes me like 4 seconds to startup. Hardly any wasted time.

embeddedt:
> As for reducing time wasted on server restarts, the paper server, with one plugin for "development" takes me like 4 seconds to startup.

Unless there has been some major improvement since 1.16, this is not the case for everyone. For development, I run Linux on a 4th gen i3 with 16GB RAM. The server still takes a good 20 seconds to start at minimum, and at least 5 seconds are spent loading libraries.

By comparison, reloading a plugin is almost instantaneous.

electronicboy:
given our intent as well as upstreams intent to move stuff away from hardcoded classes towards the Mojang registries, this PR is not a behavioural change that is maintainable and much of the mechanism used here is likely going to be removed in the future

I don't see the benefits of trying to maintain this API behavioural change for something which is going to change in the "near" future anyways, it's 100% on plugins messing with internals to clean up after themselves, and this is not a patch that would be retainable in the future, not is it written in a manner which I feel ultimately comfortable with, there are many ways in which plugins mess with this and some would be captured by this, some wouldn't.

Issue #6727, opened 2021-10-05
Boat teleports up randomly when in soul sand bubbly waters.


mannotcool:
### Expected behavior

Boat sways in the bubbly waters and you have full control.

### Observed/Actual behavior

Boat randomly teleports up.

### Steps/models to reproduce

https://imgur.com/a/HSDhU05

### Plugin and Datapack List

Plugins: LuckPerms

And NO datapacks.

### Paper version

This server is running Paper version git-Paper-308 (MC: 1.17.1) (Implementing API version 1.17.1-R
0.1-SNAPSHOT) (Git: e155002)

e-im:
https://bugs.mojang.com/browse/MC-130734

PR #7284, opened 2022-01-05
Boat Wobble API

tahmid-23:
Related to #6727
While it's listed as a Vanilla bug in the linked Mojira issue, the Mojira issue seems more broad and the code seems very deliberate, so this makes it modifiable via event


patches/server/0842-Boat-Wobble-API.patch
```diff
@@ -0,0 +1,70 @@
...
+     }
++
++    @Override
++    public int getWobbleTicks() {
++        return this.getHandle().getBubbleTime();
++    }
++
++    @Override
++    public void setWobbleTicks(int ticks) {
++        com.google.common.base.Preconditions.checkArgument(0 <= ticks, "Ticks cannot be less than 0");
```
NoahvdAa, line 55: Message says less than, check is less than or equal.

electronicboy (commented):
overall looks fine to may, may need some phraseology cleanups, etc, but, that can be done my the merger or whomever

PR #1494, opened 2018-09-27
Optimize redstone algorithm

egg82:
Patch that implements theosib's redstone algorithms.

Tons faster, more stable, etc. Includes a config option to turn it on; off by default.

PR #7291, opened 2022-01-06
Add Alternate Current redstone implementation (and make Eigencraft default)

### LINKED ISSUE #6975: Consider making Eigencraft Redstone default?

The Eigencraft Redstone engine is faster than the stock Redstone implementation, but is not default because it may cause edge incompatibilities with Redstone (though a quick questioning of devs on Discord didn't remember anything in particular). It is most likely that any incompatibilities are strictly for extremely large or technical Redstone builds, and that the vast majority of standard Redstone builds would be unaffected.

I would like to give Eigencraft a second-look for compatibility, and if no issues are found, I don't see any reason why Eigencraft couldn't be made the default Redstone engine. In my mind, we already patch out dupe glitches but give a way out for anyone personally affected; similarly, the Eigencraft Redstone engine would just be a significant performance boost with an opt-out for anyone doing something really niche (like technical builds).

Basically, in `paper.yml`, change the default from:
`    use-faster-eigencraft-redstone: false`
to
`    use-faster-eigencraft-redstone: true`

I don't think this feature request has alternatives.

### Other

_No response_

Titaniumtown:
(Closes https://github.com/PaperMC/Paper/issues/6975)

This PR does 3 main things:
1. (As the title describes) brings SpaceWalker's fantastic redstone implementation. [Alternate Current](https://github.com/SpaceWalkerRS/alternate-current) to Paper.
2. Makes Eigencraft the default redstone algorithm.
3. Adds a new config option in place of `use-faster-eigencraft-redstone`: `redstone-algo`

TODO:
- [X] Migrate previous `use-faster-eigencraft-redstone` config to the new `redstone-algo` option.
- [X] Use enums for config option
- [X] Properly implement config file changes in their respective patches
- [X] PaperDocs PR (https://github.com/PaperMC/PaperDocs/pull/146)

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -298,7 +298,7 @@ index 0000000000000000000000000000000000000000..bee2fa2bfbb61209381f24ed6508d3d1
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..e4368db074da7b5e48b47d41875c1e63b9745c2a
+index 0000000000000000000000000000000000000000..5c7b2850d311e4d54236ba9acce148bca05672fd
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
 @@ -0,0 +1,188 @@
@@ -364,8 +364,8 @@ index 0000000000000000000000000000000000000000..e4368db074da7b5e48b47d41875c1e63
 +        commands = new HashMap<String, Command>();
 +        commands.put("paper", new PaperCommand("paper"));
 +
-+        version = getInt("config-version", 24);
-+        set("config-version", 24);
++        version = getInt("config-version", 25);
++        set("config-version", 25);
 +        readConfig(PaperConfig.class, null);
 +    }
 +
--- a/patches/server/0381-Add-tick-times-API-and-mspt-command.patch
+++ b/patches/server/0381-Add-tick-times-API-and-mspt-command.patch
@@ -75,7 +75,7 @@ index 0000000000000000000000000000000000000000..d0211d4f39f9d6af1d751ac66342b42c
 +    }
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
-index e683e5bf47abe7bd3d2f7e9811a377549308ded4..c20fe23174d8a12bfc5acb4b0e947c6fca5ab897 100644
+index a3c3656ba4e8bb85bfb3186d6284f02f09975b13..6cfaef1886b517c56ed9a96347be6e51efa84d9f 100644
 --- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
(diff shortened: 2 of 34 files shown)


patches/server/0843-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2688 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Simon Gardling <titaniumtown@gmail.com>
```
NoahvdAa: IMO SpaceWalkerRS should be credited as the patch author, with you as the co-author, similar to how egg did it [when porting eigencraft](https://github.com/PaperMC/Paper/pull/1494/files#diff-67d9f633149ac5490b2de8cd95b713ef29bd2749a41e022a40a23babcefe13a1R2-R7) (further discussion [here](https://github.com/PaperMC/Paper/pull/1494#discussion_r221082930))

Titaniumtown: Good point! I'll do that after I finished addressing some concerns that SpaceWalker brought up. Thanks!

electronicboy:
random patches shouldn't change the configuration system used for later patches, also not entirely sure about changing the default, and not exactly fond of magic values either

Titaniumtown:
> random patches shouldn't change the configuration system used for later patches, also not entirely sure about changing the default, and not exactly fond of magic values either

Ok, how should I go about this then? Should I merge the Eigencraft patch and this one together? and how should I ideally manage the config option?

Edit: SpaceWalker suggested using enums, I'll do that in a bit.

patches/server/0843-Add-Alternate-Current-redstone-implementation.patch
Titaniumtown: Fixed!

Titaniumtown:
I've now further improved the config option handling. It now uses a string that's converted into an enum and isn't just a "magic value" (as electronicboy put it) xD. It also now should migrate the old eigencraft config option!

Titaniumtown:
The config changes are now implemented properly. And all the new added classes are under `io.papermc.paper.alternatecurrent`. I'll work on a PaperDocs PR to coincide with this PR.

Edit: the original author of Alternate Current, SpaceWalker, said they preferred if Alternate Current's classes were in their own group. So they are now located in `alternate.current`.

justadudeongithub:
@Titaniumtown thank you for implementing this much awaited patch. Why was the merge blocked?

Titaniumtown:
> @Titaniumtown thank you for implementing this much awaited patch. Why was the merge blocked?

I do not know.

Owen1212055:
See: https://github.com/PaperMC/Paper/issues/6975#issuecomment-1056708772

Basically Eigencraft shouldn't be default.

LaurenceBarnes:
> See: [#6975 (comment)](https://github.com/PaperMC/Paper/issues/6975#issuecomment-1056708772)
> 
> Basically Eigencraft shouldn't be default.

Is there anything else preventing this from being merged? @Titaniumtown looks like the docs PR has to be redone since the changes happened

electronicboy:
needs testing, analysis, etc, etc

dkbay:
Why was this closed? 👀 
I was really looking forward to this.

Titaniumtown:
I'm unmotivated to work on this any longer (and minecraft as a whole), I've moved on.

PR #7294, opened 2022-01-06
fix parsing command function with entity tag selectors

### LINKED ISSUE #7293: Datapack functions using entity_type tags don't work after being loaded for the first time
### Expected behavior

Functions in Minecraft datapacks should be able to reference custom entity_type tags in their commands.
e.g. If a function at `data/mydatapack/functions/kill_arthropods.mcfunction` contains
```
kill @e[type=#mydatapack:arthropods]
```
and an entity tag at `data/mydatapack/tags/entity_types/arthropods.json` contains
```json
{
  "replace": false,
  "values": [
    "minecraft:spider",
    "minecraft:cave_spider",
    "minecraft:silverfish",
    "minecraft:endermite",
    "minecraft:bee"
  ]
}
```
and I run `/function mydatapack:kill_arthropods`, all entities of the respective types should be killed.

### Observed/Actual behavior

No entities match the selector `@e[type=#mydatapack:arthropods]` and no entities are killed.
This is not an issue if the command `/kill @e[type=#mydatapack:arthropods]` is run in-game, or after reloading the datapack for the second time.

### Steps/models to reproduce

1. Start the server.
2. Add [mydatapack.zip](https://github.com/PaperMC/Paper/files/7825235/mydatapack.zip) to the datapacks folder.
3. Run `/minecraft:reload` to enable the datapack.
4. Spawn a spider or other mob in `arthropods.json`.
5. Run `/function mydatapack:kill_arthropods`.
6. Observe that the spider does not die.
7. Repeat steps 3-5. Observe that the spider now dies.
8. Restart the server and repeat steps 4 and 5. Observe that the spider does not die.

### Plugin and Datapack List

No plugins, no datapacks other than the test datapack attached above.

### Paper version

This server is running Paper version git-Paper-134 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 55a1756)
You are 2 version(s) behind
Previous version: git-Paper-398 (MC: 1.17.1)

### Other

My suspicion is that the functions are processed before the entity tags when the datapack is first loaded. I've tested this on a vanilla 1.18.1 server and could not reproduce the issue.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7293

Datapack command functions are parsed before any tags are loaded into the tag container, and that's why in vanilla there's no "parsing" of the entity tag done when the argument is parsed, its done inside the predicate for the EntitySelector which is run on command usage (after all the stuff's been loaded)

### DIFF
--- a/patches/server/0831-Fix-entity-type-tags-suggestions-in-selectors.patch
+++ b/patches/server/0831-Fix-entity-type-tags-suggestions-in-selectors.patch
@@ -113,7 +113,7 @@ index 466d1d8d80df028ff842ce21be198be6a1c77b42..015d01242a9e8e7c6ef5b6bbf1b6d6ad
          this.level = MinMaxBounds.Ints.ANY;
          this.rotX = WrappedMinMaxBounds.ANY;
 diff --git a/src/main/java/net/minecraft/commands/arguments/selector/options/EntitySelectorOptions.java b/src/main/java/net/minecraft/commands/arguments/selector/options/EntitySelectorOptions.java
-index 061181381e4eabad5fa0122f049c4ce05996ffd2..90e023be5c38a038f1c03141ef4325abb25fd615 100644
+index 061181381e4eabad5fa0122f049c4ce05996ffd2..167acc0d469bb11039055fbd54fd8d82187a05f4 100644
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Fixed in the 1.18.2 update.

PR #7299, opened 2022-01-07
Don't tick markers

### LINKED ISSUE #7276: Option to disable marker entity ticking

I'm experimenting with the marker entities added in 1.17.

In my use case, I don't really see a point in ticking them and would love to have the option to disable the ticking, as it drastically reduces tick durations with large amounts of markers, at least if my simple test is valid:

1. Summon a large amount of markers (262,000 in my test)
2. Watch the TPS drop to ~8
3. Move far away until '/paper entity list' recognizes the marker as non-ticking entities
4. Watch the TPS instantly rise to 20 again

Some method to disable marker ticking, either individually, or all together.

None that I could think of, as it isn't really a problem and more like an improvement.

### Other

_No response_

NoahvdAa:
Fixes #7276 by not adding markers to the entity tick list at all and ignoring them
in Spigot's activation range checks. The entity tick list is only used in the tick
and tickPassenger methods, so we can safely not add the markers to it.

To make testing easier, I've attached a world with roughly 100k marker entities
at spawn (on top of the diamond blocks)

[world.zip](https://github.com/PaperMC/Paper/files/7829098/world.zip)

### DIFF
--- a/patches/server/0904-Don-t-tick-markers.patch
+++ b/patches/server/0904-Don-t-tick-markers.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Noah van der Aa <ndvdaa@gmail.com>
+Date: Fri, 7 Jan 2022 11:58:26 +0100
+Subject: [PATCH] Don't tick markers
+
+Fixes https://github.com/PaperMC/Paper/issues/7276 by not adding markers to the entity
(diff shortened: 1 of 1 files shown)


NoahvdAa:
Rebased for 1.18.2.

yannicklamprecht (approved):
Seem fine.

Machine-Maker (approved):

MiniDigger (approved):

kennytv (approved):

PR #4245, opened 2020-08-29
Add PlayerInsertLecternBookEvent

Proximyst:
Surprisingly big diff for such a small event...

Closes #2286.

PR #7305, opened 2022-01-09
Add PlayerInsertLecternBookEvent [1.20 port]

### LINKED ISSUE #2286: PlayerLecternPutEvent
Event that will be fired when player put book on Lectern.

Add event that will be fired when player puts book on lectern.

Canceling right click will also block option to read book by players.

NoahvdAa:
Ports #4245 (fixes #2286) to 1.18 and addresses comments. The diff
has been simplified a lot because the vanilla already passes the player
into the resolveBook and setBook methods.


patches/server/0846-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,50 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
+index 5d0eb0dbfa1f6c2a8e49f8e7b7b0089d7e27764d..124c68ffcc4e9c968dadb6274a2ac5275e67799c 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
+@@ -246,6 +246,17 @@ public class LecternBlockEntity extends BlockEntity implements Clearable, MenuPr
+         if (this.level instanceof ServerLevel && book.is(Items.WRITTEN_BOOK)) {
+             WrittenBookItem.resolveBookComponents(book, this.createCommandSourceStack(player), player);
+         }
++        // Paper start
++        if (player != null && getLevel() != null) {
```
Machine-Maker:
```suggestion
+        if (player != null && this.level != null && player instanceof ServerPlayer serverPlayer) {
```

patches/api/0354-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,73 @@
...
++        this.book = book;
++    }
++
++    @Override
++    public boolean isCancelled() {
++        return cancelled;
++    }
++
++    @NotNull
++    public Lectern getLectern() {
```
Machine-Maker: Probably a javadoc comment saying this is a not a snapshot blockstate, but a live one

patches/api/0354-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,73 @@
...
++        return lectern;
++    }
++
++    @NotNull
++    public ItemStack getBook() {
```
Machine-Maker: comment saying changes will be reflected on the itemstack

patches/server/0846-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,50 @@
...
++            io.papermc.paper.event.player.PlayerInsertLecternBookEvent event = new io.papermc.paper.event.player.PlayerInsertLecternBookEvent(
++                (org.bukkit.entity.Player) player.getBukkitEntity(),
++                (org.bukkit.block.Lectern) getLevel().getWorld().getBlockAt(worldPosition.getX(), worldPosition.getY(), worldPosition.getZ()).getState(false),
++                org.bukkit.craftbukkit.inventory.CraftItemStack.asCraftMirror(book)
++            );
++            if (!event.callEvent()) return ItemStack.EMPTY;
++            return ((org.bukkit.craftbukkit.inventory.CraftItemStack) event.getBook()).handle;
```
Machine-Maker: Since there is a setter for the book itemstack on the event, you cannot cast it to CraftItemStack since there is no guarantee that it will be an instance of it.

patches/server/0846-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,50 @@
...
++++ b/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
+@@ -246,6 +246,17 @@ public class LecternBlockEntity extends BlockEntity implements Clearable, MenuPr
+         if (this.level instanceof ServerLevel && book.is(Items.WRITTEN_BOOK)) {
+             WrittenBookItem.resolveBookComponents(book, this.createCommandSourceStack(player), player);
+         }
++        // Paper start
++        if (player != null && getLevel() != null) {
++            io.papermc.paper.event.player.PlayerInsertLecternBookEvent event = new io.papermc.paper.event.player.PlayerInsertLecternBookEvent(
++                (org.bukkit.entity.Player) player.getBukkitEntity(),
++                (org.bukkit.block.Lectern) getLevel().getWorld().getBlockAt(worldPosition.getX(), worldPosition.getY(), worldPosition.getZ()).getState(false),
```
Machine-Maker:
```suggestion
+                (org.bukkit.block.Lectern) CraftBlock.at(this.level, this.worldPosition).getState(false),
```

patches/server/0846-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,50 @@
...
+--- a/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/LecternBlockEntity.java
+@@ -246,6 +246,17 @@ public class LecternBlockEntity extends BlockEntity implements Clearable, MenuPr
+         if (this.level instanceof ServerLevel && book.is(Items.WRITTEN_BOOK)) {
+             WrittenBookItem.resolveBookComponents(book, this.createCommandSourceStack(player), player);
+         }
++        // Paper start
++        if (player != null && getLevel() != null) {
++            io.papermc.paper.event.player.PlayerInsertLecternBookEvent event = new io.papermc.paper.event.player.PlayerInsertLecternBookEvent(
++                (org.bukkit.entity.Player) player.getBukkitEntity(),
```
Machine-Maker:
```suggestion
+                serverPlayer.getBukkitEntity(),
```

Machine-Maker (changes requested):
I'm not sure the itemstack should be a mirror AND have a setter in the event. I think usually you use mirrors if you can't reassign the nms itemstack where the event is called or if the event doesn't have a setter, it seems odd to have both.

NoahvdAa:
Shoo, rebased for 1.18.2

patches/server/0877-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,50 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/block/LecternBlock.java b/src/main/java/net/minecraft/world/level/block/LecternBlock.java
+index 25ed6dfef2887612a02fcf8884dc8dac4fbd64ff..7d1544f51bed51e625440d197df953a0ed831dae 100644
+--- a/src/main/java/net/minecraft/world/level/block/LecternBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/LecternBlock.java
+@@ -134,6 +134,7 @@ public class LecternBlock extends BaseEntityBlock {
+         if (!(Boolean) state.getValue(LecternBlock.HAS_BOOK)) {
+             if (!world.isClientSide) {
+                 LecternBlock.placeBook(player, world, pos, state, stack);
++                return ((LecternBlockEntity) world.getBlockEntity(pos)).hasBook(); // Paper
```
Machine-Maker: I would do an instance check on the result of world.getBlockEntity(pos) here as something else might cancel the block placement, or even in the PlayerInsertLecternBookEvent they change the block and remove the block entity. Like all over Mojang changes blocks, and then they check if the tile entity was created, they don't assume it was. Look at the first couple lines of placeBook

Machine-Maker (changes requested):

NoahvdAa:
Rebased for 1.19, comment has been addressed.

patches/api/0387-Add-PlayerInsertLecternBookEvent.patch
```diff
@@ -0,0 +1,93 @@
...
++    public void setCancelled(boolean cancelled) {
++        this.cancelled = cancelled;
++    }
++
++    /**
++     * Sets the itemstack to insert into the lectern.
++     *
++     * @param book the book to insert
++     */
++    public void setBook(@Nullable ItemStack book) {
```
Machine-Maker: I think we are trying to move away from nullable ItemStack in new stuff.

Owen1212055 (approved):
But see MM's comment

lynxplay:
Rebased.

For now, removed BlockState clone from API and inserted itemstack defensive clones.

Machine-Maker (approved):

NoahvdAa:
GH (obviously) won't let me review my own PR, but the rebase changes look good

PR #7315, opened 2022-01-11
Port SendSignEvent event to Paper

electronicboy:
after many years on the fabled todo list, a port of the SendSignEvent from EMC

TODO: needs testing

### DIFF
--- a/patches/api/0356-SendSignEvent.patch
+++ b/patches/api/0356-SendSignEvent.patch
@@ -1,0 +1,81 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aikar <aikar@aikar.co>
+Date: Sun, 4 Dec 2016 22:06:10 -0500
+Subject: [PATCH] SendSignEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/block/SendSignEvent.java b/src/main/java/io/papermc/paper/event/block/SendSignEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..8cc0e1eb7a1a851b3c5ea3e002149b0d96bc5972
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/block/SendSignEvent.java
+@@ -0,0 +1,69 @@
++/*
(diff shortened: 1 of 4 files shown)


patches/api/0356-SendSignEvent.patch
```diff
@@ -0,0 +1,81 @@
...
++     * @return the lines of the sign
++    **/
++    @NotNull
++    public abstract List<Component> getLines();
++
++    /**
++     * @param line the line in which to modify
++     * @param component the component in which to set
++     */
++    public abstract void setLine(int line,  @Nullable net.kyori.adventure.text.Component component);
```
imDaniX: Adventure's methods naming (`lines`, `line`)? Also there's two whitespaces before `@Nullable`.

Machine-Maker:
Something feels off about this event. Like why don't we add events anytime a block entity has its state sent to a client? (And maybe we should do that, but then lets do it in a generic way that is easily extendable to other block entities).

If we shouldn't, this feels like protocol API stuff, just modifying packets and what not.

electronicboy:
This was long on aikars todo list to port over and was something I'd modified for my own fork, and have used it myself when I had my own setup as it was really useful in terms of magical signs, e.g. populating data from the PDC and such instead of tryna mangle state on the sign

signs where just a stupidly common area to want to do this stuff in a supported manner, without having to rely on packet libraries; a more generic approach might be nice in some regards, just there didn't really seem to be much demand there

Owen1212055:
I'm a bit iffy on this myself as well.

This seems like something better off doing with a protocol library plugin.

electronicboy:
The nature of this being in the chunk packet makes handing this with a protocol library a general nightmare, hence why aikar has had this in his own server for years;

Machine-Maker:
> The nature of this being in the chunk packet makes handing this with a protocol library a general nightmare, hence why aikar has had this in his own server for years;

But this also applies to every other block entity. they are all tied up in chunk packets. I just don't think it's good to add an event for one of them and then, when someone comes along with a PR adding 20 more events for each of the other block entities, say "nah".

electronicboy:
Closing, this has generally been rejected in favour of a block entity send event

PR #7317, opened 2022-01-11
Update ASM EventExecutor generator patch to respect event handler return types. Fixes #7311

### LINKED ISSUE #7311: Bug in the ASM EventExecutor bytecode generator
### Stack trace

```
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    com/destroystokyo/paper/event/executor/asm/generated/GeneratedEventExecutor1.execute(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Event;)V @11: pop
  Reason:
    Type long_2nd (current frame, stack[1]) is not assignable to category1 type
  Current Frame:
    bci: @11
    flags: { }
    locals: { 'com/destroystokyo/paper/event/executor/asm/generated/GeneratedEventExecutor1', 'org/bukkit/event/Listener', 'org/bukkit/event/Event' }
    stack: { long, long_2nd }
  Bytecode:
    0000000: 2bc0 000e 2cc0 0010 b600 1457 b1       

	at java.lang.Class.getDeclaredConstructors0(Native Method) ~[?:?]
	at java.lang.Class.privateGetDeclaredConstructors(Class.java:3373) ~[?:?]
	at java.lang.Class.getConstructor0(Class.java:3578) ~[?:?]
	at java.lang.Class.newInstance(Class.java:626) ~[?:?]
	at org.bukkit.plugin.EventExecutor.create(EventExecutor.java:65) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.java.JavaPluginLoader.createRegisteredListeners(JavaPluginLoader.java:336) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.SimplePluginManager.registerEvents(SimplePluginManager.java:660) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at com.janboerman.repro.invalidgeneratedeventexecutor.Demo.onEnable(Demo.java:9) ~[InvalidGeneratedEventExecutor-1.0-SNAPSHOT.jar:?]
	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:264) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:370) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:500) ~[paper-api-1.18.1-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.craftbukkit.v1_18_R1.CraftServer.enablePlugin(CraftServer.java:564) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at org.bukkit.craftbukkit.v1_18_R1.CraftServer.enablePlugins(CraftServer.java:478) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at net.minecraft.server.MinecraftServer.loadWorld0(MinecraftServer.java:727) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at net.minecraft.server.MinecraftServer.loadLevel(MinecraftServer.java:503) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:313) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1202) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:317) ~[paper-1.18.1.jar:git-Paper-"87d8ef9"]
```

### Plugin and Datapack List

just my repro plugin: https://github.com/Jannyboy11/GeneratedEventExecutorBug

Install a plugin that registers a Listener with a method annotated with `@EventHandler` that returns either a double or long.

### Paper version

This server is running Paper version git-Paper-"87d8ef9" (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 87d8ef9)

### Other

The JVM spec specifies that `long` and `double` (unboxed) are so called 'category 2 types' which occupy twice as much space on the operand stack. See: https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-2.html#jvms-2.11.1
Paper's EventExecutor generator however only check whether the return type of the event handler is not void, and generates a POP instruction in that case. However, if the type is `long` "J" or `double` "D", then a POP2 instruction should be generated instead. (https://github.com/PaperMC/Paper/blob/master/patches/api/0026-Use-ASM-for-event-executors.patch#L156)

This bug does not occur on CraftBukkit/Spigot since they simply use reflection to invoke the method.

I can see two ways to fix this issue:
1. Generate a pop2 instruction if the return type of the event handler is `double` or `long`.
2. Require that event handlers always return void (and don't generate the pop instruction at all) and throw an exeption otherwise, because eventhandlers with a return value are nonsensical anyway.

What's more:
ASM's [Type](https://asm.ow2.io/javadoc/org/objectweb/asm/Type.html#getSize()) api has a method which returns the size on the operand stack of a type.

Jannyboy11:
The 0-case is intentionally left out, because no extra instruction needs to be generated if the return type is void.

I verified that [my demo plugin](https://github.com/Jannyboy11/GeneratedEventExecutorBug) now loads correctly, the VerifyError no longer occurs, and I installed a plugin with a 'normal' event handler as well to verify that nothing broke by this change.

### DIFF
--- a/patches/api/0026-Use-ASM-for-event-executors.patch
+++ b/patches/api/0026-Use-ASM-for-event-executors.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Use ASM for event executors.
 Uses method handles for private or static methods.
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 0e8e827b5a05ac9423386eaea4188d132cc7f954..70d6503745942ea8637146369aeab8f9357c42d5 100644
+index 80fdd05dd593455ca89b66636ed30f1d9facf4ed..d8d459561cc75935136f8f9888ff27b45ad98f9e 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -38,6 +38,9 @@ dependencies {
@@ -118,10 +118,10 @@ index 0000000000000000000000000000000000000000..c83672427324bd068ed52916f700b684
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/event/executor/asm/ASMEventExecutorGenerator.java b/src/main/java/com/destroystokyo/paper/event/executor/asm/ASMEventExecutorGenerator.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..b6e7d8ee8d903ebf975d60bec0e08603d9a49fdb
+index 0000000000000000000000000000000000000000..c5db5d06b28a418efa327dd2f3d34fe265455daa
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/executor/asm/ASMEventExecutorGenerator.java
-@@ -0,0 +1,47 @@
+@@ -0,0 +1,48 @@
 +package com.destroystokyo.paper.event.executor.asm;
 +
 +import java.lang.reflect.Method;
@@ -147,14 +147,15 @@ index 0000000000000000000000000000000000000000..b6e7d8ee8d903ebf975d60bec0e08603
 +        methodGenerator.returnValue();
 +        methodGenerator.endMethod();
 +        // Generate the execute method
-+        methodGenerator = new GeneratorAdapter(writer.visitMethod(ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Event;)V", null, null), ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Listener;)V");;
++        methodGenerator = new GeneratorAdapter(writer.visitMethod(ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Event;)V", null, null), ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Listener;)V");
 +        methodGenerator.loadArg(0);
 +        methodGenerator.checkCast(Type.getType(m.getDeclaringClass()));
 +        methodGenerator.loadArg(1);
 +        methodGenerator.checkCast(Type.getType(m.getParameterTypes()[0]));
 +        methodGenerator.visitMethodInsn(m.getDeclaringClass().isInterface() ? INVOKEINTERFACE : INVOKEVIRTUAL, Type.getInternalName(m.getDeclaringClass()), m.getName(), Type.getMethodDescriptor(m), m.getDeclaringClass().isInterface());
-+        if (m.getReturnType() != void.class) {
-+            methodGenerator.pop();
++        switch (Type.getType(m.getReturnType()).getSize()) {
++            case 1: methodGenerator.pop(); break;
++            case 2: methodGenerator.pop2(); break;
 +        }
 +        methodGenerator.returnValue();
 +        methodGenerator.endMethod();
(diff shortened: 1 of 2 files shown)


electronicboy (approved):
given that this is pseudo supported, I am somewhat inclined to approve this

I do think that we should maybe consider deprecating this, however :/

electronicboy (approved):
given that this is pseudo supported, I am somewhat inclined to approve this

I do think that we should maybe consider deprecating this, however :/

jpenilla (approved):
I agree with cat that this should be unsupported, especially given zmls comments on the original issue. Approving given we merge https://github.com/PaperMC/Paper/pull/7372 or something similar to make it clear that this isn't supported and that future issues like this won't be fixed

patches/api/0026-Use-ASM-for-event-executors.patch
```diff
@@ -147,14 +147,15 @@ index 0000000000000000000000000000000000000000..b6e7d8ee8d903ebf975d60bec0e08603
...
-+        methodGenerator = new GeneratorAdapter(writer.visitMethod(ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Event;)V", null, null), ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Listener;)V");;
++        methodGenerator = new GeneratorAdapter(writer.visitMethod(ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Event;)V", null, null), ACC_PUBLIC, "execute", "(Lorg/bukkit/event/Listener;Lorg/bukkit/event/Listener;)V");
 +        methodGenerator.loadArg(0);
 +        methodGenerator.checkCast(Type.getType(m.getDeclaringClass()));
 +        methodGenerator.loadArg(1);
 +        methodGenerator.checkCast(Type.getType(m.getParameterTypes()[0]));
 +        methodGenerator.visitMethodInsn(m.getDeclaringClass().isInterface() ? INVOKEINTERFACE : INVOKEVIRTUAL, Type.getInternalName(m.getDeclaringClass()), m.getName(), Type.getMethodDescriptor(m), m.getDeclaringClass().isInterface());
-+        if (m.getReturnType() != void.class) {
-+            methodGenerator.pop();
++        switch (Type.getType(m.getReturnType()).getSize()) {
```
jpenilla, line 158: Should add a comment clarifying for people unfamiliar with ASM that void is length 0, and that this code is here only to handle the unsupported cases of non-void return types

jpenilla (dismissed):

patches/api/0026-Use-ASM-for-event-executors.patch
Jannyboy11: I hope these comments will suffice.

PR #7318, opened 2022-01-11
Add poses to the api

LoaiDev:
This pull requests adds the ability to change entiy's pose through the api.
since by default the entity's pose isn't persisted between loading and unloading, any changes to the pose outside the api will not be persisted to retain the correct behaviour however any pose changed through the api will be persisted.
it also adds the ability to make the pose fixed (can not be changed by vanilla ways) such as sleeping, riding, etc.
This is my first time contributing so please point out any mistakes or better ways to further improve on this (since I don't know much about the server's code)
I know of some annotations issues that I will try to fix when I can
_This has not been tested for players and to be honest I don't think poses will ever work well for players_


patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+ public abstract class CraftEntity implements org.bukkit.entity.Entity {
+     private static PermissibleBase perm;
+@@ -1124,6 +1125,23 @@ public abstract class CraftEntity implements org.bukkit.entity.Entity {
+         return Pose.values()[this.getHandle().getPose().ordinal()];
+     }
+ 
++    // Paper start - Pose api
++    @Override
++    public void setPose(Pose pose) {
++        this.setPose(pose, false);
```
kennytv: Move this into the interface as a default method

patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+     public void setPose(net.minecraft.world.entity.Pose pose) {
++        if (this.fixedPose) {
++            return;
++        }
++        setPose(pose, false, false);
++    }
++
++    public void setPose(net.minecraft.world.entity.Pose pose, boolean persist, boolean fixed) {
++        this.fixedPose = fixed;
++        this.persistPose = persist;
```
kennytv: Paper end comment here

patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+ 
+     public void setOrigin(@javax.annotation.Nonnull Location location) {
+         this.origin = location.toVector();
+@@ -612,7 +614,17 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+ 
+     public void onClientRemoval() {}
+ 
++    // Paper start - Pose api
+     public void setPose(net.minecraft.world.entity.Pose pose) {
++        if (this.fixedPose) {
```
kennytv: move the start comment down a line

patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+ import net.minecraft.world.level.block.SoundType;
+-import net.minecraft.world.level.block.state.BlockBehaviour;
+ import net.minecraft.world.level.block.state.BlockState;
+ import net.minecraft.world.level.block.state.properties.BlockStateProperties;
+ import net.minecraft.world.level.border.WorldBorder;
+@@ -326,6 +325,9 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+     private org.bukkit.util.Vector origin;
+     @javax.annotation.Nullable
+     private UUID originWorld;
++    @Nullable
```
kennytv: ?

patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index 8ca6388ddd834c4f0e94450c8a6ce6b5e73a5f60..2750ca3c4ea174e2ec78403b9776b65ce179e808 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -92,7 +92,6 @@ import net.minecraft.world.level.block.Mirror;
+ import net.minecraft.world.level.block.RenderShape;
+ import net.minecraft.world.level.block.Rotation;
+ import net.minecraft.world.level.block.SoundType;
+-import net.minecraft.world.level.block.state.BlockBehaviour;
```
kennytv: Don't remove imports

patches/server/0846-Pose-Api.patch
```diff
@@ -0,0 +1,121 @@
...
+             fromNetherPortal = nbt.getBoolean("Paper.FromNetherPortal");
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
+index 69223b13e894d86d9529f2ef8b60a08a1f7a9267..49dbbc47f725dc4a4dfd5736d669afc04c28efae 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
+@@ -181,6 +181,7 @@ import org.bukkit.plugin.Plugin;
+ import org.bukkit.util.BoundingBox;
+ import org.bukkit.util.NumberConversions;
+ import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
```
kennytv: inline the import where it's used

kennytv (commented):
Missing a bunch of Paper comments in the Entity class

LoaiDev:
how can I edit the same patch?

patches/server/0846-Pose-Api.patch
LoaiDev: probably my ide auto optimize imports I will fix it

patches/server/0846-Pose-Api.patch
LoaiDev: I was originally storing the actual pose but then decided to use a boolean instead

kennytv:
Assuming that's your latest patch; move into the paper api/server dir, edit, commit --amend the changes and rebuild patches (see contributing guide)

LoaiDev:
fixed the problems and managed to mess up the repository in the process
should be fixed though

WcaleNieWolny:
Did you tested this pull request? When I was working on doing that there was a server side check for setting players pose (at method tick). Please be aware of that.

patches/server/0850-Pose-Api.patch
```diff
@@ -0,0 +1,100 @@
...
++        if (this.fixedPose) {
++            return;
++        }
++        setPose(pose, false, false);
++    }
++
++    public void setPose(net.minecraft.world.entity.Pose pose, boolean persist, boolean fixed) {
++        this.fixedPose = fixed;
++        this.persistPose = persist;
++        // Paper end
```
Owen1212055, line 34: Additionally here, I wouldn't update the method signature.
Instead, on CraftEntity#setPose directly access and set the fields yourself.

So it would look like
```java
public void setPose(net.minecraft.world.entity.Pose pose) {
       if (this.fixedPose) { 
           return;
        }
        // CraftBukkit start
        if (pose == this.getPose()) {
            return;
        }
        this.level.getCraftServer().getPluginManager().callEvent(new EntityPoseChangeEvent(this.getBukkitEntity(), Pose.values()[pose.ordinal()]));
        // CraftBukkit end
        this.entityData.set(Entity.DATA_POSE, pose);
    }
```

(api)

```java
     @Override
      public void setPose(Pose pose, boolean fixed) {
          CraftEntity handle = this.getHandle();
          handle.persist = true;
          handle.fixedPose = false;
          handle.setPose(net.minecraft.world.entity.Pose.values()[pose.ordinal()));
          handle.fixedPose = fixed;
     }
```

This change is a bit iffy, but the idea is you want to try to minimize the diff. (Although it might not matter that much)

patches/server/0850-Pose-Api.patch
```diff
@@ -0,0 +1,100 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index 8ca6388ddd834c4f0e94450c8a6ce6b5e73a5f60..e1e3790f831a2f4aca953c3e7d79a9b0cff01c84 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -326,6 +326,8 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+     private org.bukkit.util.Vector origin;
+     @javax.annotation.Nullable
+     private UUID originWorld;
++    private boolean persistPose;
++    private boolean fixedPose;
```
Owen1212055, line 16: Make these public, and remove the getter.

Owen1212055 (changes requested):
Currently poses aren't serialized at all with the entity as well, and it seems a little strange to me how this api method would cause it to be saved.

I personally think it shouldn't save with the entity, but I would perhaps see about adding another boolean to control if it's saved or not. But I would like to see what other people think on that.

WcaleNieWolny:
> Currently poses aren't serialized at all with the entity as well, and it seems a little strange to me how this api method would cause it to be saved.
> 
> I personally think it shouldn't save with the entity, but I would perhaps see about adding another boolean to control if it's saved or not. But I would like to see what other people think on that.

In my personal opinion we should not save pose with the entity. If someone needs that he will do it on his own with setPose api

yannicklamprecht:
Is per player weather and time saved? I don't think so. The upcoming upstream api for per player worldborder doesn't save the border.

So I'm with both of you. Persistence is a plugin task.

LoaiDev:
> Currently poses aren't serialized at all with the entity as well, and it seems a little strange to me how this api method would cause it to be saved.
> 
> I personally think it shouldn't save with the entity, but I would perhaps see about adding another boolean to control if it's saved or not. But I would like to see what other people think on that.

for the public property part, none of the entity properties related to any kind of api is made as a public one and therefore I decided to keep this when creating this and I dont believe it will make any kind of problem. as for the saving part as I mentioned in the first comment on this pr.
> since by default the entity's pose isn't persisted between loading and unloading, any changes to the pose outside the api will not be persisted to retain the correct behaviour however any pose changed through the api will be persisted.
it also adds the ability to make the pose fixed (can not be changed by vanilla ways) such as sleeping, riding, etc.

meaning the entity's pose will never be persisted unless some plugin sets it.

I mean we could always expose the "persistPose" to the api but I decided that any pose done through the api should be persisted anyways. most (if not all) of the entity api doesn't get reset when the entity is reloaded so that's why I have chosen to do this.
If you still think that these points have to be changed I will do them but in my opinion I think it doesn't need changing

bluelhf:
Although this has little repository activity, it seems there's a disproportionate amount of demand for this from users of the API: For example GSit, a plugin which provides this very feature using six separate NMS implementations, has gotten an average of over 600 downloads a day since its release.

PR #7321, opened 2022-01-12
Add PlayerInventorySlotChangeEvent

dawon:
Useful event for detecting user acquired/lost an item in any way (e.g. custom advancement). 

Otherwise, it is needed to listen for several events and I am not sure whether it is possible to listen for example for /give.

Originally I wanted it to be InventoryEvent, but since there is no InventoryView and it would have to be null, I thought PlayerEvent would be better.

### DIFF
--- a/patches/api/0399-Add-PlayerInventorySlotChangeEvent.patch
+++ b/patches/api/0399-Add-PlayerInventorySlotChangeEvent.patch
@@ -1,0 +1,96 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jakub Zacek <dawon@dawon.eu>
+Date: Sun, 24 Apr 2022 22:56:31 +0200
+Subject: [PATCH] Add PlayerInventorySlotChangeEvent
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerInventorySlotChangeEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerInventorySlotChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..f106d2c6fb349c57ed00f1d7f96efacb1ade3959
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerInventorySlotChangeEvent.java
+@@ -0,0 +1,84 @@
++package com.destroystokyo.paper.event.player;
++
++import org.bukkit.entity.Player;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a slot contents change in a player's inventory.
++ */
++public class PlayerInventorySlotChangeEvent extends PlayerEvent {
++    private static final HandlerList handlers = new HandlerList();
++    private final int slot;
++    private final ItemStack oldItemStack;
++    private final ItemStack newItemStack;
++    private boolean triggerAdvancements = true;
++
++    public PlayerInventorySlotChangeEvent(@NotNull Player player, int slot, @NotNull ItemStack oldItemStack, @NotNull ItemStack newItemStack) {
++        super(player);
++        this.slot = slot;
++        this.oldItemStack = oldItemStack;
++        this.newItemStack = newItemStack;
++    }
++
++    /**
++     * The slot number that was changed.
++     *
++     * @return The slot number.
++     */
++    public int getSlot() {
++        return slot;
++    }
++
++    /**
++     * Clone of ItemStack that was in the slot before the change.
++     *
++     * @return The old ItemStack in the slot.
++     */
++    @NotNull
++    public ItemStack getOldItemStack() {
++        return oldItemStack;
++    }
++
++    /**
++     * Clone of ItemStack that is in the slot after the change.
++     *
++     * @return The new ItemStack in the slot.
++     */
++    @NotNull
++    public ItemStack getNewItemStack() {
++        return newItemStack;
++    }
++
++    /**
++     * Gets whether the slot change advancements will be triggered.
++     *
++     * @return Whether the slot change advancements will be triggered.
++     */
++    public boolean isTriggerAdvancements() {
++        return triggerAdvancements;
++    }
++
++    /**
++     * Sets whether the slot change advancements will be triggered.
++     *
++     * @param triggerAdvancements Whether the slot change advancements will be triggered.
++     */
++    public void setTriggerAdvancements(boolean triggerAdvancements) {
++        this.triggerAdvancements = triggerAdvancements;
++    }
++
++    @NotNull
++    @Override
++    public HandlerList getHandlers() {
++        return handlers;
++    }
++
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return handlers;
++    }
++}
--- a/patches/server/0925-Add-PlayerInventorySlotChangeEvent.patch
+++ b/patches/server/0925-Add-PlayerInventorySlotChangeEvent.patch
@@ -1,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jakub Zacek <dawon@dawon.eu>
+Date: Sun, 24 Apr 2022 22:56:59 +0200
+Subject: [PATCH] Add PlayerInventorySlotChangeEvent
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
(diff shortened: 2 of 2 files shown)


dawon:
Rebased.

Machine-Maker:
If this event is going to be a thing, which I’m not 100% sold on yet, it should be in `AbstractContainerMenu#triggerSlotListeners` so you can call it with the old and new stacks. Also, the new stack should be a mirror I think, not a copy.

dawon:
Well I know the Paper team generally don't like events, that can be replaced by other events, but I would say this is not the case. If I wanted to only check Player obtained an item, I would have to listen for (and I am sure I forgot about some things):

- Item pickup
- menu click (would have to think also about Bundles)
- menu drag
- bucket fill
- bucket empty (obtaining empty bucket)
- item consume (empty bottle/bowl)
- entity interact (not sure if milking a cow calls bucket fill, also filling bucket of fish)
- block interact (lodestone compass?)
- signing/editing a book
- cloning a block (creative middle clicking on a block)
- /give
- plugins adding items into player's inventory

Most of the stuff can be detected by other events, but I am not sure how to detect last three mentioned. And even if I was, as you can see, for such simple task - detect whether/when player got an item, I would have to check that many events. This event would solve it.

Alternative would be to skim through all players inventories every tick and check for the item, which is solution I really don't like and I try to avoid doing stuff that often...

About your remarks: I also originally though putting it in there, but then the event would be bound to Inventory and not to player, and also I didn't see player instance in there, I guess maybe checking if Inventory is player's inventory and then casting it?

I did not know that mirroring was a option, found this in other event, will definitely fix it. I guess for old itemstack I should use copy though, right?

Machine-Maker:
> About your remarks: I also originally though putting it in there, but then the event would be bound to Inventory and not to player, and also I didn't see player instance in there, I guess maybe checking if Inventory is player's inventory and then casting it?

Ah yeah, in that case, I would add a default method to the ContainerListener interface that takes an additional ItemStack argument that delegates to the existing slotChanged method. Then in ServerPlayer, override that method to fire the event.

EDIT: Also, this event can be cancellable, but make clear in the event docs, that you are not cancelling the changing of the itemstacks, but logic that runs after that, such as advancement triggers for inventory changing.

Phoenix616:
> Also, this event can be cancellable, but make clear in the event docs, that you are not cancelling the changing of the itemstacks, but logic that runs after that, such as advancement triggers for inventory changing.

I'm not sure if that is the right way to go about that if the action that gets stopped isn't the one from the name of the event (the slot actually changing) I feel like if it was made cancellable that it should also stop the action that caused the event to be called (the slot changing) from occurring, at least that's how pretty much all events work.

If this should really include a toggle for the actions after the change then I feel like this should be a separate toggle (e.g. "triggerAdvements" or something along that line)

Machine-Maker:
> If this should really include a toggle for the actions after the change then I feel like this should be a separate toggle (e.g. "triggerAdvements" or something along that line)

Yeah, that might be better.

dawon:
> If this should really include a toggle for the actions after the change then I feel like this should be a separate toggle (e.g. "triggerAdvements" or something along that line)

Ok, will do that...

> Ah yeah, in that case, I would add a default method to the ContainerListener interface that takes an additional ItemStack argument that delegates to the existing slotChanged method. Then in ServerPlayer, override that method to fire the event.

If I add a default method with an additional parameter and override that, then the original method will not be overridden and thus it will be uncompilable. I might override both, and use the new only for the event, but then I can't cancel the other one.

I can make the old one default calling the new one with null as the argument. Would that be sufficient?

Edit: Another option would be to just change the method, it is only referenced twice, when called and overridden. But that might break stuff in the future, I guess...

Machine-Maker:
> If I add a default method with an additional parameter and override that, then the original method will not be overridden and thus it will be uncompilable. I might override both, and use the new only for the event, but then I can't cancel the other one.

The original one will still be there, just add an additional override where the instance of ContainerListener is created. Then call that new method from AbstractContainerMenu. that should work fine.

EDIT: see https://paste.gg/p/anonymous/55a51399e00f4de0afc768de37a7f1c1

dawon:
Rebased and fixed remarks...

Edit: thanks for the paste, noticed it after I committed, so hopefully it is fine this way... :D

dawon:
Rebased

dawon:
Rebased

Katzen48:
I would really like to see this happening.
For I project, I work on, I need an event, that detects any kind of inventory change.
No matter, if by a pickup or /give command.

Me and my team use that also for making the command blocks and data packs communicate with the plugin, that is responsible for the story quests.
Currently, our only solution is tracking item pickups and also /give command executions.
When a /give command happens, the hash codes for the slots have to be saved and compared against the now possible updated inventory, one tick after.

Other solutions, like the one in use, are not very performant but the only way without crashing maintainability.

electronicboy:
Really far from fond of events which allow mutation of the results from other events with 0 awareness towards other events

This breaks the general bukkit contract that API won't trigger events outside of niche cases, and introduces many risks for things tripping over one another and doesn't allow for many concerns of context, etc

Katzen48:
What about some immutable event?
Otherwise, there is no good way, to catch all changes to the player's inventory?

electronicboy:
only way for it to be immutable would be for all getter methods to return a clone

dawon:
AFAIK this PR does not allow mutation of the stacks, or does it? It is a while when I did it, but I still miss sine event in this matter...

electronicboy:
it does because you're mirroring, i.e. wrapping, the internal stack

dawon:
And if I cloned it? I really need just the information, don't need to modify the item.

Katzen48:
> And if I cloned it? I really need just the information, don't need to modify the item.

Same for me

dawon:
Rebased and changed it so both old and new stacks are copied.

dawon:
> If this event is going to be a thing, which I’m not 100% sold on yet, it should be in `AbstractContainerMenu#triggerSlotListeners` so you can call it with the old and new stacks. Also, the new stack should be a mirror I think, not a copy.

Just skimmed through the PR, I was changing it to mirror the ItemStack based on Machine's remarks. But yet again, I really don't care whether it will be a mirror, a copy or the ItemStack will be missing at all.

Katzen48:
Any updates on this?

dawon:
> Any updates on this?

Well, now we need to wait for reviews. But since there is quite a lot of PRs and the Paper team works on Paper in their free time, we need to be patient...

patches/server/0902-Add-PlayerInventorySlotChangeEvent.patch
```diff
@@ -0,0 +1,61 @@
...
+ 
+                 }
+             }
++            // Paper start
++            @Override
++            public void slotChanged(AbstractContainerMenu handler, int slotId, ItemStack oldStack, ItemStack stack) {
++                Slot slot = handler.getSlot(slotId);
++                if (!(slot instanceof ResultSlot)) {
++                    if (slot.container == ServerPlayer.this.getInventory()) {
++                        com.destroystokyo.paper.event.player.PlayerInventorySlotChangeEvent event = new com.destroystokyo.paper.event.player.PlayerInventorySlotChangeEvent(ServerPlayer.this.getBukkitEntity(), slotId, CraftItemStack.asBukkitCopy(oldStack), CraftItemStack.asBukkitCopy(stack));
```
electronicboy: Not 100% if needed, but maybe debating if this needs to check if there are event handlers registered for this event before actually doing stuff like cloning stacks, etc, otherwise, looks ok

electronicboy (approved):

patches/server/0902-Add-PlayerInventorySlotChangeEvent.patch
dawon: Good point, will rebase and tackle that once I come home, thank you!

dawon: Rebased and fixed.

dawon:
rebased for 1.19

dawon:
rebased for 1.19.2

dawon:
rebased on master

patches/server/0947-Add-PlayerInventorySlotChangeEvent.patch
```diff
@@ -0,0 +1,65 @@
...
+                 }
+             }
++            // Paper start
++            @Override
++            public void slotChanged(AbstractContainerMenu handler, int slotId, ItemStack oldStack, ItemStack stack) {
++                Slot slot = handler.getSlot(slotId);
++                if (!(slot instanceof ResultSlot)) {
++                    if (slot.container == ServerPlayer.this.getInventory()) {
++                        if (com.destroystokyo.paper.event.player.PlayerInventorySlotChangeEvent.getHandlerList().getRegisteredListeners().length == 0) {
++                            slotChanged(handler, slotId, stack);
```
Lulu13022002: This should only trigger the advancement trigger. No need to duplicate the whole logic here.

dawon: Good catch, thank you, fixed!

dawon:
rebased onto master

dawon:
rebased

PR #7324, opened 2022-01-13
Optimize explosion and made async this

Dysaido:
I haven't test this yet. But I think this works well.


kennytv:
Not interested in merging something that's basically just "I made the thing async, guys" with no testing or any further explanation given, especially when explosions are already optimized in Paper and aren't really a point of contention

PR #7329, opened 2022-01-14
Fix/improve Affects Spawning API

Machine-Maker:
The affects spawning API had some method names, parameter names, and copied old decompiler code in it, this fixes that (none of which changes behavior), but what does, is that this was missing the Skeleton Traps and Zombie reinforcements part of "affects spawning". This also adds checks if the player affects spawning before creating either of those.


Machine-Maker:
Rebased for 1.18.2

kennytv (approved):

PR #7330, opened 2022-01-14
Undeprecate Bed's tile entity interface

Machine-Maker:
I think the ~~quickest~~simplest way currently to get the DyeColor of a bed, is
```java
if (block.getState() instanceof Bed bed) {
	DyeColor color = bed.getDyeColor()
}
```
which is just backed by a size 16 switch statement in CraftBed, but the method is still totally valid.

In addition, Beds **do still** have an associated tile entity, and I think it doesn't make sense to have it deprecated if it does in fact still exist. Take a look at the `Conduit` interface, which exists but has declared methods, and just extends `TileState`.

### DIFF
--- a/patches/api/0054-Fix-upstream-javadocs.patch
+++ b/patches/api/0054-Fix-upstream-javadocs.patch
@@ -46,6 +46,36 @@ index 5e4e6e83ac6b52493cb285561425bed53ffff2b6..2bfbb0ce71c8c5f8bd9bbd908488831b
       *
       * @return an array containing all previous players
       */
+diff --git a/src/main/java/org/bukkit/block/Bed.java b/src/main/java/org/bukkit/block/Bed.java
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased and updated

patches/api/0054-Fix-upstream-javadocs.patch
```diff
@@ -46,6 +46,35 @@ index acd69a5d946974e0b50439a98712750698768ce5..0053327f3df85a568befc280b53a0eb3
...
++    @Override
++    @org.jetbrains.annotations.NotNull org.bukkit.DyeColor getColor();
++
++    /**
++     * <b>Unsupported</b>
++     *
++     * @throws UnsupportedOperationException not supported, set the block type
++     */
++    @Override
++    @org.jetbrains.annotations.Contract("_ -> fail")
```
kennytv, line 74: Should add deprecated here since IJ has no special hightlighting for this

kennytv (approved):
Else good

PR #7332, opened 2022-01-14
Added and implemented PlayerLootDropEvent.

NeonOverflow:
As a result of how the `PlayerDeathEvent` was implemented, when a player dies their items are simply dropped using the `world.dropItem()` method. This makes it very annoying for plugin developers to track down the Item entities dropped when a player dies. This patch implements a `PlayerLootDropEvent`, making tracking down the aforementioned Item entities simple. Furthermore, the item entities are marked as being thrown by the dead player.


patches/server/0852-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,57 @@
...
+@@ -907,14 +908,32 @@ public class CraftEventFactory {
+         victim.expToDrop = event.getDroppedExp();
+         victim.newExp = event.getNewExp();
+ 
++        // Paper start - implement player loot drop event + set item thrower as dead player
++        List<Item> items = new ArrayList<Item>();
++
+         for (org.bukkit.inventory.ItemStack stack : event.getDrops()) {
+             if (stack == null || stack.getType() == Material.AIR) continue;
+-
```
kennytv: don't remove the empty line

patches/server/0852-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,57 @@
...
+Item entities dropped on player death are marked as being thrown by the player that died.
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index 35cc150adf51f79e2fccef8b094c90554aafbee2..7621847b6b777e6c66ac49fc76cb8569a9dcd3ff 100644
+--- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -1,5 +1,6 @@
+ package org.bukkit.craftbukkit.event;
+ 
++import com.destroystokyo.paper.event.player.PlayerLootDropEvent;
```
kennytv: Inline the import

patches/server/0852-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++        // Paper end
++
++        return event;
++    }
++
++    // Paper start - callPlayerLootDropEvent
++    public static PlayerLootDropEvent handlePlayerLootDropEvent(Player player, List<Item> items) {
++        PlayerLootDropEvent event = new PlayerLootDropEvent(player, items);
++        Bukkit.getPluginManager().callEvent(event);
++        if(event.isCancelled()) {
```
kennytv: No need to return the event in the method, so just do
```suggestion
+        if (!new PlayerLootDropEvent(player, items).callEvent()) {
```

patches/server/0852-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++            items.forEach(Item::remove);
```
kennytv: Only remove an item if it hasn't been removed already

patches/api/0357-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,69 @@
...
++        this.items = items;
++    }
++
++    /**
++     * Gets the {@link org.bukkit.entity.Item}s dropped.
++     *
++     * @return the {@link org.bukkit.entity.Item}s dropped.
++     */
++    public List<Item> getItems() {
++        return items;
```
kennytv: Make the list immutable and make that clear in the docs

kennytv (changes requested):

NeonOverflow:
done

kennytv:
The static getHandlerList method has to be readded, and getItems is missing a NotNull annotation

NeonOverflow:
> The static getHandlerList method has to be readded

Sorry about that, I have no clue how that got removed. I'm still getting used to Paper's build system. 
I'll add the method and annotation when I get home.

NeonOverflow:
Alright. I triple-checked everything this time and it all looks good.

NeonOverflow:
For whatever reason it still says "1 change requested" on my end. Is there a requested change I failed to address or is that just showing up because my changes haven't been accepted?

NoahvdAa:
It's showing up because kenny hasn't re-reviewed (or dismissed his old review) yet.

patches/api/0357-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,73 @@
...
+making it easy for developers to track down the Item entities dropped when a player dies. Furthermore,
+Item entities dropped on player death are marked as being thrown by the player that died.
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerLootDropEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerLootDropEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..739ef6e6a06df07f4031da8cca3255948ab63a8c
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerLootDropEvent.java
+@@ -0,0 +1,56 @@
++package com.destroystokyo.paper.event.player;
```
NoahvdAa: This should be in the `io.papermc.paper` package instead, apart from that, LGTM.

NeonOverflow:
Ah alright. Thanks. I'm a total noob in terms of handling git and pull requests lol.

kennytv (approved):

patches/server/0854-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,52 @@
...
+ 
++        // Paper start - implement player loot drop event + set item thrower as dead player
++        List<Item> items = new ArrayList<Item>();
++
+         for (org.bukkit.inventory.ItemStack stack : event.getDrops()) {
+             if (stack == null || stack.getType() == Material.AIR) continue;
+ 
+-            world.dropItem(entity.getLocation(), stack);
++            Item item = world.dropItem(entity.getLocation(), stack);
++            item.setThrower(victim.getUUID());
```
Machine-Maker: This is a behavioral change. In vanilla, item's dropped by players do not have their thrower set. They specifically don't. The thrower, in vanilla, is used for the `thrown_item_picked_up_by_entity` advancement trigger, and items dropped on death are not counted as "thrown by the player dying", only items intentionally dropped.

Machine-Maker (changes requested):

Machine-Maker:
I've also noticed that the way item drops are spawned into the world for player deaths it not at all in line with how vanilla does it, the pickup delay is wrong, there isnt the "random" movement delta added to spread out items.

I think <https://paste.gg/p/anonymous/3fa1fe0ca0984f549cb272412131cb4d> is a better solution to this whole thing, no need for a new event, just listen for PlayerDropItemEvent and check if the player is dead.

NeonOverflow:
> I've also noticed that the way item drops are spawned into the world for player deaths it not at all in line with how vanilla does it, the pickup delay is wrong, there isnt the "random" movement delta added to spread out items.

It does the same thing upstream. Guess it's a CraftBukkit issue. 

> I think https://paste.gg/p/anonymous/3fa1fe0ca0984f549cb272412131cb4d is a better solution to this whole thing, no need for a new event, just listen for PlayerDropItemEvent and check if the player is dead.

The only issue with that is that you'd have to link each Item to a certain player's death and then bundle them together manually in order to collect all the items a player had dropped as opposed to them just being bundled into an ArrayList to begin with. I also think it's kinda weird for the API to have a method under PlayerDeathEvent for getting all of the player's loot in the form of an ItemStack whilst not having something similar for Items dropped on death.

electronicboy (approved):

patches/api/0369-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,73 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: NeonOverflow <nneonoverflow@protonmail.com>
+Date: Thu, 13 Jan 2022 22:58:50 -0500
+Subject: [PATCH] Added and implemented PlayerLootDropEvent.
+
+As a result of how the PlayerDeathEvent was implemented, when a player dies their items are simply dropped
+through the "world.dropItem()" method. This makes it very annoying for plugin developers to ultimately
+track down the Item entities dropped when a player dies. This patch implements a "PlayerLootDropEvent",
+making it easy for developers to track down the Item entities dropped when a player dies. Furthermore,
+Item entities dropped on player death are marked as being thrown by the player that died.
```
Machine-Maker: the marked as being thrown was removed

patches/server/0872-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,72 @@
...
+         victim.newExp = event.getNewExp();
+ 
++        // Paper start - implement player loot drop event
++        List<Item> items = new ArrayList<Item>();
++
+         for (org.bukkit.inventory.ItemStack stack : event.getDrops()) {
+             if (stack == null || stack.getType() == Material.AIR) continue;
+ 
+-            world.dropItem(entity.getLocation(), stack);
++            Item item = (Item) victim.drop(CraftItemStack.toNMSStack(stack), true, false);
```
Machine-Maker: Did you actually test this? You can't just cast an nms ItemEntity to the bukkit Item interface.

patches/api/0369-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,73 @@
...
++
++    public PlayerLootDropEvent(@NotNull Player player, @NotNull List<Item> items) {
++        super(player);
++        this.items = items;
++    }
++
++    /**
++     * Gets the {@link org.bukkit.entity.Item}s dropped.
++     *
++     * @return an immutable list of the {@link org.bukkit.entity.Item}s dropped.
```
Machine-Maker: The list is an unmodifiable view, not immutable.

patches/api/0369-Added-and-implemented-PlayerLootDropEvent.patch
```diff
@@ -0,0 +1,73 @@
...
++
++import org.bukkit.entity.Item;
++import org.bukkit.entity.Player;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Is called when a {@link org.bukkit.entity.Player} drops their {@link org.bukkit.entity.Item}s on death.
```
Machine-Maker: Unlike most events this actually happens **after** all the item entities are created, it might be good to mention that here, that while it all happens in the same tick, and right after each other, technically the event is being fired after the item entities exist in the world.

Machine-Maker (changes requested):

patches/server/0872-Added-and-implemented-PlayerLootDropEvent.patch
NeonOverflow: That was.... a pretty dumb mistake. If I had to guess I probably didn't build probably and ended up using an earlier version before that change was made when I tried testing it. My bad.

PR #7336, opened 2022-01-14
Better Potion API

### LINKED ISSUE #5366: Improvements to PotionType + PotionData
Currently retrieving potion effects is quite limiting and a hassle to deal with. It seems like what is currently in place does not support the modern-day potions that we have. For example, there are potion types that now can contain multiple effects. (See, turtle master potion)

It is also not very easy to get the duration/amplifier of potions without dealing with "extended" and "upgraded".
image

No

- An easy way to get List\<PotionEffect\> from PotionData. This would take into account extended + upgraded and give the corresponding effects. So potions like turtle master would return two effects.
- Adding support for multiple potion effect types from PotionType. Currently certain potions can have multiple effects on them by default, for example the turtle master potion. The system currently in place does not support this, as it only displays the slowness effect instead.
image

There really aren't any very easy alternatives that work nicely. You would need to store all of this yourself, so instead, I just use NMS to deal with this. However, it would be great if this could be officially supported because I feel that others may benefit from this as well.

**Additional context**
N/A

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/5366

Re-does bukkit's very outdated `PotionType` and `PotionData` classes into a single `Potion` class.

### DIFF
--- a/patches/api/0428-Better-Potion-API.patch
+++ b/patches/api/0428-Better-Potion-API.patch
@@ -1,0 +1,343 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Mon, 16 Aug 2021 16:59:30 -0700
+Subject: [PATCH] Better Potion API
+
+
+diff --git a/src/main/java/io/papermc/paper/potion/Potion.java b/src/main/java/io/papermc/paper/potion/Potion.java
+new file mode 100644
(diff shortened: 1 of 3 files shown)


Machine-Maker:
Removed `PaperRegistry` from this

Owen1212055 (approved):
Tested with arrows, compared returned API values, and compared items between the old/new API.

Seems to work fine, this API is a lot nicer than what we currently have.
Nice work!

Machine-Maker:
Moved this over to the PaperRegistry/Reference system.

Owen1212055:
Partially resolved by upstream. Will be fully resolved once enums are yeeted.

PR #7341, opened 2022-01-14
Sync offhand slot in menus

### LINKED ISSUE #4869: Swapping Items to your off-hand and inventories
When creating a GUI menu in a plugin and canceling all the necessary events required to prevent a player from taking items from the menu I would have expected that to also block players being able to press F on an item within the menu and getting that item (a copy of that item) placed in their off-hand.

You can press F to put the current item over your cursor in your off-hand regardless of the events being canceled, In Survival mode, this item is a "ghost" item and will vanish once the players inventory updates, however, in Creative mode this item is real and can be used. 

How to reproduce: Install or make a plugin that has an inventory menu, open the menu and press F while your cursor is over one of the items within the menu. 

Plugin List:
- InventoryTestMenu (made to test this)

[17:15:37] [Server thread/INFO]: This server is running Paper version git-Paper-319 (MC: 1.16.4) (Implementing API version 1.16.4-R0.1-SNAPSHOT)
[17:15:37] [Server thread/INFO]: Previous version: git-Paper-220 (MC: 1.16.3)
[17:15:37] [Server thread/INFO]: You are running the latest version

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/4869

Menus don't add slots for the offhand, so on sendAllDataToRemote calls the
offhand slot isn't sent. This is not correct because you *can* put stuff into the offhand
by pressing the offhand swap item

### DIFF
--- a/patches/server/0950-Sync-offhand-slot-in-menus.patch
+++ b/patches/server/0950-Sync-offhand-slot-in-menus.patch
@@ -1,0 +1,51 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Fri, 14 Jan 2022 10:20:40 -0800
+Subject: [PATCH] Sync offhand slot in menus
+
+Menus don't add slots for the offhand, so on sendAllDataToRemote calls the
+offhand slot isn't sent. This is not correct because you *can* put stuff into the offhand
+by pressing the offhand swap item
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 3b122f521c166253f20d233c0fcebdede6660be5..9c9b4d7e0637348a94befce9377fdb69c1239694 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -287,6 +287,13 @@ public class ServerPlayer extends Player {
+ 
+             }
+ 
++            // Paper start
++            @Override
++            public void sendOffHandSlotChange() {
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

patches/server/0852-Sync-offhand-slot-in-menus.patch
```diff
@@ -0,0 +1,51 @@
...
+                 ServerPlayer.this.connection.send(new ClientboundContainerSetSlotPacket(handler.containerId, handler.incrementStateId(), slot, stack));
+diff --git a/src/main/java/net/minecraft/world/inventory/AbstractContainerMenu.java b/src/main/java/net/minecraft/world/inventory/AbstractContainerMenu.java
+index 1f4d3a48553a467bcbd4799735d1950c9c2dbe23..e8c6aa9d645bbbf4aa80e08b83b5c63ff123b260 100644
+--- a/src/main/java/net/minecraft/world/inventory/AbstractContainerMenu.java
++++ b/src/main/java/net/minecraft/world/inventory/AbstractContainerMenu.java
+@@ -194,6 +194,7 @@ public abstract class AbstractContainerMenu {
+             this.synchronizer.sendInitialData(this, this.remoteSlots, this.remoteCarried, this.remoteDataSlots.toIntArray());
+         }
+ 
++        this.synchronizer.sendOffHandSlotChange(); // Paper - update player's offhand since the offhand slot is not added to the slots for menus but can be changed by swapping from a menu slot
```
scarfacered: This needs to be inside the null check above it.

Machine-Maker: Probably right. While I'm not sure its possible for an NPE to be thrown there under normal circumstances, there prob are some plugins that might.

Machine-Maker:
Rebased for 1.18.2

bendgk:
This is a legitimate bug.

Easy reproduction case would be to create a listener for `InventoryClickEvent` and cancel the event when the click type is `SWAP_OFFHAND`.

Canceling the event should set the offhand to be null on the client, but instead, a desync happens.

Issue #5344, opened 2021-03-11
[future] [comments] configuration migration 


electronicboy:
Papers configuration file has generally gotten pretty huge, 270 lines is what's currently sitting in my paper.yml, and then 168 lines in spigot.yml,
paper.yml includes several options in which overrides settings in bukkit.yml (saving), or spigot.yml (e.g. item despawn), which really creates a clusterf*** for trying to configure servers

In the future, when we diverge from spigot, there will be no real valid reason to stick to having 3 separate configuration files, it creates a lot of confusion and in general is pretty messy, the existing layout of these files is also severely lacking in general organization, in part due to how hard it is to justify moving stuff around, every "move" inside of the yaml files would generally involve a few lines of code to facilitate that and maintain

In the future, it's worth looking into separating the configuration files into a more "friendly" format, multiple config files with consistent intents, I think the line count of these files is somewhat concerning if it comes in regards to throwing it all in a single file vs separating it out into a more "clean" modular layout, would also give us the time to modify the layout of these configurations to be more useful, we would however need to consider a migration plan, which doesn't sound all too pretty :/

the consideration of this is, do we consider also switching the format of these configurations, hocon, for example, offers some useful features such as the ability to retain comments, which would make it much easier to add documentation to stuff that could be used for automatic generation of the documentation on the site, something which is often overlooked when adding new options;

LaurenceBarnes:
everyone liked that

kashike:
👍 for cleaning up configurations
👎 for moving from YAML - YAML is what 95% of Bukkit plugins use, and is what people are most familiar with in this community.

electronicboy:
Yea, hocon was just an idea I tossed around, not entirely too fond of it given the community, and apparently configurate may be able to help us here in the future if we wanted; being able to throw comments in the config classes, even if not persisted to the files, would maaaybe potentially be super nice, not 100% on how "nice" it would look however all in all, but would potentially provide means to auto-generate some config docs

BillyGalbreath:
Idk how I feel about the modular approach, tbh. Like, it sounds like it would be a good thing but I can't help but imagine myself having to go to 15 different config files in 15 different locations to change 15 different things.

I'm definitely in favor of getting rid of the redundanct bukkit/spigot ymls, though. Simply squashing it all down to a single yml should fix those confusing redundant options. (maybe even getting rid of the vanilla server.properties, too, but not too concerned with that one)

[...]

PR #7344, opened 2022-01-14
Only overworld spawn chunks get a ticket.

NeonOverflow:
Only overworld spawn chunks are intended to have a ticket added to them. A mistake must've been made at some point upstream.


electronicboy:
this should really be dealt with in the config rather than hard-coding it, but, will probs have to wait on #5344

Machine-Maker:
Yeah, vanilla only adds spawn tickets for the overworld, but the change to have it exist for all worlds might be intended due to craftbukkit splitting each world up into its own "thing". You can disable it per-world in `paper.yml` with the world setting `keep-spawn-loaded`

NeonOverflow:
> this should really be dealt with in the config rather than hard-coding it, but, will probs have to wait on #5344

Fair enough. When #5344 is complete I'm more than willing to code in a configuration option in another commit.

> but the change to have it exist for all worlds might be intended due to craftbukkit splitting each world up into its own "thing".

That's true. This has a pretty high likelihood of confusing server admins and players who expect only overworld spawn chunks to stay loaded though (I know from experience).

PR #7350, opened 2022-01-15
Add unsupported field to disable username validation

Camotoy:
Allows for bypassing the username filter without globally disabling it.


MiniDigger:
I think this should be part of the OG patch, no point in adding a two liner like this

Camotoy:
Done.

kashike (approved):

PR #7353, opened 2022-01-16
Fix xp reward for baby zombies

### LINKED ISSUE #7352: Biased baby zombie dropped exp cancelling EntityDeathEvent
### Expected behavior

Default dropped experience _(around 12 exp for a baby zombie + bonuses from items)_ even if its death event has been cancelled several times earlier.

### Observed/Actual behavior

Dropped exp progressively growing _(*2.5, *2.5, etc.)_ cancelling death.

### Steps/models to reproduce

- Create a small plugin that cancels EntityDeathEvent and logs out ``EntityDeathEvent#getDroppedExp``.
- The more zombies you will kill, the bigger the exp will be, until reaching ``Integer.MAX_VALUE``. This is a problem if the mob is later allowed to die, it drops a ton of exp.

Assumptions:
This event uses an internal method ``LivingEntity#getExpReward``, that uses ``LivingEntity#getExperienceReward``.

```java
 EntityDeathEvent event = new EntityDeathEvent(entity, drops, victim.getExpReward());
```

```java
    // From Zombie class.
    @Override
    protected int getExperienceReward(Player player) {
        if (this.isBaby()) {
            this.xpReward = (int) ((float) this.xpReward * 2.5F);
        }

        return super.getExperienceReward(player);
    }
```
As you can see, if the zombie is a baby, the exp is multiplied by 2.5, default behavior suits with this code, however, cancelling death doesn't reset this value. This is in my opinion a specific issue with baby zombies.

### Plugin and Datapack List

```
Plugins (1): DomiTest
```

```
There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
```

### Paper version

```
This server is running Paper version git-Paper-152 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 78999b5)
Previous version: git-Paper-151 (MC: 1.18.1)
```

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7352

The field that tracks the xpReward was not getting reset if the death was cancelled so this resets it after each call to `Zombie#getExperienceReward(Player)`

### DIFF
--- a/patches/server/0852-Fix-xp-reward-for-baby-zombies.patch
+++ b/patches/server/0852-Fix-xp-reward-for-baby-zombies.patch
@@ -1,0 +1,32 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 16 Jan 2022 10:34:02 -0800
+Subject: [PATCH] Fix xp reward for baby zombies
+
+The field that tracks the xpReward was not
+getting reset if the death was cancelled
+so this resets it after each call to
+Zombie#getExperienceReward
+
+diff --git a/src/main/java/net/minecraft/world/entity/monster/Zombie.java b/src/main/java/net/minecraft/world/entity/monster/Zombie.java
+index e72e9b748b3f3e34baddf01366c703efba50c67c..35f0203d260c11b729c30e6241316fda4b70bfd7 100644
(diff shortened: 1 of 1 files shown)


kennytv (approved):
Could also use or at least hint at Enemy.XP_REWARD_MEDIUM (but is just fine as is)

kennytv:
or actually, better would be to store the previous value in a local var and use that so no extra diff is needed

Machine-Maker:
Yeah, refactored it to a "previous" local var

PR #7355, opened 2022-01-16
Fix entity armor not showing on death animation

Machine-Maker:
Was previously clearing all hand/armor slots on deaths when only slots that had their items dropped should be cleared so as to not update the client to the change in entity armor status.


molor:
omg I thought that it's a client render bug..

electronicboy (approved):

PR #7366, opened 2022-01-18
Add World Time API

DeJayDev:
Adds an enum with the constant phases of the daylight cycle, to be used as reference or with setTime and setFullTime.

I considered adding helpers to World for this, but wasn't sure if it'd be overstepping or doing too much. Decided it was best to get conversation started about it first.

These times are sourced from https://minecraft.fandom.com/wiki/Daylight_cycle#Daytime, and the overlapping time enums are on purpose.


Machine-Maker (changes requested):
I don't think this fits into the API. If there was a counterpart enum, or something that mapped a time to a name in the server, this might fit. 

Just taking some values from the wiki isn't a great way to design an API. It's easy enough for plugins to create constants for such things.

DeJayDev:
As mentioned in paper-dev, I could modify these to the times in the /time command but plenty of actions are triggered by certain times of day that plugins may wish to have control over. If it still feels inappropriate feel free to close.

PR #7367, opened 2022-01-18
1.18 misc performance dev branch

Spottedleaf:
Includes:
 - Further optimised collisions. I went pretty hard, there might be collision problems.
 - Reintroduction of player chunk loader patch. Helps rate limit chunk sending to clients.

 - Reversion of Mojang's internal chunk system changes to manage what chunks are ticking. They added a system on top of the existing one just to make chunks outside simulation distance not tick, obviously a complete waste. This should help bring certain aspects of chunk/entity/block ticking back under control, at least compared to 1.17. It also brings in some fixes, notably, mob spawning now _correctly_ occurs only in simulation distance.
Expected behavioral changes are that plugins adding tickets to load chunks will now cause them to tick, as they did in 1.17. Considering that's how the API was designed to work, and has been done so since 1.14, it will remain that way.
API for simulation distance is not implemented yet. This means any plugins trying to work with dynamically changing ticking distance _will not_ work. setViewDistance, unlike 1.17, will follow vanilla behavior and set a "no ticking" view distance.
As it stands, simulation distance is kept vanilla and will stay that way - this means that you do not have to change configs for this branch.

As always, dev builds are generally experimental and are expected to have issues. Keep those issues on this PR _only_.

The jar can be found here:
https://cdn.discordapp.com/attachments/876902758366736457/932942650657632266/paper-paperclip-1.18.1-R0.1-SNAPSHOT-reobf.jar
A comment will be made per update.

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -221,6 +221,8 @@ public net.minecraft.world.entity.animal.Fox isDefending()Z
 public net.minecraft.world.entity.animal.Fox setDefending(Z)V
 public net.minecraft.world.entity.animal.Fox isFaceplanted()Z
 public net.minecraft.world.entity.animal.Fox setFaceplanted(Z)V
+public net.minecraft.world.entity.animal.Panda getEatCounter()I
+public net.minecraft.world.entity.animal.Panda setEatCounter(I)V
 
 # Cook speed multipler API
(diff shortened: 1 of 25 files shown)


Spottedleaf:
wrong branch

PR #7368, opened 2022-01-18
1.18 misc performance dev branch

Spottedleaf:
This time but the right branch.

Includes:
 - Further optimised collisions. I went pretty hard, there might be collision problems.
 - Reintroduction of player chunk loader patch. Helps rate limit chunk sending to clients.

 - Reversion of Mojang's internal chunk system changes to manage what chunks are ticking. They added a system on top of the existing one just to make chunks outside simulation distance not tick, obviously a complete waste. This should help bring certain aspects of chunk/entity/block ticking back under control, at least compared to 1.17. It also brings in some fixes, notably, mob spawning now _correctly_ occurs only in simulation distance.
Expected behavioral changes are that plugins adding tickets to load chunks will now cause them to tick, as they did in 1.17. Considering that's how the API was designed to work, and has been done so since 1.14, it will remain that way.
API for simulation distance is not implemented yet. This means any plugins trying to work with dynamically changing ticking distance _will not_ work. setViewDistance, unlike 1.17, will follow vanilla behavior and set a "no ticking" view distance.
As it stands, simulation distance is kept vanilla and will stay that way - this means that you do not have to change configs for this branch.

As always, dev builds are generally experimental and are expected to have issues. Keep those issues on this PR _only_.

The jar can be found here:
https://cdn.discordapp.com/attachments/876902758366736457/935408751576502322/paper-paperclip-1.18.1-R0.1-SNAPSHOT-reobf.jar
A comment will be made per update.


ImVexed:
Not an issue, but a thank you for the work you guys are doing. Our server was having major issues with people falling through the world for 20+ seconds, and some with bad internet even getting disconnected before 1.18.1 came out. 

This dev build completely solved it for us, players instantly have ground to walk on and no longer fall through the world *at all*, and I've received no reports of any bugs observed. 

Thanks for understanding chunk loading and the optimizations around it better than Mojang!

Spottedleaf:
Updated to include async catchers for ChunkHolder block/light update

Spottedleaf:
Updated to fix an issue with citizens

EterNityCH:
Thank you for your hard work leafmento

patches/server/0854-fixup-Highly-optimise-single-and-multi-AABB-VoxelSha.patch
```diff
@@ -0,0 +1,720 @@
...
+ 
+     public static final double COLLISION_EPSILON = 1.0E-7;
+ 
++    public static final long KNOWN_EMPTY_BLOCK = 0b00; // known to always have voxelshape of empty
++    public static final long KNOWN_FULL_BLOCK = 0b01; // known to always have voxelshape of full cube
++    public static final long KNOWN_UNKNOWN_BLOCK = 0b10; // must read the actual block state for info
++    public static final long KNOWN_SPECIAL_BLOCK = 0b11; // caller must check this block for special collisions
++
++    public static final List<String> BAD_PLUGINS = java.util.Arrays.asList(
++            "FastAsyncWorldEdit"
```
rymiel: Why is this plugin specifically singled out here?

kashike: Because it does stupid things to chunks.

Valyrox:
Hello @Spottedleaf,
Would you be interested in the result of different profilers (YourKit, Spark, Flare, etc.), for information/improvements, of a server that I am running with your patches, and with around 220 players?

patches/server/0854-fixup-Highly-optimise-single-and-multi-AABB-VoxelSha.patch
SirYwell: To add an actual explanation: FAWE does not use the very inefficient methods to set single block states in chunk sections. Instead, ticking blocks are updated all at once per chunk section. Similar to every plugin that e.g. uses the approach 4 describe in https://www.spigotmc.org/threads/methods-for-changing-massive-amount-of-blocks-up-to-14m-blocks-s.395868/, this will not call the methods required by this patch.

Spottedleaf:
I've tightened some logic around notifying the client of their current chunk center, not sure if it fixes anything

dalandis:
We experienced high ping after switching to 1.18.
I don't know if this will help or not. But for now, turning off anti-xray really helped.
Anti-xray paper, I mean. With engine-mode: 2

The problem manifested itself when flying on elytras or when moving to the nether.

The amount of traffic from the server greatly decreased after that. Before disabling anti-xray, many people who had weak computers or routers had problems with ping.

Anti-xray enable
anti-xray_enable
Anti-xray disable
anti-xray_disable

We also tried the kernel from this PR. And tried this plugin https://papermc.io/forums/t/paperlagfix-fixes-high-ping-bug-on-1-18-unofficial/9913
This did not give any noticeable effect.

I hope this will be helpful.

amaranth:
> We experienced high ping after switching to 1.18. I don't know if this will help or not. But for now, turning off anti-xray really helped. Anti-xray paper, I mean. With engine-mode: 2
> 
> The problem manifested itself when flying on elytras or when moving to the nether.
> 
> The amount of traffic from the server greatly decreased after that. Before disabling anti-xray, many people who had weak computers or routers had problems with ping.
> 

Anti-xray makes chunk data basically random and random data doesn't compress well (it shouldn't compress at all, if it's truly random). This means to use anti-xray either your players need better internet connections or lower view distances or you need to turn down the `chunk-loading.target-player-chunk-send-rate` in your `paper.yml` so chunks are sent to them slower.

LucidAPs:
For the love of god, merge it upstream please!

LucidAPs:
@dalandis Did you see the same results with engine 1?

dalandis:
> @dalandis Did you see the same results with engine 1?

I have not tried engine 1

Teslicek:
I'm running a server on pterodactyl and allocated 20G using aikar flags (12GB+). I'm preloading chunks with chunky while using this dev build of paper and it seems like the RAM usage went overboard a little. Pterodactyl is reporting 32.2GB usage. Not sure what this means, so I've decided to report it here. (Using Oracle java 17.0.2)

NamiUni:
@Teslicek 
Screenshot_20220201-025053

and

Screenshot_20220201-025445

Teslicek:
> @Teslicek Screenshot_20220201-025053
> 
> and
> 
> Screenshot_20220201-025445

Hetzner i9 12900K + 128GB ram, I allowed unlimited RAM for the server, 0 swap

davispuh:
I rebased this on latest `master` in https://github.com/davispuh/Paper/tree/1.18-perf1
This PR improves latency a lot when using Velocity. Like night and day difference.

froobynooby:
Would you consider adding an API to adjust the 'fog distance'? i.e. tell the client the server's view distance is the larger of the 'fog distance' and their send distance. Set it to -1 to have the fog distance equal to the send distance.

It would be helpful for dynamic view distance plugins as whenever the client receives an updated view distance from the server it refreshes ALL chunks, which is quite jarring. For example, if the plugin knows it won't set the view distance higher than 20 it could set the fog distance to 20, meaning the client won't have to do a hard refresh when the server's view distance changes

radiant-ai:
Using this in production for more than 2 weeks with 30-50 players online with a lot of teleportation on a 60mil x 60mil world. I think it is stable. I did not notice any major issues such as crashes, lags or memory leaks.

kev626:
I've pulled this patch and several users of my fork have reported that these patches break the `prevent-moving-into-unloaded-chunks` option.

patches/server/0854-fixup-Highly-optimise-single-and-multi-AABB-VoxelSha.patch
electronicboy: to my understanding this concern was actually resolved so no longer an issue here

electronicboy (approved):

PR #4184, opened 2020-08-23
Add MonsterEggSpawnEvent

Proximyst:
Updated #1981.

Issue #6487, opened 2021-08-25
Create SummonEntityEvent with Entity/Block involved in Spawn


Doc94:
### Is your feature request related to a problem?

Currently i need detect when a entity or dispenser "summon" a entity for griefing... but the entities what can be "summon" are hard to detect in plugins because use shapes (EnderDragon, IronGolem, SnowGolem, Wither) and depend of another rules (difficulty, dimension, spawn, etc)

### Describe the solution you'd like.

A Cancellable event called when a entity or block (dispenser) summon a entity. with the Block or Entity involved in this summon (Nullable)

In teory the entities that can be summon are.

- Ender Dragon (using Ender Crystals)
- Wither (Player Building or with Dispenser)
- IronGolem (Player Building or with Dispenser)
- SnowGolem (Player Building or with Dispenser)

Not sure if is better and extension of CreatureSpawnEvent or a independed Event.

### Describe alternatives you've considered.

The only alternative i see in plugins is use NMS and when any place a block run the Shape and detect if can be the cause for summon and check the rules in vanilla for spawn.

### Agreements

- [X] I have searched for and ensured there isn't already an open issue regarding this.
- [X] I have ensured the feature I'm requesting isn't already in the latest supported Paper build.

### Other

i try to create this event in Craftbukkit but the entity summon not work the same form in all entities...

For example (based in craftbukkit):

- Wither
In BlockWitherSkull its a method (a) for check shape and summon this method is called in IDispenseBehavior when a Wither Skull is placed then is more easy make a method for add the entity/block involved. for Players in ItemStack is called the static method in BlockWitherSkull for check wither summon.
- IronGolem/SnowGolem
In BlockPumpkinCarved its the method for check but if im not wrong is only called by onPlace and this method not has the entity or block involved
- EnderDragon
In ItemEndCrystal every time a Ender Crystal its placed then check if the world has a battle and check for respawn (if in portal is with 4 ender crystal) and with another methods respawn the dragon because need regenerate the pilars and another things...

lynxplay:
[...]

PR #7377, opened 2022-01-20
Add ItemSpawnEntityEvent and BlockPlaceEntityEvent and expand EntityPlaceEvent 

### LINKED ISSUE #4795: EntitySpawnEvent with optional source ItemStack
I'd love to be able to modify the entity when it gets spawned from some ItemStack (spawn egg, boat, item frame, ...) and act on that.

Ability to retrieve to retrieve source ItemStack which caused the Entity to spawn.
I think it'd make most sense to add it as nulllable value to existing EntitySpawnEvent, since other events like CreatureSpawnEvent and ProjectileLaunchEvent might potentially also use this.

Patching the server source code directly :/

### LINKED ISSUE #6898: Add PlayerSpawnEntityEvent

I'm trying to modify the entities which can be spawned by players, according to certain player bound criteria, for example some nbt data of the item that was used to spawn the entity, or the level of the player.

A generic event which covers all the entities that can be spawned by a player using an item. More specifically:
- All SpawnEggs
- Both ItemFrame variants
- All fish buckets
- Axolotl Bucket
- Armorstand
- Painting

I would also find the following methods very useful:
- getItemStack() → The ItemStack used to spawn the entity
- getEntity() → The Entity being spawned
- getPlayer() → The Player who spawned the entity
- getSpawnReason() → The SpawnReason, to more easily differentiate without necessarily checking the itemstack

Listening to the player interact event, and either cancelling it, and spawning the entity myself, or trying to match it to the next entity spawn event. Both ways aren't ideal in my opinion.

### Other

_No response_

Machine-Maker:
WIP so far, still adding spots these events should be called, but should be a catch-all now for non-hanging entities spawned as a result of item "placement".

Resolves https://github.com/PaperMC/Paper/issues/4795
Resolves https://github.com/PaperMC/Paper/issues/6898
Supersedes https://github.com/PaperMC/Paper/pull/4184

### Changes
Adds ItemSpawnEntityEvent
Adds PlaceEntityEvent (extends ItemSpawnEntityEvent)
Adds BlockPlaceEntityEvent (extends PlaceEntityEvent)
Makes EntityPlaceEvent extend PlaceEntityEvent

Added an ItemStack field on ItemSpawnEntityEvent since that applies to all uses of that event and EntityPlaceEvent.

##### ItemSpawnEntityEvent
- Spawn eggs from dispensers and players
##### EntityPlaceEvent (player only)
- mob eggs from players 
##### BlockPlaceEntityEvent (dispenser only)
- boats
- armor stand
- mob buckets
- tnt

```[tasklist]
### Entity Placements
- [x] Spawn Eggs - ItemSpawnEntityEvent (dispenser w/null player)
- [x] Boats - EntityPlaceEvent/BlockPlaceEntityEvent
- [x] ~~Item Frame, Painting, Leash~~ (HangingPlaceEvent) (can't change due to API break)
- [x] Armor Stand - EntityPlaceEvent/BlockPlaceEntityEvent
- [x] End Crystals - EntityPlaceEvent
- [x] Fish Buckets - EntityPlaceEvent/BlockPlaceEntityEvent
- [x] ~~Fireworks~~ already tracks its spawning stack and shooter
- [x] ~~Fire Charges~~ already tracks its spawning stack and shooter
- [x] PrimedTNT BlockPlaceEntityEvent
```

### DIFF
--- a/patches/api/0480-Add-ItemSpawnEntityEvent-and-expand-EntityPlaceEvent.patch
+++ b/patches/api/0480-Add-ItemSpawnEntityEvent-and-expand-EntityPlaceEvent.patch
@@ -1,0 +1,304 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 20 Jan 2022 10:54:54 -0800
+Subject: [PATCH] Add ItemSpawnEntityEvent and expand EntityPlaceEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/entity/BlockPlaceEntityEvent.java b/src/main/java/io/papermc/paper/event/entity/BlockPlaceEntityEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..7287ebcf3338ca30cc23efa22084cdaefbcad2da
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/entity/BlockPlaceEntityEvent.java
+@@ -0,0 +1,48 @@
++package io.papermc.paper.event.entity;
++
++import org.bukkit.block.Block;
++import org.bukkit.block.BlockFace;
++import org.bukkit.block.Dispenser;
++import org.bukkit.entity.Entity;
++import org.bukkit.entity.Player;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Called when a block, like a dispenser, places an
++ * entity. {@link #getPlayer()} will always be null.
++ * @see org.bukkit.event.hanging.HangingPlaceEvent for paintings, item frames, and leashes.
++ * @see org.bukkit.event.entity.EntityPlaceEvent for a player-only version with more context
++ * @see PlaceEntityEvent to listen to both blocks and players placing entities
++ */
++public class BlockPlaceEntityEvent extends PlaceEntityEvent {
++
++    private final Dispenser dispenser;
++
++    @ApiStatus.Internal
++    public BlockPlaceEntityEvent(final @NotNull Entity entity, final @NotNull Block block, final @NotNull BlockFace blockFace, final @NotNull ItemStack spawningStack, final @NotNull Dispenser dispenser) {
++        super(entity, null, block, blockFace, spawningStack);
++        this.dispenser = dispenser;
++    }
++
++    /**
++     * Get the dispenser responsible for placing the entity.
++     *
++     * @return a non-snapshot Dispenser
++     */
++    public @NotNull Dispenser getDispenser() {
++        return this.dispenser;
++    }
++
++    /**
++     * Player will always be null on this event.
++     */
++    @Override
++    @Contract("-> null")
++    public @Nullable Player getPlayer() {
++        return null;
++    }
++}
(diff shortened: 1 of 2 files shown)


molor:
This is not the same as https://github.com/PaperMC/Paper/pull/4184?

Machine-Maker:
@molor a couple differences with that
1. Its type inhertiance doesn't integerate with EntityPlaceEvent
2. this provides access to the blockface
3. no "entity setter", that really doesn't make sense with how the API deals with entities IMO
4. This PR expands EntityPlaceEvent in addition to adding a new event

aerulion:
This would probably also close #6898, at least in regards to what I initially wanted to achieve.

Machine-Maker:
Something I'm not sure about, are fire charges from dispeners? those technically do spawn an entity from an itemstack. but its different because a player can't do the same thing.

aerulion:
I would say the event should be called in this case, as it otherwise would be inconsistent to call the events for something like fireworks and spawn eggs, but not for fire charges. And to me it would be more likely for someone wanting to modify the dispensed fire charge, then someone wondering why there's always a null player on an entity which cannot be "summoned" by a player.

Machine-Maker:
Ok, that makes sense, but should it be EntityPlaceEvent or ItemSpawnEntityEvent? I’m leaning to EntityPlaceEvent even tho it’s more of a launch. (Like fireworks I guess)

Machine-Maker:
I've decided against adding Fire Charges and fireworks. I don't want to add any projectiles. They already have api that tracks their spawning stack and shooter. I did add PrimedTNT to the list, since a dispenser is placing that and its creating an entity.

patches/server/0918-Add-ItemSpawnEntityEvent-and-expand-EntityPlaceEvent.patch
```diff
@@ -0,0 +1,277 @@
...
+diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
+index 3d2b5f040715a0e4fac0e6786bd11a4d715330ce..d61a565657bdc333218d1d6a64f9367ac2ead047 100644
+--- a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
++++ b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
+@@ -211,7 +211,7 @@ public interface DispenseItemBehavior {
+             @Override
+             public ItemStack execute(BlockSource pointer, ItemStack stack) {
+                 Direction enumdirection = (Direction) pointer.getBlockState().getValue(DispenserBlock.FACING);
+-                EntityType entitytypes = ((SpawnEggItem) stack.getItem()).getType(stack.getTag());
++                EntityType<?> entitytypes = ((SpawnEggItem) stack.getItem()).getType(stack.getTag()); // Paper - decompile fix
```
Lulu13022002: Should probably be moved in MC-Dev-fixes.patch ?

Machine-Maker:
Rebased for 1.19

Doc94:
This can consider the staled #6487 ? its like the #6898 created later.. mostly for the Wither/Golems created using blocks detect how and who build (for avoid things like try listen place blocks and spawn for try make that)

Machine-Maker:
Rebased for 1.20.4

electronicboy:
gave this a look over as much as I can with my level of sleep and sight, overall looks fine of what I've seen

Machine-Maker:
Rebased for 1.20.6

Noogear:
Half a year has passed. Has this PR been forgotten? Up to now, there is still a lack of an event for tracking the generation of fish using buckets.

PR #7378, opened 2022-01-20
Make tag presets unmodifiable

Machine-Maker:
Tag presets that Paper adds should not be modifiable as they are global which can lead to unexpected behavior.

If this breaks any behavior for plugins, its bad behavior. Tags are not locked by default, so no proper behavior with custom tags plugins add should be affected.


electronicboy:
ensure size is a validation check, not a "last function" type call, really should just use an explicit lock call rather than hijacking ensureSize, imho

Machine-Maker:
It’s not hijacking it, it’s a whole other method. Default is that the existing ensureSize won’t lock it.

electronicboy:
you're adding a param onto a method which does 1 thing to make it do another behind the context of a boolean, I don't feel that locking should be part of the ensureSize method

Machine-Maker:
Ok, changed it to just use the `lock` method. Also added a unit test to make sure all the tag presets are locked.

electronicboy (approved):

PR #7379, opened 2022-01-20
Add more breakNaturally functionality

Machine-Maker:
Break naturally was missing the experience drops from several blocks, as well as not handling turtle eggs and ice blocks correctly.
I think that should be all functionality that doesn't require a player. For example, if breaking a beehive with a non silk touch tool, the bees should get angry at the player, but since there's no player present, that's skipped here.


Owen1212055 (approved):
Before behavior: (holding diamond in main hand represents Block#breakNaturally(true)
https://user-images.githubusercontent.com/23108066/156930119-0b5c8c50-f06d-4f61-9b66-4d1f741dddbd.mp4

After:
https://user-images.githubusercontent.com/23108066/156930237-9ec03e1e-161b-4186-85fd-43e5e63d830d.mp4


I also agree with the reason for excluding the beehive from these special conditions.
Tested!

patches/server/0827-Improves-Block-breakNaturally.patch
```diff
@@ -0,0 +1,67 @@
...
+         net.minecraft.world.level.block.Block block = iblockdata.getBlock();
+         net.minecraft.world.item.ItemStack nmsItem = CraftItemStack.asNMSCopy(item);
++        net.minecraft.world.level.material.FluidState fluidState = this.world.getFluidState(this.position); // Paper
+         boolean result = false;
+ 
+         // Modelled off EntityHuman#hasBlock
+         if (block != Blocks.AIR && (item == null || !iblockdata.requiresCorrectToolForDrops() || nmsItem.isCorrectToolForDrops(iblockdata))) {
+             net.minecraft.world.level.block.Block.dropResources(iblockdata, this.world.getMinecraftWorld(), position, this.world.getBlockEntity(position), null, nmsItem);
+             if (triggerEffect) world.levelEvent(org.bukkit.Effect.STEP_SOUND.getId(), position, net.minecraft.world.level.block.Block.getId(block.defaultBlockState())); // Paper
++            block.popExperience(this.world.getMinecraftWorld(), this.position, block.getExpDrop(iblockdata, this.world.getMinecraftWorld(), this.position, nmsItem)); // Paper - drop experience
```
electronicboy: should this one maybe be configurable?

electronicboy (approved):

patches/server/0827-Improves-Block-breakNaturally.patch
Machine-Maker: you mean a param in the method? I guess it can be. I just figured it should happen by default since its part of the block breaking.

Machine-Maker:
Rebased for 1.18.2

Also merged with pre-existing server/api patch for extending breakNaturally functionality.

PR #7380, opened 2022-01-20
Fix EntityEquipment and related javadocs

Machine-Maker:
Sooooo many things wrong here.

The EntityEquipment javadocs state that various getters return "a copy of" which is not true for the CraftInventoryPlayer impl.

The much better fix, would be to make them actually return copies since there are setters, and I would be totally fine with doing that, but it might be considered too big of a break even tho the contract (via the javadoc) says they are copies.

Another issue here, is the nullability annotations are wrong on EntityEquipment because the two different impls of it use different logic. Would also prefer to make them all notnull, but that might also be a behavioral break.

### DIFF
--- a/patches/api/0054-Fix-upstream-javadocs.patch
+++ b/patches/api/0054-Fix-upstream-javadocs.patch
@@ -124,16 +124,248 @@ index 1119e26e270bb45f517955b19d95a9ec3d113634..4631647c64c89ffdde2d9b63bdab974a
       * @param sz The new size of the slime.
       */
      public void setSize(int sz);
-diff --git a/src/main/java/org/bukkit/event/player/PlayerMoveEvent.java b/src/main/java/org/bukkit/event/player/PlayerMoveEvent.java
-index 1b2267f4e8ebded198773ec80e2bff2c861c7084..1a58734d919fae247eeb85dd785fd59990856505 100644
---- a/src/main/java/org/bukkit/event/player/PlayerMoveEvent.java
-+++ b/src/main/java/org/bukkit/event/player/PlayerMoveEvent.java
-@@ -78,7 +78,7 @@ public class PlayerMoveEvent extends PlayerEvent implements Cancellable {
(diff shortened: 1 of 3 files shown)


patches/api/0054-Fix-upstream-javadocs.patch
```diff
@@ -122,11 +122,258 @@ index 1b2267f4e8ebded198773ec80e2bff2c861c7084..1a58734d919fae247eeb85dd785fd599
...
+ 
+     /**
+      * Get all additional ItemStacks stored in this inventory.
+@@ -26,8 +25,7 @@ public interface PlayerInventory extends Inventory {
+      *
+      * @return All additional ItemStacks. Individual items can be null.
+      */
+-    @NotNull
+-    public ItemStack[] getExtraContents();
++    public @Nullable ItemStack @NotNull[] getExtraContents();
```
Owen1212055: Paper comment

patches/api/0054-Fix-upstream-javadocs.patch
```diff
@@ -122,11 +122,258 @@ index 1b2267f4e8ebded198773ec80e2bff2c861c7084..1a58734d919fae247eeb85dd785fd599
...
 --- a/src/main/java/org/bukkit/inventory/PlayerInventory.java
 +++ b/src/main/java/org/bukkit/inventory/PlayerInventory.java
-@@ -106,7 +106,7 @@ public interface PlayerInventory extends Inventory {
+@@ -14,8 +14,7 @@ public interface PlayerInventory extends Inventory {
+      *
+      * @return All the ItemStacks from the armor slots. Individual items can be null.
+      */
+-    @NotNull
+-    public ItemStack[] getArmorContents();
++    public @Nullable ItemStack @NotNull[] getArmorContents();
```
Owen1212055: Paper comment

Owen1212055 (changes requested):
Looks good when looking at the underlying implementation.

Some missing paper comments tho

Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

PR #7382, opened 2022-01-21
Expand PlayerItemMendEvent

### LINKED ISSUE #7313: Introduce XPConsumed property to PlayerItemMendEvent

I've created a plugin that wants to be able to change the max durability of a given individual item stack (It emulates custom items with customModelData).

As a part of this, it emulates durability via tracking it's own property when the item is used, and updating the real damage as a 'display' value that is interpolated.

However mending throws a huge wrench into this, as a CarrotOnAStick only has 25 durability, so it is VERY cheap to repair with mending.

So expanding this out to, say, 2000 uses, you can get instant repairs of very expensive items.

I think the minimal patch to spigot, would be allowing paper plugins to set the value of k,
or maybe this.durabilityToXP(j)

image

So, the event would expose both the repairAmount, as well as the XP remaining.

This would allow for people to change the 'scaling' rate of mending universally, making it more or less buffed/nerfed, or implement interesting mechanics.

For this reason, I think it's worthy on it's own right, than the alternative proposed below.

However, I foresee a problem though, on whether setting the RepairAmount would also mutate the XPRemaining for spigot compatibility.

So maybe it would need to be XPConsumed?

Or maybe it would need to be deprecated under paper, since setting the repairAmount would need to mutate both, and expose a replacement property that does not get changed whenever repairAmount is?

Anyway, I leave this to the implementer to consider.

Exposing a general purpose durability event, that mending, item use and repair recipes would trigger.

However it's my belief that this could be painful to maintain / upkeep.

### Other

_No response_

### LINKED ISSUE #7449: Changing PlayerItemMend's repairAmount doens't change repair:experience ratio
### Expected behavior

Changing the event's `repairAmount` would change the experience to repair amount ratio.

For example, given an experience orb with value 10, this event would have `repairAmount = 20`: if I change it to 10, I expect it to consume the same 10 experience but changing the repair amount.

### Observed/Actual behavior

No matter what the value of `repairAmount` given, NMS' recursion on `repairPlayerItems` (on Experience Orb entity) will always give the same repair amount, making this event a bit useless.

For instance, if I set `repairAmount` to 2 no matter what the initial value is, it would consume 1 experience from the orb, and recursion would call the event again until all the experience is consumed on the same 2:1 ratio.

### Paper version

`This server is running Paper version git-Paper-187 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 0cc2503)`

### Other

```java

public class MendingNerfListener implements Listener {

    private final @NotNull JavaPlugin plugin;

    public MendingNerfListener(@NotNull JavaPlugin plugin) {
        this.plugin = plugin;
    }

    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
    private void nerfMending(PlayerItemMendEvent event) {
        plugin.getServer()
              .sendMessage(Component.text("id = %d experience = %d, repair amount = %d, item = %s with %d".formatted(
                      event.getExperienceOrb().getEntityId(),
                      event.getExperienceOrb().getExperience(),
                      event.getRepairAmount(),
                      event.getItem().getType().name(),
                      ((Damageable) event.getItem().getItemMeta()).getDamage())));
        //if (!plugin.getConfiguration().isMendingNerfed()) {
        if (false) { // replace with true for normal behaviour
            return;
        }
        event.setRepairAmount(Math.min(event.getRepairAmount(), 2));
    }

    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
    private void checkNerf(PlayerItemMendEvent event) {
        plugin.getServer()
              .sendMessage(Component.text("=> id = %d experience = %d, repair amount = %d, item = %s with %d".formatted(
                      event.getExperienceOrb().getEntityId(),
                      event.getExperienceOrb().getExperience(),
                      event.getRepairAmount(),
                      event.getItem().getType().name(),
                      ((Damageable) event.getItem().getItemMeta()).getDamage())));
    }
}
```

**Edit:** I noticed how my understanding of Mending was wrong and updated the issue accordingly (I believe it stills needs changing)

### LINKED ISSUE #7815: PlayerItemMendEvent#setRepairAmount(0) causes recursion
### Expected behavior

Unsure?

### Observed/Actual behavior

Doing
```
    @EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
    public void a(PlayerItemMendEvent event) {
        event.setRepairAmount(0);
    }
```
will cause a recursion and kick you.

I'm unsure if this should do this and if so I feel like there should be at least a note.

### Steps/models to reproduce

Do the code above.

### Plugin and Datapack List

N/A

### Paper version

> ver
[03:44:05 INFO]: Checking version, please wait...
[03:44:05 INFO]: This server is running Paper version git-Paper-333 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 9fd870d)
Previous version: git-Paper-331 (MC: 1.18.2)

### Other

_No response_

Edit: https://hub.spigotmc.org/jira/browse/SPIGOT-6708 Related

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/7313
Resolves https://github.com/PaperMC/Paper/issues/7449
Closes https://github.com/PaperMC/Paper/issues/7815

Adds a getter/setter for the operation used to turn the amount of durability repaired, into an XP value to be subtracted away from the total value of the experience orb.

### DIFF
--- a/patches/api/0416-Expand-PlayerItemMendEvent.patch
+++ b/patches/api/0416-Expand-PlayerItemMendEvent.patch
@@ -1,0 +1,71 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 20 Jan 2022 18:11:44 -0800
+Subject: [PATCH] Expand PlayerItemMendEvent
+
+
+diff --git a/src/main/java/org/bukkit/event/player/PlayerItemMendEvent.java b/src/main/java/org/bukkit/event/player/PlayerItemMendEvent.java
+index 5b2415c2c92127947a21dfe8d672d2b88ea457df..533531f69a8d549e184161eefe6e5bf8a9e85c05 100644
+--- a/src/main/java/org/bukkit/event/player/PlayerItemMendEvent.java
++++ b/src/main/java/org/bukkit/event/player/PlayerItemMendEvent.java
+@@ -23,14 +23,60 @@ public class PlayerItemMendEvent extends PlayerEvent implements Cancellable {
+     private final ExperienceOrb experienceOrb;
+     private int repairAmount;
+     private boolean cancelled;
++    private java.util.function.IntUnaryOperator durabilityToXpOp; // Paper
+ 
++    @Deprecated // Paper
+     public PlayerItemMendEvent(@NotNull Player who, @NotNull ItemStack item, @NotNull EquipmentSlot slot, @NotNull ExperienceOrb experienceOrb, int repairAmount) {
++        // Paper start
++        this(who, item, slot, experienceOrb, repairAmount, amount -> amount / 2);
++    }
++
++    @org.jetbrains.annotations.ApiStatus.Internal
++    public PlayerItemMendEvent(@NotNull Player who, @NotNull ItemStack item, @NotNull EquipmentSlot slot, @NotNull ExperienceOrb experienceOrb, int repairAmount, @NotNull java.util.function.IntUnaryOperator durabilityToXpOp) {
++        // Paper end
+         super(who);
+         this.item = item;
+         this.slot = slot;
+         this.experienceOrb = experienceOrb;
+         this.repairAmount = repairAmount;
++        // Paper start
++        this.durabilityToXpOp = durabilityToXpOp;
++    }
++
++    /**
++     * Get the operation used to calculate xp used based on
++     * the set repair amount. Used to calculate how much of
++     * an XP orb will be consumed by this mend operation.
++     *
++     * @return the durability-to-xp operation
++     */
++    public @NotNull java.util.function.IntUnaryOperator getDurabilityToXpOperation() {
++        return this.durabilityToXpOp;
++    }
++
++    /**
++     * Sets the operation used to calculate xp used based on
++     * the set repair amount. Used to calculate how much of
++     * an XP orb will be consumed by this mend operation.
++     *
++     * @param durabilityToXpOp the durability-to-xp operation
++     */
++    public void setDurabilityToXpOperation(@NotNull java.util.function.IntUnaryOperator durabilityToXpOp) {
++        com.google.common.base.Preconditions.checkNotNull(durabilityToXpOp);
++        this.durabilityToXpOp = durabilityToXpOp;
++    }
++
++    /**
++     * Helper method to get the amount of experience that will be consumed.
++     * This method just returns the result of inputting {@link #getRepairAmount()}
++     * into the function {@link #getDurabilityToXpOperation()}.
++     *
++     * @return the amount of xp that will be consumed
++     */
++    public int getConsumedExperience() {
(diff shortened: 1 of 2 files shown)


ghost (approved):

patches/api/0359-Expand-PlayerItemMendEvent.patch
```diff
@@ -0,0 +1,58 @@
...
+     private boolean cancelled;
++    private java.util.function.IntUnaryOperator durabilityToXpOp; // Paper
+ 
++    @Deprecated // Paper
+     public PlayerItemMendEvent(@NotNull Player who, @NotNull ItemStack item, @NotNull ExperienceOrb experienceOrb, int repairAmount) {
++        // Paper start
++        this(who, item, experienceOrb, repairAmount, amount -> amount / 2);
++    }
++
++    public PlayerItemMendEvent(@NotNull Player who, @NotNull ItemStack item, @NotNull ExperienceOrb experienceOrb, int repairAmount, java.util.function.IntUnaryOperator durabilityToXpOp) {
```
NoahvdAa: Needs a nullability annotation on the UnaryOperator (also why the tests are failing)

Machine-Maker:
Rebased for 1.18.2

Now resolves 2 open issues.

rafaelsms:
> Now resolves 2 open issues.

Now resolves 3 open issues.

patches/server/0884-Expand-PlayerItemMendEvent.patch
```diff
@@ -0,0 +1,68 @@
...
+@@ -1222,10 +1222,16 @@ public class CraftEventFactory {
+         return event;
+     }
+ 
++    @Deprecated // Paper - use method with durabilityToXpOp parameter
+     public static PlayerItemMendEvent callPlayerItemMendEvent(net.minecraft.world.entity.player.Player entity, net.minecraft.world.entity.ExperienceOrb orb, net.minecraft.world.item.ItemStack nmsMendedItem, int repairAmount) {
++        // Paper start
++        return CraftEventFactory.callPlayerItemMendEvent(entity, orb, nmsMendedItem, repairAmount, amount -> amount / 2);
++    }
++    public static PlayerItemMendEvent callPlayerItemMendEvent(net.minecraft.world.entity.player.Player entity, net.minecraft.world.entity.ExperienceOrb orb, net.minecraft.world.item.ItemStack nmsMendedItem, int repairAmount, java.util.function.IntUnaryOperator durabilityToXpOp) {
```
kennytv: No need to create a new method, just add the parameter to the existing one

kennytv (approved):
I'd write out "Operation" in the getter and setter, even if it does make it pretty long, but otherwise fine

Machine-Maker:
Ok, tested and this fixes the recursion issue, and the new API works as expected. I'll wait for a 2nd approval since this is new API and not a simple fix.

oddlama:
Any updates on this PR? It seems to be fully implemented and functional already - is there anything that is blocking this from moving forward?

kennytv (approved):

PR #6478, opened 2021-08-23
Fix cancelling EntityDeathEvent for Armor Stands killed by players

NoahvdAa:
Fixes #6476.

This also fixes the event double firing sometimes, which spigot claims to fix.

PR #7385, opened 2022-01-21
Don't fire EntityDeathEvent twice for armor stands

Machine-Maker:
Previously, Armor Stands had their EntityDeathEvent fired twice
if they were killed by an explosion or by being on/in fire.

Perhaps this should go in another patch? Phoenix's patch to "Improve death events" maybe, idk.


Machine-Maker:
Accidental duplicate of https://github.com/PaperMC/Paper/pull/6478

PR #7387, opened 2022-01-22
Tab completion for basic aliases

Machine-Maker:
So this adds tab completions to *some* command aliases. 

Basically, only ones that end in `$1-` to capture all arguments. Implementing tab completions beyond that is... I think much more involved and ambiguous. Like if the format string takes the `$2` argument, how is a tab completion for that going to work?.

But I think this should really cover really all the use cases.


Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

electronicboy:
approved, always kinda wondered if just having a plain aliases section would be the best in commands.yml, given that majority of the usecases of this is just tryna deal with command conflicts

Machine-Maker:
You think it'd be better if I just made it a separate field? with 1-to-1 aliases that all then would be guaranteed to support tab completions?

electronicboy:
That was my thought, as then we wouldn't need to guide people on how to setup the simplest of aliases and stuff would *just work*

PR #7392, opened 2022-01-22
Don't load plugins prefixed with a dot

### LINKED ISSUE #7301: Bukkit's plugin loader doesn't ignore `.jar` files that have names beginning with `._` (Mac OS)
### Expected behavior

**Context**
(issue template doesn't contain a text box for this)
Mac OS sometimes creates a file on a storage drive which contains metadata for another file. This metadata file has the same name as the file it contains metadata for, except it is prefixed with `._`, and it is hidden to users (I've used TinkerTool to show hidden files). Why it doesn't use a different file extension, like adding `.temp` at the end? I don't know. :smile:
This is very common on external drives as many are not formatted using Mac's file system. [This webpage](https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them) contains a decent explanation of why they are created.

***

**Expected behaviour**

Bukkit's plugin loader should ignore any plugins that have file names beginning with `._` - simple as that. No plugin would or should have a name beginning with that.

### Observed/Actual behavior

Bukkit's plugin loader attempts to load the metadata files, resulting in an error akin to:
```
[16:26:28 ERROR]: Could not load 'plugins/._Terra-bukkit-5.4.1-BETA+efd1665c-shaded.jar' in folder 'plugins'
org.bukkit.plugin.InvalidDescriptionException: Invalid plugin.yml
	at org.bukkit.plugin.java.JavaPluginLoader.getPluginDescription(JavaPluginLoader.java:186) ~[patched_1.17.1.jar:git-Paper-408]
	at org.bukkit.plugin.SimplePluginManager.loadPlugins(SimplePluginManager.java:158) ~[patched_1.17.1.jar:git-Paper-408]
	at org.bukkit.craftbukkit.v1_17_R1.CraftServer.loadPlugins(CraftServer.java:419) ~[patched_1.17.1.jar:git-Paper-408]
	at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:287) ~[patched_1.17.1.jar:git-Paper-408]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1220) ~[patched_1.17.1.jar:git-Paper-408]
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:319) ~[patched_1.17.1.jar:git-Paper-408]
Caused by: java.util.zip.ZipException: zip END header not found
	at java.util.zip.ZipFile$Source.findEND(ZipFile.java:1469) ~[?:?]
	at java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1477) ~[?:?]
	at java.util.zip.ZipFile$Source.<init>(ZipFile.java:1315) ~[?:?]
	at java.util.zip.ZipFile$Source.get(ZipFile.java:1277) ~[?:?]
	at java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:709) ~[?:?]
	at java.util.zip.ZipFile.<init>(ZipFile.java:243) ~[?:?]
	at java.util.zip.ZipFile.<init>(ZipFile.java:172) ~[?:?]
	at java.util.jar.JarFile.<init>(JarFile.java:347) ~[?:?]
	at java.util.jar.JarFile.<init>(JarFile.java:318) ~[?:?]
	at java.util.jar.JarFile.<init>(JarFile.java:284) ~[?:?]
	at org.bukkit.plugin.java.JavaPluginLoader.getPluginDescription(JavaPluginLoader.java:174) ~[patched_1.17.1.jar:git-Paper-408]
	... 6 more
```
However, the *actual* plugin file actually loads completely fine. Everything works great. Just annoying errors on startup which make it hard to identify any errors which would actually be of interest since I now have to ignore startup errors.

### Steps/models to reproduce

* Plugin in an external drive or any drive that is not formatted under Mac's own file system
* Put a local server on the drive, install a plugin
  * Unsure if this happens to all plugins. For example, it happened for Terra but not PhantomWorlds on the test I did. 
* Check startup logs to see if there is a plugin.yml error

### Plugin and Datapack List

```
> plugins
[16:18:02 INFO]: Plugins (2): PhantomWorlds, Terra
```

```
> datapack list
[16:18:04 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
[16:18:04 INFO]: There are no more data packs available
```

### Paper version

```
> version
[16:26:33 INFO]: Checking version, please wait...
[16:26:33 INFO]: This server is running Paper version git-Paper-408 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: cc43ba7 on ver/1.17.1)
Previous version: git-Paper-124 (MC: 1.18.1)
```

### Other

I hope reporting this to Paper was the right thing to do. I had a feeling that reporting this to Spigot's tracker would end up in the issue being dormant for a long time, perhaps never being addressed. Any fix to this issue could be merged upstream as well.

NoahvdAa:
Fixes #7301. Not sure if there should be some sort of toggle for this, since there probably aren't a lot of (non-malicous) use-cases for loading plugin files starting with a dot.

### DIFF
--- a/patches/api/0366-Don-t-load-plugins-prefixed-with-a-dot.patch
+++ b/patches/api/0366-Don-t-load-plugins-prefixed-with-a-dot.patch
@@ -1,0 +1,18 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Noah van der Aa <ndvdaa@gmail.com>
+Date: Sat, 22 Jan 2022 16:35:44 +0100
+Subject: [PATCH] Don't load plugins prefixed with a dot
+
+
+diff --git a/src/main/java/org/bukkit/plugin/SimplePluginManager.java b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
+index 1366496271c4c7f72d1e5f990e51775b1c371f99..42da20011544075a9bea63a12ae86f2f21720667 100644
+--- a/src/main/java/org/bukkit/plugin/SimplePluginManager.java
++++ b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
+@@ -142,6 +142,7 @@ public final class SimplePluginManager implements PluginManager {
+         final List<File> pluginJars = new ArrayList<>(java.util.Arrays.asList(directory.listFiles()));
+         pluginJars.addAll(extraPluginJars);
+         for (File file : pluginJars) {
++            if (file.getName().startsWith(".") && !extraPluginJars.contains(file)) continue; // Don't load plugin if the file name starts with a dot, except if it's a extra plugin jar.
+             // Paper end
+             PluginLoader loader = null;
+             for (Pattern filter : filters) {


Chew:
If we're fixing explicitly this issue, it should really be anything starting with `._`, just in case.

NoahvdAa:
Went with everything with a dot because of sulu's comment:

> Honestly, this is more of a feature than a bug; for whoever addresses this: worth considering if files prefixed with . should be loaded at all and if other systems do any similar things.

I can't imagine there are many non-malicious reasons for loading invisible plugin files, would like to hear some other opinions on this.

patches/api/0360-Don-t-load-plugins-prefixed-with-a.patch
```diff
@@ -0,0 +1,18 @@
...
+
+diff --git a/src/main/java/org/bukkit/plugin/SimplePluginManager.java b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
+index 1366496271c4c7f72d1e5f990e51775b1c371f99..51bfae5f96b048482651cbbd2eade6beacbae80d 100644
+--- a/src/main/java/org/bukkit/plugin/SimplePluginManager.java
++++ b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
+@@ -142,6 +142,7 @@ public final class SimplePluginManager implements PluginManager {
+         final List<File> pluginJars = new ArrayList<>(java.util.Arrays.asList(directory.listFiles()));
+         pluginJars.addAll(extraPluginJars);
+         for (File file : pluginJars) {
++            if (file.getName().startsWith(".")) continue; // Don't load plugin if the file name starts with a .
```
ghost: This should also check if it isn’t an extra plugin jar

Chew: It _can_ but what if it loads the dot file first?

lokka30: > It _can_ but what if it loads the dot file first?

I assume this is likely to happen because of alphabetical ordering.
So I don't believe we can check for the extra plugin jar .. that is unless we loop thru each jar to see if the plugin.yml name value is the same but that seems too extra to achieve the desired outcome of the PR :)
What do you think?

NoahvdAa: Thanks, added an exemption if the File is in the extraPluginJars list.

ghost (approved):

patches/api/0360-Don-t-load-plugins-prefixed-with-a.patch
lokka30: Oh, far out, I had the complete wrong perception of what an extra plugin JAR was. I apologize.

lokka30 (approved):

kennytv (approved):

lokka30:
Thank you all. 🎊

PR #7398, opened 2022-01-24
Add 2 missing feature seed configs

### LINKED ISSUE #6978: Add Mineshafts and Strongholds to the feature-seeds: Configuration

Using a website like **https://www.chunkbase.com/** you can find **Mineshafts** and **Strongholds** if you were to crack the seed. There are 18 structure features but Spigot only has 16 options, no support to change seeds for Mineshafts or Strongholds. Even after changing all possible seed-x in spigot.yml and changing **generate-random-seeds-for-all:** to **true** it still won't work for the 2 options left out. 

Add configurations to Paper to change the seed of Mineshafts and Strongholds (optional).

No alternatives known.

### Other

_No response_

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/6978

Related (added stronghold seed): https://github.com/PaperMC/Paper/pull/7334

Tested same was as earlier PR, found location using https://www.chunkbase.com/apps for mineshafts and buried treasure, ensured they still generated on default settings, then changed the seed for each, deleted region files, and then rejoined in same spot to find them missing.


electronicboy (approved):

Machine-Maker:
Rebased for 1.18.2

patches/server/0852-Add-missing-structure-set-seed-configs.patch
```diff
@@ -0,0 +1,120 @@
...
+     private void initWorldGenSeeds()
+     {
+         this.villageSeed = this.getInt( "seed-village", 10387312 );
+@@ -381,6 +390,11 @@ public class SpigotWorldConfig
+         this.mansionSeed = this.getInt( "seed-mansion", 10387319 );
+         this.fossilSeed = this.getInt( "seed-fossil", 14357921 );
+         this.portalSeed = this.getInt( "seed-portal", 34222645 );
++        // Paper start - add missing feature seed configs
++        this.mineshaftSeed = this.getSeed("seed-mineshaft", Integer::parseInt);
++        this.buriedTreasureSeed = this.getInt("seed-buried-treasure", 10387320); // def taken from BuriedTreasureFeature#RANDOM_SALT field
```
NoahvdAa, line 115: Maybe add a diff on update to that field? Seems to already be gone in the latest snapshot though, so I'm not sure how long that'd last.

EDIT: Looks like it was moved to StructurePlacement#HIGHLY_ARBITRARY_RANDOM_SALT

NoahvdAa (approved):

patches/server/0852-Add-missing-structure-set-seed-configs.patch
Machine-Maker: You know, its not super clear that this can be changed in the latest snapshot at all. The reason you can't define a seed for buried treasure the same way you can for other structure sets, is that the default `spacing` and `separation` values are `1` and `0` which means the salt used has 0 effect because the random will *always* return 0. (the exclusive upper bound is `spacing - separation`. So the salt has no effect. there doesn't seem to be a replacement for it in the snapshots, no second random check.

Machine-Maker:
Superseded by https://github.com/PaperMC/Paper/pull/8030

PR #7399, opened 2022-01-24
Furnace RecipesUsed API

Machine-Maker:
stored in furnaces to calculate XP to give player's on stack removal from the result slot, or when the block is broken.

### DIFF
--- a/patches/api/0368-Furnace-RecipesUsed-API.patch
+++ b/patches/api/0368-Furnace-RecipesUsed-API.patch
@@ -1,0 +1,51 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 13 Jan 2022 15:21:08 -0800
+Subject: [PATCH] Furnace RecipesUsed API
+
+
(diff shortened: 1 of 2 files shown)


patches/server/0860-Furnace-RecipesUsed-API.patch
```diff
@@ -0,0 +1,43 @@
...
++        if (count > 0) {
++            this.getSnapshot().recipesUsed.put(location, count);
++        } else {
++            this.getSnapshot().recipesUsed.removeInt(location);
++        }
++    }
++
++    @Override
++    public java.util.Collection<org.bukkit.NamespacedKey> getRecipesUsed() {
++        return this.getSnapshot().recipesUsed.keySet().stream().map(org.bukkit.craftbukkit.util.CraftNamespacedKey::fromMinecraft).toList();
```
ghost: Shouldn’t this return a Set?

Machine-Maker: it could be, but idk if it really matters. It's an unmodifiable list anyways.

ghost (approved):

Machine-Maker:
Upstream did it badly imo, so this can still be a thing.

electronicboy (approved):

PR #7403, opened 2022-01-24
Allow blocks above y 255 to be randomticked

Titaniumtown:
It seems that the 1.18 worldheight changes weren't taken into account when porting a patch that optimizes random block ticking. This PR corrects that and now allows blocks above 255 to be random ticked.


electronicboy (approved):

kennytv (changes requested):
That's not correct, ~~pretty sure there's also other multiple other places this haven't been updated~~

Spottedleaf:
The y position is local to the chunk section, and then is turned into a global position via the OR with the chunk section's y * 16.
Even if the y position was global, the right shift would need to be changed to account for sign.
getMaxBuildHeight is also exclusive, so this patch would have caused _no blocks to tick_.

PR #7404, opened 2022-01-24
Fix cmd permission levels for command blocks

Machine-Maker:
In vanilla, command blocks (block and minecarts) have a perm level of `2`, restricting them from running `/stop`, `/ban`, `/kick`, and similar "more important" commands. craftbukkit attempted to "simulate" that with
```java
        String cmd = args[0];
        if (cmd.startsWith("minecraft:")) cmd = cmd.substring("minecraft:".length());
        if (cmd.startsWith("bukkit:")) cmd = cmd.substring("bukkit:".length());

        // Block disallowed commands
        if (cmd.equalsIgnoreCase("stop") || cmd.equalsIgnoreCase("kick") || cmd.equalsIgnoreCase("op")
                || cmd.equalsIgnoreCase("deop") || cmd.equalsIgnoreCase("ban") || cmd.equalsIgnoreCase("ban-ip")
                || cmd.equalsIgnoreCase("pardon") || cmd.equalsIgnoreCase("pardon-ip") || cmd.equalsIgnoreCase("reload")) {
            return 0;
        }
```
called when a command block tried to run a command. but that's not a great idea, for one, its very outdated, missing several newer commands that also have a 3+ perm level, **and** on a bigger note, it doesn't take into account command redirections like the `/execute ... run ...` command does. 

This PR removes that check, and just uses vanilla's built-in function levels for checking those perms. It also adds configs to turn off this functionality, or set the command block perm level away from the default `2`.

~~This also fixes a UOE when loading a datapack with a function that has a perm level greater than the one configured in `server.properties`.~~ *this was done in a different PR already, and this replaces and includes that change*


Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

Machine-Maker:
Rebased for 1.19.2

Machine-Maker:
Rebased for 1.20.2. Updated to replace an existing patch that improved vanilla permission level checking.

Owen1212055 (approved):

PR #7405, opened 2022-01-25
Add DISMOUNT teleport cause

Machine-Maker:
Also fixes an issue where the TeleportEvent is fired on disconnect. Its **possible** that it should fire an event there, but since from the player's perspective, they never actually left the entity, I'm not sure. Cause when you log back on, you are still riding the entity, but the server does dismount you internally. Either way, that part can be removed, but the teleport cause can remain.


Machine-Maker:
Rebased for 1.18.2

electronicboy (approved):

PR #7406, opened 2022-01-25
Fix player flight being disabled on world change

mxson:
Currently, when a player is flying while in survival mode and teleports to a different world, their flight will be disabled. This is due to being set to survival mode while already in survival mode, which resets the player's "mayfly" and "flying" abilities to the default "false" state. This behavior began in 1.17, 1.16 did not have this issue.

This patch resolves this issue, allowing players to teleport between worlds without their flight status being changed.

I've been running this patch in a production environment without any side effects.

### DIFF
--- a/patches/server/0860-Fix-player-flight-disabling-on-world-change.patch
+++ b/patches/server/0860-Fix-player-flight-disabling-on-world-change.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: mxson <m@son.xxx>
+Date: Mon, 24 Jan 2022 18:54:45 -0800
+Subject: [PATCH] Fix player flight disabling on world change
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 7b23535a680d2a8534dcb8dd87770f66fb982c13..30cc6c77856a93d5642832ba91dc0bb7e5e4eb6b 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1721,7 +1721,13 @@ public class ServerPlayer extends Player {
+ 
+     public void restoreFrom(ServerPlayer oldPlayer, boolean alive) {
+         this.textFilteringEnabled = oldPlayer.textFilteringEnabled;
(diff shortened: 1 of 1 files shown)


patches/server/0860-Fix-player-flight-disabling-on-world-change.patch
```diff
@@ -0,0 +1,23 @@
...
+index 7b23535a680d2a8534dcb8dd87770f66fb982c13..3ee8d55ae71a8b598e131261a7f7de52106e3eab 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1721,7 +1721,11 @@ public class ServerPlayer extends Player {
+ 
+     public void restoreFrom(ServerPlayer oldPlayer, boolean alive) {
+         this.textFilteringEnabled = oldPlayer.textFilteringEnabled;
+-        this.gameMode.setGameModeForPlayer(oldPlayer.gameMode.getGameModeForPlayer(), oldPlayer.gameMode.getPreviousGameModeForPlayer());
++
++        if (this.gameMode.getGameModeForPlayer() != oldPlayer.gameMode.getGameModeForPlayer()) { // Paper - only change the player's gamemode if it's different from their current one
```
HexedHero: Paper comment is wrong.
You need a
// Paper start - (Some change comment here)
**(Code here)
(More code)**
// Paper end

electronicboy:
This was a behavioural change in vanilla, I'm not sure I feel inclined to induce a behavioural change in paper for this

patches/server/0860-Fix-player-flight-disabling-on-world-change.patch
HexedHero: That being said, the setGameModeForPlayer shouldn't be moved down or moved with indentation, simply put the if statement around the current code. Adding a single line // Paper comment on the if statement start and end.

mxson: Was using this as reference: https://github.com/PaperMC/Paper/pull/7344/commits/debbdcca320df1a8a117a9b442bc1d1a25aae2c0
Possibly that was the wrong commit to use, I've adjusted that now.

mxson:
> This was a behavioural change in vanilla, I'm not sure I feel inclined to induce a behavioural change in paper for this

Would this perhaps be better suited as a config option? Gameplay wise it's been important for the gamemodes that I've been involved with (think skyblock, teleporting between your island in the overworld then into the nether), but I could see the desire for retaining vanilla's quirks.

imDaniX:
> Would this perhaps be better suited as a config option?

I think it would, with the original behavior enabled by default. Some time ago I was using it as a basic anti-fly measure. While we don't depend on it anymore, some other small servers still may use it the same way.

PR #7414, opened 2022-01-26
Make Panda implement Sittable

Machine-Maker:
Can probably just remove that `setIsSitting` method next mc version update? idk. its annoying. I can't keep track of all of bukkit's various interfaces


HexedHero:
The API has only been here for a week so anyone using it would be changing over pretty fast so I'd say it's safe to remove the old method after a week or so.

electronicboy (approved):

Owen1212055 (approved):

PR #5191, opened 2021-02-14
Damage with any cause

NeumimTo:
Continuation from https://github.com/PaperMC/Paper/pull/4702#issuecomment-778854350

PR #7423, opened 2022-01-30
Damage with any cause

### LINKED ISSUE #7415: Provide cleaner way to damage/kill Player

Currently, we cannot kill [Player](https://papermc.io/javadocs/paper/1.18/org/bukkit/entity/Player.html) with certain [Damage Cause](https://papermc.io/javadocs/paper/1.18/org/bukkit/event/entity/EntityDamageEvent.DamageCause.html).

There is workaround for 'damaging'(not killing) player like this:
```java
player.setLastDamageCause(new EntityDamageEvent(player, EntityDamageEvent.DamageCause.DROWNING, 1000));
player.setHealth(1); // Set health to 0 will print 'Player died', not 'Player drowned'. You cannot kill Player with drowning damage.
```
But this doesn't applied if you want to 'kill' player because what actually kills the player is `player.setHealth(0);`. This will just print `Player died', not other death message like 'Player drowned', 'Player suffocated' etc.

Provide method like `Player#damage(double damage, DamageCause cause)`.  
This method will damage the player with damage cause provided, or if the damage is higher than Player's current health, it will kill the player with privided damage cause.

I think this would be more fit to [`Damagable`](https://papermc.io/javadocs/paper/1.18/org/bukkit/entity/Damageable.html) as it has `setHealth(double health)` method, but I'm not sure that whether it would be useful or not to damage living entities other than Player. If it is, `Damagable#damage(double damage, DamageCause cause)` would be what I suggest.

I currently cannot think of alternative way of this. If you have, please let me know about it.

### Other

If you think 'Oh, that could be done with this simple code! No need to add complexity to API!', let me know about that code.

I think using `Player#damage(damage, cause)` would be much cleaner than using multiple event listener, hashset etc.

TetraTheta:
Continuation of [#5191](https://github.com/PaperMC/Paper/pull/5191). This will resolve #7415. 

Even though I tried to mimic (and enhance?) the original PR, since this is my first trial of contributing to Paper, there should be some error present.

If this patch needs patch note, tell me and I will try to add information it might need as lot as possible I can think of.

Any criticism is welcomed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
     java
     `maven-publish`
     id("com.github.johnrengelman.shadow") version "7.1.2" apply false
-    id("io.papermc.paperweight.core") version "1.3.3"
+    id("io.papermc.paperweight.core") version "1.3.4"
 }
 
 allprojects {


patches/api/0364-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,52 @@
...
+diff --git a/src/main/java/org/bukkit/entity/Damageable.java b/src/main/java/org/bukkit/entity/Damageable.java
+index fc4d3bcd9b16097086fef7975274d825b65adb10..cb6ff499d79deb3d304559c8f1f7872321309010 100644
+--- a/src/main/java/org/bukkit/entity/Damageable.java
++++ b/src/main/java/org/bukkit/entity/Damageable.java
+@@ -1,6 +1,8 @@
+ package org.bukkit.entity;
+ 
+ import org.bukkit.attribute.Attribute;
++import org.bukkit.event.entity.EntityDamageEvent;
++import org.jetbrains.annotations.NotNull;
```
tahmid-23: use fully qualified names rather than imports outside of paper classes I think?

patches/server/0860-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,459 @@
...
+index 4584dcec835f97e94dcf12788b66703075dafe45..421a4b13a7965280b793f92a7229b7e7542a82a0 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -1873,7 +1873,13 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+         this.hurtMarked = true;
+     }
+ 
++    // Paper start
+     public boolean hurt(DamageSource source, float amount) {
++        return hurt(source, amount, null);
```
tahmid-23: the diff starts one line under, can move the // Paper start right above the diff

patches/server/0860-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,459 @@
...
+         } else {
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftEnderDragonPart.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftEnderDragonPart.java
+index a407e802164603c74ec1f4ad3c98bc50398954bc..8c7934408cb28fb4d52e798d0a351b296593b329 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEnderDragonPart.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEnderDragonPart.java
+@@ -4,6 +4,7 @@ import org.bukkit.craftbukkit.CraftServer;
+ import org.bukkit.entity.EnderDragon;
+ import org.bukkit.entity.EnderDragonPart;
+ import org.bukkit.entity.Entity;
++import org.bukkit.event.entity.EntityDamageEvent;
```
tahmid-23: fully qualified name

patches/server/0860-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,459 @@
...
+ import org.bukkit.craftbukkit.entity.memory.CraftMemoryMapper;
++import org.bukkit.craftbukkit.event.CraftEventFactory;
+ import org.bukkit.craftbukkit.inventory.CraftEntityEquipment;
+ import org.bukkit.craftbukkit.inventory.CraftItemStack;
+ import org.bukkit.craftbukkit.potion.CraftPotionUtil;
+@@ -72,6 +75,7 @@ import org.bukkit.entity.TippedArrow;
+ import org.bukkit.entity.Trident;
+ import org.bukkit.entity.WitherSkull;
+ import org.bukkit.entity.memory.MemoryKey;
++import org.bukkit.event.entity.EntityDamageEvent;
```
tahmid-23: fully qualified names + remove random whitespace

patches/server/0860-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,459 @@
...
++
++import io.papermc.paper.util.DamageUtil;
+ import net.minecraft.core.BlockPos;
+ import net.minecraft.core.Direction;
+ import net.minecraft.network.protocol.game.ServerboundContainerClosePacket;
+@@ -76,7 +77,6 @@ import org.bukkit.craftbukkit.block.CraftBlock;
+ import org.bukkit.craftbukkit.block.CraftBlockState;
+ import org.bukkit.craftbukkit.block.CraftBlockStates;
+ import org.bukkit.craftbukkit.block.data.CraftBlockData;
+-import org.bukkit.craftbukkit.entity.CraftEntity;
```
tahmid-23: fully qualified names + remove random whitespace + don't need to delete imports I think

NeumimTo:
nice to see that somebody picked it up after me. 

https://github.com/NeumimTo/paperprtests 
here i made simple test plugin for my former pr. 

Should be still useful.

NeumimTo:
Maybe it would be also worth to discuss whenever it would be better instead of 

```
void damage(double amount, @NotNull DamageCause damageCause, @Nullable Entity source);
```

create a damage builder

```
void damage(DamageBuilder damageBuilder);
```

```
class DamageBuilder {
   double damage;
   Entity source;
   DamageCause cause;
   ...
}
```

In future, if somebody would need to add an additional arguments to damage the method they would not need to create another overloaded damage method in damageable interface,  but only add a new field/method to damagebuilder.

TetraTheta:
Anyone insterested in this PR, please reply your opinion of [this](https://github.com/PaperMC/Paper/pull/7423#issuecomment-1048560788) opinion.  
Because, I think both current and suggested way are good, so I cannot decide what to choose.

patches/server/0860-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,456 @@
...
++}
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index 4584dcec835f97e94dcf12788b66703075dafe45..5df294a300baa552266adb53da0f6e83088eab31 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -1874,6 +1874,12 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+     }
+ 
+     public boolean hurt(DamageSource source, float amount) {
++    // Paper start
```
amaranth: Why does this change exist? You never seem to add a call to your new three argument version of `hurt`.

TetraTheta: It is used at [`CraftEnderDragonPart`](https://github.com/PaperMC/Paper/pull/7423/files#diff-e407df4e89be01749a6a72eed330111c8649ad20abc09e49a8fea72f7f80023fR213) which implements `EnderDragonPart` which extends `Damagable` which extends `Entity`.
```java
// Paper start
@Override
public void damage(double amount, DamageCause cause, Entity source) {
    getParent().damage(amount, cause, source);
}
// Paper end
```
It is used to make EnderDragonPart to be damaged with specific source(entity).

That method is also used at `CraftLivingEntity` too.

I hope I understood the concept of the original PR(#5191)

amaranth: That's `CraftEntity.damage()`, not `Entity.hurt()`. The former calls the two argument version of the latter. I think either the three argument version isn't supposed to exist or that was where it was supposed to get called although the third argument doesn't appear to _do_ anything so I'm leaning toward shouldn't exist.

TetraTheta: If I understood the definition of the `abstract class` correctly, it is somewhat hybrid of `interface` and actual `class`, thus making actual method (not structure only) can be inherited via implementation.  
`CraftEntity` doesn't override `damage()` from `Entity`, and since `Entity` is abstract class, full `Entity#damage()` will be inherited to `CraftEntity`.

Please correct me if I am wrong.

NeumimTo: The hurt method hurt(DamageSource, float) was referenced from some non spigot code, which i simply did not want to refactor in my original pr - it was not even needed. 

Therefore i only overloaded the method and made the old method call new overloaded method with null argument, that way the pr do not have to do that many changes all over the place

TetraTheta: > pr do not have to do that many changes all over the place

Well, that makes me kind of confused. When I first look at the original PR, I thought the codes were nice and clean, no clutterness present, no need to refactor the whole logic of the PR... In this case, I don't have any idea of 'how to make these codes cleaner'.

Any advise is welcomed. I must be misunderstood something wrong.

TetraTheta:
Oops, I closed PR by mistake. I will open new PR with some... improvement later.  
Next time, I should really open new PR in my own branch rather than master...

PR #7425, opened 2022-01-31
Implement World#regenerateChunk

### LINKED ISSUE #2551: Reimplement Chunk-Regeneration
We are hosting a minecraft-network with 500-1000 concurrent players and we love to use paper especially because of the performance problems in the new versions (our servers are running on 1.14.4).

Now we have a problem: With version 1.14 it is no longer posible to regenerate chunks like before: https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#regenerateChunk-int-int-

We need an implementation of that for the version 1.14.4 but our dev-team has not the time and the knowledge to do this. So we decided to give the person that can do this for us a donation of 150€. The condition is a working implementation for regenerating chunks that works like World#regenerateChunk in version 1.13.2 or better.

If you need any proof for this offer, add me on Discord: Cytoox#3169

Thanks for you time!

kennytv:
Co-authored-by: Jason Penilla <11360596+jpenilla@users.noreply.github.com>

Closes #2551

### DIFF
--- a/patches/api/0365-Implement-regenerateChunk.patch
+++ b/patches/api/0365-Implement-regenerateChunk.patch
@@ -1,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Nassim Jahnke <jahnke.nassim@gmail.com>
+Date: Sat, 5 Feb 2022 20:25:28 +0100
+Subject: [PATCH] Implement regenerateChunk
+
+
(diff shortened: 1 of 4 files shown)


kennytv:
now it's actually ready for final review

electronicboy (approved):

Angeschossen:
Just a note that this does crash the server, in any use case, since the first build of 1.19. I'll be able to provide more details next month.

MiniDigger:
please open an issue once you got the details @Angeschossen

PR #7431, opened 2022-02-01
fix randomTick blocks above y=255 and below y=0

Titaniumtown:
Continuation of #7403 (now that I actually understand what the code does):
> It seems that the 1.18 worldheight changes weren't taken into account when porting a patch that optimizes random block ticking. This PR corrects that and now allows blocks above 255 to be random ticked.

I've tested and have confirmed this PR works (tested with grass and dirt in a void world). Would love further input!


kashike:
https://github.com/PaperMC/Paper/pull/7403#issuecomment-1020968737

PR #7432, opened 2022-02-02
API for creating command sender which forwards feedback

jpenilla:


kashike (approved):

patches/server/0861-API-for-creating-command-sender-which-forwards-feedb.patch
```diff
@@ -0,0 +1,161 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/command/ServerCommandSender.java b/src/main/java/org/bukkit/craftbukkit/command/ServerCommandSender.java
+index eaff8df6c8c12c64e005a68a02e2e35ed88f757c..1c638a4b1f2c841928d8b2a7ae43e4ebb1f7eac7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/command/ServerCommandSender.java
++++ b/src/main/java/org/bukkit/craftbukkit/command/ServerCommandSender.java
+@@ -13,9 +13,15 @@ import org.bukkit.plugin.Plugin;
+ 
+ public abstract class ServerCommandSender implements CommandSender {
+     private static PermissibleBase blockPermInst;
+-    private final PermissibleBase perm;
++    public final PermissibleBase perm; // Paper
```
Machine-Maker, line 152: AT?

JRoy (approved):
troll face

molor:
Hey, can you add second argument to `Consumer` that holds the command string so we will know to what is that `Component` related? Or if it possible to do different actions inside `Consumer` with `Component` based on command that passed to `Server#dispatchCommand`, let me know how, please

electronicboy:
Consumer is a JDK class which only takes one arg, nor do we know what command is being executed when people call the send methods on that, so that's impossible. There is nothing stopping you passing the command you're running into a lambda or a class which extends Consumer and stores that info, etc

PR #7436, opened 2022-02-02
Trigger bee_nest_destroyed trigger in the correct place

Machine-Maker:
Reported here: https://forums.papermc.io/threads/datapack-using-advancements-to-run-functions-on-newly-spawned-item-not-working.119

Issue here, is that drops were being captured for the BlockDropItemEvent and not actually in the world. The `bee_nest_destroyed` trigger is expecting to be run after those drops are in the world.


jpeterik12:
I have a datapack that relies on this. Any status on merging?

electronicboy (approved):

Issue #1002, opened 2018-02-08
Container blocks causing "Skipping BlockEntity with id" when using shift-rightclick


LogGits:
##Want to know if anyone else is experiencing this or if its just me.

### Description of issue:
When a container block is held in your hand and you shift+right click on the ground or in the air it causes the console to be flooded with "[19:16:23 WARN]: Skipping BlockEntity with id" (with no id).

### Paper build number:
Paper version git-Paper-1322 (MC: 1.12.2)

### What behaviour is expected:
It shouldn't print spam to console when right clicking with a container block in hand.
### What behaviour is observed:
console spam when right clicking with a container block in hand.
### Actions to reproduce (if known):
Hold container block in your hand (testing found hoppers caused it, but other container blocks also caused it)

### Link/paste of stack trace
Spams:
```
[19:16:23 WARN]: Skipping BlockEntity with id 
[19:16:23 WARN]: Skipping BlockEntity with id 
[19:16:23 WARN]: Skipping BlockEntity with id 
[19:16:23 WARN]: Skipping BlockEntity with id 
[19:16:23 WARN]: Skipping BlockEntity with id 
[19:16:24 WARN]: Skipping BlockEntity with id 
[19:16:24 WARN]: Skipping BlockEntity with id 
```

LogGits:
Nvm figured out the issue it was with a plugin i will alert the dev.

PR #7450, opened 2022-02-06
Updated Upstream (Bukkit/CraftBukkit/Spigot)

yannicklamprecht:

Bukkit Changes:
04c7e13c #719: Add Player Profile API
71564210 SPIGOT-6910: Add BlockDamageAbortEvent

CraftBukkit Changes:
bdac46b0 SPIGOT-6782: EntityPortalEvent should not destroy entity when setTo() uses same world as getFrom()
8f361ece #1002: Add Player Profile API
911875d4 Increase outdated build delay
e5f8a767 SPIGOT-6917: Use main scoreboard for /trigger
a672a531 Clean up callBlockDamageEvent
8e1bdeef SPIGOT-6910: Add BlockDamageAbortEvent

Spigot Changes:
7fbc6a1e Rebuild patches

### DIFF
--- a/patches/api/0006-Timings-v2.patch
+++ b/patches/api/0006-Timings-v2.patch
@@ -2782,10 +2782,10 @@ index 0000000000000000000000000000000000000000..5989ee21297935651b0edd44b8239e65
 +    }
 +}
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index 072bcbbbf167e02ce5014b672f2cf63122d29b21..de7f303bce9e2454eaec12131cd1439a54281c7e 100644
+index dff191d4b5f8866837500aca955e72d0fbe1d82c..7135cf6ac532d8f9b0958830454db92d2cec8e87 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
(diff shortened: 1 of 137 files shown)


yannicklamprecht:
I guess that the reason is that no one created an upstream update pr for more than 3 days is the decision how to handle the upstream PlayerProfile API addition without breaking the Paper one.

electronicboy:
Yeaaa, that was kinda a "somebody can figure that out" and not gotten around to looking into it
I forget how ABI works here, but, I thought that you can make it use a higher class, so, guess that we could maybe make papers extend upstream and see if that works, or, just yank theirs out, not fond of that idea but there is much more precedent set on our API vs theirs, and the fact that the class names line up perfectly is more than an annoyance...

yannicklamprecht:
> Yeaaa, that was kinda a "somebody can figure that out" and not gotten around to looking into it
> I forget how ABI works here, but, I thought that you can make it use a higher class, so, guess that we could maybe make papers extend upstream and see if that works, or, just yank theirs out, not fond of that idea but there is much more precedent set on our API vs theirs, and the fact that the class names line up perfectly is more than an annoyance...

From the naming it should work. I'll address that.

EDIT:

OfflinePlayer has the only clashing method

PR #7459, opened 2022-02-09
Stop large look changes from crashing the server

MWHunter:
If a plugin teleports a player with a large pitch or yaw, the server will crash due to some weird looping code.  I've replaced this while loop with identical code that accomplishes the same math.  I've verified this code to be identical with the following program: https://gist.github.com/MWHunter/3690afc61f290314c4e1eb8f4a2c28af


kennytv (changes requested):
Just remove all the commented out parts so the diff is cleaner and it properly conflicts on change

PR #7465, opened 2022-02-11
Added CommandPostprocessEvent

LoreSchaeffer:
Added CommandPostprocessEvent an event that is called after a command is executed with the result of the command

### DIFF
--- a/patches/api/0366-Added-CommandPostprocessEvent.patch
+++ b/patches/api/0366-Added-CommandPostprocessEvent.patch
@@ -1,0 +1,124 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: LoreSchaeffer <lm.lorenzo@outlook.it>
+Date: Fri, 11 Feb 2022 13:31:13 +0100
+Subject: [PATCH] Added CommandPostprocessEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/server/CommandPostprocessEvent.java b/src/main/java/io/papermc/paper/event/server/CommandPostprocessEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..e00afe59cd3ac499e03be28332541c33a7047eb2
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/server/CommandPostprocessEvent.java
+@@ -0,0 +1,74 @@
++package io.papermc.paper.event.server;
++
++import org.bukkit.command.CommandSender;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called after a player executes a command.
++ */
++public class CommandPostprocessEvent extends Event {
++    private static final HandlerList handlers = new HandlerList();
++    private final CommandSender sender;
++    private final String command;
++    private final String[] args;
++    private final boolean success;
++
++    public CommandPostprocessEvent(@NotNull final CommandSender sender, @NotNull final String command, @NotNull String[]args, final boolean success) {
++        this.sender = sender;
++        this.command = command;
++        this.args = args;
++        this.success = success;
++    }
++
++    /**
++     * Gets the CommandSender that executed the command.
++     *
++     * @return CommandSender that executed the command
++     */
++    @NotNull
++    public CommandSender getCommandSender() {
++        return sender;
++    }
++
++    /**
++     * Gets the command that the player has executed.
++     *
++     * @return Command the player has executed
++     */
++    @NotNull
++    public String getCommand() {
++        return this.command;
++    }
++
++    /**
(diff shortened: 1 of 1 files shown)


patches/api/0366-Added-CommandPostprocessEvent.patch
```diff
@@ -0,0 +1,124 @@
...
+         }
+         // Paper end
+ 
++        boolean success = false; // Paper
++
+         try {
+             try (co.aikar.timings.Timing ignored = target.timings.startTiming()) { // Paper - use try with resources
+             // Note: we don't return the result of target.execute as thats success / failure, we return handled (true) or not handled (false)
+-            target.execute(sender, sentCommandLabel, Arrays.copyOfRange(args, 1, args.length));
++            success = target.execute(sender, sentCommandLabel, Arrays.copyOfRange(args, 1, args.length)); // Paper
```
ryantheleach, line 109: Read the comment directly above this line. this is NOT the success value! It's merely whether the command was handled or not handled.

ryantheleach:
I think you need a slightly stricter contract for this to be useful to people.

You have provided neither the Success value as you call it in the API, nor the Result as you call it in the PR.

Can plugins rely on the output of this event never changing in the future? 

Like, is it a Informational only event, and thus is there a guarantee it won't be mutated in the future?

Or is the capability there for this to control CommandBlock decision making, by changing the success values?

I think you could add a lot more to this event, to make it more useful either way.



Your success value isn't whether the command succeeded, it's whether the command was handled.

`Success` has a definition in vanilla Minecraft, that I believe should be exposed here, if possible, in addition to whether the command was handled.

`Result` also has a definition in vanilla Minecraft, but this will be much harder to expose, as it can have varying types, including NBT. That said, it would provide a lot of compatibility if the result of any vanilla command was able to be represented.

LoreSchaeffer:
You are right, i said that badly about the success / failure thing it was clearly handled / not handled.

My thought was to make an informational only event, to know when a command finish its execution and to know if it has been handled or not.

If you have any suggestions on how to make this better or more useful i can try to code them

electronicboy:
post process is somewhat of a flawed name given that it's post-execution, not processing, processing was already done a good moment ago in the scheme of things

This event will not be fired if an exception is thrown by the command, which, creates many issues if this is solved in that you have no understanding of why it failed, etc

The nature of "Success" is also flawed here as already pointed out

LoreSchaeffer:
> post process is somewhat of a flawed name given that it's post-execution, not processing, processing was already done a good moment ago in the scheme of things
> 
> This event will not be fired if an exception is thrown by the command, which, creates many issues if this is solved in that you have no understanding of why it failed, etc
> 
> The nature of "Success" is also flawed here as already pointed out

So changing the event to "CommandPostexecuteEvent" an handling the exception inside the dispatch method in SimpleCommandMap in this way could work better?

```java
boolean handled = false; // Paper

        try {
            try (co.aikar.timings.Timing ignored = target.timings.startTiming()) { // Paper - use try with resources
            // Note: we don't return the result of target.execute as thats success / failure, we return handled (true) or not handled (false)
            handled = target.execute(sender, sentCommandLabel, Arrays.copyOfRange(args, 1, args.length)); // Paper
            } // target.timings.stopTiming(); // Spigot // Paper
        } catch (CommandException ex) {
            server.getPluginManager().callEvent(new ServerExceptionEvent(new ServerCommandException(ex, target, sender, args))); // Paper
            // Paper start
            CommandPostexecuteEvent event = new CommandPostexecuteEvent(sender, target.getName(), Arrays.copyOfRange(args, 1, args.length), false, ex);
            server.getPluginManager().callEvent(event);
            // Paper end
            //target.timings.stopTiming(); // Spigot // Paper
            throw ex;
        } catch (Throwable ex) {
            //target.timings.stopTiming(); // Spigot // Paper
            String msg = "Unhandled exception executing '" + commandLine + "' in " + target;
            server.getPluginManager().callEvent(new ServerExceptionEvent(new ServerCommandException(ex, target, sender, args))); // Paper
            // Paper start
            CommandException ex1 = new CommandException(msg, ex);
            CommandPostexecuteEvent event = new CommandPostexecuteEvent(sender, target.getName(), Arrays.copyOfRange(args, 1, args.length), false, ex1);
            server.getPluginManager().callEvent(event);
            throw ex1;
            // Paper end
        }

        // Paper start
        CommandPostexecuteEvent event = new CommandPostexecuteEvent(sender, target.getName(), Arrays.copyOfRange(args, 1, args.length), handled);
        server.getPluginManager().callEvent(event);
        // Paper end
```

Machine-Maker:
If something like this exists, it should be part of paper’s mojang api I think. And use brig’s ParseResults and what not.

LoreSchaeffer:
> If something like this exists, it should be part of paper’s mojang api I think. And use brig’s ParseResults and what not.

I'm not very familiar with brigadier, any help is appreciated

PR #7467, opened 2022-02-12
Log exceptions thrown during chat processing

### LINKED ISSUE #7466: AsyncChatEvent#renderer() swallows exceptions thrown inside it
### Expected behavior

Exceptions thrown inside `renderer()` call to be logged.

### Observed/Actual behavior

The exception is being swallowed somewhere, nothing was logged.

### Steps/models to reproduce

1. Throw an exception inside `AsyncChatEvent#renderer()`:
```java
@EventHandler
public void onAsyncChat(AsyncChatEvent event) {
    event.renderer((source, sourceDisplayName, message, viewer) -> {
        throw new IllegalArgumentException();
    });
}
```
2. Send a chat message in-game, observe nothing being logged.

### Plugin and Datapack List

Test plugin only

### Paper version

[02:47:56 INFO]: Checking version, please wait...
[02:47:57 INFO]: This server is running Paper version git-Paper-187 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 0cc2503)

### Other

_No response_

lynxplay:
Previously the async chat executor service would take chat handling
using the #submit method, which wraps the logic in a future task.
The future takes full ownership of the task, including any potential
exception, meaning that the uncaught exception handler never gets
notified about potential exceptions thrown during async chat logic.

As the chat task does neither need to be cancelled nor returns something
required later on, this commit moves from #submit to #execute, skipping
any future task creation. This properly propagates any exception upwards
to the worker thread in the executor service, allowing the server to
catch and deobfuscate the exception before properly logging it into the
servers console.

Resolves: #7466

### DIFF
--- a/patches/server/0863-Log-exceptions-thrown-during-chat-processing.patch
+++ b/patches/server/0863-Log-exceptions-thrown-during-chat-processing.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bjarne Koll <lynxplay101@gmail.com>
+Date: Sat, 12 Feb 2022 03:20:36 +0100
+Subject: [PATCH] Log exceptions thrown during chat processing
+
+Previously the async chat executor service would take chat handling
+using the #submit method, which wraps the logic in a future task.
+The future takes full ownership of the task, including any potential
+exception, meaning that the uncaught exception handler never gets
+notified about potential exceptions thrown during async chat logic.
+
+As the chat task does neither need to be cancelled nor returns something
+required later on, this commit moves from #submit to #execute, skipping
+any future task creation. This properly propagates any exception upwards
+to the worker thread in the executor service, allowing the server to
+catch and properly log the exception to the console.
+
+diff --git a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java b/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
+index 8a0ced3f9b9099913ade4b71181aff6cafbc4ee6..21588ce5a408fed3454c317b56c05439ad3af27d 100644
+--- a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
++++ b/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
+@@ -31,7 +31,7 @@ public class ServerboundChatPacket implements Packet<ServerGamePacketListener> {
+     public void handle(final ServerGamePacketListener listener) {
+         if ( !this.message.startsWith("/") )
+         {
+-            ServerboundChatPacket.executors.submit( new Runnable()
++            ServerboundChatPacket.executors.execute( new Runnable() // Paper - Use #execute to propagate exceptions up instead of swallowing them
+             {
+ 
+                 @Override


jpenilla:
I feel like this belongs in it's own patch (or in some general logging fixes patch if one exists). For context the default uncaught exception handler will print stacktraces, that patch is only replacing it with one that uses a logger so it goes through the deobfuscator.

The proposed change here impacts whether the exception is logged at all, not whether it's deobfuscated.

lynxplay:
The deobf patch adds the entire uncaught exception handler to the executor service as well tho:
https://github.com/PaperMC/Paper/blob/0cc2503b88343c4d10d9e6ecf7592d56762b4cae/patches/server/0417-Deobfuscate-stacktraces-in-log-messages-crash-report.patch#L590-L591

//edit[0]. Tho I guess it would just hit the normal exception handler without the previous patch line.

lynxplay:
Alright, my bad brain lag, I will open a new patch for this (and look around a bit for a logging patch)

lynxplay:
Updated by moving the changes into their own patch. If (in the future) further plain logging changes like this need to be implemented, this patch may simply be renamed and used as a general purpose patch.

patches/server/0863-Allow-async-chat-to-leverage-exception-handler.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java b/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
+index 8a0ced3f9b9099913ade4b71181aff6cafbc4ee6..0619b8d5f02d9875a0f88cf2f6ae8880ef6684f6 100644
+--- a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
++++ b/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
+@@ -31,7 +31,7 @@ public class ServerboundChatPacket implements Packet<ServerGamePacketListener> {
+     public void handle(final ServerGamePacketListener listener) {
+         if ( !this.message.startsWith("/") )
+         {
+-            ServerboundChatPacket.executors.submit( new Runnable()
++            ServerboundChatPacket.executors.execute( new Runnable() // Paper - Deobfuscate stacktrace by calling #execute to leverage uncaught exception handler
```
jpenilla:
```suggestion
+            ServerboundChatPacket.executors.execute( new Runnable() // Paper - Use execute to propagate exceptions up instead of swallowing them
```

patches/server/0863-Allow-async-chat-to-leverage-exception-handler.patch
```diff
@@ -0,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bjarne Koll <lynxplay101@gmail.com>
+Date: Sat, 12 Feb 2022 03:20:36 +0100
+Subject: [PATCH] Allow async chat to leverage exception handler
```
jpenilla:
```suggestion
Subject: [PATCH] Log exceptions thrown during chat processing
```

patches/server/0863-Allow-async-chat-to-leverage-exception-handler.patch
```diff
@@ -0,0 +1,31 @@
...
+using the #submit method, which wraps the logic in a future task.
+The future takes full ownership of the task, including any potential
+exception, meaning that the uncaught exception handler never gets
+notified about potential exceptions thrown during async chat logic.
+
+As the chat task does neither need to be cancelled nor returns something
+required later on, this commit moves from #submit to #execute, skipping
+any future task creation. This properly propagates any exception upwards
+to the worker thread in the executor service, allowing the server to
+catch and deobfuscate the exception before properly logging it into the
```
jpenilla: deobfuscation isn't relevant to this patch and mentioning it makes the actual purpose (ensuring exceptions are propagated up and logged) less clear and concise

jpenilla (changes requested):

lynxplay:
Removed any traces of the previous commit message and body related to deobfuscation :+1:

jpenilla (approved):

PR #7468, opened 2022-02-12
Use access transformers for player profile API

lynxplay:
During the previous upstream update, craftbukkit classes and methods had
their modifiers changed through in-patch modifications, which is an
outdated way of modifying access modifiers in server internals.

This commit moves all of the in-patch access transformers into the
paper.at file and unites them with the single existing access
transformer used for the player profile api.


kennytv (approved):
thanks, I was too lazy for that 🛌

kennytv:
I added more lazy diffs =^]

lynxplay:
Cleaned those up for you too :sweat_smile:

electronicboy (approved):

PR #7469, opened 2022-02-12
Ignore BaseSpawner impossible spawn tick

dannyball710:
Some server use spawner block to make custom block. That spawner never spawn entity , but still scanning all player every tick.
If `spawnCount` or `maxNearbyEntities` <= 0 ,then will stop to try spawn entity

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,9 +18,9 @@ jobs:
                 java: [17]
             fail-fast: true
         steps:
-            - uses: actions/checkout@v2.4.0
+            - uses: actions/checkout@v2.5.0
             - name: JDK ${{ matrix.java }}
-              uses: actions/setup-java@v3.1.0
+              uses: actions/setup-java@v3.6.0
               with:
                   java-version: ${{ matrix.java }}
                   cache: 'gradle'


Owen1212055 (approved):
Tested with ~7000 furnace minecarts, it should be noted that there really isn't any gain visible until there is a large number of players.

``/summon spawner_minecart ~ ~ ~ {SpawnCount:0,RequiredPlayerRange:0,MinSpawnDelay:0}``

electronicboy (approved):

kennytv:
Plz rebase, making prs from your fork's master branch makes that ugly for us

patches/server/0866-Ignore-impossible-spawn-tick.patch
```diff
@@ -0,0 +1,23 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/BaseSpawner.java b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+index 03726227fdd60e9cf77213d50184abff438e01ef..295fd153bc6d14a6dfecceaaf6c04af54b3fe6aa 100644
+--- a/src/main/java/net/minecraft/world/level/BaseSpawner.java
++++ b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+@@ -76,6 +76,12 @@ public abstract class BaseSpawner {
+     }
+ 
+     public void serverTick(ServerLevel world, BlockPos pos) {
++        // Paper start - Ignore impossible spawn tick
++        if (spawnCount <= 0 || maxNearbyEntities <= 0) {
```
Lulu13022002: Inline this in one line when it's just a return with low complexity to avoid large diff

PR #7476, opened 2022-02-13
automatically disable online mode when using either bungee or velocity

yannicklamprecht:
I don't know if that is a wanted change.

Use case:
Many Velocity or Bungee/Waterfall user fail to disable online-mode.

Feel free to share some insides why this PR doesn't work out. Side-effects or something else.

### DIFF
--- a/patches/server/0867-automatically-disable-online-mode-when-using-either-.patch
+++ b/patches/server/0867-automatically-disable-online-mode-when-using-either-.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Yannick Lamprecht <yannicklamprecht@live.de>
+Date: Sun, 13 Feb 2022 13:13:54 +0100
+Subject: [PATCH] automatically disable online mode when using either bungee or
+ velocity
+
+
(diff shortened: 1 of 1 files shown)


ryantheleach:
the extra step IMO encourages players to actually understand that their server is vulnerable to cracked clients connecting directly...

OTOH it means if disabling proxy support, it would spring back to online mode.

ryantheleach:
This is still required for modern forwarding with velocity right? they mention setting a second online-mode settings specifically for velocity in the paper.yml settings.

https://docs.velocitypowered.com/en/latest/users/player-info-forwarding.html#:~:text=modern%20forwarding%20is%20a%20Velocity,for%20Minecraft%201.13%20or%20higher.

PureGero:
The server prints this message while in offline mode:
```
**** SERVER IS RUNNING IN OFFLINE/INSECURE MODE!
The server will make no attempt to authenticate usernames. Beware.
Whilst this makes it possible to use BungeeCord, unless access to your server is properly restricted, it also opens up the ability for hackers to connect with any username they choose.
Please see http://www.spigotmc.org/wiki/firewall-guide/ for further information.
To change this, set "online-mode" to "true" in the server.properties file.
```

Users may not understand why this warning has appeared if all they did was enable bungeecord.

electronicboy:
disabling security features should be an explicit thing, IMHO

PR #7479, opened 2022-02-15
.

ghost:


JRoy:
Editing the properties file manually is not the correct way to update the gradle wrapper, you need to use the wrapper task. It's clear you manually edited the file from the web editor default branch name and the lack of any change to the wrapper jar when it had changes in 7.4.

Additionally the team is capable of updating the gradle wrapper when we deem it necessary, we don't accept PRs like this in the same way we don't accept trivial upstream update PRs.

yannicklamprecht:
@JRoy i had side effet with Gradle 7.4 in my fork. It failed running apply patches because Gradle detected an already locked file in Paper project itself. Likely an issue to be fixed in paperweight, by either removing that lock file or having a possibility to ignore that.

PR #7483, opened 2022-02-16
Be able to set how many threads are allocated for handling chat

OOP-778:
Since I don't even know when chat has been handled by CachedThreadPool which from my knowledge should be only used in cases where you do a lot of stuff await for results then close it.
I really think making the chat handled by a configurable ForkJoinPool gives a faster experience for the handling of the messages and improves the performance of machine & server.

### DIFF
--- a/patches/server/0867-Be-able-to-set-how-many-threads-are-allocated-for-ha.patch
+++ b/patches/server/0867-Be-able-to-set-how-many-threads-are-allocated-for-ha.patch
@@ -1,0 +1,59 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: oop-778 <oskardhavel@gmail.com>
+Date: Wed, 16 Feb 2022 23:22:25 +0200
+Subject: [PATCH] Be able to set how many threads are allocated for handling
+ chat
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index 61ea1c9881ea30c05580044af9496a65fe95d94e..8c72f7e4df1e20ebb52c338f2441a0fdbdc08a35 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+@@ -650,4 +650,9 @@ public class PaperConfig {
+     private static void timeCommandAffectsAllWorlds() {
+         timeCommandAffectsAllWorlds = getBoolean("settings.time-command-affects-all-worlds", timeCommandAffectsAllWorlds);
+     }
++
++    public static int chatThreads;
++    private static void chatThreads() {
++        chatThreads = getInt("settings.chat-threads", 4);
++    }
+ }
+diff --git a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java b/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
+index 21588ce5a408fed3454c317b56c05439ad3af27d..dc4dbc76e82e29f49239c5cd3f518dcc4b063fdd 100644
+--- a/src/main/java/net/minecraft/network/protocol/game/ServerboundChatPacket.java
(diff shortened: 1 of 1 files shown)


electronicboy:
I mean, the big aspect here is gonna be the resizing factor, e.g. the cached thread pool can drop to the minimum number of threads which depending on servers can widly swing from a message every minute towards a flood in one ago

Given that we're not actually using the features of an FJP it seems pretty redundant to use one vs just tweaking the thread pool to retain a thread or two or using a fixed size thread

OOP-778:
Well I think that FJP actually is better for use here, since fixed sized thread pool executor would just reject a task at some point. While FJP would never. Resizing factor is gonna be per server, honestly I'd rather have it take those threads than having it to continously grow I think other would also agree on that.

electronicboy:
I don't see where a fixed size pool would reject tasks if setup properly vs a FJP?

OOP-778:
Right we could actually move it into a fixed thread pool. Or I guess we could also do a way of setting x threads per x player. ex. 100 players = 2 threads, 200 = 4, etc. I feel like that would be much better too.

electronicboy:
issue is that fixed size has its own set of issues, am not sure if there is a fixed sized pool which can resize cleanly, which is why I kinda suggested on tweaking the config for the cached pool;

right now the entire thing can close off iirc, we could set a minimum number of threads to stick around vs requiring one to be fired up as people start chatting, etc, can also tweak how long a thread is allowed to idle for to allow for some sane time window before stuff starts scaling back

OOP-778:
So do I just change the ForkJoinPool into fixed thread pool and and edit this PR?

electronicboy:
Not too sure as there is still merits for the cached pool in that it can resize, but, maybe we should be taking advantage of the fact that you can control some aspects of the resize, e.g. core thread count, thread keepalive

an FJP for the most part acts the same as a cached pool when you're not using the work-stealing aspects, so am not exactly too sure where you'd be seeing any improvements using one

would be interested to see what others think on this though

patches/server/0867-Be-able-to-set-how-many-threads-are-allocated-for-ha.patch
```diff
@@ -0,0 +1,59 @@
...
++        PaperConfig.chatThreads,
++        pool -> {
++            final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
++            worker.setName(String.format("Async Chat Thread - #%s", worker.getPoolIndex()));
++            return worker;
++        },
++        new DefaultUncaughtExceptionHandler(MinecraftServer.LOGGER),
++        true
++    ); // Paper
++    public void handle(ServerGamePacketListener listener) {
```
Owen1212055, line 56: You need to mark the entire diff in paper comments (paper start/paper end), not just the statement.

Additionally, there's no need to change the handle method.

patches/server/0867-Be-able-to-set-how-many-threads-are-allocated-for-ha.patch
```diff
@@ -0,0 +1,59 @@
...
+@@ -1,7 +1,12 @@
+ package net.minecraft.network.protocol.game;
+
++import com.destroystokyo.paper.PaperConfig;
++import java.util.concurrent.ForkJoinPool;
++import java.util.concurrent.ForkJoinWorkerThread;
++import net.minecraft.DefaultUncaughtExceptionHandler;
+ import net.minecraft.network.FriendlyByteBuf;
+ import net.minecraft.network.protocol.Packet;
++import net.minecraft.server.MinecraftServer;
```
Owen1212055, line 35: I would use fully qualified imports

Owen1212055 (changes requested):

electronicboy:
https://gist.github.com/electronicboy/9e89c9898e0ec2347ece8adc292f8e5b - I haven't tested this, but, this seems like a less invasive method which does what's needed here, (and a bit more)
Setting the core size would allow you to have X number of chat threads always sitting around, as well as allowing to limit the size of the thread pool rather than the default of "unlimited", setting max == core, would also simulate a "fixed" size pool

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/7878

PR #7486, opened 2022-02-18
Call bucket events for cauldrons

Machine-Maker:
the BucketFill and Empty events should be called for cauldron interactions.

I had to create a bit of a hack-around to track the BlockFace, that's what the static field on AbstractCauldronBlock is, but I think the only alternative would be to overload the interact method in CauldronInteraction which would mean a way bigger diff.


Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased for 1.19

Machine-Maker:
Rebased for 1.21.1

lynxplay (approved):

PR #7490, opened 2022-02-19
Add CPU name and JVM vendor info in timings

Phoenix616:
This tries to get the CPU name (as requested here: aikar/timings#80, should be working on the most common Linux distributions as well as on Windows) as well as vendor info of the JVM to add to the exported timings data. (Requires changes in the timings website repo to display: https://github.com/aikar/timings/pull/137)

Imo. adding a full blown extra library like [OSHI](https://github.com/oshi/oshi) just to get the CPU info is a bit overkill if most servers will run on linux or windows anyways and the methods used here should produce proper results on the majority of systems.

### DIFF
--- a/patches/server/0012-Timings-v2.patch
+++ b/patches/server/0012-Timings-v2.patch
@@ -163,10 +163,10 @@ index 0000000000000000000000000000000000000000..b47b7dce26805badd422c1867733ff4b
 +}
 diff --git a/src/main/java/co/aikar/timings/TimingsExport.java b/src/main/java/co/aikar/timings/TimingsExport.java
 new file mode 100644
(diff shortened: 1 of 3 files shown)


kyngs:
I would like to add, that OSHI should be present in the Notchian server. Therefore, it might be a relevant option.

electronicboy:
Yea, oshi is already in paper, it's something I was actually looking into but had some headaches tryna work out what was what in terms of info, and then tryna figure out what was wanted, etc

Phoenix616:
TIL. I can look into using that instead then I guess.

Phoenix616:
Does this need any further changes?

electronicboy (approved):

kennytv (approved):

PR #7503, opened 2022-02-24
Fix setting the view distance to 32 with the API

froobynooby:
Currently, setting the view distance to 32 with `World#SetViewDistance` actually sets the view distance to 31. This is because the view distance has 1 added to it, is clamped between 3 and 33, and then is clamped again between 2 and 32. This PR removes the second clamp - which I don't think should be there?


froobynooby:
Rebased

Owen1212055 (approved):
Tested and looks good.

That 2nd clamp indeed is the issue here as it already has +1 added to it.

electronicboy (approved):

kennytv (approved):

PR #7505, opened 2022-02-24
Added byte array serialization/deserialization for PersistentDataContainers

CyanNex:
Added methods for converting a PersistentDataContainer to a byte array (using NBT format) and back.


patches/server/0878-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,35 @@
...
++    public byte[] serializeToBytes() throws java.io.IOException {
++        net.minecraft.nbt.CompoundTag root = this.toTagCompound();
++        java.io.ByteArrayOutputStream byteArrayOutput = new java.io.ByteArrayOutputStream();
++        java.io.DataOutput dataOutput = new java.io.DataOutputStream(byteArrayOutput);
++        net.minecraft.nbt.NbtIo.write(root, dataOutput);
++        return byteArrayOutput.toByteArray();
++    }
++
++    @Override
++    public void readFromBytes(byte[] bytes) throws java.io.IOException {
```
Machine-Maker: I feel like this method should clear the pdc first, or at least provide an option to clear it before writing to it.

CyanNex: Yea that's probably a good idea

electronicboy (approved):

Machine-Maker (approved):

CyanNex:
Rebased on master

patches/server/0925-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,38 @@
...
+@@ -168,5 +168,26 @@ public class CraftPersistentDataContainer implements PersistentDataContainer {
+ 
+         return this.customDataTags.containsKey(key.toString());
+     }
++
++    @Override
++    public byte[] serializeToBytes() throws java.io.IOException {
++        net.minecraft.nbt.CompoundTag root = this.toTagCompound();
++        java.io.ByteArrayOutputStream byteArrayOutput = new java.io.ByteArrayOutputStream();
++        java.io.DataOutput dataOutput = new java.io.DataOutputStream(byteArrayOutput);
```
Lulu13022002: Close the data output stream to avoid memory leak (maybe in a try with block but it's just an opinion)

patches/server/0925-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,38 @@
...
++        net.minecraft.nbt.NbtIo.write(root, dataOutput);
++        return byteArrayOutput.toByteArray();
++    }
++
++    @Override
++    public void readFromBytes(byte[] bytes, boolean clear) throws java.io.IOException {
++        if (clear) {
++            this.clear();
++        }
++        java.io.DataInput dataInput = new java.io.DataInputStream(new java.io.ByteArrayInputStream(bytes));
```
Lulu13022002: Same here see above

Lulu13022002 (commented):
Looks good just some comment about the fact that toByteArray will copy the bytes so it's safe to close the stream

lynxplay (approved):

patches/api/0389-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,50 @@
...
+     boolean has(@NotNull NamespacedKey key);
++
++    /**
++     * Serialize this {@link PersistentDataContainer} instance to a
++     * byte array.
++     *
++     * @return a binary represenation of this container
++     * @throws java.io.IOException if we fail to write this container to a byte array
++     */
++    byte[] serializeToBytes() throws java.io.IOException;
```
electronicboy: needs nullability annotation

patches/server/0933-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,41 @@
...
+@@ -168,5 +168,29 @@ public class CraftPersistentDataContainer implements PersistentDataContainer {
+ 
+         return this.customDataTags.containsKey(key.toString());
+     }
++
++    @Override
++    public byte[] serializeToBytes() throws java.io.IOException {
++        net.minecraft.nbt.CompoundTag root = this.toTagCompound();
++        try (java.io.ByteArrayOutputStream byteArrayOutput = new java.io.ByteArrayOutputStream()) {
++            try (java.io.DataOutputStream dataOutput = new java.io.DataOutputStream(byteArrayOutput)) {
```
Lulu13022002: You don't need to nest two try with block, java will close all the parent stream from the lowest stream which mean when the data output stream will be close the byte array one will be also automatically close, so no need to close two time the byte stream

patches/server/0933-Added-byte-array-serialization-deserialization-for-P.patch
```diff
@@ -0,0 +1,41 @@
...
++    @Override
++    public void readFromBytes(byte[] bytes, boolean clear) throws java.io.IOException {
++        if (clear) {
++            this.clear();
++        }
++        try (java.io.DataInputStream dataInput = new java.io.DataInputStream(new java.io.ByteArrayInputStream(bytes))) {
++            net.minecraft.nbt.CompoundTag compound = net.minecraft.nbt.NbtIo.read(dataInput);
++            for (String key : compound.getAllKeys()) {
++                this.put(key, compound.get(key));
++            }
```
Lulu13022002: You can simplify these lines to this.putAll(compound); to reduce diff

Lulu13022002 (approved):

PR #7507, opened 2022-02-24
Fix issues with CreeperIgniteEvent

### LINKED ISSUE #7461: Bug or inconclusive documentation on CreeperIgniteEvent
### Expected behavior

https://papermc.io/javadocs/paper/1.17/com/destroystokyo/paper/event/entity/CreeperIgniteEvent.html

Expecting this event to fire anytime the fuse of a creeper is ignited.

### Observed/Actual behavior

The event fires when a creeper is manually ignited, via right-clicking on it with a flint and steel item. It does not fire when a creeper "naturally fuses" by approaching the targeted player and being within the minimum range.

### Steps/models to reproduce

    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = false)
    public void onCreeperIgnite(@NotNull CreeperIgniteEvent event) {

        System.out.println("CreeperIgniteEvent");
}

### Plugin and Datapack List

Nothing installed that could possibly interfere with that test code

### Paper version

This server is running Paper version git-Paper-399 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: https://github.com/PaperMC/Paper/commit/d4318a6ebca2a3f2859ca0ef1b274192633c7d5e on ver/1.17.1)
You are 9 version(s) behind

### Other

If this is intentional behaviour of the event, please either update the documentation to reflect this behaviour, or add the firing of this event for the latter case mentioned earlier.

Thank you

Machine-Maker:
Clarified documentation for when the event is fired, and
prevented the event from firing when a Creeper is loaded.

Fixes https://github.com/PaperMC/Paper/issues/7461

### DIFF
--- a/patches/api/0142-Add-More-Creeper-API.patch
+++ b/patches/api/0142-Add-More-Creeper-API.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add More Creeper API
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/entity/CreeperIgniteEvent.java b/src/main/java/com/destroystokyo/paper/event/entity/CreeperIgniteEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..ff10251b6ded533b08048ec533525176eff03707
+index 0000000000000000000000000000000000000000..e9768c919a1860881802ab68eff559874590ac1c
 --- /dev/null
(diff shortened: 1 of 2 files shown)


Owen1212055 (approved):
Tested and seems to be good.

Machine-Maker:
Rebased for 1.18.2

Machine-Maker:
Rebased for 1.19.1

kennytv (approved):

PR #7510, opened 2022-02-26
Clarify exception messages for invalid view distances

NoahvdAa:
I've seen a few cases in #paper-help where people set their view/simulation distance to a number below two, only to then get an unclear exception message. This should hopefully help in these scenarios.

### DIFF
--- a/patches/server/0865-Replace-player-chunk-loader-system.patch
+++ b/patches/server/0865-Replace-player-chunk-loader-system.patch
@@ -123,7 +123,7 @@ index 66501c3b0eb92d946ef77bbd3f36ebcc0d3023af..153f07bac06093b43a1f5b0f8e1a46ff
  }
 diff --git a/src/main/java/io/papermc/paper/chunk/PlayerChunkLoader.java b/src/main/java/io/papermc/paper/chunk/PlayerChunkLoader.java
 new file mode 100644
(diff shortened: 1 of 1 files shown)


patches/server/0868-Replace-player-chunk-loader-system.patch
```diff
@@ -326,7 +326,7 @@ index 0000000000000000000000000000000000000000..9ad84dc6f0463dbbbdd53723edcc2a3e
 +
 +    public void setSendDistance(final int distance) {
 +        if (distance != -1 && (distance < MIN_VIEW_DISTANCE || distance > MAX_VIEW_DISTANCE + 1)) {
-+            throw new IllegalArgumentException(Integer.toString(distance));
++            throw new IllegalArgumentException("Send distance must be a number between " + MIN_VIEW_DISTANCE + " and " + MAX_VIEW_DISTANCE + ", or -1, got: " + distance);
```
kennytv: missing a +1

kennytv (approved):

patches/server/0868-Replace-player-chunk-loader-system.patch
NoahvdAa: Thanks, fixed.

electronicboy (approved):

NoahvdAa:
Rebased for 1.18.2

PR #5567, opened 2021-05-03
Add PlayerGlassBottleFillEvent

JonMagon:
Recreate PaperMC/Paper#5152, rebased

PR #7513, opened 2022-02-27
Add BottleFillEvents for players and dispensers

Machine-Maker:
Recreation of https://github.com/PaperMC/Paper/pull/5567

If acceptable, I'd like there to be a common interface between `PlayerFillBottleEvent` and `BlockFillBottleEvent` just to cut down on the duplicate documentation. Both are Cancellable, and share several common methods. But I don't think there's another event interface that does that, so idk if its something the API should add.

### DIFF
--- a/patches/api/0421-Add-FillBottleEvents-for-player-and-dispenser.patch
+++ b/patches/api/0421-Add-FillBottleEvents-for-player-and-dispenser.patch
@@ -1,0 +1,208 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 26 Feb 2022 16:43:33 -0800
+Subject: [PATCH] Add FillBottleEvents for player and dispenser
+
+Adds 2 events, PlayerFillBottleEvent and BlockFillBottleEvent. These
+events do not have a common superclass due to needing to extend
+PlayerEvent and BlockEvent respectively.
+
+TODO: Add a common interface between Block and AreaEffectCloud,
+something like BottleSource to add a "source" field to each event to get
+the source of the bottle fill, water block, cauldron, beehive, or
+AreaEffectCloud.
+
+Co-authored-by: Dmitry Sidorov <jonmagon@gmail.com>
+
+diff --git a/src/main/java/io/papermc/paper/event/block/BlockFillBottleEvent.java b/src/main/java/io/papermc/paper/event/block/BlockFillBottleEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..e6c7592affa3424bd4ce3b45f6c97398df61a5ed
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/block/BlockFillBottleEvent.java
+@@ -0,0 +1,62 @@
++package io.papermc.paper.event.block;
++
++import io.papermc.paper.event.common.FillBottleEvent;
++import org.bukkit.block.Block;
++import org.bukkit.event.HandlerList;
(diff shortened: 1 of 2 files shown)


patches/api/0372-Add-FillBottleEvents-for-player-and-dispenser.patch
```diff
@@ -0,0 +1,202 @@
...
++    public @NotNull ItemStack getResultItem() {
++        return this.resultItem;
++    }
++
++    /**
++     * Sets the result of the glass bottle being filled.
++     *
++     * @param resultItem the result of the filling
++     */
++    public void setResultItem(@Nullable ItemStack resultItem) {
```
electronicboy: prefer to just not be nullable

patches/api/0372-Add-FillBottleEvents-for-player-and-dispenser.patch
```diff
@@ -0,0 +1,202 @@
...
++    public @NotNull ItemStack getResultItem() {
++        return this.resultItem;
++    }
++
++    /**
++     * Sets the result of the bottle being filled.
++     *
++     * @param resultItem the result of the filling
++     */
++    public void setResultItem(@Nullable ItemStack resultItem) {
```
electronicboy: ditto

electronicboy (approved):

patches/server/0903-Add-FillBottleEvents-for-player-and-dispenser.patch
```diff
@@ -0,0 +1,175 @@
...
+                 itemstack.shrink(1);
+                 world.playSound(player, player.getX(), player.getY(), player.getZ(), SoundEvents.BOTTLE_FILL, SoundSource.NEUTRAL, 1.0F, 1.0F);
+                 if (itemstack.isEmpty()) {
+-                    player.setItemInHand(hand, new ItemStack(Items.HONEY_BOTTLE));
+-                } else if (!player.getInventory().add(new ItemStack(Items.HONEY_BOTTLE))) {
++                    // Paper start - glass bottle events
++                    player.setItemInHand(hand, resultItem);
++                } else if (!player.getInventory().add(resultItem)) {
++                    // Paper end
+                     player.drop(new ItemStack(Items.HONEY_BOTTLE), false);
```
SoSeDiK: Shouldn't this drop `resultItem` instead?

Edit: resolved, no idea why it didn't pick up the files the first time.

SoSeDiK:
Does this correctly handle setting result to air? (and could `setResultItem` also allow passing null too?)
Also, the prior PR had `updateInventory` in places. Is this not needed anymore?

Machine-Maker:
@SoSeDiK I went back and checked, and there are 2 places where an empty itemstack would cause a problem, I will fix that. 

As for setResultItem with null, no. We want to move away from the concept of null ItemStacks. If you want to set it to nothing, use an Air itemstack.

I replaced the updateInventory calls with a more direct sendAllDataToRemote calls. That's what updateInventory did, but I think its better to use the native method instead of the api one.

SoSeDiK:
Nice, hope to see this merged!

I agree with moving away from null, just too used to `this.resultItem = resultItem == null ? new ItemStack(Material.AIR) : resultItem` logic in other highly used places like inventory/equipment management, etc. I thought it's for the sake of simplicity (easier to type and no need for manual checks if ItemStack comes from other place in the code where null might be passed).

SoSeDiK (approved):

SoSeDiK (approved):

Machine-Maker:
Am exploring a better alternative to the BlockFill event which will just use the existing dispense event. So blocked for that.

PR #7522, opened 2022-03-01
Readd StructureLocateEvent

Machine-Maker:
So there are some things to look at here.

StructureLocateEvent is now fired multiple times since its searching for a collection of `ConfiguredStructureFeature`s instead of just a `StructureFeature`. This means that only the final event fire's radius, search-only-explored values are taken into account, and I don't see a clean way of having those values be per-configued structure without just having it run the logic multiple times which would be a significant perf hit I assume.

Also, the `setType(StructureType)` method now throws a UOE because it doesn't really make sense in this context anymore.


electronicboy (approved):

kennytv:
As per internal discussion, https://github.com/PaperMC/Paper/pull/7524 will replace the old event

PR #7524, opened 2022-03-01
Add StructuresLocateEvent as replacement for StructureLocateEvent

Machine-Maker:
Probably going to be https://github.com/PaperMC/Paper/pull/7522 or this, I don't think we should try to keep both as that would just make things even more complex.


patches/server/0579-Add-StructureLocateEvent.patch
```diff
@@ -5,27 +5,138 @@ Subject: [PATCH] Add StructureLocateEvent
...
-+        // Get origin location (re)defined by event call.
-+        center = new BlockPos(event.getOrigin().getBlockX(), event.getOrigin().getBlockY(), event.getOrigin().getBlockZ());
-+        // Get radius and whether to find unexplored structures (re)defined by event call.
++        // Paper start - StructureLocateEvent
++        final org.bukkit.World world = worldserver.getWorld();
++        final org.bukkit.Location origin = net.minecraft.server.MCUtil.toLocation(worldserver, center);
++        final Registry<ConfiguredStructureFeature<?, ?>> configuredStructureFeatureRegistry = worldserver.registryAccess().registryOrThrow(Registry.CONFIGURED_STRUCTURE_FEATURE_REGISTRY);
++        final List<io.papermc.paper.world.structure.ConfiguredStructure> configuredStructures = new ArrayList<>();
++        for (Holder<net.minecraft.world.level.levelgen.feature.ConfiguredStructureFeature<?, ?>> holder : holderset) {
++            final net.minecraft.resources.ResourceLocation configuredStructureResourceLoc = configuredStructureFeatureRegistry.getKey(holder.value());
```
kennytv: Could use holder.unwrapKey().location() I believe

kennytv (commented):
Looks fine other than missing docs

patches/api/0256-Add-StructureLocateEvent.patch
```diff
@@ -30,7 +30,9 @@ index 0000000000000000000000000000000000000000..45b6694fc5741831e2df638b1f760a3c
 + *     <li>An Explorer/Treasure Map is activated.</li>
 + *     <li>{@link World#locateNearestStructure(Location, StructureType, int, boolean)} is invoked.</li>
 + * </ul>
++ * @deprecated no longer used, see {@link StructuresLocateEvent}
 + */
++@Deprecated
```
kennytv: for removal

patches/server/0579-Add-StructureLocateEvent.patch
Machine-Maker: But that fails if it’s a Direct Holder. And I don’t see the unwrapKey method being called anywhere on holders from that HolderSet. That’s why I was hesitant to use unwrapKey until I really knew exactly when you can call it.

patches/server/0579-Add-StructureLocateEvent.patch
```diff
@@ -5,27 +5,138 @@ Subject: [PATCH] Add StructureLocateEvent
...
++            if (configuredStructureResourceLoc == null) {
++                continue;
++            }
++            final io.papermc.paper.world.structure.ConfiguredStructure configuredStructure = org.bukkit.Registry.CONFIGURED_STRUCTURE.get(org.bukkit.craftbukkit.util.CraftNamespacedKey.fromMinecraft(configuredStructureResourceLoc));
++            if (configuredStructure != null) {
```
electronicboy: should this maybe warn if it's not found? seems bad if that info/context was lost

electronicboy (approved):

patches/server/0579-Add-StructureLocateEvent.patch
Machine-Maker: No, because datapacks can add their own configured structures and those wouldn't have API representations. So I think this is a perfectly plausible scenario.

Machine-Maker: This can be mitigated if ConfiguredStructure moved to a PaperConfiguredStructure -> ConfiguredStructure w/ConfiguredStructureWrapper system like enchantments.

Then a new instance of PaperConfiguredStructure could be ctor'd when a datapack creates one.

patches/api/0256-Add-PaperRegistry.patch
```diff
@@ -0,0 +1,104 @@
...
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++record ReferenceImpl<T extends Keyed>(Registry<T> registry, NamespacedKey key) implements Reference<T> {
++
++    @Override
++    public @NotNull T value() {
++        final T value = this.registry.get(this.key);
++        if (value == null) {
++            throw new java.util.NoSuchElementException("No such value with key " + this.key);
```
kashike: Why fully qualified?

Machine-Maker: Oh, that's just cause I copied it over from where it was before in the Registry interface. ill fix that.

patches/api/0256-Add-PaperRegistry.patch
```diff
@@ -0,0 +1,106 @@
...
+@@ -0,0 +1,37 @@
++package io.papermc.paper.registry;
++
++import org.bukkit.Keyed;
++import org.bukkit.NamespacedKey;
++import org.bukkit.Registry;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public interface Reference<T extends Keyed> extends Keyed {
```
jpenilla, line 27: needs doc

patches/server/0580-Add-StructuresLocateEvent.patch
```diff
@@ -0,0 +1,221 @@
...
++        out = System.out;
++        System.setOut(Bootstrap.STDOUT);
++        for (Field field : BuiltinStructures.class.getDeclaredFields()) {
++            if (field.getType().equals(ResourceKey.class)) {
++                BUILT_IN_STRUCTURES.put(((ResourceKey<ConfiguredStructureFeature<?, ?>>) field.get(null)).location(), field.getName());
++            }
++        }
++
++        for (Field field : StructureType.class.getDeclaredFields()) {
++            if (field.getType().equals(StructureType.class)) {
```
jpenilla: does this need to be changed to account for References? also is the DEFAULTS map really necessary given we scan the fields and build the API_STRUCTURE_TYPES map for testing? another thought is would it be possible/reasonable to make the tests more generic for when we add more similar registries in the future?

patches/server/0580-Add-StructuresLocateEvent.patch
```diff
@@ -0,0 +1,221 @@
...
++        missing.forEach((key, configuredFeature) -> {
++            joiner.add("public static final Reference<ConfiguredStructure> " + BUILT_IN_STRUCTURES.get(key) + " = create(\"" + key.getPath() + "\");");
++        });
++
++        return joiner.toString();
++    }
++
++    @AfterClass
++    public static void after() {
++        System.setOut(out);
```
jpenilla, line 222: kind of out of scope for this PR but would be nice to fix the testing setup so stuff like this isn't necessary

patches/server/0580-Add-StructuresLocateEvent.patch
```diff
@@ -0,0 +1,221 @@
...
++}
+diff --git a/src/test/java/org/bukkit/support/AbstractTestingBase.java b/src/test/java/org/bukkit/support/AbstractTestingBase.java
+index acde0f0170a34a8497e81fedb7944e2ef73213ae..37934f0da922d696373e7a3a8cf976fcb9015271 100644
+--- a/src/test/java/org/bukkit/support/AbstractTestingBase.java
++++ b/src/test/java/org/bukkit/support/AbstractTestingBase.java
+@@ -40,7 +40,7 @@ public abstract class AbstractTestingBase {
+         RegistryAccess.Frozen registry = RegistryAccess.builtinCopy().freeze();
+         // Paper start
+         try {
+-            java.lang.reflect.Field field = io.papermc.paper.PaperRegistry.class.getDeclaredField("REGISTRY_ACCESS");
```
jpenilla: does this diff belong in the paperregistry patch?

jpenilla (changes requested):

patches/server/0580-Add-StructuresLocateEvent.patch
Machine-Maker: Oh actually, that's not needed anymore. And that's `StructureType` not `ConfiguredStructure`. I needed that when I was generating the code to create the instances of `COnfiguredStructure` but since that's not happening now, it can just be removed.

patches/server/0580-Add-StructuresLocateEvent.patch
Machine-Maker: I *think* it should be here. its not specific to the `ConfiguredStructure` thing, but its to make sure PaperRegistry works in tests (even tho no tests use it at this point in the patch history)

patches/server/0580-Add-StructuresLocateEvent.patch
Machine-Maker: Will do this in a separate PR that can be discussed later. cause there are places were we do this, and they'd need to be reverted, plus, its unknown if there'd be some side-effect in tests.

jpenilla (approved):

kennytv (approved):

PR #7527, opened 2022-03-02
Provide ComponentSerializer services

Machine-Maker:


kashike (approved):

kezz (changes requested):
This should probably also include a provider for MiniMessage as well, will just be a no-op for now but will be good to get in for future work with custom tags/etc.

Also, is it worth deprecating the entire `PaperComponents` class for removal at some point?

Machine-Maker:
> Also, is it worth deprecating the entire PaperComponents class for removal at some point?

It's still valid to get the ComponentFlattener right? Unless there's no need anymore but I think there prob is.

kezz:
Ah yeah, forgot about the flattener. I can't reaaallly see a use case for it, but it can't hurt to expose it I suppose?

zml2008:
The flattener's useful for anyone trying to do custom serialization formats -- discord markdown, html, etc, where they need to lose the structure of different component types.

kezz:
bad bot been approved

PR #7530, opened 2022-03-03
Add DispenserReplaceItemEvent

xDec0de:
Currently handled for any bucket and bottle type, which should be all items that can be replaced on a dispenser as far as I know. This event could be used for servers with a rarity system on items (As mine), as currently I had to use the normal dispense event and delay 1 tick a loop through the dispenser inventory to change the item lore.


patches/server/0879-Add-DispenserReplaceItemEvent.patch
```diff
@@ -0,0 +1,75 @@
...
+diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
+index 59db245fe11384282af84ec1d5be08ab0e9484ca..40beb96b083e716d5c349c96ee316f0d462122d5 100644
+--- a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
++++ b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
+@@ -4,6 +4,8 @@ import java.util.Iterator;
+ import java.util.List;
+ import java.util.Optional;
+ import java.util.Random;
++
++import io.papermc.paper.event.block.DispenserReplaceItemEvent;
```
tahmid-23: use fully qualified imports instead of import statements in vanilla classes to reduce diff

patches/server/0879-Add-DispenserReplaceItemEvent.patch
```diff
@@ -0,0 +1,75 @@
...
+@@ -632,6 +634,18 @@ public interface DispenseItemBehavior {
+                     dispensiblecontaineritem.checkExtraContent((Player) null, worldserver, stack, blockposition);
+                     // CraftBukkit start - Handle stacked buckets
+                     Item item = Items.BUCKET;
++
++                    // Paper start - DispenserReplaceItemEvent
++                    org.bukkit.block.Block block = worldserver.getWorld().getBlockAt(pointer.getPos().getX(), pointer.getPos().getY(), pointer.getPos().getZ());
++                    DispenserReplaceItemEvent replaceItemEvent = new DispenserReplaceItemEvent(block, stack.asBukkitCopy(), new ItemStack(item).asBukkitCopy());
++                    replaceItemEvent.callEvent();
++                    if (!replaceItemEvent.isCancelled()) {
```
tahmid-23: replaceItemIvent.callEvent() returns a boolean for cancellation - same thing for later usages

Machine-Maker:
I would prefer adding a "replacement stack" field to the existing BlockDispenseEvent over adding a whole event for this. 

This also slightly (or significantly) conflicts with https://github.com/PaperMC/Paper/pull/7513 I think.

xDec0de:
> I would prefer adding a "replacement stack" field to the existing BlockDispenseEvent over adding a whole event for this.
> 
> This also slightly (or significantly) conflicts with #7513 I think.

That makes sense, now that you mention it I'm not really sure as why I decided to go with a whole new event as a solution, also, yeah, I saw that PR but wasn't really sure on how to work on it because it isn't merged. I could add a field but doing so will result on another conflict I guess.

electronicboy:
https://github.com/PaperMC/Paper/pull/7513 - not sure where this stands in regards to this PR and what needs doing here (too brain gone right now)

Owen1212055:
Are you planning on perhaps adding this to the already pre-existing event instead of making a new one? The PR mentioned only covers a specific use case (bottles), and I feel like a generic solution like this one is better.

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/8660, thank you regardless for your contribution. 😄

PR #7533, opened 2022-03-03
Add option for disabling nerfing portal-based mob farms

gillesbraun:
Adds a configuration option `game-mechanics.nerf-portal-mob-farms` defaulting to true (current behavior), which prevents mobs that travelled through a portal to be instantly despawned.

Rationale: Mob farms that [use portals](https://youtu.be/uh3X0Pmr9yc?t=904) to transport them to the overworld require 2 players, one in nether and one in the overworld. This is due to the hard despawn range limit of 128 for monsters.

### DIFF
--- a/patches/server/0879-Add-setting-to-disable-nerfing-of-portal-based-mob-f.patch
+++ b/patches/server/0879-Add-setting-to-disable-nerfing-of-portal-based-mob-f.patch
@@ -2,6 +2,9 @@ From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Gilles Braun <450164+gillesbraun@users.noreply.github.com>
 Date: Thu, 3 Mar 2022 16:44:14 +0100
 Subject: [PATCH] Add setting to disable nerfing of portal based mob farms
+Marks entities that teleport through a nether portal with `fromNetherPortal`. Mobs with that flag that are prevented
+from despawning if the config value is set, effectively working around the hard entity despawn range limit. Defaults
+to the old behaviour of just deleting mobs that are beyond the hard range limit.
 
 
 diff --git a/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java b/src/main/java/com/destroystokyo/paper/PaperWorldConfig.java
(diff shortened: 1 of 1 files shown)


patches/server/0879-Add-setting-to-disable-nerfing-of-portal-based-mob-f.patch
```diff
@@ -0,0 +1,56 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Mob.java b/src/main/java/net/minecraft/world/entity/Mob.java
+index 850a4f8a9bd1dc30b13205e158fcbfaa872d6157..0f4d95b82f852a11621bdae560e6c444572008b6 100644
+--- a/src/main/java/net/minecraft/world/entity/Mob.java
++++ b/src/main/java/net/minecraft/world/entity/Mob.java
+@@ -804,7 +804,8 @@ public abstract class Mob extends LivingEntity {
+                 int i = this.level.paperConfig.hardDespawnDistances.getInt(this.getType().getCategory()); // Paper - custom despawn distances
+                 int j = i * i;
+ 
+-                if (d0 > (double) j) { // CraftBukkit - remove isTypeNotPersistent() check
++                // Paper - nerf portal mob farms
```
NoahvdAa: This comment should be on the line it changes (after CB)

NoahvdAa (commented):
Welcome to Paper! :)

Machine-Maker:
What paper change makes it so mobs despawn instantly? I know people have mentioned it before, but idk if the exact cause was determined. (Not at computer so I can’t try and find it myself rn)

electronicboy:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/server/level/WorldServer.patch#128

gillesbraun:
Just confirmed that this also happens with vanilla server (Escape -> Open to Lan)
In my testing I found that if you are the only player in the server, are in the nether and push mobs through to the overworld, they will stay there. But if there is at least one player in the overworld while you are in the nether, it despawns them if the player in the overworld is far enough away

patches/server/0879-Add-setting-to-disable-nerfing-of-portal-based-mob-f.patch
gillesbraun: Done

ghost:
> What paper change makes it so mobs despawn instantly? I know people have mentioned it before, but idk if the exact cause was determined. (Not at computer so I can’t try and find it myself rn)

Its not a Paper related issue, its down to Pure vanilla mechanics and despawn ranges. Im pretty sure a redstone chunk loader would also work here, thus eliminating the entire need for a paper patch.

gillesbraun:
> Its not a Paper related issue, its down to Pure vanilla mechanics and despawn ranges.

Indeed it appears to be vanilla behavior.

> Im pretty sure a redstone chunk loader would also work here

I tried using the `/forceload` command and [KeepChunks](https://www.spigotmc.org/resources/keepchunks.23307/) plugin, but the mobs would still despawn.

PR #7538, opened 2022-03-04
Add javadoc linking for adventure-text-minimessage

Joo200:
Looks like the javadocs from text-minimessage are missing.

This will add the javadocs from here: https://jd.adventure.kyori.net/text-minimessage/4.10.0/


kezz:
This should be being done through the `apiAndDocs` configuration instead - see a few lines above your PR. Has this stopped working or something?

jpenilla:
> This should be being done through the `apiAndDocs` configuration instead - see a few lines above your PR. Has this stopped working or something?

that's not how it works. apiAndDocs is used to supply jars to the javadoc -sourcepath argument, which is used for inheriting docs (ie when we implement Translatable on some API type and don't override the javadoc), among other things.

jd linking is completely separate

jpenilla (approved):

PR #7541, opened 2022-03-04
Remove Patches

Owen1212055:
Removes 5 patches:

- Remove invalid mob spawner tile entities
- Option to remove corrupt tile entities

If the block at the location doesn't support a tile entity, MC will not put it in the world. This meant that the "Option to remove corrupt tile entities" patch has not worked in a while either, as it doesn't exist in the world any more anyways.

- Remove FishingHook reference on Craft Entity removal
- Don't let fishing hooks use portals

Fishing hooks remove their "fisher" when they are removed. 
Also, there is a getter preventing fish hooks from being able to move dimensions via portals.

- ~~Prevent Frosted Ice from loading/holding chunks~~

~~Tested this by checking if chunks were randomly ticked if ``chunk.areNeighboursLoaded(1)`` was false. This never occurred, confirming that these chunks are not ticked if there are no neighbors loaded. This patch I am a little uncertain on, so please comment if anything seems weird and I can revert this one.~~

- Workaround for setting passengers on players

Seemed old, removing this causes no issues.
```java
@EventHandler
    public void onInteract(PlayerInteractEntityEvent event) {
        event.getPlayer().addPassenger(event.getRightClicked());
    }
```

If I am missing anything with these patches, let me know, however, I'm fairly confident these should be able to be removed.


patches/removed/1.18.2/0109-Option-to-remove-corrupt-tile-entities.patch
```diff
@@ -3,6 +3,7 @@ From: Zach Brown <zach.brown@destroystokyo.com>
 Date: Wed, 5 Oct 2016 16:27:36 -0500
 Subject: [PATCH] Option to remove corrupt tile entities
 
+Corrupt tile entities are never added to the world in the first place anymore.
```
kennytv, line 6: The point is that someone *tried* to add them. the option might not be needed anymore, but the check and exception are

kennytv (changes requested):
Keep the frosted ice patch

patches/removed/1.18.2/0109-Option-to-remove-corrupt-tile-entities.patch
Owen1212055: What check here? This patch does not touch the exception, the only thing you are losing here is the "Removing corrupt tile entity" message that is logged (which again, doesn't work because the block entity isn't added anyways).

kennytv (approved):

PR #7543, opened 2022-03-04
Remove ChatColor usages

Machine-Maker:
Removes ChatColor usages from paper patches.

Had to split PaperCommand from the paper config files patch since it has to come after Adventure


patches/api/0005-Adventure.patch
```diff
@@ -1357,7 +1357,36 @@ index 4bfc214685164a38ba4261b2bae7faa8a3bd297e..50e6d45ec9349b56d20e2ffc664eca15
...
          broadcastCommandMessage(source, message, true);
+     }
+ 
+     public static void broadcastCommandMessage(@NotNull CommandSender source, @NotNull String message, boolean sendToSource) {
+-        String result = source.getName() + ": " + message;
++        // Paper start
++        broadcastCommandMessage(source, io.papermc.paper.text.PaperComponents.legacySectionSerializer().deserialize(message), sendToSource);
++    }
++    public static void broadcastCommandMessage(@NotNull CommandSender source, @NotNull net.kyori.adventure.text.Component message, boolean sendToSource) {
++        net.kyori.adventure.text.TextComponent.Builder result = net.kyori.adventure.text.Component.text().decoration(net.kyori.adventure.text.format.TextDecoration.ITALIC, false)
```
kennytv: why the explicit italics false?

kennytv (approved):

patches/api/0005-Adventure.patch
Doc94: pretty sure because by default all the ComponentText has the italic in true.. not sure why.

kezz: That is only the case for item names and lore. The explicit italic to false is not necessary unless the resulting component is being put into item name/lore and you don't want it to have italics.

Doc94: > That is only the case for item names and lore. The explicit italic to false is not necessary unless the resulting component is being put into item name/lore and you don't want it to have italics.

really i remember use this in chat message and get a italic result.. but was so many time since this...

Machine-Maker: @kennytv because the component is being used in a component with italics set to true just a few lines down.

electronicboy (approved):

PR #7548, opened 2022-03-05
Expose Furnace Minecarts' `Push` Property.

Pheotis:
## Background
PoweredMinecarts are unique insofar as they are the only entities able to produce their own thrust (albeit at the cost of fuel).
To accomplish this, mojang used two sets of properties:

#### `Fuel` represents the amount of charcoal/coal the Minecart has been fed
* This amount determines the number of ticks before the Minecart stops producing thrust.
* The `setFuel` and `getFuel` methods are already implemented upstream to handle this.

#### `PushX` and `PushY` are the other two major properties (internally `xPush` and `zPush`)
* These represent the magnitude of the Minecart's thrust, and the direction of that thrust.
  * These values are functionally independent from velocity.
* These Push properties take a few seconds to wind down, even after the Minecart runs out of fuel.

### Current API Limitations
* If a furnace Minecart's momentum is artificially stopped (as is necessary for teleportation), it is very difficult to restore.
  * Even if a new velocity vector is applied, it will not restore the Minecart's thrust (Push) momentum.
  * An example use case can be found [here](https://github.com/stargate-rewritten/Stargate-Bukkit/commit/852ac1e4f52b7346222595cb048fc5a0f77625bf).
* At present, there is no API method through which to interact with a Powered Minecart's Push value.
  * To restore a minecart's momentum, one must either use NMS, vanilla command execution, or an artificial player collision.

## This Pull Request
This pull request intends to add two methods that expose the internal `Push` values; namely, `getPush` and `setPush`.
Through these, it will be possible to directly modify a Furnace Minecart's thrust.

### Future Expansions
This specific pull request was filed on behalf of a singular plugin/organisation;
It *only* adds the minimum functionality needed to facilitate the teleportation use case.

Having stated that, the API still lacks several other powered-minecart related things.
Although outside the scope of this PR, there are several other features that we believe would help others in the future.
In no particular order, they are as follows:

- An event for fuel being added to a Powered Minecart
  - As a workaround, it is possible to just use PlayerInteractEntityEvent
- A method for stopping a Powered Minecart; doing so is surprisingly difficult and confusing.
  - The only known way to reliably do this is to:
    - Set the minecart's fuel to zero
    - On a 2L scheduler delay, set the minecart's velocity to a 0 vector.
    - Wait another 1L and restore the minecart's fuel.
- Although the existing solution is fine, being able to feed a vector to setPush may also be convenient in future.


patches/api/0373-Makes-it-possible-to-change-the-xPush-and-zPush-of-p.patch
```diff
@@ -0,0 +1,38 @@
...
++
++    /**
++     * Set the push (xPush, zPush) of the minecart.
++     *
++     * <p>Push should be a 2-dimensional vector x, z</p>
++     *
++     * @param push The push (xPush, zPush) of the minecart.
++     */
++    public void setPush(double[] push);
++    // End paper
```
Owen1212055: Please read the contribution guide on how to write proper paper comments.
https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md

Owen1212055 (changes requested):
You can just make two separate methods (set/getPushZ & set/getPushX). An array isn't needed, I feel.

You could also just make the patch name something like "Expose furnace minecart push values"

NoahvdAa:
Welcome to Paper! The comments you added in [941a284](https://github.com/PaperMC/Paper/pull/7548/commits/941a2846118aed8c7bd4a64812f678c0ada87565) and [5cb922b](https://github.com/PaperMC/Paper/pull/7548/commits/5cb922bb21a30bcea7be0c06221ba4224f3a7940) (should be `Paper start` and `Paper end` btw) will cause the patch to fail when applying, since the line numbers don't match up. See the [contributing guide](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md) for information about how to edit patches.

Pheotis:
Sorry about that!

I had thought that might be the case, but I guess I was a bit hopeful.
Only one of our devs uses linux and has a system setup for patches / paper's workflow.

The applicable commits can be reverted in the meantime.

EpicKnarvik97:
Git rebase didn't want to work with me, but I've reverted and re-created the patch with the mentioned fixes and suggestions.

electronicboy:
when editing the last patch in -Server/-API, you can just `git commit --amend`

electronicboy (approved):

NoahvdAa (approved):
Nice work, LGTM

kennytv (approved):

PR #7549, opened 2022-03-05
Allow usage of MiniMessage in server configuration files

### LINKED ISSUE #7520: Allow using MiniMessage in '...-message' configuration options

Since [Paper 1.18.2 will include Adventure 4.10.0](https://github.com/PaperMC/Paper/commit/9a73213f888b0dd9e4d3777068095747c6cd563d), MiniMessage will now also be included in Paper. Allowing the use of MiniMessage in the message config options would allow people to use hex colors with ease and use other features not available in the legacy text format, such as translatable components and clickable/hoverable text.

A new config option (e.g. `messages.use-minimessage`), that when set to true serializes the messages in bukkit.yml, spigot.yml and paper.yml using MiniMessage instead of with the legacy ampersand serializer.

Creating gradients with legacy color codes (yuck)
image

### Other

_No response_

Astralchroma:
Allows the messages in `bukkit.yml`, `spigot.yml`, and `paper.yml` to be written using the MiniMessage format.

~~In order to maintain compatibility with existing configuration files the option `configuration-uses-minimessage` option has been added to `paper.yml`. This option defaults to `false` for any configuration files before config version 26.~~

New configuration files use MiniMessage by default, and the default messages have been converted to use it.

This also includes 2 minor API changes:
- `Bukkit#getPermissionMessage` which returned a `String` has been deprecated in favor of `Bukkit#permissionMessage` which returns a `Component`
- `Server#getPermissionMessage` which returned a `String` has been deprecated in favor of `Server#permissionMessage` which returns a `Component`

Closes #7520

### DIFF
--- a/patches/api/0163-Make-the-default-permission-message-configurable.patch
+++ b/patches/api/0163-Make-the-default-permission-message-configurable.patch
@@ -5,45 +5,65 @@ Subject: [PATCH] Make the default permission message configurable
 
 
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index 76d379cfc00157dd76ae981ebd839ac0954b4773..686235a2347ebeaa5654a14cdd717009f2c0105f 100644
+index 76d379cfc00157dd76ae981ebd839ac0954b4773..f504ab7d075c1bb49c3cdd50d3bcd2b7b8f7628e 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
-@@ -2098,6 +2098,15 @@ public final class Bukkit {
+@@ -2098,6 +2098,26 @@ public final class Bukkit {
          return server.suggestPlayerNamesWhenNullTabCompletions();
      }
  
 +    /**
 +     *
++     * @deprecated Use {@link #permissionMessage()}
 +     * @return the default no permission message used on the server
 +     */
 +    @NotNull
++    @Deprecated
 +    public static String getPermissionMessage() {
 +        return server.getPermissionMessage();
++    }
++
++    /**
++     *
++     * @return the default no permission message used on the server
++     */
++    @NotNull
++    public static net.kyori.adventure.text.Component permissionMessage() {
++        return server.permissionMessage();
 +    }
 +
      /**
       * Creates a PlayerProfile for the specified uuid, with name as null
       * @param uuid UUID to create profile for
 diff --git a/src/main/java/org/bukkit/Server.java b/src/main/java/org/bukkit/Server.java
-index 818c37490d98c287ade9b56f3fc01001db1df773..f43720d07e80e3d2937f5b271664b5268d7af027 100644
+index 818c37490d98c287ade9b56f3fc01001db1df773..79995da5de73e0d7441549ba83349826791d9047 100644
 --- a/src/main/java/org/bukkit/Server.java
 +++ b/src/main/java/org/bukkit/Server.java
-@@ -1841,6 +1841,13 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
+@@ -1841,6 +1841,22 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
       */
      boolean suggestPlayerNamesWhenNullTabCompletions();
  
 +    /**
 +     *
++     * @deprecated Use {@link #permissionMessage()}
 +     * @return the default no permission message used on the server
 +     */
 +    @NotNull
++    @Deprecated
 +    String getPermissionMessage();
++
++    /**
++     *
++     * @return the default no permission message used on the server
++     */
++    @NotNull
++    net.kyori.adventure.text.Component permissionMessage();
 +
      /**
       * Creates a PlayerProfile for the specified uuid, with name as null
       * @param uuid UUID to create profile for
 diff --git a/src/main/java/org/bukkit/command/Command.java b/src/main/java/org/bukkit/command/Command.java
-index b1152f4dede61383232cc6713d448840612eac13..95847fcb1ac2e430ee192f4f7ac94e981151c5db 100644
+index b1152f4dede61383232cc6713d448840612eac13..9e2e4d8082b014229976f92c213b3084e09e11eb 100644
 --- a/src/main/java/org/bukkit/command/Command.java
 +++ b/src/main/java/org/bukkit/command/Command.java
 @@ -184,10 +184,9 @@ public abstract class Command {
@@ -54,7 +74,7 @@ index b1152f4dede61383232cc6713d448840612eac13..95847fcb1ac2e430ee192f4f7ac94e98
 -            target.sendMessage(ChatColor.RED + "I'm sorry, but you do not have permission to perform this command. Please contact the server administrators if you believe that this is a mistake.");
              // Paper start - use components for permissionMessage
 -        } else if (!permissionMessage.equals(net.kyori.adventure.text.Component.empty())) {
-+        net.kyori.adventure.text.Component permissionMessage = this.permissionMessage != null ? this.permissionMessage : io.papermc.paper.text.PaperComponents.legacySectionSerializer().deserialize(Bukkit.getPermissionMessage());
++        net.kyori.adventure.text.Component permissionMessage = this.permissionMessage != null ? this.permissionMessage : Bukkit.permissionMessage();
 +        if (!permissionMessage.equals(net.kyori.adventure.text.Component.empty())) {
              target.sendMessage(permissionMessage.replaceText(net.kyori.adventure.text.TextReplacementConfig.builder().matchLiteral("<permission>").replacement(permission).build()));
              // Paper end
--- a/patches/api/0183-Expose-the-internal-current-tick.patch
+++ b/patches/api/0183-Expose-the-internal-current-tick.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] Expose the internal current tick
 
 
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index cf7f8a8f03adcbe466b59ea8b98b527fb54a0803..61877d9d64840408a7aec7bcc2a54779a9e820d8 100644
+index 9215375d0813840b06d11f6e14f8e660865f58cc..3e08b312895710584dca6d62becba2a31e297cfd 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
-@@ -2140,6 +2140,10 @@ public final class Bukkit {
+@@ -2151,6 +2151,10 @@ public final class Bukkit {
      public static com.destroystokyo.paper.profile.PlayerProfile createProfile(@Nullable UUID uuid, @Nullable String name) {
          return server.createProfile(uuid, name);
      }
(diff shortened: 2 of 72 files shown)


patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,73 @@
...
++++ b/src/main/java/org/bukkit/Bukkit.java
+@@ -2226,13 +2226,24 @@ public final class Bukkit {
+ 
+     /**
+      *
++     * @deprecated Use org.bukkit.Bukkit#permissionMessage()
+      * @return the default no permission message used on the server
+      */
+     @NotNull
++    @Deprecated
```
NoahvdAa: ~~Needs a `// Paper` comment, same for server~~

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
+             getDouble("settings.packet-limiter.limits.all.interval", 7.0),
+             getDouble("settings.packet-limiter.limits.all.max-packet-rate", 500.0),
+@@ -684,4 +735,9 @@ public class PaperConfig {
+     private static void useDimensionTypeForCustomSpawners() {
+         useDimensionTypeForCustomSpawners = getBoolean("settings.use-dimension-type-for-custom-spawners", false);
+     }
++
++    public static boolean configurationUsesMiniMessage;
++    private static void configurationUsesMiniMessage() {
++        configurationUsesMiniMessage = getBoolean("settings.configuration-uses-minimessage", version > 25);
```
MiniDigger: Not sure I like this. How about instead, we check if the message contains ampersand, then revert to legacy, else default to MiniMessage

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,73 @@
...
+     }
+ 
++    /**
++     *
++     * @return the default no permission message used on the server
++     */
++    @NotNull
++    public static net.kyori.adventure.text.Component permissionMessage() {
++        return server.permissionMessage();
++    }
```
NoahvdAa: ~~Needs `// Paper start` and `// Paper end`, same for server~~

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
+-        return msg != null ? io.papermc.paper.adventure.PaperAdventure.LEGACY_SECTION_UXRC.deserialize(msg) : null;
++        String msg = this.configuration.getString("settings.shutdown-message");
++
++        if (com.destroystokyo.paper.PaperConfig.configurationUsesMiniMessage)
++            // MiniMessage >> Component
++            return msg != null ? net.kyori.adventure.text.minimessage.MiniMessage.miniMessage().deserialize(msg) : null;
++
++        else
++            // Legacy >> Component
++            return msg != null ? io.papermc.paper.adventure.PaperAdventure.LEGACY_SECTION_UXRC.deserialize(msg) : null;
```
NoahvdAa: Curly braces

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: Just checking for usage of & could cause issues incase someone is using it other then formatting, having a config option is much more clear and leaves no room for it to get confused.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
++
++        if (string == null) {
++            authenticationServersDownKickMessage = null;
++
++        } else {
++            if (configurationUsesMiniMessage)
++                authenticationServersDownKickMessage = miniMessage().deserialize(string);
++
++            else
++                authenticationServersDownKickMessage = legacyAmpersand().deserialize(string);
```
NoahvdAa: Same here and below, curly braces

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
Machine-Maker: I don't think there should be a config option at all. And if the config version is < 26, all the messages should be force updated to minimessage by using the legacy and then minimessage serializers

NoahvdAa (commented):
Congrats on making your first paper PR, welcome!

electronicboy:
as I said on discord, really we should consilidate all of the messages into a singular patch file, rather than this patch which tweaks all of the old options, we'd also have to look into migrating bukkits messages over, but, am not sure 100% on that's based done without the config migration stuff

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Machine-Maker: Its already inside a paper comment block

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Machine-Maker: Already inside a paper comment block

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: They are already in a `//paper start` and `//paper end` block from another patch

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,73 @@
...
++    @Deprecated
+     String getPermissionMessage();
+ 
++    /**
++     *
++     * @return the default no permission message used on the server
++     */
++    @NotNull
++    net.kyori.adventure.text.Component permissionMessage();
++
```
Machine-Maker: I think this change (and the similar ones) should be made in the patch that added them.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,73 @@
...
+ 
++    /**
++     *
++     * @return the default no permission message used on the server
++     */
++    @NotNull
++    public static net.kyori.adventure.text.Component permissionMessage() {
++        return server.permissionMessage();
++    }
++
```
Machine-Maker: Should be done in the patch that added this.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,73 @@
...
+diff --git a/src/main/java/org/bukkit/command/Command.java b/src/main/java/org/bukkit/command/Command.java
+index 95847fcb1ac2e430ee192f4f7ac94e981151c5db..9e2e4d8082b014229976f92c213b3084e09e11eb 100644
+--- a/src/main/java/org/bukkit/command/Command.java
++++ b/src/main/java/org/bukkit/command/Command.java
+@@ -185,7 +185,7 @@ public abstract class Command {
+         }
+ 
+             // Paper start - use components for permissionMessage
+-        net.kyori.adventure.text.Component permissionMessage = this.permissionMessage != null ? this.permissionMessage : io.papermc.paper.text.PaperComponents.legacySectionSerializer().deserialize(Bukkit.getPermissionMessage());
++        net.kyori.adventure.text.Component permissionMessage = this.permissionMessage != null ? this.permissionMessage : Bukkit.permissionMessage();
```
Machine-Maker: Should be inside that patch that added it.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
+ 
+     private static File CONFIG_FILE;
+@@ -71,8 +75,8 @@ public class PaperConfig {
+         commands.put("paper", new PaperCommand("paper"));
+         commands.put("mspt", new MSPTCommand("mspt"));
+ 
+-        version = getInt("config-version", 25);
+-        set("config-version", 25);
++        version = getInt("config-version", 26);
++        set("config-version", 26);
```
Machine-Maker: This change needs to be done in the very first config patch

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
++        String kickVehicle = getString("messages.kick.flying-vehicle", "Flying is not enabled on this server");
++
++        if (configurationUsesMiniMessage) {
++            flyingKickPlayerMessage = miniMessage().deserialize(kickPlayer);
++            flyingKickVehicleMessage = miniMessage().deserialize(kickVehicle);
++
++        } else {
++            flyingKickPlayerMessage = legacyAmpersand().deserialize(kickPlayer);
++            flyingKickVehicleMessage = legacyAmpersand().deserialize(kickVehicle);
++        }
```
Machine-Maker: Should be done in the patch that adds this.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
```diff
@@ -0,0 +1,410 @@
...
++            authenticationServersDownKickMessage = null;
++
++        } else {
++            if (configurationUsesMiniMessage)
++                authenticationServersDownKickMessage = miniMessage().deserialize(string);
++
++            else
++                authenticationServersDownKickMessage = legacyAmpersand().deserialize(string);
++        }
+     }
```
Machine-Maker: Should be done in the patch that adds thsi.

Machine-Maker (changes requested):
Ok, basically all of this should be done in the patch that makes these changes originally.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: How do I find out what patch added something to begin with?

Astralchroma: Mvm I think I found it.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: Moved.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: Moved.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
Peter-Crawley: Moved.

patches/api/0373-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: Moved.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
Peter-Crawley: Moved.

patches/server/0880-Allow-usage-of-MiniMessage-in-server-configuration.patch
Astralchroma: Okay, I guess I will just force all of the messages up to MiniMessage then.

Astralchroma:
> Ok, basically all of this should be done in the patch that makes these changes originally.

So I have been working on doing this, but when it comes to messages added by Spigot, that has to be done in a new patch, which is fine, but... patch 147 depends on one of these messages, so the patch has to go before 147.

How would I go about adding a patch before an existing patch?

Machine-Maker:
> How would I go about adding a patch before an existing patch?

I wouldn't put it right before, I'd maybe put it in the adventure patch. @kashike might want to weigh in on that. but to add a patch, just interactive rebase to the commit you want to add a patch after, and then just make a new commit, and then continue rebasing.

Astralchroma:
So after basically rewriting this PR, usage of MiniMessage is no longer optional, I will add code to automatically convert Legacy messages to MiniMessage in a bit.

I have also changed everything to be done as part of previous patches. *Which was very annoying and took forever.*

Astralchroma:
*Not dead, just have alot of things to be doing, I will get back to this later.*

Astralchroma:
Closed because I am not sure how to go about handling configuration migration, so I will leave someone else to deal with it.

PR #7553, opened 2022-03-06
Prevent entity loading causing async lookups

### LINKED ISSUE #7137: failed main thread check
### Expected behavior

Normal server shutdown

### Observed/Actual behavior

See https://paste.gg/p/anonymous/cea81e06151641859f447ee1c8149ca9

### Steps/models to reproduce

Start the server and after it begins plugins loading, but before it is finished, press CTRL+C, so the server will start shutdown immediately after it finishes loading and enabling

### Plugin and Datapack List

```
[17:55:05] Plugins (18): AutoMessage, Chunky, ChunkyBorder, CoreProtect, dynmap, Essentials, EssentialsGeoIP, EssentialsSpawn, LuckPerms, MobArena, NoCheatPlus, OpenInv, Orebfuscator, ProtocolLib, TitleManager, Vault, WorldEdit, WorldGuard
```
```
[17:54:21] There are 2 data packs enabled: [vanilla (built-in)], [file/Bukkit (world)]
[17:54:21] There are no more data packs available
```

### Paper version

```
[17:54:39] This server is running Paper version git-Paper-81 (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 8c189d0)
```

### Other

_No response_

Owen1212055:
Fixes https://github.com/PaperMC/Paper/issues/7137

This just moves the code to instead run when the entity is ticked for the first time instead.

Tested with wolves & enderman and it appears there is no behavior change.

### DIFF
--- a/patches/server/0881-Prevent-entity-loading-causing-async-lookups.patch
+++ b/patches/server/0881-Prevent-entity-loading-causing-async-lookups.patch
@@ -17,10 +17,10 @@ index 84fce7dccf9232209f939a32acfc3131e62eb27c..34847add2a35e535b001f581677c4953
          if (this.isPassenger() && this.getVehicle().isRemoved()) {
              this.stopRiding();
 diff --git a/src/main/java/net/minecraft/world/entity/NeutralMob.java b/src/main/java/net/minecraft/world/entity/NeutralMob.java
-index dedf76de5d6f46b9626ca4a98cfffe125b90dd0c..f02224c1993b7953f6d7c4bad1edc6c41cc8bbf2 100644
+index dedf76de5d6f46b9626ca4a98cfffe125b90dd0c..78632fd681049fbd49d0030c23ed204dbc515a44 100644
 --- a/src/main/java/net/minecraft/world/entity/NeutralMob.java
 +++ b/src/main/java/net/minecraft/world/entity/NeutralMob.java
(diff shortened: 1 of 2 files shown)


patches/server/0881-Prevent-entity-loading-causing-async-lookups.patch
```diff
@@ -0,0 +1,68 @@
...
+-                if (entity != null) {
+-                    if (entity instanceof Mob) {
+-                        this.setLastHurtByMob((Mob) entity);
+-                    }
+-
+-                    if (entity.getType() == EntityType.PLAYER) {
+-                        this.setLastHurtByPlayer((Player) entity);
+-                    }
+-
+-                }
```
Machine-Maker: Add a replacement paper comment just saying that this was moved to its own method to do earlier.

Machine-Maker (changes requested):

patches/server/0881-Prevent-entity-loading-causing-async-lookups.patch
Owen1212055: Lmk if I did this right.

Machine-Maker (approved):

electronicboy (approved):

Owen1212055:
Rebased

PR #7555, opened 2022-03-06
Add basic Minecraft version update steps to CONTRIBUTING

kennytv:


Machine-Maker (changes requested):
I agree with jmp that we shouldn't list these in the repo like this.

PR #7556, opened 2022-03-06
Added smelt checking in "Material" class

TrevorMickelson:
I did this to add a faster time complexity for checking the smelt result. And it just makes life better. I needed this for a CustomEnchants plugin a while back, and ended up having to create a custom solution because it wasn't built in. Figured I'd try to build it in, and see if you guys like it! Thank you! <3

### DIFF
--- a/patches/api/0374-Implemeted-ability-to-obtain-smelt-result-of-a-mater.patch
+++ b/patches/api/0374-Implemeted-ability-to-obtain-smelt-result-of-a-mater.patch
@@ -1,0 +1,85 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Trevor <tmick28@protonmail.com>
+Date: Sun, 6 Mar 2022 17:46:29 -0500
+Subject: [PATCH] Implemeted ability to obtain smelt result of a material, in
+ the Material class
+
+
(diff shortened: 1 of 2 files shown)


electronicboy:
given the plans to basically rewrite this entire class, definitely not fond of adding a static block to the class

TrevorMickelson:
> 

Oh, I wasn't aware you were going to rewrite the class. That's interesting.

qixils:
the main issue i have with this is that recipes are not static, as they can be added by both datapacks and plugins, and the Sponge+Paper APIs are generally moving away from hardcoding dynamic data like this. it also doesn't quite make sense to just ignore the NBT data of the input/output items as is being done here. i think this is better left to plugins to handle

TrevorMickelson:
> the main issue i have with this is that recipes are not static, as they can be added by both datapacks and plugins, and the Sponge+Paper APIs are generally moving away from hardcoding dynamic data like this. it also doesn't quite make sense to just ignore the NBT data of the input/output items as is being done here. i think this is better left to plugins to handle

Interesting, I hadn't considered plugins implementing custom recipes. I really like the idea of getting the smelt result easily though. I may try to think of a more dynamic way to do it. It's definitely possible.

PR #7559, opened 2022-03-08
Strip enderpearl owners when they are no longer ticked

### LINKED ISSUE #6920: Ender pearls not despawning upon owner death if lazy loaded

Owen1212055:
Fixes https://github.com/PaperMC/Paper/issues/6920
Supercedes https://github.com/PaperMC/Paper/pull/6924

Currently, the fix is to check if the entity is no longer ticked... in the tick method?
This causes it to never actually be run because the entity tick method is no longer called when the chunk is in this state.

This change makes it so that this logic is run directly when they are no longer ticked (the callback method).

### DIFF
--- a/patches/server/0191-Block-Enderpearl-Travel-Exploit.patch
+++ b/patches/server/0191-Block-Enderpearl-Travel-Exploit.patch
@@ -26,6 +26,23 @@ index ef9f6712046a22a8d0dfcf95d8e2d12403345532..ea426792b751fd7591bf09f2256dd6bd
 +        log("Disable Unloaded Chunk Enderpearl Exploit: " + (disableEnderpearlExploit ? "enabled" : "disabled"));
 +    }
  }
+diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
+index c0ec21a8110ed54489361cfae1e1112fe448150a..164e3261fb6b7a30c86baffde62045b1c6081232 100644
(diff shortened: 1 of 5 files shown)


electronicboy (approved):

Machine-Maker (approved):

Owen1212055:
Rebased and ready

Owen1212055:
Can the rebase label be removed here? This can be pushed whenever

lynxplay:
Got ya 👍

sudofox:
how does this affect the stasis chamber feature that mojang introduced a while back?

e-im:
As long as you have `disable-unloaded-chunk-enderpearl-exploit` set to `false` in `paper.yml` (these were already blocked by default on paper) there is no change.

PR #7560, opened 2022-03-09
Various improvements to console color formatting

jpenilla:
- Use Adventure's [char]#rrggbb RGB format internally instead of the unusual x repeated characters format
- Option (default true) to use actual color for NamedTextColors instead of closest ANSI code
- Use a rarely used character (0x7f) instead of section symbol for internal format

### DIFF
--- a/patches/server/0596-Add-support-for-hex-color-codes-in-console.patch
+++ b/patches/server/0596-Add-support-for-hex-color-codes-in-console.patch
@@ -5,20 +5,82 @@ Subject: [PATCH] Add support for hex color codes in console
 
 Converts upstream's hex color code legacy format into actual hex color codes in the console.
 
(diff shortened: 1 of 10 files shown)


kennytv (approved):

JRoy:
you're incredibly stupid

kangarko:
This shit just broke tons of plugins using § color formatting, and should be reversed as soon as possible

electronicboy:
`§` has been deprecated for years and there's little interest trying to shim back in support for this, long term goals is to support logging components which do allow us to deal with this stuff without funky magic hiding in the logger chain, you can also do the generally supported thing for loggers and just log ansi in supported environments

Astralchroma:
> This shit just broke tons of plugins using § color formatting, and should be reversed as soon as possible

Good, maybe then they will stop using deprecated APIs.

iGabyTM:
> This shit just broke tons of plugins using § color formatting, and should be reversed as soon as possible

I think someone has to update their calendar, is 2022, not 2014 😬

PR #5725, opened 2021-05-25
Itemstack Glint API

Machine-Maker:
Closes #5274

First time touching item meta stuff

Unsure of a few things, mainly the behavior of this elsewhere in the codebase, like anvils/grindstones. Currently you can remove the glint with the grindstone (and get no xp), and you can put enchanted books on the item, *BUT* that removes the empty NBTTagCompound, so removing that enchant will make the item lose it's glint.

Issue relating to this: #5723

PR #7564, opened 2022-03-09
Add more entity api

kennytv:
Also put the Wither API into the larger patch, will probably squash more of those simple patches at a later date


patches/server/0682-Missing-Entity-Behavior-API.patch
```diff
@@ -35,6 +37,52 @@ index cd278a859c87fc89c421378ffab1bd36a45bd65d..a726006888bbbdb290bcda3ac4fd45d6
...
+--- a/src/main/java/net/minecraft/world/entity/boss/wither/WitherBoss.java
++++ b/src/main/java/net/minecraft/world/entity/boss/wither/WitherBoss.java
+@@ -84,6 +84,11 @@ public class WitherBoss extends Monster implements PowerableMob, RangedAttackMob
+         return entityliving.getMobType() != MobType.UNDEAD && entityliving.attackable();
+     };
+     private static final TargetingConditions TARGETING_CONDITIONS = TargetingConditions.forCombat().range(20.0D).selector(WitherBoss.LIVING_ENTITY_SELECTOR);
++    // Paper start
++    private boolean canPortal = false;
++
++    public void setCanTravelThroughPortals(boolean canPortal) { this.canPortal = canPortal; }
```
Owen1212055, line 51: Is there any reason why this isn't just a public field here? It prolly doesn't rly matter but I'm just curious why.

Owen1212055 (approved):
Looks good regardless!

patches/server/0682-Missing-Entity-Behavior-API.patch
electronicboy: was just a consilidated patch, no changes here; could be AT'd now given we have that stuff but 🤷

kennytv:
also added methods to set trident glow and loyalty without having to change or look into the itemstack's enchantments

patches/server/0682-Missing-Entity-Behavior-API.patch
```diff
@@ -287,3 +393,113 @@ index 2d2620dbb16aec850e8afda02174508a4be5a313..ba4e6deaaa725296be830324d2c64868
...
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftTrident.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftTrident.java
+@@ -37,4 +37,27 @@ public class CraftTrident extends CraftArrow implements Trident {
+     public EntityType getType() {
+         return EntityType.TRIDENT;
+     }
++
++    // Paper start
++    @Override
++    public boolean hasGlint() {
```
MiniDigger, line 449: both glint and foil are kinda strange names, why not glowing?

MiniDigger (approved):
just one api am not sure about, but if others don't care this is fine

patches/server/0682-Missing-Entity-Behavior-API.patch
kennytv: Fun things, https://papermc.io/javadocs/paper/1.18/org/bukkit/entity/Entity.html#isGlowing()

Owen1212055: Yea glowing and if it has a glint or not isn't the same thing. I think glint is fine, as it even matches with this PR for itemstack glint (if we are still waiting on mojang for this one as well) 

https://github.com/PaperMC/Paper/pull/5725/files

PR #7569, opened 2022-03-09
Fix setting certain custom inventory names

Owen1212055:
image

For some inventories, their name isn't set when they are created. This is because most "named" inventories rely on being set by ``CraftTileInventoryConverter#createInventory``, where if they extend RandomizableContainerBlockEntity they are correctly set. 

This is not an issue for inventories like Furnaces + Brewing Stands, as they have their own custom converter which manually sets the title (See Furnace/BrewingStand inner class).
This change uses a better parent class, which should correctly set block entities that can be named instead of creating custom converters for each one.

Also, should I have the patch that fixes it for the deprecated spigot method? I'm not sure if we should just fix it for the adventure one or not.

### DIFF
--- a/patches/server/0008-Adventure.patch
+++ b/patches/server/0008-Adventure.patch
@@ -7,7 +7,7 @@ Co-authored-by: zml <zml@stellardrift.ca>
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 
 diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
-index 5c7b2850d311e4d54236ba9acce148bca05672fd..15927e8b747e536a5136d9b59edc8d6f50354c42 100644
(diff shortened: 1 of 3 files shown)


Machine-Maker:
Does this actually fix anything (that wouldn't be fixed by an already open PR)? the only 4 block entities that **DO** extend `BaseContainerBlockEntity` but **DO NOT** extend `RandomizableContainerBlockEntity` all have their own converters (or *will* have their own converters [when this is merged](https://github.com/PaperMC/Paper/pull/7249))

Owen1212055:
In that case, nope that will fix them! I was just thinking this was a blanket fix for any other possible tile entities that are added in the future as well. This can be closed if you think it’s unnecessary however.

Edit: Relevant Convo: https://canary.discord.com/channels/289587909051416579/925530366192779286/951271666628059176

Machine-Maker:
I think this can be closed and then maybe implemented when any inventories that aren't covered are added. (if upstream doesn't handle it first)

PR #7575, opened 2022-03-11
Don't tick despawned entities

Owen1212055:
Before 1.17, mc checked if the entity was not dead before ticking it, as there was despawn logic that could be run which could cause the entity to be removed.
image

However, as of 1.17 there is no longer a check, causing entities that were despawned to be ticked.
image

It's unknown if there is any behavior that relies on this, but looking myself I couldn't find anything. If anyone can find any logic here that may rely on this, it would be appreciated. 

Context can be seen at: https://canary.discord.com/channels/289587909051416579/925530366192779286/951663121146191983


Owen1212055:
Because this behavior was only added in part of 1.17 might this be worth reverting?

Owen1212055:
LivingEntity#tick now checks if the entity is removed before executing aiStep.
```java
if (!this.isRemoved()) {
            this.aiStep();
}
```
This gives me the feeling that this was intentional then or something.

PR #7587, opened 2022-03-13
Implement Translatable on CreativeCategory

Machine-Maker:
The 3 skipped `CreativeModeTabs` aren't implemented by the API and I don't really think they need to. The api is more of a categorization, and the other 3 aren't really categories in that sense.


Machine-Maker:
Rebased

electronicboy (approved):

PR #7588, opened 2022-03-13
fix datapacks not being able to modify nether/end

### LINKED ISSUE #6880: Datapacks modifying nether/end generation don't apply properly
### Expected behavior

Flat terrain to be generated.

Version: Paper-325

### Observed/Actual behavior

Normal terrain generated.

Version: Paper-377

### Steps/models to reproduce

1. Add datapack to \<world\>\datapack and enable it.
2. Restart server.
3. Generate new terrain.
4. Normal terrain generated instead of flat.

Datapack: [FlatOverworld.zip](https://github.com/PaperMC/Paper/files/7515357/FlatOverworld.zip)
This datapack makes terrain in overworld flat.

### Plugin and Datapack List

\> plugins
[23:58:58 INFO]: Plugins (0):
\> datapack list
[23:59:18 INFO]: There are 3 data packs enabled: [vanilla (built-in)], [file/bukkit (world)], [file/FlatOverworld (world)]
[23:59:18 INFO]: There are no more data packs available

### Paper version

\> version
[23:57:24 INFO]: This server is running Paper version git-Paper-377 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 3f17694)
Previous version: git-Paper-325 (MC: 1.17.1)

### Other

Server was updated from paper-325 to paper-377 and datapack stopped working.
Datapack works in vanilla server.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/6880

Ok, so root of the issue here is... the `WorldGenSettings` codec uses something called a "data pack aware codec" to create its registry of `LevelStem`. This is like the *only* place on the server where the MappedRegistry ctor is used except in the internal register methods. And then the registry is filled up with defaults, and then sent to the `RegistryLoader` to be overriden by any possible data packs. Now this `RegistryLoader` has a cache, and if it's read in something at a particular location before, it skips it, and doesn't bother adding it to the registry, assuming its done that before. The problem is, craftbukkit runs this loading process 4 times. once at the beginning (like vanilla), but then one more time for each of the default worlds. This means that the override LevelStem isn't added to the registry in the WorldGenSettings for the other two worlds. 

All this does, is ignore any cached results if its trying to fill a LevelStem registry. As far as I can tell, the only time a new instance of registry will be passed into this `RegistryLoader`, will be a LevelStem one, others may get loaded multiple times, but it will be the same *instance* of registry, so it won't matter that it's skipped.

### DIFF
--- a/patches/server/0005-MC-Dev-fixes.patch
+++ b/patches/server/0005-MC-Dev-fixes.patch
@@ -182,6 +182,30 @@ index 2a6969baca7c07c52223672de58886dd05c032eb..a620e53021e02da1663b8d8b68d429b2
      @Override
      public void write(FriendlyByteBuf buf) {
          buf.writeInt(this.playerId);
+diff --git a/src/main/java/net/minecraft/resources/RegistryLoader.java b/src/main/java/net/minecraft/resources/RegistryLoader.java
+index 858037070db0b0fed07bf542294b76028f0018f9..75f72b568ee2b27c5042da688b62d4e3c5732e09 100644
+--- a/src/main/java/net/minecraft/resources/RegistryLoader.java
++++ b/src/main/java/net/minecraft/resources/RegistryLoader.java
+@@ -73,7 +73,7 @@ public class RegistryLoader {
+     }
+ 
+     private <E> RegistryLoader.ReadCache<E> readCache(ResourceKey<? extends Registry<E>> registryRef) {
+-        return this.readCache.computeIfAbsent(registryRef, (ref) -> {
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Alternatively, some other caching solution can be implemented that takes into account the registry instance. That way there is still caching, but only if that *instance* of registry hasn't been run through the loader before. But like I said, I'm pretty sure its just the LevelStem registries that this affects.

Machine-Maker:
Also alternatively, a new instance of `RegistryLoader`, via a new instance of `RegistryOps` instead of using the global one, would make it a new cache for each world. but then you really do kinda lose out on caching all the rest of it.

jpenilla (approved):

PR #7591, opened 2022-03-13
Llama API

Owen1212055:
Pulls some of the API from https://github.com/PaperMC/Paper/pull/6773


Owen1212055:
See: https://github.com/PaperMC/Paper/pull/7632

PR #7595, opened 2022-03-13
Add allow server listing client option

NoahvdAa:
Adds the `ALLOWS_SERVER_LISTINGS` client option from #7268. Also addresses [MM's comment](https://github.com/PaperMC/Paper/pull/7268#discussion_r777680170) by adding a constructor that takes a map as an argument. I'm not entirely sure how I feel about this map approach, but it's definitely better than adding a new constructor every time.

### DIFF
--- a/patches/api/0194-Add-Player-Client-Options-API.patch
+++ b/patches/api/0194-Add-Player-Client-Options-API.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add Player Client Options API
 
 diff --git a/src/main/java/com/destroystokyo/paper/ClientOption.java b/src/main/java/com/destroystokyo/paper/ClientOption.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..cedb51f9f3a9150035c2b44970a096448c441dd9
+index 0000000000000000000000000000000000000000..f89bfeba29e6988db849957a508ca97ff5322242
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/ClientOption.java
-@@ -0,0 +1,50 @@
+@@ -0,0 +1,52 @@
 +package com.destroystokyo.paper;
 +
 +import net.kyori.adventure.translation.Translatable;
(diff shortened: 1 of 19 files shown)


electronicboy (approved):

patches/api/0194-Add-Player-Client-Options-API.patch
```diff
@@ -127,6 +130,19 @@ index 0000000000000000000000000000000000000000..f7f171c4ee0b8339b2f8fbe82442d65f
 +        this.chatColors = chatColors;
 +        this.skinparts = skinParts;
 +        this.mainHand = mainHand;
++        this.allowsServerListings = false;
++    }
++
++    public PlayerClientOptionsChangeEvent(@NotNull Player player, @NotNull Map<ClientOption<?>, ?> options) {
```
Machine-Maker, line 139: Add an ApiStatus.Internal annotation to the ctor. no plugin should really need to construct this event I think.

patches/api/0194-Add-Player-Client-Options-API.patch
```diff
@@ -127,6 +130,19 @@ index 0000000000000000000000000000000000000000..f7f171c4ee0b8339b2f8fbe82442d65f
...
++        super(player);
++
++        this.locale = (String) options.get(ClientOption.LOCALE);
++        this.viewDistance = (int) options.get(ClientOption.VIEW_DISTANCE);
++        this.chatVisibility = (ChatVisibility) options.get(ClientOption.CHAT_VISIBILITY);
++        this.chatColors = (boolean) options.get(ClientOption.CHAT_COLORS_ENABLED);
++        this.skinparts = (SkinParts) options.get(ClientOption.SKIN_PARTS);
++        this.mainHand = (MainHand) options.get(ClientOption.MAIN_HAND);
++        this.allowsServerListings = (boolean) options.get(ClientOption.ALLOW_SERVER_LISTINGS);
```
Machine-Maker, line 148: wrap these in requireNonNulls. yeah, the map isn't great. but its better than ending up with 40 constructors in 2 years I think

Machine-Maker (changes requested):

Machine-Maker (approved):

NoahvdAa:
Rebased for 1.19, label can be removed!

Machine-Maker:
is the `textFilteringEnabled` field in the ServerboundClientInformationPacket not used by the client yet? or can that be added here too?. I see it checking a flag on the userApiService on the client, so maybe that's something that can be set on the account?

NoahvdAa:
Could've sworn that was part of the original PR, it has now been added!

NoahvdAa:
Rebased.

kennytv (approved):

Owen1212055 (approved):

Owen1212055:
~~Out of the scope of this PR, but PlayerClientOptionsChangeEvent could use similar love.~~ I am blind.

Owen1212055 (approved):

PR #7603, opened 2022-03-15
Fix console completions on invalid commands

Machine-Maker:
Both vanilla and plugin commands that completions were requested for would complete if the command started with a `/` which craftbukkit doesn't allow even tho vanilla does. I think its probably for the best anyways. Because commands can have labels of `/whatever`.


Machine-Maker:
Rebased for 1.21.1

lynxplay (approved):

PR #7606, opened 2022-03-15
Properly lookup random-ticked precipitation blocks

### LINKED ISSUE #7605: Cauldrons not filling from rain
### Expected behavior

While it's raining, a cauldron is expected to slowly fill up as long as nothing's obstructing it. In a full real-life day, at _least_ a single full cauldron is expected.

### Observed/Actual behavior

I ran a skyblock world with two cauldrons in an otherwise empty chunk, with a player loading said chunks for over 48 real-life hours. Both cauldrons remained dry the entire time. Gamerule randomTickSpeed was set to 15.

### Steps/models to reproduce

Create an empty singleplayer world using a custom flat world preset with only 1 layer of air. Copy this world to a Paper server, and place a cauldron at Y=128 in an otherwise empty chunk. No matter how long you wait, the cauldron will remain dry.

### Plugin and Datapack List

Plugins:
- Dynmap (https://www.spigotmc.org/resources/dynmap%C2%AE.274/)
- ProtectTheDirt (https://github.com/cosrnic/ProtectTheDirt)

Datapacks:
- `vanilla` (built-in)
- `file/bukkit` (world)

### Paper version

This server is running Paper version git-Paper-235 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: e0b3f3f)
You are 15 version(s) behind

### Other

_No response_

lynxplay:
As part of the random block tick optimisation, a mutable block position
was introduced in favour of two block positions, one previously pointing
towards the highest motion blocking block at the x/z column while the
other points towards the empty block above.

To convert the mutable block position between the two previous positions
when executing logic, the optimisation patch switches their y value
between the lower and higher y value (which are one block apart).

However, the patch failed to properly mutate the shared block position
to the lower block after checking for needed snow layers, causing the
logic that is usually responsible for handling precipitation on blocks
to fetch the block data of the higher block (usually air) for executing
its logic.

Hence, previously to this commit, the game fails to handle any
precipitation on blocks (such as cauldrons) as their logic is never
executed (rather the non-motion blocking block above is used instead).

Resolves: #7605

### DIFF
--- a/patches/server/0764-Optimise-random-block-ticking.patch
+++ b/patches/server/0764-Optimise-random-block-ticking.patch
@@ -71,7 +71,7 @@ index 0000000000000000000000000000000000000000..e8b4053babe46999980b926431254050
 +    }
 +}
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index 9fa7a01db61ba91b79eda4f7e4fd4f0001712808..6dee3e414e378538e6b5ef93dc1414e0128cffbb 100644
+index 8f27ef86cac71e0a0002093f4eded83f8891876f..6b90582dd0aefccab7e1b5a1718f96854d1cdd81 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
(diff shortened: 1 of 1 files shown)


NoahvdAa (approved):
Can confirm this fixes the issue, the first cauldron started filling after a few seconds, whereas without this change nothing would happen, even after 5+ minutes of rain.

Pretty useless picture

electronicboy (approved):

kennytv (approved):

PR #7609, opened 2022-03-15
Convert to configurate for paper.yml

Machine-Maker:
### Key Points
* This does away with the traditional way (started by spigot) of reading configs in favor of using configurate.
* Splits up the config into multiple files, a global file, a world-defaults file, and a file for each world that users can place overrides.
* The current plan, is to have 1 patch at the beginning that new options are added to, and then those options are "implemented" in their own patch. This means the `GlobalConfiguration` and `WorldConfiguration` classes should only ever be modified inside one patch. 

#### New directory structure
```
root/
	config/
		paper-global.yml
		paper-world-defaults.yml
	world/
		level.dat
		...
		paper-world.yml
	world_nether/
		level.dat
		paper-world.yml
```

### TODO

- [ ] Configurate 4.2 (full YAML comment support)
- [ ] (depends on ^) Add `@Setting` to all options.
- [x] remove PaperConfig & PaperWorldConfig
- [x] edit patches that add config options to make them use the new config objects
- [x] write migrations from current format to new one
- [x] leave backup of legacy paper.yml for reference (moved to `config/legacy-backup/paper.yml`)


jacksyyy:
I think this is very good idea :shipit: 👍 

My only issue is the per world config file in the world folder. is this not confusing for the admin? How will this work with #7594?

This seems like it has pro of portable world. but if everything is one levelstem with vanilla this becomes very messy.

here is my idea for better
```
root/
	config/
		paper-global.yml
		paper-world-defaults.yml
                worlds/
                    paper-world.yml
                    paper-world_the_end.yml
                    paper-world_nether.yml
```

it is just idea. I think it work for the server admin and #7594 more.

electronicboy:
There are pros and cons to both approaches, we somewhat figured that config files in the worlds seemed to be the best solution, especially in terms of things like minigame servers which carry worlds around, etc, it's much easier for them if the world is bundled with the config

the levelstem issue is complex, but, individual dimensions will still need their own level.dat, so it's not like everything is gonna be shared across stuff, just there are many complexities there as the entire systems both go against one another, data packs expect to be global state yet CB stuff expects some level of isolation

that file structure would also have to look more akin to

```
root/config/worlds/minecraft/world/paper-world.yml
root/config/worlds/minecraft/world_the_end/paper-world.yml
root/config/worlds/minecraft/world_nether/paper-world.yml
root/config/worlds/mymagicalpack/magical_world/paper-world.yml
etc
```

which is also somewhat ugly and doesn't really solve much vs keeping them with the world data

jacksyyy:
it does not have to be like that, you can even use full namespace
```
root/config/worlds/minecraft:world.yml
root/config/worlds/minecraft:world_the_end.yml
root/config/worlds/minecraft:world_nether.yml
root/config/worlds/mymagicalpack:magical_world.yml
etc
```
I think this general idea looks much better. but `:` in filename is not working well probabably.
so maybe replace with `-` or other better char not valid as namespacedkey.

This is just my idea. it is your project. I like it!

electronicboy:
`:` in filenames is not supported, - is a valid seperator elsewhere and that form of substitution just often creates its own weird compat issues in niche cases which people often forget about and then spend 2 weeks debugging

jacksyyy:
I mention this in my comment. even with replace with another char it is still to me more user friendly. but i respect whatever final decision. i just bringing more ideas hopefully.

PhilCCMG:
This is a good move providing everything migrates fully and automatically (as it seems to be so far) and the old paper.yml has a huge notice saying configs have moved to avoid confusion.

Regarding location, I personally think something like
/root/world/paper.yml, /root/world_nether/paper.yml etc would be best, as it means optimised and configured worlds can easily be transferred between people in archive form.

Machine-Maker:
Updated for latest paper config changes

patches/api/0002-Build-system-changes.patch
```diff
@@ -5,14 +5,15 @@ Subject: [PATCH] Build system changes
...
+index b81a893bd5418779544872eb4006adc6b3017a43..1624492b5c76a57ba69416b888baaff9c60de524 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
-@@ -14,15 +14,27 @@ dependencies {
+@@ -14,15 +14,28 @@ dependencies {
      api("com.google.code.gson:gson:2.8.9")
      api("net.md-5:bungeecord-chat:1.16-R0.4")
      api("org.yaml:snakeyaml:1.30")
 +    // Paper start
++    api("org.spongepowered:configurate-yaml:4.1.2")
```
kennytv: Don't expose it yet

patches/server/0004-Paper-config-files.patch
```diff
@@ -1,133 +1,3019 @@
...
++
++    public static void registerCommands(final MinecraftServer server) {
++        COMMANDS.forEach((s, command) -> {
++            server.server.getCommandMap().register(s, "Paper", command);
++        });
++
++        if (!metricsStarted) {
++            Metrics.PaperMetrics.startMetrics();
++            metricsStarted = true;
++        }
```
kennytv: Move commands and registration somewhere else

patches/server/0004-Paper-config-files.patch
```diff
@@ -1,133 +1,3019 @@
...
++
++            If you need help with the configuration or have any questions related to Paper,
++            join us in our Discord or IRC channel.
++
++            Configuration options here apply to all worlds, unless you specify overrides inside
++            the world-specific config file inside each world folder.
++
++            Discord: https://discord.gg/papermc
++            IRC: #paper @ irc.esper.net ( https://webchat.esper.net/?channels=paper )
++            Website: https://papermc.io/
```
kennytv, line 816: I'd specifically mention the docs in the first paragraph of each and remove irc from the text and links

kennytv (commented):
Also, making configurable messages use the minimessage format would be good to have, there won't be a better time to do that than now

Roan-V:
Are there plans to/would it be possible to also move bukkit.yml and spigot.yml into `root/config/`?

electronicboy:
atm, no; There is some plans to merge them all into a single file post hardfork

Roan-V:
Doesn't that basically defeat the point of having a config folder?
A config folder is nice, but not if it's inconsistent and a bunch of other files are somewhere else.

> plans to ... post hardfork

Also doesn't say much sadly, a hard fork has always been this nebulous thing in *the future*

Machine-Maker:
> A config folder is nice, but not if it's inconsistent and a bunch of other files are somewhere else.

It leaves open the opportunity to have additional files if the need arises. Like what if there is a new config thing that ends up being super big. It might warrant having its own file. If we already have a directory, it makes that super easy. 

Also we already have 2 files, the world defaults and the global. So those are best not in the root of the server

PR #7610, opened 2022-03-16
Fix world saving in unloadWorld

ComputerNerd100:
This patch updates the unloadWorld method so that savingDisabled is set to false, which should then mean ServerLevel's saving logic is executed correctly.
PR in draft status as I'm still testing this change.

### DIFF
--- a/patches/server/0883-Fix-saving-in-unloadWorld.patch
+++ b/patches/server/0883-Fix-saving-in-unloadWorld.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Philip Kelley <philip@thoriumcube.org>
+Date: Wed, 16 Mar 2022 12:05:59 +0000
+Subject: [PATCH] Fix saving in unloadWorld
+
+Change savingDisabled to false to ensure ServerLevel's saving logic gets called when unloadWorld is called with save = true
+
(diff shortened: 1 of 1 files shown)


MiniDigger:
good catch! this seems to be broken since 1.14......

ComputerNerd100:
Since 1.14!!!! Damn...
I'm just starting testing of this at the moment, probably gonna take a couple of hours because I want to make sure this doesn't break anything else before releasing for review.
If it does fix this, then I'll open a bug report upstream in Spigot too - its fundamentally a CraftBukkit bug... But at least it'll work on Paper (hopefully!)

electronicboy:
My brain short circuted and I hit the "ready to review" button because apparently I thought it was the "do testing" thing

electronicboy (approved):

ComputerNerd100:
Haha no worries!
I'll mark it ready shortly, wanted to get the PR opened once it was in progress so track anything I find in testing, but yeah I'm pretty confident it should (should TM) be OK :D

ComputerNerd100:
PR released for merging - extensively tested and doesn't break anything else, and fixes worlds being saved when Bukkit.unloadWorld is called

ComputerNerd100:
Just an additional comment on this PR that, as this is an upstream CB issue, I've opened a Spigot issue for it (SPIGOT-6967).
I'll keep an eye on this issue as this patch may no longer be required if CB fix it upstream.

Machine-Maker (approved):

PR #7613, opened 2022-03-16
Fix worldgen datapacks not taking effect on first load

Machine-Maker:
If the `level.dat` file is missing for a dimension, it generates a default one, but that default one fails to account for custom stuff added from data packs. this just runs the default LevelStem registry through the bound registry loader taking in changes from data packs.


Machine-Maker:
Rebased and tested on 1.19.1

Machine-Maker:
Fixed in 22w42a (first snapshot for 1.19.3)

PR #7615, opened 2022-03-17
Add enchantWithLevels API (closes #7536)

### LINKED ISSUE #7536: Enchanting Table like enchantment Application

In some cases you want to have an item enchanted randomly like a enchantment table would do an a specific level. In my case, I want to give a wandering trader a tool, that was enchanted on Lvl 30. But there might be many more applications. Yes, I could try to replicate the behavior of the enchantment table but I guess it would be interesting for some more people.

Give access to the enchantment table functions or create similar code to the one of the enchantment table for enchanting an item.

I could try to replicate the behavior of the enchantment table but I guess it would be interesting for some more people and I don't think it is so much effort to implement it in paper. But I might be wrong.

### Other

_No response_

jpenilla:
Technically some levels values over 30 will work, however in vanilla it's always capped to 30 and at a certain point it just stops working. I think it's simpler to just limit to 30.

closes #7536

### DIFF
--- a/patches/api/0375-Add-enchantItem-API.patch
+++ b/patches/api/0375-Add-enchantItem-API.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add enchantItem API
 
 
 diff --git a/src/main/java/org/bukkit/inventory/ItemFactory.java b/src/main/java/org/bukkit/inventory/ItemFactory.java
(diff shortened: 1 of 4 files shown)


Machine-Maker (approved):
I think ItemFactory makes sense here. (I think some of the stuff I added to UnsafeValues should really be here too). You are literally creating items from some sort of process or factory.

kennytv (approved):

PR #7621, opened 2022-03-18
API for locating configured structures

Machine-Maker:


electronicboy (approved):

electronicboy:
somewhat iffy on exposing lang3 just for this, but, it's also been bundled for years, etc

jpenilla:
I feel like a LocateResult type would be cleaner than exposing Lang3 for it's Pair

Machine-Maker:
ConfiguredStructure was replaced

PR #7622, opened 2022-03-18
Add raw iron & raw copper blocks to anti xray defaults

Warriorrrr:
Raw iron and raw copper blocks can generate naturally but these seem to never have been added to the anti xray defaults


Machine-Maker:
So the concern from @stonar96 with adding more blocks to the hidden list is the effect is has on performance under engine mode 2. But seeing as these blocks are always located in the super large ore veins of iron and copper, maybe it is worth it to have them included.

Warriorrrr:
Could this PR be re-opened? Didn't notice it got stale'd

lynxplay:
Got you :+1:

Warriorrrr:
Rebased for 1.19

Owen1212055 (approved):

electronicboy (approved):

PR #7625, opened 2022-03-19
Add EntityDyeEvent and CollarColorable interface

### LINKED ISSUE #7477: Make Cat and Wolf implement Colorable, add master coloring event

Instances of `Cat` and `Wolf` and colorable just like `Sheep` and `Shulker`, however they dont implement `Colorable`. And despite both sharing a collar to be colored, they don't share an interface like "`ColorableCollar`" or something of the likes.

1. Make `Cat` and `Wolf` implement the `Colorable` class
2. Make `Colorable#getColor()` an alias of `Cat/Wolf#getCollarColor`
3. Make `Colorable#setColor()` an alias of  `Cat/Wolf#setCollarColor`
4. If points 1-3 are out of the question -> `ColorableCollar` interface for `Cat` and `Wolf`

5. Add a master color change event for players changing the colors of entities ingame, for example: `PlayerDyeEntityEvent`, of which the existent `SheepDyeWoolEvent` should be a sub class. This new event should fire whenever a player changes the color of an entity.
6. And if possible, a parent event for `PlayerDyeEntityEvent` -> `EntityChangeColorEvent`, for whenever the color of an entity is changed, which would also include shulkers.

To clarify, the intended hierarchy of new coloring events, from parent -> child:
```
EntityEvent -> EntityChangeColorEvent -> PlayerDyeEntityEvent -> SheepDyeWoolEvent
```

Right now when i want to check for a cat or wolf having their collar color changed, you gotta do real hacky jazz via PlayerInteractEvent. Need i say more?

### Other

Thank you

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/7477

Adds EntityDyeEvent and makes it the super type for SheepDyeWoolEvent. Also adds CollarColorable interface for Wolf/Cat. 

EntityDyeEvent is called for sheep, wolf, and cat currently. Shulkers can be dyed, but not naturally so they were excluded.

### DIFF
--- a/patches/api/0381-Add-EntityDyeEvent-and-CollarColorable-interface.patch
+++ b/patches/api/0381-Add-EntityDyeEvent-and-CollarColorable-interface.patch
@@ -1,0 +1,262 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Fri, 18 Mar 2022 21:16:38 -0700
+Subject: [PATCH] Add EntityDyeEvent and CollarColorable interface
+
+
+diff --git a/src/main/java/io/papermc/paper/entity/CollarColorable.java b/src/main/java/io/papermc/paper/entity/CollarColorable.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..bb68e6a2528eee81eb3f26f22b9c35508f1e69c1
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/CollarColorable.java
+@@ -0,0 +1,25 @@
++package io.papermc.paper.entity;
++
++import org.bukkit.DyeColor;
++import org.bukkit.entity.LivingEntity;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Entities that can have their collars colored.
++ */
++public interface CollarColorable extends LivingEntity {
++
++    /**
++     * Get the collar color of this entity
++     *
++     * @return the color of the collar
++     */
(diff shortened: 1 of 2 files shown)


patches/api/0376-Add-EntityDyeEvent-and-CollarColorable-interface.patch
```diff
@@ -0,0 +1,257 @@
...
++
++    /**
++     * Get the collar color of this entity
++     *
++     * @return the color of the collar
++     */
++    @NotNull DyeColor getCollarColor();
++
++    /**
++     * Set the collar color of this wolf
```
Doc94:
```suggestion
+     * Set the collar color of this entity
```

patches/api/0376-Add-EntityDyeEvent-and-CollarColorable-interface.patch
```diff
@@ -0,0 +1,257 @@
...
++    public EntityDyeEvent(@NotNull Entity entity, @NotNull DyeColor dyeColor, @Nullable Player player) {
++        super(entity);
++        this.dyeColor = dyeColor;
++        this.player = player;
++    }
++
++    /**
++     * Gets the DyeColor the sheep is being dyed
++     *
++     * @return the DyeColor the sheep is being dyed
```
Doc94: This can be more generic? based in this event its not only for sheeps?
this for all the methods in the class

kennytv (changes requested):
missing static handlerlist

Machine-Maker:
Sounds like there should be a test for that... :). Added that method

kennytv (approved):

PR #7628, opened 2022-03-19
Improve CraftChunk#getEntities

### LINKED ISSUE #5872: Getting entities in just loaded chunk returns an empty collection
So that it's documented before someone forgets:
https://hub.spigotmc.org/jira/browse/SPIGOT-6526 was fixed on Spigot and is confirmed to still be unfixed on Paper, likely due to the async chunk loading patch.

Sync loading a chunk and getting the entities returns an empty collection.
```java
    @Override
    public void onEnable() {
        World world = Bukkit.getWorlds().get(0);
        Chunk chunk = world.getChunkAt(world.getSpawnLocation()); // Sync load
        System.out.println("Chunk loaded: " + chunk.isLoaded() + " entities: " + chunk.getEntities().length); // will be 0, even if entities exist in the chunk

        // Spawn an entity for next startup
        String name = "ID: " + System.currentTimeMillis();
        ArmorStand itemFrame = (ArmorStand) world.spawnEntity(world.getSpawnLocation(), EntityType.ARMOR_STAND);
        itemFrame.setCustomName(name);
        itemFrame.setCustomNameVisible(true);

        // See what happens after a manual unload/load
        System.out.println("Chunk loaded: " + chunk.isLoaded() + " " + chunk.getEntities().length);
        boolean unloaded = chunk.unload();
        System.out.println("Chunk unloaded: " + unloaded);
        chunk.load();
        System.out.println("Chunk loaded: " + chunk.isLoaded() + " " + chunk.getEntities().length);
        System.out.println("Please restart server");
    }
```

Machine-Maker:
This is what I came up with that fixes https://github.com/PaperMC/Paper/issues/5872 (the whole can't getEntities from CraftChunk). Just gonna do a draft cause I'm not super solid on how this part of the server works, and *if* we should even fix the issue since mojang separated entity and chunk loading.

### DIFF
--- a/patches/server/0908-improve-CraftChunk-getEntities.patch
+++ b/patches/server/0908-improve-CraftChunk-getEntities.patch
@@ -1,0 +1,153 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Fri, 5 Nov 2021 14:18:22 -0700
+Subject: [PATCH] improve CraftChunk#getEntities
+
+
+diff --git a/src/main/java/io/papermc/paper/world/ChunkEntitySlices.java b/src/main/java/io/papermc/paper/world/ChunkEntitySlices.java
+index 47b5f75d9f27cf3ab947fd1f69cbd609fb9f2749..b92e703abdff9e9d6f7543e6a689112ab40704b1 100644
+--- a/src/main/java/io/papermc/paper/world/ChunkEntitySlices.java
++++ b/src/main/java/io/papermc/paper/world/ChunkEntitySlices.java
+@@ -58,7 +58,7 @@ public final class ChunkEntitySlices {
+                 continue;
+             }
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Almighty-Satan:
This fixes the issue for me

Machine-Maker:
Made some more changes addressing some issues I found. So there are 3 situations in which Chunk#getEntities can be called.

1. The chunk entity status is FRESH which means no entity load has been requested *yet*. In this case a full sync load of the entities is performed.
2. The chunk entity status is PENDING which means that something has already requested the entities be loaded. In this case, the entities are sync loaded like before, but the pending load is removed from the loadingInbox. I added a check in EntityStorage after the file IO operation happens that if the chunk isn't pending anymore to just cancel the op. I think in this case its possible for the entities to be read from the file twice but only 1 load will happen.
3. The chunk entities are already loaded in which case nothing happens.


Should we add a CompletableFuture version of getEntities that just completes when the entities are loaded as normal?

lynxplay:
I am closing this after the leaf chunk rewrite, the issue this PR intends to fix is already resolved :+1:

PR #7634, opened 2022-03-20
Clean up removed tags after datapack unload

Machine-Maker:
*see patch description*

There is [this mojira issue](https://bugs.mojang.com/browse/MC-249007), but its not *exactly* what this fixes since they are reporting it as a client issue (which it probably still is)


Machine-Maker:
Turns out the client does still maintain a reference to the old tag, even if you leave the server and join back. The protocol just doesn't support removing tags. It's possible its worth fixing it just on the server as at least the server-side functionality which might depend on that tag.

Unsure tho.

PR #7639, opened 2022-03-20
Add Position

Machine-Maker:
Position should be the eventual *long-term* replacement for lots of `Location` stuff. But for the moment, it shouldn't hurt to add it and start using it in new API where appropriate.

**Feedback very encouraged as this interface will be used quite a bit.**

Am not 💯 on the "aboveBlock, belowBlock, above, below" methods. I pretty much just took them from nms' BlockPos class, but Position isn't exactly the same since it has to account for more than just int block positions.

An alternative is to literally just have a BlockPos. I'm not sure if there are use cases for double coordinates separate from the world. Like NMS doesn't have that as a type anyways


patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,361 @@
...
++     */
++    @Contract("-> new")
++    default @NotNull Position blockAbove() {
++        return this.blockAbove(1);
++    }
++
++    /**
++     * Gets a position for the block the specified distance above
++     *
++     * @param distance the distance up to travel
```
NoahvdAa:
```suggestion
+     * @param distance the distance to travel up
```
same for blockBelow

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++    default @NotNull BlockPosition toBlock() {
++        return new BlockPositionImpl(this.blockX(), this.blockY(), this.blockZ());
++    }
++
++    /**
++     * Converts this position to a vector
++     *
++     * @return a new vector
++     */
++    @Contract("-> new")
```
kashike:
```suggestion
+    @Contract(value = "-> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++        return new Vector(this.x(), this.y(), this.z());
++    }
++
++    /**
++     * Creates a new location object at this position with the specified world
++     *
++     * @param world the world for the location object
++     * @return a new location
++     */
++    @Contract("_ -> new")
```
kashike:
```suggestion
+    @Contract(value = "_ -> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++
++    /**
++     * Creates a position at the coordinates
++     *
++     * @param x x coord
++     * @param y y coord
++     * @param z z coord
++     * @return a position with those coords
++     */
++    @Contract("_, _, _ -> new")
```
kashike:
```suggestion
+    @Contract(value = "_, _, _ -> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++        return new BlockPositionImpl(x, y, z);
++    }
++
++    /**
++     * Creates a position from the location.
++     *
++     * @param location the location to copy the position of
++     * @return a new position at that location
++     */
++    @Contract("_ -> new")
```
kashike:
```suggestion
+    @Contract(value = "_ -> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++
++    /**
++     * Creates a position at the coordinates
++     *
++     * @param x x coord
++     * @param y y coord
++     * @param z z coord
++     * @return a position with those coords
++     */
++    @Contract("_, _, _ -> new")
```
kashike:
```suggestion
+    @Contract(value = "_, _, _ -> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++        return new FinePositionImpl(x, y, z);
++    }
++
++    /**
++     * Creates a position from the location.
++     *
++     * @param location the location to copy the position of
++     * @return a new position at that location
++     */
++    @Contract("_ -> new")
```
kashike:
```suggestion
+    @Contract(value = "_ -> new", pure = true)
```

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++        return new FinePositionImpl(this.blockX() + 0.5, this.blockY() + 0.5, this.blockZ() + 0.5);
++    }
++
++    /**
++     * Returns the block position of this position
++     * or itself if it already is a block position
++     *
++     * @return the block position
++     */
++    default @NotNull BlockPosition toBlock() {
```
kashike: `@Contract(pure = true)`

patches/api/0377-Add-Position.patch
```diff
@@ -0,0 +1,382 @@
...
++    default @NotNull FinePosition below(double distance) {
++        return new FinePositionImpl(this.x(), this.y() - distance, this.z());
++    }
++
++    /**
++     * Returns a new position at the center of the block position this represents
++     *
++     * @return a new center position
++     */
++    @Contract("-> new")
```
kashike: Invalid annotation, change to:
```suggestion
+    @Contract(pure = true)
```

kashike (dismissed):

kennytv (approved):

kezz:
I think both the fine/block position interfaces should override a lot more methods (like both the isBlock/is fine methods and the various toX methods) just to avoid unnecessary instanceofs and new object creation?

Machine-Maker:
@kezz ok, I moved the isFine, isBlock methods, and the toBlock impl. but I don't see any other methods that would benefit from that? could you clarify (if there are indeed more of them)

patches/api/0377-Add-Position.patch
zml2008: maybe it's time for contract-validator? :p

Owen1212055 (approved):
Unsure if we want to make a vector equivalent, but at least this initial api is a good start.

Machine-Maker:
Rebased for 1.19.3 and added javadocs explaining the Experimental annotation

jaylawl:
Is it going to be immutable?! 😍

olijeffers0n:
It *is*

imDaniX:
> Is it going to be immutable?! heart_eyes

Well, it's still up to implementation. While new default implementations are indeed immutable, `Location` now implements `Position` too, so `x()` and other getters may return different results each call.

jaylawl:
> > Is it going to be immutable?! heart_eyes
> 
> Well, it's still up to implementation. While new default implementations are indeed immutable, `Location` now implements `Position` too, so `x()` and other getters may return different results each call.

In any case this is a step in the right direction. Despite being aware of mutability of locations, vectors, and bounding boxes, "accidental mutation" that was hard to find in the code, just keeps on happening. I've always wished all those classes were immutable.

PR #7644, opened 2022-03-21
Grant temporary immunity from EAR to moving entities (Fixes #7637)

### LINKED ISSUE #7637: EAR culls ticks for items pushed over ice, breaking fast sorters in the Nether when out of player activation misc range
### Expected behavior

Item stacks pushed by a piston + slime over blue ice take about 10 seconds to dissipate their momentum. This is frequently used to make high capacity sorters in the Nether by moving the items over hopper filters. 

### Observed/Actual behavior

Once an item stack is more than 10 seconds old and out of player misc activation range the items start behaving erratically with a significnat desync between server and client. The item stacks stop short of their normal range after the 10 seconds. 

### Steps/models to reproduce

I use a test path of 1 glazed terracotta and 33 blue ice blocks in a line with a piston+slime at the terracotta end. Bumping a stack of items with the piston+slime makes them reach the end if the stack is younger than the 10s EAR immunity or the entire track is within the player activation range (misc category). Otherwise the items reach far shorter.

### Plugin and Datapack List

Reproduced without any plugins

### Paper version

This server is running Paper version git-Paper-265 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 993f828)
Previous version: git-Paper-259 (MC: 1.18.2)

### Other

I can thing of two approaches off the top of my head: 
- Increase the immunity to tick culling after an EntityItem is pushed by a piston from 0.5s to 10s or make it configurable
- Increase immunity to tick culling when on ice. The default 0.5s is sufficient for non-slippery surfaces. This is more in line with how water streams are currently handled.

petersv5:
Fix for issue  #7637:
Moving entities behave erratically when EAR tick decimation activates. This

breaks e.g. ice based sorters in the Nether. To counter this a test based
on the test-for-movement in ItemEntity.tick() is added to to the
ActivationRange, granting a temporary immunity. Once the momentum is lost
the item again goes inactive.

### DIFF
--- a/patches/server/0352-Entity-Activation-Range-2.0.patch
+++ b/patches/server/0352-Entity-Activation-Range-2.0.patch
@@ -14,7 +14,7 @@ Adds flying monsters to control ghast and phantoms
 Adds villagers as separate config
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index 7f2b61523ac2293fdc197cf3eaa341357cf6fc31..75d91cb0f432fb9a2b5b9a796ad2a60f4c891060 100644
+index 9b883af58fd87751bdad909a015cb78ca5647e90..12162ff2dc7c82f50f1d892bc807985ebcd44333 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
 @@ -2,7 +2,6 @@ package net.minecraft.server.level;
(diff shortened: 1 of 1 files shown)


petersv5:
In case you are happy with this approach in this patch, do you want me to provide this as an updated vesion of the EAR patch (0364) instead of as a new patch as I have now?

kennytv:
Yes please, and a rebase

petersv5:
Oh, I seem to not have squashed my patches. Let me fix that as well. One moment.

petersv5:
Ok, should be all done moving into the EAR 2.0 patch and rebase.

MiniDigger (approved):

petersv5:
Want me to do anything else or will you resolve the conflict?
I presume a rebase is needed?

petersv5:
Rebased to new patch numbers.

petersv5:
Anything else that needs doing?

electronicboy (approved):

PR #7648, opened 2022-03-22
Add api to resolve components

### LINKED ISSUE #5413: SelectorComponents do not get resolved
### Expected behavior

Get a message that contains the full list of entities

### Observed/Actual behavior

Get message `@e`

### Steps/models to reproduce
```java
@Override
public boolean onCommand(CommandSender sender, Command command, String label, String[] args) {
    Component component = Component.selector("@e");
    sender.sendMessage(component);
}
```
### Plugin list

None, except the one I used to verify this behavior.

### Paper version

> version
[00:16:21 INFO]: Checking version, please wait...
[00:16:21 INFO]: This server is running Paper version git-Paper-567 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)

### Agreements

Creating and sending `SelectorComponent`s does currently not work properly as the selector does not get resolved. I haven't tested it but same problem might apply to all component types listed [here](https://minecraft.fandom.com/wiki/Raw_JSON_text_format#Component_resolution) (i.e. `score`, `selector`, and `nbt`)

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/5413

Perhaps some mention of the permission craftbukkit adds `minecraft.command.selector` since that will block the use of some components that make use of selectors if the CommandSender doesn't have that permission?

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -553,10 +553,10 @@ index 0000000000000000000000000000000000000000..eb179aae1e1d2ce842442e49fe275827
 +}
 diff --git a/src/main/java/io/papermc/paper/text/PaperComponents.java b/src/main/java/io/papermc/paper/text/PaperComponents.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..bff9a6295db367c6b89d69fb55459a40828265ea
+index 0000000000000000000000000000000000000000..d6ec841c3878bd682c52307356aeebca9666ac3a
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/text/PaperComponents.java
-@@ -0,0 +1,112 @@
+@@ -0,0 +1,146 @@
 +package io.papermc.paper.text;
 +
 +import net.kyori.adventure.text.Component;
@@ -567,7 +567,12 @@ index 0000000000000000000000000000000000000000..bff9a6295db367c6b89d69fb55459a40
 +import net.kyori.adventure.text.serializer.plain.PlainComponentSerializer;
 +import net.kyori.adventure.text.serializer.plain.PlainTextComponentSerializer;
 +import org.bukkit.Bukkit;
++import org.bukkit.command.CommandSender;
++import org.bukkit.entity.Entity;
 +import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++import java.io.IOException;
 +
 +/**
 + * Paper API-specific methods for working with {@link Component}s and related.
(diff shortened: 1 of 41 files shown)


patches/api/0005-Adventure.patch
```diff
@@ -383,6 +388,19 @@ index 0000000000000000000000000000000000000000..15ecb12fd2fefcac96edbaef7cdd487a
...
 +    /**
++     * Resolves a component with a specific command sender and subject
++     *
++     * @param input the component to resolve
++     * @param context the command sender to resolve with
++     * @param scoreboardSubject the scoreboard subject to use (for use with {@link net.kyori.adventure.text.ScoreComponent}s)
++     * @return the resolved component
++     * @throws IOException if a syntax error tripped during resolving
++     */
++    public static @NotNull Component resolve(@NotNull Component input, @Nullable CommandSender context, @Nullable Entity scoreboardSubject) throws IOException {
```
kashike:
```suggestion
+    public static @NotNull Component resolveWithContext(@NotNull Component input, @Nullable CommandSender context, @Nullable Entity scoreboardSubject) throws IOException {
```

patches/api/0005-Adventure.patch
```diff
@@ -1114,6 +1132,7 @@ index 01e11f882abb6c631f810584aa23646042688435..f43ad19a44f3b7368a1a110cc5524727
 +    net.kyori.adventure.text.serializer.gson.GsonComponentSerializer gsonComponentSerializer();
 +    net.kyori.adventure.text.serializer.gson.GsonComponentSerializer colorDownsamplingGsonComponentSerializer();
 +    net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer legacyComponentSerializer();
++    net.kyori.adventure.text.Component resolve(net.kyori.adventure.text.Component component, org.bukkit.command.CommandSender context, org.bukkit.entity.Entity scoreboardSubject) throws java.io.IOException;
```
kashike:
```suggestion
+    net.kyori.adventure.text.Component resolveWithContext(net.kyori.adventure.text.Component component, org.bukkit.command.CommandSender context, org.bukkit.entity.Entity scoreboardSubject) throws java.io.IOException;
```

patches/api/0005-Adventure.patch
Machine-Maker: renamed as such

kashike (approved):

Machine-Maker:
Rebased for 1.19.1

Owen1212055 (approved):
Similar to the instrument PR, maybe this can also be moved to its own provider api.
ComponentResolverProvider

Something like that. Logic looks good tho.

jpenilla:
The selectEntities method does something similar with creating a CommandSourceStack and it adds a boolean param to override permissions, might want to do the same here

Machine-Maker:
> The selectEntities method does something similar with creating a CommandSourceStack and it adds a boolean param to override permissions, might want to do the same here

Ok did this, entity selector permission check is bypassed. 


I also expanded the javadocs to explain what components can be resolved, and what is required/optional for each type.

jpenilla (approved):
Are there situations where we would want to check permissions? Maybe a boolean param for the override would be good. Looks good otherwise

PR #7652, opened 2022-03-22
Add restrict-player-reloot-time config

### LINKED ISSUE #3945: Request - Time limit for restrict-player-reloot to allow looting again after a set time
I like loot chests to refill often for newer players but I also see the same regular players logging in just to reloot. I don't want to disable reloot forever as it would be a negative for long term players. 

I'd like a setting in the config to go alongside the restrict player reloot setting so I can restrict player reloot for a set amount of time before it allows them to loot the same chests again. For example restrict-player-reloot-time: 7d

This would allow refilling of chests often and also allow players to reloot eventually but less often, discouraging logging in just for that reason. 

The only other options I can see are either preventing reloot forever or disabling loot refill, both of which are undesirable. 

**Additional context**
N/A

Machine-Maker:
Closes https://github.com/PaperMC/Paper/issues/3945

### DIFF
--- a/patches/api/0040-LootTable-API.patch
+++ b/patches/api/0040-LootTable-API.patch
@@ -58,10 +58,10 @@ index 0000000000000000000000000000000000000000..b387894fe8001edb41ad2ad2b70ebabe
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/loottable/LootableInventory.java b/src/main/java/com/destroystokyo/paper/loottable/LootableInventory.java
 new file mode 100644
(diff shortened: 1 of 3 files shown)


kennytv (approved):
Not sure about `0s` as a default for "permanent", but otherwise good

Machine-Maker:
Well I did `0` cause having a time of 0 is the same as not enabling `restrict-player-reloot` as the player can just reboot instantly.

electronicboy (approved):

PR #7666, opened 2022-03-27
Add (very strict) option to hide entities by default. Fixes #7395

### LINKED ISSUE #7395: Create Entity#hideEntityByDefault

### Other

_No response_

CrystallDEV:
Added an option to hide entities by default. This option is
 currently very strict and overrides all other changes to visibility and
 interaction. 
 
 Tested everything locally and seems to work fine.
 Could maybe be improved by adding a list of players to which entities that are hidden by default are still visible to


patches/api/0377-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,36 @@
...
++
++    /**
++     * Hides this entity from all players by default
++     */
++    void hideEntityByDefault();
++
++    /**
++     * Shows this entity to all players by default
++     */
++    void showEntityByDefault();
```
NoahvdAa: This should be one setter, `setHiddenByDefault`.

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,118 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+index 6feeadad9aecb7d63e24d5daae115a93e39aeb3d..f861b695aac3c8633e3bb308cdda42fe5e7c5f06 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -1638,7 +1638,7 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+ 
+     @Override
+     public boolean canSee(org.bukkit.entity.Entity entity) {
+-        return !this.hiddenEntities.containsKey(entity.getUniqueId());
++        return !this.hiddenEntities.containsKey(entity.getUniqueId()) || !entity.isHiddenByDefault();
```
u9g: wouldn't you want an && here not || ?

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,118 @@
...
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -1524,6 +1524,11 @@ public class ServerLevel extends Level implements WorldGenLevel {
+ 
+     @Override
+     public void playSound(@Nullable Player except, Entity entity, SoundEvent sound, SoundSource category, float volume, float pitch) {
++        // Paper start - dont play sounds of entities that are hidden by default
++        if(entity.isHiddenEntityByDefault()){
++            return;
++        }
```
Owen1212055: I would instead return if the entity is hidden at all (CraftPlayer#canSee)

electronicboy:
Thanks for the PR, however; "by default" here is a massive misnomer given that it just hides entirely
"by default" would imply that it auto adds, but, am not 100% sure about this behavior, would probs need to be tied into the hide/show player methods which take a plugin and replace that with a key rather than just a plugin, but, not 100% there, then that can be tracked and managed better

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
CrystallDEV: No, since an entity can be hidden overall or specifically hidden just for this player

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
CrystallDEV: This would require to have the player that is listening to the sound, which would require me to modify the broadcast method and therefore have a completely different implementation

patches/api/0377-Add-option-to-hide-entities-by-default.-This-option-.patch
CrystallDEV: Indeed, missed updating these in the API when changing the base functions

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
Owen1212055: Ahh, my bad I misunderstood the method parameters. Disregard my comment.

CrystallDEV:
Updated the code to (hopefully) be more clear

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,145 @@
...
++    }
++
+     @Override
+     public Location getOrigin() {
+         Vector originVector = this.getHandle().getOriginVector();
+@@ -1270,7 +1280,7 @@ public abstract class CraftEntity implements org.bukkit.entity.Entity {
+     public boolean isInWaterOrBubbleColumn() {
+         return getHandle().isInWaterOrBubble();
+     }
+-
```
kennytv: Don't remove the spaces here

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,145 @@
...
+ import org.bukkit.craftbukkit.map.RenderData;
+-import org.bukkit.craftbukkit.profile.CraftPlayerProfile;
+ import org.bukkit.craftbukkit.scoreboard.CraftScoreboard;
+ import org.bukkit.craftbukkit.util.CraftChatMessage;
+ import org.bukkit.craftbukkit.util.CraftMagicNumbers;
+@@ -139,7 +137,6 @@ import org.bukkit.map.MapView;
+ import org.bukkit.metadata.MetadataValue;
+ import org.bukkit.plugin.Plugin;
+ import org.bukkit.plugin.messaging.StandardMessenger;
+-import org.bukkit.profile.PlayerProfile;
```
kennytv: Don't remove imports

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,145 @@
...
+diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
+index 3b1a34b34979ab436ccd33f0a85bfae537cbecb4..e1fb9dbfbec34b1f69c9456a5559aafb6a7be275 100644
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -1524,6 +1524,11 @@ public class ServerLevel extends Level implements WorldGenLevel {
+ 
+     @Override
+     public void playSound(@Nullable Player except, Entity entity, SoundEvent sound, SoundSource category, float volume, float pitch) {
++        // Paper start - don't play sounds of entities that are fully hidden
++        if(entity.isFullyHidden()){
```
kennytv:
```suggestion
+        if (entity.isFullyHidden()) {
```

u9g:
not stale

patches/server/0887-Add-option-to-hide-entities-by-default.-This-option-.patch
```diff
@@ -0,0 +1,104 @@
...
+     // Paper start
++
++    @Override
++    public boolean isFullyHidden() {
++        return getHandle().isFullyHidden();
++    }
++
++    @Override
++    public void setFullyHidden(boolean hide) {
++        getHandle().setFullyHidden(true);
```
Lulu13022002, line 86: The parameter hide is not used here so we can't revert back the hidden status to false

CrystallDEV: Good catch

PR #7670, opened 2022-03-27
Fix NPE for BlockDataMeta#getBlockData

Machine-Maker:
idk if it should be fixed here, or fixed in CraftMetaItem to not have a null CompoundTag for blockData


kennytv:
It'd be nicer to patch it in CraftMetaItem.getBlockData, to use `CraftMagicNumbers.getBlock(material).defaultBlockState()` in `fromData()` instead of passing it into `BlockItem.getBlockState` if the block data is null

Machine-Maker:
Ok, made that change.

electronicboy (approved):

PR #7673, opened 2022-03-28
WorldCreator#keepSpawnLoaded

electronicboy:
some other discussion reminded me that I had this branch, and the TriState generally just made this much nicer looking


Machine-Maker:
don't we have a paper.yml setting for this? how does that play into this?

electronicboy:
I apparently forgot javadocs...

defaults to the value in paper.yml, as that's set in the constructor for Level, so we just pick up after that

Machine-Maker:
Right but doesn't that seem a little off? idk, it seems weird that no matter what I do to specify the keep spawn loaded setting in paper.yml, I can't change it. shouldn't the end user have the final say, not the plugin?

patches/api/0377-WorldCreator-keepSpawnLoaded.patch
```diff
@@ -0,0 +1,48 @@
...
++    }
++
++    /**
++     * Controls if a world should be kept loaded or not, default (NOT_SET) will use the servers standard
++     * configuration, otherwise, will act as an override towards this setting
++     *
++     * @param keepSpawnLoaded the new value
++     */
++    public void keepSpawnLoaded(@NotNull net.kyori.adventure.util.TriState keepSpawnLoaded) {
++        com.google.common.base.Preconditions.checkNotNull(keepSpawnLoaded);
```
kashike:
```suggestion
+        java.util.Objects.requireNonNull(keepSpawnLoaded);
```

kashike (dismissed):

electronicboy:
> Right but doesn't that seem a little off? idk, it seems weird that no matter what I do to specify the keep spawn loaded setting in paper.yml, I can't change it. shouldn't the end user have the final say, not the plugin?

This is generally for cases where plugins feel that there is stronger reasons over controlling this, especially when the default option is for *all* worlds to have their spawn loaded which does have implications on the time taken to load worlds, plugins like WorldEdit get around this by modifying the end users configuration file, which is really far from a great solution here.

OFC, it's up to plugin devs to utilise this properly, providing config options where deemed suitable, but, the primary usecase for this is speeding up the time taken to load worlds without having to do the async pipelining magic, and, it's not like there isn't an unsupported path in which plugins can already override this, this just offers a more maintainable solution for plugins

wizjany:
WE actually only use the config to turn off the verbose cuz it spams up the log if you use regen. We bypass the whole keep spawn loaded one way or the other by interacting directly with WorldServer and not touching bukkit api.

kennytv (approved):

PR #7681, opened 2022-03-30
Fix Nullability Annotations for PrepareItemEnchantEvent

Astralchroma:
The return value is marked as NotNull causing Kotlin to disallow me from setting an item in the array to null, despite the JavaDoc saying the contents of the array may be null. This should fix that.

### DIFF
--- a/patches/api/0175-Fix-Spigot-annotation-mistakes.patch
+++ b/patches/api/0175-Fix-Spigot-annotation-mistakes.patch
@@ -9,7 +9,7 @@ a ton of noise to plugin developers.
 These do not help plugin developers if they bring moise noise than value.
 
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index 686235a2347ebeaa5654a14cdd717009f2c0105f..cf7f8a8f03adcbe466b59ea8b98b527fb54a0803 100644
+index 52f5750e1cbce6000c72f5f1b4bde41b9fa66b23..19771a9a7a212a71f4cad33981c3b72341d80093 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
 @@ -1684,7 +1684,7 @@ public final class Bukkit {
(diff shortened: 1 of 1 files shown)


patches/api/0175-Fix-Spigot-annotation-mistakes.patch
```diff
@@ -85,17 +85,43 @@ index f124b35ec76e6cb6a1a0dc464005087043c3efd0..94a2fef0dc9e13c754cd31d5eabc1bde
...
      @NotNull
      public int[] getExpLevelCostsOffered() {
          int[] levelOffers = new int[offers.length];
+@@ -85,7 +87,7 @@ public class PrepareItemEnchantEvent extends InventoryEvent implements Cancellab
+      *
+      * @return list of available enchantment offers
+      */
+-    @NotNull
++    @Nullable // Paper - offers can contain null values
+     public EnchantmentOffer[] getOffers() {
```
Machine-Maker: Well the correct signature is actually `@Nullable EnchantmentOffer @NotNull[] getOffers()`

Machine-Maker (changes requested):

patches/api/0175-Fix-Spigot-annotation-mistakes.patch
Astralchroma: I am not used to nullability notations with Java as I normally use Kotlin, so i'll fix that.

Astralchroma: Fixed.

patches/api/0175-Fix-Spigot-annotation-mistakes.patch
```diff
@@ -96,6 +105,19 @@ index 2ff1b1308571d8f8056d3359e8a8ba4a589c3726..e669ad8ecd182c6899c7820414e6ee1f
...
+@@ -85,8 +86,10 @@ public class PrepareItemEnchantEvent extends InventoryEvent implements Cancellab
+      *
+      * @return list of available enchantment offers
+      */
+-    @NotNull
+-    public EnchantmentOffer[] getOffers() {
++    // Paper start - offers can contain null values
++    // @NotNull
++    public @org.jetbrains.annotations.Nullable EnchantmentOffer @NotNull [] getOffers() {
++    // Paper end
```
Machine-Maker: just put the paper comment at the end of the line. no need to leave the old commented-out annotation or have start/end comments.

Astralchroma: This has been fixed by the way

Machine-Maker (approved):

kennytv (approved):

DragonX1301 (approved):

PR #7684, opened 2022-04-01
Remove World#refreshChunk deprecation

Owen1212055:
Was fixed in https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/c86a3f7a5877acb5406147923ac48d91c2f6e7d4

This deprecation reason doesn't really make sense either.


Machine-Maker:
on a slightly related note, the impl of that method uses a deprecated ctor for `ClientboundLevelChunkWithLightPacket`. Is that an issue? the anti X-ray patch adds a new one that should be used.

Owen1212055:
> on a slightly related note, the impl of that method uses a deprecated ctor for `ClientboundLevelChunkWithLightPacket`. Is that an issue? the anti X-ray patch adds a new one that should be used.

Yeah, see https://github.com/PaperMC/Paper/pull/7214

kennytv (approved):

electronicboy (approved):

PR #7690, opened 2022-04-02
Deprecate getEntityId as a magic value

Machine-Maker:
getEntityId is left over when entities didn't have UUID (I'm guessing) and its only used for network stuff. like the network ids for everything in registries. getUniqueId (which returns a UUID) should be used by plugins.


NoahvdAa (approved):

electronicboy:
This was always a misnomer due to the fact that there are still internal uses for this ID, and, we literally just recently added API which would take this number in order to attribute an action towards a specific entity

PR #7692, opened 2022-04-02
Fix NBT pieces overriding a block entity during worldgen deadlock

### LINKED ISSUE #7669: MC-246262 possible
### Expected behavior

There isn't a patch in PaperMC to my knowledge fixing https://bugs.mojang.com/browse/MC-246262 (Searched through bug and pull requests, empty handed)

### Observed/Actual behavior

IDK

### Steps/models to reproduce

https://bugs.mojang.com/browse/MC-246262

### Plugin and Datapack List

Not applicable

### Paper version

Not applicable

### Other

Not applicable

etil2jz:
https://bugs.mojang.com/browse/MC-246262

By checking if the world passed into StructureTemplate's placeInWorld is not a WorldGenRegion, we can bypass the deadlock entirely.

Closes #7669

### DIFF
--- a/patches/server/0889-Fix-NBT-pieces-overriding-a-block-entity-during-worl.patch
+++ b/patches/server/0889-Fix-NBT-pieces-overriding-a-block-entity-during-worl.patch
@@ -6,39 +6,29 @@ Subject: [PATCH] Fix NBT pieces overriding a block entity during worldgen
 
 
 diff --git a/src/main/java/net/minecraft/world/level/levelgen/structure/templatesystem/StructureTemplate.java b/src/main/java/net/minecraft/world/level/levelgen/structure/templatesystem/StructureTemplate.java
-index c5827e4d870a0bba483649d54cae23072eab2b18..c89081729e46d61268f6c0d80e1d43d7196325cd 100644
+index c5827e4d870a0bba483649d54cae23072eab2b18..332d080ad722a0252d2ae8eb83f7697c1323b4ce 100644
 --- a/src/main/java/net/minecraft/world/level/levelgen/structure/templatesystem/StructureTemplate.java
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Put some reference to the mojira issue inside the patch. This is helpful when someone comes along later and wonders if mojang has fixed the bug and this patch is no longer needed.

patches/server/0889-Fix-NBT-pieces-overriding-a-block-entity-during-worl.patch
```diff
@@ -0,0 +1,37 @@
...
+@@ -380,7 +384,11 @@ public class StructureTemplate {
+                         if (pair1.getSecond() != null) {
+                             tileentity = world.getBlockEntity(blockposition6);
+                             if (tileentity != null) {
+-                                tileentity.setChanged();
++                                // Paper start - Fix NBT pieces overriding a block entity during worldgen deadlock
++                                if (!(world instanceof net.minecraft.world.level.WorldGenLevel)) {
++                                    tileentity.setChanged();
++                                }
++                                // Paper end
```
Machine-Maker: I assume its a non-issue that setChanged isn't called marking the chunk as unsaved? I don't know the particulars but I'm guessing during worldgen that doesn't matter as its all unsaved anyways.

etil2jz: That's what I was thinking of, but yeah it seems to be a non-issue doing this way.
Also, I tried to reproduce with datapack attached on Mojira, no deadlock happened and I didn't encounter any issue that may have been introduced doing this way.

kennytv (approved):

PR #7694, opened 2022-04-02
Add Alternate Current's redstone implementation as an alternative to vanilla and Eigencraft's.

SpaceWalkerRS:
This PR adds [Alternate Current](https://github.com/SpaceWalkerRS/alternate-current)'s redstone implementation as an alternative to vanilla and Eigencraft's. A previous attempt at this was made by [Titaniumtown](https://github.com/Titaniumtown) (#7291), but as they moved on, I decided to have a crack at it.

Alternate Current is many times faster than vanilla, and even exceeds Eigencraft. I will do some testing in the coming days and update this post with some numbers.

Just like Eigencraft, Alternate Current changes redstone dust's update order. This means any redstone contraption that is location dependent in vanilla has a significant chance of breaking with either of the faster implementations. Beyond that, parity issues should be rare, though Alternate Current has not been as thoroughly tested as Eigencraft, so I cannot compare the two in that aspect.

### Performance

[Grids](https://imgur.com/DvIH6fm)
- Vanilla: 295 mspt
- Eigencraft: 54 mspt
- Alternate Current: 17

[Lines](https://imgur.com/Uf5FiIH)
- Vanilla: 202 mspt
- Eigencraft: 65 mspt
- Alternate Current: 21 mspt

The second test is probably closest of the two to a 'real-world' scenario.

### DIFF
--- a/patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
+++ b/patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Space Walker <spacedoesrs@gmail.com>
 Date: Sun, 3 Apr 2022 00:09:40 +0200
-Subject: [PATCH] Add Alternate Current's redstone implementation as an alternative to vanilla and Eigencraft's
+Subject: [PATCH] Add Alternate Current redstone implementation
 
 Author: Space Walker <spacedoesrs@gmail.com>
 
@@ -16,7 +16,7 @@ cannot comment on how the two compare in that aspect.
 
 Alternate Current is more invasive than Eigencraft, though some of the modifications can be removed with only a minor
 performance penalty in a small number of cases. Alternate Current needs the following modifications:
-* Level/ServerLevel: Each level has its own 'wire handler' that handles redstone dust power changes. 
+* Level/ServerLevel: Each level has its own 'wire handler' that handles redstone dust power changes.
 * RedStoneWireBlock: Replace calls to vanilla's or Eigencraft's methods for handling power changes with calls to
 Alternate Current's wire handler.
 * (optional) Every power emitter's block class: new methods added to these classes make it easier for Alternate Current
@@ -293,10 +293,10 @@ index 0000000000000000000000000000000000000000..191ef2fdcbfa4a4bd4707905c9abaf4d
 +}
 diff --git a/src/main/java/alternate/current/wire/PowerQueue.java b/src/main/java/alternate/current/wire/PowerQueue.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..943a529ad88964b902a8f5080da1f37084bbf2d9
+index 0000000000000000000000000000000000000000..4b89ef08947b0adb6cb97afcb7ecd9d06518d05b
 --- /dev/null
 +++ b/src/main/java/alternate/current/wire/PowerQueue.java
-@@ -0,0 +1,227 @@
+@@ -0,0 +1,229 @@
 +package alternate.current.wire;
 +
 +import java.util.AbstractQueue;
@@ -501,13 +501,15 @@ index 0000000000000000000000000000000000000000..943a529ad88964b902a8f5080da1f370
 +	}
 +
 +	private void linkAfter(WireNode prev, WireNode wire) {
-+		prev.next = wire;
-+		wire.prev = prev;
++		linkBetween(prev, wire, prev.next);
++	}
 +
-+		if (prev != tail) {
-+			wire.next = prev.next;
-+			prev.next.prev = wire;
-+		}
++	private void linkBetween(WireNode prev, WireNode wire, WireNode next) {
++		prev.next = wire;
++		wire.next = next;
++
++		wire.prev = prev;
++		next.prev = wire;
 +	}
 +
 +	private WireNode findPrev(WireNode wire) {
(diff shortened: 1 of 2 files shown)


patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2741 @@
...
++	private final ServerLevel level;
++
++	LevelAccess(ServerLevel level) {
++		this.level = level;
++	}
++
++	/**
++	 * A slightly optimized version of Level.getBlockState.
++	 */
++	BlockState getBlockState(BlockPos pos) {
```
kennytv, line 78: The performance improvement this method has seems negligible/unecessary

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2741 @@
...
++ * 
++ * @author Space Walker
++ */
++public class Node {
++
++	// flags that encode the Node type
++	private static final int CONDUCTOR = 0b01;
++	private static final int SOURCE    = 0b10;
++
++	final LevelAccess level;
```
kennytv, line 216: Most of these package private fields would look cleaner as private fields with a getter and/or setter

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2741 @@
...
++
++import alternate.current.wire.redstone.RedstoneWireType;
++
++public class WireTypes {
++
++	public static final WireType REDSTONE;
++
++	static {
++
++		REDSTONE = new RedstoneWireType("Redstone Wire", 1, true, true);
```
kennytv, line 2062: To quote someone:
>what other type of wire is there?
string cheese?

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2741 @@
...
+diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
+index 3b1a34b34979ab436ccd33f0a85bfae537cbecb4..dc114848e72c186c1ac07555913be5cb6ebc3ed0 100644
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -6,6 +6,9 @@ import com.google.common.collect.Lists;
+ import com.mojang.datafixers.DataFixer;
+ import com.mojang.datafixers.util.Pair;
+ import com.mojang.logging.LogUtils;
++
++import alternate.current.wire.WireHandler; // Paper - optimize redstone (Alternate Current)
```
kennytv, line 2170: Instead of adding imports that are only used a few times (~< 5), inline the imports

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2741 @@
...
+index 0000000000000000000000000000000000000000..04b355e2eba47439efd81471a19ac04663f57fea
+--- /dev/null
++++ b/src/main/java/alternate/current/util/BlockUtil.java
+@@ -0,0 +1,14 @@
++package alternate.current.util;
++
++import net.minecraft.core.Direction;
++import net.minecraft.world.level.block.state.BlockBehaviour;
++
++public abstract class BlockUtil extends BlockBehaviour {
```
kennytv, line 39: Why does this need to exist?

kennytv (commented):
Thank you for the pr. Before we continue, we'd like the following to be clear:
1) We want as little maintenance burden on ourselves and updating this patch on alternate-current as well as MC updates or to fix bugs would always get back to you, otherwise it might temporarily (or in case of major issues, permanently) be removed (which the eigencraft patch has taught us)
2) This, as well as eigencraft, should clearly be marked as experimental or temporary options that may disappear at any time, especially between major version updates
3) We'd like some changes on the code, which we realize will either mean you have to change it in your mod or push more maintenance burden onto you to merge your mod's changes into this

If all of that is fine by you, we'd be open to further review and testing (especially regarding redstone events being fired about as expected). If you do want to continue, please reopen this pr from a branch that isn't the master branch on your fork, otherwise that makes the rebase/testing process a bit more tedious for us

SpaceWalkerRS:
> Thank you for the pr. Before we continue, we'd like the following to be clear:
> 
> 1. We want as little maintenance burden on ourselves and updating this patch on alternate-current as well as MC updates or to fix bugs would always get back to you, otherwise it might temporarily (or in case of major issues, permanently) be removed (which the eigencraft patch has taught us)
> 2. This, as well as eigencraft, should clearly be marked as experimental or temporary options that may disappear at any time, especially between major version updates
> 3. We'd like some changes on the code, which we realize will either mean you have to change it in your mod or push more maintenance burden onto you to merge your mod's changes into this
> 
> If all of that is fine by you, we'd be open to further review and testing (especially regarding redstone events being fired about as expected). If you do want to continue, please reopen this pr from a branch that isn't the master branch on your fork, otherwise that makes the rebase/testing process a bit more tedious for us

I understand. I recognize there are parts of the code that exist only because of the constraints that come with modding using Mixin. For the Paper patch these do not need to exist or could be replaced by a small patch. There are also some design choices that I made with the idea that other mods that add their own wire blocks could take advantage of the algorithm for propagating power changes. It's something that I'm still developing, but that also does not need to exist in the Paper patch. I could strip it down to remove some clutter. 

I'll open a PR from a different branch and we can continue from there.

PR #7695, opened 2022-04-03
Allow set custom podium for EnderDragon

### LINKED ISSUE #7693: Add option to override EnderDragon origin / target location

I'm currently working on a project, which requires multiple simultaneous EnderDragon fights (not a real DragonBattle, just killing a "manually" spawned entity). Since the EnderDragon pathfinding is hardcoded to be centered around [0,0] the easiest solution for me was just loading multiple worlds. But now I'm trying to expand on the whole concept, and it would be easier to just have multiple arenas in the same world instead of loading a new world for each fight. But that would require having multiple EnderDragons with different origin locations.

Some possibility to define an override for the origin / target location of an EnderDragon entity.

Maybe something like:

- EnderDragon#setOrigin(@ Nullable Location origin)

which would override the default origin. Setting this value to null could restore the default behavior. (Which is targeting the EndPodiumFeature if I understood the dragon phase pathfinding correctly)

Modifying the pathfinding on my own and recreating default behavior, since I want 1:1 default AI, just not centered at [0,0].

### Other

_No response_

Doc94:
This PR resolve #7693 adding a API for set a custom podium this is handle for the most of logic for the EnderDragon Fight.

### DIFF
--- a/patches/api/0377-Allow-to-change-the-podium-of-the-EnderDragon.patch
+++ b/patches/api/0377-Allow-to-change-the-podium-of-the-EnderDragon.patch
@@ -1,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sat, 2 Apr 2022 23:03:32 -0300
+Subject: [PATCH] Allow to change the podium of the EnderDragon
+
+
+diff --git a/src/main/java/org/bukkit/entity/EnderDragon.java b/src/main/java/org/bukkit/entity/EnderDragon.java
+index 856015b08bfa3f18b3df11e25efd079d4a4f7eca..23cbaf886e230d38b3023923f74d1a6c34cd9b9d 100644
+--- a/src/main/java/org/bukkit/entity/EnderDragon.java
++++ b/src/main/java/org/bukkit/entity/EnderDragon.java
+@@ -104,4 +104,22 @@ public interface EnderDragon extends ComplexLivingEntity, Boss, Mob {
+      * @return this dragon's death animation ticks
+      */
+     int getDeathAnimationTicks();
++
++    // Paper start
++
++    /**
++     * Get the podium location used by the ender dragon.
++     *
++     * @return the podium location of the dragon
++     */
++    @NotNull
++    org.bukkit.Location getPodium();
++
++    /**
++     * Sets the location of the podium for the ender dragon.
++     *
++     * @param location the location of the podium or null to use the default podium location (exit portal of the end)
++     */
++    void setPodium(@Nullable org.bukkit.Location location);
++    // Paper end
+ }
--- a/patches/server/0890-Allow-to-change-the-podium-for-the-EnderDragon.patch
+++ b/patches/server/0890-Allow-to-change-the-podium-for-the-EnderDragon.patch
@@ -1,0 +1,151 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 3 Apr 2022 11:31:42 -0400
+Subject: [PATCH] Allow to change the podium for the EnderDragon
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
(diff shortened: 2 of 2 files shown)


patches/server/0889-Allow-change-origin-for-EnderDragon.patch
```diff
@@ -0,0 +1,80 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEnderDragon.java
+@@ -79,4 +79,16 @@ public class CraftEnderDragon extends CraftMob implements EnderDragon {
+     public int getDeathAnimationTicks() {
+         return this.getHandle().dragonDeathTime;
+     }
++
++    // Paper start
++    @Override
++    public org.bukkit.Location getOrigin() {
++        return null;
```
Owen1212055: You need to call the NMS methods in here.

Doc94: i keep this for the final and forget add when commit, xd

Added logic

Doc94:
Ok i make a few changes because i forget the logic about `Heightmap.Types` then now this PR allow change the position of "Podium" and in base of this the rest of NMS use this for apply the logic for the differents `Heightmap.Types` for keep the logic from vanilla about the relative position.

patches/server/0889-Allow-change-podium-for-EnderDragon.patch
```diff
@@ -0,0 +1,153 @@
...
++    // Paper start
++    @Override
++    public org.bukkit.Location getPodium() {
++        net.minecraft.core.BlockPos blockPosOrigin = this.getHandle().getPodium();
++        return new org.bukkit.Location(getWorld(), blockPosOrigin.getX(), blockPosOrigin.getY(), blockPosOrigin.getZ());
++    }
++
++    @Override
++    public void setPodium(org.bukkit.Location location) {
++        if (location != null) {
```
kennytv: Wrong way around?

patches/server/0889-Allow-change-podium-for-EnderDragon.patch
```diff
@@ -0,0 +1,153 @@
...
++            this.getHandle().setPodium(null);
++        } else {
++            if (location.getWorld() != null) {
```
kennytv: Validate.isTrue(location.getWorld() == nul || ...

patches/api/0377-Allow-change-podium-for-EnderDragon.patch
```diff
@@ -0,0 +1,33 @@
...
++    /**
++     * Get the location considered by the dragon how podium.
++     *
++     * @return the podium location of the dragon
++     */
++    @NotNull
++    org.bukkit.Location getPodium();
++
++    /**
++     * Sets the location of podium for the dragon.
```
kennytv:
```suggestion
+     * Sets the location of the podium for the ender dragon.
```

patches/api/0377-Allow-change-podium-for-EnderDragon.patch
```diff
@@ -0,0 +1,33 @@
...
++++ b/src/main/java/org/bukkit/entity/EnderDragon.java
+@@ -104,4 +104,22 @@ public interface EnderDragon extends ComplexLivingEntity, Boss, Mob {
+      * @return this dragon's death animation ticks
+      */
+     int getDeathAnimationTicks();
++
++    // Paper start
++
++    /**
++     * Get the location considered by the dragon how podium.
```
kennytv:
```suggestion
+     * Get the podium location used by the ender dragon.
```

patches/api/0377-Allow-change-podium-for-EnderDragon.patch
```diff
@@ -0,0 +1,33 @@
...
++     *
++     * @return the podium location of the dragon
++     */
++    @NotNull
++    org.bukkit.Location getPodium();
++
++    /**
++     * Sets the location of podium for the dragon.
++     *
++     * @param location the location of the podium or null for use the default podium (Exit Portal of the END)
```
kennytv:
```suggestion
+     * @param location the location of the podium or null to use the default podium location (exit portal of the end)
```

kennytv (changes requested):

patches/server/0889-Allow-change-podium-for-EnderDragon.patch
Doc94: im pretty sure i use ==.. maybe a confusion when WSL still not opening popups :/ fixed

kennytv (approved):
For future reference; don't merge master, either rebase, or just don't until asked to rebase, else we can do it on merging

PR #7701, opened 2022-04-05
Add Alternate Current's redstone implementation as an alternative to vanilla and Eigencraft's.

SpaceWalkerRS:
(Re-submitted #7694 from a non-master branch)

This PR adds [Alternate Current](https://github.com/SpaceWalkerRS/alternate-current)'s redstone implementation as an alternative to vanilla and Eigencraft's. A previous attempt at this was made by [Titaniumtown](https://github.com/Titaniumtown) (#7291), but as they moved on, I decided to have a crack at it.

Alternate Current is many times faster than vanilla, and even exceeds Eigencraft. I will do some testing in the coming days and update this post with some numbers.

Just like Eigencraft, Alternate Current changes redstone dust's update order. This means any redstone contraption that is location dependent in vanilla has a significant chance of breaking with either of the faster implementations. Beyond that, parity issues should be rare, though Alternate Current has not been as thoroughly tested as Eigencraft, so I cannot compare the two in that aspect.

### Performance

The world I used can be downloaded [here](https://www.mediafire.com/file/94xf0s80w1k7dg3/world%25282%2529.zip/file) for anyone wishing to replicate these tests.

Idle load: 0.7 mspt

[Grids](https://imgur.com/DvIH6fm)
- Vanilla: 292.6 mspt
- Eigencraft: 54.0 mspt
- Alternate Current: 14.0 mspt

[Lines](https://imgur.com/Uf5FiIH)
- Vanilla: 203.8 mspt
- Eigencraft: 63.1 mspt
- Alternate Current: 17.9 mspt

The second test is probably closest of the two to a 'real-world' scenario.

### DIFF
--- a/patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
+++ b/patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
@@ -26,19 +26,17 @@ to check if any block is a potential power source.
 The use-faster-eigencraft-redstone config has been replaced with the redstone-implementation config, which can be used to
 switch between the vanilla, Eigencraft, and Alternate Current implementations.
 
-diff --git a/src/main/java/alternate/current/wire/LevelAccess.java b/src/main/java/alternate/current/wire/LevelAccess.java
+diff --git a/src/main/java/alternate/current/wire/LevelHelper.java b/src/main/java/alternate/current/wire/LevelHelper.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..4be46b13ebb1b7e7a41a1853a16a70f58e4fcbe2
+index 0000000000000000000000000000000000000000..186e0a86ceaed007d394378157ce1596a0601908
 --- /dev/null
-+++ b/src/main/java/alternate/current/wire/LevelAccess.java
-@@ -0,0 +1,126 @@
++++ b/src/main/java/alternate/current/wire/LevelHelper.java
+@@ -0,0 +1,64 @@
 +package alternate.current.wire;
 +
 +import net.minecraft.core.BlockPos;
-+import net.minecraft.core.Direction;
 +import net.minecraft.server.level.ServerLevel;
 +import net.minecraft.world.level.block.Block;
-+import net.minecraft.world.level.block.Blocks;
 +import net.minecraft.world.level.block.state.BlockState;
 +import net.minecraft.world.level.chunk.ChunkAccess;
 +import net.minecraft.world.level.chunk.ChunkStatus;
@@ -46,19 +44,9 @@ index 0000000000000000000000000000000000000000..4be46b13ebb1b7e7a41a1853a16a70f5
 +
 +import org.bukkit.event.block.BlockRedstoneEvent;
 +
-+public class LevelAccess {
++public class LevelHelper {
 +
-+    private final ServerLevel level;
-+
-+    LevelAccess(ServerLevel level) {
-+        this.level = level;
-+    }
-+
-+    BlockState getBlockState(BlockPos pos) {
-+        return level.getBlockState(pos);
-+    }
-+
-+    int doRedstoneEvent(BlockPos pos, int prevPower, int newPower) {
++    static int doRedstoneEvent(ServerLevel level, BlockPos pos, int prevPower, int newPower) {
 +        BlockRedstoneEvent event = new BlockRedstoneEvent(level.getWorld().getBlockAt(pos.getX(), pos.getY(), pos.getZ()), prevPower, newPower);
 +        level.getCraftServer().getPluginManager().callEvent(event);
 +
@@ -70,11 +58,7 @@ index 0000000000000000000000000000000000000000..4be46b13ebb1b7e7a41a1853a16a70f5
 +     * update redstone wire block states, lighting checks, height map updates, and
 +     * block entity updates are omitted.
 +     */
-+    boolean setWireState(BlockPos pos, BlockState state, boolean updateNeighborShapes) {
-+        if (!state.is(Blocks.REDSTONE_WIRE)) {
-+            return false;
-+        }
-+
++    static boolean setWireState(ServerLevel level, BlockPos pos, BlockState state, boolean updateNeighborShapes) {
 +        int y = pos.getY();
 +
 +        if (y < level.getMinBuildHeight() || y >= level.getMaxBuildHeight()) {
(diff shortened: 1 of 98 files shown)


minhh2792:
This comment is unrelated but can I get the world download of the 2 images, I want to do some testing too

justmangoou:
> This comment is unrelated but can I get the world download of the 2 images, I want to do some testing too

Just build a floor and then use the structure block

minhh2792:
The server will crash when I built this 
Here is the crash report https://pastebin.com/LLe26aur

image

SpaceWalkerRS:
> The server will crash when I built this Here is the crash report https://pastebin.com/LLe26aur
> 
> image

Thanks for testing! I pushed a fix, and also added a world download for the testing world I used to the original post.

minhh2792:
> > The server will crash when I built this Here is the crash report https://pastebin.com/LLe26aur
> > image
> 
> Thanks for testing! I pushed a fix, and also added a world download for the testing world I used to the original post.

Thanks for the hard work 👍

molor:
Can some1 test, will this also fix https://github.com/PaperMC/Paper/issues/7725?

SpaceWalkerRS:
> Can some1 test, will this also fix #7725?

I will expand more on this on the original issue, but it is not a bug with Paper/Eigencraft. It is normal Vanilla behavior, as a consequence of quasi-connectivity, the range in which blocks are updated around redstone wire, and the location dependent order in which blocks are updated around redstone wire. This can result in pistons remaining 'stuck' extended or retracted in certain setups.

The effect that Eigencraft has (and Alternate Current will have, too) is that this behavior is no longer location dependent. Instead, each setup will behave the same in every location it is built.

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++		currentPower = level.doRedstoneEvent(pos, currentPower, power);
++		state = state.setValue(RedStoneWireBlock.POWER, currentPower);
++
++		return level.setWireState(pos, state, added);
++	}
++}
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index c9c08e20f729b40b1d90f3ac144f5245f4f35230..33adaac9449862879724da25f0c138709a536dbe 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
++++ b/src/main/java/com/destroystokyo/paper/PaperConfig.java
```
kennytv: The version goes into the original PaperConfig patch

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++		if (!state.isAir() && !state.is(Blocks.REDSTONE_WIRE)) {
++			level.updateNeighborShape(pos, state, fromDir, fromPos, fromState);
++		}
++	}
++
++	/**
++	 * Emit block updates around the given wire. The order in which neighbors are
++	 * updated is determined as follows:
++	 * <br>
++	 * 1. The direction of power flow through the wire is to be considered'forward'.
```
kennytv:
```suggestion
+	 * 1. The direction of power flow through the wire is to be considered 'forward'.
```

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++	private final WireConnection[] heads;
++
++	private WireConnection head;
++	private WireConnection tail;
++
++	/** The total number of connections. */
++	int total;
++
++	/**
++	 * A 4 bit number that encodes which in direction(s) the owner has connections
```
kennytv:
```suggestion
+	 * A 4 bit number that encodes in which direction(s) the owner has connections
```

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++import net.minecraft.world.level.block.Block;
++import net.minecraft.world.level.block.Blocks;
++import net.minecraft.world.level.block.state.BlockState;
++import net.minecraft.world.level.chunk.ChunkAccess;
++import net.minecraft.world.level.chunk.ChunkStatus;
++import net.minecraft.world.level.chunk.LevelChunkSection;
++
++import org.bukkit.event.block.BlockRedstoneEvent;
++
++public class LevelAccess {
```
kennytv: Everything in this class but `doRedstoneEvent` and `setWireState` seem like they should just be called directly instead of the wrapper methods (either by keeping the ServerLevel instead of a LevelAccess, or with a `level()` getter) of the same exact name/a two-liner that is only used once or twice.

The remaining two could possibly also be moved into a utility class with static methods or to where they are actually used

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++            } else {
++                implementation = "vanilla";
++            }
+         } else {
+-            log("Using vanilla redstone algorithm.");
++            implementation = this.getString("redstone-implementation", "vanilla").toLowerCase().trim();
++        }
++        switch (implementation) {
++            default:
++                logError("Warning: Invalid redstone-implementation config " + implementation + " - must be one of: vanilla, eigencraft, alternate-current");
```
kennytv:
```suggestion
+                logError("Invalid redstone-implementation config " + implementation + " - must be one of: vanilla, eigencraft, alternate-current");
```

patches/server/0890-Add-Alternate-Current-redstone-implementation.patch
```diff
@@ -0,0 +1,2474 @@
...
++            case "vanilla":
++                redstoneImplementation = RedstoneImplementation.VANILLA;
++                log("Using the vanilla redstone implementation.");
```
kennytv:
```suggestion
+                log("Using the Vanilla redstone implementation.");
```

kennytv (changes requested):
If it's not intrusive to you updating this patch in the future, please replace tabs with 4 spaces in the newly added classes in this patch

kennytv:
... that was another code review without actual tests yet; please also create a pr to [our docs repository](https://github.com/PaperMC/docs/blob/main/docs/paper/reference/paper-per-world-configuration.md) explaining that non-Vanilla redstone methods may disappear between Minecraft updates, or permanently if issues are found, and any differences in behavior of AC compared to Vanilla's redstone implementation (or just a link to your documentation, if that exists)

kennytv (approved):
Tested a whole bunch of contraptions and code smells have been cleaned up, so seems good to me. Waiting on more opinions tho

kennytv:
Is there anything else you would like to add or change here? Otherwise this will most likely be rebased and merged at some point in the coming days

MiniDigger (approved):
:shipit:

NoahvdAa (approved):
Looks good from web (and from Cub's test server) :shipit:

SpaceWalkerRS:
> Is there anything else you would like to add or change here?

I found an issue during testing the other day, so I'm gonna push a small update to fix that.

kennytv:
Thank you for the pr! If anything major comes up, we'll forward things to you, otherwise we'll be looking forward to people using the option and future updates of the mod/patch (‾◡◝)

PR #7703, opened 2022-04-05
Prevent healing by negative amounts

booky10:
This happens on healing potion amplifiers 29, 61, 93 and 125 because of bit shifting. Negative potion amplifiers are not included, as they are automatically being removed.

A check is made for drinkable potions, but apparently not for throwable ones.


MiniDigger:
is there a mojira for this?

booky10:
[MC-179180](https://bugs.mojang.com/browse/MC-179180) and [MC-30540](https://bugs.mojang.com/browse/30540) are covering this, but both are marked as a duplicate of [MC-10755](https://bugs.mojang.com/browse/MC-10755), which is marked as "Won't Fix". [MC-10755](https://bugs.mojang.com/browse/MC-10755) is also a more general issue for potion effects and enchantments.

Doc94:
Maybe this can be in https://github.com/PaperMC/Paper/blob/master/patches/server/0827-Load-effect-amplifiers-greater-than-127-correctly.patch ?

booky10:
I don't think so. Loading effects and applying healing effects are two seperate things.

electronicboy (approved):

electronicboy:
provisionally approved, but, not 100% on this
the mojira issue should probably be noted in the patch file, but there's generally 0 commitment towards ensuring that stuff doesn't go sideways when doing illegal stuff

Owen1212055:
It should be noted that the behavior of this has been used for a very long time and can be considered "instant death" potions. And looking at the fact that this can really only be achieved in creative mode, I would honestly just use a plugin to prevent these types of potions from being added instead.

booky10:
I have added the "instant death" potion note and MOJIRA issues to the patch description.

Machine-Maker:
Even tho mojang has marked this as won't fix, paper has already fixed an effect amplifier issue that would be marked as won't fix on mojira. @Owen1212055 can't you have instant death potions using the actual instant harm potions? Or do these potions provide a function that can't be replicated any other way.

> I don't think so. Loading effects and applying healing effects are two seperate things.

In this case, they are actually. they both result of incorrect handling of certain effect amplifier values. I think they can go in the same patch, the patch can just be renamed to something more general.

booky10:
> Or do these potions provide a function that can't be replicated any other way.

Heal potions don't check if the player is invulnerable, because they are supposed to heal.
Harming potions check if the player is invulnerable and then do nothing.

booky10:
Rebased for 1.19

electronicboy:
I'm still not sold on this, my understanding is that the bug this is trying to fix has, and is being used, for the unintended behavior in the real world, for items which aren't attainable without intentionally creating "invalid" items?

booky10:
Yes, "invalid" items are the most common usecase of this bug. I still think paper could fix this, because the underlying issue is just an inconsistency between drinking a potion and using a splash potion.

electronicboy:
this doesn't abait the concern that to some map makers, this is a feature

booky10:
I could add a configuration option for map makers, but I don't think any map maker uses this "feature". Most don't set the player in creative mode (where players are immune to harming potions), and even there you could still use `/kill` to kill/respawn them.

dlee13:
We're not map makers, but players use these potions on each other in my creative server because it's funny. What benefit is there for fixing the bug?

Owen1212055:
I mean paper kinda doesn’t really add patches for creative items that create buggy behavior like this. I would just make your own filter or something that properly prevents these from happening by using a plugin.

booky10:
> I mean paper kinda doesn’t really add patches for creative items that create buggy behavior like this. I would just make your own filter or something that properly prevents these from happening by using a plugin.

Yes, you can do that pretty easily by cancelling the `EntityRegainHealthEvent` if the amount is negative, but that's not the point.

If this would just be a fix for the instant-kill-potions, I wouldn't have opened this PR. I have opened this PR to fix an inconsistency between the handling of drinking a heal potion and splashing someone else with a heal potion. If you don't like this and still think, this is something which doesn't fit Paper, then close the PR.

electronicboy:
Closing, while this is fixing an inconsistency, it's only detectable through the use of items using unsupported attributes (in which has already been written off by mojang as something they don't care to invest time into). If there was no real usage of this behavior, I could much easily support pulling this patch, but, as is, this would need a config, and given the unsupported attributes, I'm just not particularly compelled by this change

PR #7704, opened 2022-04-05
Add EntityPushedByEntityAttackEvent

aerulion:
(Continuation of #7058 from non-master branch)

Since the EntityKnockbackByEntityEvent can't be reused on non-living entities due to being implemented with the use of the LivingEntity interface, this PR adds a new EntityPushedByEntityAttackEvent that does get called on pushes caused by an entity attack. Closes issue #5886.

This is my first time working with patches, so I have no idea if I did it correctly. Please tell me if I did something wrong, and I will try to fix it as fast as I can.


kennytv (dismissed):

kennytv:
Hm looking into it, I think you should remove the collision ones (plus the push reason) and make the EntityKnockbackEvent extend this new event, also considering vehicles already have collision events.

(Since I already force pushed a rebase to your branch, feel free to do the changes and force push over them with your local changes)

aerulion:
That sounds like a far better solution, should this still be a new patch or should I modify the existing one for the EntityKnockbackEvent?

kennytv:
Probably best in the original, you can then add yourself as a co-author there

aerulion:
I hope I did everything correctly and covered all the requested changes. I went ahead and renamed the event slightly in order to make it clearer that it only covers non-collision pushes.

aerulion:
Any update on this?

electronicboy (approved):

aerulion:
Rebased and updated for 1.19 (Hopefully correctly? I've never rebased patches before...)

Owen1212055 (approved):
Tested, and looks good.

I updated this to include the new entity attacks, and also added a bit of documentation as this can sometimes be fired multiple times (not this event, but the previously made event).

Owen1212055:
Thanks for your contribution, and sorry for the wait. 😄

PR #7706, opened 2022-04-06
Use patched spigot decompiler

jpenilla:
https://github.com/PaperMC/patched-spigot-fernflower

Brings `spigotDecompileJar` from >1min to 17-22 seconds on my 3800x

We will want to check for decompiler changes whenever pulling BuildData updates going forward.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -48,10 +48,17 @@ subprojects {
     }
 }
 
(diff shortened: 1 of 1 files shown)


kennytv (approved):
Pre-approving for when the jar is published/removed from the file tree here

PR #7709, opened 2022-04-07
List world on tile entity placement exception

Astralchroma:
An exception is called whenever there is an attempt to place a tile entity where there is something else there, such as this:
```
[03:29:22 WARN]: com.destroystokyo.paper.exception.ServerInternalException: Attempted to place a tile entity (net.minecraft.world.level.block.entity.TileEntityCampfire@135dea23) at 6876,131,5939 (Block{minecraft:red_sand}) where there was no entity tile!
[03:29:22 WARN]: Chunk coordinates: 6864,5936
```
This minor change will cause the world to be listed when that exception occurs, this is helpful for people who have dozens of worlds on one server, and want to know where the issue occurred without checking every single one.

### DIFF
--- a/patches/server/0055-Add-exception-reporting-event.patch
+++ b/patches/server/0055-Add-exception-reporting-event.patch
@@ -143,7 +143,7 @@ index 2d290d2e17517bc39828e05369ca5bc5ce892c27..fee21a585b95448a5edab70002e9c4ea
                                  }
  
 diff --git a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
-index ed6203119fe2e4ee47a2d51c84df5b7c236f32da..108c7ec010f826f9b9bb2db0987d2bcc18266482 100644
+index ed6203119fe2e4ee47a2d51c84df5b7c236f32da..7a63c26e360fd054bf237df3eeffc466d73d5dfb 100644
 --- a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
 +++ b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
 @@ -1,6 +1,7 @@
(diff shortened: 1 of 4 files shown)


patches/server/0055-Add-exception-reporting-event.patch
```diff
@@ -167,14 +167,15 @@ index ed6203119fe2e4ee47a2d51c84df5b7c236f32da..108c7ec010f826f9b9bb2db0987d2bcc
 +                "Attempted to place a tile entity (" + blockEntity + ") at " + blockEntity.getBlockPos().getX() + ","
 +                    + blockEntity.getBlockPos().getY() + "," + blockEntity.getBlockPos().getZ()
 +                    + " (" + getBlockState(blockposition) + ") where there was no entity tile!\n" +
-+                    "Chunk coordinates: " + (this.chunkPos.x * 16) + "," + (this.chunkPos.z * 16));
++                    "Chunk coordinates: " + (this.chunkPos.x * 16) + "," + (this.chunkPos.z * 16) +
++                    "\nWorld: " + this.level.getWorld().getName());
```
electronicboy: Should probably use the key and not the name

electronicboy (changes requested):

patches/server/0055-Add-exception-reporting-event.patch
Astralchroma: Do you mean `level.serverLevelData.getLevelName()`?

Astralchroma: Might be `level.getWorld().getKey()`...

Machine-Maker: Do `this.level.dimension().location()`

Astralchroma: Done

Astralchroma:
*is this gonna end up going stale?*

kennytv:
Not after 3 days, please don't bump prs unnecessarily

Astralchroma:
At what point does a PR go stale anyway?

NoahvdAa:
After sixty days.

electronicboy (approved):

Machine-Maker (approved):

Astralchroma:
Rebased, at least I think I did it correctly. Please check to make sure I did not screw up anything.

Astralchroma:
Rebased again

PR #5867, opened 2021-06-18
Banner custom name API

Swedz:

PR #7711, opened 2022-04-07
Nameable Banner API

Owen1212055:
Supercedes https://github.com/PaperMC/Paper/pull/5867


patches/server/0890-Nameable-Banner-API.patch
```diff
@@ -0,0 +1,36 @@
...
++
++    @Override
++    public String getCustomName() {
++        return io.papermc.paper.adventure.PaperAdventure.LEGACY_SECTION_UXRC.serializeOrNull(this.customName());
++    }
++
++    @Override
++    public void setCustomName(String name) {
++       this.customName(io.papermc.paper.adventure.PaperAdventure.LEGACY_SECTION_UXRC.deserializeOrNull(name));
++    }
```
Machine-Maker: How do we feel about adding new deprecated methods? I know its on the interface, but is there a better way to do this? Like should we make those methods on the Nameable interface be defaults that delegate to the component methods with the right serializer?

Owen1212055: I mean id say we can just make these default methods in the adventure patch yeah.

Owen1212055:
Unstale

Machine-Maker (approved):

electronicboy (approved):

Owen1212055:
Rebased, tested and updated

PR #7713, opened 2022-04-08
Fix tick function tag running before load

etil2jz:
When enabling or reloading a datapack that has functions for both tick and load, the former runs before the latter, which causes problems in some cases such as initializing scoreboard values before increasing them per tick.
See https://bugs.mojang.com/browse/MC-187539

### DIFF
--- a/patches/server/0891-Fix-tick-function-tag-running-before-load.patch
+++ b/patches/server/0891-Fix-tick-function-tag-running-before-load.patch
@@ -4,7 +4,7 @@ Date: Fri, 8 Apr 2022 21:28:40 +0200
 Subject: [PATCH] Fix tick function tag running before load
 
 When enabling or reloading a datapack that has functions for both
-#tick and #load, the former runs before the latter, which causes
+tick and load, the former runs before the latter, which causes
 problems in some cases such as initializing scoreboard values before
 increasing them per tick.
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Ugh, I'm a little conflicted. Like I see its marked as a bug, but doesn't this introduce more inconsistency with data pack behavior from vanilla? I almost want to say that this shouldn't be fixed before Vanilla does since data packs are developed with Vanilla in mind. According to that mojira issue, this has been around for a while despite being a confirmed bug.

Perhaps this should have a config option? But I so don't like having config options for bug fixes.

etil2jz:
This seems to fix more inconsistencies than it can add. I believe performing the load check before the tick function is what was intended to be. Having a config option is a possibility, but as you said it's *a lot* for a single bugfix.

Machine-Maker:
> This seems to fix more inconsistencies than it can add.

I'm talking about inconsistencies in behavior between a data pack running on a vanilla server and one on a paper server. Ideally data packs should perform the same on both, but this would change the order of those functions on paper servers where its still "broken" on vanilla servers. So I'm not sure how it fixes more inconsistencies than it adds.

jpenilla (changes requested):
I agree with Machine, this should not be fixed before vanilla.

PR #7714, opened 2022-04-08
Improvements in PlayerReadyArrowEvent

Doc94:
This PR aims a few things like..

- Call the event when check OFF_HAND (Currently if player has a arrow in OFF_HAND the event its not called) and avoid double call in base to this
- Edit docs for mention the case for CrossBow

### DIFF
--- a/patches/api/0113-PlayerReadyArrowEvent-and-PlayerReadyProjectileEvent.patch
+++ b/patches/api/0113-PlayerReadyArrowEvent-and-PlayerReadyProjectileEvent.patch
@@ -153,6 +153,8 @@ index 0000000000000000000000000000000000000000..0b7e02609900578d6c8462d7a1f2f325
 +        this.cancelled = cancel;
 +    }
 +
++    @Override
++    @NotNull
 +    public HandlerList getHandlers() {
 +        return handlers;
 +    }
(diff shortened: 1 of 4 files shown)


Doc94:
Currently the side effect for this is for the event been called when a Crossbow try to load a firework...
If necesary maybe can be good in the future rename the event to make sense to been called for any projectile weapon with "munition"

Machine-Maker:
I think instead of expanding this event to cover those cases, a new event should be created that PlayerReadyArrowEvent extends. Like PlayerReadyProjectileEvent or something. That way PlayerReadyArrowEvent can be deprecated and no functionality changes. It could just be expanded to cover crossbows, but like you said with the fireworks, I think the new event approach addresses all those concerns with the added plus of not changing any existing behavior.

Doc94:
> I think instead of expanding this event to cover those cases, a new event should be created that PlayerReadyArrowEvent extends. Like PlayerReadyProjectileEvent or something. That way PlayerReadyArrowEvent can be deprecated and no functionality changes. It could just be expanded to cover crossbows, but like you said with the fireworks, I think the new event approach addresses all those concerns with the added plus of not changing any existing behavior.

Yeah.. for this i have a questions...
- I can edit the same patch for deprecate the event and add the new if so its not issue if rename the patch?
- im not sure how can keep the two behavior or only with the change of deprecate and make the new event solve all this?

Machine-Maker:
To keep behavior, you just move all the code including the HandlerList to the new event from PlayreArrowReadEvent, and then make PlayerArrowReadyEvent extend that new event. That should keep all behavior. Then you can deprecate the old event. 

And yeah, this can all be done in that patch, and rename the patch to Add-Whatever-Event

Doc94:
> To keep behavior, you just move all the code including the HandlerList to the new event from PlayreArrowReadEvent, and then make PlayerArrowReadyEvent extend that new event. That should keep all behavior. Then you can deprecate the old event.
> 
> And yeah, this can all be done in that patch, and rename the patch to Add-Whatever-Event

Okay, i remake all the patch based in the comments.

Doc94:
EY EY EY... Fuera de aqui... i mean this still work :c

Issue #7707, opened 2022-04-07
Add a way to change what drops from an Arrow entity


u9g:
### Is your feature request related to a problem?

I'd like to make custom arrows, so I need to be able to change what the item is when picking them up.

### Describe the solution you'd like.

```java
((Arrow) entity).setItemStack()
```

### Describe alternatives you've considered.

I have literally no idea

Xemorr:
Couldn't you use this event: https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerPickupArrowEvent.html
and then give the player the item you want to give them.

u9g:
> Couldn't you use this event: https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerPickupArrowEvent.html
> and then give the player the item you want to give them.

Yes I realized this later, but then you have to add a way to identify the arrow that the player threw so you can listen for that same arrow later, rather than just adding a setter so you can just easily add what the arrow drops when spawning the arrow.

Xemorr:
Yeah it wouldn't be too hard to track you can just use PersistentData but you're right that would be the more elegant solution if it's just not being exposed by the API.

Machine-Maker:
@u9g can you listen to PlayerPickupArrowEvent and change the itemstack there? it just seems odd to add a whole separate field for pickup item.

PR #7715, opened 2022-04-08
Allow for setting the Pickup Item of an Arrow

tl0x:
The feature proposed in issue [#7707](https://github.com/PaperMC/Paper/issues/7707).


HexedHero:
Should this extend to other projectiles that can be picked up like Tridents and any future ones?

patches/api/0378-allow-setting-of-arrowpickup.patch
```diff
@@ -0,0 +1,30 @@
...
+
+diff --git a/src/main/java/org/bukkit/entity/Arrow.java b/src/main/java/org/bukkit/entity/Arrow.java
+index 8814519df13c5e9475774f4bce4c8c6a462c2978..7f80a6d1a0f1f2d1889062a1f451e311bdc92e1a 100644
+--- a/src/main/java/org/bukkit/entity/Arrow.java
++++ b/src/main/java/org/bukkit/entity/Arrow.java
+@@ -2,6 +2,7 @@ package org.bukkit.entity;
+ 
+ import java.util.List;
+ import org.bukkit.Color;
++import org.bukkit.inventory.ItemStack;
```
Machine-Maker, line 15: Inline this import

patches/api/0378-allow-setting-of-arrowpickup.patch
```diff
@@ -0,0 +1,30 @@
...
+@@ -94,4 +95,11 @@ public interface Arrow extends AbstractArrow {
+      * in a state with no Custom Effects and PotionType.UNCRAFTABLE
+      */
+     void clearCustomEffects();
++
++    /**
++     * Sets the ItemStack given when the arrow is picked up.
++     *
++     * @param itemStack
++     */
```
Machine-Maker, line 28: I would make some mention that this isn't persisted across entity unload/loads. aka its not saved.

patches/server/0891-allow-setting-of-arrowpickup.patch
```diff
@@ -0,0 +1,63 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java b/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java
+index 53d0024daf6963ac4dab575666b0d6a74a39a958..00fd2849e4dd31b506645dbb9bd7128a453ac638 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java
+@@ -603,6 +603,8 @@ public abstract class AbstractArrow extends Projectile {
+ 
+     public abstract ItemStack getPickupItem();
+ 
++    public abstract void setPickupItem(ItemStack stack);
```
Machine-Maker, line 15: I think the field should be here on AbstractArrow and then you can have a pickup item override for all things that extend AbstractArrow.

Machine-Maker (changes requested):
You've got two server patches. You can do `git commit --amend` in `Paper-Server` to add changes to the last commit. 


Does this compile? I don't see the setPickupItem method implemented on ThrownTrident.

patches/api/0378-allow-setting-of-arrowpickup.patch
u9g: why not save the itemstack as bytes in it’s pdc?

PR #7719, opened 2022-04-09
Correct javadocs on PlayerArmorStandManipulateEvent

Machine-Maker:
In this case, the javadocs are wrong, and the nullability annotations correct. The term `null` shouldn't be used here, but `air` instead. At first I thought `empty` would be best, but since there isn't actually an `isEmpty` method on ItemStack, I figured saying air would make more sense in the context of the API


Owen1212055 (approved):

Machine-Maker:
Rebased for 1.19

electronicboy (approved):

PR #7558, opened 2022-03-07
Properly create profiles with custom name/uuid

lynxplay:
Ensure that the server properly creates custom player profiles if either
the uuid or name match an online player. Previously the server would
directly shortcut to the players profile if either their uuid or name
matched, ignoring differences in the respective alternative identifier
like a custom name or a different uuid.

To keep the previous functionality, the method does populate the profile
with the online players properties if either the uuid or name match.

Resolves: #7557

PR #7723, opened 2022-04-10
Revert to old createProfile(UUID, String) logic

### LINKED ISSUE #7700: New profile creation logic breaks the plugins
### Expected behavior

Recently merged [pull request](https://github.com/PaperMC/Paper/pull/7558) that alters the profile creation has broken the UUID generation in offline mode, for example in WorldGuard. I couldn't investigate it much as WorldGuard is quite a big plugin, but I guess the problem is that it tries to get the UUID from the lowercase nickname - before the fix it was the same as the real one, now it differs.

I agree that it definitely shouldn't be so and this is a plugin issue, I [reported](https://github.com/EngineHub/WorldGuard/issues/1896) it to WorldGuard, but my previous issue was closed just for being related to offline mode, so it's very likely the maintainer will close this one too.

Hence, maybe some compatibility workaround should be implemented?

### Observed/Actual behavior

Absolutely different offline UUIDs are generated for some plugins

### Steps/models to reproduce

Using Paper >= 263

### Plugin and Datapack List

`Plugins (17): AuthMe, Chunky, CoreProtect, Essentials, EssentialsChat, EssentialsSpawn, Harbor, LuckPerms, MoneyFromMobs, NoCheatPlus, ProtocolLib, SafeTrade, SkinsRestorer, Vault, WorldBorder, WorldEdit, WorldGuard`

`There are no more data packs available`

### Paper version

```
This server is running Paper version git-Paper-274 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 04c7b16)
You are 3 version(s) behind
Previous version: git-Paper-271 (MC: 1.18.2)
```

### Other

_No response_

lynxplay:
The recent change to the createProfile method set, which allowed plugins
to properly create profiles with differing names/uuids for players
currently on the server broke an implicit method contract for plugins
providing the lowercased version of a player name to the method.

Previously, the returned profile would simply be pulled from an online
player with the same name (when compared case-insensitively), however
the previously mentioned fix to the creation logic currently returns
a profile with the lowercased name instead.

To not break plugins that relied on this specific internal logic, the
ability to create pre-filled profiles with different names or uuids for
players that are currently online is moved by this commit to a new
method called createProfileExact, which simply mirrors the current
createProfile logic.

The createProfile logic is hence returned to its old functionality, once
again allowing plugins to grab an online players profile from their
lowercased name.

See: #7558
Resolves: #7700

-------------------------------------------------------------------------

Potential points of discussions:

This PR actively brings back (in my opinion) unwanted behaviour to the `createProfile` method. While I see the use of the logic in said method, I am uncertain whether moving the (in my opinion) correct logic to a new method like `createProfileExact` is the way to go here or if we should move the old logic into a new method. Both sides have their benefits, however the patch currently introduces a new method as not break plugins that relied on the lowercase hack.

### DIFF
--- a/patches/api/0058-Basic-PlayerProfile-API.patch
+++ b/patches/api/0058-Basic-PlayerProfile-API.patch
@@ -289,10 +289,10 @@ index 0000000000000000000000000000000000000000..7b3b6ef533d32169fbeca389bd61cfc6
 +    }
 +}
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index ece84330d2700db8708d2ae2ab7badf4acb428a8..621420d35378e0038c33892c185216894912f023 100644
+index ece84330d2700db8708d2ae2ab7badf4acb428a8..90c875ad60e473c3ec25f209933d48615d0fd6c0 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
-@@ -2092,6 +2092,50 @@ public final class Bukkit {
+@@ -2092,6 +2092,83 @@ public final class Bukkit {
      public static boolean suggestPlayerNamesWhenNullTabCompletions() {
          return server.suggestPlayerNamesWhenNullTabCompletions();
      }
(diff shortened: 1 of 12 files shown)


electronicboy (approved):

molor:
Merge that please, I and my server team also affected by the lowercase issue :c

r4nx:
So is anyone against? Me and, I guess, other people who liked the PR would be thankful if it could be merged

And thanks for your work guys, I'm very surprised that you not only not refused the issue, but provided the fix. That deserves respect.

electronicboy:
the PR needs rebasing

lynxplay:
Rebased onto latest 👍

PR #7724, opened 2022-04-10
Add pre-unbreaking amount to PlayerItemDamageEvent

HexedHero:
Adds the amount of damage that the item was going to take before any Unbreaking enchantment has rolled its reduction.
It's currently not possible to tell how much damage an item was going to take before its Unbreaking enchantment, this also allows options to replace and/or simulate your own unbreaking system.

Tested with a sword breaking Grass Blocks:

[07:40:02 WARN]: Damage | OG Damage: 2 | 2
[07:40:08 WARN]: Damage | OG Damage: 2 | 2
[07:40:17 INFO]: [HexedHero: Applied enchantment Unbreaking III to HexedHero's item]
[07:40:19 WARN]: Damage | OG Damage: 2 | 2
[07:40:22 WARN]: Damage | OG Damage: 0 | 2
[07:40:28 WARN]: Damage | OG Damage: 1 | 2
[07:40:30 WARN]: Damage | OG Damage: 0 | 2


patches/api/0378-Add-pre-unbreaking-amount-to-PlayerItemDamageEvent.patch
```diff
@@ -0,0 +1,52 @@
...
+     }
+ 
+     /**
+@@ -42,6 +51,19 @@ public class PlayerItemDamageEvent extends PlayerEvent implements Cancellable {
+         return damage;
+     }
+ 
++    // Paper start - Add pre-reduction damage
++    /**
++     * Gets the amount of durability damage this item would of been taking
```
NoahvdAa:
```suggestion
+     * Gets the amount of durability damage this item would have taken
```

NoahvdAa (approved):

kennytv (approved):

patches/api/0378-Add-pre-unbreaking-amount-to-PlayerItemDamageEvent.patch
HexedHero: Thanks

Machine-Maker (commented):
I feel like "original" isn't the right descriptor here? Maybe "base"? or "unmodified". Original seems weird because it implies before external changes, and I kinda consider unbreaking, since its a vanilla thing, is part of original.

HexedHero:
I like unmodified damage, should I change it to that?
**Edit:** That being said, it seems Bukkit uses getOriginalDamage https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/entity/EntityDamageEvent.html#getOriginalDamage(org.bukkit.event.entity.EntityDamageEvent.DamageModifier) so I feel like the naming may be in-line with Bukkit?

HexedHero:
@Machine-Maker Let me know what I should change it to, I'm fine with whatever I just want this merged thanks

patches/api/0378-Add-pre-unbreaking-amount-to-PlayerItemDamageEvent.patch
```diff
@@ -0,0 +1,52 @@
...
+     private static final HandlerList handlers = new HandlerList();
+     private final ItemStack item;
+     private int damage;
++    private int originalDamage; // Paper - Add pre-reduction damage
+     private boolean cancelled = false;
+ 
++    @Deprecated // Paper - Add pre-reduction damage
+     public PlayerItemDamageEvent(@NotNull Player player, @NotNull ItemStack what, int damage) {
++        // Paper start - Add pre-reduction damage
++        this(player, what, damage, damage);
```
Machine-Maker, line 21: is it safe to do this `damage, damage` thing? This might be an event where just replacing the ctor is OK. idk if there are plugins that call this event manually. `damage, damage` is probably fine.

Machine-Maker (approved):

PR #7727, opened 2022-04-10
Fix cancelling PlayerDropItemEvent for carried items

### LINKED ISSUE #7726: Possible item loss when cancelling PlayerDropItemEvent
### Expected behavior

Assuming a player stands inside a pile of dropped items, then starts moving an item in his inventory with his cursor and thereby filling the empty slot with any of the dropped items. If the player now clicks outside his inventory to drop the cursor item, the event should get cancelled and the item should remain attached to the cursor.

### Observed/Actual behavior

When clicking outside the inventory the item will simple get deleted if the players inventory is full.

### Steps/models to reproduce

1. Cancel all PlayerDropItemEvents.
2. Stand in a pile of dropped items with a full inventory.
3. Move an item in your inventory with your cursor.
4. Click outside the inventory in order to drop the item.
5. Watch the item being deleted.

### Plugin and Datapack List

Just a test plugin cancelling the event.

### Paper version

This server is running Paper version git-Paper-283 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: f8e8d6c)
Previous version: git-Paper-277 (MC: 1.18.2)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7726

Now if the PlayerDropItemEvent is cancelled and the drop is trigged by clicking outside a menu, the item remains on the cursor instead of going back into the inventory or being deleted if the inventory is full.

### DIFF
--- a/patches/api/0416-Fix-cancelling-PlayerDropItemEvent-for-carried-items.patch
+++ b/patches/api/0416-Fix-cancelling-PlayerDropItemEvent-for-carried-items.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 15 Apr 2023 17:11:01 -0700
+Subject: [PATCH] Fix cancelling PlayerDropItemEvent for carried items
+
+
+diff --git a/src/main/java/org/bukkit/event/player/PlayerDropItemEvent.java b/src/main/java/org/bukkit/event/player/PlayerDropItemEvent.java
+index aca500f32e2ae69095a1abce9f14a164df442638..6eb9c12d225a9bea0d2c79dce62670f5e2438416 100644
+--- a/src/main/java/org/bukkit/event/player/PlayerDropItemEvent.java
++++ b/src/main/java/org/bukkit/event/player/PlayerDropItemEvent.java
+@@ -8,11 +8,15 @@ import org.jetbrains.annotations.NotNull;
+ 
+ /**
+  * Thrown when a player drops an item from their inventory
++ * <p>
++ * Look at the documentation of {@link #setOverflowConsumer(java.util.function.Consumer)} for important
++ * information regarding cancelling of this event.
+  */
+ public class PlayerDropItemEvent extends PlayerEvent implements Cancellable {
+     private static final HandlerList handlers = new HandlerList();
+     private final Item drop;
+     private boolean cancel = false;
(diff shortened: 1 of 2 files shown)


kennytv (dismissed):

Machine-Maker:
There needs to be a consensus on what happens if the player closes the inventory with the item in their cursor and the drop item event is cancelled (and the inventory is full). Thoughts on what should happen?

Owen1212055 (changes requested):
https://user-images.githubusercontent.com/23108066/180033453-ddefdbe1-82aa-4f0e-a654-dd6332b91d85.mp4
Items outside of guis still seem to disappear.
(cancelling the event)

It also seems that now dropping items outside of your inventory just causes them to disappear.

https://user-images.githubusercontent.com/23108066/180038817-bd0bb20c-95fa-4ca2-8a9c-b8139ecedc17.mp4

(Swapping hands toggles between cancelling the event)

Machine-Maker:
Yeah, I made some changes cause I spotted a bug in what I was doing before, but I didn't test those. I guess I broke it more.

Machine-Maker:
Ok, fixed that issue. Also I changed what the cursor is set to if the event is cancelled. It gets the itemstack from the item drop. So if an event handler changes the Item entity and then cancels the event it uses that itemstack in the item entity. Is that right? I don't know if we do that elsewhere, take an action based on the event if its cancelled.

Machine-Maker:
I like this idea. But what should the default behavior be? Just delete the item?

Machine-Maker:
Ok, added that overflow consumer. Tested it as well, setting the consumer to drop the itemstack in the world anyways. By default, it just sets the count of the stack to 0 to "delete" the itemstack.

patches/api/0416-Fix-cancelling-PlayerDropItemEvent-for-carried-items.patch
```diff
@@ -0,0 +1,50 @@
...
+ 
+     public PlayerDropItemEvent(@NotNull final Player player, @NotNull final Item drop) {
+         super(player);
+@@ -29,6 +33,23 @@ public class PlayerDropItemEvent extends PlayerEvent implements Cancellable {
+         return drop;
+     }
+ 
++    /**
++     * In certain cases, when this is event is cancelled via {@link #setCancelled(boolean)}, the
++     * server is unable to handle cancelling the event. An example of this might be canceling the
```
aerulion, line 33: In the first sentence cancelling is spelled with 2 L, in the second only with 1

bridgelol:
Any chance this can get merged at some point? Adding an overflow consumer seems like the best option imo.

PR #7731, opened 2022-04-12
Use Mojang code to validate usernames

MagicalSheep:
Paper will validate the characters of usernames when players join the server but offer a configuration to disable the feature. However, Mojang will do the validation for usernames no matter whether the server is in the online mode since 1.18.2.

For this reason, there is no need for Paper to do the same thing again. Considering many Paper servers, especially in China or Japan, have used non-ASCII characters as usernames by disabling the validation Paper offers, it is better to keep the configuration that can disable the validation of usernames.

This commit replaces the Paper's validating usernames logic as Mojang's, and it will work only when the server is in the online mode (Just like how it worked before 1.18.2).

### DIFF
--- a/patches/server/0832-Validate-usernames.patch
+++ b/patches/server/0832-Validate-usernames.patch
@@ -22,7 +22,7 @@ index 24ddf8cfdbe6ed2fb148f57f0d7dd98446b07bbc..da6346cacf08e12f7f1fabe2d5b1c66c
      public static int maxPlayerAutoSavePerTick = 10;
      private static void playerAutoSaveRate() {
 diff --git a/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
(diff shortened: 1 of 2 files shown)


electronicboy:
Keeping our own validation logic was 100% intentional as mojangs choices there allows a lot more than standard characters in any language
maybe make it so that disabling the paper one uses the vanilla one which is less secure but more open? idk, but, defo not fond of the laxer set of checks for a setup we don't even support

MagicalSheep:
> Keeping our own validation logic was 100% intentional as mojangs choices there allows a lot more than standard characters in any language maybe make it so that disabling the paper one uses the vanilla one which is less secure but more open? idk, but, defo not fond of the laxer set of checks for a setup we don't even support

Thanks for your work! It's ok to keep the Paper's logic of validating, that's not the main problem. 

The main problem is that Mojang will do the validation before Paper but Paper will validate usernames only when the online mode is enabled.

It caused:

1. Paper's configuration `unsupported-settings.perform-username-validation` becomes invalid. No matter whether Paper does the validation, Mojang will do it.
2. Usernames are validated twice when the server is in the online mode.

I think maybe we can remove the changes Mojang has made in 1.18.2 and just keep the Paper’s validation and configuration as before?

PR #7734, opened 2022-04-14
Don't send itemmeta for books when hide-itemmeta is enabled

NoahvdAa:
Expands `hide-itemmeta` to include unneeded information from [Written Books](https://minecraft.fandom.com/wiki/Written_Book#Item_data).


Machine-Maker (approved):

Machine-Maker:
Just a quick question, does the client not having the title or author or text change the appearance of the icon? Just wondering if maybe the glint relied on that somehow.

NoahvdAa:
Nope, a writeable book and a written book are a separate item, `WrittenBookItem#isFoil` always returns true.

kennytv (approved):

PR #7743, opened 2022-04-16
Add Event Trigger Event

Nuckerr:
Adds a new event that is called when an event triggers (Obviously excepting the Event Trigger Event). It allows you to access all the registration information like the plugin, listener and priority as well as being able to cancel the event. (Forgive any mess ups I've made, this is my first pr)


electronicboy:
This is a no go
doubling the number of events fired is far from a sane idea, removing the ability for plugins to deal with events is just downright dangerous

PR #7752, opened 2022-04-19
fix: EndDragonFight killed statuses should be false for newly created worlds

### LINKED ISSUE #7750: EnderDragonBattle#hasBeenPreviouslyKilled() returns true for a new created world
### Expected behavior

When creating a new world the `World#getEnderDragonBattle().hasBeenPreviouslyKilled()` should return false.

### Observed/Actual behavior

When world is created it first returns true.
After a player go in the world and the ender dragon has spawned it returns false.
After the death of the dragon it returns true.

### Steps/models to reproduce

I used a simple plugin to output the result :

```
@Override
public boolean onCommand(@NotNull CommandSender sender, @NotNull Command command, @NotNull String label, @NotNull String[] args) {
        sender.sendMessage(Component.text("DEBUG").color(NamedTextColor.RED));
        for (World world : Bukkit.getServer().getWorlds())
            if(world.getEnvironment().equals(World.Environment.THE_END))
                sender.sendMessage(Component.text("hasBeenPreviouslyKilled() ? " + (world.getEnderDragonBattle().hasBeenPreviouslyKilled())));

        return true;
    }
```

Steps :

- Create a new server
- Check hasBeenPreviouslyKilled() output
- Go to end
- Check hasBeenPreviouslyKilled() output again

https://user-images.githubusercontent.com/54907162/163856964-755e3423-053a-4fb2-bb4a-802eba14b8d1.mp4

### Plugin and Datapack List

image

### Paper version

image

### Other

The issue occurs too when I use another plugin taht create worlds with the `Bukkit#createWorld(WorldCreator)`.

Don't know if it can help but after going into the Paper code it seems the `EndDragonFight(ServerLevel, long, CompoundTag)` set by default the previouslyKilled field to true if the given CompoundTag doesn't contains "DragonKilled" key which could be the case for a new created world.

elmital:
Fixes #7750

### DIFF
--- a/patches/server/0008-MC-Utils.patch
+++ b/patches/server/0008-MC-Utils.patch
@@ -5914,7 +5914,7 @@ index 6c98676827ceb6999f340fa2b06a0b3e1cb4cae2..f08089b8672454acf8c2309e850466b3
  
          while (objectiterator.hasNext()) {
 diff --git a/src/main/java/net/minecraft/server/level/ServerChunkCache.java b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
-index ce88976db29b9e9524dbe45b16721ef90afb692b..96323a83d917516edf4b6951d1e881d2f5513bd0 100644
+index ce88976db29b9e9524dbe45b16721ef90afb692b..1a8c5ce3ecce9cbbc8496ea3882b18c297964e33 100644
 --- a/src/main/java/net/minecraft/server/level/ServerChunkCache.java
 +++ b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
 @@ -50,6 +50,7 @@ import net.minecraft.world.level.storage.LevelStorageSource;
@@ -5925,7 +5925,7 @@ index ce88976db29b9e9524dbe45b16721ef90afb692b..96323a83d917516edf4b6951d1e881d2
      public static final List<ChunkStatus> CHUNK_STATUSES = ChunkStatus.getStatusList();
      private final DistanceManager distanceManager;
      final ServerLevel level;
-@@ -68,6 +69,151 @@ public class ServerChunkCache extends ChunkSource {
+@@ -68,6 +69,231 @@ public class ServerChunkCache extends ChunkSource {
      @Nullable
      @VisibleForDebug
      private NaturalSpawner.SpawnState lastSpawnState;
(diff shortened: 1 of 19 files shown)


patches/server/0896-fix-EndDragonFight-previouslyKilled-should-be-false-.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+index ed6ff35a2112cea4faa6e1458d1effc40fce5dd2..cddd14d8642de7c53c68ccc792d4775abb1c24b6 100644
+--- a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
++++ b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+@@ -113,7 +113,7 @@ public class EndDragonFight {
+             }
+         } else {
+             this.dragonKilled = true;
+-            this.previouslyKilled = true;
++            this.previouslyKilled = nbt.contains("PreviouslyKilled") && nbt.getBoolean("PreviouslyKilled");// Paper - Should be false for newly created worlds
```
Doc94:
```suggestion
+            this.previouslyKilled = nbt.contains("PreviouslyKilled") && nbt.getBoolean("PreviouslyKilled"); // Paper - Should be false for newly created worlds
```

patches/server/0896-fix-EndDragonFight-previouslyKilled-should-be-false-.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+index ed6ff35a2112cea4faa6e1458d1effc40fce5dd2..cddd14d8642de7c53c68ccc792d4775abb1c24b6 100644
+--- a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
++++ b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+@@ -113,7 +113,7 @@ public class EndDragonFight {
+             }
+         } else {
+             this.dragonKilled = true;
+-            this.previouslyKilled = true;
++            this.previouslyKilled = nbt.contains("PreviouslyKilled") && nbt.getBoolean("PreviouslyKilled"); // Paper - Should be false for newly created worlds
```
Owen1212055: Any reason why you can't just make this false? It looks the only way this could happen is if the entire nbt compound is empty. However since needsStateScanning should be true, it'll cause the dragon to correctly be detected and reset to the correct values.

Owen1212055 (changes requested):

patches/server/0896-fix-EndDragonFight-previouslyKilled-should-be-false-.patch
elmital: I figured out the same as you but in the case I've missed a possibility that the nbt contains this key or if it could be for any other future/current reason it's handled. So for me it's just the safer way but can modify if well confirmed that is not needed.

Owen1212055: I guess for future proofing. But I mean, the only way for this to happen currently is if you manually remove it from the nbt.

elmital: Just waiting for a Paper team feedback(thx for yours) and I'll apply the necessary changes and a rebase if it's all good.

elmital: Reconsidering what you said I've made changes after more check and test with stateScanning option I changed the dragonKilled field too. I hope I didn't screw up with git, I know 1.19 is the priority but at least it's done ;)

elmital:
Rebased for 1.19
~~edit: just saw I fucked up with comment~~ (fixed)

elmital:
Rebased for 1.19.1

PR #7754, opened 2022-04-21
Configurable map update interval

minhh2792:
Make the map update faster, so everytime a player moving or change direction it will update instantly


electronicboy:
This should modify the original patch instead of adding on another

The reason for these changes was that the data wouldn't send anyways and thus calling it felt redundant to call logic to calculate data, which was particularly expensive in some cases, if the data isn't going to be used anyways

I don't think that breaking vanilla behavior here is the answer, this should probably be exposed in API better so that plugins can determine what they wanna do vs having wider implications for something which hasn't been cared about in years

minhh2792:
I believed that this patch is important to some plugin developer which utilizes the map, having it update faster and smoother is better, so I made the value configurable via world config now (default is 5 ticks)

One thing I notice is that on 1.12.2 it didn't have this problem, you can watch this video (based on Paper 1.12.2) and see how smooth the map is when player rotating their head https://www.youtube.com/watch?v=i_0NrTb2Rsc&t=52s

electronicboy:
No need to bump config version for adding in a property, but, as I say, I'd much prefer there to be a way to expose this via the API rather than a config option which people aren't gonna touch unless prodded

PR #7759, opened 2022-04-23
Set send-namespaced to false by default

HexedHero:
Who uses this? I feel like 95% of servers have this disabled. You're way more likely to go turn it off vs going to turn it on.


MiniDigger:
the whole point of the namespaced stuff is to be able to escape conflicts, so idk how I feel about this

HexedHero:
You can still use the commands, just isn't sent in the completion to make it not look like a mess

Machine-Maker (changes requested):
I don't think this should be changed. I feel like its just going to increase the amount of "I can't use /give with nbt" questions.

kennytv:
I don't like it either; it's easy enough to just toggle if you need it

PR #7763, opened 2022-04-24
Add onReload() to Plugin to make /reload safe to use

turtley12:
The reload command has been unsafe in several ways since it was introduced. This PR modifies the command so it can safely be used for it's intended purpose

This PR changes the functionality of reload so that instead of disabling and re-enabling plugins, the reload command will call `Plugin#onReload()`. All other functionality such as reloading server configuration has been kept.

`onReload()` is used the same was as `onEnable()` and the other similar methods in `Plugin`.

Running the Reload command now displays a modified warning message. Instead of informing the user that the command is no longer supported, it says "Please note that while this command is no longer unsafe, it may not be supported by some plugins". 

For plugins that have overrided `onReload()`, It will run their `onReload()` methods. This can be used instead or with plugin specific reload commands. Using WorldEdit as an example, WorldEdit has a `/worldedit reload` command that will reload the worldedit configuration. with this PR, the same configuration reload code could be called in `onReload()`. This will enable users to have a easy way to reload all configurations and/or other things, without any issues with things breaking like the old reload command had.

This was implemented as a method in `Plugin` instead of an event because I felt that it was similar to `onLoad()`. `onLoad()` might be able to be implemented as an event, but because it was implemented as a method in `Plugin`, I went the same way and put `onReload()` as a method in `Plugin`.

### DIFF
--- a/patches/api/0381-Add-onReload-to-make-reload-safe.patch
+++ b/patches/api/0381-Add-onReload-to-make-reload-safe.patch
@@ -1,0 +1,84 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Turtley12 <76929530+Turtley12@users.noreply.github.com>
+Date: Sat, 23 Apr 2022 17:09:28 -0700
+Subject: [PATCH] Add onReload() to make /reload safe
+
+
+diff --git a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+index 0c7ba0718de2b93d013968ca0fec34ffd423990f..aef971d0fe565fd57089f318aed0f3503c9cb8fa 100644
+--- a/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
++++ b/src/main/java/org/bukkit/command/defaults/ReloadCommand.java
+@@ -13,7 +13,7 @@ public class ReloadCommand extends BukkitCommand {
+     public ReloadCommand(@NotNull String name) {
+         super(name);
+         this.description = "Reloads the server configuration and plugins";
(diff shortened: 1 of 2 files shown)


electronicboy:
Changing the contract of the reload command is basically a non starter
without any of the wider changes needed for the long term future proofing
of plugin management, this doesn't really solve anything, maybe if it was a
separate command, but, i.e. sreload, but, any plugin which supports
reloading it's config safely already has its own command

turtley12:
I feel like the reload command's plugin reload functionality should either be replaced with something like this or removed entirely.

If everyone says not to use a feature, than why does the feature exist?

The idea with this was to have an easy way to reload any plugin that supported reloading it's config, along with the server configuration, all at the same time. This is what I thought Bukkit's reload command was originally supposed to do, just without unloading running code and the issues with plugins assuming the server was starting/stopping

What are some of the wider changes that this could go with?

What do you think about just removing everything plugin related from the reload command?

electronicboy:
> If everyone says not to use a feature, than why does the feature exist?

Because the feature was added many years ago and has its own benefits to existing, e.g. developers; reloading stuff is generally safe but can have side effects, not everything can be reloaded, and there is room for stuff to go wrong, hence why this is deemed as unsupported, for the most part it generally works, just, nobody wants to deal with the cases where it doesn't

> What are some of the wider changes that this could go with?

vanillas reload command reloads a good chunk of data, data which will one day be populatable by plugins using the API, hopefully; this entire reload cycle would need to be exposed. Ideally, we'd generally intermix reloading of plugins into the vanilla pathways of reloading the server and its datapacks through a proper plugin lifecycle

> What do you think about just removing everything plugin related from the reload command?

it's a feature which has existed for years and if we want the server to keep compat with vanilla, will likely need to stay; as much as we dislike the reload command, it exists and has done for years, and there's generally little argument for getting rid of it outside of the fact that we don't provide support for servers using it

Issue #7662, opened 2022-03-25
Color Formatting in Console broken


Chris6ix:
### Expected behavior

Colors look normal?

### Observed/Actual behavior

Sometimes the color formatting just doesnt work at all and so the color codes just stay infont of the message like this:
![Screenshot 2022-03-23 231354](https://user-images.githubusercontent.com/65610536/160198475-49f8a2fe-4a2c-4333-aeff-525cab4a3e94.png)
Some of the colored text gets underlined (although the item/prefix doesn't have underline in name)...here are some examples:

![Screenshot 2022-03-23 233137](https://user-images.githubusercontent.com/65610536/160198988-71b1e7a3-eab0-4c4a-9d89-64904e61c64d.png)
![Screenshot 2022-03-23 233159](https://user-images.githubusercontent.com/65610536/160198995-8ab80782-47ca-486f-9b64-35ee5ced9241.png)
Sometimes this underlining looks very weird:
![Screenshot 2022-03-22 235323](https://user-images.githubusercontent.com/65610536/160199068-5b5618af-b880-448f-af43-9d516ebb4de4.png)

### Steps/models to reproduce

I'm not really sure...maybe:
-Kick a player from the server?
-Assign yourself to a group (Using Luckperms) and write something in chat?
-Give yourself an item with a custom name and kill somebody with it?

### Plugin and Datapack List

Plugins (49): ArmoredElytra, BackToTheRoots*, BetterConcrete, BetterTridents, BigDoors, Bookshelf, BreakingSlab, ClientDetector, CommandCooldown, DeluxeMenus, Duels, DynamicShop, dynmap, Essentials, EssentialsChat, ExecutableItems, FlyPerms, GSit, HolographicDisplays, HoloMobHealth, HomeGUI, InteractionVisualizer, JoinRules, KittenForever, LightAPI, LinkedPortals, LinksPlus, LuckPerms, Multiverse-Core, Multiverse-Inventories, NotTooExpensive, PL-Hide, PlaceholderAPI, PlayerPortal, PlayerWarps, Playtimes*, ProtocolLib, QuickShop, RotationWrench, SCore, SilkSpawners, TAB, UltimateCatcher, UltimateTimber, UltraCosmetics, Vault, WorldEdit, WorldEditSelectionVisualizer, WorldGuard

There are 3 data packs enabled: [vanilla (built-in)], [file/bukkit (world)], [file/Chris6ix_Custom_Server_datapack.zip (world)]
[...]

Issue #7590, opened 2022-03-13
Color problem in console for Bukkit.broadcast()


x9nico:
### Expected behavior

It should set in colors my broadcasts. There, it don't want to set the colors in console

### Observed/Actual behavior

![image](https://user-images.githubusercontent.com/8091015/158067019-62145747-a52b-4e70-921f-f64acfab71a2.png)
![image](https://user-images.githubusercontent.com/8091015/158067052-90676e5b-b80c-4385-acaf-8c4684d76de2.png)

### Steps/models to reproduce

Make a broadcast in your code, like this:
`Bukkit.broadcast(Component.text(PvPBoxPlugin.getInstance().getConfig().getString("reset_map_announce").replace("&", "§").replace("{time}", Integer.toString(time))));
`

### Plugin and Datapack List

![image](https://user-images.githubusercontent.com/8091015/158067078-8003f95a-b248-4044-bdaf-bdf3a2cbead9.png)
PvPBox is a custom plugin

### Paper version

[16:30:54 INFO]: This server is running Paper version git-Paper-247 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 1358d1e)
You are running the latest version
Previous version: git-Paper-232 (MC: 1.18.2)

### Other

It does that bug from I updated to build 247, it don't affect the 232.

_No response_

NoahvdAa:
You should not mix adventure components and legacy formatting (§). Either deserialize your legacy messages using the [LegacyComponentSerializer](https://jd.adventure.kyori.net/text-serializer-legacy/4.9.1/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializer.html) (or use minimessage instead of legacy), or use the legacy method, [Bukkit#broadcastMessage](https://papermc.io/javadocs/paper/1.18/org/bukkit/Bukkit.html#broadcastMessage(java.lang.String)). This will cause the proper behaviour, as shown below:

```java
Bukkit.broadcast(Component.text("§cE")); // WRONG
Bukkit.broadcast(Component.text("E", NamedTextColor.RED));
Bukkit.broadcastMessage("§cE");
```
results in (e's without § are colored)
```
[17:11:36 INFO]: §cE
[17:11:36 INFO]: E
[17:11:36 INFO]: E
```

x9nico:
[...]

PR #7767, opened 2022-04-25
Restore deprecated legacy color codes in console with config

jamespglines:
I know the reason for remove this. It cause issue and confuse for some messages in console and legacy color is deprecated and legacy. Many plugin use this. It is hard to remove at random time with no warning (ik color are deprecated. there is no direct warning about log).

In kyori I hear about "ansi component serializer". I think maybe this should wait until ansi serializer exists.. By restore with config this provide the better deprecation period and notice. small brown out.

I default to true. I do this because if it is not, many plugin will auto set to true. This is not good idea. I put deprecated in the config name to be clear for users who read.

Thank you for consideration. This is my first PR. I hope I read the contrib guidelines correctly 😅

This provide temp fix for #7590, #7560, #7662. To kangaro: this is not made because of you. do not put me on memecademy. let ur ego suffer.

### DIFF
--- a/patches/server/0902-Restore-deprecated-legacy-color-codes-in-console-wit.patch
+++ b/patches/server/0902-Restore-deprecated-legacy-color-codes-in-console-wit.patch
@@ -1,0 +1,131 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: "James P. Glines" <jamespglines@gmail.com>
+Date: Mon, 25 Apr 2022 14:01:09 -0500
+Subject: [PATCH] Restore deprecated legacy color codes in console with config
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/PaperConfig.java b/src/main/java/com/destroystokyo/paper/PaperConfig.java
+index c9c08e20f729b40b1d90f3ac144f5245f4f35230..10284eecc7b452b2a7f366e7195aacc9cd796388 100644
+--- a/src/main/java/com/destroystokyo/paper/PaperConfig.java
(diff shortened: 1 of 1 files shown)


zml2008:
There's already a way to log colored text to console -- with the ConsoleCommandSender.

An ANSI serializer is purely an implementation detail -- the real solution for full-featured `Component` logging is via a Component-supporting SLF4J wrapper, see KyoriPowered/adventure#731. This is a relatively simple feature to implement, so I see no reason to implement workarounds rather than the correct solution

jamespglines:
Yes. you can do this. but change is made so quickly and there is NOT good alternative of logging component to console yet. this is what i mean by ansi serializer. sorry for bad sentences.

This even break paper dev own plugin EssentialsX!

jamespglines:
Why is this closed? If this workaround is not to be considered please say. this is ok. I want to know if my code is bad or if workaround never will exist.

jpenilla:
This workaround will not be considered, however if it was to be considered, then the issue would be with the poor implementation.

electronicboy:
This workaround mangles much more than it looks like it should need to and just looks to create its own set of issues in the process

I'm somewhat iffy around adding extra legacy handling for something we've advised against for years, I'm not *hard* against this, but, most instances of colour on the console are either generally abuse of the logger which can create issues, or stuff like chat, in which people should generally be using the console sender as vanilla does rather than using the logger

jamespglines:
This is understood. Thank you for your time.

zml2008:
Here's a more long-term solution to this issue: https://github.com/KyoriPowered/adventure/pull/757

PR #7773, opened 2022-04-27
Undeprecate getOfflinePlayer

HexedHero:


electronicboy:
wasn't there a discouraged annotation, at the very least we should notate that this can perform blocking I/O on the callers thread

Machine-Maker:
Is it even discouraged? There isn’t a replacement right? If I want an offline player instance by the string name, how do I get it?

electronicboy:
Well, right now it's discouraged through a deprecation, I'm in favour of removing that but I do believe that there should be some form of notice about the I/O, I thought that there was a discouraged annotation that would minimally warn you about this in the IDE

HexedHero:
The Javadoc already says
"_This method may involve a blocking web request to get the UUID for the given name._"
Not sure if that is a good enough warning.

Owen1212055:
Look at something like (which isn’t deprecated) Bukkit#getOfflinePlayers()… just a little javadoc despite a LARGE amount of IO.

electronicboy:
Well, I was thinking that IDEs would minimally hint on discouraged, if one existed, as I kiiinda believed it did but don't remember 100%
less invasive vs deprecated but, given 

> 2022
> reading

idk

electronicboy (approved):

PR #7777, opened 2022-04-30
cleanup filtered sign text

electronicboy:
Paper was at some point writing data inside of the filtered text,
which in the past was not an issue, however, it appears that mojang
has flipped something which is now causing this information to be
relevant and thus causing interesting effects on the client.

This patch will clean up said data for existing signs, given that
there is realistically 0 chance that nobody was even using this data,
this should be a none issue for old signs

### DIFF
--- a/patches/server/0902-Sign-cleanup-filtering.patch
+++ b/patches/server/0902-Sign-cleanup-filtering.patch
@@ -1,0 +1,38 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Shane Freeder <theboyetronic@gmail.com>
+Date: Sat, 30 Apr 2022 00:38:57 +0100
+Subject: [PATCH] Sign cleanup filtering
+
+TODO: This logic here should be moved into a DataFixer in the next MC release
+noting to ensure to apply the cleanup logic here, and clean up the added
+tags
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/SignBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/SignBlockEntity.java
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
Testing this fix with ``/setblock ~ ~ ~ oak_sign{Text1:'{"text":"Hello! :)"}',FilteredText1:'[{"text":""}]'} replace`` shows that this works.

However, I feel that when a new datafixer version is added that this patch should be moved to something that DFU should fix instead (like in 1.19). This would allow older signs to be updated correctly, and us not needing to upkeep this tag for a long while.

kennytv (approved):

NoahvdAa (approved):
Can confirm this works too :)
The only way I could get this to break was by placing signs with NBT that had the incorrect filteredtext set; `/give @s oak_sign{BlockEntityTag:{Text1:'{"text":"Hello! :)"}',FilteredText1:'[{"text":""}]'}}`
This is pretty much a non-issue though, since I highly doubt that (m)any of these broken signs still exist.

lordofpipes:
For anyone needing a quick and dirty fix, I've written this open source plugin to auto-fix sign NBT data. Currently being used in production on three big servers so I assume the method used is sane https://github.com/UltraVanilla/cleanupstupidmojangshit

Once this pull request is merged, however, it will be obsolete and you can simply update Paper instead.

kennytv (changes requested):
Actually, I don't *really* like this, since we'd basically have to keep this forever and keep breaking things with this; at the very least this should ~~check the world/data version to be <= 1.18.2, *possibly* add a toggle for this (god knows this might actually break few people's things already)~~ be removed later and have it replaced by a datafixer

kennytv (approved):
a

PR #7792, opened 2022-05-05
Hide ores with deepslate at Y=4 and below (#7771)

### LINKED ISSUE #7771: Anti-xray engine mode 1 is not effective below y 0
### Expected behavior

In engine mod 1, below y level 0, ores are replaced by deepslate.

### Observed/Actual behavior

In engine mod 1, below y level 0, ores are replaced by stone.

If you use xray to find a stone below y level 0, it is an ore.

### Steps/models to reproduce

https://docs.papermc.io/paper/anti-xray#engine-mode-1

Use engine 1 and observe the stone using xray below y level 0.

### Plugin and Datapack List

none

### Paper version

This server is running Paper version git-Paper-312 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: ce059b4)

### Other

_No response_

NoahvdAa:
Fixes #7771 by hiding ores at Y=4 and below with deepslate instead of stone. Y=4 was chosen because stone transitions into deepslate from Y=8 until Y=0. Hiding from 8 or 0 would make ores stand out, since there'd be an out of place stone/deepslate cube, so 4 seemed like a good average.

ccing @stonar96

<details>
<summary>Before & after comparison</summary>

**Before:**

**After:**

</details>

### DIFF
--- a/patches/server/0355-Anti-Xray.patch
+++ b/patches/server/0355-Anti-Xray.patch
@@ -258,10 +258,10 @@ index 0000000000000000000000000000000000000000..aabad39d13ead83042ec2e4dd7f4ed49
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java b/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..5fa7a7a3584d7078aff260184d9834f34296322c
+index 0000000000000000000000000000000000000000..8e2baa21f71e7105a8e72cba4ded6aa99370fca0
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java
-@@ -0,0 +1,663 @@
+@@ -0,0 +1,666 @@
 +package com.destroystokyo.paper.antixray;
 +
 +import com.destroystokyo.paper.PaperWorldConfig;
@@ -307,6 +307,7 @@ index 0000000000000000000000000000000000000000..5fa7a7a3584d7078aff260184d9834f3
 +    private final BlockState[] presetBlockStatesEndStone;
 +    private final int[] presetBlockStateBitsGlobal;
 +    private final int[] presetBlockStateBitsStoneGlobal;
++    private final int[] presetBlockStateBitsDeepslateGlobal;
 +    private final int[] presetBlockStateBitsNetherrackGlobal;
 +    private final int[] presetBlockStateBitsEndStoneGlobal;
 +    private final boolean[] solidGlobal = new boolean[Block.BLOCK_STATE_REGISTRY.size()];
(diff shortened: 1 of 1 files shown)


patches/server/0355-Anti-Xray.patch
```diff
@@ -407,7 +410,7 @@ index 0000000000000000000000000000000000000000..3916d99c746d952e7ae6c3dde750d08f
 +                return switch (level.getWorld().getEnvironment()) {
 +                    case NETHER -> presetBlockStatesNetherrack;
 +                    case THE_END -> presetBlockStatesEndStone;
-+                    default -> presetBlockStatesStone;
++                    default -> bottomBlockY <= 4 ? presetBlockStatesDeepslate : presetBlockStatesStone; // Deepslate starts mixing with stone from y=8 and below. At y=0 all stone is deepslate, for an average of y=4.
```
stonar96: This won't work as you might think. This method is called per chunk section. `bottomBlockY` can only be `..., -16, 0, 16, ...`. (That's why I said in #paper-contrib that only contexts per chunk section are feasible. Everything beyond that would be far more complex. E.g. you'd have to register deepslate and stone in the palette of the y=0 chunk section and then you'd still have to check for `<= 4` in the obfuscation logic. We can avoid all this by just using `< 0` here.)

You can also see what's going wrong with `<= 4` by doing the same thing as in your before & after comparison. Just configure it to show deepslate instead of stone and you will see that deepslate is used up to y=15 and not 4.

patches/server/0355-Anti-Xray.patch
```diff
@@ -302,6 +302,7 @@ index 0000000000000000000000000000000000000000..3916d99c746d952e7ae6c3dde750d08f
 +    private final BlockState[] presetBlockStates;
 +    private final BlockState[] presetBlockStatesFull;
 +    private final BlockState[] presetBlockStatesStone;
++    private final BlockState[] presetBlockStatesDeepslate;
 +    private final BlockState[] presetBlockStatesNetherrack;
 +    private final BlockState[] presetBlockStatesEndStone;
 +    private final int[] presetBlockStateBitsGlobal;
```
stonar96, line 308: Unfortunately github doesn't allow to comment outside of the diff. Read the lines below this comment. This must also be implemented for deepslate, otherwise it will fail for chunk sections that use the global palette (this case is handled separately because it can be optimized).

stonar96 (changes requested):

Pantera07:
https://github.com/NoahvdAa/Paper/pull/1

NoahvdAa:
Thanks for letting me know, I'm not watching that repo so I didn't get a notification.

stonar96 (approved):

patches/server/0355-Anti-Xray.patch
```diff
@@ -472,7 +478,7 @@ index 0000000000000000000000000000000000000000..3916d99c746d952e7ae6c3dde750d08f
 +        LevelChunkSection[] nearbyChunkSections = new LevelChunkSection[4];
 +        LevelChunk chunk = chunkPacketInfoAntiXray.getChunk();
 +        Level level = chunk.getLevel();
-+        int maxChunkSectionIndex = Math.min((maxBlockHeight >> 4) - chunk.getMinSection(), chunk.getSectionsCount() - 1);
++        int maxChunkSectionIndex = Math.min((maxBlockHeight >> 4) - chunk.getMinSection(), chunk.getSectionsCount()) - 1;
```
kennytv, line 481: my brain went a little smooth here but looks fine

kennytv (approved):

patches/server/0355-Anti-Xray.patch
stonar96: See https://github.com/NoahvdAa/Paper/pull/1

Pantera07:
This PR works fine.

Pantera07 (approved):

PR #7798, opened 2022-05-08
Fix mobs attacking themselves

etil2jz:
If an entity is provoked by a second one using commands, the second will join in the fight against itself, causing it to attack itself repeatedly.
See [MC-110386](https://bugs.mojang.com/browse/MC-110386).


Owen1212055 (approved):
Tested and looks good

kennytv:
We generally only want to maintain patches that fix issues people can legitimately run into, not in some self-constructed, edge-case scenarios

PedroMPagani:
That would most likely never happen in a default env, the mob attacking the other is obviously induced.

PR #7799, opened 2022-05-08
Fix brewing stands resetting their brewTime when being unloaded

etil2jz:
When a brewing stand is unloaded, such as saving and quitting in a world, the cycle is reset to the start. This includes consuming another blaze powder.
See [MC-26304](https://bugs.mojang.com/browse/MC-26304).

### DIFF
--- a/patches/server/0907-Fix-brewing-stands-resetting-their-brewTime-when-bei.patch
+++ b/patches/server/0907-Fix-brewing-stands-resetting-their-brewTime-when-bei.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: etil2jz <blanchot.arthur@protonmail.ch>
+Date: Sun, 8 May 2022 16:51:32 +0200
+Subject: [PATCH] Fix brewing stands resetting their brewTime when being
+ unloaded
+
+When a brewing stand is unloaded, such as saving and quitting in a world,
+the cycle is reset to the start. This includes consuming another blaze powder.
+See https://bugs.mojang.com/browse/MC-26304.
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
+index 5c0f1488c8a8100cd39a03adeccded9984722249..ef99e202ffad483f71702d931767cbde3cab93b0 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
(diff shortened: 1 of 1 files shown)


NoahvdAa (changes requested):
NBT tags added by Paper should start with `Paper.`

patches/server/0907-Fix-brewing-stands-resetting-their-brewTime-when-bei.patch
```diff
@@ -0,0 +1,30 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
+index 5c0f1488c8a8100cd39a03adeccded9984722249..ef99e202ffad483f71702d931767cbde3cab93b0 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java
+@@ -296,6 +296,7 @@ public class BrewingStandBlockEntity extends BaseContainerBlockEntity implements
+         ContainerHelper.loadAllItems(nbt, this.items);
+         this.brewTime = nbt.getShort("BrewTime");
+         this.fuel = nbt.getByte("Fuel");
++        this.ingredient = net.minecraft.core.Registry.ITEM.get(new net.minecraft.resources.ResourceLocation(nbt.getString("Paper.ingredient"))); // Paper - fix brewing stands resetting their brewTime when being unloaded
```
Owen1212055, line 19: May be wrong here, but do you need to check if this nbt contains this nbt tag in the first place? Cause if it doesn't exist won't it throw a null pointer or something?

Owen1212055 (changes requested):

patches/server/0907-Fix-brewing-stands-resetting-their-brewTime-when-bei.patch
NoahvdAa: iirc `Registry#get` should allow a null argument and will return null when that is passed in

Owen1212055: Well it's more so the resource location, as it has logic for parsing and might not like it.

NoahvdAa: Ah, my bad, I didn't see the resourcelocation and thought it was just passed into `Registry#get`. This will definitely throw an NPE when null is passed then.

etil2jz: The ResourceLocation parsing catches exception and, in case of a null argument, return null;

NoahvdAa: The constructor (or the methods therein) doesn't appear to do any error handling, maybe you mean `ResourceLocation#tryParse`?

PR #7801, opened 2022-05-08
ItemStack damage API

### LINKED ISSUE #7755: Add ItemStack#damage

Currently, the only way of simulating damage to an ItemStack is by
- Calling the PlayerItemDamageEvent, passing the damage you want to do
- Check if the PlayerItemDamageEvent was cancelled and continue if not
- Adding the processed damage from the PlayerItemDamageEvent to ItemStack
- Check if the ItemStack's current damage is equal to or great than the ItemStack's max durability and continue if so
- Set the ItemStack to AIR and play particles/sounds to simulate it breaking.

Add an ItemStack#damage method to tell and have Bukkit handle the amount of damage you want to deal to the ItemStack similar to how [Player#breakBlock](https://papermc.io/javadocs/paper/1.18/org/bukkit/entity/Player.html#breakBlock(org.bukkit.block.Block)) works for block-breaking instead of manually calling each event.

Doing the messy above.

### Other

_No response_

Machine-Maker:
Adds methods notify clients about item breaks and to simulate damage done to an itemstack and all the logic associated with damaging them

Resolves https://github.com/PaperMC/Paper/issues/7755

### DIFF
--- a/patches/api/0410-ItemStack-damage-API.patch
+++ b/patches/api/0410-ItemStack-damage-API.patch
@@ -1,0 +1,86 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 8 May 2022 13:35:58 -0700
+Subject: [PATCH] ItemStack damage API
+
+Adds methods notify clients about item breaks and
+to simulate damage done to an itemstack and all
+the logic associated with damaging them
+
+diff --git a/src/main/java/org/bukkit/entity/LivingEntity.java b/src/main/java/org/bukkit/entity/LivingEntity.java
+index 3bd1d100d0c481ae7edaa251869640ab370aeb42..f6cbada38ca48638e5ad0bd99d0b6ea65f6b02de 100644
+--- a/src/main/java/org/bukkit/entity/LivingEntity.java
++++ b/src/main/java/org/bukkit/entity/LivingEntity.java
+@@ -1015,5 +1015,48 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
+      * @param directionZ The relative z position of the knockback source direction
+      */
+     void knockback(double strength, double directionX, double directionZ);
++
++    /**
++     * Notifies all clients tracking this entity that the item in
++     * the slot of this entity broke.
++     * <p>
++     * NOTE: this does not mutate any entity state
++     *
++     * @param slot the slot
++     */
++    void broadcastSlotBreak(org.bukkit.inventory.@NotNull EquipmentSlot slot);
++
++    /**
++     * Notifies specified players that the item in the slot
++     * of this entity broke.
++     * <p>
++     * NOTE: this does not mutate any entity state
++     *
++     * @param slot the slot
++     * @param players the players to notify
++     */
++    void broadcastSlotBreak(org.bukkit.inventory.@NotNull EquipmentSlot slot, @NotNull Collection<Player> players);
++
++    /**
++     * Damages this itemstack by the specified amount. This
++     * runs all logic associated with damaging an itemstack like
++     * events and stat changes.
++     *
++     * @param stack the itemstack to damage
++     * @param amount the amount of damage to do
++     * @return the damaged itemstack, or an empty stack if it broke. It may be the same instance as the stack passed it, but it may not.
++     */
++    @NotNull ItemStack damageItemStack(@NotNull ItemStack stack, int amount);
++
++    /**
++     * Damages the itemstack in this slot by the specified amount.
++     * This runs all logic associated with damaging an itemstack like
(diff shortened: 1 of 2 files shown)


patches/server/0907-Add-ItemStack-damage.patch
```diff
@@ -0,0 +1,38 @@
...
++                return stack;
++            }
++            nmsStack = craftItemStack.handle;
++        } else {
++            nmsStack = CraftItemStack.asNMSCopy(stack);
++            stack = CraftItemStack.asCraftMirror(nmsStack); // mirror to capture changes in hurt logic & events
++        }
++        nmsStack.hurtAndBreak(amount, this.getHandle(), livingEntity -> {
++            if (slot != null) {
++                livingEntity.broadcastBreakEvent(org.bukkit.craftbukkit.CraftEquipmentSlot.getNMS(slot));
```
kennytv: what happens if I just pass the wrong slot?

Machine-Maker: I'm not sure what you mean the "wrong" slot?

HexedHero: probably by doing something like `damageItemstack(player.getInventory().getItemInMainHand(), 1, EquipmentSlot.FEET);`

Machine-Maker: Yeah, that just tells all the clients to play the break animation and sound for the feet slot. it doesn't change any client state. so it doesn't really matter.

HexedHero: Loop the players armor and hand items to check for the matching slot for the ItemStack given 5Head - Would a `LivingEntity#damageItemstack(Int, EquipmentSlot)` make sense to damage any ItemStack in that slot?

Machine-Maker: Oh that does look better. instead of having the nullable slot, just have a method for just a stack, and one for a slot

kennytv (approved):

patches/api/0384-ItemStack-damage-API.patch
```diff
@@ -0,0 +1,85 @@
...
++     * Damages this itemstack by the specified amount. This
++     * runs all logic associated with damaging an itemstack like
++     * events and stat changes.
++     *
++     * @param stack the itemstack to damage
++     * @param amount the amount of damage to do
++     * @param slot optional slot to broadcast break packets to other players (if it breaks)
++     * @return the damaged itemstack, or an empty stack if it broke. It may be the same instance as the stack passed it, but it may not.
++     */
++    @NotNull ItemStack damageItemstack(@NotNull ItemStack stack, int amount, @Nullable org.bukkit.inventory.EquipmentSlot slot);
```
HexedHero: Should it be damageItemStack? Capital S

electronicboy: yes

HexedHero (approved):
Did brief testing and works as expected

patches/server/0905-ItemStack-damage-API.patch
```diff
@@ -0,0 +1,76 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+index bd5825c5b5c81e0694a3635b981588f0d2ba83cb..fb83564dee8c564cd1359afbb070439df832e958 100644
+--- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
++++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+@@ -4224,7 +4224,7 @@ public abstract class LivingEntity extends Entity {
+ 
+     }
+ 
+-    private static byte entityEventForEquipmentBreak(EquipmentSlot slot) {
++    public static byte entityEventForEquipmentBreak(EquipmentSlot slot) { // Paper - private -> public
```
NoahvdAa: AT

NoahvdAa (approved):

patches/api/0384-ItemStack-damage-API.patch
```diff
@@ -0,0 +1,86 @@
...
+index 62841846ec3e14daa46564509671cab146984cc6..5d5eaca2533354bed9bf1dbfc67993bd7aee876a 100644
+--- a/src/main/java/org/bukkit/inventory/ItemStack.java
++++ b/src/main/java/org/bukkit/inventory/ItemStack.java
+@@ -978,5 +978,19 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, net.kyor
+     public boolean canRepair(@NotNull ItemStack toBeRepaired) {
+         return Bukkit.getUnsafe().isValidRepairItemStack(toBeRepaired, this);
+     }
++
++    /**
++     * Damages this itemstack by the specified amount.This
```
Lulu13022002: Missing space "."[SPACE]"This"

Issue #7770, opened 2022-04-27
Add Event in Paper API about PacketPlayInSteerVehicle


TheCalypso:
### Is your feature request related to a problem?

Hello, the addition that I propose would be in the API paper. In newer versions of Minecraft, clients send a packet called PacketPlayInSteerVehicle. It allows to detect when a player presses a key on his keyboard when he is on a vehicle. We ask you if it would be possible to add an event in order to be able to use it more easily. In our case we would like to use it to be able to more easily control the movement of an entity in our different projects.

We want to extend control of an entity when we're on it. For example we could make a MarioKart like :)

### Describe the solution you'd like.

In order to be able to better access this package, we think it would be wise to add it as an event in the API provided by Paper. This event would be a listen on this packet.

You can find the corresponding code in the PacketPlayerInSteerVehicle class in package minecraft.network.protocol.game .

### Describe alternatives you've considered.

We have already tried to use Protocolibs, only the use of the latter adds a dependency to a plugin.

### Other

Thank you very much for your attention.

Good end of the day to you!

lynxplay:
Previous attempt at implementing this in #7274. If anyone is attempting to implement this, the discussion in there might be useful for you.

TheCalypso:
Thank you for your answer. I don't see what is causing concern in the conversation you sent. Why has this feature not been implemented?

lynxplay:
The author decided against it. I don't know if they simply did not want to deal with the critique or what exactly their reasoning was. As mini said, and I agree with his assessment, there is definitely a usecase for this.

We are just waiting for someone to pick up the torch basically. If you want to, feel free to open it yourself and move the PR forward! Any kind of questions we will be happy to help you.

//edit[0]: For help, just hop by on our discord if you want to PR this :+1:

TheCalypso:
Great thank you, I'll see what I can do. But it is true that the people who answered may lacked a little flexibility in their words, it's a shame. Finally we will try to add this functionality because we think it can really be used for developers wanting to control entities.

lynxplay:
Glad to hear you want to try it 👍  I am sure we can get cat to share his thoughts on a "better suited for proper API over different mechanisms exposed in the game" if he wants to chime in on it if that was the major issue in the last PR.

TheCalypso:
I'm so glad you offered to help me and to see that you are open to discussion! Thanks again, I'm thinking of contacting you directly on Discord to avoid putting too many messages in this conversation.

[...]

PR #7802, opened 2022-05-08
Add PlayerPassengerInputEvent

TheCalypso:
Hello everyone !

I am very happy to be able to present to you the little work that our team has done. Indeed this PR makes it possible to add the recovery of the inputs that a player makes when he is on an entity. To try to best meet expectations, we worked with @lynxplay .

Here is the link to the first discussion I had via my #7770 

As what has been said, we have done everything so that this event is done before the server logic. the location we have chosen allows us to better manage the event. 

We hope this PR will be a help for a potential future full entity control API. In the meantime, we thank you in advance for your feedback.

### DIFF
--- a/patches/server/0907-Add-PlayerPassengerInputEvent-call-in-handlePlayerIn.patch
+++ b/patches/server/0907-Add-PlayerPassengerInputEvent-call-in-handlePlayerIn.patch
@@ -16,7 +16,7 @@ index 1a87f61d534ed531132fb43a9d2a45a4b604a6fc..683ebe21465bb6762c771284924e1bf0
 +        // Paper start - create event control mounted entity
 +        var event = new io.papermc.paper.event.player.PlayerPassengerInputEvent(player.getBukkitEntity(),
 +                packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown());
-+        if (event.callEvent()) return;
++        if (!event.callEvent()) return;
 +        //this.player.setPlayerInput(packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown());
 +        this.player.setPlayerInput(event.getCoefVelocityX(), event.getCoefVelocityZ(), event.isJumping(), event.isCrouching());
 +        // Paper end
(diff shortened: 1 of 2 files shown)


patches/server/0907-Add-PlayerPassengerInputEvent-call-in-handlePlayerIn.patch
```diff
@@ -0,0 +1,28 @@
...
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -494,9 +494,16 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Ser
+     @Override
+     public void handlePlayerInput(ServerboundPlayerInputPacket packet) {
+         PacketUtils.ensureRunningOnSameThread(packet, this, this.player.getLevel());
+-        this.player.setPlayerInput(packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown());
++        // Paper start - create event control mounted entity
++        var event = new io.papermc.paper.event.player.PlayerPassengerInputEvent(player.getBukkitEntity(),
++                packet.getXxa(), packet.getZza(), packet.isJumping(), packet.isShiftKeyDown());
++        if (event.callEvent()) return;
```
lynxplay: This should be `if(!event.callEvent())`. `Event#callEvent` returns `false` if the event was cancelled, e.g. we would want to return at that point.

lynxplay (changes requested):

patches/server/0907-Add-PlayerPassengerInputEvent-call-in-handlePlayerIn.patch
electronicboy: Is this cancellation safe, i.e. it's not gonna cause a desync?

TheCalypso: Oh thank you very much! Sorry I just fixed this :)

TheCalypso: > Is this cancellation safe, i.e. it's not gonna cause a desync?

What do you mean by desync?

PR #7803, opened 2022-05-08
Fix villager cure exploit patch

Warriorrrr:
Fixes the villager cure exploit not working by wrapping the part that awards positive gossip in an else, instead of only removing any previous gossip for the player that cured the villager.


jpenilla (changes requested):
The fix is not broken. The exploit which the patch fixes is the stacking of positive gossip from curing multiple times, a single cure granting a non-stacking positive gossip is not an exploit or bug and is not meant to be broken by this patch (which your change does break).

Warriorrrr:
I see, I wrongly assumed the curing discount was the exploit but missed the stacking part.

PR #7804, opened 2022-05-09
Allow customizing hopper minecart cooldowns

Owen1212055:
Currently, hopper minecarts ignore the cooldown completely, as they never update the last position variable. (It only follows the cooldown if the hopper minecart is at 0,0,0) Mojang has this issue marked as [Works as Intended](https://bugs.mojang.com/browse/MC-65029?jql=text%20~%20%22TransferCooldown%22).

Therefore, I decided to fix this by not using this system and instead just correctly respecting when the cooldown is set. This is because paper sets the cooldown in some places, and I feel it's appropriate to have those applied to these hoppers as well.
This should ensure that the optimizations to hoppers properly have cooldowns applied to hopper minecarts (for example, if the hopper is full).

Another fix is to possibly just update the lastPosition, however, this will cause the optimization to perhaps skip the cooldown in some cases if the hopper minecart moves, which seems a bit silly, especially since this system isn't even used.

Feedback is needed on this if this should be the correct fix, a separate config to change the cooldown on minecart hoppers (currently the normal cooldown of hoppers / 2), or if we just fix the vanilla system.

Also, should this go in a separate patch? Or the optimized hopper patch?

### DIFF
--- a/patches/server/0333-Optimize-Hoppers.patch
+++ b/patches/server/0333-Optimize-Hoppers.patch
@@ -24,6 +24,37 @@ index 1a21f7e590aaeca131256dd7079b9546710ca9ad..eeb794d96ac8cbe36b788d390e638192
  
              this.profiler.push(() -> {
                  return worldserver + " " + worldserver.dimension().location();
+diff --git a/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java b/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
+index 3f8990d15bd6815db0ccb924e621dfcc3220e0e0..4ce983f131ff190c4c9b35ae0a6be46f6655b971 100644
+--- a/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
++++ b/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
+@@ -21,7 +21,7 @@ public class MinecartHopper extends AbstractMinecartContainer implements Hopper
+     public static final int MOVE_ITEM_SPEED = 4;
+     private boolean enabled = true;
+     public int cooldownTime = -1;
+-    private final BlockPos lastPosition = BlockPos.ZERO;
++    private BlockPos lastPosition = BlockPos.ZERO; // Paper
+ 
+     public MinecartHopper(EntityType<? extends MinecartHopper> type, Level world) {
+         super(type, world);
+@@ -90,14 +90,15 @@ public class MinecartHopper extends AbstractMinecartContainer implements Hopper
+             BlockPos blockPos = this.blockPosition();
+             if (blockPos.equals(this.lastPosition)) {
+                 --this.cooldownTime;
+-            } else {
++            } else if (this.level.paperConfig().hopper.resetHopperMinecartCooldownOnMove) { // Paper
++                this.lastPosition = blockPos; // Paper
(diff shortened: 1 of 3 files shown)


Owen1212055:
Added ``hopperMinecartTransferCooldown`` + ``resetHopperMinecartCooldownOnMove`` to the paperconfig.

hopperMinecartTransferCooldown will now control the cooldown that is applied, which because of the fact that this has been non-functional has been set to 0 in order to prevent any breaking of mob farm contraptions. 

resetHopperMinecartCooldownOnMove determines if the cooldown should be reset if the hopper moves blocks. This by default is true, as this is vanilla behavior (that doesn't correctly function).

This in the end causes this to have no change unless configured.

electronicboy (approved):

electronicboy:
I've posted on the issue, and asked privately, about the "WAI" status of that issue but not heard back,  makes very little sense...

patches/server/0341-Optimize-Hoppers.patch
```diff
@@ -46,6 +56,30 @@ index 25fb2789bcec7cf864b3a401610e765daa1abe22..bec5328f1ab05faab83584f75d54a5a6
...
+--- a/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
++++ b/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
+@@ -88,8 +88,9 @@ public class MinecartHopper extends AbstractMinecartContainer implements Hopper
+         super.tick();
+         if (!this.level.isClientSide && this.isAlive() && this.isEnabled()) {
+             BlockPos blockPos = this.blockPosition();
+-            if (blockPos.equals(this.lastPosition)) {
++            if (blockPos.equals(this.lastPosition) && !this.level.paperConfig.resetHopperMinecartCooldownOnMove) { // Paper
+                 --this.cooldownTime;
++                this.lastPosition = blockPos; // Paper
```
electronicboy: redundant?

Owen1212055:
Rebased

patches/server/0333-Optimize-Hoppers.patch
```diff
@@ -24,6 +24,37 @@ index 1a21f7e590aaeca131256dd7079b9546710ca9ad..eeb794d96ac8cbe36b788d390e638192
...
+diff --git a/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java b/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
+index 3f8990d15bd6815db0ccb924e621dfcc3220e0e0..4ce983f131ff190c4c9b35ae0a6be46f6655b971 100644
+--- a/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
++++ b/src/main/java/net/minecraft/world/entity/vehicle/MinecartHopper.java
+@@ -21,7 +21,7 @@ public class MinecartHopper extends AbstractMinecartContainer implements Hopper
+     public static final int MOVE_ITEM_SPEED = 4;
+     private boolean enabled = true;
+     public int cooldownTime = -1;
+-    private final BlockPos lastPosition = BlockPos.ZERO;
++    private BlockPos lastPosition = BlockPos.ZERO; // Paper
```
NoahvdAa: AT

Owen1212055:
This logic no longer exists in vanilla

PR #7808, opened 2022-05-09
Make all recipes implement keyed

u9g:
Allows for doing `recipe.getKey()` from the recipeIterator


Machine-Maker (changes requested):
I think not. First of all randomKey is def not good. If villager trades were to have keys (as I imagine they may in the future when mojang makes them data pack-able) they would have to be generated algorithmically like Forge does with making keys for brewing mixes.

Also I think this should be a post hardfork thing, making villager trades not implement Recipe in the first place. That doesn't match the nms class structure at all.

EDIT: @u9g you can just cast it to Keyed. I know its slightly annoying, but fixing this without just removing Recipe from the villager trade hierarchy is just bandages on an already broken thing.

PR #7585, opened 2022-03-12
deprecate old MaterialData api

Machine-Maker:

PR #7817, opened 2022-05-13
Mark MaterialData related methods in ItemStack as deprecated

Leguan16:
These methods are for the MaterialData class which is already marked as deprecated.


NoahvdAa:
👋 Welcome to Paper! This can probably just go in the [Fix Spigot annotation mistakes](https://github.com/PaperMC/Paper/blob/master/patches/api/0175-Fix-Spigot-annotation-mistakes.patch) patch.

Tiny nitpick: Paper comments should have a capital P and a space between the p and the slashes

Machine-Maker:
There is already an open PR that does this https://github.com/PaperMC/Paper/pull/7585.

Leguan16:
well, i am blind then. Thanks for letting me know.

MiniDigger:
thank you anyways!

PR #7822, opened 2022-05-14
Fix exact choice recipe book clicks

### LINKED ISSUE #7821: Recipe Book clicking only matches material when placing items into the crafting grid
### Expected behavior

With a recipe that requires an item with custom model data (or other modified attributes), when clicking a recipe in the Recipe Book, it will result in the exact item being placed into the crafting grid.

### Observed/Actual behavior

With a recipe that requires an item with custom model data (or other modified attributes), when clicking a recipe in the Recipe Book, it only matches the material of the item and will place any item that matches that material into the crafting grid.

Example can be found below.
image
image

### Steps/models to reproduce

1. Create a recipe that requires an ingredient with attributes/custom model data/nbt.
2. Put an item in your inventory that has the same material, but not the attributes/custom model data/nbt.
3. Click on the recipe book to move the items from your inventory to the crafting grid, it will move that item that matched the material.

### Plugin and Datapack List

None, other than the plugin that registered the recipe.

### Paper version

Paper #333 1.18.2

### Other

Discussed this with Machine Maker in #paper-help on Discord on 5/14/2022 at 1:58 P.M. PST, further context can be found there.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/7821

Anytime a recipe match is going to be computed by the server to try and place the recipe in the crafting area (grid, furnace, etc.), the recipe ingredients are looked at and any ingredients that are marked as "exact" are noted. When the inventory is accounted and unique int ids are created for each item type, any item that matches the exact ingredient tracked before is given an extra separate unique int id. Then that is used to calculate if the recipe is valid.

This also removes the need for the patch added [here](https://github.com/PaperMC/Paper/commit/ac5b19c55e0327e4400d995e7f59a4c58de845c1). This patch fixed shapeless recipes with exact recipe choices due to it uses the same StackedContents system. But now that that system has been integrated with the ExactChoices, we can just use the same thing there and it works.

### Caveats
There are some client-side limitations that are unavoidable as far as I can figure. If you have an exact ingredient that has a custom display name, enchantments or a damage value, the client will show the recipe as red even if you can click on it and it will put the ingredients in the correct slots. This is because the client normally ignores items with names, damage, or enchants when calculating the ingredients for a recipe.

---
Download the paperclip jar for this pull request: [paper-7822.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/879420484.zip)


Boy0000:
Tried building this locally to test myself but couldn't get it to work?
Atleast if its meant to have the autofill work with custom nbt items.

Recipe still only shows as "craftable" when non-nbt variant is in players inventory.
Clicking it with all the NBT items in the inventory does not autofill (manually placing them makes the recipe work ofc)
Unlike before though, the vanilla equivalent items are not autofilled when they are present

Clip below should show everything, triple checked the nbt of items in inv, the recipe ones and result items.
All were identical.

**With patch:**

https://user-images.githubusercontent.com/62521371/170740226-0b7b1a6e-3890-4def-90e1-f2e7e68864b9.mp4

**Without:**

https://user-images.githubusercontent.com/62521371/170740937-74d68244-b20b-4be9-98de-adb5cf46d1e4.mp4

Machine-Maker:
Items with damage values, enchantments, or custom display names do not work with auto filling.

Boy0000:
Ah alright, kinda sad but seems to work atleast without displayname like intended.
Would this only be when the item registered in the recipe has a displayname?
So if i register the recipe with a clone of the itemstack but remove the displayname it would still work?

also when thinking about it yeah displayname and damage makes sense to not be taken into account 💀 
forgot anvil renaming and enchanting was a thing apparently

image

Machine-Maker:
> So if i register the recipe with a clone of the itemstack but remove the displayname it would still work?

No. When the game tries to find stacks that match the recipe it just ignores itemstacks in your inventory that have a display name, enchants, or damage. It might be possible to address this in the future, but that's not what this PR does.

partydev:
I can confirm that this PR fixes #7821.

Previous behavior was that the correct recipe would not be sent to the player if it was marked as craftable. The material for the items in that recipe are sticks, so the recipe is seen as craftable by the client.

image

EDIT:

After 24 hours of testing, including running it on production, I have not seen any side effects from this PR's changes, and that the issue is fixed. I will continue posting updates.

EDIT:

After a a few days of production use, no issues stemming from this.

electronicboy (approved):

Machine-Maker:
I wonder if there's some better way to track items like this, via some registering system for plugins to say, "This item should be treated separately in recipes". Right now, every unique nbt tag on an itemstack will create a separate id, and that could get rather big as there's really infinite nbt combinations.

Machine-Maker:
Dramatically refactored this to an, imo, much better and more extensible system for predicate choice coming later. Details of how this works in PR description.

kennytv (approved):

PR #7825, opened 2022-05-15
Add EntityBrainMemoryRemoveEvent

Togira123:
This adds an event named EntityBrainMemoryRemoveEvent which fires when an entity's memory is removed from its Brain. It can be cancelled, which causes this memory not to be deleted.

### DIFF
--- a/patches/api/0384-Add-EntityBrainMemoryRemoveEvent.patch
+++ b/patches/api/0384-Add-EntityBrainMemoryRemoveEvent.patch
@@ -6,14 +6,13 @@ Subject: [PATCH] Add EntityBrainMemoryRemoveEvent
 
 diff --git a/src/main/java/io/papermc/paper/event/entity/EntityBrainMemoryRemoveEvent.java b/src/main/java/io/papermc/paper/event/entity/EntityBrainMemoryRemoveEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..18d6c9c70e849c984dbe3c657df0e5b3bc384a0a
+index 0000000000000000000000000000000000000000..e9738dd17ff894fa1bdfb82eeae9591db9e3a402
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/entity/EntityBrainMemoryRemoveEvent.java
-@@ -0,0 +1,42 @@
+@@ -0,0 +1,46 @@
 +package io.papermc.paper.event.entity;
 +
 +import org.bukkit.entity.Entity;
-+import org.bukkit.entity.LivingEntity;
 +import org.bukkit.entity.memory.MemoryKey;
 +import org.bukkit.event.Cancellable;
 +import org.bukkit.event.HandlerList;
@@ -22,7 +21,7 @@ index 0000000000000000000000000000000000000000..18d6c9c70e849c984dbe3c657df0e5b3
 +
 +public class EntityBrainMemoryRemoveEvent extends EntityEvent implements Cancellable {
 +
-+    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private static final HandlerList handlers = new HandlerList();
 +
 +    private boolean cancelled;
 +    private final MemoryKey<?> memoryToBeRemoved;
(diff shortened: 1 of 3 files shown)


Togira123:
I'm getting the following errors when running `./gradlew runDev`:
```
/Users/dariollarden/Desktop/java/Paper.nosync/Paper-Server/src/main/java/net/minecraft/world/entity/ai/Brain.java:87: error: cannot infer type arguments for Brain<>
                return DataResult.success(new Brain<>(memoryModules, sensors, immutableList, mutableObject::getValue));
                                          ^
  reason: cannot infer type-variable(s) E
    (actual and formal argument lists differ in length)
  where E is a type-variable:
    E extends LivingEntity declared in class Brain
/Users/dariollarden/Desktop/java/Paper.nosync/Paper-Server/src/main/java/net/minecraft/world/entity/ai/Brain.java:178: error: incompatible types: inference variable U#1 has incompatible bounds
        return this.memories.get(type).map(ExpirableValue::getValue);
                                          ^
    equality constraints: U#2
    lower bounds: CAP#1
  where U#1,T,U#2 are type-variables:
    U#1 extends Object declared in method <U#1>map(Function<? super T,? extends U#1>)
    T extends Object declared in class Optional
    U#2 extends Object declared in method <U#2>getMemory(MemoryModuleType<U#2>)
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Object from capture of ?
/Users/dariollarden/Desktop/java/Paper.nosync/Paper-Server/src/main/java/net/minecraft/world/entity/ai/Brain.java:361: error: cannot infer type arguments for Brain<>
        Brain<E> brain = new Brain<>(this.memories.keySet(), this.sensors.keySet(), ImmutableList.of(), this.codec);
                         ^
  reason: cannot infer type-variable(s) E
    (actual and formal argument lists differ in length)
  where E is a type-variable:
    E extends LivingEntity declared in class Brain
/Users/dariollarden/Desktop/java/Paper.nosync/Paper-Server/src/main/java/net/minecraft/world/entity/ai/Brain.java:472: error: cannot infer type arguments for MemoryValue<>
            return new Brain.MemoryValue<>(type, data);
                   ^
  reason: cannot infer type-variable(s) U
    (argument mismatch; Optional<CAP#1> cannot be converted to Optional<? extends ExpirableValue<U>>)
  where U is a type-variable:
    U extends Object declared in class MemoryValue
  where CAP#1 is a fresh type-variable:
    CAP#1 extends ExpirableValue<?> from capture of ? extends ExpirableValue<?>
```

I managed to fix error 2 & 4 just by adding some casts. The problem with the other two is, that I changed the constructor of `Brain` to also take a `LivingEntity` to store the entity this brain belongs to on the brain itself. So I was able to add the `this.owner` property to the constructor call which was causing error number 3. But now I'm stuck with the first one, since there's no `LivingEntity` submitted anywhere... does anyone have a solution for this?

Togira123:
An alternative would be to pass the entity instance to the `eraseMemory` method directly... I just checked, and it would work, it's just not such an elegant solution. Should I still do it?

maestro-denery:
Imo changing of the `makeBrain()` method parameters and adding additional owner field to the `Brain` class to have an owner is a bad practice. This changes how classes relate to each other. Imo it'll be a PITA to support in the future. It also conflicts with `WIP Brain API` PR. (This PR doesn't have brain events for now, but it relays on minecraft's Brain system, and I think there will be something bad in its logic with the `Brain` having `owner` field). And you also have 2nd API patch where you fix annotations, you should modify the first patch instead of creating another one. (It is described in `CONTRIBUTING.md`)

Togira123:
> Imo changing of the `makeBrain()` method parameters and adding additional owner field to the `Brain` class to have an owner is a bad practice. This changes how classes relate to each other. Imo it'll be a PITA to support in the future. It also conflicts with `WIP Brain API` PR. (This PR doesn't have brain events for now, but it relays on minecraft's Brain system, and I think there will be something bad in its logic with the `Brain` having `owner` field). And you also have 2nd API patch where you fix annotations, you should modify the first patch instead of creating another one. (It is described in `CONTRIBUTING.md`)

alright! So, instead of adding the owner field, what if I always pass the entity with the `eraseMemory` method?

Togira123:
I'll be adding an event which does not also pass the entity because it involves too much changes to do so. Instead, there is just going to be an event which has the memory only as it's parameter

PR #7828, opened 2022-05-16
Explosion Resistance API

Astralchroma:
Adds API functions to `Bukkit`, `Server`, and `Material` to allow changing the explosion resistance of blocks without using NMS or Reflection which is the current method used to acomplish this.

As `Material` does not have a CraftBukkit implementation the functions to alter the resistance is part of `Server`, these functions are aliased in `Bukkit` and `Material` for easy access.

The existing `getBlastResistance` function has been deprecated as the name "explosion resistance" is used by Mojang Mappings. The function also previously used a switch statement to define the explosion resistance which has been commented out in favor of using the actually values.

A `defaultExplosionResistance` has been added to BlockBehaviour which stores the initial value, allowing a plugin to reset the explosion resistance to the value specified by vanilla minecraft.

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -336,3 +336,5 @@ public net.minecraft.world.entity.projectile.FishingHook timeUntilLured
 # Teleport API
 public net.minecraft.server.network.ServerGamePacketListenerImpl internalTeleport(DDDFFLjava/util/Set;Z)V
 
+# Explosion Resistance API
+public-f net.minecraft.world.level.block.state.BlockBehaviour explosionResistance
--- a/patches/api/0392-Explosion-Resistance-API.patch
+++ b/patches/api/0392-Explosion-Resistance-API.patch
@@ -1,0 +1,161 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Peter Crawley <peterzcrawley@gmail.com>
+Date: Mon, 16 May 2022 13:56:53 +0100
+Subject: [PATCH] Explosion Resistance API
+
+
+diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
+index 840aaf9e8fc828b5a7ea02252038c6524680f2e0..b2a59bc124041ee40a4b8af4d6b008054d8d12f5 100644
+--- a/src/main/java/org/bukkit/Bukkit.java
++++ b/src/main/java/org/bukkit/Bukkit.java
+@@ -2436,6 +2436,45 @@ public final class Bukkit {
+     public static @NotNull org.bukkit.potion.PotionBrewer getPotionBrewer() {
+         return server.getPotionBrewer();
+     }
++
++    /**
++     * Gets the explosion resistance of the material
++     * @param material Material to get the explosion resistance of
++     * @return The explosion resistance
++     * @throws IllegalArgumentException if Material is not a block
++     */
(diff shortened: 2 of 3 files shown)


Astralchroma:
No idea why the tests are failing, the code works fine with runDev, may need some help figuring out the issue.

Doc94:
> No idea why the tests are failing, the code works fine with runDev, may need some help figuring out the issue.

the CI shows errors with test for testBlastResistance.

Astralchroma:
> > No idea why the tests are failing, the code works fine with runDev, may need some help figuring out the issue.
> 
> the CI shows errors with test for testBlastResistance.

I know that much, but the values it is returning is correct, so the tests should be passing, but they don't. I have never worked with tests before, so I don't really know how to resolve it.

electronicboy:
See the DummyServer class

electronicboy:
I think part of the reality is that those methods kinda shouldn't be on the Server itself, idk where they'd really be best suited though

Astralchroma:
> I think part of the reality is that those methods kinda shouldn't be on the Server itself, idk where they'd really be best suited though

I agree, however putting them directly in the Material class is not really an option, and it didn't make sense to add a new singleton just for managing explosion resistance.

Astralchroma:
The test probably is not even needed anymore as all we are doing is comparing if the function is returning the very same variable we are comparing against.

Astralchroma:
Also what should be done about the switch statement I commented out, should I leave it commented out, or just remove it?

Astralchroma:
Rebased for 1.19

Astralchroma:
Fixed the issue with the test, was caused by DummyServer not having an implementation for `getExplosionResistance()`.

patches/server/0918-Explosion-Resistance-API.patch
```diff
@@ -0,0 +1,72 @@
...
+ }
+diff --git a/src/test/java/org/bukkit/support/DummyServer.java b/src/test/java/org/bukkit/support/DummyServer.java
+index 1f3705543353adc6e6d4c4a9dd06032255bf5089..21057b91e3228a6075c616f83af4f82891be2090 100644
+--- a/src/test/java/org/bukkit/support/DummyServer.java
++++ b/src/test/java/org/bukkit/support/DummyServer.java
+@@ -107,6 +107,9 @@ public final class DummyServer implements InvocationHandler {
+                     }
+                 }
+             );
++            methods.put(Server.class.getMethod("getExplosionResistance", Material.class),
```
Lulu13022002: Missing Paper comments here

Astralchroma: Inside pre-existing `// Paper start` block.

Lulu13022002: Oops sorry

Astralchroma:
Rebased

electronicboy (approved):
Generally looks fine, I'm somewhat iffy on methods being added to Bukkit for this, but, as said, there is no real general place for this stuff outside of Unsafe...

Astralchroma:
Rebased

Astralchroma:
Rebased for 1.19.1

Astralchroma:
Rebased

Owen1212055:
Thank you for your contribution, however, generally api like this is tricky as this is mutating data that really isn't expected to be mutated at all. Being able to modify a part of a block like this without notifying the client is a bit iffy. And in general, being able to change properties like this is something that we want to better support in the future. For now, it's better that these hard coded values stay immutable for the time being. If you want something like this it's most likely better to do this in your own fork.

PR #7830, opened 2022-05-16
Fix POI max search distance

Samsuik:
This fixes the maxDistance provided to the methods in PoiAccess. 

The radius provided is squared and then squared again in PoiAccess causing it to check if the POI is in range with a cubed radius. This often causes it to search from the bottom to the top of the world for POI's and also deviates from the vanilla radius.


kennytv (changes requested):
Considering the use in PortalForcer, it'd be better if the squaring were removed from the PoiAccess method (and the parameter renamed), not the method calls

Owen1212055:
Closing, please feel free to leave a comment if you are interested in continuing this PR and it can be reopened. Thank you 😄 

If anyone else would like to pickup what has been done here, feel free to do so!

PR #7833, opened 2022-05-20
Add metrics for datapack usage 

Malfrador:
This PR adds a bstats metrics chart for the amount of enabled datapacks, very similar to the [already existing chart](https://bstats.org/plugin/server-implementation/Paper/580#legacy_plugins) for legacy plugins.

This can be used to make data-based decisions on changes like the dimension storage rework (https://github.com/PaperMC/Paper/pull/7594).

Of course the `datapacks` chart needs to be added to bstats as well for this to work. 
Also data about the type of datapack (e.g. worldgen, loot or dimensions) would be very useful as well, but I have not found a way to get it other than looking at the files directly, and that seems a bit messy. 

cat talked about _maybe_ doing this on Discord so I thought I just make a quick PR. I added the changes to the existing metrics patch, that seemed to be the right place

### DIFF
--- a/patches/server/0010-Paper-Metrics.patch
+++ b/patches/server/0010-Paper-Metrics.patch
@@ -15,10 +15,10 @@ decisions on behalf of the project.
 
 diff --git a/src/main/java/com/destroystokyo/paper/Metrics.java b/src/main/java/com/destroystokyo/paper/Metrics.java
 new file mode 100644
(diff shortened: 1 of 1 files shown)


MiniDigger:
this reminds me of https://github.com/PaperMC/Paper/pull/6882...

Owen1212055 (approved):
Looks fine to me, but I know that there is a pr open to already redo the metrics... so I am not sure which of these two will come first.

PR #7835, opened 2022-05-20
[ci skip] Update repository and API urls

NoahvdAa:
Same as [9936e2b](https://github.com/PaperMC/papermc.io/commit/9936e2b85c65edb5a7ecb06c960988ec0097793d) and [0f41894](https://github.com/PaperMC/docs/commit/0f41894b9b2d41568f0ab0e4670117a4996a84bb) but for paper itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ How To (Plugin Developers)
 ```xml
 <repository>
     <id>papermc</id>
-    <url>https://papermc.io/repo/repository/maven-public/</url>
+    <url>https://repo.papermc.io/repository/maven-public/</url>
 </repository>
 ```
  * Artifact Information:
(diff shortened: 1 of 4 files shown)


Doc94:
Maybe this can include the change for javadocs url? https://jd.papermc.io/paper/1.18/

NoahvdAa:
That link points to the page with javadocs for multiple versions and doesn't have to be changed. (Unless we want to link to a specific version, which is worse than the current solution IMO)

Doc94:
> That link points to the page with javadocs for multiple versions and doesn't have to be changed. (Unless we want to link to a specific version, which is worse than the current solution IMO)

oh this make sense... but i mean point to a target version dont sounds bad if the branch is only for one version, the only thing is remember update the URL when release a new version (?

kashike (approved):

PR #7836, opened 2022-05-20
[ci skip] Update WSL docs in CONTRIBUTING

Doc94:
This PR change a few things in the WSL docs
- Change the URL for install and filesystems and latest in Windows Docs
- Mention the update things only necesary for Windows 10

PR #7838, opened 2022-05-21
[ci skip] Update outdated windows version info

nostalfinals:


electronicboy:
Theres already a PR looking to change that up #7836

PR #7839, opened 2022-05-21
Team color parameter may be null, causing NPE

rgnter:
Implementation must conform to declaration of method.
```java
void color(@Nullable net.kyori.adventure.text.format.NamedTextColor color);
```
When null is passed as method argument, it must be handled by implementation not to cause NPE. 
In this case, when this class was upgraded from 1.18.R1 to R2, this check was removed.


kennytv:
Please edit the original patch adding this method (the adventure patch), see the contribution guide for details on how to do that

electronicboy:
Thanks for the PR, manually merged into https://github.com/PaperMC/Paper/commit/d8108b5416ff5482546c380d2cf5eae9cfda9eed

In the future, please don't PR from the master branch

PR #7840, opened 2022-05-22
Add BlockLockCheckEvent

Machine-Maker:
Adds an event to further extend vanilla's locked block feature.

Couple things to address. Not a huge fan of the generic hack to combine `TileState`, `Nameable`, and `Lockable`. Could either make `Lockable` extend both `TileState` and `Nameable` or make a new interface `LockableTileState` to combine them.

### DIFF
--- a/patches/api/0420-Add-BlockLockCheckEvent.patch
+++ b/patches/api/0420-Add-BlockLockCheckEvent.patch
@@ -1,0 +1,242 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 21 May 2022 20:59:56 -0700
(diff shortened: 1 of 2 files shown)


kennytv (dismissed):
1-2 lines of docs on the event itself would be nice, otherwise looks good

kennytv (approved):

electronicboy (approved):

Owen1212055 (approved):
Looks good, but that deprecated method can probably be replaced with the DoNotUse when rebased.

jpenilla:
> Looks good, but that deprecated method can probably be replaced with the DoNotUse when rebased.

donotuse does not replace deprecated, they are both needed

Owen1212055:
> > Looks good, but that deprecated method can probably be replaced with the DoNotUse when rebased.
> 
> donotuse does not replace deprecated, they are both needed

Ah, thanks for the clarification... makes sense.

PR #7841, opened 2022-05-22
Error:  If you wish to use IP forwarding, please enable it in your BungeeCord config as well!

Orel31:
The problem is that on version 1.18.2, some players can log into the server, and for some it says "/192.168.88.104:36871 lost connection: If you wish to use IP forwarding, please enable it in your |
BungeeCord config as well! " issuing a local ip registered in bungeecord. We have a configured bungeecord and set ip_forwarding to true, and also configured in spigot.yml
bungeecord: true. I repeat, some players log in calmly, and some give these errors with a local IP address.

P.S. We have servers on other versions, everything is configured the same on them and no one has such an error.


electronicboy:
Please use the discord/forums for support

PR #7845, opened 2022-05-23
Add BrainMemoryUpdateEvent

Togira123:
This PR adds an event that fires whenever an entity updates its memory. It differentiates between memory add/change/delete. The event can be cancelled which causes this memory not to be updated. The event has two parameters, the first one being the MemoryKey of the memory that is being updated and the second one being either "ADD", "CHANGE" or "DELETE", which are properties of an enum declared in the event class. It's unfortunately not possible to also pass an Entity in that event since there is no reference from the brain class to the entity this brain belongs to. If I wanted to pass the entity directly with the methods that update the memory instead I'd have to change like 80 files, which seemed a bit overkill.


patches/api/0384-Add-BrainMemoryUpdateEvent.patch
```diff
@@ -0,0 +1,71 @@
...
++import org.jetbrains.annotations.NotNull;
++
++public class BrainMemoryUpdateEvent extends Event implements Cancellable {
++    public enum type {
++        ADD, CHANGE, DELETE
++    }
++
++    public static final type ADD = type.ADD;
++    public static final type CHANGE = type.CHANGE;
++    public static final type DELETE = type.DELETE;
```
kennytv: Remove these

patches/api/0384-Add-BrainMemoryUpdateEvent.patch
```diff
@@ -0,0 +1,71 @@
...
++package io.papermc.paper.event.entity;
++
++import org.bukkit.entity.memory.MemoryKey;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++public class BrainMemoryUpdateEvent extends Event implements Cancellable {
++    public enum type {
```
kennytv: Uppercase T

patches/server/0905-Fix-decompile-errors.patch
```diff
@@ -0,0 +1,49 @@
...
+                     DataResult<? extends Brain.MemoryValue<?>> dataResult2 = dataResult.flatMap((memoryModuleType) -> {
+                         return this.captureRead(memoryModuleType, dynamicOps, (T)pair.getSecond());
+                     });
+-                    mutableObject.setValue(mutableObject.getValue().apply2(Builder::add, dataResult2));
++                    mutableObject2.setValue(mutableObject2.getValue().apply2(Builder::add, dataResult2)); // Paper - decompile error
+                 });
+-                ImmutableList<Brain.MemoryValue<?>> immutableList = mutableObject.getValue().resultOrPartial(Brain.LOGGER::error).map(Builder::build).orElseGet(ImmutableList::of);
+-                return DataResult.success(new Brain<>(memoryModules, sensors, immutableList, mutableObject::getValue));
++                ImmutableList<Brain.MemoryValue<?>> immutableList = mutableObject2.getValue().resultOrPartial(Brain.LOGGER::error).map(Builder::build).orElseGet(ImmutableList::of); // Paper - decompile error
++                return DataResult.success(new Brain<>(memoryModules, sensors, immutableList, mutableObject::getValue)); // Paper - decompile error
```
kennytv: This line didn't change

patches/server/0906-Add-BrainMemoryUpdateEvent.patch
```diff
@@ -0,0 +1,50 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/ai/Brain.java b/src/main/java/net/minecraft/world/entity/ai/Brain.java
+index 91d266bd1fc37c9e284a99bdbb465d0e3100305a..8e0024694c8fe76f44c9211c050522dc65827cf0 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/Brain.java
++++ b/src/main/java/net/minecraft/world/entity/ai/Brain.java
+@@ -163,7 +163,38 @@ public class Brain<E extends LivingEntity> {
+             if (memory.isPresent() && this.isEmptyCollection(memory.get().getValue())) {
+                 this.eraseMemory(type);
+             } else {
+-                this.memories.put(type, memory);
++                // Paper start - check whether memory is being added/changed/deleted
```
kennytv, line 16: Clean this up by only putting the change type into a var since the rest of the call is always the same

kennytv (changes requested):

Togira123:
Why did the check fail this time on that commit, even though it succeeded a week ago? it still results in "build successful" for me when running ./gradlew test on my local repo...

electronicboy:
The caching is probably screwing it up, as this is technically trying to do a build with older stuff, i.e. a decompiler update which was recently done on the main branch

But, I'm not fond of this, I really fail to understand the merit of such an event especially within the context, i.e. none, not even the entity is available

Togira123:
> The caching is probably screwing it up, as this is technically trying to do a build with older stuff, i.e. a decompiler update which was recently done on the main branch
> 
> But, I'm not fond of this, I really fail to understand the merit of such an event especially within the context, i.e. none, not even the entity is available

Alright, I knew the use cases for this event would probably be very small. I'll close this pull request then.

PR #7847, opened 2022-05-24
Allow using degrees for ArmorStand rotations

SoSeDiK:
Introduces the new `Rotations` class that is used for storing rotations in degrees on each axis (X, Y, Z).

Currently, the patch adds usage for it for ArmorStand rotations, which makes manipulations easier and more precise 
(since it follows vanilla logic of using degrees instead of radians).

Example usage to set the right arm's rotation to 90 degrees on the X axis:
```java
armorStand.setRightArmRotations(Rotations.ofDegrees(90, 0, 0));
```

Simple! Before you'd have to use `EulerAngle` with a workaround like `Math.toRadians(90)`.

Feel free to give any feedback :)

### DIFF
--- a/patches/api/0131-Expand-ArmorStand-API.patch
+++ b/patches/api/0131-Expand-ArmorStand-API.patch
@@ -3,12 +3,15 @@ From: willies952002 <admin@domnian.com>
 Date: Thu, 26 Jul 2018 02:22:44 -0400
 Subject: [PATCH] Expand ArmorStand API
 
-Add the following:
+Adds the following:
 - Add proper methods for getting and setting items in both hands. Deprecates old methods
 - Enable/Disable slot interactions
+- Allow using Vectors for ArmorStand rotations
+
+Co-authored-by: SoSeDiK <mrsosedik@gmail.com>
 
 diff --git a/src/main/java/org/bukkit/entity/ArmorStand.java b/src/main/java/org/bukkit/entity/ArmorStand.java
-index 2ee3814a52945f541e049b621b9552f8ae9e261d..707639096657f995cc812c7b50108eeed48e8181 100644
+index 2ee3814a52945f541e049b621b9552f8ae9e261d..b5ec4b980b5596dca438919bd2e97cbcc6147fef 100644
 --- a/src/main/java/org/bukkit/entity/ArmorStand.java
 +++ b/src/main/java/org/bukkit/entity/ArmorStand.java
 @@ -14,7 +14,7 @@ public interface ArmorStand extends LivingEntity {
@@ -29,7 +32,7 @@ index 2ee3814a52945f541e049b621b9552f8ae9e261d..707639096657f995cc812c7b50108eee
       */
      @Deprecated
      void setItemInHand(@Nullable ItemStack item);
-@@ -379,5 +379,71 @@ public interface ArmorStand extends LivingEntity {
+@@ -379,5 +379,173 @@ public interface ArmorStand extends LivingEntity {
       * @param tick {@code true} if this armour stand can tick, {@code false} otherwise
       */
      void setCanTick(final boolean tick);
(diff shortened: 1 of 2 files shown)


patches/server/0230-Implement-Expanded-ArmorStand-API.patch
```diff
@@ -87,6 +90,74 @@ index 82b9ee993b0d2e7e0685231f7bad2b85756ec959..f4065938bbfd04519d1363ee8781c316
...
++    public org.bukkit.util.Vector getHeadRotation() {
++        return fromNMSRotation(getHandle().headPose);
++    }
++
++    @Override
++    public void setHeadRotation(org.bukkit.util.Vector rotation) {
++        getHandle().setHeadPose(toNMSRotation(rotation));
++    }
++
++    private static org.bukkit.util.Vector fromNMSRotation(Rotations old) {
```
DaRacci: ~~I think these nms methods for converting the rotation should be moved from the api module to server such as with potion effects have methods in CraftPotionUtil that convert to and from nms types.~~
Disregard this is already in the server side lol

DaRacci (commented):
Disregarding my previous comment, i think this should allow easier manipulation of armour stands for plugins which seems like a great idea to me.

kennytv:
Using the Vector class here doesn't really make sense

SoSeDiK:
For me it just does the job of storing degrees for each axis 😅
Though I understand what you mean

Would you prefer a new class that would store 3 numbers?
Possibly similar to EulerAngle, but for degrees?
If so, any name/additional method suggestions?

kennytv:
Could just go with the Vanilla name of Rotations

SoSeDiK:
Made a test version, open to suggestions.
`Rotations` could be a record in theory, but I'm not 100% on this.

kennytv:
Sure, making it a record would fit here

SoSeDiK:
Should be ready now. Namings/javadocs can be tweaked if needed.

Owen1212055:
I’m curious what was the reasoning behind adding an entirely new type rather than just doing something like a new utility method? Because at least currently it doesn’t look immediately obvious what is the difference between these two classes… not to mention you now have to add getters and setters for each limb. Idk, just seems cluttered.

SoSeDiK:
> I’m curious what was the reasoning behind adding an entirely new type rather than just doing something like a new utility method? Because at least currently it doesn’t look immediately obvious what is the difference between these two classes… not to mention you now have to add getters and setters for each limb. Idk, just seems cluttered.

I'm not sure what you mean. What kind of utility method?
Currently, the only way to change ArmorStand's rotations is via `EulerAngle` class which uses radians. Newly added `Rotations` follows vanilla and works with degrees.
If you meant expanding `EulerAngle` then it'd require adding static `fromDegrees()` method with additional conversion methods for each axis as well as already present add/substruct (at least 9 in total) that would constantly rely on `Math.toDegrees()` and `Math.toRadians()`. I don't think this would be a better approach and I'd say would clutter `EalerAngle` instead.

kennytv:
Since we have plans on fixing the clusterfuck that are Location and Vector usage in Bukkit, use-specific classes that can also be used elsewhere are the way to go... tho also with that in mind, one last request: Please make `Rotations` an interface with a private inner record implementing it. Instead of the public constructor, you then use a static `ofDegrees` method on the interface. Please also move it from the `/util` package to `/math`

SoSeDiK:
> make `Rotations` an interface with a private inner record implementing it

I suppose you meant package-private implementation since it's not possible to have an inner private record inside the interface (classes are forced to be public static in there by default and records can't have a non-public constructor).

I tried to do it similarly to what adventure does. I'm open to any suggestions.

patches/api/0131-Expand-ArmorStand-API.patch
```diff
@@ -3,12 +3,160 @@ From: willies952002 <admin@domnian.com>
...
++package io.papermc.paper.math;
++
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Rotations is an immutable object that stores rotations
++ * in degrees on each axis (X, Y, Z).
++ */
++@ApiStatus.NonExtendable
```
kennytv: you can remove this annotation

patches/api/0131-Expand-ArmorStand-API.patch
```diff
@@ -3,12 +3,160 @@ From: willies952002 <admin@domnian.com>
...
++public interface Rotations {
++
++    /**
++     * Returns Rotations instance with every axis set to 0
++     */
++    static @NotNull Rotations zero() {
```
kennytv: ZERO you can keep as a (static final) field in the Interface

kennytv (approved):
Looks good

Owen1212055:
Might be a little late, but any reason why we use doubles instead of Floats? Minecraft uses floats for armor stand rotations.

SoSeDiK:
> Might be a little late, but any reason why we use doubles instead of Floats? Minecraft uses floats for armor stand rotations.

Just for ease of use. No need to specify that the value `45.5` is float `45.5F` every time or cast `(float) someValue` wherever you have double in your code. In theory, `Rotations` can be used in other places too.

Owen1212055 (approved):

PR #7855, opened 2022-05-28
feat: Add Multiply total operation

DaRacci:
This allows the use of proper multiplication values, as none of the pre-existing bukkit enums properly convey the way to do this. And as exampled by me, The multiply_scalar_1 operation is frankly quite dumb, why add 1 to the users pre-existing input its just asking for unwanted pain while trying to figure out why the players health it 36 and not 18.

### DIFF
--- a/patches/server/0906-implement-real-multiplication-operation.patch
+++ b/patches/server/0906-implement-real-multiplication-operation.patch
@@ -1,0 +1,42 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DaRacci <racci@racci.dev>
+Date: Sat, 28 May 2022 23:44:17 +1000
+Subject: [PATCH] implement real multiplication operation
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeInstance.java b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeInstance.java
+index 9c58d1190377ca887aa703cb8a96a2b028b63af1..c57578d3b8f66ed1bf355939c6a9340b9ec7ebcc 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeInstance.java
++++ b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeInstance.java
+@@ -147,7 +147,11 @@ public class AttributeInstance {
+         for(AttributeModifier attributeModifier3 : this.getModifiersOrEmpty(AttributeModifier.Operation.MULTIPLY_TOTAL)) {
+             e *= 1.0D + attributeModifier3.getAmount();
+         }
+-
++        // Paper start - implement proper multiplication operation
++        for(AttributeModifier attributeModifier4 : this.getModifiersOrEmpty(AttributeModifier.Operation.MULTIPLY_TOTAL_REAL)) {
++            e *= attributeModifier4.getAmount();
++        }
++        // Paper end
+         return this.attribute.sanitizeValue(e);
+     }
+ 
+diff --git a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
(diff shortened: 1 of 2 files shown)


tomcraft:
Why don't you just substract 1 to the value in your attribute modifier ? 😅

DaRacci:
I am currently, but the fact that it took me 4 hours of messing with operations, asking the paper dev channel and it seeming like no one really knows to just say subtract 1 from it, then finally finding my answer by looking at the source code where its actually handled.
I think this means in reality that having a better defined operation like multiply total makes sense as you can see that, understand it and role with it.

patches/server/0906-implement-real-multiplication-operation.patch
```diff
@@ -0,0 +1,42 @@
...
+ 
+diff --git a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
+index 09b8fa2cee63e6bcb92ffc3694f67d0f90a3805d..ecc109d7463f903a72bc69bd3f462cde7641c197 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
++++ b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
+@@ -98,9 +98,12 @@ public class AttributeModifier {
+     public static enum Operation {
+         ADDITION(0),
+         MULTIPLY_BASE(1),
+-        MULTIPLY_TOTAL(2);
```
electronicboy, line 40: Definitely not happy with the notation of taking up the obvious next ID used by mojang if they expanded this, we really should shift off at least to 100 if we are to add custom operation types

electronicboy (changes requested):

patches/server/0906-implement-real-multiplication-operation.patch
DaRacci: Currently without my PC for afew days but I'll fix this when I'm able to get back to it next

electronicboy:
fromId needs patching to cater for the offset; I probs just wouldn't place them in there, and just add an if block inside of the else block for those special cased IDs

patches/server/0906-implement-real-multiplication-operation.patch
```diff
@@ -0,0 +1,43 @@
...
+index 09b8fa2cee63e6bcb92ffc3694f67d0f90a3805d..f49a449bfa7f5372cf89680dfea6db637fce4b18 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
++++ b/src/main/java/net/minecraft/world/entity/ai/attributes/AttributeModifier.java
+@@ -98,9 +98,13 @@ public class AttributeModifier {
+     public static enum Operation {
+         ADDITION(0),
+         MULTIPLY_BASE(1),
+-        MULTIPLY_TOTAL(2);
++        // Paper start - implement proper multiplication operation
++        MULTIPLY_TOTAL(100),
```
Lulu13022002: Only the custom operation type should be shifted, not the vanilla one. Also the bukkit javadoc is missing for the new type

DaRacci: That would appear to be a mess up on my end, ill fix that lol

DaRacci:
> fromId needs patching to cater for the offset; I probs just wouldn't place them in there, and just add an if block inside of the else block for those special cased IDs

Should this be like just an if (id == 103) or like what?

electronicboy:
> Should this be like just an if (id == 103) or like what?

Yup

DaRacci:
> > Should this be like just an if (id == 103) or like what?
> 
> Yup

Should it be removed from the OPERATIONS array aswell?

electronicboy:
I would imagine so, as it's gonna create an oddity when you can lookup a value with two different IDs

Owen1212055:
Ima bit iffy on this myself as this is an entirely new attribute type that we need to upkeep… all for just not subtracting 1? This feels like with documentation this can be fixed?

Owen1212055:
Thank you for your contribution, however, as specified above, it’s really better to improve the documentation here instead of adding our own new type entirely.

This would require up keeping as this introduces a new type entirely. And since clients won’t

PR #7857, opened 2022-05-29
Update ForgeFlower

jpenilla:


kennytv (approved):

jswashburn:
@jpenilla I got an error: "Invalid or corrupt jar file" after running `java -jar paper-1.18.2-351.jar --no-gui`
Never got the error with 350... do we need to update java or something?

electronicboy:
Nothing to do with this commit, this only changes an intermediate step in our build process

e-im:
I'd attempt downloading Paper again. This is more than likely caused by a download issue. If it persists, try opening the jar you downloaded and ensure it is a valid zip archive.

PR #7862, opened 2022-05-31
Deprecate world names

Machine-Maker:
World's should not be identified with strings, they should be identified with namespaced keys. 

### TODO

- [ ] Check that timings reports won't be affected
- [ ] World component display name
- [ ] MapItemSavedData's level name usage
- [ ] CraftServer's Map<String, World> using level name
- [ ] World adventure's Name pointer
- [ ] Move hunks into their correct patches
- [ ] Paper comments in the right places

### DIFF
--- a/patches/api/0384-Deprecate-world-names.patch
+++ b/patches/api/0384-Deprecate-world-names.patch
@@ -1,0 +1,166 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Mon, 30 May 2022 15:32:46 -0700
+Subject: [PATCH] Deprecate world names
+
+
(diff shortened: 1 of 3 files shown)


patches/api/0384-Deprecate-world-names.patch
```diff
@@ -0,0 +1,174 @@
...
++    @Deprecated(forRemoval = true)
+     @NotNull
+     String getWorldName();
+ 
++    // Paper start
++    /**
++     * @return the key for the parent world
++     */
++    @Override
++    @NotNull NamespacedKey getKey();
```
kashike: This feels wrong. Should be `getWorldKey`.

electronicboy: Yea, I'm not sure this should really be implementing Keyed as it's not really an identity for the Chunk snapshot

PR #7865, opened 2022-06-02
Move invisible methods to Entity

Owen1212055:
I'm not sure I did the paper comments for the removed diffs right, I wasn't sure if the previous ones would apply nicely or not. 
It should be noted ItemFrames currently has setVisible/isVisible.

See: https://canary.discord.com/channels/289587909051416579/555462289851940864/981740402551963688

It might be worth expanding the documentation, but opinions are appreciated!


Machine-Maker (changes requested):
don't remove the methods from LivingEntity and CraftLivingEntity.

Also make the javadocs for the methods on Entity different since they don't actually make the entity invisible

Owen1212055:
Resolved and rebased

PR #7867, opened 2022-06-02
Update Adventure to 4.11.0 and implement ComponentLogger

4drian3d:
Updated Adventure to version 4.11.0 and added the getComponentLogger method to the Plugin interface to be able to display components in the console through the respective plugin's logger instead of using ConsoleCommandSender#sendMessage


zml2008 (changes requested):
You'll need to provide a ComponentLoggerProvider service that directs components to be serialized to paper's modified legacy format to actually do coloured logging

PR #7868, opened 2022-06-02
Mark books created by players as resolved

booky10:
Fixes an exploit that causes a hang when resolving a book nbt tag on book open.

I'm not sure this is something Paper would be interested in fixing, because this can be fixed by plugins and is only caused by giving players creative mode.


Machine-Maker:
Well what if the book isn't resolved? Like wouldn't this break resolving components for written books?

PR #7870, opened 2022-06-03
Add sendMiniMessage and sendPlainMessage methods

kennytv:
Not sure about any more in-depth docs there, but there we go, exemplifying how easy it is to use minimessage and at least providing one shorthand method for the most common usecase + having something separate from the legacy text accepting sendMessage method for plain text

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -1404,7 +1404,7 @@ index 4bfc214685164a38ba4261b2bae7faa8a3bd297e..11af71edd7eaefb3cce3c2b59d390ff9
          if (sendToSource && !(source instanceof ConsoleCommandSender)) {
              source.sendMessage(message);
 diff --git a/src/main/java/org/bukkit/command/CommandSender.java b/src/main/java/org/bukkit/command/CommandSender.java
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Do we care about minimessage or plain methods for actionBar messages? or just regular chat

kennytv:
I'd only have it here before people start asking for that in even more places, just there to show it's easy

imDaniX:
Calling it `sendMiniMessage` seems a bit obscure outside the context. Also, `MiniMessage` class itself is a serializer, not a message type - probably something like `sendMiniMessageMessage` would be more appropriate, but that's, well... you can see it yourself. Can't the method be called something like `sendRichMessage` instead?

kennytv:
I like sendRichMessage more as well; any other opinions?

Machine-Maker:
I like it as well, but sendRichMessage (removing association with MiniMessage from the method name) makes it seem official so I think this should be done when mini message becomes the official text representation. when configs and whatnot default to mini message.

FawksX:
Any consideration to have the original `sendMessage`  undeprecated and use the LegacyComponentSerializer instead?

Machine-Maker:
The original sendMessage isn't deprecated, but it will be at some point in the future

kennytv:
Added in https://github.com/PaperMC/Paper/commit/09c9215b85b14be60c3a3dc22ea6f95d4b2615bb

PR #5605, opened 2021-05-10
More MemoryKey API

Machine-Maker:
So I added a bunch more of the memory keys. 
Pretty sure I tested one of every type to ensure the conversions were working correctly.

Only ones I **didn't** add keys for, are the ones without a clear bukkit equivalent class to its type. These are the nms classes I didn't know of a bukkit equivalent for:
* PositionTracker
* WalkTarget
* DamageSource (1.19.4 should make adding API for this nicer)
* Path (just needs a slight re-work of paper's path api impl)

PR #7877, opened 2022-06-06
Event/Effect API Improvements

Owen1212055:
Adds block event api, redoes the world event api,  and adds alot of missing entity effect api.
I have added 1.19 stuff because it's coming soon.

Feedback Wanted:
### **BlockEvent**
- Is it worth having the applicable materials on the event? It's not actually needed anywhere because Minecraft does the checks itself. I was debating wether or not to put it into its individual block classes, but I am not sure as it wouldn't fit EntityEffect.
### **In general**
- Most likely will convert these classes into type tokens (https://github.com/PaperMC/Paper/pull/5605), as then it would allow the nullable Axis to be an Optional.
- Any suggestions on a better way to handle converting these types instead of the big crazy instanceof chain? Might be the only way but I am not sure if there is a nicer way available for now.
- Better names? These names collide so maybe it would be better to keep them named as ``XEffect``? Although minecraft refers to them as events it might not be worth changing it, but, opinions.

TODO:
- Convert to type tokens?
- Docs

Each type has been tested (the non 1.19 ones of course).

### DIFF
--- a/patches/api/0384-Event-Effect-API-Improvements.patch
+++ b/patches/api/0384-Event-Effect-API-Improvements.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
 Date: Sun, 5 Jun 2022 21:19:40 -0400
-Subject: [PATCH] Block/World Event API
+Subject: [PATCH] Event/Effect API Improvements
 
 
 diff --git a/src/main/java/io/papermc/paper/block/event/BlockEvent.java b/src/main/java/io/papermc/paper/block/event/BlockEvent.java
@@ -169,17 +169,19 @@ index 0000000000000000000000000000000000000000..f3a0bde72d5e8b9edca46ea2f376359f
 +}
 diff --git a/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java b/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..379357a7ff1e08a49697c4c82097d0ca3f1ee980
+index 0000000000000000000000000000000000000000..59f77614f8c2e21c07be44be00f2a8b4e3b450c9
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java
-@@ -0,0 +1,24 @@
+@@ -0,0 +1,26 @@
 +package io.papermc.paper.world.event;
 +
 +import org.bukkit.block.BlockFace;
-+import org.jetbrains.annotations.Nullable;
++import org.jetbrains.annotations.NotNull;
 +import org.jetbrains.annotations.Range;
 +
-+public record SculkChargeEventConfiguration(@Range(from = 0, to = 1000) int charge, @Nullable BlockFace... blockFaces) {
++import java.util.Collection;
++
++public record SculkChargeEventConfiguration(@Range(from = 0, to = 1000) int charge, @NotNull Collection<BlockFace> blockFaces) {
 +
 +    /**
 +     * Gets the charge of this current sculk charge event.
@@ -192,8 +194,8 @@ index 0000000000000000000000000000000000000000..379357a7ff1e08a49697c4c82097d0ca
 +    /**
 +     * Gets the block faces that this sculk charge effects.
 +     */
-+    @Nullable
-+    public BlockFace[] blockFaces() {
++    @NotNull
++    public Collection<BlockFace> blockFaces() {
 +        return this.blockFaces;
 +    }
 +}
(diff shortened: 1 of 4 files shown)


kennytv (dismissed):
We definitely don't want to manually keep track of those hardcoded numbers for WorldEvents, not sure yet what that would look like (also something we'll have to do for other things in the future)... There's also not much of a point in having to update Bukkit's Effect enum given it's more than fully replaced here, just keep it with the deprecation for removal and tell people to move on

Owen1212055:
> We definitely don't want to manually keep track of those hardcoded numbers for WorldEvents, not sure yet what that would look like (also something we'll have to do for other things in the future)... There's also not much of a point in having to update Bukkit's Effect enum given it's more than fully replaced here, just keep it with the deprecation for removal and tell people to move on

I should mention that we will be able to a test to validate and make sure we have them all. See stuff like WorldEvents.
Do you want me to not update the entity effect enum then? I'll just deprecate it and replace it then.

patches/api/0384-Event-Effect-API-Improvements.patch
```diff
@@ -0,0 +1,775 @@
...
++    @ApiStatus.Internal
++    public Class<T> getConfigurationClass() {
++        return clazz;
++    }
++}
+diff --git a/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java b/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..59f77614f8c2e21c07be44be00f2a8b4e3b450c9
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/world/event/SculkChargeEventConfiguration.java
```
Machine-Maker, line 174: Should be inside a separate inner package like `world/event/data/`

patches/api/0384-Event-Effect-API-Improvements.patch
```diff
@@ -0,0 +1,775 @@
...
++}
+diff --git a/src/main/java/io/papermc/paper/world/event/SimpleWorldEvent.java b/src/main/java/io/papermc/paper/world/event/SimpleWorldEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..543b040e93b3e1fe875ac79508ffb8fe69ddeb95
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/world/event/SimpleWorldEvent.java
+@@ -0,0 +1,8 @@
++package io.papermc.paper.world.event;
++
++public class SimpleWorldEvent extends WorldEvent {
```
Machine-Maker, line 210: Should be a non-public innerclass of WorldEvent. The less public types we expose to the API, the better. Although, what is the point of this? Why not make WorldEvent not-abstract and still extend it for the other types?

patches/api/0384-Event-Effect-API-Improvements.patch
```diff
@@ -0,0 +1,775 @@
...
++import org.bukkit.Material;
++import org.jetbrains.annotations.ApiStatus;
++
++import java.util.Set;
++
++/**
++ * Represents an event that will cause
++ * an effect to occur for a block.
++ */
++public abstract class BlockEvent {
```
Machine-Maker, line 24: `@ApiStatus.NonExtendable`

or just make it a concrete type that is sealed to inner `Configurable` type (or a better name).

patches/api/0384-Event-Effect-API-Improvements.patch
```diff
@@ -0,0 +1,775 @@
...
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/block/event/SimpleBlockEvent.java
+@@ -0,0 +1,12 @@
++package io.papermc.paper.block.event;
++
++import org.bukkit.Material;
++
++import java.util.Set;
++
++public class SimpleBlockEvent extends BlockEvent {
```
Machine-Maker, line 140: Is this needed? Just make BlockEvent concrete and sealed.

patches/api/0384-Event-Effect-API-Improvements.patch
```diff
@@ -0,0 +1,775 @@
...
+new file mode 100644
+index 0000000000000000000000000000000000000000..f3a0bde72d5e8b9edca46ea2f376359f8ae4b898
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/world/event/ConfigurableWorldEvent.java
+@@ -0,0 +1,18 @@
++package io.papermc.paper.world.event;
++
++import org.jetbrains.annotations.ApiStatus;
++
++public class ConfigurableWorldEvent<T> extends WorldEvent {
```
Machine-Maker, line 156: same issue here with `Configurable`

Machine-Maker (changes requested):

Owen1212055:
Should I separate this into fixes for entity effect api and then the world/block event api?

Machine-Maker:
Yeah, I think that would be best. One doesn't need new api, the other is a set of new API.

Owen1212055:
If someone wants to continue this, please feel free to take my patches. :)

PR #7878, opened 2022-06-06
Configurable chat thread limit

electronicboy:
By default, spigot shifts chat over to an unbounded thread pool,
on a normal server, this really offers no gains, the creation of a thread
on submitting to the pool on these servers eats more time vs just running it in
the netty pipeline, however, on servers using plugins which do work in here, there
could be some overall benefits to moving this stuff outside of the pipeline.

In general, this patch does two things:
1) Exposes the core size for the pool, this allows for ensuring that a number of threads
sit around in the pool, mitigating the need for creating new threads; This IS however
caveated, the ThreadPoolExecutor will ONLY create core threads as they're needed, it
just won't allow for us to dip back under the # of core threads, this can potentially
be mitigated by calling prestartCoreThread, however, I'm not sure if there is much justification
for this
2) Exposes a max size for the pool, as stated, by default, this is unbounded, for most
servers limiting the size of the pool is going to have 0 effects given how fast chat
is actually processed, this is honestly really just exposed for the misnomers or people
who just wanna ensure that this won't grow over a specific size if chat gets stupidly active

This is basically my own take on #7483

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -431,10 +431,10 @@ index 0000000000000000000000000000000000000000..c2dca89291361d60cbf160cab77749cb
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..456595e4b7e0c7f50617aa2694b0d2dfc368ab81
+index 0000000000000000000000000000000000000000..aaecd691922a28e971b859175574c80a330edb8e
 --- /dev/null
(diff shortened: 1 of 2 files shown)


kennytv (approved):

NoahvdAa:
Spigot no longer uses a thread pool for chat as of the [1.19.1 update](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/commits/21831450a43f67f1730dd0dec987c7a7ff19caa9#CraftBukkit-Patches/0029-Fix-Broken-Async-Chat.patch).

NoahvdAa:
Closed this too quickly, apologies. In 1.19.1, a thread pool is still used, but it was moved into CraftBukkit, into the `MinecraftServer` class.

PR #7888, opened 2022-06-08
Check if is light status or above / isLightOn / starlight version fits

Owen1212055:
This check was removed, so this would basically always try to read the light data.


Owen1212055:
It should be noted that minecraft does not check this anymore, but this is a check that starlight requires, for now, in order to prevent loading weird light data. This fix may want to be revised in the future.

PR #7893, opened 2022-06-08
Handle exception on convert to api for locate structure

Owen1212055:


Machine-Maker:
yeah.... idk if this is what we want. There are actually 2 missing structure types, and upstream specifically ignore the tests that they made to check this. I didn't add the api then because they are def gonna. and if they don't we can come back to it.

Owen1212055:
> yeah.... idk if this is what we want. There are actually 2 missing structure types, and upstream specifically ignore the tests that they made to check this. I didn't add the api then because they are def gonna. and if they don't we can come back to it.

Should I just skip structures not supported by the api then?

Machine-Maker (changes requested):
see my comment

Machine-Maker:
> Should I just skip structures not supported by the api then?

Skip them where? tests? the tests have the `@Ignore` annotation so they don't run. I just wouldn't do this at all until some time has passed and upstream hasn't done it.

Owen1212055:
image
See paper-contrib

Owen1212055:
See now

Machine-Maker:
Mmmm I think there's a better way. I'm refactoring PaperRegistry to support stuff like this in the future. to gracefully handle stuff like this in a more general fashion.

Owen1212055:
Alrighty

PR #7905, opened 2022-06-09
Reapply game event listener on update

Owen1212055:
Some block entities update their listener when load is called. This is bad as then when the block entity is
removed the listener will not be able to be removed. This properly removes and readds listeners when updating
block entities that have this special behavior.

See bug reports for easy testing steps.

Fixes:
- calling update() on these block entities breaks them
- trying to place a sensor/shrieker with nbt data will cause it to not work
- /data command breaks sensors/shriekers
- Underlying issue: https://bugs.mojang.com/browse/MC-252786


patches/server/0905-Reapply-game-event-listener-on-update.patch
```diff
@@ -0,0 +1,75 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
+index 9cce218623a71955687817a6317ed519f3300368..c7aa874a636f18072a4b0c373cc0573cf43ee75b 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
++++ b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
+@@ -757,7 +757,7 @@ public class LevelChunk extends ChunkAccess {
+         this.removeBlockEntityTicker(pos);
+     }
+ 
+-    private <T extends BlockEntity> void removeGameEventListener(T blockEntity, ServerLevel worldserver) {
++    public  <T extends BlockEntity> void removeGameEventListener(T blockEntity, ServerLevel worldserver) { // Paper
```
NoahvdAa: Are ATs still used or are we slowly moving away from them?

Owen1212055: I'd like feedback from other developers, but from my interpretation yes, we appear to be moving away from them.

I'm personally trying to only use them with stuff like fields, where it's just easy to remove the final and make them public.

kennytv: General gist: Use ATs for methods that are overridden and you need to cover a whole hierarchy or in high-conflict areas, otherwise just patching one method or field is easier

Owen1212055:
Rebased and updated description to show what this issue fixes.

PR #7920, opened 2022-06-10
improve packet config serializer

Machine-Maker:
Improves the packet and map serializers, allowing the map serializer to clean up extra keys if 2 keys deserialize to the same value, like `PacketPlayInAutoRecipe` and `ServerboundPlaceRecipePacket`. This also removes the need to handle conversions in the newFormatTranformations because its done in the serializer/map serializer now.

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -2191,36 +2191,49 @@ index 0000000000000000000000000000000000000000..f2f362883d1825084c277608c791f821
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/serializer/PacketClassSerializer.java b/src/main/java/io/papermc/paper/configuration/serializer/PacketClassSerializer.java
 new file mode 100644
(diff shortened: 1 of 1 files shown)


patches/server/0004-Paper-config-files.patch
```diff
@@ -2872,31 +2907,11 @@ index 0000000000000000000000000000000000000000..3be5cb126e41a533e9453bbd6326cd1c
...
-+                                final String[] split = classMapping.mojangName().split("\\.");
-+                                className = split[split.length - 1];
-+                                break;
-+                            }
-+                        }
-+
-+                        return path.with(path.size() - 1, className).array();
-+                    } else {
-+                        LOGGER.warn("Could not convert spigot-mapped packet class name {}", value);
-+                    }
```
kennytv, line 2892: Removing this will make it fail deserialization on reobfed servers again

kennytv (changes requested):

kennytv (approved):

PR #7934, opened 2022-06-10
Use configured log prefix for spigot library loading

Machine-Maker:


Owen1212055 (approved):
Diff generally looks good, can prolly be rebased and just merged into the paper plugin patch.

PR #7936, opened 2022-06-11

Machine-Maker:

Bukkit Changes:
c1d911a2 Add compat constructor to PlayerAnimationEvent

CraftBukkit Changes:
effea36a2 PR-1065: Add missing calls for BlockSpreadEvent when Sculk spreads

---

Is it worth deprecating our `PlayerArmSwingEvent`? Its point was to be able to differentiate between hands in the animation event but upstream went ahead and added another animation enum constant.


electronicboy (approved):

electronicboy:
I mean, I'm weary on duplications, especially as it's a duplicate event now, but in part I feel that ours is probably more explicit in terms of what's going on on the basis of the name, so, bleh

kennytv:
meh

PR #7943, opened 2022-06-11
Add WardenAngerChangeEvent

nopjar:
This introduces a new Event which is called when the warden is angered by another entity.

### DIFF
--- a/patches/server/0910-Add-WardenAngerChangeEvent.patch
+++ b/patches/server/0910-Add-WardenAngerChangeEvent.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: nopjar <code.nopjar@gmail.com>
+Date: Sun, 12 Jun 2022 02:26:04 +0200
+Subject: [PATCH] Add WardenAngerChangeEvent
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/monster/warden/AngerManagement.java b/src/main/java/net/minecraft/world/entity/monster/warden/AngerManagement.java
+index e9b4f29e1844f7c44e341f9b1c07c676469ca3b6..d76800a79faef26aab0cf99b28dfa4621877ecc7 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/warden/AngerManagement.java
++++ b/src/main/java/net/minecraft/world/entity/monster/warden/AngerManagement.java
+@@ -145,7 +145,7 @@ public class AngerManagement {
+     public int increaseAnger(Entity entity, int amount) {
+         boolean bl = !this.angerBySuspect.containsKey(entity);
+         int i = this.angerBySuspect.computeInt(entity, (suspect, anger) -> {
+-            return Math.min(150, (anger == null ? 0 : anger) + amount);
++            return Math.min(150, (anger == null ? 0 : anger) + amount); // Paper - diff on change
+         });
+         if (bl) {
+             int j = this.angerByUuid.removeInt(entity.getUUID());
+diff --git a/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java b/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java
+index 60410fd12e0fcfbaad3b4dc431452f8b2e6276c7..cd6fb81d14706217c92fccc7853afce4010dfd1b 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java
++++ b/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java
+@@ -361,7 +361,7 @@ public class Warden extends Monster implements VibrationListener.VibrationListen
+ 
+     @Override
+     public Brain<Warden> getBrain() {
+-        return super.getBrain();
++        return (Brain<Warden>) super.getBrain(); // Paper - decompile fix
+     }
+ 
+     @Override
+@@ -460,6 +460,15 @@ public class Warden extends Monster implements VibrationListener.VibrationListen
+     @VisibleForTesting
+     public void increaseAngerAt(@Nullable Entity entity, int amount, boolean listening) {
+         if (!this.isNoAi() && this.canTargetEntity(entity)) {
++            // Paper start
++            int activeAnger = this.angerManagement.getActiveAnger(entity);
++            io.papermc.paper.event.entity.WardenAngerChangeEvent event = new io.papermc.paper.event.entity.WardenAngerChangeEvent((org.bukkit.entity.Warden) this.getBukkitEntity(), entity.getBukkitEntity(), activeAnger, Math.min(150, activeAnger + amount));
++            this.level.getCraftServer().getPluginManager().callEvent(event);
++            if (event.isCancelled()) {
++                return;
++            }
++            amount = event.getNewAnger() - activeAnger;
++            // Paper end
+             WardenAi.setDigCooldown(this);
+             boolean bl = !(this.getBrain().getMemory(MemoryModuleType.ATTACK_TARGET).orElse((LivingEntity)null) instanceof Player);
+             int i = this.angerManagement.increaseAnger(entity, amount);
--- a/patches/api/0385-Add-WardenAngerChangeEvent.patch
+++ b/patches/api/0385-Add-WardenAngerChangeEvent.patch
@@ -1,0 +1,117 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: nopjar <code.nopjar@gmail.com>
+Date: Sun, 12 Jun 2022 00:56:45 +0200
+Subject: [PATCH] Add WardenAngerChangeEvent
+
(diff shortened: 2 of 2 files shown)


patches/server/0910-Add-WardenAngerEvent.patch
```diff
@@ -0,0 +1,62 @@
...
+=======
+index 60410fd12e0fcfbaad3b4dc431452f8b2e6276c7..feadfcddd11d89436ecfecd99ad70a66ef0322db 100644
+>>>>>>> Stashed changes
+--- a/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java
++++ b/src/main/java/net/minecraft/world/entity/monster/warden/Warden.java
+@@ -3,6 +3,7 @@ package net.minecraft.world.entity.monster.warden;
+ import com.google.common.annotations.VisibleForTesting;
+ import com.mojang.logging.LogUtils;
+ import com.mojang.serialization.Dynamic;
++import io.papermc.paper.event.entity.WardenAngerEvent;
```
NoahvdAa: This import is already inlined

patches/server/0910-Add-WardenAngerEvent.patch
```diff
@@ -0,0 +1,62 @@
...
++            int i = this.angerManagement.increaseAnger(entity, event.getNewAnger()); // Paper - was amount which would make event useless
+             if (entity instanceof Player && bl && AngerLevel.byAnger(i).isAngry()) {
+                 this.getBrain().eraseMemory(MemoryModuleType.ATTACK_TARGET);
+             }
+@@ -474,6 +483,8 @@ public class Warden extends Monster implements VibrationListener.VibrationListen
+ 
+     }
+ 
++
++
```
NoahvdAa: clean up these newlines

patches/api/0385-Add-WardenAngerEvent.patch
```diff
@@ -0,0 +1,99 @@
...
++    public int getNewAnger() {
++        return newAnger;
++    }
++
++    /**
++     * Sets the new anger resulting from this event.
++     *
++     * @param newAnger the new anger
++     * @see Warden#setAnger(Entity, int)
++     */
```
NoahvdAa: I'd personally change this to anger level instead of just anger (to be more consistent with Warden#getAnger), tiny nitpick tho.

NoahvdAa (changes requested):
Congrats on making your first PR, welcome to Paper! 🥳 

Looks good, apart from some very small comments below!

You seem to have some unsolved merge conflicts left over in your server patch.

nopjar:
I changed everything you addressed :)

patches/server/0910-Add-WardenAngerEvent.patch
```diff
@@ -0,0 +1,38 @@
...
+     @Override
+@@ -460,9 +460,17 @@ public class Warden extends Monster implements VibrationListener.VibrationListen
+     @VisibleForTesting
+     public void increaseAngerAt(@Nullable Entity entity, int amount, boolean listening) {
+         if (!this.isNoAi() && this.canTargetEntity(entity)) {
++            // Paper start
++            io.papermc.paper.event.entity.WardenAngerEvent event = new io.papermc.paper.event.entity.WardenAngerEvent((org.bukkit.entity.Warden) this.getBukkitEntity(), entity.getBukkitEntity(), this.angerManagement.getActiveAnger(entity) + amount);
++            this.level.getCraftServer().getPluginManager().callEvent(event);
++            if (event.isCancelled()) {
++                this.angerManagement.clearAnger(entity);
```
HexedHero: Could just be me but why clear it's angry? Why not just return? The WardenAngerEvent fires each time the Wardens anger value increases if we wanted to clear it wouldn't we just want to use setNewAnger(0) instead?

lynxplay: We already talked about that in #paper-contrib. Noah pointed out that, for the current layout of the event, it does make sense. The event only represents the warden being angered so you would expect cancellation to not have an angry warden.

Maybe it is worth actually moving this event to a WardenChangeAngerEvent (or something of the likes) ?
Would also make the fact that it fires more often a bit clearer.

HexedHero: I think a name change is best since the Warden can be "angry" but not actually do anything under 80 angry value.

nopjar: Jap, makes much more sense. I'll change it

nopjar:
I hope this event makes more sense now 😊

HexedHero:
What happens if you set the new anger to a negative number?

nopjar:
> What happens if you set the new anger to a negative number?

Not much. 
with `int i = this.angerManagement.increaseAnger(entity, amount);` could `i` result in a negative number, but as the AngerLevel is determined with the following:
```java
public static AngerLevel byAnger(int anger) {
    for(AngerLevel angerLevel : SORTED_LEVELS) {
        if (anger >= angerLevel.minimumAnger) {
            return angerLevel;
        }
    }

    return CALM;
}
```
A negative amount won't cause any issues.

EDIT:
And in the next tick the entry will get deleted from the map. See `net.minecraft.world.entity.monster.warden.AngerManagement.java#105`

NoahvdAa (approved):
Looks good, thanks for contributing!

dawon:
Wouldn't it make more sense to have LivingEntity in there? Or some non living entity can also anger the Warden?

nopjar:
> Wouldn't it make more sense to have LivingEntity in there? Or some non living entity can also anger the Warden?

I don't know if a warden is aware of non-living entities, but as the method `Warden#increaseAngerAt(Entity, int, boolean)` (in which the event is called) accepts an Entity instead of a LivingEntity, I would do the same so there are no ClassCastExceptions during runtime if mojang decides that a warden is aware of non-living entities.

dawon:
> > Wouldn't it make more sense to have LivingEntity in there? Or some non living entity can also anger the Warden?
> 
> I don't know if a warden is aware of non-living entities, but as the method `Warden#increaseAngerAt(Entity, int, boolean)` (in which the event is called) accepts an Entity instead of a LivingEntity, I would do the same so there are no ClassCastExceptions during runtime if mojang decides that a warden is aware of non-living entities.

Good point, was just an idea. :-)

nopjar:
But the event is only called if `Warden#canTargetEntity(Entity)` is true and this method returns false if the entity is not an instance of LivingEntity. So therefor a safe cast would be possible. I'm open-minded to change that if it makes more sense.

dawon:
> But the event is only called if `Warden#canTargetEntity(Entity)` is true and this method returns false if the entity is not an instance of LivingEntity. So therefor a safe cast would be possible. I'm open-minded to change that if it makes more sense.

In the end it won't make much difference either way. Either you or the plugin dev will make the instanceof check/cast :-)

nopjar:
> In the end it won't make much difference either way. Either you or the plugin dev will make the instanceof check/cast :-)

That's true. So we follow the safe way and are prepared for possible changes from mojang and use Entity, right? :)

dawon:
> > In the end it won't make much difference either way. Either you or the plugin dev will make the instanceof check/cast :-)
> 
> That's true. So we follow the safe and are prepared for possible changes from mojang and use Entity, right? :)

Not really up to me, I am just contributor as you are, I would stick with what you have and if paper team wants the change, they will let you know :-)

nopjar:
Just fixuped those 3 commits into one with appropriate naming as event name changed

kennytv (changes requested):
The anger is capped at 150; do that for what you pass into the event as well and add a `// Paper - diff on change` comment in the increaseAnger method where Math.min is used

nopjar:
@kennytv What should happen if the argument of `WardenAngerChangeEvent#setNewAnger(int)` is greater than 150? Should I set it to 150 or throw an IllegalArgumentException?

kennytv:
Throw the exception and note the allowed range in docs

patches/server/0910-Add-WardenAngerChangeEvent.patch
```diff
@@ -0,0 +1,48 @@
...
+     }
+ 
+     @Override
+@@ -460,6 +460,15 @@ public class Warden extends Monster implements VibrationListener.VibrationListen
+     @VisibleForTesting
+     public void increaseAngerAt(@Nullable Entity entity, int amount, boolean listening) {
+         if (!this.isNoAi() && this.canTargetEntity(entity)) {
++            // Paper start
++            int activeAnger = this.angerManagement.getActiveAnger(entity);
++            io.papermc.paper.event.entity.WardenAngerChangeEvent event = new io.papermc.paper.event.entity.WardenAngerChangeEvent((org.bukkit.entity.Warden) this.getBukkitEntity(), entity.getBukkitEntity(), activeAnger, activeAnger + amount);
```
kennytv: still needs to be min'd here

nopjar: oh, completely missed that one. Fixed in the new change

kennytv (approved):

PR #7954, opened 2022-06-12
Add option for strict advancement dimension checks

### LINKED ISSUE #7953: Subspace Bubble unlocks in all overworlds
### Expected behavior

The "Subspace Bubble" advancement should not work with all overworlds and/or nethers so as to be consistent with "Uneasy Alliance" and "Feels like home" advancements requiring the default "world" the Overworld to unlock.

As per issue #7809 there is only one Overworld (proper noun, spelled with a capital O, not a world that looks like it) and this advancement should only unlock when returning to `minecraft:overworld`.

### Observed/Actual behavior

The "Subspace Bubble" advancement unlocks when you enter any overworld dimension from any nether dimension 7km or more from where you entered.

### Steps/models to reproduce

Create a new overworld and/or nether dimension and use either in the process of getting the "Subspace Bubble" advancement.

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version git-Paper-5 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 4797e3a)

### Other

_No response_

Machine-Maker:
Craftbukkit attempts to translate worlds that use the same generation as the Overworld, The Nether, or The End to use those dimensions when checking the `changed_dimension` criteria trigger, or whether to trigger the `NETHER_TRAVEL` distance trigger. This adds a config option to ignore that and use the exact dimension key of the worlds involved.

Option in paper-global.yml: `misc.strict-advancement-dimension-check: false`

Closes https://github.com/PaperMC/Paper/issues/7953

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -419,7 +419,7 @@ new file mode 100644
 index 0000000000000000000000000000000000000000..3e7086d31b2f101b2d6e982f3935922886cadc77
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,264 @@
+@@ -0,0 +1,265 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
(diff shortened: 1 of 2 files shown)


kennytv (approved):
💩

MiniDigger:
needs docs pr, else looks good

PR #7960, opened 2022-06-13
Log IOException in plugin update

### LINKED ISSUE #7958: Papers plugin updater should log an exception if IO operations fail

As of right now, the plugin update logic uses a rather questionable spigot-made method to copy the updated plugin jar into the plugin folder: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/util/FileUtil.java#22

As a side effect of this, IOExceptions that occure while copying the updated plugin jar into the plugin folder are quietly ignored.
This makes error tracing a lot harder than it has to be.

While paper uses its own refined plugin update logic, introduced in #7687, this issue persists on paper.

There is not much need for the `FileUtil#copy` method, when [Files.copy](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/nio/file/Files.html#copy(java.nio.file.Path,java.nio.file.Path,java.nio.file.CopyOption...)) exists. The method throws an IOException, so the update logic should to catch said exception and properly log it to the console, allowing administrators to identify the error.

A server shutdown is not required, a single plugin failing should generally not fail the server.

None, this is a rather straight forward feature request.

### Other

_No response_

Doc94:
Yeah im a stalking... this handle to fix #7958 

This PR change the Bukkit util for copy things to the native java things and add a LOG when a issue is found, this not change the current behaviour where if the copy fails pass to another file.

Not sure if this is better o can be better add the log in the Bukkit util...

### DIFF
--- a/patches/api/0378-Update-Folder-Uses-Plugin-Name.patch
+++ b/patches/api/0378-Update-Folder-Uses-Plugin-Name.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Update Folder Uses Plugin Name
 
 
 diff --git a/src/main/java/org/bukkit/plugin/SimplePluginManager.java b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
-index 45114d587a8f201778adcba16c8a019f9959f472..ed07f5820281b139739f673fa4e25171de81b894 100644
+index 45114d587a8f201778adcba16c8a019f9959f472..5270e43c629fe63f42691d10c6f77dc1cc987457 100644
 --- a/src/main/java/org/bukkit/plugin/SimplePluginManager.java
 +++ b/src/main/java/org/bukkit/plugin/SimplePluginManager.java
 @@ -395,7 +395,7 @@ public final class SimplePluginManager implements PluginManager {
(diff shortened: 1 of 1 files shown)


jpenilla (dismissed):
Use proper logging templates and pass the exception instead of string concatenation. ie `LOGGER.info("Hello {}, an {} happened!", username, "exception", ex);`

Doc94:
> Use proper logging templates and pass the exception instead of string concatenation. ie `LOGGER.info("Hello {}, an {} happened!", username, "exception", ex);`

i based in another patchs and the whole file where this is the logger used, in the code i see like 3 types of logger.
- Bukkit.getLogger
- org.apache.logging.log4j.LogManager.getLogger()
- logger in plugin manager class used in another methods to show error in update plugin
image

What logger need to use?

kennytv:
The Bukkit logger is a dumb logger and doesnt support replacement parameters nicely, and since all the other usages in that class are the same, should be fine, meh

Doc94:
Rebased.

kennytv (approved):

PR #7968, opened 2022-06-13
Update settings directory path in exceptions

Lulu13022002:
The settings directory is now configurable using CLI options (since https://github.com/PaperMC/Paper/commit/8b6f4dd482dc5f152a433475f20055f7fbf1b182), but the exceptions message hasn't been updated properly to handle this new behaviour, and will show the default config folder even if the path has changed.

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -1241,11 +1241,11 @@ index 0000000000000000000000000000000000000000..1055a079a061d8761be13d3b9efbbd90
 +            final String legacyFileName = legacyConfig.getFileName().toString();
 +            try {
 +                if (Files.exists(configDir) && !Files.isDirectory(configDir)) {
-+                    throw new RuntimeException("Paper needs to create a '" + CONFIG_DIR + "' folder in the root of your server. You already have a non-directory named '" + CONFIG_DIR + "'. Please remove it and restart the server.");
++                    throw new RuntimeException("Paper needs to create a '" + configDir.getFileName() + "' folder in the root of your server. You already have a non-directory named '" + configDir.getFileName() + "'. Please remove it and restart the server.");
 +                }
 +                final Path backupDir = configDir.resolve(BACKUP_DIR);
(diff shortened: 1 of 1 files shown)


kennytv (approved):

patches/server/0004-Paper-config-files.patch
```diff
@@ -1200,11 +1200,11 @@ index 0000000000000000000000000000000000000000..518a00886f7bde1de95150366541fc99
 +        if (needsConverting(legacyConfig)) {
 +            try {
 +                if (Files.exists(configDir) && !Files.isDirectory(configDir)) {
-+                    throw new RuntimeException("Paper needs to create a '" + CONFIG_DIR + "' folder in the root of your server. You already have a non-directory named '" + CONFIG_DIR + "'. Please remove it and restart the server.");
++                    throw new RuntimeException("Paper needs to create a '" + configDir.getFileName() + "' folder in the root of your server. You already have a non-directory named '" + configDir.getFileName() + "'. Please remove it and restart the server.");
```
jpenilla: I don't think the config folder will necessarily be in the server root anymore?

jpenilla (dismissed):

patches/server/0004-Paper-config-files.patch
Lulu13022002: I don't know this option works like other file settings (paper-settings etc..) and use joptsimple. I never test if input like ../dir or absolute path directly are allowed or not. We can choose to restrict the folder to the server root or just allow both ways and change the exception message depending of where is the final path ?

Lulu13022002: After some tests, all outside paths are valid so maybe we can change this message ?

Machine-Maker: I think you can probably just do configDir.toAbsolutePath()? That should create the full exact path to the file.

Lulu13022002: Yes normally fixed in the latest changes

Lulu13022002:
Rebased and finally manage to reset the author of the commits

Issue #7852, opened 2022-05-27
Pistons does not work correctly under alternate-current redstone system


siulung201314:
### Expected behavior

Piston retracts when there is no redstone signal from redstone dust.

### Observed/Actual behavior

Piston cannot retract when there is no redstone signal from redstone dust.

### Steps/models to reproduce

1. Build what the photos show.
2. Turn on the lever. The piston extends
3. Turn off the lever. The piston does not retract.

![image](https://user-images.githubusercontent.com/83176041/170785195-d03187a1-6844-486e-80ed-ab0fe82ab51b.png)
![image](https://user-images.githubusercontent.com/83176041/170785210-10ac8590-ce20-44bb-8298-47b6bc2533eb.png)

### Plugin and Datapack List

without any plugins

### Paper version

This server is running Paper version git-Paper-350 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 3f7fbe3)
You are running the latest version
Previous version: git-Purpur-1628 (MC: 1.18.2)

### Other

1. (Using eigencraft), there is no any bug about piston.

2. (Using alternate-current) I place a block near the extended piston without redstone signal, the piston will then retract.
![image](https://user-images.githubusercontent.com/83176041/170786471-13d54df5-6ddf-4332-a394-c5fd099de16a.png)
![image](https://user-images.githubusercontent.com/83176041/170786510-c3a947c3-7468-46e0-892c-03fe5a422b8a.png)

2. (Using alternate-current) If two pistons stick together, they can retract without redstone signal.
![image](https://user-images.githubusercontent.com/83176041/170785566-f94dfaf7-93b4-4634-9fe8-cce8cc535cd8.png)
![image](https://user-images.githubusercontent.com/83176041/170785582-027c68da-80bb-4ec9-8b6a-d89ddaae20ba.png)

3. (Using alternate-current) If the end of the redstone wire is on the block above the piston, piston can retract without redstone signal.
![image](https://user-images.githubusercontent.com/83176041/170790703-d372bbfa-265d-41dc-be6c-e6ce164734a5.png)
![image](https://user-images.githubusercontent.com/83176041/170790724-01249b1e-55cd-4deb-8bed-99ca5b57df78.png)

4. (Using alternate-current) Replace the piston with a redstone lamp, the redstone lamp can be inactive without redstone signal.

molor:
[...]

PR #7971, opened 2022-06-13
Update Alternate Current patch (1.18.2)

SpaceWalkerRS:
- update to v1.3, fixing some Vanilla parity issues (like #7852)


kennytv (approved):
Thanks!

PR #7980, opened 2022-06-14
Damage with any cause

TetraTheta:
Continuation of #7423.

Since this is almost direct port of original(#5191) PR, I don't think I fully understood the original code. I will leave some comment on the code.

Any comments will be welcomed, so please, feel free to give me any feedback.

### DIFF
--- a/patches/server/0910-Damage-with-any-cause.patch
+++ b/patches/server/0910-Damage-with-any-cause.patch
@@ -1,0 +1,483 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TetraTheta <tetratheta@gmail.com>
+Date: Tue, 14 Jun 2022 20:39:50 +0900
+Subject: [PATCH] Damage with any cause
+
+
+diff --git a/src/main/java/io/papermc/paper/util/DamageUtil.java b/src/main/java/io/papermc/paper/util/DamageUtil.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..a914c6ec22bd4532aad60567616d0f9a132a27f7
+--- /dev/null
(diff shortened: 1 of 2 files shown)


patches/server/0910-Damage-with-any-cause.patch
```diff
@@ -0,0 +1,483 @@
...
+-            CraftEventFactory.blockDamage = damager; // SPIGOT-6639: Re-set blockDamage so that other entities which are also getting damaged have the right cause
++            CraftEventFactory.blockDamage = blockDamager; // SPIGOT-6639: Re-set blockDamage so that other entities which are also getting damaged have the right cause // Paper
+ 
+             if (!event.isCancelled()) {
+                 event.getEntity().setLastDamageCause(event);
+@@ -1078,6 +1087,8 @@ public class CraftEventFactory {
+                 throw new IllegalStateException(String.format("Unhandled damage of %s by %s from %s", entity, damager.getHandle(), source.msgId));
+             }
+             EntityDamageEvent event = new EntityDamageByEntityEvent(damager, entity.getBukkitEntity(), cause, modifiers, modifierFunctions, source.isCritical()); // Paper - add critical damage API
++            */
```
TetraTheta, line 432: These lines were undeleted by purpose. I wasn't sure if I can safely delete these lines. If it is ok to delete this lines, I will.

TetraTheta:
Closing mine because https://github.com/PaperMC/Paper/pull/8058 is much more 'beautiful'  
Even though mentioned PR is still draft, I think it is more productive. Mine was just 'lazy' 😄 

I think no one would think this PR should be 'revived', but if anyone thinks so, feel free to re-open this. (Don't know if it is possible though)

PR #7995, opened 2022-06-16
Deprecate Spigot's HumanEntity#getItemInUse, mark LivingEntity#getActiveItem as NotNull

Owen1212055:
LivingEntity#getActiveItem is marked as nullable incorrectly. Additionally, this is duplicate api but spigots is on HumanEntity so deprecate it in favor of paper's method.

Paper (not null, marked incorrectly as nullable) LivingEntity
image
Spigot (nullable) HumanEntity
image


electronicboy (approved):

Owen1212055:
Rebased

PR #8002, opened 2022-06-16
Add Disable Vibrations Config

Owen1212055:
Adds an option to disable vibration listeners from being registered/playing in the world.
This causes the following to disable:
- Sculk Catalyst spreading on death
- Sculk Shrieker 
- Sculk Sensor
- Warden vibration listening (they can still attack players if nearby/provoked)
- Allay note block behavior


kennytv (changes requested):
As you said yourself, you're disabling more than just vibrations; please move it to the proper place so this doesn't break things other than vibrations (or people's datapacks), especially in the future, or rename the option

Machine-Maker:
Would this better be handled by a plugin with an event? Do we need a paper config to disable them?

Owen1212055:
~~Closed after conversation in paper contributing, plugins can handle this.~~

Owen1212055:
Renamed. It was decided to keep this incase server owners want to prevent the registering/sending of game events all together.

PR #8008, opened 2022-06-17
Correctly handle interactions with items on cooldown

### LINKED ISSUE #8005: Can't open doors while goat horn is cooling down
### Expected behavior

I expected the door to open and stay open.

### Observed/Actual behavior

The door opens momentarily before closing again.

### Steps/models to reproduce

Use a goat horn, while it's cooling down you can't interact with doors.

[Video](https://www.youtube.com/watch?v=Jc6cmLlKf98)

### Plugin and Datapack List

None.

### Paper version

This server is running Paper version git-Paper-24 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: dd61319)

### Other

_No response_

Machine-Maker:
This is a potential minefield, I tested the following situations:
* No PlayerInteractEvent changes: items on cooldown no longer prevent block interactions (can open a door)
* `PlayerInteractEvent.setUseItemInHand(DENY)`: block interactions don't happen when using the item and pressing shift

Fixes https://github.com/PaperMC/Paper/issues/8005

### DIFF
--- a/patches/server/0924-Correctly-handle-interactions-with-items-on-cooldown.patch
+++ b/patches/server/0924-Correctly-handle-interactions-with-items-on-cooldown.patch
@@ -1,0 +1,60 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 16 Jun 2022 21:57:02 -0700
+Subject: [PATCH] Correctly handle interactions with items on cooldown
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

electronicboy:
This is somewhat of a behavioral change in regards to the API, so this is bleh all around, but, I can't really think of any sane alternatives here, so, approved (Also, note that I ci-skip'd the dupe patch numbers)

Machine-Maker:
Well it was a significant behavioral break from vanilla. You couldn't open a door with an item that was on cooldown, which works totally fine in vanilla. Firing a PlayerInteractEvent with useClickedBlock set to DENY isn't new, it just shouldn't happen when the item in hand is on cooldown. The interaction isn't cancelled just cause an item is on cooldown, the useItemInHand is

electronicboy:
Yea, that was my sthinking, the way that the event was being fired is just wrong, hence the fix; biggest minefield to me is at least the API side, but, I mean, I can't see any plugin expecting this to be the case, and there's no real other fix for this, so, it's got my vote

Machine-Maker:
Rebased for 1.19.2

kennytv (approved):

PR #8023, opened 2022-06-20
Call StructureGrowEvent for mangrove propagule

NoahvdAa:
Fixes #8022. Event capturing code is the exact same as in `SaplingBlock#randomTick`.

PR #8024, opened 2022-06-21
Handle growth.sapling for spigot config in Mangrove Propagule

Doc94:
Based in https://github.com/PaperMC/Paper/pull/8023#issuecomment-1160501192 this PR handle the spigot config for saplings in the mangrove.

### DIFF
--- a/patches/server/0918-Handle-growth.sapling-for-spigot-config-in-Mangrove-.patch
+++ b/patches/server/0918-Handle-growth.sapling-for-spigot-config-in-Mangrove-.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Mon, 20 Jun 2022 22:35:49 -0400
+Subject: [PATCH] Handle growth.sapling for spigot config in Mangrove Propagule
+
+
+diff --git a/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java b/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
+index 4eaf75ef7362ee781762f5e8def6bc4d986e9db7..ea1ed2e05f88ab591db45db6c1f8e51eaab4f59a 100644
+--- a/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
+@@ -89,7 +89,7 @@ public class MangrovePropaguleBlock extends SaplingBlock implements SimpleWaterl
+     @Override
+     public void randomTick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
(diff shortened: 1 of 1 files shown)


patches/server/0918-Handle-growth.sapling-for-spigot-config-in-Mangrove-.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java b/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
+index 4eaf75ef7362ee781762f5e8def6bc4d986e9db7..ea1ed2e05f88ab591db45db6c1f8e51eaab4f59a 100644
+--- a/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/MangrovePropaguleBlock.java
+@@ -89,7 +89,7 @@ public class MangrovePropaguleBlock extends SaplingBlock implements SimpleWaterl
+     @Override
+     public void randomTick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
+         if (!isHanging(state)) {
+-            if (random.nextInt(7) == 0) {
++            if (random.nextInt(Math.max(2, (int) (((100.0F / world.spigotConfig.saplingModifier) * 7) + 0.5F))) == 0) { // Paper - handle spigot config for saplingModifier
```
Machine-Maker, line 16: I think this is wrong. If I set the sapling modifier to 200 (200%), it becomes a `random.nextInt(4) == 0` or a 1/4 chance. If its left to default at 100 it is a `random.nextInt(7) == 0` or a 1/7 chance. 1/4 is not twice as likely as 1/7.

Machine-Maker (changes requested):

patches/server/0918-Handle-growth.sapling-for-spigot-config-in-Mangrove-.patch
Doc94: hmmm i based this in all the other place where this config is set then not sure if im understanding wrong or maybe all the places where this is used is bad calculed in first place :/

Lulu13022002: Not all the places but some has the same issue, maybe using float is better (and more readable: no abusive int cast) like in the kelp patch 0796-Fix-kelp-modifier-changing-growth-for-other-crops.patch.
For example
if(random.nextFloat() < Math.max((1/2), ((1/7) * (modifier / 100.0F)))

Owen1212055:
Are you still interested in resolving the comment marked above? Just want to make sure incase somebody else wants to pick this up. 😃

Doc94:
> Are you still interested in resolving the comment marked above? Just want to make sure incase somebody else wants to pick this up. 😃

you mean change the calc by the suggest by lulu? i remember not touch this because i was confuse about the correct calc (and dont wanna rebase and remake for every try)

PR #8027, opened 2022-06-21
Call PlayerLaunchProjectileEvent for trident

### LINKED ISSUE #7789: Trident doesn't trigger PlayerLaunchProjectileEvent
### Expected behavior

When I charge and release a trident I would expect the event to be triggered.

### Observed/Actual behavior

The event doesn't trigger when a player launches a trident.

### Steps/models to reproduce

- Charge trident
- Release trident
- No event

### Plugin and Datapack List

Custom plugin

### Paper version

This server is running Paper version git-Paper-308 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: https://github.com/PaperMC/Paper/commit/5b85ee37e20de15e0798391085b01229e4e0d6fd)
You are 2 version(s) behind
Previous version: git-Paper-301 (MC: 1.18.2)

### Other

ProjectileLaunchEvent does trigger when throwing a trident as a player.
However, to listen in on that event for all projectiles is not that efficient when there is a PlayerLaunchProjectileEvent.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7789
Call this event also for trident to avoid listening generic ProjectileLaunchEvent in the spawn area.
Currently i'm not really sure if the souldConsume method should do something or not
Also i think we should document the arrow case for this event (not triggered) and redirect to EntityShootBowEvent ?

### DIFF
--- a/patches/api/0126-PlayerLaunchProjectileEvent.patch
+++ b/patches/api/0126-PlayerLaunchProjectileEvent.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] PlayerLaunchProjectileEvent
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerLaunchProjectileEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerLaunchProjectileEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..13a3d0c75c337e8be105d92c4d08a6cd7f783474
+index 0000000000000000000000000000000000000000..9074b2ede01f76c0560e5318246382163cc91591
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerLaunchProjectileEvent.java
-@@ -0,0 +1,85 @@
+@@ -0,0 +1,83 @@
 +package com.destroystokyo.paper.event.player;
 +
 +import org.bukkit.entity.Player;
@@ -58,7 +58,6 @@ index 0000000000000000000000000000000000000000..13a3d0c75c337e8be105d92c4d08a6cd
 +
 +    /**
 +     * Get whether to consume the ItemStack or not
-+     * Note: this does nothing for reusable and throwable item like the trident
 +     *
 +     * @return True to consume
 +     */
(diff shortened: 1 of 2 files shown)


patches/api/0126-PlayerLaunchProjectileEvent.patch
```diff
@@ -67,6 +68,7 @@ index 0000000000000000000000000000000000000000..9074b2ede01f76c0560e531824638216
 +
 +    /**
 +     * Set whether to consume the ItemStack or not
++     * Note: this does nothing for reusable and throwable item like the trident
```
lynxplay: Is there any reason we cannot actually implement this ?
I mean, the item removal logic can technically just be skipped if the flag is set no ?

Lulu13022002: Yes i forgot why i have made that, but i think the main idea was avoid a duplicate item in survival but since it's plugin anyway, this can be cool for mini game. will implement that

lynxplay: Yea, I figured that was the concept but I agree with you here. "Should consume" set to false directly means duplication tbh.
Yea it is an item here, it might as well be xp from bottles.

Lulu13022002: The only question left is should we avoid the hurt damage on the item when this is disabled or not

lynxplay: Well, this does indeed pose an interesting question.
I'd presume we would want to leave the item undamaged as this is as close as possible to the "just don't consume whatever was used, only launch" mentality at least I think of in regards to the event.

davidmayr:
Any updates on this?

electronicboy (approved):

lynxplay (approved):

lynxplay:
Do you wanna add the javadocs in regards to arrows from bows and crossbows ? Won't stop the merge, I'll rebase and final test it tomorrow if no one else comes first, but I think that would be a good addition.

PR #8028, opened 2022-06-21
Update Alternate Current patch

SpaceWalkerRS:
Update the Alternate Current patch to v1.4.0

- New optimizations bring a 10%-20% performance boost.
- Several code modifications have been removed to make the mod less invasive.


kennytv (approved):
Thanks!

PR #8037, opened 2022-06-22
Add fire-tick-delay option

### LINKED ISSUE #7277: Fire tick speed

vytskalt:
Closes #7277

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -50,4 +50,5 @@ LemonCaramel <admin@caramel.moe>
 Noah van der Aa <ndvdaa@gmail.com>
 Doc <nachito94@msn.com>
 Nick Hensel <nickhensel25@icloud.com>
+vytskalt <vytskalt@protonmail.com>
 ```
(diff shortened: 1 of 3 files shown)


patches/server/0918-Add-fire-tick-delay-option.patch
```diff
@@ -0,0 +1,53 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+index 32f0cd29d1198fe320d10ccfe0b02f8632ac12aa..4b5372487e9a905d874da9c095351318988e118b 100644
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+@@ -334,6 +334,7 @@ public class WorldConfiguration extends ConfigurationPart {
+             public BooleanOrDefault findAlreadyDiscoveredLootTable = BooleanOrDefault.USE_DEFAULT;
+         }
+ 
++        public int fireTickDelay = 30;
```
electronicboy: This belongs in the config files patch

electronicboy (changes requested):

electronicboy (approved):

Owen1212055 (approved):
To be honest, it would probably be less of a diff if you just modify the pre-existing method and add a level param. 

But looks good here.

vytskalt:
> To be honest, it would probably be less of a diff if you just modify the pre-existing method and add a level param.

Done

ByteBeamer:
Thanks so much for this!

PR #8041, opened 2022-06-23
Add EntityFertilizeEggEvent

### LINKED ISSUE #8040: EntityBreedEvent does not fire for turtles
### Expected behavior

When breeding a turtle, org.bukkit.event.entity.EntityBreedEvent should be called as it is for all other animals at the time of breeding.

### Observed/Actual behavior

The event is never triggered when breeding turtles.

### Steps/models to reproduce

Use this snippet and try to breed turtles.
```
@EventHandler
  public void onBreed(EntityBreedEvent e){
    e.setCancelled(true);
    Bukkit.broadcastMessage("Breed");
  }
```

### Plugin and Datapack List

This can be replicated with no plugins or datapacks installed.

### Paper version

This server is running Paper version git-Paper-34 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: e52499e)
Previous version: git-Paper-33 (MC: 1.19)

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/8040
Add EntityFertilizeEggEvent, the super entity is the mother and doesn't extends EntityBreedEvent at all
Also plugin can set negative or zero experience to totally ignore experience drops ~unlike EntityBreedEvent (which is strange because there are also a negative check in the nms)~


patches/api/0387-Add-EntityMateEvent.patch
```diff
@@ -0,0 +1,141 @@
...
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Called when a Turtle or a Frog meets a new mate.
++ */
++public class EntityMateEvent extends EntityEvent implements Cancellable {
++
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker:
```diff
-    private static final HandlerList handlers = new HandlerList();
+    private static final HandlerList HANDLER_LIST = new HandlerList();
```

patches/api/0387-Add-EntityMateEvent.patch
```diff
@@ -0,0 +1,141 @@
...
++        Preconditions.checkArgument(mother != null, "Cannot have null mother");
++        Preconditions.checkArgument(father != null, "Cannot have null father");
++
++        // Breeder can be null in the case of spontaneous conception
++        this.mother = mother;
++        this.father = father;
++        this.breeder = breeder;
++        this.bredWith = bredWith;
++
++        setExperience(experience);
```
Machine-Maker: Prefer using direct field assignment here instead of calling a set method

patches/api/0387-Add-EntityMateEvent.patch
```diff
@@ -0,0 +1,141 @@
...
++import org.bukkit.entity.Player;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.entity.EntityEvent;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Called when a Turtle or a Frog meets a new mate.
```
Machine-Maker: I would generalize a bit, and say something like "Called when a mating occurs that results in a pregnancy". Since this is a different breeding mechanic to other mobs that produce offspring.

patches/api/0387-Add-EntityMateEvent.patch
```diff
@@ -0,0 +1,141 @@
...
++ */
++public class EntityMateEvent extends EntityEvent implements Cancellable {
```
Machine-Maker: I still think EntityMateEvent and EntityBreedEvent are so similar sounding. Is there a better name we could come up with that refects the difference in mechanics? EntityBreedEvent produces an entity offspring immediately while this causes one of the parent entities to take some action later.

Machine-Maker (changes requested):

patches/api/0387-Add-EntityMateEvent.patch
Lulu13022002: I think based on what you've changed on the description we can reuse the word pregnancy for the event name (means that an action will occur later but not now) ?
But i agree with you the two names are really similar it's why i have decided to say which entity can trigger this event at the starting point

Machine-Maker: Yeah, I'm not sure. possible words: impregnate, conceive, fertilize. I don't know which one is the best, open to more suggestions.

Lulu13022002: The word fertilize is already used by BlockFertilizeEvent and i think we should avoid to repeat the same name (for two different things) to avoid confusing people. But the other words looks good i let you the choice since i'm better reading english than writing (and have never seen the conceive word)

mfcfbro: Why not something like EntityFertalizeEggEvent since that's fairly descriptive of what's happening? As far as I could imagine this could cover other animals that may have this mechanic in the future.

Lulu13022002: Yes i like this idea very good to know directly which mechanics that refer

jpenilla: Fertalize is not a word, you want Fertilize like you had before.

Lulu13022002: Thank you, renamed

Owen1212055 (approved):
Looks good, but I would explain that this occurs with entities that usually lay eggs. As it's hard to tell which entities this will occur from just by the documentation.

Machine-Maker (approved):

PR #8045, opened 2022-06-24
Add xp orb spawn reason for frogs

Lulu13022002:
Track the spawn reason of the xp orb dropped when frogs breeds

### DIFF
--- a/patches/server/0125-ExperienceOrbs-API-for-Reason-Source-Triggering-play.patch
+++ b/patches/server/0125-ExperienceOrbs-API-for-Reason-Source-Triggering-play.patch
@@ -21,10 +21,10 @@ index cd3b4c97374d44f5a0e710e03f4ac38938757e25..b18bb06caf5f034dffbb72120c8f21da
  
                  return true;
 diff --git a/src/main/java/net/minecraft/world/entity/ExperienceOrb.java b/src/main/java/net/minecraft/world/entity/ExperienceOrb.java
-index b3433ce9c722bdab81848a6c2d121ca510c48509..154e86ea88425c24683df11772a677267e6498b4 100644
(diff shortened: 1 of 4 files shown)


Owen1212055 (approved):

Machine-Maker (approved):
I think this patch should mark the ExperienceOrb ctor that doesn't take a SpawnReason as deprecated. Just to increase the likelihood
that usages of it are caught earlier.

Lulu13022002:
I can deprecate the default constructor if you wants ? Also there is a decompile issue in the getBrain method return which needs to be cast into a Brain\<Frog\> before merging this one and the new event

Machine-Maker:
Yeah, add the `@Deprecated` to that constructor.

kennytv (approved):

patches/server/0125-ExperienceOrbs-API-for-Reason-Source-Triggering-play.patch
```diff
@@ -67,6 +67,7 @@ index b3433ce9c722bdab81848a6c2d121ca510c48509..227aca795efc99c4f81dfb30c00d31d2
 +        nbttagcompound.put("Paper.ExpData", comp);
 +    }
  
++    @Deprecated
```
Machine-Maker, line 71: Now add the `@DoNotUse` annotation to it and remove the `@Deprecated` annotation. That annotation will cause there to be build errors if the annotated method or constructor is used anywhere in the source.

This was just added the other week.

Machine-Maker (changes requested):

patches/server/0125-ExperienceOrbs-API-for-Reason-Source-Triggering-play.patch
Lulu13022002: Fixed i have also cast the brain frog to resolve the build error

Machine-Maker (approved):

electronicboy (approved):

PR #8049, opened 2022-06-25
Add MaterialTags for the different Types of Tools

thelooter:
Adds MaterialTags for the following Types of Tools:

- Wooden Tools
- Stone Tools
- Iron Tools
- Gold Tools
- Diamond Tools
- Netherite Tools

### DIFF
--- a/patches/api/0156-Add-Material-Tags.patch
+++ b/patches/api/0156-Add-Material-Tags.patch
@@ -117,7 +117,7 @@ new file mode 100644
 index 0000000000000000000000000000000000000000..1a78872e26f8fadbddd9af15fff063d03690077f
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/MaterialTags.java
-@@ -0,0 +1,696 @@
+@@ -0,0 +1,704 @@
 +/*
 + * Copyright (c) 2018 Daniel Ennis (Aikar) MIT License
 + *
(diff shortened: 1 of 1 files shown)


imDaniX:
Just by tags names, it's actually a bit unclear what's included into those - we also have items like shears, flint'n'steel, clock and compass, bucket.

thelooter:
What Names would you suggest instead?

cody-quinn:
I disagree in the names being unclear. I feel the sheers, flint n steel, click, compass and bucket fit into their own category independent of the main tiered tools.

patches/api/0157-Add-Material-Tags.patch
```diff
@@ -777,6 +778,41 @@ index 0000000000000000000000000000000000000000..728bc33cd6081f9462ab62d2ca693557
...
++        .add(Material.GOLDEN_AXE, Material.GOLDEN_HOE, Material.GOLDEN_PICKAXE, Material.GOLDEN_SHOVEL, Material.GOLDEN_SWORD)
++        .ensureSize("GOLDEN_TOOLS", 5).lock();
++
++    /**
++     * Covers all Diamond Tools.
++     */
++    public static final MaterialSetTag DIAMOND_TOOLS = new MaterialSetTag(keyFor("diamond_tools"))
++        .add(Material.DIAMOND_AXE, Material.DIAMOND_HOE, Material.DIAMOND_PICKAXE, Material.DIAMOND_SHOVEL, Material.DIAMOND_SWORD)
++        .ensureSize("DIAMOND_TOOLS", 5).lock();
 +}
```
cody-quinn: Where is Netherite

cody-quinn (changes requested):
Seems that netherite was ignored.

Could also be cool to include armor sets?

imDaniX:
> I disagree in the names being unclear. I feel the sheers, flint n steel, click, compass and bucket fit into their own category independent of the main tiered tools.

I mean, while bucket, compass and clock don't really fit there, shears and flint'n'steel absolutely do - both are indeed tools, made using iron, both do have durability and can be enchanted. If we don't include these in iron tools, at least is should be noted in javadocs.

cody-quinn:
> > I disagree in the names being unclear. I feel the sheers, flint n steel, click, compass and bucket fit into their own category independent of the main tiered tools.
> 
> I mean, while bucket, compass and clock don't really fit there, shears and flint'n'steel absolutely do - both are indeed tools, made using iron, both do have durability and can be enchanted. If we don't include these in iron tools, at least is should be noted in javadocs.

I can agree with the information about them being exlucded should be included in the JavaDocs. However I still don't think they should be included in the collection itself as they are not tiered. The naming of wooden, iron, exc. makes me assume the collections are refering to the tiered tools where diamond will include all the diamond equivalents of the iron tools.

Maybe an `UNTIERED_TOOLS` or `OTHER_TOOLS` collection could be made for those?

patches/api/0157-Add-Material-Tags.patch
thelooter: Should be added now

cody-quinn (approved):
LGTM :+1:

NoahvdAa (approved):
looks good from web

Owen1212055 (approved):
In the future, I recommend creating PRS on other branches (not your master branch, because it makes some things difficult) 

Looks good tho

thelooter:
> In the future, I recommend creating PRS on other branches (not your master branch, because it makes some things difficult)
> 
> Looks good tho

Yeah I usually do. Only noticed I was on master after opening the PR.

thelooter:
Any update on this?

electronicboy (approved):

PR #8061, opened 2022-06-27
Use Worldheight for Activation Ranges

Owen1212055:


patches/server/0349-Entity-Activation-Range-2.0.patch
```diff
@@ -498,19 +498,30 @@ index 7bae24598218dcf0012dd21e619e6f5f984bd6f0..c9a032c5331a918453de5e8c6a6d13f5
...
+-            ActivationRange.maxBB = player.getBoundingBox().inflate( maxRange, 256, maxRange );
+-            ActivationType.MISC.boundingBox = player.getBoundingBox().inflate( miscActivationRange, 256, miscActivationRange );
+-            ActivationType.RAIDER.boundingBox = player.getBoundingBox().inflate( raiderActivationRange, 256, raiderActivationRange );
+-            ActivationType.ANIMAL.boundingBox = player.getBoundingBox().inflate( animalActivationRange, 256, animalActivationRange );
+-            ActivationType.MONSTER.boundingBox = player.getBoundingBox().inflate( monsterActivationRange, 256, monsterActivationRange );
 +            // Paper start
-+            ActivationType.WATER.boundingBox = player.getBoundingBox().inflate( waterActivationRange, 256, waterActivationRange );
-+            ActivationType.FLYING_MONSTER.boundingBox = player.getBoundingBox().inflate( flyingActivationRange, 256, flyingActivationRange );
-+            ActivationType.VILLAGER.boundingBox = player.getBoundingBox().inflate( villagerActivationRange, 256, villagerActivationRange );
++            int worldHeight = world.getMaxBuildHeight();
```
kennytv: height, not max build height

kennytv (changes requested):

patches/server/0349-Entity-Activation-Range-2.0.patch
Owen1212055: Thanks!

kennytv (approved):

patches/server/0349-Entity-Activation-Range-2.0.patch
```diff
@@ -498,19 +498,30 @@ index 7bae24598218dcf0012dd21e619e6f5f984bd6f0..c9a032c5331a918453de5e8c6a6d13f5
...
++            int worldHeight = world.getHeight();
++            ActivationRange.maxBB = player.getBoundingBox().inflate( maxRange, worldHeight, maxRange );
++            ActivationType.MISC.boundingBox = player.getBoundingBox().inflate( miscActivationRange, worldHeight, miscActivationRange );
++            ActivationType.RAIDER.boundingBox = player.getBoundingBox().inflate( raiderActivationRange, worldHeight, raiderActivationRange );
++            ActivationType.ANIMAL.boundingBox = player.getBoundingBox().inflate( animalActivationRange, worldHeight, animalActivationRange );
++            ActivationType.MONSTER.boundingBox = player.getBoundingBox().inflate( monsterActivationRange, worldHeight, monsterActivationRange );
++            ActivationType.WATER.boundingBox = player.getBoundingBox().inflate( waterActivationRange, worldHeight, waterActivationRange );
++            ActivationType.FLYING_MONSTER.boundingBox = player.getBoundingBox().inflate( flyingActivationRange, worldHeight, flyingActivationRange );
++            ActivationType.VILLAGER.boundingBox = player.getBoundingBox().inflate( villagerActivationRange, worldHeight, villagerActivationRange );
```
Lulu13022002: This will trigger active tick for a too large aabb in the negative y because inflate expand the box both in positive and negative, i think the box needs to be expanded with:
[player.aabb.minX - range, world.minHeight, player.aabb.minZ - range,
 player.aabb.maxX + range, world.maxHeight, player.aabb.maxZ + range]
But this should happens only if the player is in the y limit for outside player the y size must be infinite with possibly an option to just ignore outside player and keep the same aabb as above
Another solution is to create a new configuration option to choose the y offset (here: worldHeight) and by default put the delta / 2 => (maxHeight - minHeight) / 2

Owen1212055: I'm not sure, the behavior has been this way. The height that picks here is the total height of the world, including the min height.

Height: 384, MinHeight: -64 

Therefore, the height is "technically" 320. However, since this bounding box is higher it will properly expand 384 UP and 384 down. This is the behavior of before, where it will expand 255 UP and 355 DOWN.

electronicboy (approved):

Owen1212055:
Rebased

PR #8060, opened 2022-06-27
Add api for disabling conduits attacking entities

u9g:
Example plugin if curious:
```java
package io.papermc.paper.testplugin;

import net.kyori.adventure.text.minimessage.MiniMessage;
import org.bukkit.Bukkit;
import org.bukkit.block.Conduit;
import org.bukkit.event.EventHandler;
import org.bukkit.event.Listener;
import org.bukkit.event.block.BlockPlaceEvent;
import org.bukkit.event.entity.EntityDeathEvent;
import org.bukkit.inventory.EquipmentSlot;
import org.bukkit.plugin.java.JavaPlugin;
import org.bukkit.scheduler.BukkitRunnable;

public final class TestPlugin extends JavaPlugin implements Listener {
    @Override
    public void onEnable() {
        this.getServer().getPluginManager().registerEvents(this, this);
    }
    @EventHandler
    public void onA(BlockPlaceEvent e) {
        var loc = e.getBlock().getLocation();
        new BukkitRunnable() {
            @Override
            public void run() {
                if (e.getHand() == EquipmentSlot.OFF_HAND) {
                    Bukkit.broadcast(MiniMessage.miniMessage().deserialize("not disabling canAttackEntities on conduit bc placed with offhand"));
                    return;
                }
                if (loc.getBlock().getState(false) instanceof Conduit c) {
                    c.setCanAttackEntities(false);
                    Bukkit.broadcast(MiniMessage.miniMessage().deserialize("disabled canAttackEntities on a conduit"));
                }
            }
        }.runTaskLater(this, 1);
    }
    @EventHandler
    void onDie(EntityDeathEvent e) {
        if (!e.getEntity().getLocation().getNearbyPlayers(10).isEmpty()) {
            Bukkit.broadcast(MiniMessage.miniMessage().deserialize(e.getEntityType()+" has died by " + e.getEntity().getLastDamageCause()));
        }
    }
}
```

PR #8063, opened 2022-06-27
Add api for disabling conduits attacking entities

u9g:
had to remake https://github.com/PaperMC/Paper/pull/8060 since I forcepushed while closed, but I fixed all reviews.

### DIFF
--- a/patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
+++ b/patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
@@ -1,0 +1,68 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: u9g <git@u9g.dev>
+Date: Mon, 27 Jun 2022 14:02:35 -0400
+Subject: [PATCH] Add api for disabling conduits attacking entities
+
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
+index 05eab04e4aec4151018f25b59f92ddbbb4c09f87..5c614a1494e505c3cbc069e41227297ce3760279 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
+@@ -51,6 +51,7 @@ public class ConduitBlockEntity extends BlockEntity {
+     @Nullable
+     private UUID destroyTargetUUID;
+     private long nextAmbientSoundActivation;
++    public boolean canAttackEntities = true; // Paper
+ 
+     public ConduitBlockEntity(BlockPos pos, BlockState state) {
+         super(BlockEntityType.CONDUIT, pos, state);
+@@ -64,6 +65,7 @@ public class ConduitBlockEntity extends BlockEntity {
+         } else {
+             this.destroyTargetUUID = null;
+         }
++        canAttackEntities = nbt.getBoolean("Paper.CanAttackEntities"); // Paper
+ 
+     }
+ 
+@@ -73,6 +75,7 @@ public class ConduitBlockEntity extends BlockEntity {
+         if (this.destroyTarget != null) {
+             nbt.putUUID("Target", this.destroyTarget.getUUID());
+         }
++        nbt.putBoolean("Paper.CanAttackEntities", this.canAttackEntities);
+ 
+     }
+ 
+@@ -214,6 +217,13 @@ public class ConduitBlockEntity extends BlockEntity {
+     }
+ 
+     private static void updateDestroyTarget(Level world, BlockPos pos, BlockState state, List<BlockPos> activatingBlocks, ConduitBlockEntity blockEntity) {
++        // Paper start
++        if (!blockEntity.canAttackEntities) {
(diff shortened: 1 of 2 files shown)


patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
```diff
@@ -0,0 +1,60 @@
...
+     private long nextAmbientSoundActivation;
++    public boolean canAttackEntities = true; // Paper
+ 
+     public ConduitBlockEntity(BlockPos pos, BlockState state) {
+         super(BlockEntityType.CONDUIT, pos, state);
+@@ -64,6 +65,7 @@ public class ConduitBlockEntity extends BlockEntity {
+         } else {
+             this.destroyTargetUUID = null;
+         }
++        canAttackEntities = nbt.contains("Paper.CanAttackEntities") && nbt.getBoolean("Paper.CanAttackEntities"); // Paper
```
NoahvdAa: `Paper.CanAttackEntities` is never saved anywhere.

Also, this will cause the conduit to default to `false` if the key doesn't exist, which would probably happen the first time existing conduits are loaded after this patch is added, meaning that these conduits will never attack entities (since that value would get saved once that is implemented).

```suggestion
+        canAttackEntities = !nbt.contains("Paper.CanAttackEntities") || nbt.getBoolean("Paper.CanAttackEntities"); // Paper
```

u9g: maybe I should just negate the naming that way everything is much simpler, rather than putting the ! everywhere

u9g: I did that^ also fixed it not being saved

patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
```diff
@@ -0,0 +1,68 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
+index 05eab04e4aec4151018f25b59f92ddbbb4c09f87..322baf49c4c86103d176f7a1f8274638df0bc9e4 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
+@@ -51,6 +51,7 @@ public class ConduitBlockEntity extends BlockEntity {
+     @Nullable
+     private UUID destroyTargetUUID;
+     private long nextAmbientSoundActivation;
++    public boolean dontAttackEntities = false; // Paper
```
Machine-Maker: don't negate this, make it canAttackEntities, just to keep it consistent with the naming scheme of the API methods. It just makes it easier to comprehend. Removes the need for the ! when setting or getting the value.

patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
```diff
@@ -0,0 +1,68 @@
...
+ 
+     public ConduitBlockEntity(BlockPos pos, BlockState state) {
+         super(BlockEntityType.CONDUIT, pos, state);
+@@ -64,6 +65,7 @@ public class ConduitBlockEntity extends BlockEntity {
+         } else {
+             this.destroyTargetUUID = null;
+         }
++        dontAttackEntities = nbt.contains("Paper.DontAttackEntities") && nbt.getBoolean("Paper.DontAttackEntities"); // Paper
```
Machine-Maker: I don't think the contains is needed here. getBoolean should return false if it doesn't exist. (make sure you adjust this if you heed my earlier suggestion)

Machine-Maker (changes requested):

patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
u9g: fixed

patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
u9g: fixed

patches/server/0919-Add-api-for-disabling-conduits-attacking-entities.patch
```diff
@@ -0,0 +1,68 @@
...
+         }
++        canAttackEntities = nbt.getBoolean("Paper.CanAttackEntities"); // Paper
+ 
+     }
+ 
+@@ -73,6 +75,7 @@ public class ConduitBlockEntity extends BlockEntity {
+         if (this.destroyTarget != null) {
+             nbt.putUUID("Target", this.destroyTarget.getUUID());
+         }
++        nbt.putBoolean("Paper.CanAttackEntities", this.canAttackEntities);
```
Lulu13022002, line 31: Missing Paper comment here

Lulu13022002:
I'm not sure if this is really needed as Owen said https://github.com/PaperMC/Paper/pull/8060#issuecomment-1167723777, you can just cancel this event, and if you need that in another event like in your example, you can use pdc on TE to check in which hand the player has placed the block.

Owen1212055:
Closing, as this is generally possible yourself by using an event and PDC to store this. 

Feel free to leave a comment if you still feel this api is justified and it can be reopened. Thanks!

PR #8066, opened 2022-06-28
Delete 0481-Prevent-headless-pistons-from-being-created.patch

Guennii:


MiniDigger:
why?
this is also not how this works.
if you are looking to contribute, please join our discord and ask for help, but I don't see why this patch should be dropped, there is already a config option anyways.

PR #8067, opened 2022-06-28
Fix experience & improvements to block events

### LINKED ISSUE #8062: Sculk Blocks Not Dropping XP
### Expected behavior

When a player breaks or uses TNT lit by a flint and steel, or uses a wither, to destroy Sculk, XP should be dropped

### Observed/Actual behavior

No XP is dropped when breaking Sculk in either of the three methods above

### Steps/models to reproduce

Place a sculk block down, use TNT and flint and steel or a wither to destroy sculk, no xp will drop

### Plugin and Datapack List

ArmourStandTools, Elevator, Essentials, EssentialsSpawn, f3NPERM, FarmProtect, MoreMobHeads, PowerlessLamps, TreeAssist, ViaVersion, WorldEdit

Vanilla-Built-in, Bukkit-World, Anti-Enderman Grief, AntiGhast, CustomNetherPortals, DoubleShulkerShells, NetherPortalCoordinates, SIlenceMobs

### Paper version

Git-Paper-8 MC:1.19

1.19-RO.1-SNAPSHOT GIT: E4f5e1c

35 versions behind on minehut

### Other

NA

### LINKED ISSUE #10080: Block breaking via command does not drop XP
### Expected behavior

Breaking a block that drops XP on mining with `/setblock x y z air destroy` does drop XP or has the chance to drop XP. Same behavior like mining with a diamond pickaxe.

### Observed/Actual behavior

Breaking a block with named command does not drop any XP

### Steps/models to reproduce

Create a fresh paper server, place or find any block that drops XP on mining and enter the setblock command with destroy property. 

[YouTube Video Comparison](https://youtu.be/qgmaTxs__jE)

### Plugin and Datapack List

No plugins, no datapacks

### Paper version

This server is running Paper version git-Paper-346 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: e035fd7)
Previous version: git-Paper-309 (MC: 1.20.2)

### Other

_No response_

Owen1212055:
This is a lot but basically adds a method to disable the dropping of experience and drops experience by default.
This way things that require XP to be dropped manually (via modification), they can drop XP themselves when needed but without touching anywhere else that may drop xp.

It should be noted this causes breakNaturally() to now drop experience.

There is alot of patches that need to touch certain methods, so it's kind of all over the place here so please feedback is appreciated.
Fix #8062
Fix #10080

---
Download the paperclip jar for this pull request: [paper-8067.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1148533055.zip)


patches/server/0295-BlockDestroyEvent.patch
```diff
@@ -11,36 +11,42 @@ floating in the air.
...
 +            // Paper start - while the above setAir method is named same and looks very similar
 +            // they are NOT used with same intent and the above should not fire this event. The above method is more of a BlockSetToAirEvent,
 +            // it doesn't imply destruction of a block that plays a sound effect / drops an item.
 +            boolean playEffect = true;
++            BlockState effectType = iblockdata;
++            int xp = iblockdata.getBlock().getExpDrop(iblockdata, (ServerLevel) this, pos, ItemStack.EMPTY, true);
 +            if (com.destroystokyo.paper.event.block.BlockDestroyEvent.getHandlerList().getRegisteredListeners().length > 0) {
-+                com.destroystokyo.paper.event.block.BlockDestroyEvent event = new com.destroystokyo.paper.event.block.BlockDestroyEvent(MCUtil.toBukkitBlock(this, pos), fluid.createLegacyBlock().createCraftBlockData(), drop);
++                com.destroystokyo.paper.event.block.BlockDestroyEvent event = new com.destroystokyo.paper.event.block.BlockDestroyEvent(net.minecraft.server.MCUtil.toBukkitBlock(this, pos), fluid.createLegacyBlock().createCraftBlockData(), drop);
++                event.setExpToDrop(xp);
```
Machine-Maker: I don't think this is a good thing to start doing. Just change or add a new constructor to the event instead of using the set method.

patches/server/0702-Add-BlockBreakBlockEvent.patch
```diff
@@ -19,12 +19,15 @@ index 718c67c630f7799c20df9850d9c2056a7218c220..224ce424b9a22548212ac157ec273c6d
 +            for (net.minecraft.world.item.ItemStack drop : net.minecraft.world.level.block.Block.getDrops(state, world.getMinecraftWorld(), pos, blockEntity)) {
 +                items.add(org.bukkit.craftbukkit.inventory.CraftItemStack.asBukkitCopy(drop));
 +            }
++            Block block = state.getBlock();
 +            io.papermc.paper.event.block.BlockBreakBlockEvent event = new io.papermc.paper.event.block.BlockBreakBlockEvent(org.bukkit.craftbukkit.block.CraftBlock.at(world, pos), org.bukkit.craftbukkit.block.CraftBlock.at(world, source), items);
++            event.setExpToDrop(block.getExpDrop(state, (ServerLevel) world, pos, net.minecraft.world.item.ItemStack.EMPTY, true));
```
Machine-Maker: Same here, see other comment on the line that instantiates BlockDestroyEvent

patches/server/0702-Add-BlockBreakBlockEvent.patch
```diff
@@ -19,12 +19,15 @@ index 718c67c630f7799c20df9850d9c2056a7218c220..224ce424b9a22548212ac157ec273c6d
...
 +            event.callEvent();
 +            for (var drop : event.getDrops()) {
 +                popResource(world.getMinecraftWorld(), pos, org.bukkit.craftbukkit.inventory.CraftItemStack.asNMSCopy(drop));
 +            }
-+            state.spawnAfterBreak(world.getMinecraftWorld(), pos, ItemStack.EMPTY, true);
++            state.spawnAfterBreak(world.getMinecraftWorld(), pos, ItemStack.EMPTY, false);
```
Machine-Maker: I think its best to leave it as true, just because that is what vanilla does. even though that dropExperience parameter does absolutely nothing.

patches/server/0919-Properly-handle-experience-dropping-on-block-break.patch
```diff
@@ -0,0 +1,84 @@
...
+     public void setPlacedBy(Level world, BlockPos pos, BlockState state, @Nullable LivingEntity placer, ItemStack itemStack) {}
+diff --git a/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java b/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
+index 771c6cf992664b65ffbf4ae0192bc7b09f77c2e6..caf924b50b9a3bb0635758873f27a99f66ed6abf 100644
+--- a/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
++++ b/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
+@@ -1022,6 +1022,7 @@ public abstract class BlockBehaviour {
+ 
+         public void spawnAfterBreak(ServerLevel world, BlockPos pos, ItemStack stack, boolean dropExperience) {
+             this.getBlock().spawnAfterBreak(this.asState(), world, pos, stack, dropExperience);
++            getBlock().popExperience(world, pos, this.getBlock().getExpDrop(asState(), world, pos, stack, true)); // Paper - spawn experience
```
Machine-Maker: doesn't this just totally undo spigot's separation of exp drops and other spawnAfter behaviors? also you aren't checking the boolean, but I think that's irrelevant as this probably shouldn't exist.

Machine-Maker (changes requested):

patches/server/0919-Properly-handle-experience-dropping-on-block-break.patch
Owen1212055: This is supposed to check on true, the idea was that this would respect it and only not drop xp if it's explicitly needed.

patches/server/0702-Add-BlockBreakBlockEvent.patch
Owen1212055: See the previous comment, this is supposed to not cause xp to drop if false.

patches/server/0919-Properly-handle-experience-dropping-on-block-break.patch
Machine-Maker: yeah, that's probably fine then. as a catchall to any other usages. I didn't see any that exist (except for whatever is going on in EnderDragon) but makes sense.

Machine-Maker (approved):

Owen1212055:
rebased

patches/api/0168-BlockDestroyEvent.patch
```diff
@@ -36,23 +37,42 @@ index 0000000000000000000000000000000000000000..051b2ef76a914228338fa28553ad739b
...
++    @NotNull
++    public BlockData getEffectBlock() {
++        return this.effectBlock;
++    }
++
++    /**
++     * Sets the effect that will be played when the block is broken.
++     *
++     * @param effectBlock block effect
++     */
```
Machine-Maker: Should mention/link the `#playEffect()` as controlling if the effect plays at all and
to disable effect, use `#setPlayEffect()`

patches/server/0288-BlockDestroyEvent.patch
```diff
@@ -22,21 +22,25 @@ index 178509c545f2872174af501bdcec3314f703739c..d315a31f6b489ee88f81eb3d1b358c92
...
-@@ -669,8 +670,21 @@ public abstract class Level implements LevelAccessor, AutoCloseable {
+@@ -669,8 +670,25 @@ public abstract class Level implements LevelAccessor, AutoCloseable {
              return false;
          } else {
              FluidState fluid = this.getFluidState(pos);
 +            // Paper start - while the above setAir method is named same and looks very similar
 +            // they are NOT used with same intent and the above should not fire this event. The above method is more of a BlockSetToAirEvent,
 +            // it doesn't imply destruction of a block that plays a sound effect / drops an item.
 +            boolean playEffect = true;
++            BlockState effectType = iblockdata;
```
Machine-Maker, line 33: I don't see this actually being used for the effect. You need to change the BlockState that goes into the levelEvent call

patches/server/1058-Properly-handle-experience-dropping-on-block-break.patch
```diff
@@ -0,0 +1,90 @@
...
+-            state.spawnAfterBreak((ServerLevel) world, pos, tool, true);
++            state.spawnAfterBreak((ServerLevel) world, pos, tool, dropExperience); // Paper
+         }
+ 
+     }
+@@ -429,11 +437,16 @@ public class Block extends BlockBehaviour implements ItemLike {
+         this.playerDestroy(world, player, pos, state, blockEntity, tool, true);
+     }
+     public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack tool, boolean includeDrops) {
++        // Paper end
```
Machine-Maker: dangling comment

patches/server/1058-Properly-handle-experience-dropping-on-block-break.patch
```diff
@@ -0,0 +1,90 @@
...
+ 
+diff --git a/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java b/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
+index 3ab8b99837b1d1faea722c598b0228b2780be8b1..401f1b418a4f57059e991cc958ebadb31239a581 100644
+--- a/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
++++ b/src/main/java/net/minecraft/world/level/block/state/BlockBehaviour.java
+@@ -1286,6 +1286,7 @@ public abstract class BlockBehaviour implements FeatureElement {
+ 
+         public void spawnAfterBreak(ServerLevel world, BlockPos pos, ItemStack tool, boolean dropExperience) {
+             this.getBlock().spawnAfterBreak(this.asState(), world, pos, tool, dropExperience);
++            if (dropExperience) {getBlock().popExperience(world, pos, this.getBlock().getExpDrop(asState(), world, pos, tool, true));} // Paper - spawn experience
```
Machine-Maker: I think this will cause xp to drop twice when a player breaks a block. This is because spawnAfterBreak is called with a true parameter via `Block#playerDestroy` -> `Block#dropResources`

patches/server/1058-Properly-handle-experience-dropping-on-block-break.patch
```diff
@@ -0,0 +1,90 @@
...
+     }
+@@ -429,11 +437,16 @@ public class Block extends BlockBehaviour implements ItemLike {
+         this.playerDestroy(world, player, pos, state, blockEntity, tool, true);
+     }
+     public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack tool, boolean includeDrops) {
++        // Paper end
++        // Paper start
++        this.playerDestroy(world, player, pos, state, blockEntity, tool, true, true);
++    }
++    public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack tool, boolean includeDrops, boolean dropExp) {
```
Machine-Maker: This overload isn't used anywhere except right above it. And **if** we are adding another parameter to it... you need to change **all** of the overrides as well. If you are going to do that, just change the method we already add to have the extra parameter.

Machine-Maker (changes requested):

patches/server/1058-Properly-handle-experience-dropping-on-block-break.patch
Owen1212055: spawnAfterBreak doesn't drop any experience, as per craftbukkit change.

Machine-Maker: It does if `dropExperience` is true, which it is in the call chain from ServerPlayerGameMode#destroyBlock -> Block#playerDestroy -> BLock#popResource -> Block#spawnAfterBreak

Owen1212055:
all fixed

patches/server/1058-Properly-handle-experience-dropping-on-block-break.patch
```diff
@@ -0,0 +1,98 @@
...
+-            state.spawnAfterBreak((ServerLevel) world, pos, tool, true);
++            state.spawnAfterBreak((ServerLevel) world, pos, tool, dropExperience); // Paper
+         }
+ 
+     }
+@@ -429,11 +437,10 @@ public class Block extends BlockBehaviour implements ItemLike {
+         this.playerDestroy(world, player, pos, state, blockEntity, tool, true, true);
+     }
+     public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack tool, boolean includeDrops, boolean dropExp) {
+-        // Paper end
```
Machine-Maker: seems to be accidentally removed

Machine-Maker (changes requested):
I found at least one case of double xp dropping still. CraftBlock#breakNaturally calls popResources and that has a true for dropping xp but it also handles xp drops later itself.

Machine-Maker (approved):

PR #8072, opened 2022-06-29
Let JavaPluginLoader load classes from plugins that the requested plugin depended on firstly

theTd:
If a plugin is using a library class packaged by a plugin they're depended on, at the same time another plugin also packaged that library but another version, problem may raise.
This patch will solve it.


Owen1212055:
If I am understanding this correctly, this is essentially causing dependencies to be iterated over first?

Owen1212055:
See https://github.com/PaperMC/Paper/pull/8108 in general

Introducing a change like this *may* cause some weird issues because of the fun mess that is spigot plugins right now. It's super hard to tell if certain plugins actually depend on the unknown behavior of this. (Notice how when we introduced prioritized classloading it is under a toggle) 

So I feel like it's better to leave the spigot classloading resolution alone for now, but with the new plugin type we are introducing we can introduce a new performant way of resolving other classloaders like this.

PR #8074, opened 2022-06-29
Fire ServerListPingEvent for secondary motd send (fixes #8073)

### LINKED ISSUE #8073: RGB colors in MOTD not supported when player is kicked
### Expected behavior

Paper supports RGB color codes in the MOTD that is displayed in the client's server list. This is an extension on vanilla servers which don't support RGB in MOTD at all. The MOTD should behave consistently and keep the defined colors regardless of how a player disconnects.

### Observed/Actual behavior

When a player is kicked from the server, the server's MOTD falls back to vanilla colors which messes up the RGB codes in the MOTD definition. Refreshing the server list fixes the colors per Paper's RGB support. When a player disconnects from the server voluntarily, the RGB colors are also respected. It points to the fact that when a player is kicked by the server, a packet is sent which updates the MOTD but the method for displaying colors here is different from the usual one which Paper uses.

### Steps/models to reproduce

1. Connect to a 1.19 server which uses RGB codes in MOTD
2. Disconnect
3. The RGB colors in the MOTD are OK
4. Connect again
5. Get kicked by the server
6. The MOTD's colors are broken
7. Refresh the server list
8. The RGB colors in the MOTD are OK

### Plugin and Datapack List

Plugins (28): ChatBubbles, Chunky, CoreProtect, DecentHolograms, dynmap, Essentials, EssentialsChat, EssentialsDiscord, EssentialsGeoIP, EssentialsProtect, EssentialsSpawn, floodgate, Geyser-Spigot, GSit, LuckPerms, LWC, Multiverse-Core, Multiverse-Inventories, Multiverse-NetherPortals, OpenInv, QuickChannels, Simple-Playerlist, SkQuery, Skript, Vault, WorldEdit, WorldEditSUI, WorldGuard

There are 3 data packs enabled: [vanilla (built-in)], [file/bukkit (world)], [file/kamienok (world)]

(Custom datapack only for custom recipes)

### Paper version

[14:59:13 INFO]: This server is running Paper version git-Paper-40 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 0d79b86)
Previous version: git-Paper-34 (MC: 1.19)

### Other

[A recent issue](https://github.com/PaperMC/Paper/issues/8036) was fixed by [changing how MOTD is parsed](https://github.com/PaperMC/Paper/commit/147a2b7f54a077447ba5afd74e4f921b497f5205). Maybe this could also be implemented in the new way servers send packets after kicking.

Vanilla 1.19 introduced something related (taken from the [release announcement](https://www.minecraft.net/en-us/article/the-wild-update-out-today-java) changelog):
> Servers will now also send an additional icon and MOTD packet after a player has connected
> This allows servers with enable-status=false to set an icon and MOTD for players that successfully connect

This might be a longshot but I also found a [bug report in Velocity](https://github.com/PaperMC/Velocity/issues/769) with a similar description, also the new vanilla MOTD behavior is mentioned in the comments.

NoahvdAa:
Fixes #8073. Not sure if there should be a flag on the event indicating whether it was fired for a connected player or just someone pinging.

### DIFF
--- a/patches/server/0178-Implement-extended-PaperServerListPingEvent.patch
+++ b/patches/server/0178-Implement-extended-PaperServerListPingEvent.patch
@@ -60,10 +60,10 @@ index 0000000000000000000000000000000000000000..d926ad804355ee2fdc5910b2505e8671
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/network/StandardPaperServerListPingEventImpl.java b/src/main/java/com/destroystokyo/paper/network/StandardPaperServerListPingEventImpl.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..4c2351b03b58511b80017b58ee9b20ab5193adc9
+index 0000000000000000000000000000000000000000..84285d5235b66aee9a4059b1d4db4a99e296f22f
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/network/StandardPaperServerListPingEventImpl.java
-@@ -0,0 +1,110 @@
+@@ -0,0 +1,121 @@
 +package com.destroystokyo.paper.network;
 +
 +import com.destroystokyo.paper.profile.CraftPlayerProfile;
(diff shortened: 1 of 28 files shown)


molor:
> there should be a flag on the event indicating whether it was fired for a connected player or just someone pinging

Maybe something like `PaperServerListPingEvent#getReceiver()` would be better in that case, e.g. if not null then it's fired for that connected player, otherwise it's some1 pinging

But there's already `PaperServerListPingEvent#getBukkitPlayer()` (mentioned in javadocs without any useful information) and idk maybe it's already can be used for that, I'm sorry in that case

electronicboy (approved):

NoahvdAa:
> But there's already `PaperServerListPingEvent#getBukkitPlayer()` (mentioned in javadocs without any useful information) and idk maybe it's already can be used for that, I'm sorry in that case

That's an internal method used for the iterator. Perhaps a nullable getter `#getPlayer` (or the name you suggested, `#getReceiver`) could be added?

Owen1212055:
I know you've mentioned this before but I'd like to reiterate this on the PR. In general, I feel like this would be better as a separate event. As this is sort of different behavior and might be nice to control it individually.

PR #8075, opened 2022-06-29
Don't fire TNTPrimeEvent for EnderDragon + fixes

Owen1212055:
This seems to have been back when the patch was originally made, but it doesn't make much sense for this to be here.

This caused every single block broken by the dragon to fire a TNTPrimeEvent (even if it wasn't TNT) IF EntityExplodeEvent yield was set to anything but 0.
Additionally, blocks that actually were TNT would fire another event because wasExploded calls the event.
This does cause a small behavior change in that this event will no longer cause the block to be removed. 
However, because this event uses an EntityExplodeEvent you can still customize these blocks.

The behavior of tnt exploding/blocks dropping from ender dragons is not vanilla behavior.

### DIFF
--- a/patches/server/0919-EnderDragon-EntityExplodeEvent-Logic-FIxes.patch
+++ b/patches/server/0919-EnderDragon-EntityExplodeEvent-Logic-FIxes.patch
@@ -1,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Wed, 29 Jun 2022 20:50:08 -0400
+Subject: [PATCH] EnderDragon EntityExplodeEvent Fixes
+
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index f872b60697525e3015979f286bfb08d29bb001df..610f5180841c44d78f1510cf8618eb169bb79f78 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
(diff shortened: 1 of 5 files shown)


Machine-Maker:
> The behavior of tnt exploding/blocks dropping from ender dragons is not vanilla behavior.

well we need to get rid of this then right?

Owen1212055:
> > The behavior of tnt exploding/blocks dropping from ender dragons is not vanilla behavior.
> 
> well we need to get rid of this then right?

It's plugin set behavior by setting the yield to anything but 0 in the EntityExplodeEvent, sorry that wasn't clarified.

Machine-Maker (approved):

Owen1212055:
Updated with another patch that shows the level effect + don't explode tnt.

patches/server/0919-EnderDragon-EntityExplodeEvent-Logic-FIxes.patch
```diff
@@ -0,0 +1,29 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index f872b60697525e3015979f286bfb08d29bb001df..610f5180841c44d78f1510cf8618eb169bb79f78 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
++++ b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -530,6 +530,7 @@ public class EnderDragon extends Mob implements Enemy {
+                 this.level.removeBlock(new BlockPos(block.getX(), block.getY(), block.getZ()), false);
+             }
+         } else {
++            flag1 = false; // Paper
```
kennytv: Needs a diff on change comment where initialized or checked for the return

Owen1212055:
Diff on change + rebased

electronicboy (approved):

Warriorrrr (approved):

Issue #8057, opened 2022-06-26
quasi connectivity


agoniaprivat:
### Expected behavior

See video:
https://youtu.be/VR_GBL1USEY

First clip is the issue 00:00 - 00:19. On the other ones it works like it should.

### Observed/Actual behavior

00:19 -> (video above)

### Steps/models to reproduce

https://youtu.be/VR_GBL1USEY

### Plugin and Datapack List
![image](https://user-images.githubusercontent.com/50961682/175840768-50f1ae7a-c214-46c6-8e7e-94ff7dd64861.png)

### Paper version

39-1.19 - on paper

### Other

worked on 1.18.2

lynxplay:
Are you using eigencraft redstone engine ? From my testing, vanilla is working fine, while I can replicate your issue on eigencraft.

If that is the case, it would probably make sense to open a new issue specifically for the eigencraft engine.

lynxplay:
Closed by a22eb678af9555e7f55a03d397178e8b87d7c1ca.
If you are still having issues after the latest update, please either comment here or open a new issue.

PR #8076, opened 2022-06-29
Update Eigencraft patch to fix Vanilla parity issues

SpaceWalkerRS:
1.19 made changes to how shape and block updates are processed, adding them to a stack before processing them. This introduces Vanilla parity issues with the Eigencraft redstone implementation. This PR updates the Eigencraft patch to bypass the neighbor update stack for any shape and block updates emitted, mostly restoring old behavior and fixing these issues. The Alternate Current implementation already does this as well.


lynxplay:
Resolves: #8057. Tested locally.

kennytv (approved):

PR #8077, opened 2022-06-30
Fix playing adventure sounds on World/Server

Machine-Maker:
playing adventure sounds on a World/Server had some unexpected side effects due to ForwardingAudience's delegation to each individual audience. This overrides all 3 playSound methods in CraftWorld/CraftServer with better implementations.

----

The issues arrises due to seeded sounds. When you play a sound on a World or Server without a seed, it should pick one seed. For location-specific sounds on World/Server, packets shouldn't be sent to all players, only players within range, so overriding also allows delegation to vanilla packet broadcast mechanics.

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -736,10 +736,10 @@ index 0000000000000000000000000000000000000000..2fd6c3e65354071af71c7d8ebb97b559
 +}
 diff --git a/src/main/java/io/papermc/paper/adventure/PaperAdventure.java b/src/main/java/io/papermc/paper/adventure/PaperAdventure.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..badde58c5151f838faa4b42db02e767eafa2da18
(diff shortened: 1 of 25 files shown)


Machine-Maker:
Rebased for 1.19.3

patches/server/0010-Adventure.patch
```diff
@@ -832,7 +840,8 @@ index 0000000000000000000000000000000000000000..01e424792f68bac73ec41726031ebbb5
 +        })
 +        .build();
 +    public static final AttributeKey<Locale> LOCALE_ATTRIBUTE = AttributeKey.valueOf("adventure:locale"); // init after FLATTENER because classloading triggered here might create a logger
-+    @Deprecated public static final PlainComponentSerializer PLAIN = PlainComponentSerializer.builder().flattener(FLATTENER).build();
++    @Deprecated
```
jpenilla, line 843: unneeded change

jpenilla (changes requested):

jpenilla (approved):
Approved

kashike (approved):

PR #8078, opened 2022-06-30
Raw Itemstack improvements

### LINKED ISSUE #5071: ItemStack#setType is heavy when the itemstack got a bunch of display name and lore
plugin in question here calls ItemStack#setType to DIAMOND_PICKAXE, to call getDrops using the tool and getting the right drops. replacing that with new ItemStack(DIAMOND_PICKAXE) avoided this issue

https://spark.lucko.me/#YnTs0gpbw1

we should be able to avoid the reparsing here, doesnt seem necessary

### LINKED ISSUE #5188: Performance regression of ItemStack#getEnchantmentLevel
Patch 0063 changes the implementation of ItemStack#getEnchantmentLevel and ItemStack#containsEnchantment. Instead of directly accessing the item's enchantment data, they now perform an expensive getItemMeta call:
 - https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0063-Handle-Item-Meta-Inconsistencies.patch#L129-L134
 - https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0063-Handle-Item-Meta-Inconsistencies.patch#L123-L124

Previously, ItemStack#getEnchantmentLevel was a fast alternative to getItemMeta, but this change makes them just as slow as each other.

Revert the implementation of getEnchantmentLevel and containsEnchantment to upstream.

Call `EnchantmentManager.getEnchantmentLevel(CraftEnchantment.getRaw(ench), handle)` directly through reflection/nms.

**Additional context**
The patch resolves item meta inconsistencies, but getEnchantmentLevel and containsEnchantment can't introduce any inconsistencies given that they don't write any enchantments. It should be safe to continue using the faster upstream implementations.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/5188
Paper has fixed enchantment order inconsistencies but has also implemented the fix on getter which is useless because no write operation is done here
In fact i think all methods outside of the ItemMeta class shouldn't call itemmeta at all even for setter (or at least it should be documented) because developer who use this method don't think that the whole tag will be reparsed

Closes https://github.com/PaperMC/Paper/issues/5071
Remove the reparse of the whole itemmeta because the tag is already here at this stage. This method just change the item in the nms class (also remove the bukkit cached itemstack impl), and the itemmeta will be reparsed when the developer recall getItemMeta() without the help of bukkit

### DIFF
--- a/patches/server/0065-Handle-Item-Meta-Inconsistencies.patch
+++ b/patches/server/0065-Handle-Item-Meta-Inconsistencies.patch
@@ -70,7 +70,7 @@ index e937186aaf819a77c80beeb9e08413a1f781c13a..0e19f49ca2496b1c42d27289bcea15d2
  
      public boolean isEnchanted() {
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-index 4b79b96579efbc4dd9a10e7183ed08b73b488794..e42a88dee9f1813d76a0c7c62bb8e46fd1339de0 100644
+index 4b79b96579efbc4dd9a10e7183ed08b73b488794..ff77f28bc26617b17587fea8c730a14f56820386 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
 @@ -126,10 +126,48 @@ public final class CraftItemStack extends ItemStack {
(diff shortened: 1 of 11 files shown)


Lulu13022002:
As said on discord, the second fix keep bad data, the new one doesn't allow any bad data including (count overflow, enchantment, tags, custom model and damage on undamageable item). I have also change all the enchantment related methods to just use the TreeMap instead of a plain ItemMeta

Machine-Maker:
I really don't like a lot of what's going on here. You have some failing item meta tests that have to be addressed, and I think you're going to encounter more that fail due to the item meta changes. There is a lot more nbt that has to be addressed than just enchantments, model data, and damage. Some nbt tags apply to more than one item type, like book pages, some doesn't. The way item meta is setup now, with the applyTo method and the constructors that take an instance of CraftMetaItem to copy stuff over... all of that logic is going to have to be replicated inside the setType method.

Lulu13022002:
Yes i know the things is that itemmeta rebuild the whole tag from the ground. And it's hard to replicate but if the test covered all the cases it should be easier to fix

Lulu13022002:
All test passed now normally

electronicboy:
There's a lot of stuff going on in here, one or two opinionated things which shouldn't be in there, and generally a lot of extra stuff which has 0 validation around a system that is already a massive arse mess...

Some of this may make sense as separate PRs, but, as is, this is unreviewable to us

Lulu13022002:
Yes that's fine, in fact the whole item meta is bad since it recreate the nbttag from the ground (and that happens even if you move your item from one pixel in your inventory). The question is can i reopen for the first issue: 5188 ? without all that mess

PR #8079, opened 2022-06-30
Fix block destroy desync

spookyoctopus:
It is quite common to have blocks get desynced between the client and
server, Especially with the use of plugins. This patch reduces the
consequences of such a desync, by sending a BlockUpdatePacket to the client on a block action
such as digging.


electronicboy:
This patch is dangerous, if it was to do this, it would need to be in the block, but I'm generally against just blindly sending such packets vs trying to actually target such issues.

PR #8082, opened 2022-07-01
Strip legacy color chars in gui

### LINKED ISSUE #7827: GUI console shows legacy formatting codes
### Expected behavior

No formatting codes in the GUI console

### Observed/Actual behavior

Legacy formatting codes show up in the GUI console
javaw_757B4jeypg

### Steps/models to reproduce

1) Run Paper by double-clicking the .jar
2) Produce a vanilla message in chat, for example by joining the server.

### Plugin and Datapack List

none

### Paper version

This server is running Paper version git-Paper-334 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 8d1de8d)
Previous version: git-Paper-71 (MC: 1.18.1)

### Other

Tested on vanilla, they do not show up there.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7827
It seems the gui doesn't receive any form of ANSI char but just the legacy char code
And spigot strip only ANSI
This new patch strip both legacy named char and most recent syntax using hexadecimal for all other rgb colors

### DIFF
--- a/patches/server/0918-Strip-color-char-in-console-gui.patch
+++ b/patches/server/0918-Strip-color-char-in-console-gui.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
+Date: Fri, 1 Jul 2022 18:58:08 +0200
+Subject: [PATCH] Strip color char in console gui
+
+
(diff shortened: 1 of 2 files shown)


Owen1212055:
Do we have any built-in utilities that may already cover this behavior? This would seem like something we may already have.

Lulu13022002:
well the ChatColor class is deprecated and doesn't strip the rgb color (formed with hex) and the other seems a little bit overkill.

Issue #8086, opened 2022-07-02
Getting the source inventory on `InventoryMoveItemEvent` the moved item size is always one.


katekerllenevich:
### Expected behavior

`event.getSource().getItem(0).getAmount()` on `InventoryMoveItemEvent` should return the amount of items that are actually in slot 0 in the inventory.

### Observed/Actual behavior

The above method will always return one, where as it should only be that `event.getItem().getAmount()` should return one.

### Steps/models to reproduce
```
package dev.omega24.test;

public class TestPlugin extends JavaPlugin, implements Listener {

    @Override
    public void onEnable() {
        getServer().getPluginManager().registerEvents(this, this);
    }

@EventHandler
    public void onInventoryMoveItem(InventoryMoveItemEvent event) {
        if (!(event.getSource().getHolder(false) instanceof Hopper hopper)) {
            return;
        }

        System.out.println("item: " + event.getItem().getAmount());
            System.out.println("source: " + event.getSource().getItem(0).getAmount());
    }
}
```

### Plugin and Datapack List

[17:36:40 INFO]: Plugins (1): HopperSort

### Paper version

```
[17:35:41 INFO]: This server is running Paper version git-Paper-43 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 9d0650e)
You are running the latest version
Previous version: git-Paper-42 (MC: 1.19)
```

### Other

The problem stems from the `HopperBlockEntity.hopperPush` function.

What happens is that on the eight line of the function the original item stack is copied to a new variable, called itemstack (`ItemStack itemstack = origItemStack;`). This is done so that `event.getItem()` always has an amount of one, but because of Java's "pass by reference" nature, this always makes all changes to origItemStack happen to itemstack.

The problem line is `origItemStack.setCount(moved);`. On this line the original item stack is set to the count to be moved (this is where the number one comes from), which allows `event.getItem()` to have the correct amount. However this also makes it for a split second that the Hopper inventory's first (or whatever slot origItemStack is in) slot has a stack size of one.

To fix this there needs to be two changes:
1. `ItemStack itemstack = origItemStack;` needs to turn into `ItemStack itemstack = origItemStack.clone();`
2. `origItemStack.setCount(moved);` needs to turn into `itemstack.setCount(moved);`

This does bring up performance questions on how badly `.clone()` will affect the server performance, but if thats a concern why is there a whole new variable being created just to reference the original?

Doc94:
[...]

PR #8087, opened 2022-07-02
Fix count in origin ItemStack for InventoryMoveItemEvent

Doc94:
This fixs #8086 based in the same issue the problem is the item passed to the event not have the change in the amount to pass.

note: need learn more git because when make https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#method-1 this i cannot modify the patch... this was a directly edition in patch.. tested with applyPatches

### DIFF
--- a/patches/server/0339-Optimize-Hoppers.patch
+++ b/patches/server/0339-Optimize-Hoppers.patch
@@ -78,7 +78,7 @@ index a05acf709735b40ca86f978508c63a86065fd405..6a1405a8630e90db3b5a3c9152259ba6
  
      double getLevelY();
 diff --git a/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java
-index a507d7f65a94e49ecd18cd18797b156474558390..a7ac6b528aecae528a17af157f8ec29371e4484c 100644
+index a507d7f65a94e49ecd18cd18797b156474558390..765c7a082e3fa4ed1324aaaf1b6395d3ede6113e 100644
 --- a/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java
 +++ b/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java
 @@ -3,7 +3,6 @@ package net.minecraft.world.level.block.entity;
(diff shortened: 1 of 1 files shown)


patches/server/0338-Optimize-Hoppers.patch
```diff
@@ -114,11 +114,11 @@ index a507d7f65a94e49ecd18cd18797b156474558390..a7ac6b528aecae528a17af157f8ec293
 +            if (!item.isEmpty()) {
 +                foundItem = true;
 +                ItemStack origItemStack = item;
-+                ItemStack itemstack = origItemStack;
++                ItemStack itemstack = origItemStack.copy();
```
Doc94: the issue says about use clone.. but NMS have a method for this.. my issue if can be this method or can be `cloneItemStack(true)`

katekerllenevich: I just said `clone` because thats what bukkit uses but `copy` should work too

kennytv:
On a quick glance there's more in the patch to be cleaned up if you clone the original item. also, here's how you properly fixup patches https://youtu.be/K_lym3WHoFA

electronicboy:
blindly cloning the original item negates a good chunk of the perf improvements here, would defo need that to at least be in the fire event, but, afaik, the patch currently maintains the behavior of upstream in terms of how the event works but I'd need to double check

Doc94:
> On a quick glance there's more in the patch to be cleaned up if you clone the original item. also, here's how you properly fixup patches https://youtu.be/K_lym3WHoFA

not sure why now works.. i try the same thing (based in contribution) and the class was reverted to the original.. maybe a WSL cache thing...

i rebased this and add the same thing to the pull...

About the copy thing not sure if can be another thing i can do in this... i dont tested only with the change in count ignoring the copy...

montlikadani:
What about if we just sets the amount of item before calling PushMoveEvent and adding the item into the hopper instead of hard-copying the item? I mean
```diff
                final int origCount = origItemStack.getCount();
                final int moved = Math.min(level.spigotConfig.hopperAmount, origCount);
                origItemStack.setCount(moved);

                // We only need to fire the event once to give protection plugins a chance to cancel this event
                // Because nothing uses getItem, every event call should end up the same result.
                if (!skipPushModeEventFire) {
+                   itemstack.setCount(origCount);
                    itemstack = callPushMoveEvent(destination, itemstack, hopper);
                    if (itemstack == null) { // cancelled
                        origItemStack.setCount(origCount);
                        return false;
                    }
                }
                final ItemStack itemstack2 = addItem(hopper, destination, itemstack, enumdirection);
+               itemstack.setCount(moved);
                final int remaining = itemstack2.getCount();
```
or we could just move this `origItemStack.setCount(moved)` after `addItem`?

Owen1212055 (changes requested):
See above. It's a bit eh all the sudden cloning items like this in code that can get very hot like this. It might be worth looking into what we can do to avoid having to reclone the entire item but properly update the count to try to keep things performant here.

PR #8090, opened 2022-07-03
Leave a paper.yml.txt stub pointing to new location

Machine-Maker:


NoahvdAa (approved):

kennytv (approved):

patches/server/0004-Paper-config-files.patch
```diff
@@ -1198,6 +1206,12 @@ index 0000000000000000000000000000000000000000..fa191f418079a8ee24326c5952d12d44
 +    }
 +
 +    public static PaperConfigurations setup(final Path legacyConfig, final Path configDir, final Path worldFolder, final File spigotConfig) throws Exception {
++        final Path legacy = Files.isSymbolicLink(legacyConfig) ? Files.readSymbolicLink(legacyConfig) : legacyConfig;
++        final Path replacementFile = legacy.resolveSibling(legacyConfig.getFileName() + ".txt");
++        if (Files.notExists(replacementFile)) {
++            Files.createFile(replacementFile);
++            Files.writeString(replacementFile, String.format(MOVED_NOTICE, configDir.toAbsolutePath()));
```
MiniDigger, line 1215: not sure if this should be absolute but rather relative to server root maybe?
thinking of hosts and stuff where the absolute path isn't really useful and could confuse users.

Lulu13022002: I think this can be relative if it's in the server root. But there is a cli option to change the config dir path that allow outside path and a relative path like ../ can be confusing

JRoy (approved):

MiniDigger (approved):

PR #5674, opened 2021-05-18
Add missing EntityDropItemEvent calls

Machine-Maker:

PR #8094, opened 2022-07-04
Call EntityDropItemEvent for goats dropping horns

### LINKED ISSUE #8093: Goat dropping a Goat Horn does not trigger EntityDropItemEvent
### Expected behavior

Goat dropping a Goat Horn triggers EntityDropItemEvent

### Observed/Actual behavior

Goat dropping a Goat Horn does not trigger EntityDropItemEvent

### Steps/models to reproduce

Listen on EntityDropItemEvent, have some sort of print message that triggers if the entity is a goat and the item is a goat horn.
Then, summon a screaming goat (they ram faster), stand still in survival mode in front of stone, and dodge the goat when it rams. It should drop a horn when it hits the wall, and if the event is triggered, should print the message.

### Plugin and Datapack List

ApolloUHC (my custom plugin), ProtocolLib

### Paper version

This server is running Paper version git-Paper-48 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 385f313)

### Other

_No response_

lynxplay:
Calls the EntityDropItemEvent when a goat drops one of its goat horns.

As with other calls of the EntityDropItemEvent, the event is published
after the modification of the entity and cancellation of it does not
restore the entities state. In this case, the goat would still loose
their horn however the item would not be dropped (See snowmen dropping
their pumpkin).

Resolves: #8093 

-----------------------------------------------------------------------------------------------------

Notes about this PR:

- ~~Most other calls of the EntityDropItemEvent are called from `Entity#spawnAtLocation` in combination with the `Entity.forceDrops` field. However this logic does not allow for custom x/y/z, which the goat computes~~

- As laid out, this event is called after the goat data was mutated to no longer have the horn dropped. This is in line with some other entities that drop something, e.g. Snowmen or Leash. However idk if this is part of the event contract, the javadocs (as with all spigot things) is rather vague.

### DIFF
--- a/patches/server/0922-Add-various-missing-EntityDropItemEvent-calls.patch
+++ b/patches/server/0922-Add-various-missing-EntityDropItemEvent-calls.patch
@@ -57,3 +57,17 @@ index 8f294f10aca2df007830b12da0506f7614206a89..6a66b5d1a3d8615dcc15057f03476e9c
      }
  
      @Override
+diff --git a/src/main/java/net/minecraft/world/entity/animal/goat/Goat.java b/src/main/java/net/minecraft/world/entity/animal/goat/Goat.java
+index 56dd01801f56c56d07101e7e22b58ac059f5f07f..31be36e6b7b6bd0c0d7fda4e1b03ecd38947f3a5 100644
+--- a/src/main/java/net/minecraft/world/entity/animal/goat/Goat.java
++++ b/src/main/java/net/minecraft/world/entity/animal/goat/Goat.java
+@@ -329,8 +329,7 @@ public class Goat extends Animal {
+             double d2 = (double) Mth.randomBetween(this.random, -0.2F, 0.2F);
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Machine-Maker:
If you merge https://github.com/PaperMC/Paper/pull/5674 before this PR, you can use the Entity#spawnAtLocation method. I think that's better than adding yet another separate location calling that event. Better to have as many as you can in one place.

lynxplay:
Perfect :+1: Yea I checked through the existing patches, missed the open PR :sweat_smile:  Do you want to just include this fix in your PR or do we pull yours and I modify the patch later once it is pulled to main ?

lynxplay:
Rebased/injected into patch introduced by ##5674

kennytv (approved):

PR #8095, opened 2022-07-05
Fix a bunch more forceDrops for dropping items

Machine-Maker:


electronicboy (approved):

Owen1212055 (approved):
This whole thing seems hacky, is this for handling capturing entity item drops? For future reference, what warrants having this boolean set around item drops?

Although, it looks good on the web.

lynxplay:
> This whole thing seems hacky, is this for handling capturing entity item drops? For future reference, what warrants having this boolean set around item drops?
> 
> Although, it looks good on the web.

Issue basically is that both normal alive drops as well as drops when dying use the `spawnAtLocation` method to drop their stuff.
Spigot captures all those items as it believes they are drops. This boolean prevents the capture when an item is dropped normally (not during death).

Whether it is worth it to move from "only prevent capture when bool" to "only capture when bool" down the line, might be worth some consideration.

Machine-Maker:
> Whether it is worth it to move from "only prevent capture when bool" to "only capture when bool" down the line, might be worth some consideration.

I think in the future, it should just be a separate method. and we change it where the death drops are. But that can really be a future, post-hard fork thing to tackle.

PR #8099, opened 2022-07-05
Fix EntityChangeBlockEvent for silverfish

Lulu13022002:
The new block data of this event is wrong when the gamerule mobGriefing is false and the silverfish try to call its friends in the nearby infested blocks


patches/server/0874-Fix-new-block-data-for-EntityChangeBlockEvent.patch
```diff
@@ -0,0 +1,40 @@
...
+-                                if (org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(this.silverfish, blockposition1, net.minecraft.world.level.block.Blocks.AIR.defaultBlockState()).isCancelled()) {
++                                // Paper start
++                                org.bukkit.event.entity.EntityChangeBlockEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(this.silverfish, blockposition1, world.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING) ? net.minecraft.world.level.block.Blocks.AIR.defaultBlockState() : ((InfestedBlock) block).hostStateByInfested(world.getBlockState(blockposition1)));
++                                if (event.isCancelled()) {
++                                    // Paper end
+                                     continue;
+                                 }
+                                 // CraftBukkit end
+-                                if (world.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)) {
++                                if (event.getBlock().isEmpty()) { // Paper
```
Machine-Maker: This isn't right. isEmpty checks if the block is air, which it definitely is not since there is `block instanceof InfestedBlock` right above it.

Machine-Maker (dismissed):

patches/server/0874-Fix-new-block-data-for-EntityChangeBlockEvent.patch
Lulu13022002: You have right i have checked the old data and not the new one, fixed in the latest changes

patches/server/0874-Fix-new-block-data-for-EntityChangeBlockEvent.patch
```diff
@@ -0,0 +1,40 @@
...
+-                                if (org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(this.silverfish, blockposition1, net.minecraft.world.level.block.Blocks.AIR.defaultBlockState()).isCancelled()) {
++                                // Paper start
++                                org.bukkit.event.entity.EntityChangeBlockEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(this.silverfish, blockposition1, world.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING) ? net.minecraft.world.level.block.Blocks.AIR.defaultBlockState() : ((InfestedBlock) block).hostStateByInfested(world.getBlockState(blockposition1)));
++                                if (event.isCancelled()) {
++                                    // Paper end
+                                     continue;
+                                 }
+                                 // CraftBukkit end
+-                                if (world.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)) {
++                                if (event.getTo().isAir()) { // Paper
```
Machine-Maker: I think there is still a problem here. destroyBlock runs different logic than setBlock and so what if the mobGriefing gamerule is set to false, but the event sets the block to air? Should it run the destroyBlock logic? That wouldn't happen in vanilla since the game rule is set to false.

Lulu13022002: Yes do you think it's better to keep the vanilla behavior or allow plugin to bypass the gamerule ?
But the things is that in vanilla there are no plugin also.

Lulu13022002: Fixed, i have kept the vanilla behavior here

electronicboy (approved):

Owen1212055 (approved):
Tested and looks good, I confirmed and this looks like vanilla behavior as well.

Owen1212055 (approved):
BEFORE:
``
EntityChangeBlockEvent{block=CraftBlock{pos=BlockPos{x=101, y=104, z=-127},type=INFESTED_CHISELED_STONE_BRICKS,data=Block{minecraft:infested_chiseled_stone_bricks},fluid=net.minecraft.world.level.material.EmptyFluid@7e19755a}, cancel=false, to=CraftBlockData{minecraft:air}, entity=CraftSilverfish}
``

AFTER:
``
EntityChangeBlockEvent{block=CraftBlock{pos=BlockPos{x=101, y=104, z=-127},type=INFESTED_CHISELED_STONE_BRICKS,data=Block{minecraft:infested_chiseled_stone_bricks},fluid=net.minecraft.world.level.material.EmptyFluid@6f49d153}, cancel=false, to=CraftBlockData{minecraft:chiseled_stone_bricks}, entity=CraftSilverfish}
``

PR #8103, opened 2022-07-06
Expose CommandSender for command blocks

Machine-Maker:


electronicboy (approved):

NoahvdAa (approved):

Owen1212055 (approved):

Machine-Maker (commented):
From jmp:
only issue with that PR is some weird stuff could happen if you use the sender after the command block is removed
might want to add better exceptions/checks
also when a command block executes commands it does some extra logic, may or may not want to call into that or have a separate method that does, idk

Machine-Maker:
@jpenilla 
> only issue with that PR is some weird stuff could happen if you use the sender after the command block is removed might want to add better exceptions/checks

We don't have checks for this on entities right? You can hold a reference to an entity that no longer exists on the world but it can be used as a CommandSender.

> also when a command block executes commands it does some extra logic, may or may not want to call into that or have a separate method that does, idk

I think that should be a separate method. Basically a method that just directly runs the command execution logic directly. Not adding that here.

PR #8104, opened 2022-07-06
1.19 Anti-Xray cleanup

stonar96:


patches/server/0004-Paper-config-files.patch
```diff
@@ -1105,6 +1106,7 @@ index 0000000000000000000000000000000000000000..fa191f418079a8ee24326c5952d12d44
 +                    .register(DoubleOrDefault.SERIALIZER)
 +                    .register(BooleanOrDefault.SERIALIZER)
 +                    .register(Duration.SERIALIZER)
++                    .register(ChunkPacketBlockControllerAntiXray.EngineMode.SERIALIZER)
```
Machine-Maker: The reason I added this in the X-ray patch, is that now the server won't compile if you only some patches before X-ray.

patches/server/0353-Anti-Xray.patch
```diff
@@ -845,9 +845,7 @@ index 0000000000000000000000000000000000000000..a5d43901595f864c8a5dd5d013aa42d7
 +
 +        HIDE(1, "hide ores"),
 +        OBFUSCATE(2, "obfuscate");
-+
```
jpenilla, line 848: personally I think the whitespace here improves readability

patches/server/0353-Anti-Xray.patch
```diff
@@ -845,9 +845,7 @@ index 0000000000000000000000000000000000000000..a5d43901595f864c8a5dd5d013aa42d7
...
 +        public static final ScalarSerializer<EngineMode> SERIALIZER = new Serializer();
-+
```
jpenilla, line 850: and here

patches/server/0353-Anti-Xray.patch
```diff
@@ -1051,11 +1031,12 @@ index 568ac283baf40e2a89f5b002ffd899eba8008ef2..23d76eb74a88610472aa0288559efbaa
          this.heightmaps = new CompoundTag();
  
          for(Map.Entry<Heightmap.Types, Heightmap> entry : chunk.getHeightmaps()) {
-@@ -43,7 +46,13 @@ public class ClientboundLevelChunkPacketData {
+@@ -43,7 +46,14 @@ public class ClientboundLevelChunkPacketData {
          }
  
          this.buffer = new byte[calculateChunkSize(chunk)];
 -        extractChunkData(new FriendlyByteBuf(this.getWriteBuffer()), chunk);
++
```
jpenilla, line 983: is this blank line intentional?

jpenilla (dismissed):
Anywhere that has a note about "make sure this isn't used on update" should be annotated with `@DoNotUse` in addition to `@Deprecated` so that the project will fail to pass `check` when any Paper or Minecraft code calls it.

bluelhf:
Does this PR account for new blocks with transparency, e.g. mangrove roots?

stonar96:
> Does this PR account for new blocks with transparency, e.g. mangrove roots?

Any issues?

Edit: Indeed, fixed.

patches/server/0004-Paper-config-files.patch
stonar96: As discussed on Discord in #paper-contrib I have moved the `EngineMode` stuff to the config patch.

patches/server/0353-Anti-Xray.patch
stonar96: Moved to config patch and done.

patches/server/0353-Anti-Xray.patch
stonar96: Yes, I always use blank lines to separate code blocks that have indentation. This style is used in the entire patch.

electronicboy (approved):

kennytv (approved):

PR #8109, opened 2022-07-07
Fix calling Player#chat asynchronously

Machine-Maker:


electronicboy (approved):

Spottedleaf (changes requested):
This changes command handling behaviour.

You need to clarify what fixing this is even supposed to do

electronicboy:
This fixes an issue where tryna chat off the main thread in a plugin causes an exception down the chain, that async boolean in retrospect is kinda dumb, maybe make it so that async is only set to true if off the main thread and the string doesn't start with a "/" is the "quick" fix for this i think

Spottedleaf:
The correct fix to that is to throw an exception immediately. The async chat event is defined to be synchronous for the fake player chat calls, so there's no option here to make this work.

Machine-Maker:
Oh that's described on the old deprecated javadoc for AsyncPlayerChatEvent. That should probably be put on the adventure event as well...

PR #8112, opened 2022-07-08
Refactor paper command

jpenilla:
Cleaned up some code + reduces merge conflicts from command patches stepping on each other

Currently subcommand labels and aliases are defined in PaperCommand, a possible change is to have a method on PaperSubcommand that returns labels/aliases. In my opinion it's fine as is, but just throwing the idea out there, open to feedback.

### DIFF
--- a/patches/server/0009-Paper-command.patch
+++ b/patches/server/0009-Paper-command.patch
@@ -4,102 +4,50 @@ Date: Mon, 29 Feb 2016 21:02:09 -0600
 Subject: [PATCH] Paper command
 
 
-diff --git a/src/main/java/com/destroystokyo/paper/PaperCommand.java b/src/main/java/com/destroystokyo/paper/PaperCommand.java
+diff --git a/src/main/java/io/papermc/paper/command/CommandUtil.java b/src/main/java/io/papermc/paper/command/CommandUtil.java
 new file mode 100644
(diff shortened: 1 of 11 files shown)


Owen1212055:
Can you make /paper dumpitem copy the nbt to clipboard if you click the message or something? Been meaning to add that, nice QOL thing.

jpenilla:
> Can you make /paper dumpitem copy the nbt to clipboard if you click the message or something? Been meaning to add that, nice QOL thing.

Did that + added the item id so that you can copy and paste directly into `/minecraft:give`

tadhgboyle:
A couple thoughts:
- `/paper debug help` seems redundant, since `/paper debug` gives the same output. I could see keeping this command here if more debug options are planned to be added in the future?
	- Same with `/paper entity ...`
- `/paper fixlight` is great, but it's not shown anywhere to the user that a `number` argument is valid, or what that number is used for?
- Maybe this has been discussed internally, but adding two new directories ("debug" (for chunk stuff) and "heap") to the root directory doesn't seem fantastic, it's pretty crowded already.
	- Maybe keep the "debug" one, and make subdirectories for chunk dumps and heap dumps?
- `/paper mobaps` only does the world the user is in at the time when a world is not specified, but `/paper chunkinfo` does all worlds when a world is not specified.

PR #5529, opened 2021-04-22
OfflinePlayer advancements

Machine-Maker:
Closes #5486

PR #8117, opened 2022-07-09
Add offline PDC API

### LINKED ISSUE #7288: Allow getting offline player's PDC

I want to be able to get a player's PDC container, offline or online.

either `Bukkit.getPDC(String username);` or `Bukkit.getPDC(UUID playerUUID);`

> Machine Maker: you'd have to manually load the player data nbt file and search for it yourself

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7288
<details>
<summary>Old description</summary>
Plugin can now gets the offline persistent data of a player. ~The current behavior is like the Machine's Advancement PR:
All setters re-save the data directly into the disk~
Note also that if spigot config disablePlayerDataSaving is enabled, the data will not be written
</details>

Plugin can now gets the offline persistent data of a player with a read-only view.


patches/server/0925-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,99 @@
...
+         return result;
+     }
++
++    @Override
++    public org.bukkit.persistence.PersistentDataContainer getPersistentDataContainer() {
++        Player player = getPlayer();
++        if(player != null) return player.getPersistentDataContainer();
++
++        CompoundTag result = getData();
++        if(result == null) return null; // never called
```
lynxplay: Am I missing something ?
This if definitely is possible if a server requests an offline player e.g. based on uuid that has not joined before.

```java
final OfflinePlayer dinnerbone = Bukkit.getOfflinePlayer(UUID.fromString("61699b2e-d327-4a01-9f1e-0ea8c3f06bc6"));
final PersistentDataContainer persistentDataContainer = dinnerbone.getPersistentDataContainer();
```
would yield a null pdc instance (unless dinnerbone played on your server before) which contradicts the `@NotNull` annotation on https://jd.papermc.io/paper/1.19/org/bukkit/persistence/PersistentDataHolder.html#getPersistentDataContainer().

patches/server/0925-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,99 @@
...
++
++        org.bukkit.craftbukkit.persistence.CraftPersistentDataContainer pdc = new org.bukkit.craftbukkit.persistence.CraftPersistentDataContainer(org.bukkit.craftbukkit.entity.CraftEntity.DATA_TYPE_REGISTRY) {
```
lynxplay: Might also want to override the `clear` method. Looking at #7505, that method might be used soon.
Could also just be part of the other PR tho.

patches/server/0925-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,99 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
+index 2a6c67634c31c332102d24bef293da1bacd0c000..424a5a8a8345c89dda54687aaf74be3b69f81ca3 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
+@@ -190,7 +190,7 @@ import org.bukkit.util.Vector;
+ 
+ public abstract class CraftEntity implements org.bukkit.entity.Entity {
+     private static PermissibleBase perm;
+-    private static final CraftPersistentDataTypeRegistry DATA_TYPE_REGISTRY = new CraftPersistentDataTypeRegistry();
++    public static final CraftPersistentDataTypeRegistry DATA_TYPE_REGISTRY = new CraftPersistentDataTypeRegistry(); // Paper private -> public
```
lynxplay: If you want, this could be in the access transformers instead of a manual change.
Up to you, I don't think we really have much of a preference.

patches/api/0387-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,38 @@
...
++    /**
++     * Returns a custom tag container capable of storing tags on the object.
++     *
++     * Note that the tags stored on this container are all stored under their
++     * own custom namespace therefore modifying default tags using this
++     * {@link org.bukkit.persistence.PersistentDataHolder} is impossible.
++     *
++     * @return the persistent metadata container
++     */
++    @NotNull
```
lynxplay: The javadocs should properly mention the side effects of an offline players persistent data container.
E.g. that every write operation (remove, set) triggers disk IO.

patches/server/0925-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,99 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/storage/PlayerDataStorage.java b/src/main/java/net/minecraft/world/level/storage/PlayerDataStorage.java
+index 601f8099f74e81c17600566b3c9b7a6dd39c9bcb..bf7c29c481539d796521de725e17a58955410866 100644
+--- a/src/main/java/net/minecraft/world/level/storage/PlayerDataStorage.java
++++ b/src/main/java/net/minecraft/world/level/storage/PlayerDataStorage.java
+@@ -47,6 +47,22 @@ public class PlayerDataStorage {
+         }
+ 
+     }
++    // Paper start
++    public void save(java.util.UUID uuid, CompoundTag tag) {
```
lynxplay: Unsure how to properly do this, this method is pretty much just a copy of the existing `PlayerDataStorage#save(Player)`.
Could reduce duplicate code here.

lynxplay (changes requested):

patches/server/0925-Add-Offline-PDC-API.patch
Machine-Maker: I'd leave it as a change here. I think we are trying to move away from making everything an AT, leaving that for larger things, or things that need to change visibility across inheritance.

patches/server/0925-Add-Offline-PDC-API.patch
Lulu13022002: Well not really i thought that the Bukkit.getOfflinePlayer call returns null for unknown player but i was wrong

patches/api/0387-Add-Offline-PDC-API.patch
Lulu13022002: Also do you think a closable persistent data can be better to avoid rewrite for each operation

u9g:
why wouldn't this just be setPersistentDataContainer and getPersis... like on itemMeta, that way you can write all your data before writing back to the file?

Lulu13022002:
In fact it's not just for this all methods in this class just save after write operation (see statistics)

electronicboy:
This is inheriently unsafe imho, and really needs to be backed by a proper async player data system (and proper flow control for plugin devs to be able to safely mess with this stuff without screwing over other data consumers)

Lulu13022002:
That's right, i see your point https://github.com/PaperMC/Paper/pull/5529#issuecomment-830155819, do you prefer that i make this system immutable waiting your async/thread safe implementation. Because this PR was just covered this feature so rewriting this part is out of this PR. Also i agree with the autocloseable part.

Lulu13022002:
I have fixed a bug when player data was already created using the offline PDC API and the that same player join for its first time (Thanks LoaiDev).
Also the interface is now AutoCloseable (no need to resave after each write op) but sadly need a cast due to how Player and OfflinePlayer are linked
 ```Java
OfflinePlayer p = Bukkit.getOfflinePlayer("Lulu13022002");
try(OfflinePersistentDataContainer pdc = (OfflinePersistentDataContainer) p.getPersistentDataContainer()) {
    pdc.set(KEY, PersistentDataType.INTEGER, 1);
}
```

File: patches/server/0933-Add-Offline-PDC-API.patch [hunk 6]
```diff
@@ -0,0 +1,212 @@
...
+     }
++    // check if an offline PDC has already been created before its first join
++    public static boolean checkOfflinePDC(Player player, CompoundTag nbt, boolean checkOnly) {
++        boolean result = checkOnly ? player.getBukkitEntity().searchAvailableData(nbt) : player.getBukkitEntity().readBukkitValues(nbt);
++        if (result && nbt.size() == 1) {
++            return true;
++        }
++        return false;
++    }
++    // Paper end
```
Machine-Maker: Be more strict with your `// Paper start` and `// Paper end` comments. Most of the time, they should only be surrounding actual diff, not a big chunk with several line changes inside it.

Machine-Maker (dismissed):

patches/server/0933-Add-Offline-PDC-API.patch
Lulu13022002: Resolved, removed some useless diff

patches/api/0391-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,65 @@
...
++     *
++     * <p>
++     * Note that the tags stored on this container are all stored under their
++     * own custom namespace therefore modifying default tags using this
++     * {@link org.bukkit.persistence.PersistentDataHolder PersistentDataHolder} is impossible.
++     *
++     * @return the persistent metadata container
++     */
++    @NotNull
++    org.bukkit.persistence.PersistentDataContainer getPersistentDataContainer();
```
lynxplay: Any reason this is not just returning `io.papermc.paper.persistence.OfflinePersistentDataContainer` then ? If you are supposed to always cast it ?

patches/api/0391-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,65 @@
...
++++ b/src/main/java/io/papermc/paper/persistence/OfflinePersistentDataContainer.java
+@@ -0,0 +1,15 @@
++package io.papermc.paper.persistence;
++
++import org.bukkit.persistence.PersistentDataContainer;
++
++/**
++ * A {@link AutoCloseable} Persistent Data Container
++ */
++public interface OfflinePersistentDataContainer extends PersistentDataContainer, AutoCloseable {
```
lynxplay: Are we expecting this interface to be used again ?
I am unsure when, but the general concept of a PDC that does not "write" until closed might come up again, at which point this naming may be too specific.

patches/server/0933-Add-Offline-PDC-API.patch
```diff
@@ -0,0 +1,215 @@
...
++    public boolean readBukkitValues(CompoundTag c) { // Paper
+         Tag base = c.get("BukkitValues");
+         if (base instanceof CompoundTag) {
+             this.persistentDataContainer.putAll((CompoundTag) base);
++            return true; // Paper
+         }
++        return false; // Paper
+     }
++    // Paper start
++    public boolean searchAvailableData(CompoundTag c) {
```
lynxplay: This method name basically says nothing. what available data is it searching ? This method is also more of a contains/has logic, not really a search.

lynxplay (commented):
Just some thoughts. In general, this PR got kinda more ugly than initially set out to be because it aims to support PDCs for players that have never joined and hence we need all this mess.

patches/api/0391-Add-Offline-PDC-API.patch
Lulu13022002: Yes ideally all data that are retrieved by a single file like the advancements and the statistics for the offline player must use a similar things. For the pdc i agree with you but i have no idea of a good name.

patches/api/0391-Add-Offline-PDC-API.patch
Lulu13022002: Like i have said before this interface is also used for the online player (CraftPlayer) and so an offline pdc doesn't make sense for online player

patches/server/0933-Add-Offline-PDC-API.patch
Lulu13022002: Yes, you've right what about hasBukkitValues/hasBukkitPersistentData name ?

patches/api/0391-Add-Offline-PDC-API.patch
lynxplay: Right, the permanent issue of players extending offline players, I forgot.
This is some ugly API ngl, the fact that Player would then return you a PDC that cannot be cast but offline players do is pretty fucked.

Lulu13022002:
I can't really see good usage of supporting offline PDC for player that has never join the server before. So if LoaiDev or anyone see a good use case then i will keep otherwise i will just drop it and document + ignore the data.

lynxplay:
I'd probably be in favour of this 👍 If this feature is required later on, more thought can be put into it but it is just way to ugly to hack in there.

Question is how we want to indicate whether or not the offline player has played before. Do we wanna throw an exception when closing the PDC ? Or just log a WARN into console ? Just quietly failing might be unfortunate.

Lulu13022002:
For now i have throw an exception when the pdc is closed because this can cause data loss and possible NPE in future. And these things are not tiny bug and shouldn't be ignored.

lynxplay (approved):
Generally, LGTM.
Moved the impl of getOrDefault back to the impl, as that is where it lives in upstreams impl.

Only thing left to discuss pre-merge imo is a potential javadocs warning about how the impl works, given the pdc getter itself does not do IO but every query to it does, which might throw people off compared to the usual behaviour of OfflinePlayer, which does all its IO on the method itself and then returns a snapshot (e.g. location or metadata lists)

PR #8119, opened 2022-07-10
Add PlayerItemConsumedEvent

nopjar:
Adding an event which is called after an item was consumed by a player, which allows building mechanics based on consumption. The `org.bukkit.event.player.PlayerItemConsumeEvent` does not allow those operations safely, as the event can still be cancelled or the consumed Item be changed.


chickeneer:
What is an example of operations that are unsafe? Or an example of something that this would enable that isn't feasible with the current PlayerItemConsumeEvent.

> the event can still be cancelled or the consumed Item be changed.
This is the whole purpose of the MONITOR level priority - is that all plugins agree to not alter the event during that stage (and it is last in order).

Not saying that there is no possible uses - I am just struggling to see the usefulness necessarily. Admittedly this is after the item has fully been used and finished - so there might be some new information available about the post-use state of the player/world. But not sure what that is useful for

nopjar:
You could rely on the Monitor-Priority, ofc. If we do, this PR is pretty useless, sure.

PR #8122, opened 2022-07-10
Don't use level random in entity constructors

### LINKED ISSUE #8116: Spawning bees in BlockPopulators causes crash due to concurrent LegacyRandomSource access.
### Expected behavior

Spawning bees in a BlockPopulator does not crash the game

### Observed/Actual behavior

Bees access the world's `random` instance, which causes issues when they are spawned in a `BlockPopulator`, as it is run during world generation, on the worldgen executor, concurrently to the main server thread. The `LegacyRandomSource` used by the world random throws an exception if it detects a race condition.

Stack trace (Server thread detects race condition):
```
java.lang.IllegalStateException: Accessing LegacyRandomSource from multiple threads
	at net.minecraft.util.ThreadingDetector.makeThreadingException(ThreadingDetector.java:79) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.world.level.levelgen.LegacyRandomSource.next(LegacyRandomSource.java:45) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.world.level.levelgen.BitRandomSource.nextInt(BitRandomSource.java:21) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.world.level.NaturalSpawner.getRandomPosWithin(NaturalSpawner.java:419) ~[?:?]
	at net.minecraft.world.level.NaturalSpawner.spawnCategoryForChunk(NaturalSpawner.java:210) ~[?:?]
	at net.minecraft.world.level.NaturalSpawner.spawnForChunk(NaturalSpawner.java:183) ~[?:?]
	at net.minecraft.server.level.ServerChunkCache.tickChunks(ServerChunkCache.java:1024) ~[?:?]
	at net.minecraft.server.level.ServerChunkCache.tick(ServerChunkCache.java:908) ~[?:?]
	at net.minecraft.server.level.ServerLevel.tick(ServerLevel.java:647) ~[?:?]
	at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1551) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.server.dedicated.DedicatedServer.tickChildren(DedicatedServer.java:446) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1414) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1187) ~[paper-1.19.jar:git-Paper-58]
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:302) ~[paper-1.19.jar:git-Paper-58]
```

Sometimes, a worldgen thread will detect the race condition and throw the exception.

### Steps/models to reproduce

1. Install this [test plugin](https://github.com/dfsek/BeeRandomAccessExample/releases/download/1.0.0/BeeRandomAccessExample-1.0-SNAPSHOT.jar) ([source](https://github.com/dfsek/BeeRandomAccessExample)) on a Paper server. This plugin simply adds a BlockPopulator that [spawns 10 bees in every chunk](https://github.com/dfsek/BeeRandomAccessExample/blob/master/src/main/java/com/dfsek/bees/BeeRandomAccessPlugin.java).
2. Run the server, join and fly around.

Sometimes the crash happens almost immediately, sometimes it takes a while.

### Plugin and Datapack List

Plugins (1): BeeRandomCrash

### Paper version

This server is running Paper version git-Paper-58 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: e294802)

### Other

This is a Minecraft issue, as Mojang spawns other entities during worldgen (this still happens even if spawning is moved to the `populateEntities` phase of worldgen, tested on Fabric).

In Terra, we solved this with [this patch](https://github.com/PolyhedralDev/Terra/blob/master/platforms/mixin-common/src/main/java/com/dfsek/terra/mod/mixin/fix/BeeMoveGoalsUnsynchronizedRandomAccessFix.java) to the bee goals. This would *slightly* change behavior, however the random is only used to determine the number of ticks to wait after spawning before attempting to find a hive, with a range of [0,10), so we believe the behavioral change is negligible. If this fix is acceptable, I would be happy to open a PR contributing it to Paper.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/8116

There are several cases where the entity constructor uses the thread-unsafe random on Level in the constructor. This hampers plugins ability to create entity instances off the main thread for world gen. Using the entity's random is fine, since paper already makes that thread safe as opposed to vanilla where it's unsafe.

I only found 3 instances of this, the 2 bee places from the linked issue, then one in ItemEntity and one in PrimedTnt. 

Just a note, you can't create EnderDragon instances off the main thread since that tries to fire EnderDragonChangePhaseEvent asynchronously when the event isn't marked as such (throwing an exception).

I don't see any apparently problem with using the entity's random over the level's. Entity constructors also use their own in places, so there isn't a pattern that I see.


electronicboy (approved):

MarijnIsN00B:
I hope this is gonna be fixed soon, these constant crashes are horrible

Vulpicula:
Hai, not sure if this is the place to say this, but if this could be merged sooner rather than later, it'd be much appreciated. This is a pretty big issue for my server, and a few others who use a certain world generation plugin.

Wound up manually building a version of this with the updated patch as a workaround, but I'm not sure if everyone is able to do that. (If anyone wants it, I can probably just share the link around. >.>)

electronicboy:
it's got my approval, just needs rebasing and such

PR #8123, opened 2022-07-11
Properly disallow async Player#chat

Machine-Maker:
properly catch chat from `Player#chat` (and any other plugins that might call `ServerGamePacketListenerImpl#chat` with async set to false while not on the main thread)

Also clarify asynchronous status of AsyncChatEvent (javadocs taken from AsyncPlayerChatEvent)


SpiritOTHawk (approved):

Machine-Maker:
Rebased for 1.19.2

Owen1212055 (approved):

lynxplay (approved):

PR #8125, opened 2022-07-11
Add copper clear on lightning strike to EntityBlockFormEvent

chrisgdt:
Hello,

I basically wanted to detect when a copper block is being cleared by a lightning strike, but I realized that there is no event to detect it. So I added this patch to call EntityBlockFormEvent everytime such a block has changed.

I didn't know if I had to call BlockForm or modify LightningStrikeEvent, but EntityBlockFormEvent seems to be the most relevant event to call. I kept static methods to minimize the diff but I can remove these static and use only `this` instead of transmit the lightning instance. I also changed the way it changes the block to call the event more easily, if it generates too much lag or it is not good, I also can change it and only keep the event call then the way it changes the block before. However, I tested the patch on my local server and everything works as attended (cancel, not cancel so copper is cleared).

### DIFF
--- a/patches/server/0925-Add-copper-clear-on-lightning-strike-to-EntityBlockF.patch
+++ b/patches/server/0925-Add-copper-clear-on-lightning-strike-to-EntityBlockF.patch
@@ -1,0 +1,75 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: ChristopheG <61288881+chrisgdt@users.noreply.github.com>
+Date: Mon, 11 Jul 2022 20:39:05 +0200
+Subject: [PATCH] Add copper clear on lightning strike to EntityBlockFormEvent
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/LightningBolt.java b/src/main/java/net/minecraft/world/entity/LightningBolt.java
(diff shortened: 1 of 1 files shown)


Owen1212055:
What about the ``EntityChangeBlockEvent``? That might make more sense here. EntityBlockFormEvent makes it seem like it's "creating" new blocks when in reality it's just being changed.

chrisgdt:
Well, I didn't find this event because I have never used it and I look for a BlockEvent, not EntityEvent... But the fact that BlockFormEvent is called when a copper oxidizes itself put me on the way to use EntityBlockFormEvent.

Should I really modify the patch to call EntityChangeBlockEvent instead of EntityBlockFormEvent ?

Owen1212055:
> Well, I didn't find this event because I have never used it and I look for a BlockEvent, not EntityEvent... But the fact that BlockFormEvent is called when a copper oxidizes itself put me on the way to use EntityBlockFormEvent.
> 
> Should I really modify the patch to call EntityChangeBlockEvent instead of EntityBlockFormEvent ?

Copper oxidizing isn't caused by any entity, just by its existence in the world, so BlockFormEvent would make sense here (they also seem to be different). Copper clearing due to lightning is very much caused by the lightning entity. So... yes, I think you should.

chrisgdt:
I don't really want to modify this patch so I will recreate a new pull request. Thank you !

PR #8127, opened 2022-07-12
Fix typo and misc improvements in Paper configuration

montlikadani:
https://github.com/PaperMC/Paper/commit/172d260d677a05c59e9e8f3f6bfe9216063eef72#r78280454

_Close this PR if you decide to leave that typo as it is._


BillyGalbreath:
This is not a typo.. It is proper English. The permissions message is not talking about _[an error](https://www.merriam-webster.com/dictionary/error)_, it is talking about if you feel the message is _[in error](https://www.merriam-webster.com/dictionary/in%20error)_. Two totally different meanings.

SoSeDiK:
Maybe it would still be better to use a direct word instead of this idiom?

People seem to be confused by this. I, as a non-native English speaker, never heard of this before as well.

I mean in cases like "we're on the same page"/"bite the bullet" you can ~~feel~~ spot the idiom, but with "in error" it's not that clear.

Issue #2541, opened 2019-09-06
Trident doesn't respect AbstractArrow


Proximyst:
### What behaviour is expected:
The trident to do no damage.

### What behaviour is observed:
Trident completely ignored all of the properties.

### Steps/models to reproduce:
Listen to `ProjectileHitEvent` (send a message or something to make sure you know it's listened to), then something along the lines of the following Kotlin snippet
```kotlin
        val projectile = event.entity
// code...
        if (projectile is AbstractArrow) {
            if (projectile is Arrow) {
                projectile.clearCustomEffects()
            }
            projectile.fireTicks = 0 // not known if ignored
            projectile.knockbackStrength = 0 // ignored
            projectile.damage = 0.0 // ignored
            projectile.isCritical = false // not known if ignored
            projectile.pierceLevel = 0 // not known if ignored
        }

        if (main.config.pvpMessage) {
            val msg = when {
                damagerState -> main.messages.pvpNoPvpTarget
                damagedState -> main.messages.pvpNoPvpHitter
                else -> main.messages.pvpNoPvpHitter
            }
            shooter.sendMessage(
                msg.process("target" to damaged)
                    .colour()
            )
        }
```

### Plugin list:
Only a plugin doing this. The project isn't touched in any other way.

### Paper build number:
![https://owo.whats-th.is/2wJBwV6.png](https://owo.whats-th.is/2wJBwV6.png)

kickash32:
According to the code, all of those values are ignored. This is because a Trident does not act like an arrow at all when it hits an entity. A trident cannot be on fire so fireTicks does not make any sense. Tridents cannot have knock back enchantment so they are unable to do any knockback. The damage value is hardcoded to 8 and tridents are unable to cause critical hits. And finally Tridents cannot pierce, which also does not make much sense but is less farfetched than a flaming trident since it doesn't need any fire animation.

Most of this can be changed but I'm wondering if you only need to be able to change the damage.

Proximyst:
@kickash32 Ideally, it'd just respect that I cancel the entire event altogether, and simply do not touch the player it hits (I forgot to include that part in the snippet it seems). If it has to be that way, I'm entirely fine with it not damaging the player and still applying knockback, however it would be ideal if it didn't do either.

kickash32:
Have you tried cancelling the ProjectileCollideEvent? This event should be called before the ProjectileHitEvent and it should be cancellable.

Proximyst:
[...]

PR #8132, opened 2022-07-12
Allow custom damage for trident

Lulu13022002:
Partial implementation of #2541
Remove the hardcoded damage value ~and allow critical damage (for plugin)~
I think plugin can implement the other things like knockback etc themselves

### DIFF
--- a/patches/server/0925-Allow-trident-custom-damage.patch
+++ b/patches/server/0925-Allow-trident-custom-damage.patch
@@ -1,0 +1,46 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
+Date: Tue, 12 Jul 2022 18:01:14 +0200
+Subject: [PATCH] Allow trident custom damage
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java b/src/main/java/net/minecraft/world/entity/projectile/AbstractArrow.java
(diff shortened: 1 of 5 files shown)


Owen1212055:
A trident can’t critical, it doesn’t really make sense why the critical api would work like this. Yes, it shares many properties cause AbstractArrow, but it isn’t an arrow. Like llamas for example, they share a lot of code with horses, you can make them do a lot of things that horses would normally do but they mostly don’t do anything.

Lulu13022002:
I know actually the critical things is more like a marker for plugin. But the damage make sense for plugin to avoid reimpl the enchantment logic

patches/server/0925-Allow-trident-custom-damage.patch
```diff
@@ -0,0 +1,46 @@
...
+     @Nullable
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/ThrownTrident.java b/src/main/java/net/minecraft/world/entity/projectile/ThrownTrident.java
+index fd4c4aef35f5412fe7df1049b3f2f7a75b655249..77d3544e282f3e7feabc17bc62a41f7152678728 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/ThrownTrident.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/ThrownTrident.java
+@@ -33,11 +33,13 @@ public class ThrownTrident extends AbstractArrow {
+ 
+     public ThrownTrident(EntityType<? extends ThrownTrident> type, Level world) {
+         super(type, world);
++        this.baseDamage = 8.0D; // Paper
```
kennytv: net.minecraft.world.item.TridentItem.BASE_DAMAGE

kennytv (approved):

PR #6371, opened 2021-08-10
fire EntityChangeBlockEvent in more places

Machine-Maker:
Closes #6348

PR #8136, opened 2022-07-13
Call EntityChangeBlockEvent for frogs egg

Lulu13022002:
Call EntityChangeBlockEvent when frogs lay eggs in the lake


Owen1212055 (approved):
Generally, I would wait for https://github.com/PaperMC/Paper/pull/6371/files to go through so then you can just modify that patch.

Looks good tho.

Lulu13022002:
Rebased into the said patch ^

electronicboy (approved):

Owen1212055:
That was most likely not intended seeing as though it was in a mc update commit, i'd recommend pulling that.

PR #8139, opened 2022-07-13
fix #8138

### LINKED ISSUE #8138: Redstone wire going up blocks does not update when powered
### Expected behavior

when the redstone is already powered and you connect a higher level piece of redstone to a lower level, the higher piece of redstone should have its texture update.

### Observed/Actual behavior

The redstone on the higher blocks textures do not update, however, their behavior does. [
image
](image)

### Steps/models to reproduce

place a block and put redstone on top of it

power the redstone

place redstone next to the block that would normally change the direction it points

watch the bug happen

### Plugin and Datapack List

AnvilColor, ColoredSigns, CoreProtect, DiscordSRV, dynmap, Dynmap-GriefPrevention, Essentials, EssentialsChat, EssentialsSpawn, FakeLeaveAndJoin, GriefPrevention, GSit, GuestMode, HasteBoost, PermissionsEx, PetMaster, PortalHelper, SafeSpectate, SmoothSleep, StaffChat, TalkingBot, Vault, ViaBackwards, ViaVersion, WorldBorder, WorldEdit, Wrenchable

### Paper version

1.19-58

### Other

_No response_

SpaceWalkerRS:
The recent update to the Eigencraft patch introduced a [bug](https://github.com/PaperMC/Paper/issues/8138) that saw indirect shape updates omitted entirely, which can leave newly placed wire blocks in an invalid state (not updating their connections). This PR re-introduces those indirect shape updates to fix that issue. This does result in a slight performance penalty (or rather, it undoes the performance gain from the recent update).

### DIFF
--- a/patches/server/0486-Eigencraft-redstone-implementation.patch
+++ b/patches/server/0486-Eigencraft-redstone-implementation.patch
@@ -20,10 +20,10 @@ Just added Bukkit's event system and took a few liberties with dead code and com
 
 diff --git a/src/main/java/com/destroystokyo/paper/util/RedstoneWireTurbo.java b/src/main/java/com/destroystokyo/paper/util/RedstoneWireTurbo.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..3e709645f5c35d16beefa4328a0a8e9c95e632aa
+index 0000000000000000000000000000000000000000..22a2547810d0c029f29685faddf7ac21cde2df0b
 --- /dev/null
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

molor:
_(very quiet: port to 1.18.2 please)_

SpaceWalkerRS:
That will not be necessary, as this bug only affected 1.19 Paper.

PR #8147, opened 2022-07-16
Add signature info to async chat event

Rigner:
Added a way to get the signature info from client chat messages.

Kept the original component as-is, to make sure it's the exact same content as sent by the client, since it may be used for validation.

A future improvement could be to forward this to the receivers if the content wasn't modified by plugins (or as unsigned content if the server wants to).


lynxplay:
Welcome to paper :)
Thanks for the PR too!

However there is no point in opening a PR for a feature that is still heavily under development by mojang at this point in time.
Beyond that, adventure is already working on signature related classes that paper will be using once available.

I would heavily suggest to leave the changes towards chat api for 1.19+ a) to a point where mojang finalized the system and b) to the core/developer team.

electronicboy:
Don't particularly mind if somebody gets to it before us, but, this is really pending on adventure to prevent headaches down the road

Rigner:
Hey, thanks for the feedback, didn't know adventure was already working on it! I guess I'll wait for that then!

_(Can't wait to have a way to delete signed messages through the 1.19.1 changes)_

PR #8150, opened 2022-07-17
Add EntityStopTickingEvent

ghost:
My first PR, hopefully it isn't too fucked up. Called for when entities move into non ticking chunks, for example if you shoot an arrow too far it just stops mid-air.


patches/api/0387-Add-EntityStopTickingEvent.patch
```diff
@@ -0,0 +1,58 @@
...
+             }
+             // Paper end
++
++            // Paper start - Call EntityStopTickingEvent
++            getCraftServer().getPluginManager().callEvent(new io.papermc.paper.event.entity.EntityStopTickingEvent(entity.getBukkitEntity()));
++            // Paper end
+         }
+ 
+         public void onTrackingStart(Entity entity) {
+
```
NoahvdAa: 👋 Welcome to Paper, congrats on making your first PR!

This section should be in a server patch, since the implementation doesn't exist as far as the API is concerned.

PR #8152, opened 2022-07-17
Configurable shutdown and restart process

### LINKED ISSUE #7446: Expose abnormal exit path to plugins.

I have a paper instance operating inside of a wrapper (presumed docker container) provided by a shared hosting provider.

The host offers configuration, as to whether the server should be automatically restarted on crash, and (presumably) keeps track of time since the last crash to prevent crash loops.

It does this via monitoring the exit codes of the various game servers it's able to host.

However it's currently not able to detect whether a plugin induced shutdown is considered normal or abnormal.

It would be useful, if there were a method of communicating shutdown intent, to the wrapper of the server, from plugins.

E.g. 

One plugin could offer a `/plannedrestart` command, that would notify users that an upcoming unscheduled but planned restart is about to occur. Once the server is shutdown, the wrapper could recognize there was intent to restart, and restart the server.

Another command that could be offered `/emergencystop` which would perform an immediate shutdown of the server, with no intent to restart, and no warning to players, or the same mechanism be  invoked within a plugins onEnable if a fatal configuration error had occurred with say, a database connection string for a large minigames network.

I would like to see plugins able to call a `Server.abnormalShutdown()` command that would indicate an error state, via an exit code, much like the watchdog is currently able to produce.

Also, an ability to call `Server.restart()` which depending on configuration, would either invoke the restart script configured, or if no restart script configured, exit with some form of signalling that a wrapper could pick up on.

Whether exitcode `0` is used for for a normal shutdown without restart, or a normal shutdown with restart, or both with some other method of signalling (making sure the last thing posted to the `stderr` or `stdout` pipe some form of restart sentinel value?)

It feels wrong to use a non zero exit code to indicate a non errored shutdown of a process, but maybe there is precedent elsewhere? It certainly would feel like the most natural method of signalling available.

Spigots `/restart` command is awkward, and slightly re-entrant in that it attempts to start the new process before the old one has ended. It's liable to hit strange and awkward bugs, and doesn't seem to be compatible with how some basic docker containers are configured without a explicit startup script, making assumptions about what shells are installed based on whether the os contains 'win'.

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7446

This allow more flexibility for plugins to restart or shutdown the server with other external app:
- Exit the process with a non zero exit code (also plugin should avoid the 70 exit code already used for watchdog)
=> server.abnormalShutdown(exitCode)

- Restart the process without a script but using a runnable this allow plugin to do more things without adding a new shutdown hook. Also plugin can easier choose the right executing order 
=> server.restart(() ->...)

Keep in mind that theses functions must be called on the **main thread** (to avoid background process, data loss and file lock).

Also i have added a new config option: predictShellOnRestart (true by default)
By default when server restart spigot predict which shell will be used for windows and "others" but this behavior can cause some trouble in container or when the predicted shell is not available for X reason


Owen1212055:
> Keep in mind that theses functions must be called on the main thread (to avoid background process, data loss and file lock).

Perhaps then this should be documented?

And if we don't want plugins using the watchdog exit code, perhaps log when a plugin does this? See StackWalkerUtil to get the caller plugin

patches/server/0981-Configurable-restart-and-shutdown-exit-code.patch
```diff
@@ -0,0 +1,166 @@
...
++        org.spigotmc.AsyncCatcher.shuttingDown = true; // Paper
++        try {
++            // Paper - extract method and cleanup
++            addRawShutdownHook(onExit);
++            System.out.println("Attempting to restart with a plugin custom hook");
++            // Stop the watchdog
++            WatchdogThread.doStop();
++
++            shutdownServer(true);
++            // Paper end
```
Warriorrrr, line 105: Forgot to remove a sysout here, also paper comments inside paper comments

patches/server/0005-Paper-config-files.patch
```diff
@@ -719,6 +719,7 @@ index 0000000000000000000000000000000000000000..52b02cb1f02d1c65b840f38cfc8baee5
 +        public boolean lagCompensateBlockBreaking = true;
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
++        public boolean predictShellOnRestart = true;
```
Warriorrrr, line 722: I'm wondering whether it might be nice to have a system property for this instead of an option so that it can be set by hosts, though I doubt many of them use the restart command

PR #8156, opened 2022-07-18
Add config setting for grindstone overstacking

uku3lig:
Title pretty much says it all.

Overstacking cursed books can be very useful for item filters, and I have played on multiple servers running Paper which sadly couldn't use this bug/inconsistency.

The maintenance cost should be pretty low, since it's just a config value.


lynxplay (dismissed):
Hi! 

Welcome to paper :smile: 
Thank you for your first PR, always great to see new faces on here :)

Generally a few issues here are the patches that you used.
Added configuration now lives in the https://github.com/PaperMC/Paper/blob/master/patches/server/0004-Paper-config-files.patch patch, so please move the addition of your configuration value into said patch.

Additionally, the behavioral change introduced to the grindstone menu should most likely live in the patch that introduced the paper limitation for this itself. https://github.com/PaperMC/Paper/blob/master/patches/server/0600-Prevent-grindstones-from-overstacking-items.patch specifically :+1:

uku3lig:
Thanks for the help, it took me a bit of time to figure out how to do everything correctly (hopefully :sweat_smile:), everything should be good now.

Machine-Maker (approved):

kennytv (approved):

Owen1212055 (approved):

PR #8161, opened 2022-07-19
Make Biome implement Translatable

NoahvdAa:
Works for every biome except Biome.CUSTOM, not sure how to handle that.


Machine-Maker:
Maybe add a test like the other translation keys?

NoahvdAa:
> Maybe add a test like the other translation keys?

Added :)

kennytv (approved):

PR #8165, opened 2022-07-20
Add SpawnReason to Tadpoles spawned by Frogspawn

### LINKED ISSUE #8164: tadpoles from frogspawn missing spawnreason
### Expected behavior

There should be a spawnreason when tadpoles are spawned from frogspawn.

### Observed/Actual behavior

When tadpoles are spawned from frogspawn they will be spawned with the spawnreason default on the CreatureSpawnEvent

### Steps/models to reproduce

-

### Plugin and Datapack List

-

### Paper version

> paper version
[17:07:32 INFO]: Checking version, please wait...
[17:07:32 INFO]: This server is running Paper version git-Paper-66 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: b7faa65)

### Other

_No response_

Doc94:
Close #8164 like mention lynx this is like the spawn of turtles then the spawnreason is the "EGG".

### DIFF
--- a/patches/server/0925-Add-SpawnReason-to-Tadpoles-spawned-by-Frogspawn.patch
+++ b/patches/server/0925-Add-SpawnReason-to-Tadpoles-spawned-by-Frogspawn.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Wed, 20 Jul 2022 12:07:36 -0400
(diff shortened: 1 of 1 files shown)


lynxplay:
Note, this is purely my opinion. I don't think a hyper specific spawn reason for a "natural" spawn reason is really the point of the enum.

This also still has the issue of upstream potentially breaking this when they do it if they don't follow my reasoning here.

kennytv (approved):

PR #8168, opened 2022-07-21
Fix errors on occurred when updating plugins

monun:
FileAlreadyExistsException occurred when the file name of the plugin being updated was the same as that of the existing plugin.

`java.nio.file.Files#copy` causes an exception when a file with the same name exists.

Use temporary files to move plugin files in a safe way.

#7960


emilyy-dev:
Shouldn't simply using `SCO.REPLACE_EXISTING` work fine? Instead of doing the whole dance around a temp file

emilyy-dev:
On a second thought... wouldn't/couldn't the whole logic be replaced with
```java
Files.delete(file);
Files.move(updateFile, pluginsDir.resolve(updateFile.getFileName()), StandardCopyOptions.REPLACE_EXISTING);
```

electronicboy:
Thanks for the PR however, for minimal dif, I just decided it'd be best to just restore the old mechanism essentially vs tryna juggle more file ops around

PR #8170, opened 2022-07-21
Separate Command Sending to Separate Thread Pool

Owen1212055:
See: https://canary.discord.com/channels/289587909051416579/925530366192779286/999447366639829093

I'm not sure if I should shutdownNow or wait, would like feedback here.
But basically, the idea is to shutdown and stop command sending from being done async and not being closed till very later in the server saving process.

There might be more as to what is causing exceptions like this, but shutting it down might be fine here?


darbyjack (approved):

electronicboy (approved):

Owen1212055:
Rebased

PR #8176, opened 2022-07-23
Handle rotation for PlayerMoveEvent

### LINKED ISSUE #8144: Overriding player movement in PlayerMoveEvent causes a PlayerTeleportEvent
### Expected behavior

If I override player movement in a `PlayerMoveEvent`, it should not call a `PlayerTeleportEvent`. I believe this is a genuine oversight, while it is niche I do believe it needs to be addressed for consistency

```java
@EventHandler
public void onPlayerMove(PlayerMoveEvent event) {
        // Why does this call PlayerTeleportEvent ?
        Location to = event.getFrom();
        to.setPitch(event.getTo().getPitch());
        to.setYaw(event.getTo().getYaw());
        event.setTo(to);
}
```

### Observed/Actual behavior

When using the above logic to override the player movement, it calls `PlayerTeleportEvent`, 

### Steps/models to reproduce

```java
public class Main extends JavaPlugin implements Listener {
    private static boolean shouldFreeze = false;

    @Override
    public void onEnable() {

        getServer().getScheduler().runTaskTimerAsynchronously(this, () -> {
            shouldFreeze = !shouldFreeze;
        }, 20, 100);
    }

    public void onPlayerTeleport(PlayerTeleportEvent event) {
        getServer().broadcastMessage(event.getCause().toString());
    }

    @EventHandler
    public void onPlayerMove(PlayerMoveEvent event) {
        if(shouldFreeze) {
            // Why does this call a teleport event?
            Location to = event.getFrom();
            to.setPitch(event.getTo().getPitch());
            to.setYaw(event.getTo().getYaw());
            event.setTo(to);
        }
    }

}
```

### Plugin and Datapack List

N/A

### Paper version

[01:23:50 INFO]: This server is running Paper version git-Paper-61 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: cd215af)

### Other

_No response_

Doc94:
This close #8144 by using the new teleport API for handle cases where in PlayerMoveEvent the event only change the rotation of the Player.

### DIFF
--- a/patches/server/0886-More-Teleport-API.patch
+++ b/patches/server/0886-More-Teleport-API.patch
@@ -5,10 +5,36 @@ Subject: [PATCH] More Teleport API
 
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index 421d7e28370085dffa415d71d4e28d3c4d38ac87..92617a19fb328320e17b58956cf0276d5a08325b 100644
+index 421d7e28370085dffa415d71d4e28d3c4d38ac87..23c518d3d0689949e5c28cddc20a5b10daa68b99 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-@@ -1729,11 +1729,17 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+@@ -760,6 +760,12 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                         // there to avoid any 'Moved wrongly' or 'Moved too quickly' errors.
+                         // We only do this if the Event was not cancelled.
+                         if (!oldTo.equals(event.getTo()) && !event.isCancelled()) {
++                            // Paper start
++                            if (oldTo.getWorld().equals(event.getTo().getWorld()) && oldTo.getX() == event.getTo().getX() && oldTo.getY() == event.getTo().getY() && oldTo.getZ() == event.getTo().getZ()) {
++                                this.player.getBukkitEntity().setRotation(event.getTo().getYaw(), event.getTo().getPitch());
++                                return;
++                            }
++                            // Paper stop
+                             this.player.getBukkitEntity().teleport(event.getTo(), PlayerTeleportEvent.TeleportCause.PLUGIN);
+                             return;
+                         }
+@@ -1613,6 +1619,12 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                                         // there to avoid any 'Moved wrongly' or 'Moved too quickly' errors.
+                                         // We only do this if the Event was not cancelled.
+                                         if (!oldTo.equals(event.getTo()) && !event.isCancelled()) {
++                                            // Paper start
++                                            if (oldTo.getWorld().equals(event.getTo().getWorld()) && oldTo.getX() == event.getTo().getX() && oldTo.getY() == event.getTo().getY() && oldTo.getZ() == event.getTo().getZ()) {
++                                                this.player.getBukkitEntity().setRotation(event.getTo().getYaw(), event.getTo().getPitch());
++                                                return;
++                                            }
++                                            // Paper stop
+                                             this.player.getBukkitEntity().teleport(event.getTo(), PlayerTeleportEvent.TeleportCause.PLUGIN);
+                                             return;
+                                         }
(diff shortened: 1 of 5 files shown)


patches/server/0926-More-Teleport-API.patch
```diff
@@ -5,10 +5,36 @@ Subject: [PATCH] More Teleport API
...
+index 9d89da66c0c5ab07252430eaff6d20e5e83db658..c869b6948db38b8e823a22b5a82d239b1fe90361 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-@@ -1660,11 +1660,17 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Ser
+@@ -719,6 +719,12 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Ser
+                         // there to avoid any 'Moved wrongly' or 'Moved too quickly' errors.
+                         // We only do this if the Event was not cancelled.
+                         if (!oldTo.equals(event.getTo()) && !event.isCancelled()) {
++                            // Paper start
++                            if (oldTo.getX() == event.getTo().getX() && oldTo.getY() == event.getTo().getY() && oldTo.getZ() == event.getTo().getZ()) {
```
Machine-Maker: This doesn't handle if the world changed. Theoretically I could change the world in the PlayerMoveEvent, and the two locations would not be equal (because the world was different), but everything else could be equal, in which case the player wouldn't change worlds.

patches/server/0926-More-Teleport-API.patch
```diff
@@ -5,10 +5,36 @@ Subject: [PATCH] More Teleport API
...
++                            // Paper end
+                             this.player.getBukkitEntity().teleport(event.getTo(), PlayerTeleportEvent.TeleportCause.PLUGIN);
+                             return;
+                         }
+@@ -1544,6 +1550,12 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Ser
+                                         // there to avoid any 'Moved wrongly' or 'Moved too quickly' errors.
+                                         // We only do this if the Event was not cancelled.
+                                         if (!oldTo.equals(event.getTo()) && !event.isCancelled()) {
++                                            // Paper start
++                                            if (oldTo.getX() == event.getTo().getX() && oldTo.getY() == event.getTo().getY() && oldTo.getZ() == event.getTo().getZ()) {
```
Machine-Maker: See other comment about changing worlds

You can use `PlayerMoveEvent#hasChangedBlock` as shorthand for this conditional too.

Machine-Maker (changes requested):

patches/server/0926-More-Teleport-API.patch
Doc94: hasChangedBlock only let me know if the from/to are different but i need check if the "to" originally send in event its the same than to set by user (world,x,y,z) then this short not works for this case.

Doc94:
Example of this patch using.
```
    public void onPlayerTeleport(PlayerTeleportEvent event) {
        getServer().broadcastMessage(event.getCause().toString());
    }
```
```    @EventHandler
    public void onPlayerMove(PlayerMoveEvent event) {
        // Why does this call a teleport event?
        getServer().broadcastMessage("test move: [" +  event.getFrom() + "] [" + event.getTo() + "]");
        if (event.getPlayer().getInventory().getItemInMainHand().getType().equals(Material.STICK)) {
            Location to = event.getFrom();
            to.setPitch(90);
            //to.setYaw(90);
            event.setTo(to);
        } else if (event.getPlayer().getInventory().getItemInMainHand().getType().equals(Material.SLIME_BALL)) {
            Location to = event.getFrom();
            to.setWorld(getServer().getWorld("world_nether"));
            to.setPitch(90);
            //to.setYaw(90);
            event.setTo(to);
        }
    }
```
https://youtu.be/mHXRMvHTpEM

patches/server/0926-More-Teleport-API.patch
Doc94: Added the world validation

Doc94:
Rebased and now are two files (thanks indexs) :3

Doc94:
Friday of PR rebased.

Doc94:
and conflicts gone again.

Owen1212055:
Generally I’m a bit iffy as this is a major behavior change. Unfortunately I’m not sure it’s worth making a breaking change like this?

Doc94:
Well this breaking is very situational (?
Because only handle the old form in spigot to make a teleport for any variation in the destination, this change is more to reduce this teleportation for rotations.

Owen1212055:
Yes, it's situational but you cannot really tell if people may have relied on this behavior or not. What I would instead do is perhaps make a new event, maybe PlayerMoveHeadEvent or something idk.

The PlayerMoveHeadEvent could act as a way to allow plugins to prevent players from moving their head while still allowing to keep velocity the same. But again, I think that changing the behavior of this event like that isn't the best here.

PR #8179, opened 2022-07-23
Add FurnaceBurnEvent.setBurnTime(short burnTime)

### LINKED ISSUE #8178: Values above SHORT.MAX_VALUE for org.bukkit.event.inventory.FurnaceBurnEvent#setBurnTime do not play burning animations inside the furnace UI for players
### Expected behavior

The furnace should play a burning animation for the furnace inside the furnace UI window

### Observed/Actual behavior

No furnace burning animation is played to show that the furnace is fueled, however the furnace continues to operate

### Steps/models to reproduce

Use org.bukkit.event.inventory.FurnaceBurnEvent#setBurnTime and set it to a value substantially higher than Short.MAX_VALUE, observe that the furnace no longer plays an animation inside the UI window

### Plugin and Datapack List

This was tested by modifying mcMMO source code, but the bug is inside Bukkit, the burnTime of a furnace should be a short (as that is how it is saved in NBT) yet instead allows an integer

### Paper version

This server is running Paper version git-Paper-77 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 44b6869)

### Other

A simple fix without breaking the API would be to apply Math.min(Short.MAX_VALUE, burnTime) in the implementation of setBurnTime

nossr50:
This also deprecates setBurnTime(int burnTime) and fixes its implementation as well
Fixes #8178

### DIFF
--- a/patches/api/0389-Add-FurnaceBurnEvent-setBurnTime-short-burnTime.patch
+++ b/patches/api/0389-Add-FurnaceBurnEvent-setBurnTime-short-burnTime.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: nossr50 <nossr50@gmail.com>
+Date: Sat, 23 Jul 2022 15:21:43 -0700
+Subject: [PATCH] Add FurnaceBurnEvent::setBurnTime(short burnTime) Fixes #8178
+
+
+diff --git a/src/main/java/org/bukkit/event/inventory/FurnaceBurnEvent.java b/src/main/java/org/bukkit/event/inventory/FurnaceBurnEvent.java
+index caef53d0f6546516fa7aabb2cb3abed70808b3ba..9581e66a984d5188e9c50f2dcc1dd3bae15e930b 100644
+--- a/src/main/java/org/bukkit/event/inventory/FurnaceBurnEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/FurnaceBurnEvent.java
+@@ -49,8 +49,19 @@ public class FurnaceBurnEvent extends BlockEvent implements Cancellable {
+      * Sets the burn time for this fuel
+      *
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Is there a better way? There's no way to write short literals in java right? So you'd now have to do 
```java
furnace.setBurnTime((short) 40);
```
just to avoid it using the deprecated integer method.
maybe just throw exception if its too big to be converted to a short?

emilyy-dev:
Even if the int variant wasn't there you need to either cast to a short or put the number into a short variable to use the method (more specifically, narrowing conversions are not allowed in a method invocation context, and int literals are, well, `int` literals).
Maybe put a `@Range` annotation on the int parameter? I think clamping the value is a reasonable approach, exceptions are nasty

Machine-Maker:
> Even if the int variant wasn't there you need to either cast to a short or put the number into a short variable to use the method

Oh I figured in the absence of another method that took an int, it'd accept a regular int in a short parameter.

nossr50:
I don't see the issue with having to cast to short since it's new API and not a change of an existing signature.

emilyy-dev:
Casting is actively annoying and, when you have a method that does not require casting, it's very easy to miss and forget.
Others might think differently but those are my two cents.

lynxplay:
I'd also be in favour of just limiting the setBurnTime(int)  method to Short.MAX_VALUE like done right now for backwards compat.
The getter is returning an integer. As pointed out by emily, the @Range annotation could be added to actively inform developers about the constraint on the values even without reading through javadocs.

A new method that requires developers to add rather ugly casts for little to no benefit seems overkill for a problem this small.

Owen1212055 (changes requested):
Many of NMS' values are represented by integers by the API. Instead, just Math.min (or clamp it to the min/max short) on the provided integer. This would help avoid creating a new method like this.

PR #8183, opened 2022-07-24
Fix entity exit position when teleport using portals

### LINKED ISSUE #8182: entity in the end portal come out 0.5 blocks off
### Expected behavior

I send entity through end portal, it comes out in center of middle block of end platform.

### Observed/Actual behavior

Entity comes off not at `.50` but at `.00` exact x z coord

### Steps/models to reproduce

send falling block (this is block so easy to see) through portal with x/z and +y velocity (easy with slime block or plugin). set up wall in end for falling block to hit, see it hits inbetween two blocks (ignoring y)

try vanilla, see it hits only one block (x/z)
try old version of paper, see it hits only one block

### Plugin and Datapack List

none

### Paper version

git-Paper-77 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 44b6869)

### Other

I try ALL the versions until it breaks (it works before). 
it start with build 48 (Git: 385f313a)

Doc94:
This close #8182 
The bug in the issue was introduced by me in upstream... the origin was a [issue](https://hub.spigotmc.org/jira/projects/SPIGOT/issues/SPIGOT-7090) where if use teleport and the location its in another world the coords are change because craftbukkit teleport use `BlockPos` (when change world its related) and this class change the coords from double to int... the fix was use `PositionImpl` and also change the logic when a entity use a portal this cause the issue where decimals was ignored again, this PR change the use of `BlockPos` to `PositionImpl` for this cases and avoid lose this values for plugins.

~~Perdon el mucho texto.. pero mire el error y me dio algo por dentro~~

### DIFF
--- a/patches/server/0929-Fix-entity-exit-position-when-teleport-using-portals.patch
+++ b/patches/server/0929-Fix-entity-exit-position-when-teleport-using-portals.patch
@@ -1,0 +1,57 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 24 Jul 2022 17:20:54 -0400
+Subject: [PATCH] Fix entity exit position when teleport using portals
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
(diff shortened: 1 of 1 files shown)


electronicboy:
if the change made in upstream borked upstream, this should probably be sent there

Doc94:
Close by upstream https://github.com/PaperMC/Paper/commit/345a6a620524ee18db251da2bbb48ff32864be2f

PR #8198, opened 2022-07-28
Initial support for signed messages

### LINKED ISSUE #8201: Component-based Preview Event

There is no preview event that supports components, only bucket's `AsyncPlayerChatPreviewEvent` which uses strings.

## Preferred Solution

If paper's preview event is used (`AsyncChatPreviewEvent`), the component set in that preview event will be made `read-only` in the following `PlayerChatEvent`->`AsyncPlayerChatEvent`->`ChatEvent`->`AsyncChatEvent` chain. That way the preview result will always be the message sent. (Cancellations from the chat events will still be respected).

## Alternative Solution 1

Based on the preferred solution, except its not read-only by default. In one of Paper's chat events (`ChatEvent` or `AsyncChatEvent`) you can tell the event to force use of the preview'd component.

## Alternative Solution 2

Have no "force" system, just add `AsyncChatPreviewEvent` and put the result through the same chat logic, if something else modifies it, so be it, and accept that its a modified message.

I don't like this solution, it means you have to use the same formatting logic in both the preview event, and chat event (this is how bukkit handles things with its events).

(see solutions)

### Other

You can see (and test) the [initial chat signing implementation here](https://github.com/PaperMC/Paper/pull/8198#issuecomment-1197709599).

Machine-Maker:
Initial support for using signed, or signed+modified messages.


Machine-Maker:
This does need some testing with chat plugins in more fleshed out environments. I'd appreciate anyone who wants to give feedback on these changes [to download this jar](https://potrebic.box.com/s/2reta427dieti62b1o938n3fvtj9gzhb), and try it out with your chat setup. Let me know here, or ping me in the Paper discord if you run into anything unexpected.

MiniDigger (dismissed):
looks good as a mvp, I like how relatively painless this is cause we already had most of the ground work done

Owen1212055 (approved):
A lot of testing was done between many other chat events and no issues seem to be present anymore.

Machine-Maker:
Ok, I think this is pretty close to done. I've done some testing with various event interactions, and things work as expected.

kennytv (changes requested):
Breaks display name styling if the message content isn't changed; download EssX and use a nick/have an op prefix and it won't be displayed

kennytv (approved):
Time to break everything

PR #8199, opened 2022-07-28
Auth system and forwarding changes

Xernium:
These changes modify the auth session handler to verify the player key early, if present.    
This potentially saves session service requests on invalid connections.    
The Velocity forwarding mode has also been updated to reflect the V2 key forwarding mode


Xernium:
I just realized I need to patch the signature enforcement field with the value defined in the config for velocity-online-mode, doing that now

electronicboy (approved):

patches/server/0273-Verify-player-key-early.patch
```diff
@@ -0,0 +1,96 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
kennytv, line 1: This patch seems like unnecessarily high-conflict diff for too little benefit in a normal server scenario

patches/server/0273-Verify-player-key-early.patch
```diff
@@ -0,0 +1,96 @@
...
+From: "FivePB (Xer)" <admin@fivepb.me>
+Date: Thu, 28 Jul 2022 11:36:36 +0000
+Subject: [PATCH] Verify player key early
+
+This patch refactors authentication with the player key in a way
+that forces the key to be verified early using the (currently unused)
+player UUID provided in the Server hello packet.
+This helps verify connections early and may reduce requests to the session
+service
```
kennytv, line 10: This patch seems like unnecessarily high-conflict diff for too little benefit in a normal server scenario, since it doesn't help in any other important way with specific targetting etc.

... or just needs fixing the first if on key != null

kennytv (changes requested):

Xernium:
Closed in favor of #8219 and following

PR #8205, opened 2022-07-29
Add extra raytracing api

PrimordialMoros:
New api allows to specify a set of blocks (or rather block positions) that will be ignored when performing a raytrace.

It also adds a builder to easily construct raytraces instead of using the world methods with a myriad args.

Should this be consolidated with previous raytrace patches?

### DIFF
--- a/patches/api/0389-Add-extra-raytracing-api.patch
+++ b/patches/api/0389-Add-extra-raytracing-api.patch
@@ -5,17 +5,13 @@ Subject: [PATCH] Add extra raytracing api
 
 Adds the option to ignore specific blocks (based on their position) when performing a raytrace.
 
-diff --git a/src/main/java/io/papermc/paper/raytrace/RayTraceBuilder.java b/src/main/java/io/papermc/paper/raytrace/RayTraceBuilder.java
+diff --git a/src/main/java/io/papermc/paper/raytrace/RayTraceContext.java b/src/main/java/io/papermc/paper/raytrace/RayTraceContext.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..99a45534f9c4b61fbae1abfb79c810ef419f730c
+index 0000000000000000000000000000000000000000..0b3a490df1e114f81fcedc04343d4b105cb498fb
 --- /dev/null
-+++ b/src/main/java/io/papermc/paper/raytrace/RayTraceBuilder.java
-@@ -0,0 +1,187 @@
++++ b/src/main/java/io/papermc/paper/raytrace/RayTraceContext.java
+@@ -0,0 +1,320 @@
 +package io.papermc.paper.raytrace;
-+
-+import java.util.Collection;
-+import java.util.Set;
-+import java.util.function.Predicate;
 +
 +import com.google.common.base.Preconditions;
 +import org.bukkit.FluidCollisionMode;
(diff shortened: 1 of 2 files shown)


patches/api/0389-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,282 @@
...
++import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Helps prepare a raytrace to be cast.
++ *
++ * Usage of the builder is preferred over the super long {@link World#rayTrace(Location, Vector, double, FluidCollisionMode, boolean, double, Predicate, Collection)} API
++ */
++public class RayTraceBuilder {
```
Machine-Maker: I think I would prefer a name like `RayTraceQuery` and `RayTraceQuery.Builder`. I think this should be a proper builder pattern, instead of what you have here, which kinda combines the too.

Basically make `RayTraceQuery` immutable (maybe with a method to create a builder using its current options), and then have an inner Builder class for actually building the query.

patches/server/0925-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,96 @@
...
+         return CraftRayTraceResult.fromNMS(this, nmsHitResult);
+     }
+ 
+     @Override
+     public RayTraceResult rayTrace(Location start, Vector direction, double maxDistance, FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, double raySize, Predicate<Entity> filter) {
++        return rayTrace(start, direction, maxDistance, fluidCollisionMode, ignorePassableBlocks, raySize, filter, Set.of()); // Paper start - add overload method
++    }
++
++    @Override
++    public RayTraceResult rayTrace(Location start, Vector direction, double maxDistance, FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, double raySize, Predicate<Entity> filter, Collection<Block> ignored) { // Paper end
```
Machine-Maker: I think in this situation, the `// Paper end` command should go on the line below. Sometimes I like doing inline ending comments, but here it would look better if it was on its own line.

patches/server/0925-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,96 @@
...
++          ignoredBlockPositions.add(new BlockPos(block.getX(), block.getY(), block.getZ()));
++        }
++        HitResult nmsHitResult = this.getHandle().clip(new ClipContext(startPos, endPos, ignorePassableBlocks ? ClipContext.Block.COLLIDER : ClipContext.Block.OUTLINE, CraftFluidCollisionMode.toNMS(fluidCollisionMode), null), ignoredBlockPositions);
++        // Paper end
+         return CraftRayTraceResult.fromNMS(this, nmsHitResult);
+     }
+ 
+     @Override
+     public RayTraceResult rayTrace(Location start, Vector direction, double maxDistance, FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, double raySize, Predicate<Entity> filter) {
++        return rayTrace(start, direction, maxDistance, fluidCollisionMode, ignorePassableBlocks, raySize, filter, Set.of()); // Paper start - add overload method
```
Machine-Maker: Paper comment on its own line

patches/server/0925-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,96 @@
...
+         Vec3 startPos = new Vec3(start.getX(), start.getY(), start.getZ());
+         Vec3 endPos = new Vec3(start.getX() + dir.getX(), start.getY() + dir.getY(), start.getZ() + dir.getZ());
+-        HitResult nmsHitResult = this.getHandle().clip(new ClipContext(startPos, endPos, ignorePassableBlocks ? ClipContext.Block.COLLIDER : ClipContext.Block.OUTLINE, CraftFluidCollisionMode.toNMS(fluidCollisionMode), null));
+-
++        Set<BlockPos> ignoredBlockPositions = new HashSet<>();
++        for (Block block : ignored) {
++          ignoredBlockPositions.add(new BlockPos(block.getX(), block.getY(), block.getZ()));
++        }
++        HitResult nmsHitResult = this.getHandle().clip(new ClipContext(startPos, endPos, ignorePassableBlocks ? ClipContext.Block.COLLIDER : ClipContext.Block.OUTLINE, CraftFluidCollisionMode.toNMS(fluidCollisionMode), null), ignoredBlockPositions);
++        // Paper end
```
Machine-Maker: Be more "picky" about starting and ending comments. This should be 2 paper comment blocks instead of 1 big one.

patches/server/0925-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,96 @@
...
+-
++    // Paper start
+     default BlockHitResult clip(ClipContext context) {
++      return clip(context, java.util.Set.of());
++    }
++
++    default BlockHitResult clip(ClipContext context, java.util.Set<BlockPos> ignored) {
+         return (BlockHitResult) BlockGetter.traverseBlocks(context.getFrom(), context.getTo(), context, (raytrace1, blockposition) -> {
+-            return this.clip(raytrace1, blockposition); // CraftBukkit - moved into separate method
++            return this.clip(raytrace1, blockposition, ignored); // CraftBukkit - moved into separate method // Paper end
```
Machine-Maker: This diff can be cleaned up a bit. Make the `// Paper start` comment right above
```java
return clip(context, java.util.Set.of());
```
Then you can put an end comment right below the new default method declaration, and then have 1 more inline Paper comment on the `this.clip(....); // CraftBukkkit -...` line.

patches/server/0925-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,96 @@
...
+         });
+     }
++    // Paper start
++    default BlockHitResult clip(ClipContext raytrace1, BlockPos blockposition) {
++      return clip(raytrace1, blockposition, java.util.Set.of());
++    }
+ 
+     // CraftBukkit start - moved block handling into separate method for use by Block#rayTrace
+-    default BlockHitResult clip(ClipContext raytrace1, BlockPos blockposition) {
++    default BlockHitResult clip(ClipContext raytrace1, BlockPos blockposition, java.util.Set<BlockPos> ignored) { // Paper end
```
Machine-Maker: One comment block for the `clip(ClipContext,BlockPos)` method,
and then 1 inline comment on the method you added a parameter too

patches/api/0389-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,282 @@
...
++   *
++   * @param world the world to cast the raytrace in
++   * @return the result
++   */
++  @Nullable
++  public RayTraceResult entities(@NotNull World world) {
++    Preconditions.checkArgument(world != null, "World cannot be null");
++    Location start = new Location(world, origin.getX(), origin.getY(), origin.getZ());
++    return world.rayTrace(start, direction, range, mode, ignorePassable, raySize, entityPredicate, ignoreBlocks);
++  }
```
Machine-Maker: I think these methods should probably be on World, and just take a `RayTraceQuery` object.

Something to also do in the future (not in this PR) would be to make all the other ray trace methods utilize builders

Machine-Maker (changes requested):
Thanks for you contribution!

I like the idea of using a builder pattern here.

patches/api/0389-Add-extra-raytracing-api.patch
PrimordialMoros: Settled for `RayTraceContext` as a name which is now immutable and has a separate builder.

patches/api/0389-Add-extra-raytracing-api.patch
PrimordialMoros: I think it's still useful to have utility methods to cast the raytrace directly from the builder.
If I add the methods with the `RayTraceContext` arg on World should I replace the new overload methods or simply add them. It would be rather odd to "lock" new api functionality behind the method with the context parameter. But on the other hand could be confusing to have multiple ways (and maybe implementations) to perform the same thing.

Machine-Maker: I generally think that a "Builder" should have one job, to build some other object type. Now the Paper API isn't consistent on that, see the ParticleBuilder


EDIT: Also why wouldn't you just put those methods on the context type? What's the purpose of having them on the Builder specifically instead of the type that's being built.

patches/api/0389-Add-extra-raytracing-api.patch
```diff
@@ -0,0 +1,415 @@
...
++import java.util.Collection;
++import java.util.Set;
++import java.util.function.Predicate;
++
++/**
++ * Represents the immutable context for performing a raytrace.
++ * <p>Usage of the {@link Builder} is preferred over the super long
++ * {@link World#rayTrace(Location, Vector, double, FluidCollisionMode, boolean, double, Predicate, Collection)} API.
++ */
++public class RayTraceContext {
```
Machine-Maker, line 36: Can this be a record?

PrimordialMoros: It could but you'd need to repeat validation for all the arguments and also override some methods due to vectors being mutable. Not sure if it would be worth it.

patches/api/0389-Add-extra-raytracing-api.patch
PrimordialMoros: Likewise, you could argue the context object has one job, storing the raytrace configuration.
I think it's fair to assume that whenever one builds a context they also want to cast a raytrace using it. Those helper methods would simply combine the build and cast in one call.

gepron1x: To be honest, i disagree; simply keeping data is not a responsibility: this turns objects into C structs. I don't see any problems of having a rayTrace method here.

Owen1212055 (commented):
I'm curious on a couple things...

1. I feel the max range is a bit too restrictive here, IMO if someone puts a huge number here too bad! I don't think it's great to limit it like this. Why was this range chosen?
2. Why not use a predicate for ignoreBlocks instead? This way you could simply (in the predicate) define your own logic and possibly extend it in the future. 

But in general, this makes that whole api MUCH nicer to handle. 😅

Owen1212055:
Superceeded by https://github.com/PaperMC/Paper/commit/23860da6c2f306fdc598fb0d09132636c7141dfd .

A builder is something that might be useful, but in general, would probably be better off implemented in a different way (avoid passing World, etc) So if someone wants to pickup on that, PRs welcome. 😄

PR #8206, opened 2022-07-29
Add new override for setChunckForceLoaded

zJanny:
I was coding a plugin and noticed that loadChunk has a override with a chunk parameter, but setChunckForceLoaded does not, so i added it.


electronicboy:
I really do not see the benefits of this, there's already a method on Chunk for setting this

Issue #7006, opened 2021-12-03
[1.18] Teleporting results in exception spam


DarkEyeDragon:
### Stack trace

```
[16:41:25 FATAL]: Error executing task on Chunk source main thread executor for minecraft:overworld
net.minecraft.ReportedException: Exception generating new chunk
        at net.minecraft.server.level.ChunkMap.lambda$scheduleChunkGeneration$34(ChunkMap.java:1260) ~[paper-1.18.jar:git-Paper-25]
        at net.minecraft.util.thread.BlockableEventLoop.doRunTask(BlockableEventLoop.java:149) ~[?:?]
        at net.minecraft.server.level.ServerChunkCache$MainThreadExecutor.doRunTask(ServerChunkCache.java:1231) ~[?:?]
        at net.minecraft.util.thread.BlockableEventLoop.pollTask(BlockableEventLoop.java:122) ~[?:?]
        at net.minecraft.server.level.ServerChunkCache$MainThreadExecutor.pollTask(ServerChunkCache.java:1243) ~[?:?]
        at net.minecraft.server.level.ServerChunkCache.pollTask(ServerChunkCache.java:831) ~[?:?]
        at net.minecraft.server.MinecraftServer.pollTaskInternal(MinecraftServer.java:1401) ~[paper-1.18.jar:git-Paper-25]
        at net.minecraft.server.MinecraftServer.pollTask(MinecraftServer.java:1384) ~[paper-1.18.jar:git-Paper-25]
        at net.minecraft.util.thread.BlockableEventLoop.managedBlock(BlockableEventLoop.java:132) ~[?:?]
        at net.minecraft.server.MinecraftServer.waitUntilNextTick(MinecraftServer.java:1362) ~[paper-1.18.jar:git-Paper-25]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1268) ~[paper-1.18.jar:git-Paper-25]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:317) ~[paper-1.18.jar:git-Paper-25]
        at java.lang.Thread.run(Thread.java:833) ~[?:?]
Caused by: java.lang.ArrayIndexOutOfBoundsException
[16:41:25 INFO]: [STDOUT] [net.minecraft.CrashReport] Negative index in crash report handler (0/12)
```

### Plugin and Datapack List

[16:44:09 INFO]: Plugins (1): RandomTeleport

### Actions to reproduce (if known)

[...]

PR #8208, opened 2022-07-29
Warn on plugins accessing faraway chunks

Owen1212055:
Warns when plugins do silly stuff with chunks.
Warns the cause of issues like: https://github.com/PaperMC/Paper/issues/7006

``[12:56:18 WARN]: [Paper-Test-Plugin] Plugin is getting a faraway chunk at (13225852, 22907744), this might cause issues.``

This adds a util method that can be used to see if code is executed by another plugin by looking for the first class loaded by a plugin in the stacktrace.


electronicboy (approved):

Owen1212055:
Rebased

Malfrador:
Wouldn't this be more useful if it used the worldborder instead of just always 30000000? Vanilla also uses that for most of its within bounds checks for things like chunk ticks, mob spawning, interact packets, structure locations etc. There is also a convenient `isChunkInBounds `method on `WorldBorder` for this exact purpose.

electronicboy:
The concern here is more about stuff crashing the server rather than worrying about stuff peaking outside of the world border, think if stuff like a map rendering plugin started throwing warns because you generated a slither outside of the world border, etc

Owen1212055:
Not to mention this is only meant for coordinates that might cause issues. In this case, chunks outside here seem to cause certain issues with generation.

PR #5691, opened 2021-05-20
Better Stats API

Machine-Maker:
Still needs some javadocs

And if this is good, replacing uses of the bukkit statistic enum.

PR #8210, opened 2022-07-30
Add existsStatistic with Material and EntityType check

petomka:
When I was trying to modify some statistics in batch, I noticed that there wasn't a way to check if a certain statistic and material/entity type combination exists, instead the methods just threw exceptions whenever a certain combination was invalid.

### DIFF
--- a/patches/api/0389-Add-existsStatistic-with-Material-and-EntityType-che.patch
+++ b/patches/api/0389-Add-existsStatistic-with-Material-and-EntityType-che.patch
@@ -1,0 +1,89 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Benedikt Karl <karl.bene.99@googlemail.com>
+Date: Sat, 30 Jul 2022 00:40:50 +0200
+Subject: [PATCH] Add existsStatistic with Material and EntityType check
+
+
+diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
+index 92a1462261029e804da73da2743bbd68e57841e9..fed9b902c05ba392994d24c7942b88acf2e3d072 100644
+--- a/src/main/java/org/bukkit/Bukkit.java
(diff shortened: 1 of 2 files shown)


Machine-Maker (commented):
I don't think this should be merged. I don't want to expand on this API, I want it replaced (See the existing [Better Stats API PR](https://github.com/PaperMC/Paper/pull/5691)).

The current stats API is just bad. its really that simple. No use expanding on a bad system.

SpiritOTHawk (commented):
Rename 0389-Add-existsStatistic-with-Material-and-EntityType-che.patch to 0389-Add-existsStatistic-with-Material-and-EntityType-check.patch :)

MiniDigger:
> Rename 0389-Add-existsStatistic-with-Material-and-EntityType-che.patch to 0389-Add-existsStatistic-with-Material-and-EntityType-check.patch :)

patch file names are automatically generated based off of the commit and have a length limit

qixils:
> I don't think this should be merged. I don't want to expand on this API, I want it replaced (See the existing [Better Stats API PR](https://github.com/PaperMC/Paper/pull/5691)).
> 
> The current stats API is just bad. its really that simple. No use expanding on a bad system.

it's easy to say "we should just use the new API" but considering that new API has been in PR hell for over a year now, i don't really see a good reason why not to merge this 🤷🏻‍♀️ this is a relatively minor and harmless PR; it just needs some cleaning up of the imports and some nullability annotations

electronicboy:
slapped an approval on the other PR, I'd forgotten I'd already given it a look over
as for this one,
1) maintaining extra API which is just deprecated on day 1 is pretty bleh, so not sure about that entirely
2) if this was to be merged, that API method needs to be renamed
3) needs cleanup

Owen1212055:
Thank you for your contribution, however, as stated by Cat and Machine it's probably better to freeze any additions to this API due to its planned replacement.

I also agree with this, because this is just annoying maintenance for us on something we want to be replaced anyways.

PR #8211, opened 2022-07-30
Allow disabling chat message signatures

Linux4:
This patch allows to disable the popup when joining a server without "secure chat", e.g offline-mode servers (also servers behind bungeecord/waterfall) and additionally converts all chat messages to system messages so they can't be reported nor will be flagged as unverified.


electronicboy:
No, we have no interest in accommodating this, use a plugin

PR #8212, opened 2022-07-30
Custom Chat Completions API

Owen1212055:
Allows you to add custom suggestions in chat.
Better names are appreciated, but it lets you do stuff like this.

image

I decided not to expose SET chat completions because plugins might override each other, I'm not sure... it might be nice to have as then you have to track which suggestions you have sent in order to reset completions. Feedback? 

(I do not endorse using legacy chars)

### DIFF
--- a/patches/api/0389-Custom-Chat-Completion-Suggestions-API.patch
+++ b/patches/api/0389-Custom-Chat-Completion-Suggestions-API.patch
@@ -1,0 +1,36 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sat, 30 Jul 2022 11:23:11 -0400
+Subject: [PATCH] Custom Chat Completion Suggestions API
+
+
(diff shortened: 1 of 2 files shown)


patches/api/0389-Custom-Chat-Completion-Suggestions-API.patch
```diff
@@ -0,0 +1,36 @@
...
+      */
+     void sendOpLevel(byte level);
++
++    /**
++     * Adds custom chat completion suggestions that the client will
++     * suggest when typing in chat.
++     *
++     * @param completions custom completions
++     */
++    void addCustomCompletionSuggestions(@NotNull java.util.List<String> completions);
```
MiniDigger: addAdditionalChatCompletions maybe?

Owen1212055: Agreed

kennytv (changes requested):
Should accept a collection and wrap in a new list (so that people modifying packet fields don't modify the provided collection)

kennytv (approved):

PR #8213, opened 2022-07-30
Add Entity#hideEntityByDefault

### LINKED ISSUE #7395: Create Entity#hideEntityByDefault

zJanny:
Closes #7395


aerulion:
I don't quite think your current implementation does what was requested in the initial feature request. If I understood the request correctly, one should be able to hide the entity be default from every player joining / on the server, _whilst still being able to show the entity to specific players manually_. And that simply won't work here if you are just setting the entity silent and more or less always returning "false" for the "canSee()" method, once the hideEntityByDefault boolean is set to "true". Also setting the entity silent would have unwanted side-effects, since that would be persisted, as opposed to the hideByDefault value.
(Haven't tested whether this works at all)

PR #8217, opened 2022-08-01
Undeprecate and fix Boat#getBoatMaterial

### LINKED ISSUE #8216: `Boat#getBoatMaterial()` is erroneously deprecated and has changed behaviour
### Expected behavior

This method previously, and should, return the Material corresponding to the held variant of that boat. Eg, `Material.OAK_BOAT`.

### Observed/Actual behavior

The method now delegates to `getBoatType().getMaterial()` from Spigot, which returns the _plank_ type that the boat is made of.

### Steps/models to reproduce

* https://github.com/PaperMC/Paper/blob/cd215af6eb784724868ea0d276e6cc0f06fc2624/patches/server/0547-Implement-API-to-get-Material-from-Boats-and-Minecar.patch#L18
* https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/entity/Boat.java#139

### Plugin and Datapack List

n/a - API bug

### Paper version

fd557b7d51bba4d5e830d6047005f7e4335f5ada

### Other

Given this should not just mirror the Bukkit method, this should not be marked as deprecated either.

me4502:
Fixes https://github.com/PaperMC/Paper/issues/8216

As explained in attached issue, this method has had a behaviour change at the time it was deprecated. The deprecation was added because the behaviour was the same as what it was changed to. Given that behaviour was incorrect, I've reverted to the correct and original behaviour, and removed the corresponding deprecation.

Tbh moving it into that enum might make sense at some point anyway - but for now this resolves any plugins that relied on getting a boat item back from this function.

### DIFF
--- a/patches/server/0547-Implement-API-to-get-Material-from-Boats-and-Minecar.patch
+++ b/patches/server/0547-Implement-API-to-get-Material-from-Boats-and-Minecar.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Implement API to get Material from Boats and Minecarts
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java
-index 5871cd149fe07e97c5d68ffd83dae4d3fc6bcf03..40debf1130a8e6cc9510061976a01050cd8ddc05 100644
+index 5871cd149fe07e97c5d68ffd83dae4d3fc6bcf03..f2896aa6fa5a5282b4be106320c0dad9dd6036c5 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java
 @@ -80,6 +80,13 @@ public class CraftBoat extends CraftVehicle implements Boat {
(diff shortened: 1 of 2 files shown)


patches/server/0547-Implement-API-to-get-Material-from-Boats-and-Minecar.patch
```diff
@@ -5,17 +5,25 @@ Subject: [PATCH] Implement API to get Material from Boats and Minecarts
...
-index 5871cd149fe07e97c5d68ffd83dae4d3fc6bcf03..40debf1130a8e6cc9510061976a01050cd8ddc05 100644
+index 5871cd149fe07e97c5d68ffd83dae4d3fc6bcf03..44ea7d4afa3f633e0b064247b51e963620eb9642 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftBoat.java
-@@ -80,6 +80,13 @@ public class CraftBoat extends CraftVehicle implements Boat {
+@@ -4,6 +4,7 @@ import com.google.common.base.Preconditions;
+ import java.util.stream.Collectors;
+ import org.bukkit.TreeSpecies;
+ import org.bukkit.craftbukkit.CraftServer;
++import org.bukkit.craftbukkit.util.CraftMagicNumbers;
```
kennytv: Should inline the import

kennytv (approved):
Else good

PR #8218, opened 2022-08-01
Don't call AsyncTabCompleteEvent from netty IO thread

kennytv:
Since forever, the AsyncTabCompleteEvent has been called from the netty IO thread, which is *very* bad, especially considering the patch's descriptions says
> Especially useful if you need to query a database in order to obtain the results for tab
completion, such as offline players.

The first and third commit are purely cosmetic to merge two patches that should've always just been one (aside from a comment and variable name change), the actual diff is in [the second commit](https://github.com/PaperMC/Paper/commit/f47a42a190a9b83987d952aca028ab1d2f099261), simply wrapping everything in an executor.


patches/server/0166-AsyncTabCompleteEvent.patch
```diff
@@ -13,12 +13,20 @@ completion, such as offline players.
...
+index f558d0f8f7b78132510f8c5e61701eccf1f8bcfa..6de0b057d61dbe5939f0c04d8a5d1469837ed305 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-@@ -783,10 +783,10 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+@@ -781,27 +781,58 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+ 
+     }
  
++    // Paper start
++    private static final java.util.concurrent.ExecutorService TAB_COMPLETE_EXECUTOR = java.util.concurrent.Executors.newCachedThreadPool(
```
electronicboy: I do wonder if this one should be bounded, given it's placement

kennytv: der you go, 4 is all I've got

electronicboy (approved):

PR #8219, opened 2022-08-01
Update modern forwarding to v3

Xernium:
- Update modern forwarding to v3
- Fix the inconsistent state of ServerData when on proxied networks
- Move forwarding patch to the end for future invasive patch (originally part of #8199)


kennytv (dismissed):

electronicboy (approved):

Aurelien30000:
I tried this pull request and I noticed that something may be wrong. _Hopefully I did not mess up with my own stuff, let me explain:_

Take a look here, ``profilepublickey`` is built (and only built here) with the player in the following method ``ServerLoginPacketListenerImpl#handleAcceptedLogin``.

```java
public void handleAcceptedLogin() {
    ProfilePublicKey profilepublickey = null;

    if (!this.server.usesAuthentication()) {
        // this.gameProfile = this.createFakeProfile(this.gameProfile); // Spigot - Moved to initUUID
        // Spigot end
    } else {
        try {
            SignatureValidator signaturevalidator = this.server.getServiceSignatureValidator();

            profilepublickey = ServerLoginPacketListenerImpl.validatePublicKey(this.profilePublicKeyData, this.gameProfile.getId(), signaturevalidator, this.server.enforceSecureProfile());
        } catch (ServerLoginPacketListenerImpl.PublicKeyValidationException loginlistener_a) {
            ServerLoginPacketListenerImpl.LOGGER.error(loginlistener_a.getMessage(), loginlistener_a.getCause());
            if (!this.connection.isMemoryConnection()) {
                this.disconnect(loginlistener_a.getComponent());
                return;
            }
        }
    }

    // CraftBukkit start - fire PlayerLoginEvent
    ServerPlayer s = this.server.getPlayerList().canPlayerLogin(this, this.gameProfile, profilepublickey, hostname);
    ...
```

Behind a Velocity proxy, using **your** branch, players still get the following pop-up warning and messages appear as non-signed _(red icon)_.
image

This is caused by a null ``profilepublickey``, only generated when the server **does** use authentication (server.properties ``online-mode``) as the code states itself. This may be wrong since a proxy does not require this setting, and breaks whole chat signing, the check should be adapted then. To let you known, I wrapped out the if condition, and it fixed the message sign status issue.

Thanks!

Xernium:
I’ll have to look at this again. I’m a tiny bit surprised because if that’s true then keys don’t work with bungeecord at all either

Xernium:
Although 
https://github.com/PaperMC/Paper/pull/8219/files#diff-9d8186f3d1fdf665b520d062d18b4ffccf16440059c4a8d58c2349d524b45a1bR105
Is already also a fix required by bungeecord too so it wouldn’t be completely unexpected.

Aurelien30000:
> I’ll have to look at this again. I’m a tiny bit surprised because if that’s true then keys don’t work with bungeecord at all either

I’m not surprised bth because bungeecord does not support chat signing and report yet: https://www.spigotmc.org/threads/spigot-bungeecord-1-19-1.566548/ (third point)

Xernium:
@Aurelien30000 Thanks, good catch. Fixed with https://github.com/PaperMC/Paper/pull/8219/commits/fbcef06ea213eb521ecf9048b710cf69fc7f637a
I also force pushed the repo

Aurelien30000:
Gladly! I went further in my testing, _it's a bit out of this PR_, but here's what I noticed for future imprs/fixes:

- This problem does exist and the fix works as expected: https://github.com/PaperMC/Velocity/pull/810
- When a player switches of server _(from A to B)_ ``LastSeenMessages`` _(chat packet)_ sent by the client on server B is not empty, the server B obviously does not recognize the messages sent before on the server A. The first message fails and the client is kicked by the server for ``Chat message validation failure``.

Xernium:
> Gladly! I went further in my testing, _it's a bit out of this PR_, but here's what I noticed for future imprs/fixes:
> 
>     * This problem does exist and the fix works as expected: [Fix command packet wrong mark as unsigned Velocity#810](https://github.com/PaperMC/Velocity/pull/810)
> 
>     * When a player switches of server _(from A to B)_ `LastSeenMessages` _(chat packet)_ sent by the client on server B is not empty, the server B obviously does not recognize the messages sent before on the server A. The first message fails and the client is kicked by the server for `Chat message validation failure`.

I will fix that, give me a second;     

Heres a deeper problem, I need the following to be tested:
- Two paper servers 
- One velocity proxy
- Three clients (A, B, C)
       
I need to know what happens when
- Client A stays on server 1
- Client B stays on server 2
- Client C logs into server 1, chats with A
- Client C switches to server 2 and chats with B
- Client C switches back to server 1 and chats with A
     
Depending on the outcome we may have to forcefully manage chat on the proxy layer.

Xernium:
The second thing should be fixed with https://github.com/PaperMC/Paper/pull/8219/commits/d2ee57cba35fb38d0048927f99d0bb203f8ac761

Xernium:
Back to the drawing board

SpiritOTHawk (approved):

electronicboy (changes requested):
This is probably getting a bit wider than what it was originally intending to cover, the v3 stuff looks fine, but the other proto stuff should probably be moved into a separate PR, that way it can be reviewed alongside the relevant velocity changes

Aurelien30000:
Yes and no, because the v3 PR itself, **with the key correction**, breaks the chat behind a proxy, when switching server.

Xernium:
To update everyone here:
The chat sync protocol will be part of another project addition. That also entails the velocity forwarding update being partially broken. This is intentional.
The fixes to make that work are to be part of the follow-up.

kennytv (approved):

electronicboy (approved):

PR #8221, opened 2022-08-02
Fix suggest command message for brigadier syntax exceptions

chickeneer:
This is a bug accidentally introduced in upstream CB.

Introduced in this commit. https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/64c15270e76475e68b2167d4bfba162a4a827fe0#nms-patches/net/minecraft/commands/CommandDispatcher.patch


Machine-Maker (approved):
Some of this craftbukkit diff is really overblown. Like I feel they make so many changes that are no longer needed. Can't want to rewrite these patches to much more closely match vanilla code.

kennytv (approved):

PR #8232, opened 2022-08-03
Fix calling UnknownCommandEvent

### LINKED ISSUE #8225: UnknownCommandEvent only gets fired when the console uses an unknown command
### Expected behavior

For the event to be fired not only when the console uses an unknown command

### Observed/Actual behavior

The event is only fired when the console uses an unknown command

### Console
(ignore the debug print to stdout)
Console
### Player
Player

### Plugin and Datapack List

Only one plugin which i made
The repo: https://github.com/Skulk-Network/Plugin
(related: https://github.com/Skulk-Network/Plugin/blob/master/src/main/java/network/skulk/plugin/extensions/message_override/listeners/UnknownCommandListener.java)

### Paper version

```
This server is running Paper version git-Paper-102 (MC: 1.19.1) (Implementing API version 1.19.1-R0.1-SNAPSHOT) (Git: e0486ae)
Previous version: git-Paper-99 (MC: 1.19.1)
```

### Other

_No response_

### LINKED ISSUE #8329: papers config no-permission message comes back a different message than the one you set
### Expected behavior

doing a command you dont have permission for  should be the message you set in papers config

also running an unknown command should be what is set in spigots config

### Observed/Actual behavior

the message sent is "Uknown or incomplete command, see below for error"

the message sent for an unknown command is also "Uknown or incomplete command, see below for error"

### Steps/models to reproduce

run an unknown command or a command you dont have permission for as a player in a 1.19.2 paper server. 

it sends the configured message when you run an unknown command in the console.

### Plugin and Datapack List

none

### Paper version

paper-1.19.2-132

### Other

this problem only happens on a paper 1.19.2 server, 1.19 does not do this

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/8225
Fixes https://github.com/PaperMC/Paper/issues/8329

### DIFF
--- a/patches/api/0058-Add-UnknownCommandEvent.patch
+++ b/patches/api/0058-Add-UnknownCommandEvent.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add UnknownCommandEvent
 
 diff --git a/src/main/java/org/bukkit/event/command/UnknownCommandEvent.java b/src/main/java/org/bukkit/event/command/UnknownCommandEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..57128c780bc73e39d2733cceeb21dc06c2422c3c
+index 0000000000000000000000000000000000000000..e4f8a0d14ae66468bbb0a5d6bf9c1ee417dc866c
 --- /dev/null
 +++ b/src/main/java/org/bukkit/event/command/UnknownCommandEvent.java
-@@ -0,0 +1,113 @@
+@@ -0,0 +1,110 @@
 +package org.bukkit.event.command;
 +
 +import net.kyori.adventure.text.Component;
@@ -17,6 +17,7 @@ index 0000000000000000000000000000000000000000..57128c780bc73e39d2733cceeb21dc06
 +import org.bukkit.command.CommandSender;
 +import org.bukkit.event.Event;
 +import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.Contract;
 +import org.jetbrains.annotations.NotNull;
 +import org.jetbrains.annotations.Nullable;
 +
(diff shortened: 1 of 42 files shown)


kennytv (dismissed):
`/tp` needs to have the brig message passed into the event, should also add a small note that this now includes invalid command syntax as unknown commands for called events

patches/server/0139-Add-UnknownCommandEvent.patch
```diff
@@ -3,9 +3,108 @@ From: Sweepyoface <github@sweepy.pw>
...
+                 }
++                // Paper end
+ 
+                 b1 = 0;
+                 return b1;
+@@ -350,6 +378,7 @@ public class Commands {
+                 b0 = 0;
+             }
+         } finally {
++            this.dispatcher.setConsumer(existing); // Paper
```
electronicboy: I would probably validate that the consumer hasn't changed in the meantime

electronicboy (approved):

patches/server/0139-Add-UnknownCommandEvent.patch
```diff
@@ -3,9 +3,108 @@ From: Sweepyoface <github@sweepy.pw>
...
++++ b/src/main/java/net/minecraft/commands/CommandSourceStack.java
+@@ -329,8 +329,13 @@ public class CommandSourceStack implements SharedSuggestionProvider {
+     }
+ 
+     public void sendFailure(Component message) {
++        // Paper start
++        this.sendFailure(message, true);
++    }
++    public void sendFailure(Component message, boolean withStyle) {
++        // Pape rend
```
Lulu13022002: Small typo

patches/server/0139-Add-UnknownCommandEvent.patch
```diff
@@ -3,9 +3,108 @@ From: Sweepyoface <github@sweepy.pw>
...
++                        .append(net.kyori.adventure.text.Component.newline())
++                        .append(io.papermc.paper.adventure.PaperAdventure.asAdventure(ichatmutablecomponent));
++                }
++                }
++                org.bukkit.event.command.UnknownCommandEvent event = new org.bukkit.event.command.UnknownCommandEvent(commandlistenerwrapper.getBukkitSender(), s, builder.build());
++                org.bukkit.Bukkit.getServer().getPluginManager().callEvent(event);
++                if (event.message() != null) {
++                    commandlistenerwrapper.sendFailure(io.papermc.paper.adventure.PaperAdventure.asVanilla(event.message()), false);
+                 }
++                // Paper end
```
Lulu13022002: This comment should probably go one line above

patches/server/0139-Add-UnknownCommandEvent.patch
Machine-Maker: I just removed the whole consumer thing. Found a better way to know if a vanilla command was run

pauhull:
Is this going to be fixed?

electronicboy (approved):

Issue #4361, opened 2020-09-24
Wrong EntitySpawnReasons when Bees come out of a beehive


aurorasmiles:
### What behaviour is expected:
When a bee comes out of a beehive / nest, the EntitySpawnReason should be BEEHIVE
### What behaviour is observed:
The spawn reason is SPAWNER_EGG
### Steps/models to reproduce:

1. Make a plugin that prints out the EntitySpawnReason for bees (https://waste.cutiepie.at/lluwenvjm414ejhe)
2. Set up a few beehives with bees and wait till they come out
### Plugin list:
only the test plugin
### Paper version:
1.16.3 204

aurorasmiles:
It seems like the bee keeps the spawn reason they originally spawned with. When I spawn the Bee mit a spawn egg, the EntitySpawnReason is Spawner_egg when they come out of the nest.
When I spawned them via breeding, the EntitySpawnReason also is breeding when it comes out of the nest

Machine-Maker:
I feel like this is intended behavior? The spawn reason is supposed to be the same through out the entities "life". Technically, it is the same entity that comes out of the beehive after you spawned in that bee with a spawn egg. 

I could be getting it backwards tho. And it's supposed to be updated every time the entity appears in the world. If thats the case, then any entity that goes through a portal should have portal be its spawn reason.

aurorasmiles:
I'm not sure. In that case, there should maybe be a seperate "BeeExitHive" event or something like that :D

Machine-Maker:
But that isn't the reason they exist. You spawned them in with an egg. That's their spawn reason. Just cause they happen to go into a box doesn't change that reason.

Proximyst:
I'd like input from @PaperMC/core-team on what the intended spawn reason here, as both reasons can be argued for.

HexedHero:
I feel like this is intended and normal behaviour like Machine since its not the reason why they were spawned into the world, just exited a block but I 100% agree we need an EntityLeaveBlockEvent of some sort, you could prob add metadata to the bee on the EntityEnterBlockEvent and check for it on Spawn event and remove it.

Chew:
I guess we're going philosophical here, what is considered "spawning"? Entity isn't loaded/rendered in the bee nest/beehive, it's stored, but it's not an actual, living entity. Does it "spawn" out of the bee hive/nest? If not, what would you call this?

HexedHero:
Spawning = Added into the game for the first time.
[...]

PR #8236, opened 2022-08-04
Implement CreatureSpawnEvent.SpawnReason.SPAWNER_BUCKET

### LINKED ISSUE #8234: Missing CreatureSpawnEvent.SpawnReason when using a bucket with tropical fish or axolotl

When a player uses a bucket with a fish or axolotl, the event `CreatureSpawnEvent` is fired with the reason `SpawnReason.SPAWNER_EGG`, that is not the real reason.
When we use a plugin that logs specific entity spawns (SPAWNER_EGG for instance), we thought that a player where using spawner-eggs (that is not obtainable in survival) but we figured out that the player was only using buckets in water.

Add a new entry in the `CreatureSpawnEvent.SpawnReason` enum, maybe like `EMPTY_BUCKET`, `BUCKET` or something named more appropriately.
And of course, calling the event with the appropriate new `SpawnReason` when a player empties a fish/axolotl bucket.

Since the player spawning the entity is not referenced by the `CreatureSpawnEvent`, it is not possible to get what’s in the hand of the player, thus it is not possible to deduce if it was a spawner-egg or a bucket. 

### Other

_No response_

Doc94:
This close #8234 adding a new SpawnReason for entity spawned using a bucket.

The name of the reason is based in another things with "spawn" logic like SPAWNER_EGG

### DIFF
--- a/patches/api/0391-Add-CreatureSpawnEvent.SpawnReason.SPAWNER_BUCKET.patch
+++ b/patches/api/0391-Add-CreatureSpawnEvent.SpawnReason.SPAWNER_BUCKET.patch
@@ -1,0 +1,23 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Wed, 3 Aug 2022 22:15:16 -0400
+Subject: [PATCH] Add CreatureSpawnEvent.SpawnReason.SPAWNER_BUCKET
+
+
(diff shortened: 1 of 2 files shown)


davidmayr:
I think just BUCKET or ENTITY_BUCKET would be better than SPAWNER_BUCKET.

The only other two entries with "spawn" in their name are actually named like that in game. E.g. the Spawner Block or for example "Blaze Spawn Egg"

Machine-Maker:
See https://github.com/PaperMC/Paper/issues/4361#issuecomment-701640428. In a similar situations, we don't change a Bee's SpawnReason when they leave a beehive, it keeps the previous one. A similar thing should be applied here. It can fallback to some other reason if it doesn't exist. (I think spawn egg is still fine for that)

PR #8242, opened 2022-08-05
Add configuration option for MiniMessage-formatted MOTD

### LINKED ISSUE #8230: MiniMessage support for MOTD

Currently, the only way to set the MOTD of a server without plugins is using the `motd` field in `server.properties` which uses legacy section formatting instead of MiniMessage.

Add a config option to paper-global.yml's `messages` section which allows specifying an MOTD using MiniMessage which overrides the one set in `server.properties`

Spending like 5 minutes to create a new Paper project in IntelliJ and setting the MOTD myself with `Component`s

### Other

_No response_

wordandahalf:
This closes #8230 by adding a new configuration option at `messages.motd` in `paper-global.yml` that overrides the value of `motd` in `server.properties` if it exists.

This is my first contribution, so please let me know if there is anything that I can change; I feel like there may have been some curfuffling in the process of rebuilding of the patches, but it functioned as expected when I tested.

Thanks!

### DIFF
--- a/patches/server/0096-Paper-config-files.patch
+++ b/patches/server/0096-Paper-config-files.patch
@@ -416,10 +416,10 @@ index 0000000000000000000000000000000000000000..31325994ab441c59a4c0bd9f3f9db3d9
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..456595e4b7e0c7f50617aa2694b0d2dfc368ab81
+index 0000000000000000000000000000000000000000..f5100edd1bd664326d7b47b50d3a9f687ecb03d2
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,265 @@
+@@ -0,0 +1,269 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
@@ -467,6 +467,10 @@ index 0000000000000000000000000000000000000000..456595e4b7e0c7f50617aa2694b0d2df
 +        }
 +
 +        public Component noPermission = Component.text("I'm sorry, but you do not have permission to perform this command. Please contact the server administrators if you believe that this is in error.", NamedTextColor.RED);
++
++        @Nullable
++        public Component motd = null;
++
 +        public boolean useDisplayNameInQuitMessage = false;
 +    }
 +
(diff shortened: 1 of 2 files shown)


Doc94:
Welcome to the world of PR :3
For config things you need add to the patch related to config like mention the contribution files https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#committing-changes and then make another patch for your changes (if im not wrong...)

patches/server/1025-Add-configuration-option-for-MiniMessage-formatted-M.patch
```diff
@@ -0,0 +1,38 @@
...
+diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+index 456595e4b7e0c7f50617aa2694b0d2dfc368ab81..446ed729a74ba08d882059980652b29cf3a872cd 100644
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -45,6 +45,8 @@ public class GlobalConfiguration extends ConfigurationPart {
+         }
+ 
+         public Component noPermission = Component.text("I'm sorry, but you do not have permission to perform this command. Please contact the server administrators if you believe that this is in error.", NamedTextColor.RED);
++        @Nullable
++        public Component motd = null;
```
Doc94: im not sure how is handled the null in the file... and what can happen if you use a empty string in the config..

wordandahalf: From my testing, the field is omitted by default in the file. If added, its value overrides that of `motd` in `server.properties` so long as it is not `null`/empty:
```yaml
# Is removed, considered null/empty
motd:
# Same story
motd: null
# Considered a value, yields an empty MOTD
motd: ''
```
Would it be preferred to make the field be an empty string by default, and have it override only when it is nonempty?

wordandahalf:
> For config things you need add to the patch related to config like mention the contribution files https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#committing-changes and then make another patch for your changes (if im not wrong...)

Thanks for that, I managed to miss that tidbit.

patches/server/1025-Add-configuration-option-for-MiniMessage-formatted-M.patch
electronicboy: I generally eer towards using a string like "default", as then we can just document the behavior of what's going on, vs empty string which, what if you do want an empty string? In the future there is an argument to move all of the config strings over into paper.yml, and I'm not 100% sure that in the long term we'll want to keep those options existing in their standard config files (outside of reading them if they exist so that we can convert them). "default" just offers a sane option which allows us to determine and change what the default behavior is in the future without worrying as much about the side-effects of it (default right now means grab from server.properties, in the future, it could be, use the default hard-coded string)

wordandahalf: I see your perspective, but making it nullable seems to be the simplest and most straightfoward approach. It's completely opt-in, otherwise it functions exactly as it did before. And it's pretty much self-documenting. You want a MiniMessage-formatted MOTD? Add the field. You want to use the value in `server.propertes`? Remove the field.

Of course, I'd be more than happy to reimplement it in the way you describe if that's what's decided is the best course of action.

wordandahalf:
I've (hopefully) fixed up the configuration commit. I've noticed that my patch numbers do not correspond with the existing numbering. Is that an issue?

Owen1212055:
This can be fixed by applying patches and rebuilding.

In general, I see why the whole "null" is a little bit iffy. What if we had a new type that represented an "override"? Basically,
if motd was equal to ``unset`` this would cause it to use the default vanilla server properties. 

This might be a useful type in general if we want to move more things to use adventure like this.

PR #8243, opened 2022-08-05
Fix inconsistencies between offline and online spawn position getter

### LINKED ISSUE #7860: Player vs OfflinePlayer getBedSpawnLocation
### Expected behavior

When calling `getBedSpawnLocation` on an instance of CraftPlayer or CraftOfflinePlayer the result should be the same (or the discrepancy should be documented and predictable)

### Observed/Actual behavior

They are different as shown here
```
test
[16:05:20 INFO]: [Paper-Test-Plugin] [STDOUT] online: true
[16:05:20 INFO]: [Paper-Test-Plugin] [STDOUT] spawn: Location{world=CraftWorld{name=world_nether},x=-7.5,y=89.0,z=-18.5,pitch=0.0,yaw=0.0}
[16:05:22 INFO]: Machine_Maker lost connection: Disconnected
[16:05:22 INFO]: Machine_Maker left the game
test
[16:05:27 INFO]: [Paper-Test-Plugin] [STDOUT] online: false
[16:05:27 INFO]: [Paper-Test-Plugin] [STDOUT] spawn: Location{world=CraftWorld{name=world_nether},x=-8.0,y=89.0,z=-18.0,pitch=0.0,yaw=0.0}
```

### Steps/models to reproduce

1. Set your spawn in the world
2. Log the player's bed spawn loc
3. log out
4. get the instance of offlineplayer
5. log the bed spawn loc

### Plugin and Datapack List

none

### Paper version

af3b3778f6de06af48de08d41a43565d3ff19976

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7860

Currently i have just added a boolean (and keep the old behavior by default)
which will load the location of the block and check if the spawn block is valid

The default behavior is the following
offline: load = false
online: load = true

I'm not sure if it's better to do that for minimal diff or to completely rewrite two new methods (with a better name cause bedSpawnLoc name isn't accurate anymore with the respawn anchor) one to only get the stored position
and another to check that position (and maybe deprecate the old method), so feedback are welcome


patches/api/0391-Fix-wrong-position-for-bed-spawn-loc.patch
```diff
@@ -0,0 +1,92 @@
...
++++ b/src/main/java/org/bukkit/OfflinePlayer.java
+@@ -172,15 +172,29 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
+      */
+     public boolean hasPlayedBefore();
+ 
++    // Paper start
+     /**
+-     * Gets the Location where the player will spawn at their bed, null if
+-     * they have not slept in one or their current bed spawn is invalid.
++     * Gets the Location where the player will spawn at their spawn block (bed, respawn anchor...), null if
```
lynxplay: If we are editing javadocs anyway, this should mention that the method will not do a proper lookup by default.

patches/api/0391-Fix-wrong-position-for-bed-spawn-loc.patch
```diff
@@ -0,0 +1,92 @@
...
++     * Sets the Location where the player will spawn at their spawn block (bed, respawn anchor...).
+      *
+      * @param location where to set the respawn location
+      * @param force whether to forcefully set the respawn location even if a
+-     *     valid bed is not present
++     *     valid spawn block is not present
+      */
+     public void setBedSpawnLocation(@Nullable Location location, boolean force);
++    // Paper end
+
```
lynxplay: This is a huge paper block for single line changes in javadocs.
Probably best to add in line comments. Html comments work fine in javadocs for that matter. (See https://github.com/PaperMC/Paper/blob/master/patches/api/0226-Clarify-the-Javadocs-for-Entity.getEntitySpawnReason.patch#L16)

patches/server/0933-Fix-wrong-position-for-bed-spawn-loc.patch
```diff
@@ -0,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
+Date: Wed, 20 Jul 2022 14:16:25 +0200
+Subject: [PATCH] Fix wrong position for bed spawn loc
```
lynxplay: There is already https://github.com/PaperMC/Paper/blob/master/patches/server/0896-Fix-OfflinePlayer-getBedSpawnLocation.patch.

Concerning this is now adding new API, I guess the existing patch should just be merged into this one or the other way around, crediting either you or MM as a co-author.

patches/server/0933-Fix-wrong-position-for-bed-spawn-loc.patch
```diff
@@ -0,0 +1,65 @@
...
+     }
+ 
+     @Override
+-    public Location getBedSpawnLocation() {
++    public Location getBedSpawnLocation(boolean load) { // Paper
+         ServerLevel world = this.getHandle().server.getLevel(this.getHandle().getRespawnDimension());
+         BlockPos bed = this.getHandle().getRespawnPosition();
+ 
+-        if (world != null && bed != null) {
++        if (load && world != null && bed != null) { // Paper
```
lynxplay: Wouldn't 
```java
@Override
public Location getBedSpawnLocation(boolean load) { // Paper
    ServerLevel world = this.getHandle().server.getLevel(this.getHandle().getRespawnDimension());
    BlockPos bed = this.getHandle().getRespawnPosition();

    if (world != null && bed != null) {
        // Paper start
        if(!load) {
            return new Location(world.getWorld(), bed.getX(), bed.getY(), bed.getZ(), this.getHandle().getRespawnAngle(), 0);
        }
        // Paper end
        Optional<Vec3> spawnLoc = net.minecraft.world.entity.player.Player.findRespawnPositionAndUseSpawnBlock(world, bed, this.getHandle().getRespawnAngle(), this.getHandle().isRespawnForced(), true);
        if (spawnLoc.isPresent()) {
            Vec3 vec = spawnLoc.get();
            return new Location(world.getWorld(), vec.x, vec.y, vec.z, this.getHandle().getRespawnAngle(), 0);
        }
    }
    return null;
}
```

just be less diff and also nicer logic wise.

lynxplay (changes requested):

patches/api/0391-Fix-wrong-position-for-bed-spawn-loc.patch
Lulu13022002: Ooh, i didn't know that was allowed

Lulu13022002:
I have discover that a method getPotentialBedLocation exists in HumanEntity which doesn't load and check the block so maybe it should be deprecated after this.

Owen1212055 (approved):
```
ONLINE: Location{world=CraftWorld{name=world},x=92.5,y=90.0,z=44.5,pitch=0.0,yaw=-25.730234}

OFFLINE:
OLD LOGIC: Location{world=CraftWorld{name=world},x=93.0,y=90.0,z=44.0,pitch=0.0,yaw=-25.730234}
NEW LOGIC: Location{world=CraftWorld{name=world},x=92.5,y=90.0,z=44.5,pitch=0.0,yaw=-25.730234}
```

patches/server/0863-Fix-OfflinePlayer-getBedSpawnLocation.patch
```diff
@@ -39,8 +49,38 @@ index d7823d7dc88cfba6f6ac9dae220e03dea4a0bcdd..6d2ba650f53de8a460857f1846401a20
 +            if (spawnWorld == null) {
 +                return null;
 +            }
-+            return new Location(spawnWorld, data.getInt("SpawnX"), data.getInt("SpawnY"), data.getInt("SpawnZ"), respawnAngle, 0);
++            if (!load) {
```
Machine-Maker: instead of exiting early, it might be possible to adapt the findRespawnPositionAndUseSpawnBlock method to not load anything not already loaded. Then if the area being searched is already loaded, the full logic can run even if the load param is false

Lulu13022002: That's so true, thanks

Machine-Maker (approved):

PR #8244, opened 2022-08-05
Missing eating regain reason

Lulu13022002:
Correct the javadoc of the eating regain reason because currently only animal trigger this and not the player
Also add the missing reason for the llama and the cat
I'm not sure if we can consider also the iron golem with the iron ingots ?


patches/api/0391-Missing-eating-regain-reason.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/org/bukkit/event/entity/EntityRegainHealthEvent.java b/src/main/java/org/bukkit/event/entity/EntityRegainHealthEvent.java
+index d51d2ec1d04d9ea8a25a70d0d856f2355ebfcb4a..7ecff9fcee19fc94be784474fea620e5dd434731 100644
+--- a/src/main/java/org/bukkit/event/entity/EntityRegainHealthEvent.java
++++ b/src/main/java/org/bukkit/event/entity/EntityRegainHealthEvent.java
+@@ -105,7 +105,7 @@ public class EntityRegainHealthEvent extends EntityEvent implements Cancellable
+          */
+         SATIATED,
+         /**
+-         * When a player regains health from eating consumables
++         * When an animal regains health from eating consumables
```
kennytv: Should go into the javadox fixup patch

kennytv (changes requested):

Machine-Maker (commented):
Should the `EATING` reason be used for Iron Ingots being given to Iron Golems? It's using the same logic, consume a resource to heal the entity, but maybe the word "eating" shouldn't describe that for iron golems.

Lulu13022002:
It depends if we consider the golem as a robot or not, If so the word REPAIR can be better

kennytv (approved):

PR #8245, opened 2022-08-05
Add BANNERS Material Tag

Insprill:


lynxplay:
https://jd.papermc.io/paper/1.19/org/bukkit/Tag.html#BANNERS and https://jd.papermc.io/paper/1.19/org/bukkit/Tag.html#ITEMS_BANNERS already exist.

I don't see a point in this PR.

Insprill:
There are quite a few MaterialTags that also exist in Bukkit's Tag like beds, signs, and coral.
But, I suppose it's just me being a bit too picky and wanting my tags to be from one place haha.

lynxplay:
Those exist because paper was faster than spigot/mojang 😛 So paper basically added the BEDS tags and stuff when spigot hasn't/minecraft didn't have those as tags in the first place.

Spigot or mojang then caught up leading to the doubled tags.

Insprill:
I see, thank you!

PR #8247, opened 2022-08-05
fix teleporting outside worldborder with enderpearl

zJanny:
Fixes being able to teleport behind the worldborder with an enderpearl

### DIFF
--- a/patches/server/0933-fix-teleporting-outside-worldborder-with-enderpearl.patch
+++ b/patches/server/0933-fix-teleporting-outside-worldborder-with-enderpearl.patch
@@ -1,0 +1,37 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: zJanny <jan.trummer@hotmail.com>
+Date: Fri, 5 Aug 2022 18:42:01 +0200
+Subject: [PATCH] fix teleporting outside worldborder with enderpearl
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java b/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
+index 23f77dfc6df93ef6a70e14a7e410263d39fae2cb..cdf2f250eb3b91d9583a89a63d78812dadb6ba3a 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
+@@ -18,6 +18,8 @@ import net.minecraft.world.phys.EntityHitResult;
+ import net.minecraft.world.phys.HitResult;
+ // CraftBukkit start
+ import org.bukkit.Bukkit;
++import org.bukkit.Location;
++import org.bukkit.WorldBorder;
+ import org.bukkit.craftbukkit.event.CraftEventFactory;
+ import org.bukkit.event.entity.CreatureSpawnEvent;
+ import org.bukkit.event.player.PlayerTeleportEvent;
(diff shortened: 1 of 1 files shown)


patches/server/0933-fix-teleporting-outside-worldborder-with-enderpearl.patch
```diff
@@ -0,0 +1,37 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java b/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
+index 23f77dfc6df93ef6a70e14a7e410263d39fae2cb..cdf2f250eb3b91d9583a89a63d78812dadb6ba3a 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/ThrownEnderpearl.java
+@@ -18,6 +18,8 @@ import net.minecraft.world.phys.EntityHitResult;
+ import net.minecraft.world.phys.HitResult;
+ // CraftBukkit start
+ import org.bukkit.Bukkit;
++import org.bukkit.Location;
++import org.bukkit.WorldBorder;
```
Owen1212055: In general, please try inlining imports and use fully qualified class names instead.

patches/server/0933-fix-teleporting-outside-worldborder-with-enderpearl.patch
```diff
@@ -0,0 +1,37 @@
...
++                    // Paper start - Check if enderpearl lands outside worldborder
++                    double size = getBukkitEntity().getWorld().getWorldBorder().getSize() / 2;
++                    Location center = getBukkitEntity().getWorld().getWorldBorder().getCenter();
++                    double x = location.getX() - center.getX();
++                    double z = location.getZ() - center.getZ();
++
++                    if((x > size || (-x) > size) || (z > size || (-z) > size)){
++                        return;
++                    }
++                    // Paper end
```
Owen1212055: Look at level.getWorldBorder().isWithinBounds() This logic is already done and will probably be generally much nicer.

There's also Level#isLoadedAndInBounds, which might be fine?

Owen1212055 (changes requested):
Welcome to Paper! 😄 

Here are some things, if you need any help feel free to ask in #paper-dev.
But a good reference is the [guide](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md).

Lulu13022002:
Maybe the teleportation should collide with the border instead of just be ignored, bc it can be frustrating in case of flee.

zJanny:
> Maybe the teleportation should collide with the border instead of just be ignored, bc it can be frustrating in case of flee.

There is already a Minecraft bug report for projectiles not colliding with the border which has been assigned low priority. 
So this will probably be fixed by Mojang in the future, this patch should just fix that players can use enderpearls to teleport out of bounds.

pop4959:
If the intended behavior is for the pearl to collide with the border I think I'd prefer that. If I'm understanding correctly this would just mean that the player loses the pearl and doesn't teleport, which feels hacky. (correct me if wrong)

zJanny:
> If the intended behavior is for the pearl to collide with the border I think I'd prefer that. If I'm understanding correctly this would just mean that the player loses the pearl and doesn't teleport, which feels hacky. (correct me if wrong)

(sry for closing the PR i missclicked)

I can try to implement the collision.

Lulu13022002:
Cool, i have already done that in an older version but basically the projectile raytrace start from the ProjectileUtil#getHitResult method
so this method should call your modified raytrace that take in account the border (in level.clip) but since this method is too generic your own logic should be implemented aside (see ClipContext.Block => COLLIDER)

PR #8248, opened 2022-08-05
Prevent wither skulls from using portals

### LINKED ISSUE #4241: Wither projectiles traveling through bedrock from the end to spawn
Wither projectile skulls stay and explode in the end when the wither is trapped in the center bedrock block of the portal such as in vanilla.

When attacking the wither the projectile skulls can travel through the portal to spawn and destroy the surroundings.

Create a wither in the end under the portal so that it's main head is in the center bedrock. Attack it and randomly the projectile skulls travel through the portal to spawn.

### Plugin list:
No plugins

### Paper version:
This server is running Paper version git-Paper-158 (MC: 1.16.2) (Implementing API version 1.16.2-R0.1-SNAPSHOT)

### Anything else:
I discovered this on a server running git-Paper-147 (MC: 1.16.2). So I then tested on my servers running git-Paper-158 (MC: 1.16.2)
Also have tested this on my 1.16.2 vanilla server and cannot recreate this issue.

In the pictures, left half is the end and right half is spawn. I had set mobGriefing to false so as to not destroy spawn for the pictures.  Attacks from players have the same results as attacks from iron golems.

Screenshot from 2020-08-28 19-59-02
Screenshot from 2020-08-28 19-59-46

JRoy:
Fixes #4241

### DIFF
--- a/patches/server/0933-Prevent-wither-skulls-from-using-portals.patch
+++ b/patches/server/0933-Prevent-wither-skulls-from-using-portals.patch
@@ -1,0 +1,17 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Josh Roy <joshroy126@gmail.com>
+Date: Fri, 5 Aug 2022 15:00:54 -0400
+Subject: [PATCH] Prevent wither skulls from using portals
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/WitherSkull.java b/src/main/java/net/minecraft/world/entity/projectile/WitherSkull.java
+index 2867e841e73a3edfdeb83af9d96e0d0cd4116a68..b601a69053f812aab354cb8379ff6e76a13f1192 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/WitherSkull.java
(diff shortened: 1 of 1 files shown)


lynxplay:
This *feels* like it should be fixed in whatever is breaking the collision checks no?
Not sure how well that would play with vanilla behaviour.
At least this would need some config option to re-enable this behaviour for whatever maniac uses wither skulls travelling through portals.

JRoy:
> This *feels* like it should be fixed in whatever is breaking the collision checks no?
> Not sure how well that would play with vanilla behaviour.
> At least this would need some config option to re-enable this behaviour for whatever maniac uses wither skulls travelling through portals.

Having the possibility for a random wither skull being teleported to the world spawn feels like something that shouldn't happen period lmfao

Owen1212055:
I mean, wouldn't this logic also apply to other projectiles? If a separate "prevent projectiles using end portals" config was added that would make sense, but, this is a bit of a behavior change that doesn't fix whatever the core issue actually is that may be causing this.

YanisBft (approved):

Owen1212055:
Closing, in the general, is just a work around and I feel that just blindly disabling wither skulls from being able to use portals it not the valid solution here. This is in general most likely caused by the collision optimizations that paper does.

Issue #8033, opened 2022-06-21
Tp out of Worldborder with Enderpearl


Chris6ix:
### Expected behavior

Not teleporting outside the worldborder

### Observed/Actual behavior

When throwing Enderpearl, you get teleported outside the worldborder

### Steps/models to reproduce

1.Tp to the Worldborder
2.Throw an Enderpearl at it
3.Get teleported outside the Worldborder

### Plugin and Datapack List

Plugins (55): ArmoredElytra, AutoTotem, BackToTheRoots*, BetterConcrete, BetterTridents, BigDoors, Bookshelf, BreakingSlab, ClientDetector, CommandCooldown, DeluxeMenus, Duels, DynamicShop, dynmap, Essentials, EssentialsChat, ExcellentEnchants, ExecutableItems, ExtraGameRules, FlyPerms, GSit, HolographicDisplays, HoloMobHealth, InteractionVisualizer, JoinRules, KittenForever, LevelledMobs, LightAPI, LinkedPortals, LinksPlus, LuckPerms, Multiverse-Core, Multiverse-Inventories, NexEngine, NotTooExpensive, PL-Hide, PlaceholderAPI, PlayerPortal, PlayerWarps, Playtimes*, ProtocolLib, QuickShop-Hikari, RotationWrench, SCore, SilkSpawners, TAB, UltimateCatcher, UltimateTimber, UltraCosmetics, Vault, WorldEdit, WorldEditSelectionVisualizer, WorldGuard, WorldGuardExtraFlags, XpBottles

There are 3 data packs enabled: [vanilla (built-in)], [file/bukkit (world)], [file/Chris6ix_Custom_Server_datapack.zip (world)]
There are no more data packs available

### Paper version

This server is running Paper version git-Paper-386 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 072f54c on ver/1.18.2)
You are running the latest version
Previous version: git-Paper-379 (MC: 1.18.2)

### Other

I'm not sure if this is a bug or if it's intended, but it's definitely not ok...

lynxplay:
Duplicate of [MC-53602](https://bugs.mojang.com/browse/MC-53602). 
I personally don't think there is any point in us fixing this, it is already on low priority on the mojang bug tracker, but I will leave the issue open for a core team/developer member to potentially close if they agree with my judgement.

Chris6ix:
[...]

PR #8252, opened 2022-08-06
Add thrown projectile collision with worldborder

zJanny:
https://streamable.com/ooz2oi

This is what i have come up with to detect if a thrown projectile hits the worldborder. 

Some input if this is a good idea would be appreciated.

### DIFF
--- a/patches/server/0933-Add-projectile-collision-with-worldborder.patch
+++ b/patches/server/0933-Add-projectile-collision-with-worldborder.patch
@@ -1,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: zJanny <jan.trummer@hotmail.com>
+Date: Sat, 6 Aug 2022 11:38:23 +0200
+Subject: [PATCH] Add projectile collision with worldborder
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/ProjectileUtil.java b/src/main/java/net/minecraft/world/entity/projectile/ProjectileUtil.java
+index 0524161f7cb414d526e6118258bddd989be6cc9a..9a5d44c057ea0cd18f2b30d49158b74d8826c41a 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/ProjectileUtil.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/ProjectileUtil.java
+@@ -34,6 +34,12 @@ public final class ProjectileUtil {
+             hitResult = hitResult2;
+         }
+ 
++        // Paper start
++        if(entity.getLevel().getWorldBorder().getCollisionShape().intersects(entity.getBoundingBox().expandTowards(entity.getDeltaMovement()))){
++            hitResult = new net.minecraft.world.phys.WorldborderHitResult(entity.position());
++        }
++        // Paper end
++
+         return hitResult;
+     }
+ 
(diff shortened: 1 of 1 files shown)


Lulu13022002:
See my comment in your old pr bc even if this can work you currently do the raytrace work on a part that will not even be used (outside of the border)

Lulu13022002:
So i have wrong in my last sentence (old pr) you don't need to add your own block type instead of collider but just to use the blockHitFactory bifunction into the clip method to implement your own logic through a new parameter in the clipcontext class
I'm still not sure how to get the dir parameter of the blockhitresult but this works (tested)
POC BlockGetter#clip method:
```Java
return (BlockHitResult) BlockGetter.traverseBlocks(context.getFrom(), context.getTo(), context, (raytrace1, blockposition) -> {
    if(context.withBorder() && finalLevel != null && !finalLevel.getWorldBorder().isWithinBounds(blockposition)) {
        return new BlockHitResult(new Vec3(blockposition.getX(), blockposition.getY(), blockposition.getZ()), [DIR], blockposition, false);
    }
    return this.clip(raytrace1, blockposition); // CraftBukkit - moved into separate method
}, ...
```

And if you want to add API into bukkit to support raytrace with border collide you can move this into the clip(raytrace, pos) method and add new parameter into the bukkit raytrace method

zJanny:
> So i have wrong in my last sentence (old pr) you don't need to add your own block type instead of collider but just to use the blockHitFactory bifunction into the clip method to implement your own logic through a new parameter in the clipcontext class I'm still not sure how to get the dir parameter of the blockhitresult but this works (tested) POC BlockGetter#clip method:
> 
> ```java
> return (BlockHitResult) BlockGetter.traverseBlocks(context.getFrom(), context.getTo(), context, (raytrace1, blockposition) -> {
>     if(context.withBorder() && finalLevel != null && !finalLevel.getWorldBorder().isWithinBounds(blockposition)) {
>         return new BlockHitResult(new Vec3(blockposition.getX(), blockposition.getY(), blockposition.getZ()), [DIR], blockposition, false);
>     }
>     return this.clip(raytrace1, blockposition); // CraftBukkit - moved into separate method
> }, ...
> ```
> 
> And if you want to add API into bukkit to support raytrace with border collide you can move this into the clip(raytrace, pos) method and add new parameter into the bukkit raytrace method

Thanks for the information, i did not know i could do it like that, but if you already implemented this, wouldnt it make more sense for you to PR this, instead of me doing work that has already been done.

Lulu13022002:
Yes that's true i can make a PR, but my feeling about this issue is the same that lynx (see https://github.com/PaperMC/Paper/issues/8033#issuecomment-1163670509), so i will not PR this. But if you really want to go ahead, here are some advices
- Go on discord, and see if the paper team is in favor of a fix or not to avoid useless work
- Make sure to stop the raytrace at the border limit (see my POC, if you use it give some credit)
- The idea to create a custom hit result is good but you can extends the blockhitresult to give more context
- My POC only teleport to the block before the border and not to the pearl, but you can improve this using the entity in the context (see ClipContext#collisionContext#getEntity() => nullable)
- The direction can be get using the nearest function from the Direction class of the two vectors (from, to) of the context
- The entity aabb expand to check entities, can cause some issue if an entity is found outside of the border, so must be restricted like the raytrace
- Also a last quirk is i think we should allow to pearl outside of the border when the origin is already outside (for that you can use the from vector)

PR #8253, opened 2022-08-06
Fix player desync on dimension change

### LINKED ISSUE #6511: Player Desync with Nether Portals
### Expected behavior

Expected player to appear in the portal.

### Observed/Actual behavior

Player appeared some distance away from the portal but did eventually correct itself. With low ping, it is possible to momentarily see yourself in a different location, but this corrects itself far quicker than when witnessing another player.

### Steps/models to reproduce

This issue has occurred on different worlds but I cannot tell you what the conditions are to replicate the issue but only describe what I have noticed. The world spawn is set to `0, 0` on both worlds, with a spawn radius of 0. The player will spawn in the same position each time, and the portal can be placed a maximum of 3 chunks away before the issue is no longer present. It seems to be related to the spawn chunks?

### Plugin list

None

### Paper version

This server is running Paper version git-Paper-222 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 159112e)
Previous version: git-Purpur-1335 (MC: 1.17.1)

### Agreements

### Other

I have narrowed down this issue to be Vanilla. I created a bug report with more detail here [https://bugs.mojang.com/browse/MC-235414](https://bugs.mojang.com/browse/MC-235414) in hope that it can be fixed for a later version of Minecraft. I was also advised to create an issue here in hopes that Paper can provide a patch in the meantime.

The issue was first noticed in Paper very early on and is still present with the latest version as of typing this. I tried with Tuinity and Purpur before trying Vanilla. The issue is present in Vanilla, and was not patched by Spigot and forks.

If a video is needed, I can provide one. I will answer any questions too to the best of my ability.

Owen1212055:
This caused a moment where the player entity is sent with the previous location, and the teleport packet which is sent shortly after is meant to correct that. Players with high ping might see it in the incorrect location.

Fixes #6511


kennytv (approved):

Owen1212055:
rebased

YanisBft (approved):

PR #5139, opened 2021-01-30
Force health update after updating attributes

Proximyst:
Currently, the health is updated in the player tick, THEN entity
trackers are ran and the attributes of a player are updated. This can
result in a health desync with the client, as the client will cap their
health at the known maximum health (e.g. we send 40.0F HP, but they think
the maximum is still 20.0F HP).

This can be alleviated by broadcasting the required attributes off the
bat in the player tick, if applicable.

An alternative implementation would be to set the lastHealthSent to
-1.0E8F, and let the player tick method send the health again the next
tick.

Fixes SPIGOT-6250.
Fixes #4793.

---

In this implementation, however, I am rather unsure if I should clear
the attributes fetched before the tracker handles it. It may result in
an extra packet being sent unnecessarily, which could potentially get
rather large as it contains all _modified_ ("tracked") attributes for
the specific entity (player in this case).

PR #8254, opened 2022-08-07
Don't send synced player data on join

### LINKED ISSUE #4793: Visual glitch with custom hearts container

Owen1212055:
Fixes https://github.com/PaperMC/Paper/issues/4793
Supercedes https://github.com/PaperMC/Paper/pull/5139

(Vanilla does not normally send entity data here at all)
The issue is it reintroduces some of the issues specified in: https://github.com/SpigotMC/BungeeCord/issues/2321

Opinions here?

### DIFF
--- a/patches/server/0933-Dont-send-synced-player-data-on-join.patch
+++ b/patches/server/0933-Dont-send-synced-player-data-on-join.patch
@@ -1,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Sat, 6 Aug 2022 22:20:32 -0400
+Subject: [PATCH] Dont send synced player data on join
+
+This was done inorder to fix an issue causing player synced data to be retained with joining another server.
+However, this seems to cause some issues, most significantly is that health on join is not sent correctly.
+
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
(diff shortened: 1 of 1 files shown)


Owen1212055:
Tux says that this issue it fixes probably doesn’t exist for velocity proxies but might still be an issue for bungeecord.

So this will need to be tested with velocity most likely, but if the issue is still prevalent with bungee im not sure what we should do here. Should this be behind a config or just activate this if a proxy is used and it isn’t velocity? (I’m not sure how you can just detect if ur using bungee) 

https://discord.com/channels/289587909051416579/908507886420910101/1023775699531739217

electronicboy:
The underlaying issue is also potentially fixed on bungeecord too, at least for recent releases with the dim registry

Owen1212055:
Ah nice, then I’ll test this on bungee when I get the chance to.

PR #8262, opened 2022-08-08
Fix Player#chat kicking all clients

JRoy:
@Machine-Maker


JRoy:
few fyis;
* i tried using an empty byte array in the header data too, didn't work. 
* the clients are disconnecting themselves for a BROKEN_CHAIN state

Machine-Maker (approved):

kennytv (approved):

lerokko:
What ever you did. In this build now all chat messages, regular, /me, /say, whispers, vanished text kick me for Chat message validation failure. Minecraft and plugin commands. Before that it was only 1 legacy plugin that did that.

electronicboy:
The change here should only impact plugins which try to faux command/chat execution, I'd be weary of plugins messing with stuff

PR #8267, opened 2022-08-08
[ci skip] Added the missing "d" character

ghost:


NoahvdAa (changes requested):
👋 Welcome to Paper, congrats on making your first PR! :)

Since this is a very small formatting change, this commit does not need its own build. To skip a build for this commit, simply add `[ci skip]` to the start of the PR title.

For this small change it probably won't really matter, but something to keep in mind for future PRs is: Patch files contain the hashes of the files they modify (see line 110). These will change if the content of the patch changes, meaning that patch files shouldn't be edited by hand since the hashes won't be updated (which can cause unnecessary conflicts when updating to a new MC version, though again, it probably doesn't matter here). See [this section](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#modifying-patches) of the contributing guide for information about how to edit patches.

MelonHell (approved):

electronicboy (approved):

Spliterash (approved):

PR #8272, opened 2022-08-09
Fix command pool submitting on shutdown

Owen1212055:
Don't submit to the command pool while on shutdown.

Submitting throws an exception, might just be better to wrap it in a try catch? Unsure.

### DIFF
--- a/patches/server/0296-Async-command-map-building.patch
+++ b/patches/server/0296-Async-command-map-building.patch
@@ -9,14 +9,18 @@ commands if the server is restarting. Using the default async pool caused issues
 due to the shutdown logic generally being much later.
 
 diff --git a/src/main/java/net/minecraft/commands/Commands.java b/src/main/java/net/minecraft/commands/Commands.java
-index 2bf67468a6c745bc6243c65210477ba129bfcb07..91b32a4856a4c9d6dc12871ed080f7a67311ead0 100644
+index 2bf67468a6c745bc6243c65210477ba129bfcb07..b89761343fd6125cef3b7949b39c56c2e1e513a2 100644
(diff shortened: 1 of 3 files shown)


Owen1212055:
Now simply has a discard policy, ignoring any submitted tasks that aren't approved.

electronicboy (approved):

patches/server/0296-Async-command-map-building.patch
```diff
@@ -9,14 +9,15 @@ commands if the server is restarting. Using the default async pool caused issues
...
+index 2bf67468a6c745bc6243c65210477ba129bfcb07..5e390e2e5e3af66cd7b596ecde62c8a29a00ffe6 100644
 --- a/src/main/java/net/minecraft/commands/Commands.java
 +++ b/src/main/java/net/minecraft/commands/Commands.java
-@@ -360,6 +360,23 @@ public class Commands {
+@@ -360,6 +360,25 @@ public class Commands {
          if ( org.spigotmc.SpigotConfig.tabComplete < 0 ) return; // Spigot
          // CraftBukkit start
          // Register Vanilla commands into builtRoot as before
 +        // Paper start - Async command map building
++        // Don't send to pool if it's shutdown
```
Machine-Maker: I think this comment was left over from before you found out about the discard policies?

Owen1212055: Yeah see paper contrib, I will change this later.

Owen1212055: done

Machine-Maker (approved):

jpenilla (approved):

PR #8274, opened 2022-08-09
Don't sync player rider

### LINKED ISSUE #7021: Most rideable entity statistic trackers (xxxx_one_cm) do not track movement correctly
### Expected behavior

Rideable entities, such as:
- Minecarts (minecart_one_cm)
- Striders (strider_one_cm)
- Pigs (pig_one_cm)
- Boats (boat_one_cm)
- Horses (horse_one_cm)
each have a statistic for tracking how far a player moves while riding them. Most of these don't track this statistic correctly.

- Minecarts function correctly.
- Striders should accurately track how far a player moves while riding them regardless of velocity or if the strider is being controlled with a fungus on a stick.
- Pigs should also accurately track this, regardless of velocity or if the pig is being controlled with a carrot on a stick.
- Boats should correctly track this statistic based on if the player is or isn't moving the boat, as well as how fast that player is going.
- Horses should also take into account if the player is moving at all, and how fast/slow that player may be moving.

### Observed/Actual behavior

While Minecarts work as expected and in parity with Vanilla Minecraft, the rest of the rideable entities do not.

"Works as intended" in all cases means that:
- If the entity you're riding isn't moving, your stat does not increase, AND
- If the entity you're riding IS moving, your stat does increase, AND
- Your statistic increases dependent on how far/fast the entity your riding moves

Minecart:
Vanilla: Works as intended
Paper: Works as intended

Strider:
Vanilla: Works as intended
Paper: Broken. If you get on the strider while not holding a "fungus on a stick", it will seemingly calculate the correct distance that the strider moves while you're riding it. If you pull out a fungus stick however, the stat will increase by a constant amount, even if that strider changes speed or is not moving because it's stuck.

Pig:
Vanilla: Works as intended
Paper: Broken. If you get on the pig while not holding a "carrot on a stick", it will seemingly calculate the correct distance that the pig moves while you're riding it. If you pull out a carrot stick however, the stat will increase by a constant amount, even if that pig changes speed or is not moving because it's stuck.

Boat:
Vanilla: Works as intended
Paper: Broken. The stat increases by a consistent amount (45 p/ tick) no matter what you're doing with the boat, even if you're not moving. The stat does not change based on motion whatsoever, it just constantly increases by 45 every tick.

Horse: 
Vanilla: Works as intended
Paper: Broken. The stat increases by a constant amount (85 p/ tick) even if you're not moving. If you jump with the horse, your stat briefly increases up to 122 p/ tick, before falling back down to 85.

### Steps/models to reproduce

I've done my best to explain the issue and what to observe in detail in this video, hope it helps:
https://www.youtube.com/watch?v=ZqKOwTL9DbE

This issue requires no prior setup and should be immediately visible once you're viewing the relevant affected statistic.

### Plugin and Datapack List

> pl
[20:13:21 INFO]: Plugins (0):

> datapack list
[20:13:31 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
[20:13:31 INFO]: There are no more data packs available

### Paper version

> version
[20:12:51 INFO]: This server is running Paper version git-Paper-36 (MC: 1.18) (Implementing API version 1.18-R0.1-SNAPSHOT) (Git: 76ee105)
Previous version: git-Paper-32 (MC: 1.18)

### Other

I found a Spigot ticket that discusses this issue from 6 months ago and it looks like it's abandoned. https://hub.spigotmc.org/jira/browse/SPIGOT-6475
One comment says they made a PR for it but the ticket is still open as unresolved and this issue still obviously persists.

According to md_5 in the same ticket, this issue goes back to at least Minecraft 1.13.2, I've been able to reproduce it on:
- Spigot 1.18
- Paper 1.17 build 391
- Paper 1.18 build 32
- Paper 1.18 build 36

This is not reproduceable in Vanilla Minecraft 1.18.

Here's the scoreboard commands so you don't need to type them out smile :^)
```
Minecart:
/scoreboard objectives add minecart_one_cm minecraft.custom:minecraft.minecart_one_cm
/scoreboard objectives setdisplay sidebar minecart_one_cm

Strider:
/scoreboard objectives add strider_one_cm minecraft.custom:minecraft.strider_one_cm
/scoreboard objectives setdisplay sidebar strider_one_cm

Pig:
/scoreboard objectives add pig_one_cm minecraft.custom:minecraft.pig_one_cm
/scoreboard objectives setdisplay sidebar pig_one_cm

Boat:
/scoreboard objectives add boat_one_cm minecraft.custom:minecraft.boat_one_cm
/scoreboard objectives setdisplay sidebar boat_one_cm

Horse:
/scoreboard objectives add horse_one_cm minecraft.custom:minecraft.horse_one_cm
/scoreboard objectives setdisplay sidebar horse_one_cm
```

### LINKED ISSUE #8214: player.getLocation() returns incorrect Y value when riding a mob
### Expected behavior

when a player mounts a horse, their y level increases by 0.85, the ride height of the horse, as indicated on their f3 and by running minecraft:tp ~ ~ ~ to get position, this should be reflected by player.getLocation().

### Observed/Actual behavior

Calling player.getLocation() returns the y value of the player's vehicles feet as when they are not on the vehicle at all.

### Steps/models to reproduce

Check every tick player.getLocation(), mount a horse, at for example y level 60, when you are on the horse it will correctly return your Y level as 60.85, but once you add a saddle, enabling control of the vehicle, the y level glitches between 60.85 and 60.0 randomly.

### Plugin and Datapack List

N/A

### Paper version

git-Paper-"f090298" (MC: 1.19)

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/8214

craftbukkit apply a sync of the position of the rider in the same time that the vehicle move (probably for the later event)
but the sync here isn't enough powerfull to handle the rider Y offset and even applying a total positionRider will not solve all the issue.

---
Download the paperclip jar for this pull request: [paper-8274.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/810655063.zip)

### DIFF
--- a/patches/server/0948-Don-t-teleport-player-rider.patch
+++ b/patches/server/0948-Don-t-teleport-player-rider.patch
@@ -14,7 +14,7 @@ But this patch prevent the teleportation that upstream
 do to sync the rider with its vehicle
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index 0c2255b6e2fb7752f85b0f83d4f84732758bd14d..5d706b007ccca4d8f8cbaddc1daa7e3dbee66534 100644
+index 0c2255b6e2fb7752f85b0f83d4f84732758bd14d..7c50b462b4c45dfcf92c6515ffdd023dba24d8e0 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 @@ -681,7 +681,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
@@ -35,6 +35,15 @@ index 0c2255b6e2fb7752f85b0f83d4f84732758bd14d..5d706b007ccca4d8f8cbaddc1daa7e3d
                      this.connection.send(new ClientboundMoveVehiclePacket(entity));
                      return;
                  }
+@@ -702,7 +702,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                 // CraftBukkit start - fire PlayerMoveEvent
+                 Player player = this.getCraftPlayer();
+                 // Spigot Start
+-                if ( !this.hasMoved )
++                if ( false && !this.hasMoved ) // Paper unneeded and will works only the first time
+                 {
+                     this.lastPosX = curPos.getX();
+                     this.lastPosY = curPos.getY();
 @@ -717,7 +717,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
  
                  // If the packet contains movement information then we update the To location with the correct XYZ.
@@ -44,3 +53,20 @@ index 0c2255b6e2fb7752f85b0f83d4f84732758bd14d..5d706b007ccca4d8f8cbaddc1daa7e3d
                  to.setZ(packet.getZ());
  
  
+@@ -739,6 +739,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                     // Skip the first time we do this
+                     if (true) { // Spigot - don't skip any move events
+                         Location oldTo = to.clone();
++                        Location currentPos = player.getLocation(); // Paper
+                         PlayerMoveEvent event = new PlayerMoveEvent(player, from, to);
+                         this.cserver.getPluginManager().callEvent(event);
+ 
+@@ -758,7 +759,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+ 
+                         // Check to see if the Players Location has some how changed during the call of the event.
+                         // This can happen due to a plugin teleporting the player instead of using .setTo()
+-                        if (!from.equals(this.getCraftPlayer().getLocation()) && this.justTeleported) {
++                        if (!currentPos.equals(this.getCraftPlayer().getLocation()) && this.justTeleported) { // Paper - fix this being called one time on the first ride even if the event does nothing
+                             this.justTeleported = false;
+                             return;
+                         }
--- a/patches/server/0948-Don-t-teleport-player-rider.patch
+++ b/patches/server/0948-Don-t-teleport-player-rider.patch
@@ -3,12 +3,18 @@ From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
 Date: Thu, 8 Dec 2022 18:52:05 +0100
 Subject: [PATCH] Don't teleport player rider
 
+In the PlayerMoveEvent, the rider Y from the to location is wrong,
+but since mojang hasn't exposed a good way
+to get the rider offset there's no way to do it without emulating
+the positionRider which isn't long term viable (cc camel)
+But this patch prevent the teleportation that upstream
+do to sync the rider with its vehicle
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index d31a345edfffe39f127073fc3aec8b3489bae79c..b3b68219e0f794ee942197b7aa666bfef4c6910c 100644
+index 0c2255b6e2fb7752f85b0f83d4f84732758bd14d..7931c339e8e6b60e2809b446c7c61e8caf24f6e5 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-@@ -689,7 +689,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+@@ -681,7 +681,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
                  Location curPos = this.getCraftPlayer().getLocation(); // Spigot
  
                  entity.absMoveTo(d3, d4, d5, f, f1);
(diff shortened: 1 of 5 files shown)


patches/server/0933-Apply-offset-when-riding-entity.patch
```diff
@@ -0,0 +1,29 @@
...
+index dbefefc123e3cb38a95bd321d74c384bc087cfba..d04344d0559ddbc20ea8f1f1465b56249ddd5f84 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -692,7 +692,8 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                 Location curPos = this.getCraftPlayer().getLocation(); // Spigot
+ 
+                 entity.absMoveTo(d3, d4, d5, f, f1);
+-                this.player.absMoveTo(d3, d4, d5, this.player.getYRot(), this.player.getXRot()); // CraftBukkit
++                double riderOffset = entity.getPassengersRidingOffset() + this.player.getMyRidingOffset(); // Paper
++                this.player.absMoveTo(d3, d4 + riderOffset, d5, this.player.getYRot(), this.player.getXRot()); // CraftBukkit // Paper
```
Owen1212055: Looking at this, do we even need to sync the players position like this? I really think we should look more indepth if this is actually needed. Because obviously it's desyncing the entire player, but it might not even be needed in the first place.

The only possible place I could think of is
```java
// Check to see if the Players Location has some how changed during the call of the event.
                        // This can happen due to a plugin teleporting the player instead of using .setTo()
                        if (!from.equals(this.getCraftPlayer().getLocation()) && this.justTeleported) {
                            this.justTeleported = false;
                            return;
                        }
```

This PlayerMoveEvent that is being fired here just seems wrong tbh, as it's not the player that's moving it's the entity. And bukkit defo tries to shove the player in the vehicle inorder to pretend this issue doesn't exist.
```java
Location from = new Location(player.getWorld(), this.lastPosX, this.lastPosY, this.lastPosZ, this.lastYaw, this.lastPitch); // Get the Players previous Event location.
                Location to = player.getLocation().clone(); // Start off the To location as the Players current location.

                // If the packet contains movement information then we update the To location with the correct XYZ.
                to.setX(packet.getX());
                to.setY(packet.getY());
                to.setZ(packet.getZ());


                // If the packet contains look information then we update the To location with the correct Yaw & Pitch.
                to.setYaw(packet.getYRot());
                to.setPitch(packet.getXRot());
```

What we need to do is "estimate" where the player is going to be moving TO, as this location is the vehicle location and not the players.
Take a look at Entity#positionRider(Entity passenger, Entity.MoveFunction positionUpdater).
What we can do is make our own MoveFunction which updates the TO location based on the offset of the player.

This should allow:
1. Not needing to move the player at all
2. Correctly estimate the new PlayerMoveEvent location

Owen1212055:
I think what we should do here is just make from and to the PREVIOUS position of the rider and then the position sent in the packet. Currently that’s the behavior of the event, except spigot actually mutates the player currently (which isn’t needed). 

So yes it won’t be the actual correct location of the player, but it matches the behavior it has been for a while since we really don’t have a nice way to simulate this and in all honesty I don’t see it being really that worth it.

patches/server/0948-Don-t-teleport-player-rider.patch
```diff
@@ -0,0 +1,49 @@
...
+                 }
+                 // Spigot End
+-                Location from = new Location(player.getWorld(), this.lastPosX, this.lastPosY, this.lastPosZ, this.lastYaw, this.lastPitch); // Get the Players previous Event location.
++                Location from = player.getLocation(); // Paper - Get the actual player location (since their position is updated in the next tick, this become the from loc)
+                 Location to = player.getLocation().clone(); // Start off the To location as the Players current location.
+ 
+                 // If the packet contains movement information then we update the To location with the correct XYZ.
+                 to.setX(packet.getX());
+-                to.setY(packet.getY());
++                to.setY(packet.getY()); // todo wrong rider Y pos
```
Owen1212055: can probably remove this

Owen1212055 (approved):
Generally, approved.

The calls to move the player are not needed when moving the vehicle's rider. 
This can be seen in vanilla, where there are no calls that teleport the player.
Additionally paper calls its own logic which does not move the player either.
```java
// Paper start - Prevent moving into unloaded chunks
                if (player.level.paperConfig().chunks.preventMovingIntoUnloadedChunks && (
                    !worldserver.areChunksLoadedForMove(this.player.getBoundingBox().expandTowards(new Vec3(toX, toY, toZ).subtract(this.player.position()))) ||
                        !worldserver.areChunksLoadedForMove(entity.getBoundingBox().expandTowards(new Vec3(toX, toY, toZ).subtract(entity.position())))
                    )) {
                    this.connection.send(new ClientboundMoveVehiclePacket(entity));
                    return;
                }
// Paper end
```


This solution does NOT change the behavior of this event, as the last position would have been the player's location anyways. (Yes, it goes from player location to vehicle location, but that is the current broken behavior which I am not sure should be fixed)

Owen1212055:
Would appreciate it if anyone would test to ensure that riding entities seem fine. 
Seems fine to me, but, just incase! 😄 

And please confirm that the rideable stats bug is fixed too.

Lulu13022002:
I re-put this pr as a draft there's more stuff i need to test for the plugin event before

Owen1212055:
I’d argue that the event isn’t that important, it’s been broken for a while… I think the main benefit here is the actual fix of not teleporting the player like this.

Lulu13022002:
nvm you're right, i think this event shouldn't be called here in the first place

Owen1212055:
It was brought up in stream. Leaf said that this is needed in order to ensure that the player position is correctly synced in order to avoid exploits. This should be looked into.

electronicboy (approved):

PR #8276, opened 2022-08-10
Fix map cursors

### LINKED ISSUE #8229: Adding text to a Map Cursor (via a plugin) causes all map cursors on that map to display a blank text field
### Expected behavior

MapCursors with set captions display their text beneath them, and cursors without captions have no text and the text field is absent.

### Observed/Actual behavior

I'm writing a plugin that adds captions (using setCaption()) to MapCursors that aren't banners. Whenever I do this in Paper, every cursor on the map, including the player location icon, starts displaying a blank text field beneath it. Removing those specific cursors again by changing the plugin's code causes the issue to go away.

### Steps/models to reproduce

Acquire or create any plugin that allows you to add a caption to a MapCursor. Add a captioned cursor to a map. You'll observe that any additional cursor added to the map, including the one marking the player's location, will now display a text field underneath it containing no text (it'll look like a dark vertical line).

### Plugin and Datapack List

Plugins: Only my own plugin that I'm writing, which applied the captions to the cursors.
Datapacks: Vanilla and the Bukkit default datapack (both unmodified).

### Paper version

This server is running Paper version git-Paper-103 (MC: 1.19.1) (Implementing API version 1.19.1-R0.1-SNAPSHOT) (Git: c032050)

### Other

I've tested this on both Spigot and Paper, using the exact same plugin, world, and map item. The issue is ONLY present in Paper. Additionally, the vanilla-Minecraft method of adding text to maps (naming a banner and then clicking it with the map) doesn't cause the issue to occur.

TheTuso:
Fix [#8229](https://github.com/PaperMC/Paper/issues/8229)

### DIFF
--- a/patches/server/0008-Adventure.patch
+++ b/patches/server/0008-Adventure.patch
@@ -4230,6 +4230,19 @@ index 1980240d3dc0331ddf2ff56e163e2bfbd3b231ab..7a7f3f53aef601f124d474d9890e23d8
          @Override
          public Inventory createInventory(InventoryHolder holder, InventoryType type, String title) {
              // BrewingStand does not extend TileEntityLootable
+diff --git a/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java b/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
+index b47f18b1e448807a17ca9f2ae9609680233da837..9683d7d103af66fffd68c11abc38fb4fd2f99482 100644
+--- a/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
++++ b/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
+@@ -42,7 +42,7 @@ public class CraftMapRenderer extends MapRenderer {
+             }
+ 
+             MapDecoration decoration = worldMap.decorations.get(key);
+-            cursors.addCursor(decoration.getX(), decoration.getY(), (byte) (decoration.getRot() & 15), decoration.getType().getIcon(), true, CraftChatMessage.fromComponent(decoration.getName()));
++            cursors.addCursor(decoration.getX(), decoration.getY(), (byte) (decoration.getRot() & 15), decoration.getType().getIcon(), true, decoration.getName() == null ? null : io.papermc.paper.adventure.PaperAdventure.asAdventure(decoration.getName())); // Paper
+         }
+     }
+ 
 diff --git a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
 index 0d25c7c03f7ac21a4b21bb95b5bd921c43430cf9..b7f0277b50a0f45c32b818bf9fe1218874aa8533 100644
 --- a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java


patches/server/0933-Fix-map-cursors.patch
```diff
@@ -0,0 +1,36 @@
...
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java b/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
+index b47f18b1e448807a17ca9f2ae9609680233da837..5ee6a376558a08fc112cfe6e8e93e96b1802dab5 100644
+--- a/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
++++ b/src/main/java/org/bukkit/craftbukkit/map/CraftMapRenderer.java
+@@ -1,14 +1,11 @@
+ package org.bukkit.craftbukkit.map;
+ 
++import io.papermc.paper.adventure.PaperAdventure;
```
lynxplay: Please inline these imports and don't remove/edit existing imports.

lynxplay (changes requested):

patches/server/0933-Fix-map-cursors.patch
```diff
@@ -0,0 +1,36 @@
...
++import org.bukkit.map.*;
+ 
+ public class CraftMapRenderer extends MapRenderer {
+ 
+@@ -42,7 +39,7 @@ public class CraftMapRenderer extends MapRenderer {
+             }
+ 
+             MapDecoration decoration = worldMap.decorations.get(key);
+-            cursors.addCursor(decoration.getX(), decoration.getY(), (byte) (decoration.getRot() & 15), decoration.getType().getIcon(), true, CraftChatMessage.fromComponent(decoration.getName()));
++            cursors.addCursor(new MapCursor(decoration.getX(), decoration.getY(), (byte) (decoration.getRot() & 15), decoration.getType().getIcon(), true, decoration.getName() == null ? null : PaperAdventure.asAdventure(decoration.getName()))); // Paper
```
lynxplay: Additionally, not much need to create the new MapCursor here.
The constructor is deprecated for the same reason the plain `addCursor` method is deprecated, it uses the magic value from `decoration.getRot()`. This will not change unless you actually parse the api representation of the rotation, which is overkill.

I'd argue to just revert back/move to the `addCursor(int x, int y, byte direction, byte type, boolean visible, @Nullable net.kyori.adventure.text.Component caption)` call

TheTuso: ✅

NoahvdAa (commented):
👋 Welcome to Paper, congrats on making your first PR! :)

Looks good code wise, but this probably doesn't need its own patch. Like lynx mentioned in his original comment, this should probably go into the [Adventure patch](https://github.com/PaperMC/Paper/blob/2a74071bcb9d633f3cddccd974b1e926e37ed326/patches/server/0008-Adventure.patch). Since you already already created a new patch, it's probably easiest to fixup the commit, see [this section in the contributing guide](https://github.com/PaperMC/Paper/blob/2a74071bcb9d633f3cddccd974b1e926e37ed326/CONTRIBUTING.md#manual-method).

NoahvdAa (approved):
Works, thanks for the PR!

lynxplay (approved):

electronicboy (approved):

kennytv (approved):

PR #8280, opened 2022-08-11
EntityCombustByBlockEvent not called when ItemEntity burned by lava.

orang3-juic3:
[Javadoc](https://jd.papermc.io/paper/1.17/org/bukkit/event/entity/EntityCombustByBlockEvent.html#getCombuster()) states that EntityCombustByBlockEvent's combuster can be lava, implying that it is called when an entity is combusted by lava. This functions fine for all LivingEntities, however when the Entity is an ItemEntity this is overlooked, even though it can be set on fire (and this event is called when an Item enters a fire block). This pr changes the server's behaviour to also fire the event when an ItemEntity is combusted by lava.

### DIFF
--- a/patches/server/0934-EntityCombustByBlockEvent-should-be-called-when-Item.patch
+++ b/patches/server/0934-EntityCombustByBlockEvent-should-be-called-when-Item.patch
@@ -1,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Alex <alexzatstone@gmail.com>
+Date: Thu, 11 Aug 2022 23:13:30 +0200
+Subject: [PATCH] EntityCombustByBlockEvent should be called when ItemEntity is
+ burned by lava, this is currently not what happens, probably because
+ ItemEntity is not a subclass of LivingEntity
+
+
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Why limit this to only LivingEntity in the first place? Why can't we just remove that instanceof check in the first place?

orang3-juic3:
> Why limit this to only LivingEntity in the first place? Why can't we just remove that instanceof check in the first place?

I agree with you, the reason the new code still has that original instanceof check was because I was scared that I overlooked something that removing it would break.

orang3-juic3:
I did a bit of investigating, and it seems that the EntityCombustByBlockEvent in BaseFireBlock (which is working correctly) is fired by the entityInside method in the same class. I looked at what classes call this method to determine which classes should fire the event for the lava case, and now their behaviour should be virtually the same, apart from the fact that the event, in the fire case, is only fired when the entity enters the block, whereas in the lava case the event seems to be called for every tick the entity is in lava.

Owen1212055:
Why do we need an instance of check at ALL? I'd think that all entities here should be fine?

Machine-Maker:
I agree with @Owen1212055. We can just remove the check. No such check exists for the fire call.

PR #8285, opened 2022-08-13
Add LocalCoordinates API

chrisgdt:


chrisgdt:
Just in case you wonder, I tested it and everything is fine (the second commit is because I forgot that a Vec2 representing yaw and pitch is x=pitch and y=yaw, not like the debug screen F3 that shows `(yaw, pitch)`, I inverted them and it's ok !)

patches/api/0392-Add-LocalCoordinates-API.patch
```diff
@@ -0,0 +1,68 @@
...
+@@ -903,6 +903,27 @@ public final class Bukkit {
+         return server.createExplorerMap(world, location, structureType, radius, findUnexplored);
+     }
+ 
++    // Paper start
++    /**
++     * Computes local coordinates according to the given location
++     * and coordinates.
++     * <br>
++     * Local coordinates are used in vanilla with <code>^x ^y ^z</code> and
```
MiniDigger, line 20: doesn't vanilla use tilde (~)?

NoahvdAa: Those are for [relative coordinates](https://minecraft.fandom.com/wiki/Coordinates#Relative_world_coordinates), [local coordinates](https://minecraft.fandom.com/wiki/Coordinates#Local_coordinates) also take rotation into account.

Lulu13022002: Yes it's just a little sad that this cannot be moved into the Location class due to nms needs

NoahvdAa: Can't we just:
a) Copy the calculation logic into the location class?
or
b) Shove the NMS call into unsafe and call that from the Location class?

Lulu13022002: Yes both will work but the unsafe class is meh...
Redoing the calculation is probably fine since it's just pure math, i think i prefer this over the unsafe use.
We should just make sure to sync with the vanilla impl

Machine-Maker: If this has anything to do with Location, it should probably wait until the Position API is in. See the pr for that.

chrisgdt: Both ~ and ^ are used in vanilla, but we cannot mix them (we can use direct or relative coordinates but if we want to use local, the three components must be local coordinates).

The exact same calculation logic using API methods is the following (partially copied from NMS) : 
```
private static final float[] SIN = make(new float[65536], (sineTable) -> {
    for (int i = 0; i < sineTable.length; ++i) {
        sineTable[i] = (float)Math.sin((double)i * Math.PI * 2.0D / 65536.0D);
    }
});

private static <T> T make(T object, Consumer<T> initializer) {
    initializer.accept(object);
    return object;
}

private static float sin(float value) {
    return SIN[(int)(value * 10430.378F) & '\uffff'];
}

private static float cos(float value) {
    return SIN[(int)(value * 10430.378F + 16384.0F) & '\uffff'];
}

public static Location getLocalCoord(double left, double up, double forwards, Location origin) {
    float f = cos((origin.getYaw() + 90.0F) * ((float)Math.PI / 180F));
    float g = sin((origin.getYaw() + 90.0F) * ((float)Math.PI / 180F));
    float h = cos(-origin.getPitch() * ((float)Math.PI / 180F));
    float i = sin(-origin.getPitch() * ((float)Math.PI / 180F));
    float j = cos((-origin.getPitch() + 90.0F) * ((float)Math.PI / 180F));
    float k = sin((-origin.getPitch() + 90.0F) * ((float)Math.PI / 180F));

    Vector vec32 = new Vector(f * h, i, (double)(g * h));
    Vector vec33 = new Vector(f * j, k, (double)(g * j));
    Vector vec34 = vec32.clone().crossProduct(vec33).multiply(-1.0D);
    double d = vec32.getX() * forwards + vec33.getX() * up + vec34.getX() * left;
    double e = vec32.getY() * forwards + vec33.getY() * up + vec34.getY() * left;
    double l = vec32.getZ() * forwards + vec33.getZ() * up + vec34.getZ() * left;
    return new Location(origin.getWorld(), origin.getX() + d, origin.getY() + e, origin.getZ() + l);
}
```

Before discovering it, I had another code but it wasn't the same, especially when `left` is not zero. When I saw that my code wasn't the same local coordinates than the vanilla behavior, I created the PR.
```
Vector dirX = new Location(origin.getWorld(), 0, 0, 0, Location.normalizeYaw(origin.getYaw()-90), origin.getPitch()).getDirection().normalize();
Vector dirY = new Location(origin.getWorld(), 0, 0, 0, origin.getYaw(), origin.getPitch()-90).getDirection().normalize();
Vector dirZ = origin.getDirection().normalize();
return origin.add(dirX.multiply(left)).add(dirY.multiply(up)).add(dirZ.multiply(forwards));
```

I just wanted to link the actual behavior by having the same *sin* table (no idea if we can just replace sin and cos methods by Math.sin and Math.cos). I also though to set this in the Location class, but since it is located in the API, get a direct link to NMS wasn't a clear idea for me. I didn't add relative coordinates here because it is really easy for developers to just add an offset (actually, an API method like `Location#parse(String)` that takes a coordinate string, e.g. `~1 10 ~-4` or `^ ^1 ^5` would be great for everybody).

There is probably some better ways to add a method for local coordinates, so yes. The Position API is #7639 right ?

Lulu13022002: Oh i have forgot that cos/sin wasn't exposed api side. So maybe use unsafe then...
For the difference between cos/sin table and the native java usage (if you see the impl of cos/sin you will see that use a native method that rely on some low level C method: if you are curious see: https://github.com/openjdk/jdk/blob/master/src/java.base/share/native/libfdlibm/s_cos.c) it's just an optimization that minecraft made for performance (it cache some values so the result will loss some accuracy). It's probably why the bukkit equivalent with getDirection() is different. But i think this should return the exact same result as vanilla one so don't use native cos/sin.

Owen1212055:
I think that this should wait until https://github.com/PaperMC/Paper/pull/7639
As then we will just be able to make a new Position object which represents an offset like this.

Warriorrrr:
With the position PR merged, this could probably be looked at again.

patches/server/0935-Add-LocalCoordinates-API.patch
```diff
@@ -0,0 +1,41 @@
...
++++ b/src/main/java/net/minecraft/commands/arguments/coordinates/LocalCoordinates.java
+@@ -24,6 +24,12 @@ public class LocalCoordinates implements Coordinates {
+     public Vec3 getPosition(CommandSourceStack source) {
+         Vec2 vec2 = source.getRotation();
+         Vec3 vec3 = source.getAnchor().apply(source);
++        // Paper start
++        return getPosition(vec3, vec2);
++    }
++
++    public Vec3 getPosition(Vec3 vec3, Vec2 vec2) {
```
lynxplay, line 19: Maybe this should be a static method also taking the x y and z (left, up, forwards).
That way we don't have to allocate another throwaway object when calling this.

PR #8287, opened 2022-08-13
Add method to get highest anger level of warden

kennytv:
Passing null into the vanilla method returns the anger level towards whatever entity it has the highest anger


kennytv:
Very cool pr, thank you @kennytv

PR #5503, opened 2021-04-17
add ability to get biomekey directly rather than using the biome Enum

ExcessiveAmountsOfZombies:
https://github.com/PaperMC/Paper/pull/4890
I fixed it after I broke it

PR #8292, opened 2022-08-14
Add NamespacedKey biome methods

JRoy:
Methods are in UnsafeValues due to the possibility of them being removed in the future when we do more chunkgen/biome stuff, but for now this is the only way to get the underlying key for actual biomes.

Rebases and Closes #5503

### DIFF
--- a/patches/api/0391-Add-NamespacedKey-biome-methods.patch
+++ b/patches/api/0391-Add-NamespacedKey-biome-methods.patch
@@ -1,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Josh Roy <10731363+JRoy@users.noreply.github.com>
+Date: Sun, 14 Aug 2022 12:22:54 -0400
+Subject: [PATCH] Add NamespacedKey biome methods
+
+Co-authored-by: Thonk <30448663+ExcessiveAmountsOfZombies@users.noreply.github.com>
+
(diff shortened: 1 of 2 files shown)


patches/api/0391-Add-NamespacedKey-biome-methods.patch
```diff
@@ -0,0 +1,52 @@
...
+
+diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
+index 4fcafddf3792b66c618f91e04d102f374de565a8..4c58fb860940947bf8fb404f717faaaa9b0f04f5 100644
+--- a/src/main/java/org/bukkit/UnsafeValues.java
++++ b/src/main/java/org/bukkit/UnsafeValues.java
+@@ -11,6 +11,7 @@ import org.bukkit.inventory.ItemStack;
+ import org.bukkit.material.MaterialData;
+ import org.bukkit.plugin.InvalidPluginException;
+ import org.bukkit.plugin.PluginDescriptionFile;
++import org.jetbrains.annotations.NotNull;
```
kennytv: Are my eyes tricking me or is that an IMPORT

electronicboy (approved):

kennytv (approved):

patches/api/0391-Add-NamespacedKey-biome-methods.patch
JRoy: BINGQLIN NO IDEA WHAT YOU'RE TALKING ABOUT BINGQLIN

Issue #8289, opened 2022-08-13
Set a maximum thread limit for the BukkitScheduler


Vankka:
### Is your feature request related to a problem?

The BukkitScheduler's async thread pool [doesn't have a maximum thread limit](https://github.com/PaperMC/Paper/blob/41647af74caed955c1fd5b38d458ee59298ae5d4/patches/server/0183-Improved-Async-Task-Scheduler.patch#L80) which can easily lead to `java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached` if a plugin/multiple plugins run(s) `runTaskAsynchronously` for hundreds/thousands (depending on os thread limit) of tasks without considering/knowing about there being no limit on threads being used concurrently.

### Describe the solution you'd like.

Limit the maximum amount of concurrent threads created by the BukkitScheduler. The limit also should not be too *low* since there might be tasks that are long lived.

Additionally it might be worth considering if the amount of concurrent tasks should be limited per plugin to avoid a single plugin holding up the tasks of all plugins.

### Describe alternatives you've considered.

Plugins may make their own thread pool with a maximum thread limit, but plugin developers might not know that using the Scheduler to run thousands of tasks might cause the above error to occur.

A log message would help plugin developers and server administrators find and fix the issue, right now you need to get a thread dump to figure out what plugin(s) is/are using a lot of threads or test removing plugins to find which one is causing the issue (if it can be consistently reproduced).

electronicboy:
I've considered storing the thread pools per plugin in the past, but generally have been iffy about it as it creates some concerns over the performance vs just one pool given how infrequently stuff runs; I think an argument could be made for having at least a system property to control that, but, setting a limit on it is a huge "piss in the wind" type thing towards tryna determine a suitable number which I'm not sure we wanna deal with

SrMonsterYT:
> ### Sua solicitação de recurso está relacionada a um problema?
> O pool de threads de Bukkit do Bukkit [não tem](https://github.com/PaperMC/Paper/blob/41647af74caed955c1fd5b38d458ee59298ae5d4/patches/server/0183-Improved-Async-Task-Scheduler.patch#L80) facilmente um plugin que pode ser facilmente a que um plugin será capaz de levar os threads a serem executados / limite em threads sendo usados ​​simultaneamente.`java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached``runTaskAsynchronously`
> 
[...]

PR #8293, opened 2022-08-14
Fix: Add async thread pool limit CraftAsyncScheduler.java

DrupalDoesNotExists:
https://github.com/PaperMC/Paper/issues/8289#issue-1338013513


electronicboy:
1) incorrect formatting, should just be a single line comment, noting the change
2) as I originally said, any form of number as a default would be a piss in the wind, i.e. the number here is 2x the total allowed threads by ptero, the no.1 contender for issues here, so this seems like a really bad default
3) if we really do add this, it must be configurable, especially given point 2

PR #8294, opened 2022-08-14
Fix set profile patch issues

Owen1212055:
The old setPlayerProfile would cause a duplicate player to appear in tab list because the previous game profile was not removed properly. This caused it to not work at all.

This also moves a bunch of the deprecation patches to the correct patch.
However, it might be better to merge those patches together (merger plz).


Lulu13022002:
Theses two patch (https://github.com/PaperMC/Paper/blob/master/patches/api/0175-Entity-getEntitySpawnReason.patch and https://github.com/PaperMC/Paper/blob/master/patches/api/0226-Clarify-the-Javadocs-for-Entity.getEntitySpawnReason.patch)
should be also merge i think

kennytv:
Please split the fix and patch switcheroos into different prs, or at least different commits, also with more comments on what changed and why; as is this is unreviewable

Owen1212055:
Yeah, i'll close this and open new ones when I get to it.

Issue #8246, opened 2022-08-05
The request is about nether portal search, and preventing players entering the roof in a better way for old servers.


TheForgTN:
### Is your feature request related to a problem?

This would be a more pleasent way, to prevent players entering the nether roof on servers, where nether roof where enabled back in time.

### Describe the solution you'd like.

The request is about nether portal search, and generating a new portal if, an existing found above the max Y height (aka abow roof).

it could be an option like 

max-nether-portal-search-height: 

or it might would be even better if the server doesn't even search for a portal in the nether above a given height

### Describe alternatives you've considered.

The void damage is a great way (actually I would use both), but on a server where, there are existing portals, it's like a gambling, if you got yourself on the roof, well thats annoying.

Owen1212055:
I really feel that this could be easily achieved with something like a plugin.

BigCrazyOfficial:
This could be achieved more easily by making portals above this limit invalid locations for entering the nether. ~~which I just made a PR for hey isn't that convenient~~

Machine-Maker:
I don't think this is a concern anymore. You will not travel to or create portals above the bedrock roof if the void damage is enabled, even if the portal existed before that option was enabled.

PR #8295, opened 2022-08-14
Only find exit portal if in height bounds

DrupalDoesNotExists:
https://github.com/PaperMC/Paper/issues/8246

(Second try)


Doc94:
Hi, for PR related to new configs you need edit this patch https://github.com/PaperMC/Paper/blob/master/patches/server/0004-Paper-config-files.patch and later create a new patch using this new config.

Machine-Maker (changes requested):
I don't think you need to add a "portal search height limit" config option. That is the same thing as the logical height for the world. This can be changed via datapacks, and those changes should be accounted for automatically instead of someone needing to also change it in the world configuration file.

Issue #8283, opened 2022-08-13
Warden spawned by shrieker uses DEFAULT SpawnReason


EvelynBeebs:
### Expected behavior

with 80+ wardens in 1 spot no other mobs should be spawning

### Observed/Actual behavior

no other mobs should be spawning

### Steps/models to reproduce

get 80 wardens spawned look around mobs there is

### Plugin and Datapack List

![image](https://user-images.githubusercontent.com/72664293/184466437-79c58a1b-5284-4a4e-ba4e-601878b039fd.png)

### Paper version

![image](https://user-images.githubusercontent.com/72664293/184466455-5eabd288-c0df-4246-bd58-9e49fadd8865.png)

e-im:
try setting `count-all-mobs-for-spawning` to `true` in the `config/paper-world-defaults.yml`. this is likely WAI -- how did you spawn the wardens?

Machine-Maker:
Should warden's spawn with the NATURAL reason?

EvelynBeebs:
@e-im 
> try setting `count-all-mobs-for-spawning` to `true` in the `config/paper-world-defaults.yml`. this is likely WAI -- how did you spawn the wardens?


ah that's what was going on was thinking this was a bug.  

to your question did with shrieker and with command block both did the same. and after doing that thing in config both were working right

e-im:
command block is working as intended shrieker is not. Currently, they just have the DEFAULT reason, not sure that NATURAL quite fits though either as they are somewhat player summoned, not from the natural spawner.

e-im:
What I gave you is a workaround, there is still an issue here. Reopening

Machine-Maker:
I think the solution is to change how the "natural" mobs are tracked. Just having a NATURAL spawn reason shouldn't be good enough. Instead, some other flag on "non-natural" mobs should be set, and only mobs without that flag should be counted.

PR #8296, opened 2022-08-14
Fix warden spawn reason

DrupalDoesNotExists:
https://github.com/PaperMC/Paper/issues/8283


Machine-Maker (changes requested):
I think this should be fixed where the sculk shrieker block entity spawns the warden, not in some util helper method. It looks like there's a method right below it that takes a spawn reason parameter, just change the warden usage to use that one.

Owen1212055:
Why was this closed?

PR #8297, opened 2022-08-14
Tadpole age lock api

Owen1212055:
Smoll separate addition, not putting this in the monster entity api PR as this adds functionality.

In the future, might be better to make a common interface that can be used. We can't use ageable since the adult is a separate entity.

### DIFF
--- a/patches/api/0313-Missing-Entity-Behavior-API.patch
+++ b/patches/api/0313-Missing-Entity-Behavior-API.patch
@@ -896,6 +896,32 @@ index a52a7af219633d575dcbe8ac4b219834bfd4d4d2..1e839b247182af6873a4d74b236d6412
   */
 -public interface Salmon extends Fish { }
 +public interface Salmon extends io.papermc.paper.entity.SchoolableFish { } // Paper - Schooling Fish API
+diff --git a/src/main/java/org/bukkit/entity/Tadpole.java b/src/main/java/org/bukkit/entity/Tadpole.java
+index d64979ebdd018f0f63b6115809b48374ba454249..8e097ad55d208a6980c320e2a849efdcc504cff1 100644
+--- a/src/main/java/org/bukkit/entity/Tadpole.java
++++ b/src/main/java/org/bukkit/entity/Tadpole.java
+@@ -18,4 +18,21 @@ public interface Tadpole extends Fish {
+      * @param age New age
+      */
+     public void setAge(int age);
++
++    // Paper start - Tadpole age lock api
++    /**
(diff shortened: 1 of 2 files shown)


darbyjack (approved):

patches/server/0670-Missing-Entity-Behavior-API.patch
```diff
@@ -24,6 +24,66 @@ index d13f3460644f635ded96bf92ddf9ecf8984c8e47..bd048cc30046f19f9eee89c6ba45d081
...
+-        if (!this.level.isClientSide) {
++        if (!this.level.isClientSide && !this.ageLocked) { // Paper
+             this.setAge(this.age + 1);
+         }
+ 
+@@ -104,12 +105,14 @@ public class Tadpole extends AbstractFish {
+     public void addAdditionalSaveData(CompoundTag nbt) {
+         super.addAdditionalSaveData(nbt);
+         nbt.putInt("Age", this.age);
++        nbt.putBoolean("AgeLocked", this.ageLocked); // Paper
```
NoahvdAa:
```suggestion
+        nbt.putBoolean("Paper.AgeLocked", this.ageLocked); // Paper
```

Owen1212055: The reason why I didn't change this was because it's already implemented under "AgeableMob", do we really want to have "Paper.AgeLocked" and then the "AgeLocked" for other entities?

kennytv (approved):

patches/server/0670-Missing-Entity-Behavior-API.patch
```diff
@@ -24,6 +24,66 @@ index d13f3460644f635ded96bf92ddf9ecf8984c8e47..bd048cc30046f19f9eee89c6ba45d081
...
+         nbttagcompound.putInt("Age", this.getAge());
++        nbttagcompound.putBoolean("AgeLocked", this.ageLocked); // Paper
+     }
+ 
+     @Override
+@@ -171,6 +175,7 @@ public class Tadpole extends AbstractFish {
+             this.setAge(nbt.getInt("Age"));
+         }
+ 
++        this.ageLocked = nbt.contains("AgeLocked") && nbt.getBoolean("AgeLocked"); // Paper
```
Lulu13022002: The getBoolean method will already returns false in case the key is not found so no need to check using the contains method before

Owen1212055:
For this PR, what if an AgeLockable interface was added separately? It's a bit yucky since the ageable interface also has set Adult/Baby but it's a bit meh to add separate methods like this.

PR #8299, opened 2022-08-15
More time entity api and fallingblock extra

Lulu13022002:
Some minimal addition.

### DIFF
--- a/patches/api/0312-Missing-Entity-Behavior-API.patch
+++ b/patches/api/0312-Missing-Entity-Behavior-API.patch
@@ -184,6 +184,81 @@ index 94f3a8c4bf8cf14263d34d866db66728e98dfdb0..7937a0e082199554d3e8db1f9811be29
 +    void setHasBeenStaredAt(boolean hasBeenStaredAt);
 +    // Paper end
  }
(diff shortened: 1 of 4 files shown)


kennytv (changes requested):
Most of these are things that don't really make much sense to have each and every vanilla field exposed since you can already do them with plugins through events or other api methods. While we appreciate every contribution, adding more things for the sake of adding them with no other comment provided makes it hard to evaluate actual use cases

Lulu13022002:
I'm not sure of that
~The set/getDespawnDelay on the llama trader cannot be replaced by a plugin without redo the whole logic each tick and is usefull
when the trader llama is spawned without the trader.~ nvm it can be replaced as well but like the trader villager has theses methods it make senses to also have theses on the llama also when we want a delay greater than vanilla this can be usefull
Same for get/setTime on the falling block due to spigot that override the setTickLived method so handle them separately is convenient
The falling block damage api can be reproduce with plugin, i must admit.

Lulu13022002:
After more thinking all can be reproduce by a plugin except the despawn delay greater than vanilla limit for llama trader (the remove event is not cancellable)
but since this limit is about 40m, i don't see any use, so i will close that.

Issue #8301, opened 2022-08-15
Crash when paper-global.yml and settings directory are read-only


karthanistyr:
### Stack trace

```
[18:01:07 ERROR]: Encountered an unexpected exception
org.spongepowered.configurate.ConfigurateException: []: java.nio.file.AccessDeniedException: /mcconfig/.22351533673970431302813050paper-global.yml.tmp
        at org.spongepowered.configurate.loader.AbstractConfigurationLoader.save(AbstractConfigurationLoader.java:205) ~[configurate-core-4.1.2.jar:?]
        at io.papermc.paper.configuration.Configurations.initializeGlobalConfiguration(Configurations.java:114) ~[paper-1.19.2.jar:git-Paper-130]
        at io.papermc.paper.configuration.Configurations.initializeGlobalConfiguration(Configurations.java:97) ~[paper-1.19.2.jar:git-Paper-130]
        at io.papermc.paper.configuration.PaperConfigurations.initializeGlobalConfiguration(PaperConfigurations.java:180) ~[paper-1.19.2.jar:git-Paper-130]
        at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:214) ~[paper-1.19.2.jar:git-Paper-130]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1125) ~[paper-1.19.2.jar:git-Paper-130]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:305) ~[paper-1.19.2.jar:git-Paper-130]
        at java.lang.Thread.run(Thread.java:833) ~[?:?]
Caused by: java.nio.file.AccessDeniedException: /mcconfig/.22351533673970431302813050paper-global.yml.tmp
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:90) ~[?:?]
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106) ~[?:?]
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111) ~[?:?]
        at sun.nio.fs.UnixCopyFile.copyFile(UnixCopyFile.java:246) ~[?:?]
        at sun.nio.fs.UnixCopyFile.copy(UnixCopyFile.java:603) ~[?:?]
        at sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:258) ~[?:?]
[...]

PR #8306, opened 2022-08-16
Fix: Allow loading if no permission to write configuration

DrupalDoesNotExists:
https://github.com/PaperMC/Paper/issues/8301

### DIFF
--- a/patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
+++ b/patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Fix: Allow loading if no permission to write configuration
 
 
 diff --git a/src/main/java/io/papermc/paper/configuration/Configurations.java b/src/main/java/io/papermc/paper/configuration/Configurations.java
-index 31325994ab441c59a4c0bd9f3f9db3d9440375d0..671fb1fb4d2fbe84717a84c61b79a78d7f211fb3 100644
+index 31325994ab441c59a4c0bd9f3f9db3d9440375d0..5ae2a5271dccc80ef2cde4239ed327abb3222418 100644
 --- a/src/main/java/io/papermc/paper/configuration/Configurations.java
 +++ b/src/main/java/io/papermc/paper/configuration/Configurations.java
 @@ -1,5 +1,6 @@
@@ -51,7 +51,7 @@ index 31325994ab441c59a4c0bd9f3f9db3d9440375d0..671fb1fb4d2fbe84717a84c61b79a78d
 +        } catch (ConfigurateException exception) {
 +            if (exception.getCause() instanceof AccessDeniedException) {
 +                LOGGER.warn("Configuration is read-only. If there any new settings, you need to apply them manually!");
-+            }
++            } else throw exception;
 +        }
          return instance;
      }
--- a/patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
+++ b/patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dmitry Lomov <drupal.doesnot.exists@mail.ru>
+Date: Tue, 16 Aug 2022 11:13:27 +0300
+Subject: [PATCH] Fix: Allow loading if no permission to write configuration
+ file
+
+
+diff --git a/src/main/java/io/papermc/paper/configuration/Configurations.java b/src/main/java/io/papermc/paper/configuration/Configurations.java
+index 31325994ab441c59a4c0bd9f3f9db3d9440375d0..671fb1fb4d2fbe84717a84c61b79a78d7f211fb3 100644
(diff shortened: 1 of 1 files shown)


patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
```diff
@@ -0,0 +1,58 @@
...
+         this.applyGlobalConfigTransformations(node);
+         final G instance = creator.apply(node);
+-        loader.save(node);
++        // Do not crash if configuration is read-only
++        try {
++            loader.save(node);
++        } catch (ConfigurateException exception) {
++            if (exception.getCause() instanceof AccessDeniedException) {
++                LOGGER.warn("Configuration is read-only. If there any new settings, you need to apply them manually!");
++            }
```
molor: Maybe should `throw exception;` again there if that condition is false, to keep the "old behavior" (let's crash if exception is not an instance of AccessDeniedException, or at least print stack trace)?

Astralchroma:
Generally adding imports in patches is discouraged, also this probably is supposed to be done as part of editing the original configuration patch. Also it may be a good idea to handle the exceptions that occur with the other configuration files `server.properties`, `bukkit.yml`, and `spigot.yml`.

Lulu13022002:
Be more explicit on the error message. Mention the name (absolute path) of the file permission issue and also it's not the configuration that is read-only but the file of the configuration.
The files server.properties, bukkit.yml/spigot.yml fail already silently so no need to handle them but you can unify the error message
for example instead of Could not save [NAME]. use your specific message for permission case.
For bukkit/spigot.yml: Check if it's an instanceof SecurityException
For server.properties: Modify the craftbukkit block: "Don't attempt writing to file if it's read only" and instead of just return, log a warning

patches/server/0935-Fix-Allow-loading-if-no-permission-to-write-configur.patch
```diff
@@ -0,0 +1,58 @@
...
+         }
+         this.applyGlobalConfigTransformations(node);
+         final G instance = creator.apply(node);
+-        loader.save(node);
++        // Do not crash if configuration is read-only
++        try {
++            loader.save(node);
++        } catch (ConfigurateException exception) {
++            if (exception.getCause() instanceof AccessDeniedException) {
++                LOGGER.warn("Configuration is read-only. If there any new settings, you need to apply them manually!");
```
karthanistyr, line 53: > If there any new settings, you need to apply them manually!

I think this is a wrong statement. Not all new settings must be added to the config file if they have a default value (which is the case for the vast majority of them), and Paper will work fine without them being in the config file at all!
A better wording would be like "Please remember to review the documentation for the complete list of configuration settings."

Maybe worth mentioning the docs URL https://docs.papermc.io/paper/configuration

karthanistyr, line 53: > If there any new settings, you need to apply them manually!

I think this is an incorrect statement. Not all new settings must be added to the config file if they have a default value (which is the case for the vast majority of them), and Paper will work fine without them being in the config file at all!
A better wording would be like "Please remember to review the documentation for the complete list of configuration settings."

Maybe worth mentioning the docs URL https://docs.papermc.io/paper/configuration

karthanistyr (changes requested):

electronicboy:
1) This should go in the configuration patch rather than being tacked on the end
2) honestly, I'd much rather have config file saving through a flag, which can maybe be applied to the other configs down the line if needed, 99% of people seeing that exception there actually have deeper issues in their environment rather than it being an elective choice; ofc, this is probs something which we'd need comments on

karthanistyr:
A few questions:

* > Generally adding imports in patches is discouraged

@Peter-Crawley It doesn't look like this is possible to avoid, at least the imports for logging, unless there is more extensive refactoring. Any reason why that is, and is it a blocker in this instance?

* > I'd much rather have config file saving through a flag,

@electronicboy Do you agree that this is perhaps outside of the scope of this fix, and could happen later with a well-defined refactoring? The goal here is to avoid the hard crash that was introduced by the migration to Configurate and return to the previous behaviour (printing the stack trace to the console but loading continued fine).

lynxplay:
You avoid imports by fully qualifying your code. E.g.
```java

import foo.Bar;

Bar.something();
```

vs
```java
foo.Bar.something();
```

karthanistyr:
ah yes I know that, I thought what was meant was "no new dependency"

electronicboy:
I don't think that the hard crash from an unexpected illegal state is a bug; Maybe a regression over expectations, but, this imho s more a feature request to me that a "fix", maybe a system property would be better as those tend to fare better in terms of compatibility with different MC environments

It's taken 2 months or so until somebody noticed this as a regression in terms of behavior, the fact that it blew up early on borked file perms honestly kinda seemed like a bit of a blessing

karthanistyr:
I disagree that this is either an illegal state or borked file perms. This is a legitimate system state, and there is no good technical reason why you should enforce writability alone with a crash. The configuration file on a host system is not Paper's property.

The fact that failing to save the config file now crashes is a regression not just on expectations but on design, this requirement (and now behaviour) has never been documented anywhere. It's more of an expectation on your part than on mine at this point.

karthanistyr:
I'm proposing #8318 as a stop-gap solution to ease the crash while a broader design is decided within the core team.

Issue #8300, opened 2022-08-15
Concurrent error no longer occurs on Spigot when command blocks load or unload worlds


willkroboth:
### Is your feature request related to a problem?

#6072 and #8080 raised the issues that command blocks loading/unloading worlds would cause the server to crash. Paper fixed this in #7653 and #8081 by throwing an IllegalStateException if code attempted to load/unload worlds while the worlds were being ticked. 

The same issue was raised in [SPIGOT-7089](https://hub.spigotmc.org/jira/browse/SPIGOT-7089), which was recently resolved. Now, if code attempts to load/unload a world while the worlds are being ticked on a Spigot server it continues as normal.

### Describe the solution you'd like.

I think the Spigot fix here is better. Throwing the IllegalStateException isn't necessary anymore because worlds can now be loaded/unloaded while the worlds are being ticked.

I think all that needs to be done is remove the changes implemented by #7653 and #8081 and make sure craftbukkit is updated (might already be done since https://github.com/PaperMC/Paper/commit/0ddd20c6f7cc545bba5b8345416ee377d5b87be2).

### Describe alternatives you've considered.

The alternative is that the Paper fix is kept and worlds would not be able to be loaded or unloaded while the worlds were being ticked.

electronicboy:
loading I can semi get behind, but, somewhat iffy on due to long term goals to properly support vanilla dimensions stuff which loading worlds in play is already going to become a headache

unloading while ticking sounds like an unmitigated disaster likely to create its own set of stupid issues

willkroboth:
Actually, an alternative I posted on [SPIGOT-7089](https://hub.spigotmc.org/jira/browse/SPIGOT-7089) might work as well:

```
3. Let CraftServer#createWorld and CraftServer#unloadWorld do their thing but delay them from modifying MinecraftServer.levels until the worlds are no longer being ticked.

This one would be the most complicated, and might
look something like this:

// do everything else
if(!worldsAreBeingTicked){
    Modify MinecraftServer.levels now because it is safe
} else {
    Add world to list so it can be loaded/unload later
}
This would allow createWorld and unloadWorld to be called while the worlds are being ticked, but not actually modify the Map of levels until it is safe to do so.
```

[...]

PR #8310, opened 2022-08-17
Modify behavior when worlds are attempted to be loaded or unloaded while being ticked (Addresses #8300)

willkroboth:
This PR changes two things:
- Allow world creation while worlds are being ticked (previously an IllegalStateException would be thrown)
- Delay world unload until worlds are not being ticked (previously an IllegalStateException would be thrown)

This aligns Paper's behavior with Spigot's after [SPIGOT-7089](https://hub.spigotmc.org/jira/browse/SPIGOT-7089) was resolved, which fixed an error that occurred if worlds were loaded or unloaded while being ticked (Previously, Paper had fixed this by throwing IllegalStateExceptions: #7653 and #8081). With this PR, Spigot and Paper agree and allow worlds to be loaded/unloaded while being ticked.

To address the concerns of [this comment](https://github.com/PaperMC/Paper/issues/8300#issuecomment-1218202104), if the worlds are being ticked Paper will not unload a world immediately. Instead, the task that unloads the world will be packaged in a Runnable and stored in a list to be run immediately after the worlds are done being ticked.


willkroboth:
#8300 has been alternatively resolved as: plugins have to deal with the differences between Spigot and Paper.

PR #8324, opened 2022-08-24
Add Entity hidden by default flag

### LINKED ISSUE #7395: Create Entity#hideEntityByDefault

jstnf:
Allows entities to be hidden by default. Entities that are hidden by default are hidden from a Player until Player#showEntity is invoked on that entity. Players that join the server or relog will also not see entities that are hidden by default. The entity's tracker maintains a set of player connections that have been shown the entity.

Entities hidden by default respect other plugins' calls to Player#hideEntity. This means that, if Player#showEntity is called on an entity hidden by default, but the method was not called by the plugin that originally hid the entity, then the entity will still not be shown. If no other plugin hides the player when it is shown, then the entity will remain hidden by default but be shown to that Player.

The functionality is expanded for players that are hidden by default. Hiding players by default will send REMOVE_PLAYER PlayerInfo packets so that they are hidden in a similar fashion to hidePlayer calls. The same can be said when a player is no longer being hidden by default.

Test with my own test plugins, that hide entities and players by default and then show them to specific players. Leaving and rejoining was also tested.

Closes #7395

### DIFF
--- a/patches/server/0936-Add-Entity-hidden-by-default-flag.patch
+++ b/patches/server/0936-Add-Entity-hidden-by-default-flag.patch
@@ -3,25 +3,25 @@ From: Justin <justin@justinf.dev>
 Date: Wed, 24 Aug 2022 05:17:20 -0700
 Subject: [PATCH] Add Entity hidden by default flag
 
-Adds a boolean to the server Entity controlling if the entity should be hidden by default. 
-The TrackedEntity maintains a set of all players that will be able to see the entity even 
-when it is hidden by default. This set is modified when Player#showEntity and Player#hideEntity 
-are invoked. Changes are made to the way that TrackedEntity updates players when the entity is 
+Adds a boolean to the server Entity controlling if the entity should be hidden by default.
+The TrackedEntity maintains a set of all players that will be able to see the entity even
+when it is hidden by default. This set is modified when Player#showEntity and Player#hideEntity
+are invoked. Changes are made to the way that TrackedEntity updates players when the entity is
 hidden by default - if a player is not present in the above set, the update is abandoned.
 
-This functionality is expanded when a Player is hidden by default. The player will send out 
-PlayerInfo packets to hide themselves from other Players when hiddenByDefault. There remains a 
-discrepancy when trying to getHiddenPlayers - players that are hidden by default are currently 
+This functionality is expanded when a Player is hidden by default. The player will send out
+PlayerInfo packets to hide themselves from other Players when hiddenByDefault. There remains a
+discrepancy when trying to getHiddenPlayers - players that are hidden by default are currently
 not returned from this method, only those hidden with hideEntity or hidePlayer.
 
-Hiding entities by default also respects precedent of keeping entities hidden when other plugins 
-explicitly hide them from a player. If an entity is hidden by default and by a plugin, the entity 
-must have the plugin that hid them from a player show them in order to be shown to a player again. 
-If no plugin had previously hidden the entity before the entity was hidden by default, then any 
+Hiding entities by default also respects precedent of keeping entities hidden when other plugins
+explicitly hide them from a player. If an entity is hidden by default and by a plugin, the entity
+must have the plugin that hid them from a player show them in order to be shown to a player again.
+If no plugin had previously hidden the entity before the entity was hidden by default, then any
 plugin that attempts to show the entity to a player will succeed in doing so.
 
 diff --git a/src/main/java/net/minecraft/server/level/ChunkMap.java b/src/main/java/net/minecraft/server/level/ChunkMap.java
-index 77c89376495d90d0e7cbf6cd02c9a1c8d9a4340b..c29c2a57b8bab3dd36af0e4ff21c6f0735c260a9 100644
+index 77c89376495d90d0e7cbf6cd02c9a1c8d9a4340b..4e506395085fc597a0544830355d4490194706f0 100644
 --- a/src/main/java/net/minecraft/server/level/ChunkMap.java
 +++ b/src/main/java/net/minecraft/server/level/ChunkMap.java
 @@ -2277,6 +2277,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
@@ -32,29 +32,23 @@ index 77c89376495d90d0e7cbf6cd02c9a1c8d9a4340b..c29c2a57b8bab3dd36af0e4ff21c6f07
  
          public TrackedEntity(Entity entity, int i, int j, boolean flag) {
              this.serverEntity = new ServerEntity(ChunkMap.this.level, entity, j, flag, this::broadcast, this.seenBy); // CraftBukkit
-@@ -2385,6 +2386,13 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+@@ -2385,6 +2386,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
                      flag = false;
                  }
                  // CraftBukkit end
-+
-+                // Paper start - Entity hiddenByDefault
-+                if (this.entity.isHiddenByDefault() && !this.showToEvenWhenHiddenByDefault.contains(player.connection)) {
-+                    flag = false;
-+                }
-+                // Paper end
 +
                  if (flag) {
                      if (this.seenBy.add(player.connection)) {
                          this.serverEntity.addPairing(player);
 diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
-index f925a8d550ecbf2044a37bfe58b30d6578c5f6af..dfda6e2f38be417f19dc9d7847f3cae80fb8fac2 100644
+index f925a8d550ecbf2044a37bfe58b30d6578c5f6af..20b2656086ba8e2e517ef0e2a6e417c068e22bc6 100644
 --- a/src/main/java/net/minecraft/world/entity/Entity.java
 +++ b/src/main/java/net/minecraft/world/entity/Entity.java
 @@ -398,6 +398,7 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
      @javax.annotation.Nullable
      private UUID originWorld;
      public boolean freezeLocked = false; // Paper - Freeze Tick Lock API
-+    public boolean hiddenByDefault; // Paper - Entity hiddenByDefault
++    private boolean hiddenByDefault; // Paper - Entity hiddenByDefault
  
      public void setOrigin(@javax.annotation.Nonnull Location location) {
          this.origin = location.toVector();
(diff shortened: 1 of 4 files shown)


patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
++        public final Set<ServerPlayerConnection> showToEvenWhenHiddenByDefault = new ReferenceOpenHashSet<>(); // Paper - Entity hiddenByDefault
+ 
+         public TrackedEntity(Entity entity, int i, int j, boolean flag) {
+             this.serverEntity = new ServerEntity(ChunkMap.this.level, entity, j, flag, this::broadcast, this.seenBy); // CraftBukkit
+@@ -2385,6 +2386,13 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+                     flag = false;
+                 }
+                 // CraftBukkit end
++
++                // Paper start - Entity hiddenByDefault
```
lynxplay: Are we not doubling logic here?
https://github.com/jstnf/Paper/blob/5e3a6cc38bf223eaf36657f7f70b364745ed7f4c/patches/server/0936-Add-Entity-hidden-by-default-flag.patch#L170-L174

already checks if a player is respecting the hidden by default state / show overwrite for hidden by default.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
+                         this.serverEntity.addPairing(player);
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index f925a8d550ecbf2044a37bfe58b30d6578c5f6af..dfda6e2f38be417f19dc9d7847f3cae80fb8fac2 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -398,6 +398,7 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
+     @javax.annotation.Nullable
+     private UUID originWorld;
+     public boolean freezeLocked = false; // Paper - Freeze Tick Lock API
++    public boolean hiddenByDefault; // Paper - Entity hiddenByDefault
```
lynxplay: Why is this public if we have setters and getters ?
One of the two should be enough, usually its a public field mutated through the craft bukkit layer directly just for a smaller NMS diff.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
+     // Paper End - Collision API
++
++    // Paper start - Entity hiddenByDefault
++    @Override
++    public void setHiddenByDefault(boolean hiddenByDefault) {
++        this.getHandle().setHiddenByDefault(hiddenByDefault);
++
++        // We need to update the players again
++        ServerLevel world = ((CraftWorld) this.getWorld()).getHandle();
++        ChunkMap.TrackedEntity entityTracker = world.getChunkSource().chunkMap.entityMap.get(this.getEntityId());
```
lynxplay: This can be fast-tracked using `getHandle().tracker`, saves you the world casting too.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
++            hidingPlugins.remove(CraftPlayer.getPluginWeakReference(plugin));
++            if (!hidingPlugins.isEmpty()) {
++                return; // Some other plugins still want the entity hidden
++            }
++        } else {
++            if (!entity.isHiddenByDefault()) {
++                return; // Entity isn't hidden
++            }
+         }
++        // Paper end
```
lynxplay: Technically speaking here, we should also not re-register the shown entity if the player is already part of the default hidden overwrite set. While it does not make a difference in testing, as the client just discards a new ClientboundPlayerInfoPacket with the exact same profile, the method should not be called if the player can already see the entity.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
+     }
+     // Paper start
+     private void reregisterPlayer(ServerPlayer player) {
+@@ -1843,7 +1855,17 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+ 
+     @Override
+     public boolean canSee(org.bukkit.entity.Entity entity) {
+-        return !this.hiddenEntities.containsKey(entity.getUniqueId());
++        // Paper start - Entity hiddenByDefault
++        boolean ret = true;
```
lynxplay: ret ? might be nice to pick a better variable name here.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
++
++        if (entity.isHiddenByDefault()) {
++            ServerLevel world = ((CraftWorld) entity.getWorld()).getHandle();
++            ChunkMap.TrackedEntity entityTracker = world.getChunkSource().chunkMap.entityMap.get(entity.getEntityId());
```
lynxplay: Can again be shortcutted using `getHandle().tracker`

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
+     // Paper end
+ 
++    // Paper start - Entity hiddenByDefault
++    @Override
++    public void setHiddenByDefault(boolean hiddenByDefault) {
++        // Update server lists
++        if (hiddenByDefault) {
++            server.getHandle().getPlayers().stream()
++                .filter(serverPlayer -> serverPlayer != this.getHandle())
++                .forEach(serverPlayer -> serverPlayer.connection.send(new ClientboundPlayerInfoPacket(ClientboundPlayerInfoPacket.Action.REMOVE_PLAYER, this.getHandle())));
```
lynxplay: Any reason we don't check if the player can see the entity in the first place here ?

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,207 @@
...
++        } else {
++            server.getOnlinePlayers().stream()
++                .filter(entityPlayer -> entityPlayer != this)
++                .filter(entityPlayer -> !entityPlayer.hiddenEntities.containsKey(this.getUniqueId()))
++                .forEach(entityPlayer -> entityPlayer.getHandle().connection.send(new ClientboundPlayerInfoPacket(ClientboundPlayerInfoPacket.Action.ADD_PLAYER, this.getHandle())));
```
lynxplay: This is also technically sent to players that can already see the entity due to a `showEntity` cakk while the entity was hidden by default.

lynxplay (changes requested):

Thanks for your first PR, looks great!
A few minor comments on the PR, most of them related to some missing checks that imo would be useful.

Beyond that, I did not have the time to test this properly yet, mostly just looking over the patches.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Good catch, put this check in before updating canSee.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: I agree. Think the thought process here was looking at the surrounding variables. Will change to private.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Thanks!

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Absolutely

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Thanks ^^

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: The assumption is that, since we just called setHiddenByDefault, that any player will not be able to see the Entity since it has not been shown yet.

jstnf: But now that you mention it, would be good if the entity is already hidden. Will add a check.

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Ah! So if showPlayer is called twice on an entity hidden by default. Good catch!

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
jstnf: Looks like canSee is the catch all then. Will use this for the check instead.

Lulu13022002:
I know that the logic is not the same as https://github.com/PaperMC/Paper/pull/7666 but it close the same issue, so you should probably find a solution together

jstnf:
> I know that the logic is not the same as #7666 but it close the same issue, so you should probably find a solution together

How should I go about doing this?

patches/server/0936-Add-Entity-hidden-by-default-flag.patch
```diff
@@ -0,0 +1,200 @@
...
++        if (hiddenByDefault) {
++            server.getOnlinePlayers().stream()
++                .filter(entityPlayer -> entityPlayer != this)
++                .filter(entityPlayer -> entityPlayer.canSee(this))
++                .forEach(entityPlayer -> entityPlayer.getHandle().connection.send(new ClientboundPlayerInfoPacket(ClientboundPlayerInfoPacket.Action.REMOVE_PLAYER, this.getHandle())));
++        } else {
++            server.getOnlinePlayers().stream()
++                .filter(entityPlayer -> entityPlayer != this)
++                .filter(entityPlayer -> !entityPlayer.canSee(this))
++                .forEach(entityPlayer -> entityPlayer.getHandle().connection.send(new ClientboundPlayerInfoPacket(ClientboundPlayerInfoPacket.Action.ADD_PLAYER, this.getHandle())));
```
Owen1212055: Streams are cringe and cause unneeded overhead, just use a for loop for this.

jstnf: Done.

jstnf:
Rebased and removed the streams.

jstnf (commented):
Seems the request for changes came back after I rebased the commits. Not sure what to do. @lynxplay

molor:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/e8010dc93fbb163876d6c745970dd2d9b587b381?

Machine-Maker:
Done by upstream https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/fb50924403915ad24e9389361a10dfbf53c1e696

PR #8326, opened 2022-08-24
Stop large look changes from crashing the server

Jaren8r:
Supersedes #7459, with a proper diff this time

### DIFF
--- a/patches/server/0936-Stop-large-look-changes-from-crashing-the-server.patch
+++ b/patches/server/0936-Stop-large-look-changes-from-crashing-the-server.patch
@@ -1,8 +1,9 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Jaren Knodel <Jaren@Knodel.com>
+From: MWHunter <s0521458@student.rockvalleycollege.edu>
 Date: Wed, 24 Aug 2022 09:54:11 -0400
 Subject: [PATCH] Stop large look changes from crashing the server
 
+Co-authored-by: Jaren Knodel <Jaren@Knodel.com>
 
 diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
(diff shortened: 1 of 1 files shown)


Lulu13022002:
This can also happens for all projectile see the Projectile#lerpRotation. But since this is just faster/simpler than looping over a large value, you should search in the code all bad pattern of this.

Jaren8r:
That's the only other use of that pattern I could find in the server code, so I went ahead and changed that too.

Lulu13022002:
Hmm what about ProjectileUtil#rotateTowardsMovement (will probably not crash) and Mth#rotlerp then? I can't test if these crash because of the two other loop that run in the tick directly. Also the fix provided for the projectile is wrong, you forgot to increment instead of replace
basically: prevRot **+**= Math.round((newRot - prevRot) / 360.0F) * 360.0F;

Jaren8r:
> Also the fix provided for the projectile is wrong, you forgot to increment instead of replace

Blah, good catch. It's fixed now.

patches/server/0936-Stop-large-look-changes-from-crashing-the-server.patch
```diff
@@ -0,0 +1,73 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jaren Knodel <Jaren@Knodel.com>
```
kennytv: Put the original patch author here

kennytv (changes requested):

patches/server/0936-Stop-large-look-changes-from-crashing-the-server.patch
Jaren8r: Done

kennytv (approved):

asquared31415:
Can this get backported to all supported versions? It affects 1.18.2 and possibly other versions as well.

electronicboy:
1.18 is EOL and has far exceeded its support lifecycle, if somebody wants to open a PR, maybe; but, I can't justify throwing any time towards it

PR #8333, opened 2022-08-28
Boat + ChestBoat constructors fix

IExploitableMan:


olijeffers0n:
Can I just ask, What does this fix?

Lulu13022002:
I'm pretty sure that this can't even compile since you have deleted the constructor (EntityType<? extends Boat> type, Level world)
which is needed for the generic EntityType class (for the summon command for example). Also please explain more about that. If the position is not set here it's bc it's set later in the summon command or other place where it's called.

Owen1212055:
If you need a change like this it is recommended that you make a fork of paper here. These types of changes are not supported and don’t really have any benefit to have on paper.

Since you haven’t put any description what this does either, it’s hard to tell what you even want here.

PR #8334, opened 2022-08-28
Only create paper.yml-README.txt when config conversion happens

### LINKED ISSUE #8332: Add a method to prevent the creation of the paper.yml-README.txt file.

Currently, this file is created in your server's files even if you have fully acknowledged the location of the new configuration files. This is really annoying, especially if you are looking to keep your server files clean.

Possibly allow users to add a Java flag to their servers to prevent the creation of this file, or, even better, only generate the `paper.yml-README.txt` file once and then once it is deleted by the user, do not re-generate it anymore.

I have tried opening a feature request on Purpur's GitHub repository about this, but ended up realizing that this should be brought up to Paper directly.

### Other

_No response_

TetraTheta:
Currently, Paper always creates `paper.yml-README.txt` if there is no such file in main directory.  

With this PR, Paper will only generate `paper.yml-README.txt` file config file conversion happens.

If there are any other situation I should check too, let me know.

This PR may resolve #8332 (forgot to mention, thanks for remind me!)

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -708,7 +708,7 @@ index 0000000000000000000000000000000000000000..84785fed0d85d78c4caf8fabe35c0e89
 +        public boolean lagCompensateBlockBreaking = true;
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
-+        public boolean doNotCreateReadme = true;
++        public boolean doNotCreateReadme = false;
 +    }
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/InnerClassFieldDiscoverer.java b/src/main/java/io/papermc/paper/configuration/InnerClassFieldDiscoverer.java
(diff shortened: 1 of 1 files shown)


Machine-Maker:
> But the method I used is very ugly. Since GlobalConfiguration.get().misc was null, I had to use this way. If there is any better way to get the value, please tell me.

Yeah I'm not sure there is really. Unless you moved the whole readme file creation logic until after the global configuration was initialized. I guess you could technically do that, move it to PaperConfigurations#initializeGlobalConfig, but your way is probably cleaner than that.

Astralchroma:
Wouldn't it be simpler to only create the file if the newer config files don't exist already?

TetraTheta:
Updated PR to reflect suggestion. I think server admin should manually check config about README file generation, but this new method is cleaner way to check things. No manual value check with plain string key.

kennytv (changes requested):
Now you're creating the info file only when the new config already exists - just move the creation into the `needsConverting(legacyConfig)` block, since it's only relevant to people who upgrade from our old configs

electronicboy:
imho, it literally just needs to create the file when creating a new config folder, nothing more, nothing less; There is literally 0 reasons to have a config for this for any reason, the literal sole intent is to just say "we moved this"

I don't think it's only relevant to those upgrading from new configs, as we see the odd person who installed a new server and is still referring to old documentation from around the web, etc

patches/server/0004-Paper-config-files.patch
```diff
@@ -1233,6 +1229,8 @@ index 0000000000000000000000000000000000000000..b2e961bbd33c6ecb7f049365b7aff6c5
 +                if (Files.isSymbolicLink(legacyConfig)) {
 +                    Files.delete(legacyConfig);
 +                }
++                Files.createFile(replacementFile);
++                Files.writeString(replacementFile, String.format(MOVED_NOTICE, configDir.toAbsolutePath()));
```
kennytv: Needs the notExists check

TetraTheta: I'm not sure if I did this right, but added notExists check.

kennytv (approved):

PR #8335, opened 2022-08-28
Also load resources from LibraryLoader

Goldmensch:
Currently resources are only loaded from the plugin classloader, which causes issues with some libraries, because they can't find their resources. This pr adds support for loading them from the `LibraryLoader` of the plugin.


Goldmensch:
~~There's currently a missing null check. I'll fix it tomorrow.~~

patches/api/0392-Also-load-resources-from-LibraryLoader.patch
```diff
@@ -0,0 +1,52 @@
...
+
+diff --git a/src/main/java/org/bukkit/plugin/java/PluginClassLoader.java b/src/main/java/org/bukkit/plugin/java/PluginClassLoader.java
+index 345394132df70593800127d34a38f8f8a4dafe00..d09811ba3aedadc20f8b37883fb5a02d81177284 100644
+--- a/src/main/java/org/bukkit/plugin/java/PluginClassLoader.java
++++ b/src/main/java/org/bukkit/plugin/java/PluginClassLoader.java
+@@ -10,6 +10,7 @@ import java.net.URL;
+ import java.net.URLClassLoader;
+ import java.security.CodeSigner;
+ import java.security.CodeSource;
++import java.util.ArrayList;
```
kennytv: Inline the import

patches/api/0392-Also-load-resources-from-LibraryLoader.patch
```diff
@@ -0,0 +1,52 @@
...
+ import java.util.Collection;
+ import java.util.Collections;
+ import java.util.Enumeration;
+@@ -90,12 +91,31 @@ public final class PluginClassLoader extends URLClassLoader { // Spigot
+ 
+     @Override
+     public URL getResource(String name) {
+-        return findResource(name);
++        // Paper start
++        var resource = findResource(name);
```
kennytv: No `var` please

Lulu13022002: The var is still here, there was two var in fact

electronicboy (approved):

patches/api/0392-Also-load-resources-from-LibraryLoader.patch
```diff
@@ -0,0 +1,44 @@
...
++        java.util.ArrayList<URL> resources = new java.util.ArrayList<>();
++        addEnumeration(resources, findResources(name));
++        if (libraryLoader != null) {
++            addEnumeration(resources, libraryLoader.getResources(name));
++        }
++        return Collections.enumeration(resources);
++        // Paper end
++    }
++
++    private <T> void addEnumeration(java.util.ArrayList<T> list, Enumeration<T> enumeration) {
```
lynxplay, line 39: This method should be enclosed in paper comments.

lynxplay (changes requested):

PR #8338, opened 2022-09-01
Re-arrange most chunk system patches to front

Spottedleaf:
Moved most chunk system hooks to a single class, so that
later patches are not hopefully reliant on specific
patches even exist.
PlayerChunkLoader patch was not re-organized due to how awful it was, that will have to be done via the chunk system merge.


electronicboy (approved):

Spottedleaf:
looks fine to me

JRoy (approved):

PR #8343, opened 2022-09-02
Allow preventing BlockDestroyEvent from dropping items

okx-code:
Useful for overriding drops from blocks such as bamboo, which fire this event when their base is broken.


kennytv:
Changing the boolean here does nothing, the value is not used in any of the actual drop logic

kennytv (approved):

kennytv:
... Although we do not accept any more feature prs to the 1.18.2 branch. You'd have to change the base branch and update the pr onto master.

okx-code:
See #8349

PR #8344, opened 2022-09-03
Supports translatable component with custom bundle for itemstack (lore/name)

Lulu13022002:
When a plugin register a custom bundle of translation. This bundle is ignored on lore/displayname of an itemstack and only the vanilla keys works. This is a problem because all logic must be duplicated just for text. Sadly this was harder than i thought and i have made a little workaround to stop the client that mutate the lore/name with the translated value. This isn't a problem with the actual really strict item meta api that just strip all bad values.


electronicboy:
Not entirely fond of something like this, just, mucho hacko

Lulu13022002:
I'm closing this since this can be done using a resource pack and i don't want to mangle the data send to the client like that.

PR #8354, opened 2022-09-08
EndDragonFight killed statuses should be false for newly created worlds

### LINKED ISSUE #7750: EnderDragonBattle#hasBeenPreviouslyKilled() returns true for a new created world

elmital:
Superseeds #7752 same fix but on a separate branch
Fixes #7750


YanisBft (approved):

patches/server/0937-Fix-EndDragonFight-killed-statuses-should-be-false-f.patch
```diff
@@ -0,0 +1,35 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+index f8d846345c1cc3c78f9ac14635b26f2affc77190..d380779f0fcdd00342b16b12f9d5e1aadb90fc46 100644
+--- a/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
++++ b/src/main/java/net/minecraft/world/level/dimension/end/EndDragonFight.java
+@@ -91,7 +91,7 @@ public class EndDragonFight {
+     public EndDragonFight(ServerLevel world, long gatewaysSeed, CompoundTag nbt) {
+         // Paper start
+         this.needsStateScanning = world.paperConfig().entities.spawning.scanForLegacyEnderDragon;
+-        if (!this.needsStateScanning) this.dragonKilled = true;
++        this.dragonKilled = !this.needsStateScanning;
```
Lulu13022002: I think this line isn't usefull since this field is already false at init stage. So the old line was enough.

elmital: You right! I don't even remember why I've made this change. I reverted it in the last commit and rebase the PR in the same time.

electronicboy (approved):
Not fond of the formatting changes, and somewhat weary that this is just hiding a deeper issue inside of upstream world management stuff, but, I guess dealing with that is down the line as another thing

PR #8360, opened 2022-09-14
Limit size of Authenticator Cache Thread Pool

Astralchroma:
@MiniDigger 
Screenshot from 2022-09-14 11-48-06

### DIFF
--- a/patches/server/0116-Cache-user-authenticator-threads.patch
+++ b/patches/server/0116-Cache-user-authenticator-threads.patch
@@ -5,29 +5,19 @@ Subject: [PATCH] Cache user authenticator threads
 
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
-index bcf189d0ae917b99fff62167740ddb0012082138..6e315a2e87bbc4b03e4e0f38ba0b0f6b592a433e 100644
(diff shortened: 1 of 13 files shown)


jpenilla:
the thread factory should not be removed

patches/server/0116-Cache-user-authenticator-threads.patch
```diff
@@ -5,29 +5,19 @@ Subject: [PATCH] Cache user authenticator threads
...
+@@ -118,6 +118,8 @@ public class ServerLoginPacketListenerImpl implements TickablePacketListener, Se
  
      }
  
-+    // Paper start - Cache authenticator threads
-+    private static final AtomicInteger threadId = new AtomicInteger(0);
-+    private static final java.util.concurrent.ExecutorService authenticatorPool = java.util.concurrent.Executors.newCachedThreadPool(
-+            r -> {
-+                Thread ret = new Thread(r, "User Authenticator #" + threadId.incrementAndGet());
++    private static final java.util.concurrent.ExecutorService authenticatorPool = new java.util.concurrent.ThreadPoolExecutor(0, 16, 60L, java.util.concurrent.TimeUnit.SECONDS, new java.util.concurrent.SynchronousQueue<>()); // Paper - Cache authenticator threads
```
electronicboy: as said on the comments, the thread factory needs to stay that we we can get thrown exceptions, and the name of the threads correct

electronicboy (changes requested):

electronicboy (approved):

PR #8367, opened 2022-09-18
Support enforce-secure-profile when behind proxy.

Gardelll:
This removes the "[Not Secure]" label when running Paper server behind a proxy like Velocity.


electronicboy:
My understanding is that this was not done because it creates many other dozen issues in the process

Gardelll:
> My understanding is that this was not done because it creates many other dozen issues in the process

Can you give some examples? I will try to solve them.

electronicboy:
Closing; This really need to be followed by a wider set of changes in between the client and the server to ensure that this remains working in various cases, i.e. server switching. There is already efforts on the velocity team and people surrounding that project working on this for velocity, which will potentially be ported over to waterfall, but, as is, we cannot support re-enabling this thing as-is in these current setups.

Issue #8118, opened 2022-07-10
markers dont update position while riding other entities


Strongis13:
### Expected behavior

When summoning a moving entity with a marker as a passenger, the marker should constantly update to the location of the entity it is riding

### Observed/Actual behavior

Passenger markers do not update their position while riding other entities, causing them to stay in the position they were spawned at

### Steps/models to reproduce

In vanilla: https://youtu.be/9YN98lzqCoM
commands used:
-tag @e[type=minecraft:snowball,tag=!modsnowball] add snowball1
-execute as @e[type=snowball,tag=snowball1] at @s run summon snowball ~ ~ ~ {Tags:["modsnowball"],Passengers:[{id:"minecraft:marker"}]}
-data modify entity @e[tag=modsnowball,limit=1] Motion set from entity @e[tag=snowball1,limit=1] Motion
-kill @e[type=snowball,tag=snowball1,limit=1]
-execute at @e[type=marker] run particle glow ~ ~ ~ 0 0 0 0 10 force

In paper 1.19: https://youtu.be/bUXehZ7mh-Q
commands used:
-same as vanilla

### Plugin and Datapack List

There are no plugins or custom datapacks

### Paper version

> version
[21:13:55 INFO]: Checking version, please wait...
[21:13:56 INFO]: This server is running Paper version git-Paper-58 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: e294802)
You are running the latest version
Previous version: git-Paper-39 (MC: 1.19)

### Other

Adding a config option to disable #7299 may fix this issue

lynxplay:
Paper actively removes the marker entity from the ticking list as an optimisation.
This either might be worth to add a configuration option for or just straight up remove said optimisation.
Really up to whoever opens the PR and the reviewers 👍

piratecody:
Currently working on a fix to add a config option to enable/disable this optimization

Owen1212055:
> Currently working on a fix to add a config option to enable/disable this optimization

I assume you are not working on this anymore? Is someone else able to take over?

piratecody:
> > Currently working on a fix to add a config option to enable/disable this optimization
> 
> I assume you are not working on this anymore? Is someone else able to take over?

I have the fix written, but I've been busy and haven't had a chance to commit & PR. Should have some time this weekend.

piratecody:
PR in progress 
https://github.com/PaperMC/Paper/pull/8372

[...]

PR #8372, opened 2022-09-18
Adding config to allow marker ticking (#8118)

piratecody:
This adds a world config option to allow markers to be ticked, rather than just having them not ticked at all. Defaults to false so the existing functionality is the same for those that don't need markers to tick.

### DIFF
--- a/patches/api/0003-Test-changes.patch
+++ b/patches/api/0003-Test-changes.patch
@@ -288,12 +288,12 @@ index 4ac3dd977e75cd8464163351d306e037ee32cb48..c26ea217927ba77611e6ae93f8df50a8
 +                return true;
 +            }
 +        }
- 
--    private static boolean isWellAnnotated(@Nullable List<AnnotationNode> annotations) {
++
 +        return false;
 +    }
 +    // Paper end
-+
+ 
+-    private static boolean isWellAnnotated(@Nullable List<AnnotationNode> annotations) {
 +    private static boolean isWellAnnotated(@Nullable List<? extends AnnotationNode> annotations) { // Paper
          if (annotations == null) {
              return false;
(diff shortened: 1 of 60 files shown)


Owen1212055:
I would edit the patch that adds this functionality instead of making a new patch.

lynxplay:
Hi :wave: congrats on your first paper PR!

As owen already pointed out, to reduce patches in general, the logic should live in the patch that originally introduced the marker-no-tick (https://github.com/PaperMC/Paper/blob/master/patches/server/0891-Don-t-tick-markers.patch).

Beyond that, changes the any of the configuration files belong in the https://github.com/PaperMC/Paper/blob/master/patches/server/0004-Paper-config-files.patch patch.
If you have any specific questions on how to achieve this, feel free to ask in our #paper-dev channel on the paper discord :)

patches/server/0943-Adding-config-to-allow-marker-ticking.patch
```diff
@@ -0,0 +1,61 @@
...
+diff --git a/src/main/java/org/spigotmc/ActivationRange.java b/src/main/java/org/spigotmc/ActivationRange.java
+index 40b382c2e0e33fe5c24a51b211cd2f9557a60c5e..fb8bed037809040003a12553ac704a0483bfb232 100644
+--- a/src/main/java/org/spigotmc/ActivationRange.java
++++ b/src/main/java/org/spigotmc/ActivationRange.java
+@@ -212,7 +212,7 @@ public class ActivationRange
+             // Paper end
+ 
+             // Paper start
+-            java.util.List<Entity> entities = world.getEntities((Entity)null, maxBB, (e) -> !(e instanceof net.minecraft.world.entity.Marker)); // Don't tick markers
++            java.util.List<Entity> entities = world.getEntities((Entity)null, maxBB, (e) -> (e instanceof net.minecraft.world.entity.Marker && world.paperConfig().entities.tickMarkers)); // Don't tick markers by default
```
Lulu13022002: This will actually just allow marker to be activated or not but all others entities will be ignored. And the old behavior activate all entities except markers
should probably be:
!(e instanceof net.minecraft.world.entity.Marker) || tickMarkers ?

Machine-Maker (changes requested):
See @Lulu13022002's review

PR #8373, opened 2022-09-19
Avoid item meta usage for itemstack enchantment getter

### LINKED ISSUE #5188: Performance regression of ItemStack#getEnchantmentLevel

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/5188
Redo of my old PR without all the mess (before part II)

"Paper has fixed enchantment order inconsistencies but has also implemented the fix on getter which is useless because no write operation is done here."
This wouldn't be a problem if the item meta was fast
You should use theses methods over the meta one if you care about performance and doesn't do bulk operation on item.

### DIFF
--- a/patches/server/0073-Handle-Item-Meta-Inconsistencies.patch
+++ b/patches/server/0073-Handle-Item-Meta-Inconsistencies.patch
@@ -70,10 +70,18 @@ index e937186aaf819a77c80beeb9e08413a1f781c13a..0e19f49ca2496b1c42d27289bcea15d2
  
      public boolean isEnchanted() {
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-index 4b79b96579efbc4dd9a10e7183ed08b73b488794..7789f3ad01df1c6fc109592ba107852c7539c982 100644
+index 4b79b96579efbc4dd9a10e7183ed08b73b488794..c9093275d2b78b6e2f59e64efab0b4775ff0b43b 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-@@ -178,28 +178,11 @@ public final class CraftItemStack extends ItemStack {
+@@ -6,7 +6,6 @@ import java.util.Map;
+ import net.minecraft.nbt.CompoundTag;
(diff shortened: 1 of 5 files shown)


patches/server/0073-Handle-Item-Meta-Inconsistencies.patch
```diff
@@ -70,18 +70,10 @@ index e937186aaf819a77c80beeb9e08413a1f781c13a..0e19f49ca2496b1c42d27289bcea15d2
...
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-index 4b79b96579efbc4dd9a10e7183ed08b73b488794..3745033afb8923ce06cbf13b55c4e96f2a89573f 100644
+index 4b79b96579efbc4dd9a10e7183ed08b73b488794..7789f3ad01df1c6fc109592ba107852c7539c982 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-@@ -6,7 +6,6 @@ import java.util.Map;
- import net.minecraft.nbt.CompoundTag;
- import net.minecraft.nbt.ListTag;
- import net.minecraft.world.item.Item;
--import net.minecraft.world.item.enchantment.EnchantmentHelper;
```
electronicboy: pointless diff

electronicboy (approved):

PR #8377, opened 2022-09-21
Improve performance of ThreadedLevelLightEngine

bluelhf:
This PR adds small changes that improve the performance of the threaded level light engine and Starlight interface.

||Old|New|Change
|---|---|---|---
|Memory allocations|The memory allocation call tree for the previous implementation|The memory allocation call tree for the new implementation|**-23.1&nbsp;%**
|CPU samples|The CPU sample call tree for the previous implementation|The CPU sample call tree for the new implementation|**-35.4&nbsp;%**

<sup>Lower is better in both allocated bytes and CPU samples.</sup>

This is done by swapping the CompletableFuture `thenCombine(...)` with a more efficient `thenCompose(...)` (we don't care about the result of the other CompletableFuture, as it does not have one). Additionally, a `HashSet` field in `StarlightInterface.LightQueue.ChunkTasks` was converted to an `ObjectOpenHashSet`. I did not measure the performance impact of this alone, but I would be surprised if the HashMap-backed java.util HashSet was faster than a 'real' implementation.

### DIFF
--- a/patches/server/0788-Rewrite-the-light-engine.patch
+++ b/patches/server/0788-Rewrite-the-light-engine.patch
@@ -3071,15 +3071,16 @@ index 0000000000000000000000000000000000000000..1b0d92c68407cdb09ed8aac271b625d9
 +}
 diff --git a/src/main/java/ca/spottedleaf/starlight/common/light/StarLightInterface.java b/src/main/java/ca/spottedleaf/starlight/common/light/StarLightInterface.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..ef8dcbb6bbc0769e9ccfdadb05e6a46c070eda98
+index 0000000000000000000000000000000000000000..be4b62c0facfd7bf20d1e5699c26a9078610069c
 --- /dev/null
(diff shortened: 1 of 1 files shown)


NoahvdAa (changes requested):
👋 Congrats on making your first PR, welcome to Paper!

This change should modify the existing [starlight patch](https://github.com/PaperMC/Paper/blob/9e614e6f9a1b625ba550a147266a474400c013c6/patches/server/0788-Rewrite-the-light-engine.patch) instead of creating a new patch, you can just fixup your current patch into it. The contributing guide has instructions on it in case you're unsure how to do it.

Since you're modifying Starlight itself, this PR should also be opened to the [Starlight repo](https://github.com/PaperMC/Starlight/), to prevent the changes getting lost in a future update.

bluelhf:
I hope I did it right...

Spottedleaf:
When https://github.com/PaperMC/Starlight/pull/160 is merged I will port the diff under your commit details to master.

bluelhf:
> When https://github.com/PaperMC/Starlight/pull/160 is merged I will port the diff under your commit details to master.

thank you this makes my life ten times easier i do not understand the patch system in the slightest

PR #8399, opened 2022-09-28
Preliminary example of WorldUnloadEvent

electronicboy:
needs organisating, commentry, and actually exposing in the API

PR is semi hasty, I don't wanna throw a bunch of time cleaning this up without some general commentary first, iffy on not using an enum, being enums have bit us in the arse before, and mostly, people mostly only care about fail/success

Should probably add the annotation to not ignore the ret value too, as many people seem to miss that one in general

### DIFF
--- a/patches/api/0398-Add-WorldUnloadResult.patch
+++ b/patches/api/0398-Add-WorldUnloadResult.patch
@@ -1,0 +1,42 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Shane Freeder <theboyetronic@gmail.com>
+Date: Wed, 28 Sep 2022 05:09:05 +0100
+Subject: [PATCH] Add WorldUnloadResult
+
+
+diff --git a/src/main/java/io/papermc/paper/world/WorldUnloadResult.java b/src/main/java/io/papermc/paper/world/WorldUnloadResult.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..5fb077e0ab5ad238763adb2a1d2a6cc95cf0f1e4
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/world/WorldUnloadResult.java
+@@ -0,0 +1,30 @@
++package io.papermc.paper.world;
++
++import net.kyori.adventure.text.Component;
++
++public class WorldUnloadResult {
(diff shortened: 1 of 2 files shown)


patches/api/0398-Add-WorldUnloadResult.patch
```diff
@@ -0,0 +1,29 @@
...
++
++import org.jetbrains.annotations.NotNull;
++
++public record WorldUnloadResult(boolean success, @NotNull Component message) {
++
++    public static final WorldUnloadResult SUCCESS = new WorldUnloadResult(true, Component.text("Success"));
++    public static final WorldUnloadResult NOT_LOADED = new WorldUnloadResult(false, Component.text("Cannot unload unloaded world"));
++    public static final WorldUnloadResult UNSUPPORTED = new WorldUnloadResult(false, Component.text("Unloading this world is not supported"));
++    public static final WorldUnloadResult NOT_EMPTY = new WorldUnloadResult(false, Component.text("Unable to load world with players!"));
++    public static final WorldUnloadResult PENDING_LOGIN = new WorldUnloadResult(false, Component.text("Unable to load world with pending login!"));
```
tomcraft, line 25: I think you have a typo on "load" (you meant "unload" right ?)

electronicboy: Yea, I was honestly just rushing this to get it down, been meaning to look into this one for a good while, general improvements to the messages welcome, etc

MiniDigger:
I assume the title is a typo and you dont actually want an event to check the result of the api call?

I generally like this, not super sure the component in the result is required, but I am sure some plugin devs will appreciate beging able to forward those messages to users instead of having to come up with their own.

electronicboy:
Yea , title is a brainfart; ideally I'd probs make those just translatable components or whatever, if we wanted to get rid of those I'd probs just turn it into an enum for better or worse, I guess it doesn't exactly matter here, not something likely to be extended, but, thoughts was is that plugins cancelling could maybe add their own reasons down the line, etc

I'm not exactly expecting that plugins will care to act differently on any of the results as they should generally be taking care of those cases ahead of time

Owen1212055:
Any reason why you are providing a component "error" message?

PR #8410, opened 2022-09-30
Load Allay duplication cooldown as a long

NoahvdAa:
It is saved as a long, and the API exposes a long, so loading it as an int may cause some unexpected issues like overflows.

This could potentially be merged into [Aikar's MC-133506 patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0237-MC-135506-Experience-should-save-as-Integers.patch).


Owen1212055:
Yeah perhaps that should be merged. But, LGTM.

electronicboy (approved):

MiniDigger:
does this fix an known issue? 
its kinda strange that mojang did it this way, they must have had a reason?

MiniDigger (approved):

NoahvdAa:
Rebased and merged with Aikars patch

lynxplay (approved):

PR #8413, opened 2022-09-30
Fix BlockDestroyEvent being fired wrong when breaking one of the blocks

### LINKED ISSUE #8341: BlockDestroyEvent always firing for same part of two-block structure when breaking
### Expected behavior

When I break a bed or a door, I expect to receive a `BlockBreakEvent` for the half of the bed/door that I actually broke through interaction, and then a `BlockDestroyEvent` for the other half that gets broken automatically through game logic.

### Observed/Actual behavior

For beds, the `BlockDestroyEvent` is always called for the bottom half, no matter which half of the bed the player actually broke. For doors it's always the upper half. Might affect other blocks as well, only tested those two, couldn't think of anything else that's made up of two (or more) blocks off the top of my head.

### Steps/models to reproduce

Add two `EventHandler`s for the events, log the coordinates whenever they're called for a Bed or Door. When breaking the bottom half of a bed or the top half of a door, both events should log the same coordinates.

### Plugin and Datapack List

```
> plugins
[23:00:03 INFO]: Plugins (1): SpigotTestBox
> datapack list
[23:00:07 INFO]: There are 2 data packs enabled: [vanilla (built-in)], [file/bukkit (world)]
[23:00:07 INFO]: There are no more data packs available
```

### Paper version

```
> version
[22:57:21 INFO]: Checking version, please wait...
[22:57:22 INFO]: This server is running Paper version git-Paper-136 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: 046466f)
Previous version: git-Paper-134 (MC: 1.19.2)
```

### Other

Could be related to #8191

louboecker:
There was a Problem with the `BlockDestroyEvent` firing weirdly sometimes (see #8341 for more).

This was the solution I found for this, it doesn't seem to cause any other issues.

Closes #8341

### DIFF
--- a/patches/server/0919-fix-BlockDestroyEvent-being-fired-wrong-when-breakin.patch
+++ b/patches/server/0919-fix-BlockDestroyEvent-being-fired-wrong-when-breakin.patch
@@ -1,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Emma=20B=C3=B6cker?= <stckoverflw@gmail.com>
+Date: Fri, 30 Sep 2022 20:11:49 +0200
+Subject: [PATCH] fix BlockDestroyEvent being fired wrong when breaking one of
+ the blocks
+
+
+diff --git a/src/main/java/net/minecraft/world/level/block/BedBlock.java b/src/main/java/net/minecraft/world/level/block/BedBlock.java
+index 654a859a37bf991c7a7fa8a44a3d20f8feb223db..a568acaa5c79bce233c8346f0617516519375d68 100644
+--- a/src/main/java/net/minecraft/world/level/block/BedBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/BedBlock.java
+@@ -213,7 +213,10 @@ public class BedBlock extends HorizontalDirectionalBlock implements EntityBlock
+                 BlockState iblockdata1 = world.getBlockState(blockposition1);
+ 
+                 if (iblockdata1.is((Block) this) && iblockdata1.getValue(BedBlock.PART) == BedPart.HEAD) {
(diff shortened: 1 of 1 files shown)


louboecker:
I forgot something haha

Lulu13022002:
~~I have tested this and it fix the issue for those block but introduce a last quirk/bug for the upside of the tallgrass and the bottom part of the bed: it fires the GenericGameEvent two times instead of one like before. This is due to the destroyBlock method that invoke indirectly that event.~~
Also please when you comment the vanilla code, don't make additional change on this code, especially see:
world.setBlock(blockposition1, iblockdata2, 35"**, 0**"); in the DoublePlantBlock

louboecker:
> Also please when you comment the vanilla code, don't make additional change on this code, especially see: world.setBlock(blockposition1, iblockdata2, 35"**, 0**"); in the DoublePlantBlock

Oh yea sorry, I tried around and forgot to change that back

louboecker:
Btw, isn't it actually alright that the Event is being called twice, once for the broken block and once for the block that gets destroyed because of game logic (the locations in the event are alright as well)?

Lulu13022002:
Oh nmv the only things that are changed in this event is the type before, one possible hit would result in two AIR block and now this case has been patched

louboecker:
I did find another issue, btw, since I set the `maxUpdateDepth` to 0 it also doesn't trigger observers pointing at the head of the bed when breaking the foot part of the bed. I'm going to be honest, I have no idea how to fix this smart.

Same issue goes for tall grass breaking the top part with an observer pointing at the bottom part.

The only solution I can find for this issue also calls the `BlockDestroyEvent` on the player-broken part of the bed… (There is a Destroy Event being fired for the not player-broken part, tho)

PR #8416, opened 2022-10-01
Avoid resizing of ArrayList in ChunkMap#getPlayers

Gerrygames:
Avoids resizing of ArrayList in ChunkMap#getPlayers and avoids allocations completely if no players are nearby. This method is only used once in Paper itself but I do use this method in plugins.
I also replaced the cast with instanceof pattern matching. I can revert this part if this is not wanted for some reason.


kennytv (approved):

MiniDigger (approved):
yey for modern java features

PR #8417, opened 2022-10-01
Add system property to dump stack on bad plugin class access

### LINKED ISSUE #8407: Show a stack trace when one plugin loads a class from another plugin which isn't a (soft)depend

It's a pain in the butt sometimes to figure out where from a class is loaded, especially when it involves injection frameworks

Log a stack trace when "Loaded class {0} from {1} which is not a depend or softdepend of this plugin." is logged

Modifying paper myself, dotting printlns all over the plugins code to find when it happens

### Other

_No response_

kennytv:
Closes #8407


MiniDigger (approved):
only comment I have is: do we document those somewhere?

electronicboy:
I don't think so, needs some "here be dragons" page

MiniDigger:
ill drop a message in the docs channel, maybe somebody has time for that 
edit: https://github.com/PaperMC/docs/issues/23

bergerkiller:
Thanks for adding! Will help me a lot now and in the future

PR #8421, opened 2022-10-01
Fix MC-179072

BillyGalbreath:


MiniDigger (approved):

Machine-Maker:
Why is it better to override the `canContinueToUse` method vs. adding another check in the `tick` method? It looks like several line of sight and distance checks are done in the tick method. Is there a benefit to "stopping" the swell one way over another?

BillyGalbreath:
> Why is it better to override the `canContinueToUse` method vs. adding another check in the `tick` method? It looks like several line of sight and distance checks are done in the tick method. Is there a benefit to "stopping" the swell one way over another?

Just ensures the tick runs at least one tick. I guess it _really_ doesnt matter, that's just where I decided to stick it.

Owen1212055 (approved):

PR #8422, opened 2022-10-01
Fix MC-238526

BillyGalbreath:


Machine-Maker:
That is marked as WAI. This is something I've talked about for a while, that spawning is really inconsistent. Some spawners take into account height/biome requirements, and some don't and the vanilla bug tracker is inconsistent with itself in marking these as bugs or WAI. So I don't know about Paper taking a stance one way or another.

MiniDigger:
maybe should be discussed with mojira people first then?

Owen1212055 (changes requested):
Seeing as though many mobs currently have an issue similar to this, I think that it's better to do a (configurable) sort of blanket fix.

See SpawnPlacements#checkSpawnRules

In general, it doesn't make sense *just* to fix this. Although this may technically be an issue changing this may have an impact on spawners/"mob grinders", as suddenly they are now able to spawn in a place they were not able to before. I'm a little iffy with making this change blindly, so, perhaps a configurable option would be better.

PR #8423, opened 2022-10-01
Fix MC-147659

BillyGalbreath:


MiniDigger (approved):

Owen1212055:
https://bugs.mojang.com/browse/MC-147659
The thing is this issue was marked as "Won't Fix", is this something that we want to change if they aren't gonna fix it?

electronicboy:
I've tried engaging somebody on this, but, not heard much, ofc, if I got any info, and told you, I'd have to kill you.

The thing makes 0 sense, if they were setting the cat types purely in the chunk gen logic, sure, I could get behind the "it's not a bug, we just want it to work that way, even if it's odd", but, the things in the finalise spawn method, which the placement of literally makes 0 sense if they're not setting the type anywhere. I can only guess this is a bug and they're just being weird about it, but, idk

electronicboy (approved):

Owen1212055 (approved):
Probably can just say "moved up" and remove the comment. But, looks good. I don't really agree with the bug report resolution.

PR #8424, opened 2022-10-01
Fix MC-210802

BillyGalbreath:


patches/server/0920-Fix-MC-210802.patch
```diff
@@ -0,0 +1,36 @@
...
+diff --git a/src/main/java/net/minecraft/server/level/ChunkMap.java b/src/main/java/net/minecraft/server/level/ChunkMap.java
+index 5461bac37b6dc0575cccd6656b48b2ef18cfaa04..ec67d9044a50fdc9da8d6dd2d9eae2688452dc34 100644
+--- a/src/main/java/net/minecraft/server/level/ChunkMap.java
++++ b/src/main/java/net/minecraft/server/level/ChunkMap.java
+@@ -983,7 +983,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+         return this.anyPlayerCloseEnoughForSpawning(this.getUpdatingChunkIfPresent(chunkcoordintpair.toLong()), chunkcoordintpair, reducedRange);
+     }
+ 
+-    final boolean anyPlayerCloseEnoughForSpawning(ChunkHolder playerchunk, ChunkPos chunkcoordintpair, boolean reducedRange) {
++    public final boolean anyPlayerCloseEnoughForSpawning(ChunkHolder playerchunk, ChunkPos chunkcoordintpair, boolean reducedRange) { // Paper - package -> public
```
NoahvdAa: ~~AT~~

EDIT: This is a Paper-added method, so ATs won't work. Modify the original patch instead.

BillyGalbreath: Working on it, but I'm having issues :3

This should work, but it's not (no error on apply, either). ATs confuse me lol

```
# Fix MC-210802
public net.minecraft.server.level.ChunkMap anyPlayerCloseEnoughForSpawning(Lnet/minecraft/server/level/ChunkHolder;Lnet/minecraft/world/level/ChunkPos;Z)Z
```

BillyGalbreath: > It's a paper added method
> So the at won't find it since it's ran before that
> In that case, just change the existing patch

Got it \o/

patches/server/0920-Fix-MC-210802.patch
```diff
@@ -0,0 +1,23 @@
...
+index 80aa539f7c6a6ee44338de084cdcdf5fb4ef996a..fa4e4ff19a41d56efedcac81d766a65e2a9f2b04 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/goal/EatBlockGoal.java
++++ b/src/main/java/net/minecraft/world/entity/ai/goal/EatBlockGoal.java
+@@ -31,6 +31,12 @@ public class EatBlockGoal extends Goal {
+ 
+     @Override
+     public boolean canUse() {
++        // Paper start - Fix MC-210802
++        net.minecraft.world.level.chunk.LevelChunk chunk = this.mob.level.getChunkIfLoaded(this.mob.blockPosition());
++        if (chunk == null || chunk.playerChunk == null || !((net.minecraft.server.level.ServerLevel) this.mob.level).getChunkSource().chunkMap.anyPlayerCloseEnoughForSpawning(chunk.playerChunk, this.mob.chunkPosition(), false)) {
```
Owen1212055, line 17: How about using ServerLevel#shouldTickBlocksAt instead of doing all of this?

PR #8425, opened 2022-10-01
Fix CraftPersistentDataTypeRegistry CME

Gerrygames:
New approach to https://github.com/PaperMC/Paper/pull/6701. This will have no performance impact after adapters are created.


Owen1212055:
Is accessing this api async something we want to encourage? Especially if we move away from having our own itemstack implementation I am not sure how safe doing any of this would be even.

Machine-Maker:
Well we definitely want to support creating and manipulating itemstacks on other threads. Specifically itemstacks that don't have an in-world representation (aren't in an inventory or container, or entity).

lynxplay:
Would also agree. While the item stack itself does not have to be thread safe as an instance itself (e.g. I don't think we need to promote editing item stacks on multiple threads) we should still allow item stack creation and their customization off the main thread if that thread owns the item stack instance and no other (including main thread) mutates the item.

CodeByCam:
Closing this because a fix was already merged ([in the PR you mentioned](https://github.com/PaperMC/Paper/pull/6701)).

PR #8427, opened 2022-10-01
Exempt players in creative/spectator from nether ceiling damage

Warriorrrr:
Makes it so that players in creative or spectator mode don't receive damage from the nether ceiling when that option is configured.


kennytv (approved):

Machine-Maker:
Technically the "out of world" damage is supposed to bypass invulnerability. Like the OUT_OF_WORLD DamageSource specifically sets the flag to bypass invulnerability. I'm not sure, but I think even spectators die in the void.

PR #8428, opened 2022-10-02
Fix MC-252892

BillyGalbreath:
This bug effects _all_ cooldown items. No one's noticed until the goat horn, though, because most item cooldowns are really short. This fixes only the goat horn, but it might be desirable to remove this line of code entirely to fix all items.


MelnCat:
Imo, it should probably be fixed for all cooldown items, since the effect is visible for other cooldown items as well, and it breaks expected vanilla behavior.

Machine-Maker:
Is this a duplicate of https://github.com/PaperMC/Paper/pull/8008?

Machine-Maker:
Closing as a duplicate of #8008 (which fixed a bug already reported here, #8005 )

PR #8429, opened 2022-10-02
Fix MC-121048

MelnCat:
This pull request fixes [MC-121048](https://bugs.mojang.com/browse/MC-121048) where the combat tracker only contains the final attack. Fixing this bug which for the doomed to fall death messages to occur again.


Owen1212055:
You have two new patches in this PR, not sure if you meant to push that.

MelnCat:
Oh, oops. Accidentally made the branch off the wrong branch

MelnCat:
Fixed

olijeffers0n (approved):
Looks good

Machine-Maker:
I would include a short summary in the patch description of what this change does. I understand it based on the mojira issue, but we shouldn't rely on outside documentation.

MelnCat:
This patch makes the CombatTracker keep the previously recorded damage when the entity dies. Previously, the health was first set to 0, then the damage was recorded. However, the method used to record the damage would reset the combat tracker if the entity was dead, so the CombatTracker was unintentionally wiped. This caused certain death messages to no longer appear, since they rely in the previous CombatTracker records (such as doomed to fall). This patch fixes that by moving the setting of the health to after the damage is recorded, so the method does not reset the CombatTracker.

MelnCat:
Oh wait, you said patch description, hold on

MelnCat:
Patch description added.

electronicboy (approved):

Owen1212055 (approved):
Looks good.
For example, punching a player off a cliff and having them die will now correctly show
that X player caused them to die.

PR #8431, opened 2022-10-02
Supports custom enchantment to use the non item meta methods

### LINKED ISSUE #8426: Inconsistence with Custom Enchantment beetween ItemStack and ItemMeta
### Expected behavior

The `ItemStack#getEnchantmentLevel` are equal to `ItemStack.ItemMeta#getEnchantmentLevel` if used with custom enchantments (extended from org.bukkit.enchantments.Enchantment)

### Observed/Actual behavior

The `ItemStack#getEnchantmentLevel` only works with Enchantments from vanilla and `ItemStack.ItemMeta#getEnchantmentLevel`  return the level correctly

### Steps/models to reproduce

- Create a Enchantment extended from org.bukkit.enchantments.Enchantment and register using org.bukkit.enchantments.Enchantment.registerEnchantment (i know the javadocs say not use this...)
- Use the methods provided by the API to get the level for the two forms directly in item and itemmeta

### Plugin and Datapack List

Plugins (15): ConterCraft, ConterSubsHelper, CoreProtect, gProtector, Lands, LibsDisguises, LiteEco, LuckPerms, PlugManX (PlugMan), ProtocolLib, SellGUI, SentryIntegration, spark, Vault, WorldEdit

### Paper version

This server is running Paper version git-Paper-189 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: ec76e36)
Previous version: git-Paper-174 (MC: 1.19.2)

### Other

Related to https://github.com/PaperMC/Paper/commit/62680d5f83467590ad26cc976ee48b9b2dda3d13 , really not sure if maybe my method (working since many versions) is correct or exists better methods...

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/8426

Even if this "non API" is old, in any case if paper/spigot implement a proper way
to register custom enchantment, paper will need something like that.

### DIFF
--- a/patches/server/0068-Handle-Item-Meta-Inconsistencies.patch
+++ b/patches/server/0068-Handle-Item-Meta-Inconsistencies.patch
@@ -69,8 +69,36 @@ index e937186aaf819a77c80beeb9e08413a1f781c13a..0e19f49ca2496b1c42d27289bcea15d2
      }
  
      public boolean isEnchanted() {
+diff --git a/src/main/java/net/minecraft/world/item/enchantment/EnchantmentHelper.java b/src/main/java/net/minecraft/world/item/enchantment/EnchantmentHelper.java
+index 22925b2f44fc510832ef07290d3109f0394f7d30..6955a1c91e9dbf06a0411c74ad5c6a3edb7dc8bc 100644
+--- a/src/main/java/net/minecraft/world/item/enchantment/EnchantmentHelper.java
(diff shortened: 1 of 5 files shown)


Machine-Maker:
Custom enchantments can just be registered to the internal registry as well as the craftbukkit map right? I always thought that was the preferred way to do custom enchants. I don't think this is an issue that needs to be solved.

PR #8432, opened 2022-10-03
Fix nothing mlg

FriwiDev:
The issue here is that when jumping and sneaking right before the end of a long fall, the client will send incorrect move packets resulting in a teleport back. Due to this, the move packet (impact) of the player will not be considered for fall damage. The following move packet to the impact packet will have a positive y-value again due to the player jumping, resulting in a fall damage reset.

Solution: Apply fall damage also while moving wrongly, but supply the real values the player sent to the damage calculation. This will not have any real negative effect, as the doCheckFallDamage method is meant to be called repeatedly anyway.

Video with demonstration:
https://www.youtube.com/watch?v=dvhoQq9wTAQ
(The explanation in the video is wrong, the behavior of the player on impact is the only relevant thing. The invalid movement packet just has to be one ahead of the "impact" packet for this to work.)

Testing: Tested before and after patch. Can reproduce without patch every second time. Did not reproduce once with patch. This does also not apply fall damage when moving wrongly in general - the onGround check and everything is still in place.

### DIFF
--- a/patches/server/0920-Fix-nothing-mlg.patch
+++ b/patches/server/0920-Fix-nothing-mlg.patch
@@ -1,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: FriwiDev <friwidev@gmail.com>
+Date: Mon, 3 Oct 2022 16:10:06 +0200
+Subject: [PATCH] Fix nothing mlg
+
+The issue here is that when jumping and sneaking right before the end of a long fall, the client will send incorrect move packets resulting in a teleport back. Due to this, this move packet (impact) of the player will not be considered for fall damage. The following move packet to the impact packet will have a positive y-value again due to the player jumping, resulting in a fall damage reset.
+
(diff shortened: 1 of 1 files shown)


olijeffers0n (changes requested):
It does fix the fact that you do not die, however the jumping animation still occurs? 
https://cdn.discordapp.com/attachments/980078018888368142/1026509932629073970/2022-10-03_16-03-01.mp4

FriwiDev:
That is correct, because the internal teleport still occurs. I do not want to remove that because porting back players on invalid move packets still is important. This is mainly a visual bug which will not affect gameplay at all. Alternatively, one could move up the call before the teleport, check if the player is still alive and then teleport. But this would not apply to not dying then - you can also try to negate smaller amounts of damage. I think this visual glitch is acceptable.

olijeffers0n (approved):
Ah, Understood!

MiniDigger (approved):

Ali1v (approved):

nopjar (approved):

kennytv (approved):
Hilarous issue

PR #8434, opened 2022-10-04
Handle legacy quirk in the console formatter

### LINKED ISSUE #7939: Adventure colors prevent formatting tags from closing in console
### Expected behavior

Closing formatting tags in console message stops the formatting, even when using colors.

Without colors:
image

With colors:
image

### Observed/Actual behavior

Closing formatting tags in console message does not stop the formatting, but only when using colors.

Without colors:
image

With colors:
image

### Steps/models to reproduce

```java
Bukkit.getConsoleSender().sendMessage(
    MiniMessage.miniMessage().deserialize("<red>Testing <bold>some</bold> <i>minimessage</i>")
);
```

This gets around the problem, but is inconvenient.
```java
Bukkit.getConsoleSender().sendMessage(
    MiniMessage.miniMessage().deserialize("<red>Testing</red> <red><bold>some</bold></red> <red><i>minimessage</i>")
);
```

### Plugin and Datapack List

Only a test plugin with the above mentioned code.

No datapacks.

### Paper version

I'm assuming since first build adventure was added in.

### Other

The fix mentioned in https://github.com/PaperMC/Paper/issues/7938 does not solve this issue, thus I made it a separate issue.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/7939

This handle correctly the needed reset ansi before each color change like bukkit do.

---
Download the paperclip jar for this pull request: [paper-8434.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/452662891.zip)

### DIFF
--- a/patches/server/0576-Add-support-for-hex-color-codes-in-console.patch
+++ b/patches/server/0576-Add-support-for-hex-color-codes-in-console.patch
@@ -65,10 +65,10 @@ index c3631efda9c7fa531a8a9f18fbee7b5f8655382b..9a3c1314d5a0aa20380662595359580b
  }
 diff --git a/src/main/java/io/papermc/paper/console/HexFormattingConverter.java b/src/main/java/io/papermc/paper/console/HexFormattingConverter.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..6ee7bcf1ecbf4a2c0909e104dcaab5ab9323c09d
+index 0000000000000000000000000000000000000000..b9922b07cb105618390187d98acdf89e728e1f5a
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/console/HexFormattingConverter.java
-@@ -0,0 +1,217 @@
+@@ -0,0 +1,213 @@
 +package io.papermc.paper.console;
 +
 +import io.papermc.paper.configuration.GlobalConfiguration;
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

patches/server/0577-Add-support-for-hex-color-codes-in-console.patch
```diff
@@ -243,7 +243,12 @@ index 0000000000000000000000000000000000000000..b4d0b7ecd56ab952319946854168c129
 +        while (matcher.find()) {
 +            int format = LOOKUP.indexOf(Character.toLowerCase(matcher.group().charAt(1)));
 +            if (format != -1) {
-+                matcher.appendReplacement(buffer, ansi ? ansiCodes[format] : "");
++                String replacement = "";
++                if (ansi && !(format >= 16 && format != 21)) { // handle the upstream quirk at CraftChatMessage L115
++                    replacement = ANSI_RESET;
++                }
++                replacement += ansi ? ansiCodes[format] : "";
```
montlikadani:
```suggestion
                if (ansi) {
                   replacement += ansiCodes[format];
                }
```
To avoid creating extra empty string

tjalp (approved):

patches/server/0577-Add-support-for-hex-color-codes-in-console.patch
```diff
@@ -243,7 +243,12 @@ index 0000000000000000000000000000000000000000..b4d0b7ecd56ab952319946854168c129
 +        while (matcher.find()) {
 +            int format = LOOKUP.indexOf(Character.toLowerCase(matcher.group().charAt(1)));
 +            if (format != -1) {
-+                matcher.appendReplacement(buffer, ansi ? ansiCodes[format] : "");
++                String replacement = "";
++                if (ansi && !(format >= 16 && format != 21)) { // handle the upstream quirk at CraftChatMessage L115
```
lynxplay: Generally, this might just be me, but this logic seems rather convoluted for what it is doing.
Especially in regards to the ansi check done again below.
Like, unless I am missing something, this logic boils down to

```java
String replacement = "";
if (ansi) {
    replacement = (format < 16 || format == 21) ? ANSI_RESET + ansiCodes[format] : ansiCodes[format];
}
matcher.appendReplacement(buffer, replacement);
``` 

This way we
a) get rid of two checks of the ansi bool for no reason
b) de morgans the negation into the expression, which I believe to arguably makes this a lot easier to understand.

Lulu13022002: No, i disagree
The two check are intended, if paper wants to remove the quirk into the bukkit parser
The negation is also intended to follow the bukkit flow (if else), and since the format is already hard to understand (bc the format is an index of a static string to represent color char), i don't think it's worth enough. But in general if people wants to understand more they can still go to the linked line

lynxplay (changes requested):
So, after looking deeper into this, I still stand by my feedback and I believe this to even have a stronger case now.

Your current code appends two ANSI_RESET strings as a replacement for a single `§r`, which is incorrect as your if statement sets `replacement` to `ANSI_RESET` for both colour codes and the reset char. However `ansiCodes[21]` is also the `ANSI_RESET`, leading to a duplication.

Fixing this reduces the if statement to a single "is this a colour code", at which point sticking to a `!(format >= 16)` because "it mirrors the craftbukkit logic" in my opinion is not valid.

Following that, I also strongly disagree with keeping the ansi check separate here. This quirk is not a bug it is a specific feature of the legacy colour format which is simply not very well documented. Colours simply reset text decorations. I don't see how we would ever remove this without breaking either this again or the entire legacy colour format. Hence keeping these checks separate is not really useful beyond just more code.

This entire logic boils down to a
```java
String replacement = "";
if (ansi) replacement = format < 16 ? ANSI_RESET + ansiCodes[format] : ansiCodes[format];
matcher.appendReplacement(buffer, replacement);
```

Lulu13022002:
Thanks i was aware of that duplicate $r ansi but i forgot why i have let it, i will make more test with your changes and push :+1:

Lulu13022002:
Rebased with your better solution

lynxplay (approved):

jpenilla (changes requested):
This doesn't handle RGB colors correctly, I think we need to `private static final String RGB_ANSI = ANSI_RESET + "\u001B[38;2;%d;%d;%dm";`

Lulu13022002:
Thanks, should be good now

jpenilla:
Is there a reason you concat the reset every time instead of just adding it to the replacement pattern?

Lulu13022002:
Yes the formatHexAnsi is also used for the named colors but named colors rely on bukkit logic and will break otherwise ~also will add duplicate reset (see above)~ (i have tested your fix before and it doesn't work)

jpenilla:
https://pastes.dev/GF5jxJ7ksA this should work then?

patches/server/0576-Add-support-for-hex-color-codes-in-console.patch
```diff
@@ -243,7 +244,11 @@ index 0000000000000000000000000000000000000000..b4d0b7ecd56ab952319946854168c129
 +        while (matcher.find()) {
 +            int format = LOOKUP.indexOf(Character.toLowerCase(matcher.group().charAt(1)));
 +            if (format != -1) {
-+                matcher.appendReplacement(buffer, ansi ? ansiCodes[format] : "");
++                String replacement = "";
++                if (ansi) {
++                    replacement = format < 16 ? ANSI_RESET + ansiCodes[format] : ansiCodes[format]; // handle the upstream quirk at CraftChatMessage L115
```
jpenilla: idk if this comment is really needed, it's a quirk of the format in general not just CB's impl (which we don't actually use for logging iirc, we use the LegacyComponentSerializer)

patches/server/0576-Add-support-for-hex-color-codes-in-console.patch
```diff
@@ -243,7 +244,11 @@ index 0000000000000000000000000000000000000000..b4d0b7ecd56ab952319946854168c129
 +        while (matcher.find()) {
 +            int format = LOOKUP.indexOf(Character.toLowerCase(matcher.group().charAt(1)));
 +            if (format != -1) {
-+                matcher.appendReplacement(buffer, ansi ? ansiCodes[format] : "");
++                String replacement = "";
```
jpenilla: would prefer a final variable and if else

jpenilla (changes requested):

patches/server/0576-Add-support-for-hex-color-codes-in-console.patch
Lulu13022002: you're right i will change this

jpenilla:
After thinking about this a little more, I think something like https://pastes.dev/2jDt8IPU6q might be the better solution?

Lulu13022002:
Yes thank you for the help ^^ also your first solution work but just doesn't handle the rgb option

jpenilla (approved):

lynxplay (approved):

lynxplay:
Concerning it works, I am pretty fine with it. Now we have the reset codes directly inside the definition of the colour which *may* prove troublesome in the future if the plain colour is ever needed for whatever reason, tho as I stated earlier that should not happen as this quirk is wanted behaviour :+1:

PR #8436, opened 2022-10-05
Fix MC-121706

5jd:
When a skeleton targets someone from a height, they just keep looking forward instead of up at their target.

This patch is directly a solution proposed by the reporter : https://bugs.mojang.com/browse/MC-121706


kennytv:
We generally don't like accepting prs that copy paste code snippets from Mojira ticket comments, as those often have unintended side-effects, and just copying those implies you haven't dealt with trying to find the underlying cause yourself

5jd:
Hi, I've tested the patch personally for several month on a private fork without any issues, and i thought it could be interesting to PR it to Paper, though if it might cause issues i understand your point of view.

Issue #7383, opened 2022-01-21
Add a "Post" variation of CraftItemEvent with total quantity


viveleroi:
### Is your feature request related to a problem?

`CraftItemEvent` doesn't contain any final-tally information about the craft event. Because it's cancellable it acts more like a "Pre" event. In order to get a full count of how items a player crafted, I currently have to manually check for all of the scenarios a craft can occur. Clicking, shift-clicking, doing both holding matching and non-matching items, factoring available inventory space, etc. It's a nightmare.

### Describe the solution you'd like.

I would like some kind of "Post" CraftItem event, that has:

- The item crafted, probably via `ItemStack getItemStack`
- The total crafted, probably via `int getTotal()`

### Describe alternatives you've considered.

I've looked thoroughly into the API to see if I've missed any existing logic. I've looked at what other devs use and it's always some terribly complicated guessing logic - calculating how many they can make, calculating how many get made based on the way they're clicking, etc.

Doc94:
hmm maybe can check in upstream (sorry) its strange why in SmithItemEvent they use the result item for the event and in CraftItemEvent not.. well in the two events the result are not set in the event...

EDIT: Why cant use CraftItemEvent::getRecipe::getResult?

Machine-Maker:
> Why cant use CraftItemEvent::getRecipe::getResult?

That just gets the result from 1 craft of the recipe, like OP said 
> I currently have to manually check for all of the scenarios a craft can occur. Clicking, shift-clicking, doing both holding matching and non-matching items, factoring available inventory space, etc. It's a nightmare.

tomcraft:
and the event is only called once, even for a "multiple-crafy" click

PR #8443, opened 2022-10-08
Support for result ItemStack in CraftItemEvent

### LINKED ISSUE #8439: Final CraftItemEvent result

The `CraftItemEvent` does not allow you to see how many items the player actually crafted. It seems like the result acts more like a copy. It would be a lot more intuitive if the event offered the number of items the player *actually* crafted.

Some way to access the final result of the craft event, like getFinalResult or just altering the current method that already exists to make it reflect the result of the event.

Hacky & imperfect guessing logic to see how many items the player crafted based on their inventory.

### Other

The original issue explains this pretty well: #7383. I am opening this issue because that one was closed due to being stale.
Also, IIRC the `PlayerTradeEvent` offers a proper result that tells you how many emeralds or other items the player received, and works when the item is shift-clicked into the player's inventory. Maybe the `CraftItemEvent` could be updated to look more like that, if that's possible? Thank you for reading!

Doc94:
This PR close https://github.com/PaperMC/Paper/issues/8439 and https://github.com/PaperMC/Paper/issues/7383 adding a few changes for possible list result item from craft event, this based in the quick move generated when a player use Shift+Click for get all the possible items crafted.

A little [example](https://youtu.be/V0fupWtk3A4) with.
```java
    @EventHandler
    public void onCraft(CraftItemEvent event) {
        HumanEntity humanEntity = event.getWhoClicked();
        humanEntity.sendMessage("You craft is a list of " + event.getItemResults().size());
        event.getItemResults().forEach(itemStack -> {
            humanEntity.sendMessage("You crafted " + itemStack);
        });
    }
```

### DIFF
--- a/patches/api/0399-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
+++ b/patches/api/0399-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 9 Oct 2022 11:58:38 -0300
+Subject: [PATCH] Support for results of ItemStacks in CraftItemEvent
+
+
+diff --git a/src/main/java/org/bukkit/event/inventory/CraftItemEvent.java b/src/main/java/org/bukkit/event/inventory/CraftItemEvent.java
+index cab13877fb38c361f18ce9a433e98fea689c9722..fef15af229bced8cfe526800ffe0591eecc6e06c 100644
+--- a/src/main/java/org/bukkit/event/inventory/CraftItemEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/CraftItemEvent.java
+@@ -11,6 +11,7 @@ import org.jetbrains.annotations.NotNull;
+  */
+ public class CraftItemEvent extends InventoryClickEvent {
+     private Recipe recipe;
++    private java.util.List<org.bukkit.inventory.ItemStack> itemResults; // Paper
+ 
+     public CraftItemEvent(@NotNull Recipe recipe, @NotNull InventoryView what, @NotNull SlotType type, int slot, @NotNull ClickType click, @NotNull InventoryAction action) {
+         super(what, type, slot, click, action);
+@@ -22,6 +23,31 @@ public class CraftItemEvent extends InventoryClickEvent {
+         this.recipe = recipe;
+     }
+ 
++    // Paper start - add more data in CraftItemEvent
++    public CraftItemEvent(@NotNull Recipe recipe, @NotNull InventoryView what, @NotNull SlotType type, int slot, @NotNull ClickType click, @NotNull InventoryAction action, @NotNull java.util.List<org.bukkit.inventory.ItemStack> itemResults) {
++        super(what, type, slot, click, action);
++        this.recipe = recipe;
++        this.itemResults = itemResults;
++    }
++
++    public CraftItemEvent(@NotNull Recipe recipe, @NotNull InventoryView what, @NotNull SlotType type, int slot, @NotNull ClickType click, @NotNull InventoryAction action, int key, @NotNull java.util.List<org.bukkit.inventory.ItemStack> itemResults) {
++        super(what, type, slot, click, action, key);
++        this.recipe = recipe;
++        this.itemResults = itemResults;
++    }
++
++    /**
++     * @return A list of copies of the items result from the craft.
++     */
++    @NotNull
++    public java.util.List<org.bukkit.inventory.ItemStack> getItemResults() {
++        if (this.itemResults == null) {
++            return java.util.Collections.singletonList(recipe.getResult().clone());
++        }
++        return this.itemResults;
++    }
++    // Paper end
++
+     /**
+      * @return A copy of the current recipe on the crafting matrix.
+      */
--- a/patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
+++ b/patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
@@ -1,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 9 Oct 2022 12:00:56 -0300
+Subject: [PATCH] Support for results of ItemStacks in CraftItemEvent
+
+
+diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+index 63c9040ed2349eec500ba6e9090440347c514a3b..4e6ba6588872172d1255c559e712be2ecdd733c6 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -3280,12 +3280,33 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                         org.bukkit.inventory.Inventory top = inventory.getTopInventory();
+                         if (packet.getSlotNum() == 0 && top instanceof CraftingInventory) {
+                             org.bukkit.inventory.Recipe recipe = ((CraftingInventory) top).getRecipe();
+-                            if (recipe != null) {
++                            // Paper start - add more data to CraftItemEvent
++                            org.bukkit.inventory.ItemStack baseResult = (((CraftingInventory) top).getResult() == null) ? null : ((CraftingInventory) top).getResult().clone();
++                            if (recipe != null && baseResult != null) {
++                                List<org.bukkit.inventory.ItemStack> results = new java.util.ArrayList<>();
++                                int itemResultCount = baseResult.getAmount();
++                                if (packet.getClickType() == net.minecraft.world.inventory.ClickType.QUICK_MOVE && action == InventoryAction.MOVE_TO_OTHER_INVENTORY) {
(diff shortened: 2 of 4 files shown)


kennytv (changes requested):
I'm a bit iffy on this, as when you for example craft pickaxes, you don't craft an itemstack of a pickaxe with the amount of 5, but 5 different items

Doc94:
> I'm a bit iffy on this, as when you for example craft pickaxes, you don't craft an itemstack of a pickaxe with the amount of 5, but 5 different items

Ohhh i forget this...
Then what can be? Replace the result method with a list maybe? And need check uf nms handke this too in the same way...

Doc94:
Okay i make a few changes...

- Change the result to a list of items for the case if items with the not default 64 stack
- The result only respect the items in max stack, the result not represent how inventory handle the split of items (in case if you have more than 1 slot where merge items)

a little example added.
https://youtu.be/V0fupWtk3A4

patches/api/0399-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
```diff
@@ -0,0 +1,50 @@
...
++
++    /**
++     * @return A list of copies of the items result from the craft.
++     */
++    @NotNull
++    public java.util.List<org.bukkit.inventory.ItemStack> getItemResults() {
++        if (this.itemResults == null) {
++            return java.util.Collections.singletonList(recipe.getResult().clone());
++        }
++        return this.itemResults;
```
Lulu13022002: This should be an immutable list for consistency

patches/api/0399-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
```diff
@@ -0,0 +1,50 @@
...
++        this.itemResults = itemResults;
++    }
++
++    /**
++     * @return A list of copies of the items result from the craft.
++     */
++    @NotNull
++    public java.util.List<org.bukkit.inventory.ItemStack> getItemResults() {
++        if (this.itemResults == null) {
++            return java.util.Collections.singletonList(recipe.getResult().clone());
```
Lulu13022002: The result of the recipe is already a clone, no need to clone a clone.

patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++                                    for (itemstack = this.player.containerMenu.quickMoveStack(this.player, packet.getSlotNum()); !itemstack.isEmpty() && ItemStack.isSame(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem(), itemstack); itemstack = this.player.containerMenu.quickMoveStack(player, packet.getSlotNum())) {
++                                        if (this.player.containerMenu.slots.get(packet.getSlotNum()).mayPickup(this.player) && (this.player.getInventory().getSlotWithRemainingSpace(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem()) != -1 || this.player.getInventory().getFreeSlot() != -1)) {
++                                            itemResultCount++;
++                                        }
++                                    }
++                                }
++                                int remainResults = itemResultCount;
++                                while(remainResults > 0) {
++                                    int resultAmount = Math.min(baseResult.getMaxStackSize(), itemResultCount);
```
Lulu13022002: This logic doesn't work for example if i try to craft redstone dust from 64 redstone block i get 9 full stacks but the event catch only two full stack (also related to above comment).
should probably replaced by: Math.min(baseResult.getMaxStackSize(), remainResults);

patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++                            if (recipe != null && baseResult != null) {
++                                List<org.bukkit.inventory.ItemStack> results = new java.util.ArrayList<>();
++                                int itemResultCount = baseResult.getAmount();
++                                if (packet.getClickType() == net.minecraft.world.inventory.ClickType.QUICK_MOVE && action == InventoryAction.MOVE_TO_OTHER_INVENTORY) {
++                                    for (itemstack = this.player.containerMenu.quickMoveStack(this.player, packet.getSlotNum()); !itemstack.isEmpty() && ItemStack.isSame(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem(), itemstack); itemstack = this.player.containerMenu.quickMoveStack(player, packet.getSlotNum())) {
++                                        if (this.player.containerMenu.slots.get(packet.getSlotNum()).mayPickup(this.player) && (this.player.getInventory().getSlotWithRemainingSpace(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem()) != -1 || this.player.getInventory().getFreeSlot() != -1)) {
++                                            itemResultCount++;
```
Lulu13022002: This is not correct you need to increment by its crafting amount not just by one so basically:
itemResultCount += baseResult.getAmount(); maybe you can make a constant for that

patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++                            if (recipe != null && baseResult != null) {
++                                List<org.bukkit.inventory.ItemStack> results = new java.util.ArrayList<>();
++                                int itemResultCount = baseResult.getAmount();
++                                if (packet.getClickType() == net.minecraft.world.inventory.ClickType.QUICK_MOVE && action == InventoryAction.MOVE_TO_OTHER_INVENTORY) {
++                                    for (itemstack = this.player.containerMenu.quickMoveStack(this.player, packet.getSlotNum()); !itemstack.isEmpty() && ItemStack.isSame(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem(), itemstack); itemstack = this.player.containerMenu.quickMoveStack(player, packet.getSlotNum())) {
++                                        if (this.player.containerMenu.slots.get(packet.getSlotNum()).mayPickup(this.player) && (this.player.getInventory().getSlotWithRemainingSpace(this.player.containerMenu.slots.get(packet.getSlotNum()).getItem()) != -1 || this.player.getInventory().getFreeSlot() != -1)) {
```
Lulu13022002: This is not needed the quick move stack already check that see the boolean result of #moveItemStackTo

patches/api/0399-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
Doc94: Correct me but pretty sure the lists passed in Collection are already immutable

Lulu13022002: I mean the first one wrapped by the singleton is already immutable but the last return isn't

Doc94: ohhh i forget.. originally in server make the list immutable but i detele this.. maybe can add the immutable list in constructor of the event for make immutable list.

Lulu13022002: That's up to you :+1:
Also if you _love_ annotation and think that this isn't too obvious you can add the Unmodifiable annot for the api

patches/server/0925-Support-for-results-of-ItemStacks-in-CraftItemEvent.patch
Doc94: i use quickmove based in how NMS handle this calling the same method.

Lulu13022002: Oh nvm i have missed this nms logic my bad

Machine-Maker (changes requested):
I don't really like shoving all that logic into the packet listener class. Also, this issue applies to all the "crafts" in the game. I feel like we need a new system of events listening to "recipe results" for all the recipe types that takes into account shift clicks and everything instead of relying on the inventory click event.

Doc94:
> I don't really like shoving all that logic into the packet listener class. Also, this issue applies to all the "crafts" in the game. I feel like we need a new system of events listening to "recipe results" for all the recipe types that takes into account shift clicks and everything instead of relying on the inventory click event.

But then is not correct add tbe correct result in CraftItemEvent?

About the code in packet maybe move this to the CraftEvent or dont remember the class with many methods to call event?

Machine-Maker:
The `CraftItemEvent` doesn't have a "result" field, so it can't be incorrect. It has a "clicked item" field, and that is correct, you only clicked on a single item stack for 1 recipe. You didn't click on anything else. 

So this is the wrong event to use to get this. A different event(s) should be added to handle "recipe use" in a general sense easily applies to all recipe types, not just 1.

Doc94:
> The `CraftItemEvent` doesn't have a "result" field, so it can't be incorrect. It has a "clicked item" field, and that is correct, you only clicked on a single item stack for 1 recipe. You didn't click on anything else.
> 
> So this is the wrong event to use to get this. A different event(s) should be added to handle "recipe use" in a general sense easily applies to all recipe types, not just 1.
> 
> This should be done without mangling the vanilla logic for handling shift clicks.

Then you mean...
- Make a RecipeUseEvent for handle when a recipe is used and have the list of items resulting of this operation like i tried here
- Move the logic from the packet.... exists another place or you mean move the code to the CraftEvent?

Machine-Maker:
The event should probably function like shift clicking multiple stacks. Instead of trying to aggregate all the stacks crafted, just call the event once per craft. So maybe it'll be called 64 times, but that similar behavior already exists with the inventory click event.

Machine-Maker:
This is smth along the lines of what I had in mind. <https://github.com/Machine-Maker/Paper/tree/feature/RecipeCraftEvent>

XLordalX:
What's the status on this?

Ayouuuu:
> This is smth along the lines of what I had in mind. https://github.com/Machine-Maker/Paper/tree/feature/RecipeCraftEvent. Not 100% satisfied, its just a mockup pretty much. And it only handles quickmoves atm, not other types of clicks

Not working in 2×2 crafting

indyteo:
Hi, any update on this?

I'm currently running roughly the same logic as in this PR in my code to compute the number of crafted items, but having to implement it myself can lead to errors (I already had to fix my code twice...), and I would love to see it implemented officially by Paper instead!

Thx

PR #8447, opened 2022-10-09
Fix MC-121706

5jd:
A follow up PR of #8436
This PR is using a seperate branch instead of using master.

Reminder :

When a skeleton targets someone from a height, they just keep looking forward instead of up at their target.

This patch is directly a solution proposed by the reporter : https://bugs.mojang.com/browse/MC-121706


MiniDigger:
I think there is more here.
```java
            if (this.strafingTime > -1) {
                // ...
                this.mob.getMoveControl().strafe(this.strafingBackwards ? -0.5F : 0.5F, this.strafingClockwise ? 0.5F : -0.5F);
                this.mob.lookAt(var0, 30.0F, 30.0F);
            } else {
                this.mob.getLookControl().setLookAt(var0, 30.0F, 30.0F);
            }
```
in the strafing path, lookAt is called, but it has no effect? 
I don't know the difference between `getLookControl().setLookAt` and `lookAt` rn, but that seems to be the root cause? 
there is a reason the control one is only called in the else, I am sure. else you could just omit the lookAt all and call lookcontrol setLookAt for both cases?

5jd:
> I think there is more here.
> 
> ```java
>             if (this.strafingTime > -1) {
>                 // ...
>                 this.mob.getMoveControl().strafe(this.strafingBackwards ? -0.5F : 0.5F, this.strafingClockwise ? 0.5F : -0.5F);
>                 this.mob.lookAt(var0, 30.0F, 30.0F);
>             } else {
>                 this.mob.getLookControl().setLookAt(var0, 30.0F, 30.0F);
>             }
> ```
> 
> in the strafing path, lookAt is called, but it has no effect? I don't know the difference between `getLookControl().setLookAt` and `lookAt` rn, but that seems to be the root cause? there is a reason the control one is only called in the else, I am sure. else you could just omit the lookAt all and call lookcontrol setLookAt for both cases?

 ```java
             if (this.strafingTime > -1) {
                 // ...
                 this.mob.getMoveControl().strafe(this.strafingBackwards ? -0.5F : 0.5F, this.strafingClockwise ? 0.5F : -0.5F);
                 this.mob.getLookControl().setLookAt(var0, 30.0F, 30.0F);
             } else {
                 this.mob.getLookControl().setLookAt(var0, 30.0F, 30.0F);
             }
 ```

It seems that `lookAt` rotates the skeleton body to its target, since replacing `this.mob.lookAt(var0, 30.0F, 30.0F);` with `this.mob.getLookControl().setLookAt(var0, 30.0F, 30.0F);` only rotates the skeleton's head towards the targeted entity and keeps the skeleton's body orientation in the first attack state.

5jd:
While this new approach works at the first glance it seems that some oriantation issues still persists.

PR #8455, opened 2022-10-11
Elder Guardian appearance API

SoSeDiK:
Allows sending elder guardian effect to the player, optionally without a sound.

Note: float value in this packet acts like a boolean, and thus it's a boolean in API too (>= 0.5 < 1.5 plays the same sound, otherwise silent).

P.S. In the server patch, the upper `// Paper` block has `// Paper start - brand support` heading, so that's why a new block was made.

### DIFF
--- a/patches/server/0925-Elder-Guardian-appearance-API.patch
+++ b/patches/server/0925-Elder-Guardian-appearance-API.patch
@@ -15,7 +15,7 @@ index 968aa80b57a31d89852c6f4bc0ec5ed4a98c6530..4712a034ea43a6dada15d68e8657c485
 +    // Paper start
 +    @Override
 +    public void showElderGuardian(boolean silent) {
-+        getHandle().connection.send(new ClientboundGameEventPacket(ClientboundGameEventPacket.GUARDIAN_ELDER_EFFECT, silent ? 0F : 1F));
++        if (getHandle().connection != null) getHandle().connection.send(new ClientboundGameEventPacket(ClientboundGameEventPacket.GUARDIAN_ELDER_EFFECT, silent ? 0F : 1F));
 +    }
 +    // Paper end
 +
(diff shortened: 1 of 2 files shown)


patches/server/0925-Elder-Guardian-appearance-API.patch
```diff
@@ -0,0 +1,24 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -2932,6 +2932,13 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+     }
+     // Paper end
+ 
++    // Paper start
++    @Override
++    public void showElderGuardian(boolean silent) {
++        getHandle().connection.send(new ClientboundGameEventPacket(ClientboundGameEventPacket.GUARDIAN_ELDER_EFFECT, silent ? 0F : 1F));
```
Lulu13022002: Generally these "packet" api also check if the connection isn't null to avoid a potential NPE call from an unsafe place

SoSeDiK: > Generally these "packet" api also check if the connection isn't null to avoid a potential NPE call from an unsafe place

I can do that if needed. IDE swears at me that it can't be null, and many methods (including Paper's) do not have it.

Lulu13022002: Yep, you've right some paper API doesn't use that but not all (see sendMultiBlockChange for example)
To test that see if this throw an NPE inside the PlayerSpawnLocationEvent ?

SoSeDiK: You are right, it did throw NPE. Will update in a minute :)
Good catch!

kennytv (approved):

PR #8456, opened 2022-10-11
Add LivingEntity#swingHand(EquipmentSlot) convenience method

SoSeDiK:
I have dozens of methods where I operate with an item from events solely based on the passed slot.
Old and ugly code (already shortened by not using brackets) would require me to do this:
```java
if (hand == EquipmentSlot.HAND) player.swingMainHand();
else if (hand == EquipmentSlot.OFF_HAND) player.swingOffHand(); // or just else considering no third hand
```
It can now be as simple as `player.swingHand(hand);` :)

### DIFF
--- a/patches/api/0399-Add-LivingEntity-swingHand-EquipmentSlot-convenience.patch
+++ b/patches/api/0399-Add-LivingEntity-swingHand-EquipmentSlot-convenience.patch
@@ -1,0 +1,34 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SoSeDiK <mrsosedik@gmail.com>
+Date: Tue, 11 Oct 2022 22:35:56 +0300
+Subject: [PATCH] Add LivingEntity#swingHand(EquipmentSlot) convenience method
+
+
+diff --git a/src/main/java/org/bukkit/entity/LivingEntity.java b/src/main/java/org/bukkit/entity/LivingEntity.java
+index 4af5e8d0cba6555f7615e4e809d9aff221c0dc4d..3dcb117664bfd170d739a4ee2af3ad30de298b25 100644
+--- a/src/main/java/org/bukkit/entity/LivingEntity.java
++++ b/src/main/java/org/bukkit/entity/LivingEntity.java
+@@ -985,5 +985,23 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
+      * @param hurtDirection hurt direction
+      */
+     void setHurtDirection(float hurtDirection);
++
++    /**
++     * Makes this entity swing their hand.
++     *
++     * <p>This method does nothing if this entity does not
++     * have an animation for swinging their hand.
++     *
++     * @param hand hand to be swung, either {@link org.bukkit.inventory.EquipmentSlot#HAND} or {@link org.bukkit.inventory.EquipmentSlot#OFF_HAND}
++     * @throws IllegalArgumentException if invalid hand is passed
++     */
++    default void swingHand(@NotNull org.bukkit.inventory.EquipmentSlot hand) {
++        com.google.common.base.Preconditions.checkArgument(hand == org.bukkit.inventory.EquipmentSlot.HAND || hand == org.bukkit.inventory.EquipmentSlot.OFF_HAND, "EquipmentSlot was not a hand!");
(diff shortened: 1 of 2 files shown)


patches/api/0399-Add-extra-swing-method-for-convenience.patch
```diff
@@ -0,0 +1,29 @@
...
++    /**
++     * Makes this entity swing their hand.
++     *
++     * <p>This method does nothing if this entity does not
++     * have an animation for swinging their hand.
++     *
++     * @param hand hand to be swung, either {@link org.bukkit.inventory.EquipmentSlot#HAND} or {@link org.bukkit.inventory.EquipmentSlot#OFF_HAND}
++     */
++    default void swingHand(@NotNull org.bukkit.inventory.EquipmentSlot hand) {
++        if (hand == org.bukkit.inventory.EquipmentSlot.HAND) swingMainHand();
```
jpenilla: add brackets to ifs and `this.` to instance method calls. also should have a preconditions check that one of the required values was passed

jpenilla (changes requested):

patches/api/0399-Add-extra-swing-method-for-convenience.patch
```diff
@@ -0,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SoSeDiK <mrsosedik@gmail.com>
+Date: Tue, 11 Oct 2022 22:35:56 +0300
+Subject: [PATCH] Add extra swing method for convenience
```
jpenilla:
```suggestion
Subject: [PATCH] Add LivingEntity#swingHand(EquipmentSlot) convenience method
```

patches/api/0399-Add-extra-swing-method-for-convenience.patch
SoSeDiK: Done & squished to a single commit.
Oh, forgot about preconditions, oops.

jpenilla: might have missed my edit on the other comment too, sorry :p
(naming the patch with swingHand (not swing) like the method)

SoSeDiK: Oh yea, just copypasted it without noticing 😅
Probably should be sleeping now instead of doing changes at 2 AM :)

SoSeDiK:
By the way, would `EquipmentSlot#isHand` be a neat addition? 🤔
(can do a separate PR or add a commit to this one with separate .patch)
 
You'd have to do a long `EquipmentSlot.HAND || EquipmentSlot.OFF_HAND` instead of simple `#isHand()` to manually check for the hand if needed.

Edit: If so, maybe `#isArmor` could also serve a purpose for checking armor slots? Since `!isHand()` might be off in case of new equipment slots in the future, and manually checking all four is "meh".

patches/api/0399-Add-LivingEntity-swingHand-EquipmentSlot-convenience.patch
```diff
@@ -0,0 +1,34 @@
...
++     * Makes this entity swing their hand.
++     *
++     * <p>This method does nothing if this entity does not
++     * have an animation for swinging their hand.
++     *
++     * @param hand hand to be swung, either {@link org.bukkit.inventory.EquipmentSlot#HAND} or {@link org.bukkit.inventory.EquipmentSlot#OFF_HAND}
++     * @throws IllegalArgumentException if invalid hand is passed
++     */
++    default void swingHand(@NotNull org.bukkit.inventory.EquipmentSlot hand) {
++        com.google.common.base.Preconditions.checkArgument(hand == org.bukkit.inventory.EquipmentSlot.HAND || hand == org.bukkit.inventory.EquipmentSlot.OFF_HAND, "EquipmentSlot was not a hand!");
```
jpenilla: Would be nice to include the value of `hand` in the exception message

as far as isHand/isArmor, it seems reasonable to me. probably do it in a separate PR, since with the aforementioned change this one is basically ready

SoSeDiK: Hopefully, this will do :)

jpenilla (approved):

PR #8108, opened 2022-07-07
Paper Plugins, Bootstrappers, Loaders and More?

Owen1212055:
This implements paper plugins, which can be created by ``paper-plugin.yml`` in your plugin resources.
At runtime, they are all in the end the same "JavaPlugins". However, in the future, it is possible that this can be expanded.
Note: You can have both paper plugin and spigot plugin ymls and paper will prioritize the paper yml. The goal is to also have paper and spigot plugins work fine together (because again, they are the same at run time).

**This was opened to collect a lot of feedback**
Difference between paper plugins:
- library loading is done at runtime by plugin loaders, not anymore in the config.
- They have a cool ``bootstrapper`` field, allowing code to be executed even before a Minecraft server is created!
- Cannot be pre-1.19
- Cannot be loaded on spigot servers (shocker)

Resolves:
https://github.com/PaperMC/Paper/issues/5992
https://github.com/PaperMC/Paper/issues/7961
https://github.com/PaperMC/Paper/issues/7955

---
### Final testing process
Hoping to target this release to allow for future API to move forward. Please test this on your servers to ensure that all spigot plugins work.
Paper plugins will be **experimental**.

---
Download the paperclip jar for this pull request: [paper-8108.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/563289954.zip)

PR #8463, opened 2022-10-12
Added PermissibleBase getters and setters

Badbird5907:
Adding this API will remove the need to use reflections in plugins to get and set `PermissibleBase` on `CraftHumanEntity` to override permissions logic

### DIFF
--- a/patches/api/0400-Added-PermissibleBase-getter-and-setter.patch
+++ b/patches/api/0400-Added-PermissibleBase-getter-and-setter.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Added PermissibleBase getter and setter
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 1d98abff1ad0116f7a2599f078aa730cb84843c1..443e63db6fc37ff5e55a09d1c0238d103f490ab6 100644
+index 1d98abff1ad0116f7a2599f078aa730cb84843c1..d0dec0999a2ffe0117b88a9bddf3b707bcd16710 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
 @@ -33,6 +33,7 @@ import org.bukkit.event.player.PlayerResourcePackStatusEvent;
@@ -43,7 +43,7 @@ index 1d98abff1ad0116f7a2599f078aa730cb84843c1..443e63db6fc37ff5e55a09d1c0238d10
      /**
       * Gets the no-ticking view distance for this player.
       * <p>
-@@ -2832,4 +2833,19 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+@@ -2832,4 +2833,21 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      @Override
      Spigot spigot();
      // Spigot end
@@ -54,12 +54,14 @@ index 1d98abff1ad0116f7a2599f078aa730cb84843c1..443e63db6fc37ff5e55a09d1c0238d10
 +     * @param base {@link PermissibleBase}
 +     * @return the old {@link PermissibleBase}
 +     */
-+    PermissibleBase setPermissibleBase(PermissibleBase base);
++    @NotNull
++    PermissibleBase setPermissibleBase(@NotNull PermissibleBase base);
 +
 +    /**
 +     * Get the player's {@link PermissibleBase}
 +     * @return the player's {@link PermissibleBase}
 +     */
++    @NotNull
 +    PermissibleBase getPermissibleBase();
 +    // Paper end
  }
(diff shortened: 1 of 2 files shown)


Owen1212055:
Note: https://github.com/PaperMC/Paper/pull/8108

Do note that this expands the whole permission overriding and allows you to override the permission logic in the SimplePluginManager.

Badbird5907:
Added the missing annotations causing tests to fail

patches/api/0400-Added-PermissibleBase-getter-and-setter.patch
```diff
@@ -0,0 +1,67 @@
...
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index 1d98abff1ad0116f7a2599f078aa730cb84843c1..d0dec0999a2ffe0117b88a9bddf3b707bcd16710 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -33,6 +33,7 @@ import org.bukkit.event.player.PlayerResourcePackStatusEvent;
+ import org.bukkit.inventory.EquipmentSlot;
+ import org.bukkit.inventory.ItemStack;
+ import org.bukkit.map.MapView;
++import org.bukkit.permissions.PermissibleBase;
```
kennytv: Inline the import

patches/api/0400-Added-PermissibleBase-getter-and-setter.patch
```diff
@@ -0,0 +1,67 @@
...
+ import org.bukkit.plugin.Plugin;
+ import org.bukkit.plugin.messaging.PluginMessageRecipient;
+ import org.bukkit.scoreboard.Scoreboard;
+@@ -1965,7 +1966,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      * @param saturationLevel the saturation level of the player
+      */
+     public void sendHealthUpdate(final double health, final int foodLevel, final float saturationLevel);
+-
```
kennytv: Don't change indents/spacing

kennytv (changes requested):

Badbird5907:
Did the requested changes
_idea didn't like the indents_

patches/server/0926-Added-PermissibleBase-getter-and-setter.patch
```diff
@@ -0,0 +1,34 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
+index e779dcc4982ff51e4d450265fd61bc26e8e74d3a..61fd80c691476c3eaec9ff182715913da3a0f051 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
+@@ -70,7 +70,7 @@ import org.bukkit.plugin.Plugin;
+ public class CraftHumanEntity extends CraftLivingEntity implements HumanEntity {
+     private CraftInventoryPlayer inventory;
+     private final CraftInventory enderChest;
+-    protected final PermissibleBase perm = new PermissibleBase(this);
++    protected PermissibleBase perm = new PermissibleBase(this);
```
Lulu13022002: Missing the paper comment

patches/api/0400-Added-PermissibleBase-getter-and-setter.patch
```diff
@@ -0,0 +1,32 @@
...
+@@ -2832,4 +2832,21 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+     @Override
+     Spigot spigot();
+     // Spigot end
++
++    // Paper start
++    /**
++     * Set the player's {@link org.bukkit.permissions.PermissibleBase}
++     * @param base {@link org.bukkit.permissions.PermissibleBase}
++     * @return the old {@link org.bukkit.permissions.PermissibleBase}
```
Lulu13022002: These links seems a little bit too redundant, maybe keep only the links into the description but not into the parameter explanation ?

zml2008:
this feels like the kind of thing that exposes internal details of api impl which makes evolving the permissions api harder, and makes it easier for plugin developers to make mistakes without any traceability.

you'd be better off reworking how permissions are set and computed in the first place to remove the numerous footguns developers encounter when customizing permissions api logic - things like state tracking across the various join/quit/kick events, the permissions impact of recalculating permissions, and all the logic with default permissions

electronicboy:
The thing with evolving the perm API is that it's generally one of the many things which has been left to rot, hence why any perm plugin of note will literally just rip the thing out in order to add basic features, such as verbose mode, wildcards, etc

I'm much open towards expanding this API, but, I'd much rather that be an effort kicked off through perm plugin devs whom know their needs here so that we can actually deliver towards it, but, at this point, people reflect into this mess anyways that I don't have a strong urge to deny this personally, but, did wanna ask people like Luck if they had any thoughts or considerations; It's not like exposing this in the API is a huge priority for us and I'd much rather see perm plugin devs discuss this sorta thing

Owen1212055:
Closing, as in general as specified above it's best to instead expand this API instead of trying to move around it like this.

PR #8474, opened 2022-10-15
Bedrock pattern fix

19MisterX98:
image
image

Paper uses a fixed number for bedrock in the same column.
Vanilla uses the y coord of the bedrock block.


Machine-Maker:
Yep, this was incorrectly copied. Another casualty of the decompiler using the same variable names. If you look at `SurfaceRules$VerticalGradientConditionSource#apply$VerticalGradientCondition you can see there are two variables named `i`. Looks like it was copied incorrectly and the slight difference in bedrock formations wasn't noticed.

Machine-Maker (approved):

kennytv (approved):

PR #8478, opened 2022-10-16
More vanilla friendly methods to update trades

### LINKED ISSUE #8471: Villager#setVillagerLevel does not update new trades
### Expected behavior

When we choose a level of trade of the villager and he has well on the profession, that it unlocks the trades of the level.

### Observed/Actual behavior

Nothing happens except for an increase in the selling price of the first 2 trades

### Steps/models to reproduce

```Java
villager.setVillagerLevel(3); // By example
```
image

### Plugin and Datapack List

Plugins (18): CMILib, Custom_items, DataManager, Eris, Essentials, FastAsyncWorldEdit (WorldEdit), Insights, MobFarmManager, NBTAPI, NMS_IMPLEMENTATION, PlaceholderAPI, ShopGUIPlus, SkyblockVP, SkyblockVP_Extends_IsValue, SkyblockvpMineralGenerator, TAB, Vault, VoidPlot

### Paper version

[00:57:32 INFO]: Checking version, please wait...
[00:57:33 INFO]: This server is running Paper version git-Paper-211 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: eb68bd4)
Previous version: git-Paper-196 (MC: 1.19.2)

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/8471

Adds two new methods increaseLevel and addTrades based on Machine's comment
The first one increase the level from a such amount like in vanilla
The second unlock new trades, in vanilla this is used:
- for villager: 2 trades are unlocked while level up
- for wandering trader: 5 trades are unlocked + 1 extra


patches/api/0403-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,67 @@
...
+@@ -74,5 +74,11 @@ public interface WanderingTrader extends AbstractVillager {
+      * @param location location to wander towards (world is ignored, will always use the entity's world)
+      */
+     void setWanderingTowards(@org.jetbrains.annotations.Nullable org.bukkit.Location location);
++
++    /**
++     * Gives to this wandering trader some potential new trades.
++     * @param amount The amount of trades to give
++     */
++    void addTrades(int amount);
```
Machine-Maker: I don't think the same method should be in WanderingTrader. Due to the difference in how trades are added, a single method with a single amount doesn't really make sense to me. Some other scheme should be here to reflect the two separate pools of trades for wandering traders (doesn't have to be in this PR).

patches/api/0403-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,67 @@
...
++     * @see #setVillagerLevel(int)
++     */
++    boolean increaseLevel(int amount);
++
++    /**
++     * Gives to this villager some potential new trades
++     * based to its profession and level.
++     * @param amount The amount of trades to give
++     */
++    void addTrades(int amount);
```
Machine-Maker: I think this should have a boolean return value. There are 2 villager professions that this method will not add trades to, the nitwit and none. The boolean would reflect if any trades were actually added.

patches/server/0928-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,124 @@
...
+ 
+         return j >= 0 && j < this.inventory.getContainerSize() ? SlotAccess.forContainer(this.inventory, j) : super.getSlot(mappedIndex);
+     }
++    // Paper start
++    protected void updateTrades() {
++        updateTrades(2);
++    }
+ 
+-    protected abstract void updateTrades();
++    public abstract void updateTrades(int amount);
```
Machine-Maker: This should return a boolean on whether trades were actually added (for use in the boolean API return value)

patches/server/0928-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,124 @@
...
+index f0b910df1ee471b4d72d97c6197ab14f2854976e..2cc03fe21c879d2cca0c2d5ac85094203fb0709e 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftVillager.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftVillager.java
+@@ -99,6 +99,30 @@ public class CraftVillager extends CraftAbstractVillager implements Villager {
+     }
+ 
+     // Paper start
++    @Override
++    public boolean increaseLevel(int amount) {
++        Preconditions.checkArgument(amount > 0, "Level earned must be positive");
```
Machine-Maker: The setVillagerLevel method throws an IAE if the amount is invalid. I think this should just check that the amount is greater than 0 and it won't set the villager level to above 5. throw an exception if it does instead of just ignoring it.

patches/server/0928-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,124 @@
...
++
++        it.unimi.dsi.fastutil.ints.Int2ObjectMap<net.minecraft.world.entity.npc.VillagerTrades.ItemListing[]> trades =
++            net.minecraft.world.entity.npc.VillagerTrades.TRADES.get(this.getHandle().getVillagerData().getProfession());
++
++        if (trades == null || trades.isEmpty()) return false;
```
Machine-Maker: If the current profession doesn't have trades, shouldn't the level still be increased? The API allows for changing the level of trade-less villagers.

Machine-Maker (changes requested):

patches/api/0403-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,57 @@
...
++     * method {@link #setVillagerLevel(int)}.
++     * <p>
++     * A villager with a level of 1 and no experience is liable to lose its
++     * profession.
++     * <p>
++     * A master villager has a level of 5 in its profession and
++     * will unlock 10 trades (2 per level).
++     *
++     * @param amount The amount of level
++     * @return If this villager has earned at least one level
```
Machine-Maker: This return value needs to be re-thought. An exception is thrown if the villager won't gain a level, so technically this should never return false. Perhaps the boolean can be if trades were added? Just spitballing.

Machine-Maker (changes requested):

patches/api/0403-More-vanilla-friendly-methods-to-update-trades.patch
Lulu13022002: The implementation already do that, this is a javadoc issue due to (a leftover i think)

kennytv (approved):

patches/server/0928-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,74 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/npc/Villager.java b/src/main/java/net/minecraft/world/entity/npc/Villager.java
+index 10b45ec24a5a0867106d1694312385ad1e267f43..a12022868ed7c6cb74557bcf1a5028c144035068 100644
+--- a/src/main/java/net/minecraft/world/entity/npc/Villager.java
++++ b/src/main/java/net/minecraft/world/entity/npc/Villager.java
+@@ -919,6 +919,12 @@ public class Villager extends AbstractVillager implements ReputationEventHandler
+ 
+     @Override
+     protected void updateTrades() {
++        // Paper start
++        updateTrades(2);
```
Machine-Maker: Use the constant `TRADES_PER_LEVEL` here instead of `2`. Probably best to make use of those constants that get unused due to compiler optimizations.

patches/server/0928-More-vanilla-friendly-methods-to-update-trades.patch
```diff
@@ -0,0 +1,74 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftVillager.java
+@@ -99,6 +99,33 @@ public class CraftVillager extends CraftAbstractVillager implements Villager {
+     }
+ 
+     // Paper start
++    @Override
++    public boolean increaseLevel(int amount) {
++        Preconditions.checkArgument(amount > 0, "Level earned must be positive");
++        int supposedFinalLevel = this.getVillagerLevel() + amount;
++        Preconditions.checkArgument(1 <= supposedFinalLevel && supposedFinalLevel <= 5, "Final level reached after the donation (%d) must be between [1, 5]".formatted(supposedFinalLevel));
```
Machine-Maker: Instead of `1` and `5`, use `VillagerData.MIN_VILLAGER_LEVEL` and `VillagerData.MAX_VILLAGER_LEVEL`

Machine-Maker (approved):

Owen1212055 (approved):

PR #8479, opened 2022-10-16
Add LivingEntity knockback API

MelnCat:
This pull request adds a knockback method on LivingEntity to knock back the entity from a specific position and a specific knockback.

Currently, knocking back an entity can be done with setting the velocity, but this method does not take in vanilla knockback mechanics into consideration such as knockback resistance. Copying the knockback method from the NMS LivingEntity works, but it causes a lot of code duplication, and won't update if the built-in vanilla knockback dynamics change. The knockback method added by this pull request will expose the built-in knockback method from the NMS LivingEntity to allow simulating vanilla knockback mechanics.

### DIFF
--- a/patches/api/0403-Add-entity-knockback-API.patch
+++ b/patches/api/0403-Add-entity-knockback-API.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add entity knockback API
 
 
 diff --git a/src/main/java/org/bukkit/entity/LivingEntity.java b/src/main/java/org/bukkit/entity/LivingEntity.java
-index 319e4571aca24d1e3f6c85b7435d65c0e77a5245..dba1729a693a6883f94896fcf16b162abe83b1fe 100644
+index 319e4571aca24d1e3f6c85b7435d65c0e77a5245..c9a44e8024f903da83181ee752c971bab22c8895 100644
 --- a/src/main/java/org/bukkit/entity/LivingEntity.java
 +++ b/src/main/java/org/bukkit/entity/LivingEntity.java
 @@ -1003,5 +1003,17 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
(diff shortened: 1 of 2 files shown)


patches/api/0403-Add-entity-knockback-API.patch
```diff
@@ -0,0 +1,27 @@
...
+             this.swingOffHand();
+         }
+     }
++
++    /**
++     * Knocks back this entity from a specific direction with a specified strength.
++     *
++     * The direction specified in this method will be the direction of the source of the knockback,
++     * so the entity will be pushed in the opposite direction.
++     * @param strength The strength of the knockback. If the entity hsa knockback resistance, this will be reduced.
```
Lulu13022002: A little typo: "If the entity **hsa** knockback resistance, this will be reduced"

patches/server/0928-Add-entity-knockback-API.patch
```diff
@@ -0,0 +1,21 @@
...
+index e807ef287136e7b3931197e45434a3f618cf3054..152b07ad1d1027d5dd5cd861a687229c04853c66 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
+@@ -977,5 +977,10 @@ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
+         }
+         throw new IllegalArgumentException(entityCategory + " is an unrecognized entity category");
+     }
++
++    @Override
++    public void knockback(double strength, double directionX, double directionZ) {
```
Lulu13022002, line 17: This should add a precondition if the strength is negative instead of ignoring it

patches/api/0403-Add-entity-knockback-API.patch
MelnCat: Fixed

patches/server/0928-Add-entity-knockback-API.patch
MelnCat: Precondition added

Lulu13022002: I mean the strength must be positive (excluding zero): strength > 0.0D

MelnCat: Ah, alright

kennytv (approved):

Issue #8481, opened 2022-10-17
Wardens confused after chunk loading after the chunk system rewrite


petersv5:
### Expected behavior

Wardens occupied by a repeating note block on chunk unload will still be occupied (aggrod) after loading the chunk again, or at least quickly reaquire the note block as a target.

### Observed/Actual behavior

Wardens lose all interest in the old target and fail to require the note block target when reloaded. The wardens eventually despawn due to a lack of aggro.

### Steps/models to reproduce

Any warden-based mob switch will show the effect.
Have a warden distracted by a repeating note block (4gt clock). Unload chunk by moving away or restarting the server. Return to the chunk and observe the warden not being interested in the repeating note block and eventually despawning.

### Plugin and Datapack List

None

### Paper version

This server is running Paper version git-Paper-"236a296" (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: 236a296)

Error was git bidect:ed to 01a13871deefa50e186a10b63f71c5e0459e7d30, the "Rewrite chunk system" committed 2022-09-26.

### Other

Error was git bidect:ed to 01a13871deefa50e186a10b63f71c5e0459e7d30, the "Rewrite chunk system" committed 2022-09-26.

petersv5:
I have instrumented the code a bit. It seems that warden::shouldListen() is never called. 
A bit further seems to point to there not being a VibrationListener for the warden, or at least not one that has VibrationListener::handleGameEvent() being called. 

I am not sure how to dig further into this. I have not really ofund/understood code to create the VibrationListner instance when restoring a chunk.

Owen1212055:
The issue is that on ``ServerLevel$EntityCallbacks#onTrackingStart`` ``entity.updateDynamicGameEventListener(DynamicGameEventListener::add);`` is called, which is supposed to register the dynamic listener into the chunk that the entity is loaded in. 

However, it seems that for whatever reason at this stage the chunk the entity is in isn't loaded yet. 
So, when ``DynamicGameEventListener#move`` is called this bit that is supposed to register the current listener into the
chunk returns null and doesn't do anything.
```java
ifChunkExists(world, this.lastSection, (dispatcher) -> {
                    dispatcher.register(this.listener);
                });
```

petersv5:
It seems that the updateDynamicGameEventListener() gets called when then chunk is updating to BORDER. Is the issue that the update to BORDER is fired prematurely, before the chunk really has entered that state?

[...]

PR #8483, opened 2022-10-18
Make a chunk accessible before entities are asked to interact with it during loading

petersv5:
While digging into issue  #8481 it seems that the order between chunk.onChunkLoad() causing ServerChunkCache#addLoadedChunk() making the chunk accesible and changeEntityChunkStatus() instruction entites to interact with the chunk is wrong. 

I m not sure if there are any non-obvious bad effects from delaying the updates to the entities until after the onChunkLoad() calls. I did some test runs and went through the code as far as I could and nothing obvious was found.


Spottedleaf:
I can't accept this fix because it moves the entity load after the chunk load. That is, this ordering will expose entities inside the chunk before they are added to the world.

PR #8485, opened 2022-10-19
Add Sneaking API for Entities

dawon:
This allows developer to sneak any entity which will make its Name Tag less visible.

### DIFF
--- a/patches/api/0421-Add-Sneaking-API-for-Entities.patch
+++ b/patches/api/0421-Add-Sneaking-API-for-Entities.patch
@@ -1,0 +1,56 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: dawon <dawon@dawon.eu>
+Date: Wed, 19 Oct 2022 23:36:42 +0200
+Subject: [PATCH] Add Sneaking API for Entities
+
+
+diff --git a/src/main/java/org/bukkit/entity/Entity.java b/src/main/java/org/bukkit/entity/Entity.java
+index 33a6b7a27dc91552799c07a7aad9b3df31ad13f7..cdbc7329cf5f67d66e31eb31e83b9e7997040f72 100644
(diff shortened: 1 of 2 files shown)


patches/server/0929-Add-Sneaking-API-for-Entities.patch
```diff
@@ -0,0 +1,52 @@
...
++        super.setSneaking(sneak);
++        // Paper end
+     }
+ 
+     @Override
+     public boolean isSneaking() {
+-        return this.getHandle().isShiftKeyDown();
++        // Paper start - move the logic to Entity class
++        return super.isSneaking();
++        // Paper end
```
Machine-Maker: Is it worth making this change? functionally its no different and so its extra useless diff, but at the same time its smart to have it in one place instead of two.

dawon: So just put it in Entity and leave the duplicated code in Player?

Machine-Maker: I don't know, which is why it's not a requested change, just a comment. Someone else might think that, but I don't have an opinion either way.

dawon: Me neither. But your point it will make more diff is valid and knowing how much stuff needs to be updated during update I guess it makes sense to just leave it in there, considering it is oneliner anyway...

electronicboy: Yea, if there was a bunch of logic in there that needed to be moved and modified, sure; but, for a singular line that is purely just delegating a call over elsewhere, it's just pointless diff

electronicboy (approved):

YanisBft (approved):

dawon:
rebased

dawon:
rebased

Machine-Maker (approved):

dawon:
rebased on master

PR #8492, opened 2022-10-22
Expand on entity serialization API

SoSeDiK:
Makes an entity serializable via API no matter what.
Previously you'd get an empty component (and an error later when trying to spawn the deserialized entity) in some cases:
- When the entity is marked as non-persistent.
- When the entity is a passenger.
- When the entity is not saveable (misc entities like lightning, knot, etc.).
- When the entity is "invalid" (dead, unloaded, etc.).

Now there's no problem serializing & deserializing & spawning entities in those cases.
___
Edit: see [the comment](https://github.com/PaperMC/Paper/pull/8492#issuecomment-1348574662) below for the new flags explanation

### DIFF
--- a/patches/api/0312-Add-Raw-Byte-Entity-Serialization.patch
+++ b/patches/api/0312-Add-Raw-Byte-Entity-Serialization.patch
@@ -3,35 +3,141 @@ From: Mariell Hoversholm <proximyst@proximyst.com>
 Date: Sun, 24 Oct 2021 16:19:26 -0400
 Subject: [PATCH] Add Raw Byte Entity Serialization
 
+Co-authored-by: SoSeDiK <mrsosedik@gmail.com>
 
+diff --git a/src/main/java/io/papermc/paper/entity/EntitySerializationFlag.java b/src/main/java/io/papermc/paper/entity/EntitySerializationFlag.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..e29dd78c1e5114915a45b02a5276e3ef7e283400
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/EntitySerializationFlag.java
+@@ -0,0 +1,38 @@
++package io.papermc.paper.entity;
++
++import org.bukkit.UnsafeValues;
++import org.bukkit.entity.Entity;
++import org.bukkit.entity.Player;
++import org.bukkit.entity.SpawnCategory;
++
++/**
++ * Represents flags for entity serialization.
++ *
++ * @see UnsafeValues#serializeEntity(Entity, EntitySerializationFlag... serializationFlags)
++ */
++public enum EntitySerializationFlag {
++
++    /**
++     * Serialize invalid (dead, unloaded, etc.) entities.
++     *
++     * @see Entity#isValid()
++     */
++    FORCE,
++    /**
++     * Serialize misc non-savable entities like lighting bolts, fishing bobbers, etc.
++     * <br>Note: players require a separate flag: {@link #PLAYER}.
++     *
++     * @see SpawnCategory#MISC
++     */
++    MISC,
++    /**
++     * Include passengers in the serialized data.
++     */
++    PASSENGERS,
++    /**
++     * Allow serializing {@link Player}s.
++     * <p>Note: deserializing player data will always fail.
++     */
++    PLAYER
++
++}
 diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
-index cbc63144e5eb35799548209f8fbee70d0c20a53d..240ac3f658ed24d3980707e146e6dfee6c9b31a0 100644
+index cbc63144e5eb35799548209f8fbee70d0c20a53d..7e6f435a060a79a45060248c26dbca7c73366240 100644
 --- a/src/main/java/org/bukkit/UnsafeValues.java
 +++ b/src/main/java/org/bukkit/UnsafeValues.java
-@@ -167,6 +167,14 @@ public interface UnsafeValues {
+@@ -167,6 +167,74 @@ public interface UnsafeValues {
  
      ItemStack deserializeItem(byte[] data);
  
-+    byte[] serializeEntity(org.bukkit.entity.Entity entity);
++    // Paper start - Raw entity serialization API
++    /**
++     * Serializes the provided entity.
++     *
++     * @param entity entity
++     * @return serialized entity data
++     * @see #serializeEntity(org.bukkit.entity.Entity, io.papermc.paper.entity.EntitySerializationFlag...)
++     * @see #deserializeEntity(byte[], World, boolean, boolean)
++     */
++    default byte[] serializeEntity(@org.jetbrains.annotations.NotNull org.bukkit.entity.Entity entity) {
++        return serializeEntity(entity, new io.papermc.paper.entity.EntitySerializationFlag[0]);
++    }
 +
-+    default org.bukkit.entity.Entity deserializeEntity(byte[] data, World world) {
++    /**
++     * Serializes the provided entity.
++     *
++     * @param entity entity
++     * @param serializationFlags serialization flags
++     * @return serialized entity data
++     * @see #deserializeEntity(byte[], World, boolean, boolean)
++     */
++    byte[] serializeEntity(@org.jetbrains.annotations.NotNull org.bukkit.entity.Entity entity, @org.jetbrains.annotations.NotNull io.papermc.paper.entity.EntitySerializationFlag... serializationFlags);
++
++    /**
++     * Deserializes the entity from data.
++     * <br>The entity's {@link java.util.UUID} as well as passengers will not be preserved.
++     *
++     * @param data serialized entity data
++     * @param world world
++     * @return deserialized entity
++     * @see #deserializeEntity(byte[], World, boolean, boolean)
++     * @see #serializeEntity(org.bukkit.entity.Entity, io.papermc.paper.entity.EntitySerializationFlag...)
++     * @see org.bukkit.entity.Entity#spawnAt(Location, org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason)
++     */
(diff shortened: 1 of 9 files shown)


patches/server/0703-Add-Raw-Byte-Entity-Serialization.patch
```diff
@@ -58,7 +38,8 @@ index 0a3d447f87698dc786d6cab6ded27eb1b5780204..d3b37adfe6d66e82db18d94f143af3ab
 +        Preconditions.checkArgument(entity instanceof org.bukkit.craftbukkit.entity.CraftEntity, "only CraftEntities can be serialized");
 +
 +        CompoundTag compound = new CompoundTag();
-+        ((org.bukkit.craftbukkit.entity.CraftEntity) entity).getHandle().serializeEntity(compound);
++        compound.putString("id", net.minecraft.world.entity.EntityType.getKey(((org.bukkit.craftbukkit.entity.CraftEntity) entity).getHandle().getType()).toString());
++        ((org.bukkit.craftbukkit.entity.CraftEntity) entity).getHandle().saveWithoutId(compound);
```
lynxplay: This would serialise passengers of the entity serialised which breaks the current return contract.
There is a reason the current logic sets the passengers to an empty list before calling save()

SoSeDiK: Passengers are not preserved upon spawning the deserialized entity, but true, this data should probably omit those values immediately upon serialization.
Brought back the workaround 👍

Owen1212055:
The issue I see is that wouldn't this allow for the serialization of entities in an invalid state and then their state will be "reset"? Not sure if that's a good thing.

SoSeDiK:
> The issue I see is that wouldn't this allow for the serialization of entities in an invalid state and then their state will be "reset"? Not sure if that's a good thing.

Yes, it would, and that's partially the goal. Before you'd get a `byte[]` of invalid data, and now you'll have the last captured state of the entity, which can be manually "reset" (spawned) into a new valid entity.

Personally, I don't think this is a problem for API which does nothing by itself and is even placed in Unsafe.
I'd say it's a dev's responsibility if they want to take into account such cases. After all, it was them who tried to serialize a dead entity, for example (which is considered invalid too).

Owen1212055:
Well, in my opinion, if you try serializing an invalid entity I wouldn't expect it to magically be deserialized as a valid entity?
It's just tricky now since you will have to account for odd cases where "the entity will be serialized as a valid entity, despite being invalid"

SoSeDiK:
> serializing an invalid entity I wouldn't expect it to magically be deserialized as a valid entity

I think it depends. My wording might've not been the best, but I'll try to explain.

Upon deserialization, you'll get the same set of data as before serialization.
So while the entity might be considered invalid, its data is valid and shouldn't be prevented from deserialization via API.

For example, if you serialize an entity with 0 health you'll also get an entity with 0 health upon deserialization, and nothing will happen if you try to spawn it, because, well, it's "invalid".
Of course, you can change the health via API and spawn now a valid entity.

One other example of an "invalid" state is when the entity was despawned.
It's invalid because it's not present in the world (someone called `entity.remove()` or a chunk was unloaded, for instance).
This, by game's means `!entity.isValid()` state, shouldn't prevent you from serialization & deserialization & manual spawn of an entity too, because the entity's data is valid.

lynxplay:
Generally after looking at this, I think it might be worth hiding these "bypasses" behind some form of flag system similar to relative teleportation. E.g. if I as a plugin wants to save an entity that is a vehicle, it could actively define that by passing some `EntitySerialisazionCheckBypassFlag.PASSENGER` to the method to indicate that specific check should be ignored.

I do not really want to end up with completely invalid entities being saved when all I want to do is passenger saving.
Alternative would be to just allow a flag to toggle on passenger/non-persist saving and never allow non-serialisable and invalid entities to be saved (knots etc).

lynxplay:
Do you have anything interest in moving forward with this PR ?
Generally seems like a worth change in my opinion :+1:

SoSeDiK:
Yes, I do! :)
However, I'm currently not at home having exams at uni, and due to the situation in Ukraine I have lack of electricity most of the spare time. I'll revisit in about a week.

As for the feedback, I guess flags might be a neat solution 🤔
1. What else besides `PASSENGER` would there be?

My proposals are:
`MISC` (serialize misc entities like knot/lightning) / `FORCE` (serialize dead/unloaded)?

2. And as for `PASSENGER`, do you mean that it'll allow including passengers?

Because I think that not being able to save a passenger is more of a bug *in this case*, and it shouldn't be prevented. Vanilla does that bacause that entity is already serialized as part of its vehicle, so no need to save the passenger separately once again.

I believe in our case we want to serialize the passenger specifically since we don't care about the vehicle.

3. If this is the case (`PASSENGER` is for including passengers), do we want to go through all serialized entities chain (vehicle -> passengers) and spawn them upon deserializing?

3. Also, what should we do when someone tries to serialize "invalid" entity without passing the bypass flag?

Previously you'd end up with *corrupted* data. Should we have the inner checks and raise IllegalArgumentException? So, for example, if we have `!isValid()` entity and no `FORCE` flag then throw?

4. One more caveat I thought of is de/serializing Players. Should we prevent that altogether?

SoSeDiK:
Alright, rebased & introduced flags & added some javadocs. Feedback would be appreciated!

For now, this PR:
1. Inrtoduces `EntitySerializationFlag`s:
- `FORCE` - serializing invalid (dead, unloaded) entities.
	- This way you can "validate" data on Entity later after deserializing it (for example, set health, so the entity is no longed considered dead).
- `MISC` - serializing misc entities (those with `!EntityType#canSerialize` (since there's no need in vanilla to, for example, save a lightning bolt)).
- `PASSENGERS` - if present, the serialized data will include serialized passengers as well.
- `PLAYER` - if anyone for whatever reason wants to save/look at serialized player's data.
	- Vanilla always skips entity with Player's id, so deserializing such data will lead to IllegalArgumentException.

2. `UnsafeValues#serializeEntity` method now has additional checks:
- If `!entity.isValid()` then there must be the `FORCE` flag.
- If `!#canSerialize` then there must be the `MISC` flag.
- If serializing a Player then there must be the `PLAYER` flag.
- If the `PASSENGERS` flag is present then the passengers are included in the serialized data.

3. Overview of behaviours this this patch:
- Non-persistent entities can now be serialized since we do not save the data to the world folder like vanilla but rather the user manually requested the serialized data.
- For the same reason, entities that ride other entities can now be serialized too (e.g. you can serialize a skeleton riding a spider and then deserialize/spawn the skeleton only; previously you'd have a corrupted data in this and the previous case).
- Spawning invalid entity does nothing, they are rejected. The data on Entity instance must be "validated" for the entity to appear in the world.
- While it is possible to serialize Players with a flag, it's not possible to deserialize them.
- `preserveUUID` boolean controlls preserving uuid for the entity and its passengers.
- `#deserializeEntity` now accepts new `preservePassengers` boolean. If passed `true`, Entity instances will be created for passengers as well, and upon calling `#spawnAt` the entity and its passengers will be spawned together.

5. Currently, all the flags affect only the passed entity and do not apply to its passengers. So, for example, even if `FORCE` and `PASSENGERS` are passed, invalid passengers will not be present in serialized data. Also, while serializing allows valid non-persistent entities, due to the inner CraftBukkit's checks non-persistent entities will be skipped in passenger serializing. This can be worked around if needed, but will require extra work.

6. Bukkit's `EntityType` enum currently does not expose `#canSerialize` boolean. Should this be introduced (in this PR in a separate patch / different PR) or should this be left to the user to figure out? Currently, leash knots, lightning bolts, players and fishing bobbers are not serializable. Player has its own flag, and others require the misc flag.

lynxplay:
I like the flags chosen :+1: player seems a bit weird, but I guess there could be some really weird point ?
I have not yet had the time to look at the impl, but API wise, while the methods in there are deprecated, I don't think we have to break the API if not needed here ?

UnsafeValues#serialiseEntity(Entity) is now gone.
While programatically the same, this is an ABI break to the new method signature UnsafeValues#serialiseEntity(Entity, FlagType...). A simple default method that passes the previous default in regards to flags (I think none ?) would keep this working for older plugins and cross versions :+1:

I'll look at the impl of things once I make it out of the train :)

Owen1212055 (approved):

patches/api/0328-Add-Raw-Byte-Entity-Serialization.patch
```diff
@@ -3,28 +3,130 @@ From: Mariell Hoversholm <proximyst@proximyst.com>
...
++    default org.bukkit.entity.Entity deserializeEntity(byte[] data, World world, boolean preserveUUID) {
++        return deserializeEntity(data, world, preserveUUID, false);
++    }
++
++    /**
++     * Deserializes the entity from data.
++     *
++     * @param data serialized entity data
++     * @param world world
++     * @param preserveUUID whether to preserve uuid
```
lynxplay: I wonder if preserveUUID in this case should also apply to passangers.
In my opinion it very much should, as this method may be used to clone entities via the API and removing and regenerating the uuid of entities is required for said behaviour.

SoSeDiK: It currently does apply to both the entity and its passengers. I guess this could be clarified in the javadoc?
One thing I still wonder about is whether non-persistent passengers should be saved with the `FORCE` flag?

lynxplay: Yea sorry, didn't see the helper method.
Javadocs should certainly reflect it.
IMO, if possible, FORCE should probably allow passangers that are non-persistent to be stored.

Warriorrrr (approved):

SoSeDiK:
Rebased, but haven't gotten to test whether everything works yet

SoSeDiK:
Will revisit once 1.21 builds are ready.

lynxplay:
Needs some testing on my end, but does look good otherwise. Will get to on monday :+1:

PR #8498, opened 2022-10-23
Deprecate unused VehicleEntityCollisionEvent methods

Owen1212055:


Machine-Maker:
Were these methods ever used? Is there any chance they could be in the future? Just thinking if the "forRemoval" is good to have here if something could be implemented that uses them in the future.

Owen1212055:
I can't see them ever being used, it seems it was changed https://github.com/Bukkit/Bukkit/commit/3d94820557409983f72125a83087257ccc0facfa
but, I can't ever see them every used https://github.com/TacoSpigot/CraftBukkit/blob/24557bc2b37deb6a0edf497d547471832457b1dd/nms-patches/EntityMinecartAbstract.patch

But, it doesn't make much sense for them to do anything.
If you ever want to cancel an entity being "picked up" by the boat, cancel VehicleEnterEvent.
If you want to cancel the collision, cancel the VehicleEntityCollisionEvent.

electronicboy (approved):

PR #8499, opened 2022-10-23
Only limit global sounds to same world if limiting radius

e-im:
I'm not sure if this is the correct way to fix this, or if a separate config option to make the sounds per world would be warranted.

To me, it looks like the patch was initially made to fix the spigot bug that would, when limiting radius sound radius, cause players in the other dimensions at the same coordinates to also hear the sound. It could have also just been making an opinionated change, though. I couldn't find any discussion of it. https://github.com/PaperMC/Paper/commit/4becba32a6d20587ecdffe9ebbc5047a3f02590a

In vanilla boss death sounds are played to all players on the server.


kennytv (approved):
Seems reasonable to me - if people still want it to be heard in just one world but in the entire world, they can increase the radius

NoahvdAa (approved):

PR #7333, opened 2022-01-14
Add Multi Block Change API

brodybeckwith:
API to send large amounts of fake block changes in one packet per chunk section.

PR #8502, opened 2022-10-24
Deprecate Player sendMultiBlockChange

brodybeckwith:
A similar method player.sendBlockChanges was added by upstream which essentially performs the same function as player.sendMultiBlockChange that I PR'd awhile ago to Paper at https://github.com/PaperMC/Paper/pull/7333.

This PR deprecates my original method in favor of upstreams.

### DIFF
--- a/patches/api/0355-Multi-Block-Change-API.patch
+++ b/patches/api/0355-Multi-Block-Change-API.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] Multi Block Change API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index b5519cf329485a2ec72586e66a8a88617c09312e..82d9cfadb00da9b7c2034938780354a573801728 100644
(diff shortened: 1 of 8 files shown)


Owen1212055:
I argue against this. I don’t see an issue with having both api methods. And I think yours is a bit more workable with it using BlockData and locations, rather than a BlockState.

BrodyBeckwith:
Is there no way to create new BlockState via the api? If not then thats kinda surprising.

electronicboy:
Only way to get a BlockState is to get the blockstate from the world, and then mutate that. For simple blocks, I guess that's not a huge issue, but, is far from ideal when dealing with BlockEntities, which, I don't recall seeing upstream even actually dealing with, so it's not like there is a huge benefit here?

Owen1212055:
But at least, you’ll need a new block state instance for each block. While for your method, you can share the same BlockData. 

But yeah I really think both are fine here. They both tackle it kinda differently. But, let’s see what others think too.

BrodyBeckwith:
tbh to me it seems the api "sendBlockChanges" is kinda pointless lol, unless you say wanted to revert blocks to a previous "snapshot"

PR #8505, opened 2022-10-25
Configurable Region Compression Format

Astralchroma:
Adds a configuration option to save region files without any compression, this is useful for any servers which wish to use compression at the file system level.

Minecraft already has the ability to save and load region files without compression, however the ability to save isn't used, and this adds a config option to enable that. As Minecraft supports regions saved this way, the worlds saved using this option remain compatible with an unmodified client / server.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -440,10 +440,10 @@ index 0000000000000000000000000000000000000000..9ef6712c70fcd8912a79f3f61e351aac
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..58cdb70c0b9de45ca28811416d871099eefec2ce
+index 0000000000000000000000000000000000000000..5cfd0bd7c09aacd7313575e30cf0a171c597e2bd
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,283 @@
+@@ -0,0 +1,291 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
@@ -589,6 +589,14 @@ index 0000000000000000000000000000000000000000..58cdb70c0b9de45ca28811416d871099
 +        public boolean allowHeadlessPistons = false;
 +        @Comment("This setting controls if grindstones should be able to output overstacked items (such as cursed books).")
 +        public boolean allowGrindstoneOverstacking = false;
++        @Comment("This setting controls what compression format is used for region files.")
++        public CompressionFormat compressionFormat = CompressionFormat.ZLIB;
++
++        public enum CompressionFormat {
++            GZIP,
++            ZLIB,
++            NONE
++        }
 +    }
 +
 +    public Commands commands;
--- a/patches/server/0019-Rewrite-chunk-system.patch
+++ b/patches/server/0019-Rewrite-chunk-system.patch
@@ -15673,7 +15673,7 @@ index 0000000000000000000000000000000000000000..962d3cae6340fc11607b59355e291629
 +
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-index 58cdb70c0b9de45ca28811416d871099eefec2ce..6a0560b1b0972f90b5260115e81f2baa84cfb40d 100644
+index 5cfd0bd7c09aacd7313575e30cf0a171c597e2bd..115c44d6f064049270f1fae07683da1da974b08e 100644
 --- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 @@ -116,21 +116,6 @@ public class GlobalConfiguration extends ConfigurationPart {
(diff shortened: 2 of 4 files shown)


patches/server/0936-Uncompressed-Regions.patch
```diff
@@ -0,0 +1,48 @@
...
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -130,6 +130,7 @@ public class GlobalConfiguration extends ConfigurationPart {
+         public double playerMaxConcurrentLoads = 20.0;
+         public double globalMaxConcurrentLoads = 500.0;
+         public double playerMaxChunkLoadRate = -1.0;
++        public boolean saveRegionsUncompressed = false;
+     }
+ 
+     public UnsupportedSettings unsupportedSettings;
```
NoahvdAa: Config changes go in the config patch now

NoahvdAa (changes requested):

patches/server/0936-Uncompressed-Regions.patch
Astralchroma: Moved to config patch

lynxplay (commented):
Just as a general thought, wouldn't it be a bit more future proof to have this more of an enum value, rather than a boolean ?

E.g. right now this option lets you switch between deflate and none, however potentially the server offers different IO wrappers in the future here.
Instead of a boolean, this could simply default to "deflate", with the possible alternative format of "plain" or something along those lines.

patches/server/0936-Uncompressed-Regions.patch
```diff
@@ -0,0 +1,36 @@
...
+ 
+diff --git a/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java b/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java
+index 97d4ae5619dcc0922e0381b1bb45a135f514e3af..62f4592b346f2ac216cb8d7ab191c7177c831da0 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java
++++ b/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java
+@@ -30,6 +30,7 @@ public class RegionFileVersion {
+     }, (outputStream) -> {
+         return outputStream;
+     }));
++    public static final RegionFileVersion ACTIVE_VERSION = io.papermc.paper.configuration.GlobalConfiguration.get().chunkLoading.saveRegionsUncompressed ? VERSION_NONE : VERSION_DEFLATE; // Paper - Uncompressed Regions
```
Machine-Maker: Make sure that the GlobalConfiguration is loaded before this static field is initialized. Otherwise you'll want to use a Supplier here. 

Actually, its probably better to make this a Supplier regardless, just to make sure no class load order issues arise.

Astralchroma:
> Just as a general thought, wouldn't it be a bit more future proof to have this more of an enum value, rather than a boolean ?
> 
> E.g. right now this option lets you switch between deflate and none, however potentially the server offers different IO wrappers in the future here. Instead of a boolean, this could simply default to "deflate", with the possible alternative format of "plain" or something along those lines.

Minecraft already has support for 3 compression formats (gzip, zlib (default), and none), although I don't see why anyone would want to use gzip over zlib or none, but I suppose if we're adding a option for default or none, we may as well just use an enum and allow all 3, so I'll do that. Although I don't think this enum should be used by Paper to add alternative formats in the future as formats added by Paper would be incompatible with Vanilla.

olijeffers0n:
> As formats added by Paper would be incompatible with Vanilla.

I know that this is something that is technically possible with some folder moving, but moving worlds from Paper to vanilla is something that I hope people can understand may cause issues. 

If enough of a warning notice is placed on it when the user selects it, I have no problem with this. 

There is potential for a converter tool to be developed to change it back anyways, so I +1 an enum for (the possibility of) new formats as ultimately Paper is here to improve the server experience.

Astralchroma:
Updated to rebase on master (apparently I forgot to do that before), as well as use a supplier like @Machine-Maker suggested, as well as changing it to use an enum.

patches/server/0934-Configurable-region-compression-format.patch
```diff
@@ -0,0 +1,45 @@
...
+index 97d4ae5619dcc0922e0381b1bb45a135f514e3af..78eec7232e223732d415e70b47f1f4302929658c 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java
++++ b/src/main/java/net/minecraft/world/level/chunk/storage/RegionFileVersion.java
+@@ -30,6 +30,16 @@ public class RegionFileVersion {
+     }, (outputStream) -> {
+         return outputStream;
+     }));
++
++    // Paper Start - Configurable region compression format
++    public static final java.util.function.Supplier<RegionFileVersion> compressionFormatSupplier = () ->
```
Machine-Maker: Use ALL_CAPS for the field name here since it’s a static final field.

Machine-Maker (dismissed):

patches/server/0934-Configurable-region-compression-format.patch
Astralchroma: Done.

electronicboy (approved):

Owen1212055 (approved):

Peter-Crawley:
This has been rebased onto master now, any chance this could be merged soon?

Owen1212055:
This was discussed somewhere, but I think that this config option should be moved to the unsupported category. It's not officially supported by mojang and I am not sure this is something we should officially support too.

Peter-Crawley:
> This was discussed somewhere, but I think that this config option should be moved to the unsupported category. It's not officially supported by mojang and I am not sure this is something we should officially support too.

That makes sense, I've gone ahead and done that

Astralchroma:
Rebased

Astralchroma:
*the rebase required label can removed btw*

PR #8514, opened 2022-10-28
Move block farther away for Player#setRotation

Owen1212055:
This resolves issues that cause this method to work incorrectly for clients with high latency.


KabanFriends:
This causes a problem where player position on server side is not updated when you set rotation in a high frequency (client <-> server desync)
Although this is something we can't avoid when using relative teleports, LookAt packets do not cause this to happen.

Desync happens only when your latency is higher than the duration of a tick, so to test this yourself, you need to connect to a server that is hosted somewhere else, or connect through ngrok or something

Owen1212055:
This patch will be reworked, the solution in the end is to multiply the direction by some large number like 99999. This causes the issue to not be very noticeable, as since it's far away it shouldn't effect the client too much.

In general however these packets aren't handled too well in high latency environments.

Owen1212055:
image

In general, what this fix is supposed to do is make any high latency not as noticeable. In this case, we shift it so far that when the player moves it should *roughly* still be in the same area and not as visible.
Because of how it is currently when the player moves, it could cause it to shift dramatically.

On the bottom, we can see how when the player moves it could cause their entire direction to shift. While the new implementation might cause only a small shift. The goal is to make it almost not noticeable.

Machine-Maker (approved):

PR #8518, opened 2022-10-29
Call missing BlockDispenseEvents

Lulu13022002:
This event should be triggered for mud transform, anchor recharge and copper waxing

### DIFF
--- a/patches/server/0936-Call-missing-BlockDispenseEvent.patch
+++ b/patches/server/0936-Call-missing-BlockDispenseEvent.patch
@@ -5,55 +5,55 @@ Subject: [PATCH] Call missing BlockDispenseEvent
 
 
 diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
-index 9d601bf258a6d209a889afa7ba7f839c01f83cd8..f2f5b033893df21acffa0d39870022d5c2d6318c 100644
+index 9d601bf258a6d209a889afa7ba7f839c01f83cd8..d57ceb21905137f94290fe23b5ce611f5f670999 100644
 --- a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
 +++ b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
-@@ -1083,6 +1083,14 @@ public interface DispenseItemBehavior {
-                 BlockState iblockdata = worldserver.getBlockState(blockposition);
+@@ -1084,6 +1084,14 @@ public interface DispenseItemBehavior {
  
                  this.setSuccess(true);
-+
-+                // Paper start
-+                ItemStack result = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockDispenseEvent(pointer, blockposition, stack, this);
-+                if (result != null) {
-+                    return result;
-+                }
-+                // Paper end
-+
                  if (iblockdata.is(Blocks.RESPAWN_ANCHOR)) {
-                     if ((Integer) iblockdata.getValue(RespawnAnchorBlock.CHARGE) != 4) {
-                         RespawnAnchorBlock.charge(worldserver, blockposition, iblockdata);
-@@ -1104,6 +1112,14 @@ public interface DispenseItemBehavior {
-                 BlockPos blockposition = pointer.getPos().relative((Direction) pointer.getBlockState().getValue(DispenserBlock.FACING));
-                 ServerLevel worldserver = pointer.getLevel();
-                 BlockState iblockdata = worldserver.getBlockState(blockposition);
-+
-+                // Paper start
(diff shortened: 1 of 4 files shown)


Owen1212055:
Is there any reason why this.setSuccess(true); should still be true for when this event is canceled? It seems like some interactions have it false if the event is canceled. 

See line 1015, 
```java
   @Override
            public ItemStack execute(BlockSource pointer, ItemStack stack) {
                this.setSuccess(false);
                ServerLevel worldserver = pointer.getLevel();
                BlockPos blockposition = pointer.getPos().relative((Direction) pointer.getBlockState().getValue(DispenserBlock.FACING));
                BlockState iblockdata = worldserver.getBlockState(blockposition);

                // CraftBukkit start
                org.bukkit.block.Block bukkitBlock = worldserver.getWorld().getBlockAt(pointer.getPos().getX(), pointer.getPos().getY(), pointer.getPos().getZ());
                CraftItemStack craftItem = CraftItemStack.asCraftMirror(stack);

                BlockDispenseEvent event = new BlockDispenseEvent(bukkitBlock, craftItem.clone(), new org.bukkit.util.Vector(blockposition.getX(), blockposition.getY(), blockposition.getZ()));
                if (!DispenserBlock.eventFired) {
                    worldserver.getCraftServer().getPluginManager().callEvent(event);
                }

                if (event.isCancelled()) {
                    return stack;
                }
```

This will be FALSE if it's cancelled.

patches/server/0936-Call-missing-BlockDispenseEvent.patch
```diff
@@ -0,0 +1,87 @@
...
+@@ -1129,6 +1145,13 @@ public interface DispenseItemBehavior {
+                     BlockPos blockposition = pointer.getPos();
+                     BlockPos blockposition1 = pointer.getPos().relative((Direction) pointer.getBlockState().getValue(DispenserBlock.FACING));
+ 
++                    // Paper start
++                    ItemStack result = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockDispenseEvent(pointer, blockposition1, stack, this);
++                    if (result != null) {
++                        return result;
++                    }
++                    // Paper end
```
Machine-Maker: This needs to be in the else clause of the conditional below otherwise it will create duplicate events firing. BlockDispenseEvent is also fired in the defaultDispenseItemBehavior#dispense method.

patches/server/0936-Call-missing-BlockDispenseEvent.patch
```diff
@@ -0,0 +1,87 @@
...
+                 BlockPos blockposition = pointer.getPos().relative((Direction) pointer.getBlockState().getValue(DispenserBlock.FACING));
+                 ServerLevel worldserver = pointer.getLevel();
+                 BlockState iblockdata = worldserver.getBlockState(blockposition);
++
++                // Paper start
++                ItemStack result = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockDispenseEvent(pointer, blockposition, stack, this);
++                if (result != null) {
++                    return result;
++                }
++                // Paper end
```
Machine-Maker: This needs to be in the if clause of the conditional below to avoid duplicate event fires.

patches/server/0936-Call-missing-BlockDispenseEvent.patch
```diff
@@ -0,0 +1,87 @@
...
+                 BlockState iblockdata = worldserver.getBlockState(blockposition);
+ 
+                 this.setSuccess(true);
++
++                // Paper start
++                ItemStack result = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockDispenseEvent(pointer, blockposition, stack, this);
++                if (result != null) {
++                    return result;
++                }
++                // Paper end
```
Machine-Maker: This needs to be in the if clause of the conditional below to avoid duplicate event fires.

Machine-Maker (changes requested):
All 3 new event fires need to be moved to avoid adding duplicate event firings.

Machine-Maker (approved):

Owen1212055 (approved):

Machine-Maker (changes requested):
So looking at how this stuff works, we need to start introducing more consistency. In **most** of the BlockDispenseEvent the stack is the actually stack that will be dispensed. As its a single item, instead of the stack in the slot that was selected to be dispensed. So this should be changed to reflect that.

patches/server/0950-Call-missing-BlockDispenseEvent.patch
```diff
@@ -0,0 +1,87 @@
...
+ 
+                 this.setSuccess(true);
+                 if (iblockdata.is(Blocks.RESPAWN_ANCHOR)) {
++                    // Paper start
++                    ItemStack result = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockDispenseEvent(pointer, blockposition, stack, this);
++                    if (result != null) {
++                        this.setSuccess(false);
++                        return result;
++                    }
++                    // Paper end
```
Machine-Maker: I think this needs to be inside the next if statement which checks if it can even consume a glowstone.

Machine-Maker (changes requested):

Machine-Maker (approved):

PR #8524, opened 2022-10-30
Fix possible StackOverflowError for some dispenser iteractions

Machine-Maker:
```java
    @EventHandler
    public void on(BlockDispenseEvent event) {
        final ItemStack stack = event.getItem();
        stack.editMeta(meta -> {
            meta.lore(List.of(Component.text("Dispensed at: " + System.nanoTime())));
        });
        event.setItem(stack);
    }
```

This will produce an SO error for lots of equipables, like saddles on pigs, chests on horses, carpets are llamas, mob/player skulls on any mob, and carved pumpkins.

This is due to the dispense event logic not properly searching for a replacement dispenser behavior if it detects the item has changed in any way. It needs to properly handle the situation where the replacement behavior is the same as the current behavior.


Machine-Maker:
Rebased for 1.19.3

Owen1212055 (approved):
Tested

Machine-Maker:
Rebased for 1.20.4

PR #8525, opened 2022-10-30
Fix duplicate dispense events for some items

Machine-Maker:
Both wither skulls and carved pumpkins fired a BlockDispenseEvent twice if they didn't spawn a mob.

This is possibly a behavior change, but as usual, it's actually a bug fix so the behavior before was wrong.


Machine-Maker:
Rebased for 1.19.3 and removed another duplicate event for glass bottles

PR #8529, opened 2022-10-31
Fix EntityCombustEvent cancellation issues

### LINKED ISSUE #8523: EntityCombustEvent cancellation can't fully prevent entities from being set on fire
### Expected behavior

Canceling every event of type `EntityCombustEvent` should guarantee no entity ever gets set on fire.

### Observed/Actual behavior

Under conditions that might trigger an `EntityCombustEvent` on the same entity very quickly (every tick), the entity may eventually get set on fire.

### Steps/models to reproduce

For the plugin:
- Use a handler that cancels any `EntityCombustEvent`
- Use a handler that cancels any `BlockIgniteEvent` (to avoid setting fire to a block, and that block, in turn, set fire to the entity)
- (Optional) To better observe the behavior, use a handler that cancels `EntityDamageByEntityEvent` if the damager is a `LightningStrike`

In-game:
- Use an entity of your choosing (e.g.: a `Sheep`)
- In a repeating command block, summon lightning at the sheep's location `execute at @e[type=sheep,limit=1,sort=nearest] run summon minecraft:lightning_bolt`

### Plugin and Datapack List

A simple plugin that includes the aforementioned handlers.
No other plugins or datapacks.

### Paper version

This server is running Paper version git-Paper-246 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: aaf5e39)

### Other

Upon following the steps required to reproduce the issue, the sheep shall be set on fire and eventually be killed by the fire.

It's important to mention that running that command **once** will effectively leave the sheep **unharmed**. The problem arises when the lightning bolts are summoned incredibly fast, every tick to be exact.

maestro-denery:
Fixes https://github.com/PaperMC/Paper/issues/8523. If any other problem with this event will be found I also fix it there.

### DIFF
--- a/patches/server/0936-Fix-EntityCombustEvent-cancellation-cant-fully-preve.patch
+++ b/patches/server/0936-Fix-EntityCombustEvent-cancellation-cant-fully-preve.patch
@@ -6,14 +6,14 @@ Subject: [PATCH]
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
-index f2908cf61ff28bef44fcf46b15cf585e942fd7ce..148034a8fc4167ff0e77a8eb1dd205c38a91d548 100644
+index f2908cf61ff28bef44fcf46b15cf585e942fd7ce..97ef2f3a7b076805b4520bc5ca771652be3f968f 100644
 --- a/src/main/java/net/minecraft/world/entity/Entity.java
 +++ b/src/main/java/net/minecraft/world/entity/Entity.java
 @@ -3137,6 +3137,11 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
              if (!entityCombustEvent.isCancelled()) {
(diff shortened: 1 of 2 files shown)


Owen1212055 (commented):
Remove test paper plugin diff

lynxplay (changes requested):
Glad to see you opening your first PR 🥳 

If you are up for it, could you also address the issue lulu found in https://github.com/PaperMC/Paper/issues/8523#issuecomment-1297050260 ?
It follows the exact logic and would fit nicely in this patch.

Beyond that, Owen already pointed out, the test plugin should not be committed :)

maestro-denery:
@lynxplay It seems that's all, I didn't find any other similar usages of the `#getRemainingTicks()` method.

lynxplay (approved):
Looks good to me! Now both the sheep smitten every tick as well as myself dancing in fire no longer combust.

electronicboy (approved):

PR #8532, opened 2022-11-01
Handle properly the custom bukkit damage cause

Lulu13022002:
The custom bukkit damage cause MELTING and POISON replace totally the vanilla ones resulting in incompatible vanilla behavior and a non future proof way. For example in vanilla when you throw a potion of poison on a silverfish near stone/infested brick it will destroy the nearby stones, but on paper it doesn't because of the custom poison replacing the magic one (so source == DamageSource.MAGIC become false). Also i know that the shallow clone that bukkit does isn't future proof but since the DamageCause will be deprecated it's not really a problem actually.

### DIFF
--- a/patches/server/0960-Implement-properly-the-custom-bukkit-damage-cause.patch
+++ b/patches/server/0960-Implement-properly-the-custom-bukkit-damage-cause.patch
@@ -1,0 +1,69 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
+Date: Tue, 1 Nov 2022 14:59:40 +0100
+Subject: [PATCH] Implement properly the custom bukkit damage cause
+
+
+diff --git a/src/main/java/net/minecraft/world/effect/MobEffect.java b/src/main/java/net/minecraft/world/effect/MobEffect.java
+index e708b2c987fac150c22b3367cec2e3e2bcb9914c..75d8dd109705697608460ffb3c581be56f65192c 100644
+--- a/src/main/java/net/minecraft/world/effect/MobEffect.java
++++ b/src/main/java/net/minecraft/world/effect/MobEffect.java
+@@ -65,7 +65,11 @@ public class MobEffect {
+             }
+         } else if (this == MobEffects.POISON) {
+             if (entity.getHealth() > 1.0F) {
+-                entity.hurt(CraftEventFactory.POISON, 1.0F);  // CraftBukkit - DamageSource.MAGIC -> CraftEventFactory.POISON
++                // Paper start
++                entity.customDamageSource = CraftEventFactory.POISON;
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
https://user-images.githubusercontent.com/23108066/202831238-326c84c9-2bbe-4ad5-85e1-6b945097f51c.mp4
https://user-images.githubusercontent.com/23108066/202831242-330c2f47-f8e1-411c-b2dd-903a2db2f2e6.mp4

Can confirm that there is a behavior change here! Vanilla vs Paper

This PR fixes this behavior.

patches/server/0937-Implement-properly-the-custom-bukkit-damage-cause.patch
```diff
@@ -0,0 +1,65 @@
...
+--- a/src/main/java/net/minecraft/world/entity/animal/SnowGolem.java
++++ b/src/main/java/net/minecraft/world/entity/animal/SnowGolem.java
+@@ -106,7 +106,9 @@ public class SnowGolem extends AbstractGolem implements Shearable, RangedAttackM
+             Biome biomebase = (Biome) this.level.getBiome(blockposition).value();
+ 
+             if (biomebase.shouldSnowGolemBurn(blockposition)) {
+-                this.hurt(CraftEventFactory.MELTING, 1.0F); // CraftBukkit - DamageSource.BURN -> CraftEventFactory.MELTING
++                this.customDamageSource = CraftEventFactory.MELTING; // Paper
++                this.hurt(DamageSource.ON_FIRE, 1.0F); // CraftBukkit - DamageSource.BURN -> CraftEventFactory.MELTING // Paper revert
++                this.customDamageSource = null; // Paper
```
Owen1212055: Paper start

Paper end

patches/server/0937-Implement-properly-the-custom-bukkit-damage-cause.patch
```diff
@@ -0,0 +1,65 @@
...
+--- a/src/main/java/net/minecraft/world/effect/MobEffect.java
++++ b/src/main/java/net/minecraft/world/effect/MobEffect.java
+@@ -65,7 +65,9 @@ public class MobEffect {
+             }
+         } else if (this == MobEffects.POISON) {
+             if (entity.getHealth() > 1.0F) {
+-                entity.hurt(CraftEventFactory.POISON, 1.0F);  // CraftBukkit - DamageSource.MAGIC -> CraftEventFactory.POISON
++                entity.customDamageSource = CraftEventFactory.POISON; // Paper
++                entity.hurt(DamageSource.MAGIC, 1.0F);  // CraftBukkit - DamageSource.MAGIC -> CraftEventFactory.POISON // Paper revert
++                entity.customDamageSource = null; // Paper
```
Owen1212055: Paper start

Paper end

olijeffers0n (approved):

Lulu13022002:
I'm closing this since mojang has fixed the issue instead of spigot :) with the rework of the damagesource

PR #8534, opened 2022-11-02
Fix rendering for translatable death messages

PrimordialMoros:
When using a `TranslatableComponent` in `PlayerDeathEvent#deathMessage` the client should view the rendered translation of the component (assuming the key is registered in the GlobalTranslator).

Currently, the component in `PlayerDeathEvent#deathMessage` is converted to a vanilla component and when written to a chat packet, the GlobalTranslator is not utilized (client will only translate keys found in its ResourcePack).

That results in the client only seeing the translation key.

Since this is related to Adventure, I figured I'd use the existing patch, let me know if it needs to be in a separate one.

Here's a plugin to test.
```java
    @Override
    public void onEnable() {
        TranslationRegistry registry = TranslationRegistry.create(Key.key("test", "translations"));
        registry.defaultLocale(Locale.ENGLISH);
        registry.register("test.death", Locale.ENGLISH, new MessageFormat("Test translation: {0} died"));
        GlobalTranslator.translator().addSource(registry);
    }

    @EventHandler
    public void onDeath(PlayerDeathEvent event) {
        event.deathMessage(Component.translatable("test.death").args(event.getEntity().name()));
    }
}
```

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -1911,16 +1911,17 @@ index 1fb62779527a228f748b49a4d2ddfc57ccb80cf8..bd808eb312ade7122973a47f4b965058
      }
  
 diff --git a/src/main/java/net/minecraft/network/protocol/game/ClientboundSystemChatPacket.java b/src/main/java/net/minecraft/network/protocol/game/ClientboundSystemChatPacket.java
-index fd5cdc33c4cc8f24265b450621a13d3ab03644c2..b0052398d7ff954c2a7943ea4618d26f45d701da 100644
+index fd5cdc33c4cc8f24265b450621a13d3ab03644c2..55e21c7b13826f60e3c656f76e1507e0242e0af3 100644
 --- a/src/main/java/net/minecraft/network/protocol/game/ClientboundSystemChatPacket.java
 +++ b/src/main/java/net/minecraft/network/protocol/game/ClientboundSystemChatPacket.java
-@@ -6,16 +6,25 @@ import net.minecraft.network.chat.Component;
+@@ -6,24 +6,42 @@ import net.minecraft.network.chat.Component;
  import net.minecraft.network.protocol.Packet;
  
  // Spigot start
 -public record ClientboundSystemChatPacket(String content, boolean overlay) implements Packet<ClientGamePacketListener> {
 +public record ClientboundSystemChatPacket(@javax.annotation.Nullable net.kyori.adventure.text.Component adventure$content, @javax.annotation.Nullable String content, boolean overlay) implements Packet<ClientGamePacketListener> { // Paper - Adventure
  
++    @io.papermc.paper.annotation.DoNotUse // Paper - No locale context
      public ClientboundSystemChatPacket(Component content, boolean overlay) {
 -        this(Component.Serializer.toJson(content), overlay);
 +        this(null, Component.Serializer.toJson(content), overlay); // Paper - Adventure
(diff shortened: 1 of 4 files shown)


Machine-Maker:
This is not the right fix. Instead, any use of the ClientboundSystemChatPacket constructor that just takes an nms Component and a boolean should instead use the main ctor, passing in a json string from PaperAdventure.asJsonString(). Then that ctor should be marked as `@DoNotUse` as there is no way to get the locale context in the packet ctor.

PrimordialMoros:
There were 3 more uses of the nms component ctor but it's all vanilla translations. If needed I can change those too but figured I'd keep the diff small for now.

Owen1212055 (changes requested):
Logic wise, looks good.

Now because of the new DoNotUse Annotation added you have to make sure to replace the usages of those methods. Go ahead. The point of this method is to be able to see future method calls added in order to properly prevent bugs like this in the future.
image
```
Method net/minecraft/network/protocol/game/ClientboundSystemChatPacket#<init>(Lnet/minecraft/network/FriendlyByteBuf;)V makes call to bad method net/minecraft/network/protocol/game/ClientboundSystemChatPacket#<init>(Lnet/minecraft/network/chat/Component;Z)V
Method net/minecraft/server/network/ServerGamePacketListenerImpl#tryHandleChat(Ljava/lang/String;Ljava/time/Instant;Lnet/minecraft/network/chat/LastSeenMessages$Update;)Z makes call to bad method net/minecraft/network/protocol/game/ClientboundSystemChatPacket#<init>(Lnet/minecraft/network/chat/Component;Z)V
Method net/minecraft/server/network/ServerGamePacketListenerImpl#broadcastChatMessage(Lnet/minecraft/network/chat/PlayerChatMessage;)V makes call to bad method net/minecraft/network/protocol/game/ClientboundSystemChatPacket#<init>(Lnet/minecraft/network/chat/Component;Z)V
Method net/minecraft/server/level/ServerPlayer#lambda$sendSystemMessage$9(Lnet/minecraft/network/chat/Component;)Lnet/minecraft/network/protocol/Packet; makes call to bad method net/minecraft/network/protocol/game/ClientboundSystemChatPacket#<init>(Lnet/minecraft/network/chat/Component;Z)V
```

PrimordialMoros (commented):
Changed as requested.

Owen1212055 (approved):

Machine-Maker (changes requested):
The vanilla components in the places which previous used the ctor now marked as `DoNotUse` will not use the adventure global renderer. PaperAdventure#asJsonString(nms.Component, Locale) just skips the locale if the component is a plain nms Component. So instead of that, just wrap the vanilla components in PaperAdventure.asAdventure() or convert the whole vanilla component to use the adventure component types.

PrimordialMoros:
Is that something that should be supported? If someone wanted to edit vanilla translation they should be using the resource pack and not the global renderer.

Machine-Maker:
I think it should be yes.

The goal of the global renderer is to NOT have to use resource packs. You can already use resource packs with custom translations, you don’t have to only use vanilla ones in resource packs. So if you are arguing for the use of them, you should also argue that the global renderer is useless “cause why not just use resource packs for them”.

PrimordialMoros (commented):
Rebased and converted components to adventure so that GlobalTranslator can be used to override vanilla death messages,

PrimordialMoros:
Rebased for 1.19.3

Machine-Maker (approved):

Owen1212055:
Thank you for the contribution, apologies for the wait. 😦

PR #8539, opened 2022-11-04
[ci-skip] Rename Paper stop to Paper end.

astro-angelfish:
Uh My friend had just found `// Paper end` in the file to be modified is just mistyped to `// Paper stop`.
It seems that it is a standard for patched minecraft server and `// Paper stop` just broke this standard.
I have retyped it back.
Not intensively tested on my computer. I don't think changing a comment would break anything.


maestro-denery:
I think you should add [ci-skip] or something like this so if this gets merged github actions won't build again the previous commit with the changed comment

electronicboy (approved):

PR #8541, opened 2022-11-04
Fix Advancement#getDisplay() api break

Machine-Maker:
Doesn't handle invokedynamic via lambdas, but neither does the rest of Commodore fixes. But I can add that if we think it should.


jpenilla:
for now I think it's fine that it doesn't handle invokedynamic for lambdas, if someone reports it being a problem we can re asses

jpenilla (approved):

PR #8543, opened 2022-11-04
Fix OfflinePlayer#getPlayerProfile returning deprecated type

Machine-Maker:
OfflinePlayer has a method, getPlayerProfile, that returns the deprecated bukkit PlayerProfile type. This changes that to use our interface which is a subtype of bukkit's. Additionally, to maintain API compatibility, this rewrites plugins still expecting the old type.

### DIFF
--- a/patches/server/0266-Hook-into-CB-plugin-rewrites.patch
+++ b/patches/server/0266-Hook-into-CB-plugin-rewrites.patch
@@ -8,7 +8,7 @@ our own relocation. Also lets us rewrite NMS calls for when we're
 debugging in an IDE pre-relocate.
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/util/Commodore.java b/src/main/java/org/bukkit/craftbukkit/util/Commodore.java
-index a679133a3de51e26eb19932ece9ade292879aefd..2c537a80951ddf98d3135ca19a34a238484c2d88 100644
+index f5cf40baaad1b055755b6c1e18452a4afcd2dfba..3c79c0ddb9bc9024a6f74d0aef273c65bc6f3228 100644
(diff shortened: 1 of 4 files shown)


TheFruxz:
Maybe a .paper() method inside the Bukkit PlayerProfile can also improve the upgrading experience to paper's solution?

electronicboy:
We've already hardline no'd a `paper()` method, it just creates so much pointless call cludge to avoid the odd cases where upstream does something which screws us over, and just creates a messy dev experience

patches/server/0266-Hook-into-CB-plugin-rewrites.patch
```diff
@@ -167,18 +167,15 @@ index a679133a3de51e26eb19932ece9ade292879aefd..89a32de0a3fbb7465b078b2aa1cc1d15
...
                          {
                              switch ( name )
-@@ -306,6 +420,14 @@ public class Commodore
-                             return;
+@@ -307,6 +421,11 @@ public class Commodore
                          }
  
-+                        // Paper start - Rewrite plugins
+                         // Paper start - Rewrite plugins
 +                        owner = getOriginalOrRewrite( owner) ;
```
TheFruxz, line 173: Is this spacing intentional?

TheFruxz: After looking again at it, it seems to be intentional

electronicboy: partially intentional due to the existing code formatting in that class; somebody needs to pick a format though in their patch ;P

electronicboy (approved):

PR #8546, opened 2022-11-05
Fix preemtive connection dropping during shutdown process.

### LINKED ISSUE #8460: Server reacts badly to any packets from client-side after shutdown process begins, preemptively dropping connections
### Expected behavior

In the shutdown process, online players should get disconnected exactly when "Notch lost connection: Server closed" appears in console log, not at any point in time before that.

Because of the nature of the early disconnect, proxies (such as BungeeCord) aren't able to handle the disconnect as a server kick, and disconnect the player from the whole network.

### Observed/Actual behavior

Players get disconnected almost immediately after the shutdown process begins, and kick messages aren't actually kicking anyone, since players are already disconnected.
This is apparently caused by client-side packets that are sent to the server and lead to this issue being reproduced: https://bugs.mojang.com/browse/MC-142590

### Steps/models to reproduce

For vanilla:
1) Fly up, begin falling down
2) Type /stop while falling down
3) Get "Disconnected", instead of "Server closed"
Custom:
1) Using any plugin, delay full shutdown for a couple seconds (like using Thread.sleep in onDisable)
2) Observe time difference between players being dropped with "Disconnected" and actual kicks in console for "Server closed"

### Plugin and Datapack List

Doesn't matter

### Paper version

[07:00:05 INFO]: This server is running Paper version git-Paper-199 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: 77a50b9)
You are 7 version(s) behind

### Other

Link to the discord talk: https://discord.com/channels/289587909051416579/289587909051416579/1029582828243005520
Video showcasing the problem: https://cdn.discordapp.com/attachments/289587909051416579/1029584461668548728/paper_shutdown.mp4

maestro-denery:
Fixes https://github.com/PaperMC/Paper/issues/8460. 

The issue is caused by the `PacketUtils#ensureRunningOnSameThread()` checks in the `ServerGamePacketListenerImpl`. The implementation of `BlockableEventLoop` (which is `MinecraftServer`) inherits `BlockableEventLoop#executeIfPossible(Runnable)` method which is:
```java
    @Override
    public void executeIfPossible(Runnable runnable) {
        if (this.isStopped()) {
            throw new RejectedExecutionException("Server already shutting down");
        } else {
            super.executeIfPossible(runnable);
        }
    }
``` 
The exception is catched in the `Connection#channelRead0(ChannelHandlerContext, Packet)` method and it looks like this:
```java
           try {
                Connection.genericsFtw(packet, this.packetListener);
            } catch (RunningOnDifferentThreadException cancelledpackethandleexception) {
                ;
            } catch (RejectedExecutionException rejectedexecutionexception) { 
                // disconnects player if something is handled in the `ServerGamePacketListenerImpl`
                this.disconnect(Component.translatable("multiplayer.disconnect.server_shutdown"));
            } catch (ClassCastException classcastexception) {
                Connection.LOGGER.error("Received {} that couldn't be processed", packet.getClass(), classcastexception);
                this.disconnect(Component.translatable("multiplayer.disconnect.invalid_packet"));
            }
```
After the testing (Falling and calling `/stop` command) it worked fine and I got the `Stopping the server` disconnect reason.
```
[01:51:48 INFO]: Player107 issued server command: /stop
[01:51:48 INFO]: [Player107: Stopping the server]
[01:51:48 INFO]: Stopping server
[01:51:48 INFO]: [Paper-Test-Plugin] Disabling Paper-Test-Plugin v1.0.0-SNAPSHOT
[01:51:49 INFO]: Saving players
[01:51:50 INFO]: Player107 lost connection: Server closed
[01:51:50 INFO]: Player107 left the game
[01:51:50 INFO]: Saving worlds
```
(Paper test plugin used `Thread#sleep` in its `#onDisable()` method to make it even clear.)

The remaining questions are:
1. Why did Mojang do that check? Maybe there are cases when it is needed.
2. Should we really yeet the usage of `BlockableEventLoop#executeIfpossible(Runnable)` method? It is the only usage of this method.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -53,4 +53,5 @@ Nick Hensel <nickhensel25@icloud.com>
 vytskalt <vytskalt@protonmail.com>
 TheFruxz <cedricspitzer@outlook.de>
 Kieran Wallbanks <kieran.wallbanks@gmail.com>
+Denery <dorofeevij@gmail.com>
 ```
--- a/patches/server/0949-Fix-player-kick-on-shutdown.patch
+++ b/patches/server/0949-Fix-player-kick-on-shutdown.patch
@@ -1,0 +1,23 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Denery <dorofeevij@gmail.com>
+Date: Sun, 6 Nov 2022 02:02:46 +0300
+Subject: [PATCH] Fix player kick on shutdown
+
+Fix preemptive player kick on a server shutdown.
+If you update minecraft version / upstream and something is changed in this method make sure that a server doesn't disconnect a player preemptively,
+also check if all packets are ignored during the shutdown process.
+See net.minecraft.network.Connection#channelRead0(ChannelHandlerContext, Packet) and net.minecraft.util.thread.BlockableEventLoop#executeIfPossible(Runnable)
+
+diff --git a/src/main/java/net/minecraft/network/protocol/PacketUtils.java b/src/main/java/net/minecraft/network/protocol/PacketUtils.java
+index 8bc0cb9ad5bb4e76d962ff54305e2c08e279a17b..edefab4e22ee7217e17eb036165ce58788d4c782 100644
+--- a/src/main/java/net/minecraft/network/protocol/PacketUtils.java
++++ b/src/main/java/net/minecraft/network/protocol/PacketUtils.java
+@@ -41,7 +41,7 @@ public class PacketUtils {
+ 
+     public static <T extends PacketListener> void ensureRunningOnSameThread(Packet<T> packet, T listener, BlockableEventLoop<?> engine) throws RunningOnDifferentThreadException {
+         if (!engine.isSameThread()) {
+-            engine.executeIfPossible(() -> {
++            engine.execute(() -> { // Paper - Fix preemptive player kick on a server shutdown. 
+                 packetProcessing.push(listener); // Paper - detailed watchdog information
+                 try { // Paper - detailed watchdog information
+                 if (MinecraftServer.getServer().hasStopped() || (listener instanceof ServerGamePacketListenerImpl && ((ServerGamePacketListenerImpl) listener).processedDisconnect)) return; // CraftBukkit, MC-142590


Machine-Maker:
So this just essentially changes the thrown error behavior to a NOOP behavior. There is a check added by craftbukkit in side the runnable in `ensureRunningOnSameThread` that just returns out if the server is stopped. I wonder if that was added at some point in the past when it didn't throw an exception?

EDIT: Actually its exactly that issue. The linked issue here mentions MC-142590 which is also directly referenced by craftbukkit with it's fix (that seems to no longer work).

maestro-denery:
This change seems totally fine? In the last commit I added more description just in case something will change there in the future. (mojang fix / upstream fix) I also debugged it a little bit, the `ensureRunningOnSameThread` method basically is needed to handle packets when netty reads them in Epoll threads. Netty calls the `channelRead0` method from the Epoll threads which has all of that exception catch stuff, hence `ensureRunningOnSameThread` is called in Epoll threads too and it inserts runnable into minecraft server's `BlockableEventLoop` which again uses `ensureRunningOnSameThread` but now from the main thread, so a packet handles succesfully. Nothing wrong should be there...

Owen1212055 (approved):
Can confirm that this does indeed fix the issue, but feedback from people who are much more knowledgeable on this would be appreciated. 👍

electronicboy (approved):

maestro-denery:
rebased

patches/server/0936-Fix-player-kick-on-shutdown.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/net/minecraft/network/protocol/PacketUtils.java b/src/main/java/net/minecraft/network/protocol/PacketUtils.java
+index 8bc0cb9ad5bb4e76d962ff54305e2c08e279a17b..2fe968dfe0d892c37db570b8f8e16aec3dba4ce1 100644
+--- a/src/main/java/net/minecraft/network/protocol/PacketUtils.java
++++ b/src/main/java/net/minecraft/network/protocol/PacketUtils.java
+@@ -41,7 +41,7 @@ public class PacketUtils {
+ 
+     public static <T extends PacketListener> void ensureRunningOnSameThread(Packet<T> packet, T listener, BlockableEventLoop<?> engine) throws RunningOnDifferentThreadException {
+         if (!engine.isSameThread()) {
+-            engine.executeIfPossible(() -> {
++            engine.execute(() -> { // Paper - Fix preemptive player kick on a server shutdown. If you update minecraft version / upstream and something is changed in this method make sure that a server doesn't disconnect a player preemptively, also check if all packets are ignored during the shutdown process. See net.minecraft.network.Connection#channelRead0(ChannelHandlerContext, Packet) and net.minecraft.util.thread.BlockableEventLoop#executeIfPossible(Runnable)
```
MiniDigger: I feel like that long as comment should go into the patch header? what do others think?

you also seem to have messaged up the rebase slighly, there are now two versions of the patch :)

(both are fine and can be fixed on merge, just noting for awareness)

PR #8555, opened 2022-11-11
Keep vanilla check for custom growth modifier for sugar cane and cactus

### LINKED ISSUE #8544: Modifying cactus growth modifier causes it to not break after growing
### Expected behavior

### Observed/Actual behavior

### Steps/models to reproduce

as above

### Plugin and Datapack List

no plugin

### Paper version

### Other

_No response_

Lulu13022002:
Closes #8544 

The spigot growth modifier is broken for both sugar cane and cactus and the mentionned issue also happens for
sugar cane but it's more complex to reproduce (basically destroy the water source once placed and observe (vanilla will break the plant but not spigot one)
The old bahavior can possibly run two roll that also break the percent modifier

The new system handle correctly the percent in four stage
When modifier = 100: keep vanilla behavior
When modifier < 100: gives less chance to the plant to increase its own age
When modifer > 100 but < 1600: increase the age by more than one depending on the percent
When modifier >= 1600: fast grow, the age will not increase but they can grow directly even for an age smaller than 15
but to keep this consistent with vanilla we need to include the grow check in the same tick and not one tick after like usually

### DIFF
--- a/patches/server/0763-Fix-Spigot-growth-modifiers.patch
+++ b/patches/server/0763-Fix-Spigot-growth-modifiers.patch
@@ -10,6 +10,40 @@ Also fix above-mentioned modifiers from having the reverse effect
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 Co-authored-by: Noah van der Aa <ndvdaa@gmail.com>
 
+diff --git a/src/main/java/net/minecraft/world/level/block/CactusBlock.java b/src/main/java/net/minecraft/world/level/block/CactusBlock.java
+index 1ec242205b82a5a1f10deb2312795cc5dc157a76..359172280fc5d8826fc23a7d9c030da0a9d146cd 100644
+--- a/src/main/java/net/minecraft/world/level/block/CactusBlock.java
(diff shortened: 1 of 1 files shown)


lynxplay (commented):
I honestly think the entire concept of directly growing a cactus that is not age 15 is rather flawed.
I have a patch in the works for this, I'll open a PR for it this weekend so if you are fine with it I might supersede your PR.

See: https://github.com/PaperMC/Paper/commit/e65b7ab469c6e4c2ef3764678318fac4e638cf32

Lulu13022002:
Yes, you're right but even growing the cactus from more than one is strange because then it will ignore some stage that will never be used, but then we can't support the greater modifier

lynxplay:
Yea, I mean if we had to actually get through every stage, we'd had to *reschedule* random ticks ?
But that tbh is a worth while thing to allow for the modifiers.
I'd be very happy if you could chime in on the PR I'll open this weekend.

Especially in regards to the then unused stages. We could look into a gaussian distribution for the age growth instead of what my patch suggests rn to cover all ages.

PR #8561, opened 2022-11-12
Strip coordinates from lodestone compasses

NoahvdAa:
Stops the coordinates of a lodestone compass from being sent to other players, preventing possible base leaks.

Since the other client doesn't know the position, it'll show the compass in the 'spinning' state, similar to what happens when you use a compass in the nether, as shown in the gif below. Without hide itemmeta enabled the compass just points to the same location as on the screen of the user holding the compass.

[Click to view gif, it's too large for GH to proxy](https://i.noah.pm/xWg6hF.gif)

Another possible solution would be to make the compass point in the general direction of the location, similar to what craftbukkit does for far-away sounds, but this would not only require re-sending the compass quite often, but it would also allow users to triangulate the position the compass is pointing towards.

### DIFF
--- a/patches/server/0004-Paper-config-files.patch
+++ b/patches/server/0004-Paper-config-files.patch
@@ -1435,10 +1435,10 @@ index 0000000000000000000000000000000000000000..1bb16fc7598cd53e822d84b69d6a9727
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..a1bd848bbf924267e74e61dabdb840628712b1ad
+index 0000000000000000000000000000000000000000..7e995bb83a06e302fd70a8e6f079e3ac55c60473
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
-@@ -0,0 +1,467 @@
+@@ -0,0 +1,468 @@
 +package io.papermc.paper.configuration;
 +
 +import com.google.common.collect.HashBasedTable;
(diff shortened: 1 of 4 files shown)


Chris6ix:
Hey :)
If I understand this right, then this makes it so that only the player that holds the Lodestone Compass can see where it points to?
If that is so, then it would be nice if you could please add a config option to disable/enable this.

NoahvdAa:
Yes, it would do that. This is already behind a config option, [hide-itemmeta](https://docs.papermc.io/paper/reference/world-configuration#hide-itemmeta).

Machine-Maker:
This one might deserve a separate config option, like durability. None of the other stripped info changes the visual look to a vanilla default client. But this one does.

Owen1212055 (approved):

Owen1212055:
The general logic looks good, but yeah, maybe a separate config option?

NoahvdAa:
Added a config option and rebased.

olijeffers0n (approved):
I'm not sure how "exact" the compass actually is for tracking a base, but I see no harm in adding this as an option as there is a use-case for many people

Owen1212055:
I mean, you’d be able to get exact coordinates from the item NBT?

olijeffers0n:
Oh lol, I was just thinking about looking at the direction of the arrow. Yeah no - this PR has more merit then

Chris6ix:
Just don't like the last sentence of the config description, as it sounds like It's not sure if it changes the looks of an item, while it's certain that it changes. Also, it makes it sound like it's just some minor tiny detail that changes, while in reality it makes the compass look like it's broken for other people.
Instead of `This may affect small details about how an item looks when held by another player.` it should say `This affects how an item looks when held by another player.` imo.

NoahvdAa:
Thanks, I've updated the docs text

MiniDigger (approved):
Assuming there is a docs PR later, this looks like a nice addition

olijeffers0n:
Already Is:
<https://github.com/PaperMC/docs/pull/113>

PR #8564, opened 2022-11-13
Properly grow plants with growth modifier

### LINKED ISSUE #8544: Modifying cactus growth modifier causes it to not break after growing

lynxplay:
Spigots implementation of the growth modifier boils down to a plain chance for a plant to either grow immediately or to grow a single age value towards its fully grown age.

This behaviour is specifically troubling for plants like the cactus or the sugar cane, as they change an unrelated block when growing up (the free air block above their tip). This blocks validity is checked during neighbour updated which are triggered by resetting the previous tip of the plant back to [age=0].

With spigots modifier implementation, blocks at the age of zero are able to grow a new block above them however, causing the block change of the previous tip to change a block of age zero to a block of age zero, which does **not** cause a neighbour update as nothing actually changed. This means that the about block is not updated and hence may remain in an otherwise invalid state. (E.g. a cactus block next to a solid block).

This commit rewrites the existing spigot growth modifier implementation specifically for the plants that grow new blocks like the cactus or the sugar can block.

The modifier provided for the plants is separated into two distinct values. One being the remains after a floored division by 100, e.g. the original modifier without its 10^0 and 10^1 digits, the other the modifier modulo 100.
The modifier 475 would, for example, be split into a base growth of 4 and a percentage chance of an additional age progression of 75%. This way, on average, the plant would grow 4.75 times as fast as a normal plant, which would only grow a single age per random tick.

Resolves: #8544

### DIFF
--- a/patches/server/0719-Fix-Spigot-growth-modifiers.patch
+++ b/patches/server/0719-Fix-Spigot-growth-modifiers.patch
@@ -12,6 +12,37 @@ Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 Co-authored-by: Noah van der Aa <ndvdaa@gmail.com>
 Co-authored-by: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
 
+diff --git a/src/main/java/net/minecraft/world/level/block/CactusBlock.java b/src/main/java/net/minecraft/world/level/block/CactusBlock.java
+index 0003fb51ae3a6575575e10b4c86719f3061e2577..08047eb3abad49e388c99520146e3aa8e2620da6 100644
+--- a/src/main/java/net/minecraft/world/level/block/CactusBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/CactusBlock.java
+@@ -57,15 +57,22 @@ public class CactusBlock extends Block {
+             if (i < world.paperConfig().maxGrowthHeight.cactus) { // Paper - Configurable growth height
+                 int j = (Integer) state.getValue(CactusBlock.AGE);
+ 
+-                int modifier = world.spigotConfig.cactusModifier; // Spigot - SPIGOT-7159: Better modifier resolution
+-                if (j >= 15 || (modifier != 100 && random.nextFloat() < (modifier / (100.0f * 16)))) { // Spigot - SPIGOT-7159: Better modifier resolution
++                // Paper start - fix block creating growth modifiers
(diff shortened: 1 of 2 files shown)


lynxplay:
I am opening this as a draft pull request as I'd like more input here before I finalise the system (and hence also ship it to the sugar cane implementation).

As @Lulu13022002 pointed out in #8555, while this system works, it will *always* grow the plant by the base growth **at least**.
A plant with a modifier of 350 would hence *always* at least grow 3 age steps per random tick and in 50% of the random ticks it would grow 4 age steps per random tick.

This change however means that certain age values are never met in the current implementation. E.g. age 1 and 2 are always completely skipped. So is age 5.
We could change this implementation hence to potentially instead use a gaussian or normal distribution, however that also comes with some caveats:
1. We need to decide on the limits for such a random distribution. E.g. a modifier of 300 should imply an average growth of 3 age levels per random tick. What are the limits of the Gaussian ? At best obviously the average itself, the plant cannot grown a negative age, but it is still a question to consider.
2. This would most certainly require a specific exemption for the default modifier of 100, as it (at least in my opinion) should act as a form of NOOP, always leading to one (and only one) age grown per tick.

Feedback would be greatly appreciated :)

Lulu13022002:
Sorry for the late reply
I don't think we need to overthink this and imo this impl should be straightforward it's just spigot that has made a unwise decision in the first place. But i really like your solution and think that just a note on the paperdoc about what we change for theses config is enough and say that even if we fixed spigot it can still break niche plugin/command/datapack/resource pack for map maker purpose i guess when the percent overflow 100 percent since it skip some natural step.
There potentially some similar issue for the beetroot and crops where the percent change the moisture rate and so is not independent like the other. The beetroot has even a bigger issue since its class override the parent method with a 1/3 chance.

thiagogebrimm:
This fix would be great, I'm waiting

SwartZCoding:
Maybe up this and PR we waiting since 2 years lol

DR4GONSTEAR:
I'm still waiting for this. If it's not going to become master, is there a plugin that can achieve this functionality? Is there a way to hack this into your own private server? I really don't want to move everything over to fabric.

thiagogebrimm:
Any eta to add this fix?

lynxplay:
Closed in favour of https://github.com/PaperMC/Paper/pull/12264

PR #8572, opened 2022-11-19
Add No Damage Tick Utilities

Owen1212055:
Generally, this API is a bit misleading, 
```java
if ((float) this.invulnerableTime > (float) this.invulnerableDuration / 2.0F) { // CraftBukkit - restore use of maxNoDamageTicks
```

The invulnerability duration is only considered valid until it reaches the invulnerability duration / 2. 
Therefore, the no damage tick actions don't work as expected.
When setting it to 20 ticks, the entity will only have an actual invulnerability time frame of 10 ticks due to this invulnerability duration.

This adds new methods, which calculate the limit on top of your specified value in order to make it so when you do ``setExactNoDamageTicks(20)``, the entity will actually have 20 ticks and not 10 ticks of invulnerability. 

Old methods:
setNoDamageTicks(10)
getNoDamageTicks() -> 10
(getExactNoDamageTicks() will return 0) 
Behavior: The entity will have no invulnerability ticks

New methods:
setExactNoDamageTicks(10)
getExactNoDamageTicks() -> 10
(getNoDamageTicks() will return 20) 
Behavior: The entity will have 10 invulnerability ticks

This can probably be expanded to the invulnerability duration too, as it's currently divided by 2. But, it's not something I see useful enough and in general could be PRed by someone else.

---
Download the paperclip jar for this pull request: [paper-8572.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/449215361.zip)

### DIFF
--- a/patches/api/0407-Add-No-Damage-Tick-Utilities.patch
+++ b/patches/api/0407-Add-No-Damage-Tick-Utilities.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Fri, 18 Nov 2022 21:42:10 -0500
+Subject: [PATCH] Add No Damage Tick Utilities
+
+
+diff --git a/src/main/java/org/bukkit/entity/LivingEntity.java b/src/main/java/org/bukkit/entity/LivingEntity.java
+index 3bd1d100d0c481ae7edaa251869640ab370aeb42..a0a36c7a44ee33e0c532587e0e40d7b861c4c922 100644
+--- a/src/main/java/org/bukkit/entity/LivingEntity.java
++++ b/src/main/java/org/bukkit/entity/LivingEntity.java
+@@ -405,6 +405,7 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
+     /**
+      * Returns the living entity's current no damage ticks.
+      *
++     * @see LivingEntity#getExactNoDamageTicks()
+      * @return amount of no damage ticks
+      */
+     public int getNoDamageTicks();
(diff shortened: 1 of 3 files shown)


Machine-Maker (approved):

patches/api/0407-Add-No-Damage-Tick-Utilities.patch
```diff
@@ -0,0 +1,58 @@
...
++     * Gets exactly how many ticks this entity will be
++     * invulnerable with its current no damage ticks.
++     * <p>
++     * Note: The difference between this and {@link LivingEntity#getNoDamageTicks()} is that this method
++     * will correctly calculate and account for this entity's current maximum damage ticks in order get
++     * the exact value of how many ticks the current entity will be invulnerable.
++     * @return exact tick count
++     */
++    default int getExactNoDamageTicks() {
++        return Math.max(0, this.getNoDamageTicks() - (this.getMaximumNoDamageTicks() / 2));
```
lynxplay: I wonder if we want to expose this logic in the API, concerning we are making use of an internal "fact" that might change in the future.

Wouldn't internal implementation plus a `// Paper - diff on change - getExactNoDamageTIcksAPI`  make more sense as this is a craftbukkit/upstream implemented change:

```patch
-            if ((float) this.invulnerableTime > 10.0F) {
+            if ((float) this.invulnerableTime > (float) this.invulnerableDuration / 2.0F) { // CraftBukkit - restore use of maxNoDamageTicks
```

Owen1212055: Certainly doable, and was thinking of something like that. Opinions?

Owen1212055:
Thinking about this again, should these really be separate methods, or should we fix the current API? As, for the current behavior, I am not sure what use cases would even require older behavior like that. Because in general, it doesn't make sense.

PR #8576, opened 2022-11-19
Avoid a cubed max search distance for POIs

Lulu13022002:
Redo of #7830 
The max search distance used to get the closest Poi data on X/Z axis is cubed instead of squared (one time in the parameter and another time in the function) for almost all search.
Generally this has been hidden by another check that already do a pre distance check between the poi pos and the source pos for individual component but the issue still happens for diagonal distance search.
Discovered by Samsuik

### DIFF
--- a/patches/server/0709-Optimise-general-POI-access.patch
+++ b/patches/server/0709-Optimise-general-POI-access.patch
@@ -32,10 +32,10 @@ had to be specifically modified.
 
 diff --git a/src/main/java/io/papermc/paper/util/PoiAccess.java b/src/main/java/io/papermc/paper/util/PoiAccess.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..966c08e80f26c283c28292c9d34fc5e9ffa7e76f
+index 0000000000000000000000000000000000000000..05640f5f70e81833530e8098d30c400fed7ba6e1
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/util/PoiAccess.java
-@@ -0,0 +1,801 @@
+@@ -0,0 +1,800 @@
 +package io.papermc.paper.util;
 +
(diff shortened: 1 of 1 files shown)


Owen1212055 (changes requested):
There seems to be another group of methods which require the same fixing.

See AquirePoi#start
```java
io.papermc.paper.util.PoiAccess.findNearestPoiPositions(poiManager, this.poiType, predicate, entity.blockPosition(), 48, 48*48, PoiManager.Occupancy.HAS_SPACE, false, 5, poiposes);
```

This calls PoiAccess#findNearestPoiRecords
```java
 public static void findNearestPoiRecords(final PoiManager poiStorage,
                                             final Predicate<Holder<PoiType>> villagePlaceType,
                                             // position predicate must not modify chunk POI
                                             final Predicate<BlockPos> positionPredicate,
                                             final BlockPos sourcePosition,
                                             final int range, // distance on x y z axis
                                             final double maxDistance,
                                             final PoiManager.Occupancy occupancy,
                                             final boolean load,
                                             final int max,
                                             final List<PoiRecord> ret) {
        final Predicate<? super PoiRecord> occupancyFilter = occupancy.getTest();

        final double maxDistanceSquared = maxDistance * maxDistance;
        final Double2ObjectRBTreeMap<List<PoiRecord>> closestRecords = new Double2ObjectRBTreeMap<>();
        int totalRecords = 0;
        double furthestDistanceSquared = maxDistanceSquared;
```

Notice how these methods seem to square it too. So, most likely this should be fixed here too? 
Please correct me if I am wrong..

Lulu13022002:
:smile: I was literally relooking at the code and figured that now

Owen1212055 (approved):

kennytv (approved):

lynxplay (approved):

PR #8577, opened 2022-11-19
Expand Instrument API

Machine-Maker:
Trying out a new pattern. Instead of shoving tons of methods into Unsafe, this uses services to interact with nms from a non Craft- extended type.

### DIFF
--- a/patches/api/0407-Expand-Instrument-API.patch
+++ b/patches/api/0407-Expand-Instrument-API.patch
@@ -1,0 +1,62 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 19 Nov 2022 12:38:15 -0800
+Subject: [PATCH] Expand Instrument API
+
+
+diff --git a/src/main/java/org/bukkit/Instrument.java b/src/main/java/org/bukkit/Instrument.java
+index 92194803bcdfbcfdb99567835906ce9219eabd04..6fbdc2a177e7361e2e3e55d47714deec5944f907 100644
+--- a/src/main/java/org/bukkit/Instrument.java
++++ b/src/main/java/org/bukkit/Instrument.java
+@@ -78,6 +78,7 @@ public enum Instrument {
+ 
+     private final byte type;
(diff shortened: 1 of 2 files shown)


jpenilla:
Would be nice if the service interface wasn't public somehow (similar to PaperBrigadier) - the downside of that is mockbukkit etc would be forced to put their mock impl in the bukkit/paper package.

Machine-Maker:
That's not super simple because of the `craftbukkit` package mismatch. If this were all in paper packages, it could work. But it can't be package-private because there is no equivalent bukkit package (unless we add this one thing to org.bukkit).

patches/api/0407-Expand-Instrument-API.patch
```diff
@@ -0,0 +1,62 @@
...
+
+diff --git a/src/main/java/org/bukkit/Instrument.java b/src/main/java/org/bukkit/Instrument.java
+index 92194803bcdfbcfdb99567835906ce9219eabd04..6fbdc2a177e7361e2e3e55d47714deec5944f907 100644
+--- a/src/main/java/org/bukkit/Instrument.java
++++ b/src/main/java/org/bukkit/Instrument.java
+@@ -78,6 +78,7 @@ public enum Instrument {
+ 
+     private final byte type;
+     private static final Map<Byte, Instrument> BY_DATA = Maps.newHashMap();
++    private static final java.util.Optional<org.bukkit.Instrument.InstrumentUtil> UTIL = net.kyori.adventure.util.Services.service(InstrumentUtil.class); // Paper
```
MiniDigger, line 15: if we are only throwing on call anyways, might as well throw here, once, right? if we cant find that class, might as well stop start startup since something is fucked
generally, really like using services for stuff like this, nice idea

Machine-Maker: Api tests don’t work then

jpenilla:
> That's not super simple because of the `craftbukkit` package mismatch. If this were all in paper packages, it could work. But it can't be package-private because there is no equivalent bukkit package (unless we add this one thing to org.bukkit).

I think a couple methods in unsafe is better in this case then, rather than the methods and an additional interface

patches/api/0407-Expand-Instrument-API.patch
MiniDigger: fiar enough

PR #8579, opened 2022-11-20
Cache ResourceLocation#toString

montlikadani:


MiniDigger:
Do you have a profiler report that shows this as a hot code path? I am not convinced such a mirco optimization is worth the maintenance burden (and memory footprint I guess)

montlikadani:
No unfortunately. But if I saw correctly, this method is used for serializing for example PlayerAdvancements, like in `net\minecraft\advancements\Advancement`
```java
public JsonObject serializeToJson() {
...
    if (this.parent != null) {
        jsonobject.addProperty("parent", this.parent.getId().toString());
    } else if (this.parentId != null) {
        jsonobject.addProperty("parent", this.parentId.toString());
    }
```

kennytv:
A method being used a bunch doesn't matter if it doesn't do anything but simple string concatenation, even if it were somewhat of a hot spot - we have no indication of that being the case either

PR #8583, opened 2022-11-22
[ci skip] Rework workflows to support optional paperclip build

Machine-Maker:
## How this works
Add a label called whatever we want (workflows are configured for one named `build-pr-jar`).

When the label is added and anytime the PR is updated a paperclip artifact will be built and uploaded. 

Then at the bottom of the PR text body (what you are reading now), a link will be automatically added and updated to download the jar. 

This is entirely optional to use as we don't want every PR to have a paperclip jar, but if we want to provide an easy way to test changes, this is it.

### Drawbacks (but not really)
The `pr_comment` workflow will run for every `build` workflow that runs. But nothing will happen as most `build` workflows either aren't tied to a PR or don't have an uploaded artifact attached to them. I can't find a way to skip the `pr_comment` workflow.

I also tried having 1 workflow to also handle the comment's BUT, I can't figure out a way to get the artifact ID within the currently running workflow. The artifact doesn't seem to exist in the api until the workflow has ended. Hence the need for a second workflow to interact with the first's artifact.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,33 +1,86 @@
 # Here lie dragons!
 #
-# Note that there is no artifact step in this script. We do not want Paperclip
-# jars to be built for every push & PR; our CI handles pushes to branches, while
-# PRs can themselves link to Paperclip jars if it is necessary. Official such
-# PRs will take use of testing builds.
+# This action either builds the server or
+# builds a paperclip jar to be updated in the body
+# of the PR relating to this action.
 
 name: Build Paper
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
 
 jobs:
-    build:
-        # Only run on PRs if the source branch is on someone else's repo
-        if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
-        runs-on: ubuntu-latest
-        strategy:
-            matrix:
-                java: [17]
-            fail-fast: true
-        steps:
-            - uses: actions/checkout@v3
-            - name: JDK ${{ matrix.java }}
-              uses: actions/setup-java@v3.6.0
-              with:
-                  java-version: ${{ matrix.java }}
-                  cache: 'gradle'
-                  distribution: 'temurin'
-            - name: Patch and build
-              run: |
-                  git config --global user.email "no-reply@github.com"
-                  git config --global user.name "Github Actions"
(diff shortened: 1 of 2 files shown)


olijeffers0n:
This seems like a useful tool, one thing I would say is that it could get a bit obnoxious if someone pushes a lot of small updates - Maybe having a message that gets edited? I’m not sure if that’s possible :)

I see that it runs on the label event, is this called when you remove the label? This way you could just toggle it on/off by adding and removing the label to just get one build.

Doc94:
Maybe the workflow for pr-comment need more validations? not sure but like mention this workflow run for every success build.



> This seems like a useful tool, one thing I would say is that it could get a bit obnoxious if someone pushes a lot of small updates - Maybe having a message that gets edited? I’m not sure if that’s possible :)
> 
> I see that it runs on the label event, is this called when you remove the label? This way you could just toggle it on/off by adding and removing the label to just get one build.

the event is only for label, for remove the label they need to add unlabeled [source](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/events-that-trigger-workflows)

NoahvdAa:
>  Maybe having a message that gets edited? I’m not sure if that’s possible :)

> Then at the bottom of the PR text body (what you are reading now), a link will be automatically added and updated to download the jar.

Doc94:
> > Maybe having a message that gets edited? I’m not sure if that’s possible :)
> 
> > Then at the bottom of the PR text body (what you are reading now), a link will be automatically added and updated to download the jar.

maybe but pretty sure is more complex have a thing like in docs?
image

olijeffers0n:
> > Maybe having a message that gets edited? I’m not sure if that’s possible :)
> 
> > Then at the bottom of the PR text body (what you are reading now), a link will be automatically added and updated to download the jar.



> > Maybe having a message that gets edited? I’m not sure if that’s possible :)
> 
> > Then at the bottom of the PR text body (what you are reading now), a link will be automatically added and updated to download the jar.

Yeah I did read that Noah, it’s just what i mean is there could potentially being like 20 build messages on a PR which is more than a simple merge. Take the chunk system for example, there were a tonne of comments on that PR, and build messages would end up cluttering it more. 

I like what Doc mentioned, with the deployment, but i don’t think this is possible.

Machine-Maker:
What are you talking about? None of the workflows add any messages. How would that clutter anything?

olijeffers0n:
My bad, misread. Sounds good to me

kennytv (approved):
That's pretty hot

PR #8584, opened 2022-11-22
Add displayName methods for advancements

Machine-Maker:


patches/api/0331-Add-more-advancement-API.patch
```diff
@@ -107,6 +108,15 @@ index 0000000000000000000000000000000000000000..67341bb70762a2326030abd2548372b9
 +    NamespacedKey backgroundPath();
 +
 +    /**
++     * Gets the formatted display name for this display. This
++     * is the component that would be shown in chat when a player
```
jpenilla: is a part of?

patches/api/0331-Add-more-advancement-API.patch
```diff
@@ -189,6 +199,18 @@ index 17527c2f7bd5b8a5528388a53f2472bc1869c7f3..792d5ab90e9252b4f660434f72f004fa
 +    io.papermc.paper.advancement.AdvancementDisplay getDisplay();
 +
 +    /**
++     * Gets the formatted display name for this display. This
```
jpenilla, line 203: jd should include that when a display is present it will return the same as AD#displayName

patches/server/0694-Add-more-advancement-API.patch
```diff
@@ -73,6 +80,47 @@ index 0000000000000000000000000000000000000000..0567e500c40d3d424ddc19062c4f6da9
...
++            throw new IllegalArgumentException("can't both be null");
++        } else if (display == null) {
++            return Component.literal(id.toString());
+         } else {
+-            Component ichatbasecomponent = display.getTitle();
+-            ChatFormatting enumchatformat = display.getFrame().getChatColor();
+-            MutableComponent ichatmutablecomponent = ComponentUtils.mergeStyles(ichatbasecomponent.copy(), Style.EMPTY.withColor(enumchatformat)).append("\n").append(display.getDescription());
+-            MutableComponent ichatmutablecomponent1 = ichatbasecomponent.copy().withStyle((chatmodifier) -> {
+-                return chatmodifier.withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, ichatmutablecomponent));
++            Component title = display.getTitle();
```
jpenilla: this seems like more diff than necessary?

jpenilla (changes requested):

patches/server/0694-Add-more-advancement-API.patch
Machine-Maker: yeah, that turned out way worse than I had hoped for. I got it better tho.

patches/api/0331-Add-more-advancement-API.patch
```diff
@@ -189,6 +200,21 @@ index 17527c2f7bd5b8a5528388a53f2472bc1869c7f3..792d5ab90e9252b4f660434f72f004fa
...
++     * is the component that would be shown in chat when a player
```
jpenilla: same thing here, `is a part of`

jpenilla (changes requested):

patches/api/0331-Add-more-advancement-API.patch
Machine-Maker: done

jpenilla (approved):

PR #8585, opened 2022-11-22
Fixes potential issues arising from optimizing getPlayerByUUID

Machine-Maker:
Vanilla's getPlayerByUUID only can return player's in the level, but due to an optimization by paper, this changed to return any player with the UUID across all levels. This can lead to unexpected side effects for sculk events and tamable animal owners. 

This PR adds a level check to paper's optimization and replaces some uses of getPlayerByUUID with a getGlobalPlayerByUUID where you do want to return any matching player on the whole server.

### DIFF
--- a/patches/server/0338-Optimise-EntityGetter-getPlayerByUUID.patch
+++ b/patches/server/0338-Optimise-EntityGetter-getPlayerByUUID.patch
@@ -1,15 +1,15 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Spottedleaf <Spottedleaf@users.noreply.github.com>
 Date: Sat, 11 Jan 2020 21:50:56 -0800
-Subject: [PATCH] Optimise IEntityAccess#getPlayerByUUID
+Subject: [PATCH] Optimise EntityGetter#getPlayerByUUID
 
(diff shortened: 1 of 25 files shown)


jpenilla (approved):

Owen1212055 (approved):

patches/server/0338-Optimise-EntityGetter-getPlayerByUUID.patch
```diff
@@ -1,23 +1,24 @@
...
      }
      // Paper end - rewrite chunk system
  
 +    // Paper start - optimise getPlayerByUUID
 +    @Nullable
 +    @Override
 +    public Player getPlayerByUUID(UUID uuid) {
-+        return this.getServer().getPlayerList().getPlayer(uuid);
++        final Player player = this.getServer().getPlayerList().getPlayer(uuid);
++        return player != null && player.level == this ? player : null;
```
Bobcat00, line 21: Don't you need parentheses here?  Because what you have is essentially:
`return (player != null && player.level == this) ? player : null;`
so you're comparing a boolean to a player and the return will always be null.
I think you want:
`return player != null && (player.level == this ? player : null);`
I didn't look to see if this is wrong elsewhere in the PR.

PR #8590, opened 2022-11-23
Fix rogue armswing detection

### LINKED ISSUE #8540: PlayerInteractEvent fails to fire when left clicking an entity from 4 blocks away
### Expected behavior

I expect PlayerInteractEvent to be called whenever I left click (I DO NOT ignore cancelled events) with an item in main hand.

### Observed/Actual behavior

PlayerInteractEvent is called in every case expected EXCEPT when I am 3-4 blocks away from an entity and facing them (I used a zombie to test, may not be all entities). It starts calling when I'm 5 blocks away.

### Steps/models to reproduce

0. Have a plugin that sends a message anytime PlayerInteractEvent occurs, make sure it does not ignore cancelled events.
1. Create a box that can fit a zombie (1x1, 2 blocks tall) and be in survival mode after.
2. Put 3 blocks in between you and the zombie (screenshot: https://gyazo.com/3a0c6a4be3710149e306f33fa4178536)
3. Left click while looking at the zombie. At this point, you should see that the plugin does NOT send any messages.
4. Left click while looking elsewhere for a control group. The plugin should be sending messages at this point. The PlayerInteractEvent ONLY fails to fire when you're aiming at an entity (only tested zombie).

I have a hypothesis that this is because spigot is incorrectly using creative mode melee range (which is 1 block longer than survival mode melee range) to calculate when to fire the event. My evidence and reproduction is below:

0. Do step 0, 1, and 2 of above, but be in CREATIVE mode.
1. After putting 3 blocks between you and the zombie, left click. If the plugin doesn't send a message, take a very small step back. Left click again. Continue until you receive a message from the plugin. Notice that as SOON as you are not in range to hit the zombie, the plugin starts sending messages again. This is at around 4-5 blocks from the entity. This persists regardless of game mode, it's just easy to illustrate that it seems to use creative mode's range.

### Plugin and Datapack List

For the sake of this test, the ONLY plugin I ran was one that let me see if the event fired. I did not have any datapacks. The code is below:
```
public void onEnable() {
Bukkit.getServer().getLogger().info("NeoDebugger Enabled");
}

public void onDisable() {
	  org.bukkit.Bukkit.getServer().getLogger().info("NeoDebugger Disabled");
	  super.onDisable();
}

@EventHandler(ignoreCancelled = false)
public void onInteract(PlayerInteractEvent e) {
System.out.println("HERE: " + e.getAction() + " " + e.getHand());
}
```

### Paper version

[16:28:24 INFO]: This server is running Paper version git-Paper-258 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: ab1c072)
Previous version: git-Paper-220 (MC: 1.19.2)

### Other

_No response_

Owen1212055:
Fix https://github.com/PaperMC/Paper/issues/8540

The range for survival is actually 3 blocks.
This also fixes the issue where if you look "nearby" an entity it will incorrectly assume you
are attacking that entity.

TODO: Properly have check for blocks/entities separate

### DIFF
--- a/patches/server/0942-Fix-rouge-armswing-detection.patch
+++ b/patches/server/0942-Fix-rouge-armswing-detection.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Wed, 23 Nov 2022 13:19:25 -0500
+Subject: [PATCH] Fix rouge armswing detection
+
+
+diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+index d31a345edfffe39f127073fc3aec8b3489bae79c..90780fbe7da58ef9076c258dd8ac265cc6cd6a33 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -2750,9 +2750,9 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+         double d2 = this.player.getZ();
+         Location origin = new Location(this.player.level.getWorld(), d0, d1, d2, f2, f1);
+ 
+-        double d3 = player.gameMode.getGameModeForPlayer() == GameType.CREATIVE ? 5.0D : 4.5D;
++        double d3 = player.gameMode.getGameModeForPlayer() == GameType.CREATIVE ? 5.0D : 3.0D; // Paper
+         // SPIGOT-5607: Only call interact event if no block or entity is being clicked. Use bukkit ray trace method, because it handles blocks and entities at the same time
+-        org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.1, entity -> entity != this.player.getBukkitEntity());
++        org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0, entity -> entity != this.player.getBukkitEntity()); // Paper
+ 
+         if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode
+             CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_AIR, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);


Owen1212055:
Reported upstream, most likely will see a fix there. If not will reopen.

Moulberry:
Hello!
This fix is incorrect. While the entity reach for survival is 3.0, the block reach is 4.5
Changing the raycast distance to 3.0 will cause both LEFT_CLICK_AIR and LEFT_CLICK_BLOCK to be fired when a block is between 3.0 and 4.5 blocks away.

The correct fix is to do a raycast with 4.5 length, but allow the event to be fired if the distance to the entity is further than 3.0 blocks away.

A sample patch is provided below.

```
From: Moulberry <james.jenour@protonmail.com>
Date: Sun, 30 Apr 2023 22:40:51 +0800
Subject: [PATCH] Fix Paper/8540


diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
index 9c5754ac3b00d3f0cb80ec83a2beefadee6d0a14..89db79d3ba4d1725a65935b089e6ec48de7667b8 100644
@@ -2606,7 +2606,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
         // SPIGOT-5607: Only call interact event if no block or entity is being clicked. Use bukkit ray trace method, because it handles blocks and entities at the same time
         org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.1, entity -> entity != this.player.getBukkitEntity() && this.player.getBukkitEntity().canSee(entity));
 
-        if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode
+        if (result == null || (result.getHitEntity() != null && origin.toVector().distanceSquared(result.getHitPosition()) > 3.0D*3.0D) || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode // Paper - call PlayerInteractEvent when entity distance is greater than 3.0
             CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_AIR, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);
         }
```

Moulberry:
Hello!

Unfortunately, this fix is incorrect. It will result in both LEFT_CLICK_AIR and LEFT_CLICK_BLOCK firing if there is a block greater than 3.0 units away.

The correct fix is to perform the raycast with length 4.5, then allow the PlayerInteractEvent to be fired as long as the entity is further than 3.0 blocks away. As far as I can tell, this is the same logic that the vanilla GameRenderer#pick uses.

```
From: Moulberry <james.jenour@protonmail.com>
Date: Sun, 30 Apr 2023 23:06:24 +0800
Subject: [PATCH] Fix Paper/8540


diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
index 9c5754ac3b00d3f0cb80ec83a2beefadee6d0a14..7c113741fd06e400b4bfc0caf9a6a8221505f220 100644
@@ -2606,7 +2606,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
         // SPIGOT-5607: Only call interact event if no block or entity is being clicked. Use bukkit ray trace method, because it handles blocks and entities at the same time
         org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.1, entity -> entity != this.player.getBukkitEntity() && this.player.getBukkitEntity().canSee(entity));
 
-        if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode
+        if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE || (player.gameMode.getGameModeForPlayer() != GameType.CREATIVE && result.getHitEntity() != null && origin.toVector().distanceSquared(result.getHitPosition()) > 3.0D*3.0D)) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode // Paper - call PlayerInteractEvent when entity distance is greater than 3.0
             CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_AIR, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);
         }
```

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/9211

PR #8592, opened 2022-11-24
Add `/paper dumplisteners tofile` and increase detail of command output

jpenilla:
File gets dump in the `debug` dir similar to chunk debug

---
Download the paperclip jar for this pull request: [paper-8592.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/451990382.zip)

### DIFF
--- a/patches/server/0943-Add-paper-dumplisteners-command.patch
+++ b/patches/server/0943-Add-paper-dumplisteners-command.patch
@@ -27,10 +27,10 @@ index 724592234e2a178a518f6ab7d09c3180780371a7..92154550b41b2e1d03deb1271b71bb3b
              .flatMap(entry -> entry.getKey().stream().map(s -> Map.entry(s, entry.getValue())))
 diff --git a/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java b/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..5ddf123508ccf4e578ebbb3be086fcf85d4a31e1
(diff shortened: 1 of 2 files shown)


kennytv (approved):

patches/server/0943-Add-paper-dumplisteners-command.patch
```diff
@@ -93,12 +127,19 @@ index 0000000000000000000000000000000000000000..dab8d6fb489f8b3acc7c8fdaa1b5b6b8
 +
 +            sender.sendMessage(text("Listeners for " + args[0] + ":"));
 +
-+            Stream.of(handlers.getRegisteredListeners())
-+                .map(listener -> text(listener.getPlugin().getName(), GREEN)
++            for (final RegisteredListener listener : handlers.getRegisteredListeners()) {
++                final Component hoverText = text("Priority: " + listener.getPriority().name() + " (" + listener.getPriority().getSlot() + ")", WHITE)
++                    .append(newline())
++                    .append(text("Listener: " + listener))
```
Warriorrrr: The things included by RegisteredListener#toString are already part of the (hover) message, so this shouldn't be needed.

jpenilla: It's not needed anymore yes, however I kept it intentionally so that some basic info can be seen at a glance without hovering.

jpenilla: Misread this earlier, it was meant to be `listener.getListener()`, fixed now

Warriorrrr (approved):

MiniDigger (approved):

Owen1212055 (approved):
Tested and looks good. Maybe put it in the ``debug`` directory? 
The ChunkDebugCommand for example uses this.

PR #8593, opened 2022-11-24
Clarify working directories for patch modification

nullishamy:
This clarifies which repositories should be operated upon when editing existing patches, as this was previously unclear.

(Wording to be confirmed)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,9 @@ Your commit will be converted into a patch that you can then PR into Paper.
 
 Modifying previous patches is a bit more complex:
 
+> ! When modifying patches, you should be operating on the sources directory
+> until the modifications are complete, then you should operate on the root directory.
+
 ### Method 1
 
 This method works by temporarily resetting your `HEAD` to the desired commit to


CONTRIBUTING.md
```diff
@@ -99,6 +99,9 @@ Your commit will be converted into a patch that you can then PR into Paper.
 
 Modifying previous patches is a bit more complex:
 
+> ! When modifying patches, you should be operating on the sources directory
```
jpenilla, line 102: I feel like this is covered in the section above?

jpenilla (changes requested):

CONTRIBUTING.md
nullishamy: Hmm, it's stated there, but its still not clear (to me at least) that the same should be followed for modification.

Owen1212055:
It's tricky because although it's implied I do see why there can be some confusion. Perhaps say "similarly to adding patches" or something. As the whole process of editing a patch is being in the git repo for the pserv/api, which is pretty much needed when touching the patches at all.

Owen1212055:
Closing as in general this seems to be something that doesn't need to be clarified. Thank you for your contribution regardless.

PR #8600, opened 2022-12-01
remove duplicate animate packet for records

Machine-Maker:
Returning SUCCESS here will send an animate packet to all players tracking the "using" player which will be a duplicate packet because the server has already sent one in response to the serverbound arm swing packet. Verified the duplicate packet behavior on 1.19 using pakkit and a vanilla server vs a 1.19 paper server.

### DIFF
--- a/patches/server/0972-remove-duplicate-animate-packet-for-records.patch
+++ b/patches/server/0972-remove-duplicate-animate-packet-for-records.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 1 Dec 2022 12:42:18 -0800
+Subject: [PATCH] remove duplicate animate packet for records
+
+
+diff --git a/src/main/java/net/minecraft/world/item/RecordItem.java b/src/main/java/net/minecraft/world/item/RecordItem.java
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
For clarity's sake it may make sense to just put ``InteractionResult.sidedSuccess(world.isClientSide);`` instead of the actual interaction result.

But, looks good.

patches/server/0947-remove-duplicate-animate-packet-for-records.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/net/minecraft/world/item/RecordItem.java b/src/main/java/net/minecraft/world/item/RecordItem.java
+index 4c6f60cc1269ed0315b7e5a54235b9c95d1f039c..fa3868689f0435c991b82f2e3a943bb5ce404018 100644
+--- a/src/main/java/net/minecraft/world/item/RecordItem.java
++++ b/src/main/java/net/minecraft/world/item/RecordItem.java
+@@ -43,7 +43,7 @@ public class RecordItem extends Item {
+             ItemStack itemstack = context.getItemInHand();
+ 
+             if (!world.isClientSide) {
+-                if (true) return InteractionResult.SUCCESS; // CraftBukkit - handled in ItemStack
++                if (true) return InteractionResult.sidedSuccess(world.isClientSide);; // CraftBukkit - handled in ItemStack // Paper - fix duplicate animate packet
```
Krakenied: double semicolon

PR #8602, opened 2022-12-02
Fix command function perm checking

Machine-Maker:
When a command function tried to parse a command above it's configured permission level (see `function-permission-level` in `server.properties), it would throw a UOE due to a badly structured conditional in CommandSourceStack from upstream. Now instead of a UOE it throws an unknown command error, as it would in vanilla on an invalid command.


Machine-Maker:
Apparently opened another PR that makes this change... idk how I ended up opening 2 prs separated by 6 months that fix the same problem, but whatever. Was done in https://github.com/PaperMC/Paper/pull/9411

PR #8606, opened 2022-12-04
Include previous spawn angle in SpawnChangeEvent

Lulu13022002:
Include the previous spawn angle in the SpawnChangeEvent like spigot, actually it's always zero.

### DIFF
--- a/patches/server/0462-Fix-SpawnChangeEvent-not-firing-for-all-use-cases.patch
+++ b/patches/server/0462-Fix-SpawnChangeEvent-not-firing-for-all-use-cases.patch
@@ -5,29 +5,35 @@ Subject: [PATCH] Fix SpawnChangeEvent not firing for all use-cases
 
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index 95c53189e32aec6aca254fa3b157df7bc68d4d1a..0637fd424e98175f864f15c8bfa8bf696ed47fe7 100644
+index 95c53189e32aec6aca254fa3b157df7bc68d4d1a..b046d08f2c9cc78935e5b75b5367aa45d540465c 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
-@@ -1866,6 +1866,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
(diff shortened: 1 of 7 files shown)


patches/server/0466-Fix-SpawnChangeEvent-not-firing-for-all-use-cases.patch
```diff
@@ -5,26 +5,32 @@ Subject: [PATCH] Fix SpawnChangeEvent not firing for all use-cases
...
-index 5537903d6e1233489d31127f3af3d99c40258eb4..12534be05aa56930cde39e6fef90d402b5eb7c4e 100644
+index 5537903d6e1233489d31127f3af3d99c40258eb4..4d5203493472dde8d80a463027156f356400f269 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
-@@ -1893,6 +1893,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -1890,9 +1890,11 @@ public class ServerLevel extends Level implements WorldGenLevel {
+     public void setDefaultSpawnPos(BlockPos pos, float angle) {
+         // Paper - configurable spawn radius
+         BlockPos prevSpawn = this.getSharedSpawnPos();
++        float prevSpawnAngle = this.getSharedSpawnAngle(); // Paper
```
lynxplay: If we cannot completely rely on the prevSpawn anymore, is there any reason to not just use the existing logic in
CraftWorld#getSpawnLocation here ? That location includes the angle (hi doc lol)  (it would invoke ServerLevel#getSharedSpawnPos again, this isn't hot code tho).

Tho tbh, does not really matter, just kinda sad we don't have a MCUtil constructor for locations that includes yaw and pitch. Dunno if that is worth adding ? I am sure it has its usages down the line. 

Purely stylistic feedback tho, functionality looks great :+1: 
I am rather bed bound due to COVID getting me after three years so i don't know if I can test this today, if not I'll try by tomorrow to get this approved and merged :)

Lulu13022002: Oops i wasn't aware that spigot has implemented that so recently, i thought the reverse things. There's no reason i use that instead of the bukkit way (except maybe the previous statement).
We could maybe find more usage for a MCUtil helper but i think the new Position API should clean this a little bit more in general.
Take your time if you need some rest :heart:

Lulu13022002:
Yes i have rebased and also updated the mapping names :)

Machine-Maker (approved):

Owen1212055 (approved):

PR #8607, opened 2022-12-04
fixed npe

NonSwag:
Fixed Issue with Allay#stopDancing causing a null pointer exception if no jukebox was set before


Doc94:
I remember this PR.. maybe in this time the check for jukeboxPos == null was first... or im confuse..
i dont tested but looks good.

patches/server/0952-fixed-null-pointer-exception-when-invoking-Allay-sto.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftAllay.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftAllay.java
+index db3b0a9d3696f31d473342312d437e1eae97404a..29b25ee3081060b53358683a193196c01c80afb4 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftAllay.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftAllay.java
+@@ -82,8 +82,11 @@ public class CraftAllay extends CraftCreature implements org.bukkit.entity.Allay
+     @Override
+     public void stopDancing() {
+         this.getHandle().forceDancing = false;
+-        this.getHandle().jukeboxPos = null;
```
MiniDigger, line 16: why did you remove this line?

patches/server/0952-fixed-null-pointer-exception-when-invoking-Allay-sto.patch
```diff
@@ -0,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: NonSwag <mrminecraft00@gmail.com>
+Date: Sun, 4 Dec 2022 13:58:15 +0100
+Subject: [PATCH] fixed null pointer exception when invoking Allay#stopDancing
```
MiniDigger, line 4: nitpick, but can changed to "fixed NPE ..." to fit more text into the patch file name

patches/server/0952-fixed-null-pointer-exception-when-invoking-Allay-sto.patch
Doc94: the setJukeboxPlaying already set null the jukebox.. but see again the changes maybe its better only make the jukebox null and in the next tick the aiStep when check the jukebox stop the dance... make a thing like..
```java
@Override
public void stopDancing() {
  getHandle().forceDancing = false; // cancel the CB field for force dance
  getHandle().jukeboxPos = null; // remove any jukebox
  getHandle().setDancing(false); // save the nbt for dance in false
 }
```
this like the same than the setJukebox used in the original implementation.

PR #8609, opened 2022-12-05
Add FluidState API

Machine-Maker:
Not sure what to call one of the interfaces... right now its called `IdkWhat`. There are also probably some methods on nms FluidState we can put on FluidData, just not sure which ones so this PR is blocked until there is some feedback on that. Also methods for getting a FluidData need to be added to RegionAccessor

```[tasklist]
### TODO
- [x] https://github.com/PaperMC/Paper/pull/8608
- [ ] Rename `IdkWhat` to something good
- [ ] Add more methods to FluidData that point to methods on FluidState
- [ ] add FluidData methods to RegionAccessor
- [x] ~~Can you actually "set" fluid states? If not, remove the set methods, only have get methods.~~ (Nope, can't set, removed set methods)
```

### DIFF
--- a/patches/api/0415-Add-FluidState-API.patch
+++ b/patches/api/0415-Add-FluidState-API.patch
@@ -1,0 +1,95 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 4 Dec 2022 19:21:52 -0800
+Subject: [PATCH] Add FluidState API
+
+
+diff --git a/src/main/java/io/papermc/paper/block/fluid/FluidData.java b/src/main/java/io/papermc/paper/block/fluid/FluidData.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..251e5b5a9b5ec5c71fefb53af48eed423228f49c
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/block/fluid/FluidData.java
+@@ -0,0 +1,21 @@
++package io.papermc.paper.block.fluid;
++
++import org.bukkit.Fluid;
++import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


viciscat:
Hey I have been working on the API locally on my PC, I copied by hand what you did from the files in your PR because the patch just wouldn't work. Should I do a list of what I added here or make a PR of my own? Anyways I'm mostly here to ask a few questions:

I am not sure what is not enough or too many methods for `FluidData`. There is a method in `FluidState` that returns the direction of the flowing liquid as vector and I wonder if I should add that to `FluidData`?

By "add FluidData methods to RegionAccesor" you mean stuff like `getFluidData(x, y, z)`?

And finally a more general question, I've seen a lot that in interfaces some things are marked as NonNull but they aren't in their implemation, why is that?

Hope you can respond to my questions!

Machine-Maker:
> I am not sure what is not enough or too many methods for FluidData. There is a method in FluidState that returns the direction of the flowing liquid as vector and I wonder if I should add that to FluidData?

Yeah, I haven't looked too closely at what methods on FluidState could be easily added to FluidData. Probably `getHeight`, maybe getFlow.

> By "add FluidData methods to RegionAccesor" you mean stuff like getFluidData(x, y, z)?

Yes, like all the methods BlockData has on RegionAccessor. Currently the only way to get FluidData is via that method on BlockData.

> And finally a more general question, I've seen a lot that in interfaces some things are marked as NonNull but they aren't in their implemation, why is that?

Well upstream hasn't cared. Whenever **I** add a new type, I do add annotations to the implementation, but not directly. As you can see in the patch, a `package-info.java` file was added with checker-qual's annotations which makes all files in that package default to have `@NonNull` applied to everything. That is way better imo than manually annotating everything.

Machine-Maker:
Superseded by https://github.com/PaperMC/Paper/pull/9951

PR #8615, opened 2022-12-07
Fix a handful of 1.19.3 compile issues

NoahvdAa:


patches/server/0183-Player.setPlayerProfile-API.patch
```diff
@@ -132,7 +132,7 @@ index 865ba836d8a62fa08786e41f8434eb322885e965..32170da591ac2430ac08e327d759e3d7
 +
 +        //Respawn the player then update their position and selected slot
 +        ServerLevel worldserver = handle.getLevel();
-+        connection.send(new net.minecraft.network.protocol.game.ClientboundRespawnPacket(worldserver.dimensionTypeId(), worldserver.dimension(), BiomeManager.obfuscateSeed(worldserver.getSeed()), handle.gameMode.getGameModeForPlayer(), handle.gameMode.getPreviousGameModeForPlayer(), worldserver.isDebug(), worldserver.isFlat(), true, this.getHandle().getLastDeathLocation()));
++        connection.send(new net.minecraft.network.protocol.game.ClientboundRespawnPacket(worldserver.dimensionTypeId(), worldserver.dimension(), BiomeManager.obfuscateSeed(worldserver.getSeed()), handle.gameMode.getGameModeForPlayer(), handle.gameMode.getPreviousGameModeForPlayer(), worldserver.isDebug(), worldserver.isFlat(), (byte) 1, this.getHandle().getLastDeathLocation()));
```
Machine-Maker, line 135: Is 1 right here? Previously the `true` was to keep all player data, `3` is the value now to keep all player data. And it should use the static field inside the respawn packet instead of the magic value.

Machine-Maker (changes requested):

patches/server/0183-Player.setPlayerProfile-API.patch
kennytv: 1 is the same behavior as 1.19.2. the second bit is a new option

NoahvdAa: ~~I originally used 1 because CB used it somewhere (and I didn't see the constants, lol), changed it to the constant now.~~

EDIT: reverting, 2 is a new option, 1 is the old behaviour, according to Mr. Television in vc.

Machine-Maker: Just gonna clarify here, 1 is **not** the old behavior. The old behavior always kept entity data, the boolean just kept attributes. Keeping this as 1 just keeps attributes, not entity data. 3 keeps both. 

Kenny said he'd fix it.

PR #8622, opened 2022-12-08
Account for splash water bottles now extinguishing entities

jpenilla:
Still wip, not sure if we want to extend PotionSplashEvent or just ProjectileHitEvent... also not sure if we want our own handlerlist for this event or to use the old one for some sort of backwards compat


patches/api/0415-Add-SplashWaterBottleSplashEvent.patch
```diff
@@ -0,0 +1,97 @@
...
++import java.util.List;
++import java.util.Map;
++import org.bukkit.entity.Axolotl;
++import org.bukkit.entity.LivingEntity;
++import org.bukkit.entity.ThrownPotion;
++import org.bukkit.event.entity.PotionSplashEvent;
++import org.jetbrains.annotations.NotNull;
++
++public class SplashWaterBottleSplashEvent extends PotionSplashEvent {
++    private final @NotNull List<Axolotl> rehydrate;
```
electronicboy: I probably wouldn't use Axolotl incase they extend this out

patches/api/0415-Add-SplashWaterBottleSplashEvent.patch
```diff
@@ -0,0 +1,97 @@
...
+diff --git a/src/main/java/org/bukkit/event/entity/PotionSplashEvent.java b/src/main/java/org/bukkit/event/entity/PotionSplashEvent.java
+index 80f31a267ef4be88718811484e91a6097106e8d2..97d94aba0a8e86e055abd836dd868b8fe8a486bd 100644
+--- a/src/main/java/org/bukkit/event/entity/PotionSplashEvent.java
++++ b/src/main/java/org/bukkit/event/entity/PotionSplashEvent.java
+@@ -16,7 +16,7 @@ import org.jetbrains.annotations.NotNull;
+ public class PotionSplashEvent extends ProjectileHitEvent implements Cancellable {
+     private static final HandlerList handlers = new HandlerList();
+     private boolean cancelled;
+-    private final Map<LivingEntity, Double> affectedEntities;
++    protected final Map<LivingEntity, Double> affectedEntities; // Paper
```
Owen1212055: Do ATs not work on api patches?

electronicboy: no, ATs only apply to the server

jpenilla:
Also note this new event and PotionSplashEvent don't cancel their associated particle effects when the event is cancelled, might be better to address that in a future PR after the update though

electronicboy (approved):

jpenilla:
Changes lgtm

PR #8623, opened 2022-12-08
Take in account waterlogged block for the event

Lulu13022002:
There's some inconsistencies between the EntityChangeBlockEvent and the BlockDestroyEvent. The destroy event has always the right final type but the change block event has always air even for watterlogged blocks. This fix this behavior only for possible case to avoid a bigger patch. I have maybe missed some of them.


Machine-Maker:
The "to" BlockData isn't ever used to change the world right? This is just fixing it in the event, so the event has the correct information.

Lulu13022002:
Yes it's just the event block data which is wrong doesn't affect the world

Machine-Maker (approved):

Owen1212055 (approved):

PR #8625, opened 2022-12-08
Update adventure

Owen1212055:
Unsure if we should have deleteMessage implemented either, but this is a baseline and contains no signed message sending support.

### DIFF
--- a/patches/server/0009-Adventure.patch
+++ b/patches/server/0009-Adventure.patch
@@ -3179,7 +3179,7 @@ index 446fdca49a5a6999626a7ee3a1d5c168b15a09dd..f9863e138994f6c7a7975a852f106faa
      public boolean isOp() {
          return true;
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
-index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb046c7a4f6c 100644
+index d2be98416529aea3bdbedd0ea7131bd8de2a0162..3a7667fd0a53649b6b9f21ea1c1fd490532d2313 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
 @@ -277,14 +277,39 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
@@ -3405,7 +3405,7 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
      @Override
      public int getPing() {
          return this.getHandle().latency;
-@@ -1955,6 +2048,193 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+@@ -1955,6 +2048,247 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
          return this.getHandle().allowsListing();
      }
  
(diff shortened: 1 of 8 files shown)


patches/server/0009-Adventure.patch
```diff
@@ -3530,12 +3572,18 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
 +
 +    @Override
 +    public void playSound(final net.kyori.adventure.sound.Sound sound, final double x, final double y, final double z) {
++        final long seed;
++        if (sound.seed().isPresent()) {
++            seed = sound.seed().getAsLong();
++        } else {
++            seed = this.getHandle().getRandom().nextLong();
++        }
```
Machine-Maker: Probably should use the one-liner for all this

patches/server/0009-Adventure.patch
```diff
@@ -3549,13 +3597,19 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
 +        } else {
 +            throw new IllegalArgumentException("Sound emitter must be an Entity or self(), but was: " + emitter);
 +        }
++        final long seed;
++        if (sound.seed().isPresent()) {
++            seed = sound.seed().getAsLong();
++        } else {
++            seed = this.getHandle().getRandom().nextLong();
++        }
```
Machine-Maker: same as above

patches/server/0009-Adventure.patch
```diff
@@ -3530,12 +3572,18 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
...
 +        if (event.isPresent()) {
-+            this.getHandle().connection.send(new ClientboundSoundEntityPacket(BuiltInRegistries.SOUND_EVENT.wrapAsHolder(event.get()), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), entity, sound.volume(), sound.pitch(), this.getHandle().getRandom().nextLong())); // TODO adventure sound seed
++            this.getHandle().connection.send(new ClientboundSoundEntityPacket(BuiltInRegistries.SOUND_EVENT.wrapAsHolder(event.get()), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), entity, sound.volume(), sound.pitch(), seed));
 +        } else {
-+            this.getHandle().connection.send(new ClientboundSoundPacket(Holder.direct(SoundEvent.createVariableRangeEvent(name)), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), x, y, z, sound.volume(), sound.pitch(), this.getHandle().getRandom().nextLong())); // TODO adventure sound seed
++            this.getHandle().connection.send(new ClientboundSoundPacket(Holder.direct(SoundEvent.createVariableRangeEvent(name)), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), x, y, z, sound.volume(), sound.pitch(), seed));
 +        }
```
Machine-Maker: Why is this using the sound entity packet in the first place? That looks wrong but if its right, it supports custom sounds

patches/server/0009-Adventure.patch
```diff
@@ -3549,13 +3597,19 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
...
++            this.getHandle().connection.send(new net.minecraft.network.protocol.game.ClientboundSoundEntityPacket(BuiltInRegistries.SOUND_EVENT.wrapAsHolder(event.get()), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), entity, sound.volume(), sound.pitch(), seed));
 +        } else {
-+            this.getHandle().connection.send(new ClientboundSoundPacket(Holder.direct(SoundEvent.createVariableRangeEvent(name)), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), entity.position().x(), entity.position().y(), entity.position().z(), sound.volume(), sound.pitch(), this.getHandle().getRandom().nextLong())); // TODO adventure sound seed
++            this.getHandle().connection.send(new ClientboundSoundPacket(Holder.direct(SoundEvent.createVariableRangeEvent(name)), io.papermc.paper.adventure.PaperAdventure.asVanilla(sound.source()), entity.position().x(), entity.position().y(), entity.position().z(), sound.volume(), sound.pitch(), seed));
 +        }
```
Machine-Maker: the entity sound packet supports custom sounds now

patches/server/0009-Adventure.patch
```diff
@@ -3422,6 +3422,48 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
...
++//        net.minecraft.network.chat.PlayerChatMessage playerChatMessage = new net.minecraft.network.chat.PlayerChatMessage(
++//            null, // TODO:
++//            new net.minecraft.network.chat.MessageSignature(signedMessage.signature().bytes()),
++//            null, // TODO
++//            io.papermc.paper.adventure.PaperAdventure.asVanilla(signedMessage.unsignedContent()),
++//            net.minecraft.network.chat.FilterMask.PASS_THROUGH
++//        );
++//
++//        this.getHandle().sendChatMessage(net.minecraft.network.chat.OutgoingChatMessage.create(playerChatMessage), this.getHandle().isTextFilteringEnabled(), this.toHandle(boundChatType));
++//    }
```
Machine-Maker: If we aren't going to do the main impl yet, we should at least override the default impl and point system signed messages to the method that takes a bound chat type. See the default Audience#sendMessage(SignedMessage,ChatType.Bound) impl

Machine-Maker (changes requested):

patches/server/0009-Adventure.patch
Owen1212055: Ahh. I thought someone already made the change lol, similar names.

patches/server/0009-Adventure.patch
```diff
@@ -3422,6 +3422,54 @@ index d2be98416529aea3bdbedd0ea7131bd8de2a0162..e12b4ec0d14901a0246cfea16c65fb04
 +    }
 +
 +    @Override
++    public void deleteMessage(net.kyori.adventure.chat.SignedMessage.Signature signature) {
++        if (getHandle().connection == null) return;
++        net.minecraft.network.chat.MessageSignature sig = new net.minecraft.network.chat.MessageSignature(signature.bytes());
++
++        this.getHandle().connection.send(new net.minecraft.network.protocol.game.ClientboundDeleteChatPacket(new net.minecraft.network.chat.MessageSignature.Packed(0, sig)));
```
Machine-Maker: 0 is not the "ignore" message id. See the other ctor for MessageSignature.Packed which uses -1. Just use that other ctor that only takes the MessageSignature object.

PR #8632, opened 2022-12-09
fix ArmorStandMeta not applying false flags

TehBrian:
Currently, trying to set one of ArmorStandMeta's flags (invisible, small, etc.) to false with the API does nothing. This PR fixes that.

One issue with this solution is that explicitly setting flags to false now saves the flags to item, which is technically useless data, so perhaps something like this for the flag handlers in `#applyToItem` might be better:
```java
if (this.bool != null) {
  if (this.bool) {
    this.entityTag.putBoolean(BOOL.NBT, this.bool);
  else {
    this.entityTag.remove(BOOL.NBT);
  }
}
```
But I'm not sure if that might cause issues with compatibility.

### DIFF
--- a/patches/server/0176-Add-ArmorStand-Item-Meta.patch
+++ b/patches/server/0176-Add-ArmorStand-Item-Meta.patch
@@ -13,7 +13,7 @@ starting point for future additions in this area.
 Fixes GH-559
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
-index 4017933f2244fca32cf9d39444f3a4f550e8af01..06bd5e6dd7ca6e6a44dca6ff133233b53c8ae22d 100644
+index 4017933f2244fca32cf9d39444f3a4f550e8af01..e721517ce7b52a1aa10d039aa9f309eb69db4733 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
 @@ -8,9 +8,22 @@ import org.bukkit.Material;
@@ -131,7 +131,7 @@ index 4017933f2244fca32cf9d39444f3a4f550e8af01..06bd5e6dd7ca6e6a44dca6ff133233b5
  
      boolean isArmorStandEmpty() {
 -        return !(this.entityTag != null);
-+        return !(this.isInvisible() || this.hasNoBasePlate() || this.shouldShowArms() || this.isSmall() || this.isMarker() || this.entityTag != null);
++        return !(this.invisible != null || this.noBasePlate != null || this.showArms != null || this.small != null || this.marker != null || this.entityTag != null); // Paper
      }
  
      @Override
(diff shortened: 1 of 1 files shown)


electronicboy:
Shouldn't call remove, `not_set, true, false` are all valid states; technically if null you should call remove (though, that can't be set)

TehBrian:
Hm. We could make the API's parameters `@Nullable Boolean` to reflect that tristate, but I doubt that'd be worth it. (Also, does switching from a primitive to a wrapper break API compatibility?)

Anyway, as long setting the flags to true/false works, I'm happy. I don't think any plugin developers need to care about trimming very small amounts of data from the item. (Or, if they are, there are other ways to do that.)

electronicboy:
checking to a Boolean would be an ABI break, as well as a behavioral break; That's ofc the joy of this stuff is that CraftMeta doesn't understand these nuances, which becomes an issue because this stuff matters at the implementation level, even if it rarely matters to plugins

TehBrian:
Alright, gotcha. In its current state, though, will this not break stuff?

electronicboy (approved):

electronicboy:
yea, in its current state this looks fine

patches/server/0176-Add-ArmorStand-Item-Meta.patch
```diff
@@ -13,11 +13,14 @@ starting point for future additions in this area.
...
-index 4017933f2244fca32cf9d39444f3a4f550e8af01..0c40a4a18cfc6f4ed6473a475f307f5c75ab56c5 100644
+index 4017933f2244fca32cf9d39444f3a4f550e8af01..d5a63727912f6775dd0eb3f4acbcbc553d1dc0b0 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
-@@ -8,9 +8,22 @@ import org.bukkit.Material;
+@@ -6,11 +6,25 @@ import net.minecraft.nbt.CompoundTag;
+ import net.minecraft.nbt.Tag;
+ import org.bukkit.Material;
  import org.bukkit.configuration.serialization.DelegateDeserialization;
++import org.checkerframework.checker.nullness.qual.Nullable; // Paper
```
Owen1212055: Idk if the nullable is needed, it's set as null so i'm guessing it can be implied. Just, import diffs are yucky

TehBrian: Hm, what do you think about using the FQN?

Machine-Maker: I think he means you just don't have to annotate the Type in the field as Nullable. Simply initializing it to null should be enough. We aren't picky about nullability annotations on the server, just api.

Machine-Maker (approved):
See the comment about the Nullable annotation, but otherwise looks good.

TehBrian:
Alrighty, yoinked the Nullable import and annotations.

Owen1212055 (approved):

Owen1212055:
While testing I discovered a little issue that I will be adding to this patch.
These values will not be saved if there is no entity tag on the armor stand, as it checks to make sure that armor stand "isn't empty" (which is a flawed approach)

is empty will now return if any of the values are set

Owen1212055:
Thanks for your contribution! 😄

PR #8636, opened 2022-12-10
Improve vibration events

Machine-Maker:
Adds EntityReceiveGameEvent and expands BlockReceiveGameEvent

BlockReceiveGameEvent was not called for all block listeners, and was called for entity listeners. This separates them out into two events and adds missing event calls. It also adds more api for bypassing line of sight/occlusion checks (not sure what to call that).

Still needs testing, but wanted to open this while I did that.


Machine-Maker:
Updated for 1.19.4 and removed the "vibration" part of the event. GameEvent are not equivalent to vibrations. There could be a use for a future sub event of the block/entity events for vibration specifically, but this isn't that PR.

PR #8642, opened 2022-12-11
Replace ConfiguredStructure api with Structure

Machine-Maker:
So this is an API break, but I really don't see any way around it. In `StructuresLocateEvent` there was a list of configured structures. That list was mutable, and the event supported removing structures from it. I didn't see a good way to keep that mutable functionality when replacing that list with the list of upstream's StructureType.


Machine-Maker:
Tests fixed, and diff significantly cleaned up, removing old server types that we don't need anymore. At some point in the future we can come back through and delete `ConfiguredStructure` entirely, maybe in a few months/mc versions.

kennytv (approved):

PR #8643, opened 2022-12-11
Deprecate and replace methods using deprecated StructureType

### LINKED ISSUE #9477: Server#createExplorerMap() uses deprecated StructureType
### Expected behavior

org.bukkit.generator.structure.StructureType or/and org.bukkit.generator.structure.Structure in Server#createExplorerMap()

### Observed/Actual behavior

Deprecated org.bukkit.StructureType in the Server#createExplorerMap() method

### Steps/models to reproduce

Try using the org.bukkit.generator.structure.StructureType or org.bukkit.generator.structure.Structure in Server#createExplorerMap()

### Plugin and Datapack List

N/A

### Paper version

[21:30:27 INFO]: Checking version, please wait...
[21:30:27 INFO]: This server is running Paper version git-Paper-72 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: a3c760e)
Previous version: git-Paper-71 (MC: 1.20.1)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/9477

Well this change certainly got more complicated. In order to deprecate and add replacements for the methods on Server/Bukkit that used the deprecated StructureType, I had to completely switch to mockito for testing. Otherwise there's a static initializer loop created. Apparently the default heap sizes then weren't enough for the tests, so I bumped those up (this seems weird, maybe I did something wrong with mockito).

### DIFF
--- a/patches/server/1004-Deprecate-and-replace-methods-with-old-StructureType.patch
+++ b/patches/server/1004-Deprecate-and-replace-methods-with-old-StructureType.patch
@@ -1,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 10 Dec 2022 17:52:38 -0800
+Subject: [PATCH] Deprecate and replace methods with old StructureType
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+index 249d76acac9a91cd46f0b8a477511974a75d6f4a..ec4b73321205b472f19fa5bd4ad95893020d1340 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -1872,6 +1872,11 @@ public final class CraftServer implements Server {
+ 
+         ServerLevel worldServer = ((CraftWorld) world).getHandle();
+         Location structureLocation = world.locateNearestStructure(location, structureType, radius, findUnexplored);
++        // Paper start - don't throw NPE
++        if (structureLocation == null) {
++            throw new IllegalStateException("Could not find a structure for " + structureType);
++        }
++        // Paper end
+         BlockPos structurePosition = CraftLocation.toBlockPosition(structureLocation);
+ 
+         // Create map with trackPlayer = true, unlimitedTracking = true
+@@ -1882,6 +1887,31 @@ public final class CraftServer implements Server {
+ 
+         return CraftItemStack.asBukkitCopy(stack);
+     }
++    // Paper start - copied from above (uses un-deprecated StructureType type)
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored) {
++        Preconditions.checkArgument(world != null, "World cannot be null");
++        Preconditions.checkArgument(location != null, "Location cannot be null");
++        Preconditions.checkArgument(structureType != null, "StructureType cannot be null");
++        Preconditions.checkArgument(mapIcon != null, "mapIcon cannot be null");
++
++        ServerLevel worldServer = ((CraftWorld) world).getHandle();
++        final org.bukkit.util.StructureSearchResult structureSearchResult = world.locateNearestStructure(location, structureType, radius, findUnexplored);
++        if (structureSearchResult == null) {
++            return null;
++        }
++        Location structureLocation = structureSearchResult.getLocation();
++        BlockPos structurePosition = new BlockPos(structureLocation.getBlockX(), structureLocation.getBlockY(), structureLocation.getBlockZ());
++
++        // Create map with showIcons = true, unlimitedTracking = true
++        net.minecraft.world.item.ItemStack stack = MapItem.create(worldServer, structurePosition.getX(), structurePosition.getZ(), MapView.Scale.NORMAL.getValue(), true, true);
++        MapItem.renderBiomePreviewMap(worldServer, stack);
++        // "+" map ID taken from VillagerTrades$TreasureMapForEmeralds
++        MapItem.getSavedData(stack, worldServer).addTargetDecoration(stack, structurePosition, "+", MapDecoration.Type.byIcon(mapIcon.getValue()));
++
++        return CraftItemStack.asBukkitCopy(stack);
++    }
++    // Paper end
+ 
+     @Override
+     public void shutdown() {
--- a/patches/api/0426-Deprecate-and-replace-methods-with-old-StructureType.patch
+++ b/patches/api/0426-Deprecate-and-replace-methods-with-old-StructureType.patch
@@ -1,0 +1,159 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 10 Dec 2022 17:52:45 -0800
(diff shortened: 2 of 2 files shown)


Machine-Maker:
The mockito refactor was done in a separate PR so this one is a lot simpler now, just adds replacement methods to Bukkit and Server for methods that used the now-deprecated `StructureType` type instead of the new-ish replacement. Upstream didn't add them when they replaced the type.

Machine-Maker:
Rebased for 1.20.1

patches/server/0988-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,42 @@
...
+     }
++    // Paper start - copied from above (uses un-deprecated StructureType type)
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon) {
++        return this.createExplorerMap(world, location, structureType, mapIcon, 100, true);
++    }
++
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored) {
++        Preconditions.checkArgument(world != null, "Wgsorld cannot be null");
```
Sytm: Is "wgsorld" some special name in this context or is it a typo?

Machine-Maker: that is me typing `gs` which is my alias for `git status` while the cursor is not in the terminal 🤣

patches/server/0988-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,42 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -1844,6 +1844,31 @@ public final class CraftServer implements Server {
+ 
+         return CraftItemStack.asBukkitCopy(stack);
+     }
++    // Paper start - copied from above (uses un-deprecated StructureType type)
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon) {
++        return this.createExplorerMap(world, location, structureType, mapIcon, 100, true);
```
Lulu13022002: This should be in a default method on the Server interface

patches/server/0988-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,42 @@
...
++    }
++
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored) {
++        Preconditions.checkArgument(world != null, "World cannot be null");
++        Preconditions.checkArgument(structureType != null, "StructureType cannot be null");
++        Preconditions.checkArgument(mapIcon != null, "mapIcon cannot be null");
++
++        ServerLevel worldServer = ((CraftWorld) world).getHandle();
++        Location structureLocation = world.locateNearestStructure(location, structureType, radius, findUnexplored).getLocation();
```
Lulu13022002: This will throw a NPE if the structure is not found in the radius specified, that was already the case on the old methods. Maybe the method should be nullable?

Machine-Maker: It did do this before... I will make this method nullable and change the deprecated one to throw an exception if it doesn't find a location instead of a random NPE.

Zymo5221:
What about [Structure](https://jd.papermc.io/paper/1.20/org/bukkit/generator/structure/Structure.html)? In [World#locateNearestStructure()](https://jd.papermc.io/paper/1.20/org/bukkit/World.html#locateNearestStructure(org.bukkit.Location,org.bukkit.generator.structure.Structure,int,boolean)) you can use it.[ StructureType.JIGSAW](https://jd.papermc.io/paper/1.20/org/bukkit/generator/structure/StructureType.html#JIGSAW) describes all structures that use jigsaw blocks. I'm currently learning java and I'm not familiar with all these things but I made this:
[0989-Deprecate-and-replace-methods-with-old-StructureType.patch](https://github.com/PaperMC/Paper/files/12062874/0989-Deprecate-and-replace-methods-with-old-StructureType.patch)
[0422-Deprecate-and-replace-methods-with-old-StructureType.patch](https://github.com/PaperMC/Paper/files/12062875/0422-Deprecate-and-replace-methods-with-old-StructureType.patch)
By using that you can create explorer map for ancient city, villages, pillager outposts and trail ruins. I'm ready to hear what you guys think 👍

patches/api/0422-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,146 @@
...
++    default @Nullable ItemStack createExplorerMap(@NotNull World world, @NotNull Location location, @NotNull org.bukkit.generator.structure.StructureType structureType, @NotNull org.bukkit.map.MapCursor.Type mapIcon) {
++        return this.createExplorerMap(world, location, structureType, mapIcon, 100, true);
++    }
++
++    /**
++     * Create a new explorer map targeting the closest nearby structure of a
++     * given {@link org.bukkit.generator.structure.StructureType}.
++     * <br>
++     * This method uses implementation default values for radius and
++     * findUnexplored (usually 100, true).
```
Lulu13022002: This doesn't seems to be true here, it use the parameters and not the default values. It's also the case for the old methods and the method in the other class

lynxplay (approved):

patches/server/0989-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,53 @@
...
+         // Create map with trackPlayer = true, unlimitedTracking = true
+@@ -1844,6 +1849,30 @@ public final class CraftServer implements Server {
+ 
+         return CraftItemStack.asBukkitCopy(stack);
+     }
++    // Paper start - copied from above (uses un-deprecated StructureType type)
++    @Override
++    public ItemStack createExplorerMap(World world, Location location, org.bukkit.generator.structure.StructureType structureType, org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored) {
++        Preconditions.checkArgument(world != null, "World cannot be null");
++        Preconditions.checkArgument(structureType != null, "StructureType cannot be null");
```
NoahvdAa: location is also annotated notnull but is missing a check here

NoahvdAa (changes requested):

NoahvdAa (approved):

patches/server/1004-Deprecate-and-replace-methods-with-old-StructureType.patch
```diff
@@ -0,0 +1,54 @@
...
++        if (structureSearchResult == null) {
++            return null;
++        }
++        Location structureLocation = structureSearchResult.getLocation();
++        BlockPos structurePosition = new BlockPos(structureLocation.getBlockX(), structureLocation.getBlockY(), structureLocation.getBlockZ());
++
++        // Create map with trackPlayer = true, unlimitedTracking = true
++        net.minecraft.world.item.ItemStack stack = MapItem.create(worldServer, structurePosition.getX(), structurePosition.getZ(), MapView.Scale.NORMAL.getValue(), true, true);
++        MapItem.renderBiomePreviewMap(worldServer, stack);
++        // "+" map ID taken from EntityVillager
```
Lulu13022002: The comments could be updated with the new names/classes in this method

Lulu13022002 (approved):

PR #8644, opened 2022-12-11
Add Position methods to pathfinder api

Machine-Maker:
Deprecates methods that return location, but left methods with Location parameters as undeprecated because Location implements Position, and would therefore require a cast to not use the deprecated methods (if you still used a Location).

Also made PaperPath a static class instead of inner class. This will be useful later for a certain memory key which PaperPath is the type we would use for it (can't use it if its an inner class in PaperPathfinder).


Owen1212055 (approved):

Warriorrrr (approved):

Owen1212055:
Superseded, will be done in bulk later.

PR #8651, opened 2022-12-11
Add and remove missing or obsolete vanilla tags

Machine-Maker:
Brings the Tag class up to date with current vanilla tags. Also adds a test to check that it is always up to date in the future.

Something this PR doesn't do, is address the 6 "Experimental" tags that are "null" if the datapack isn't enabled. I'm not sure what to do about that.


Machine-Maker:
Updated to fix tests

Machine-Maker:
tags are autogenerated now, no need to rebase.

PR #8657, opened 2022-12-12
Fix BlockDispenseEvent being called when nothing actually happens

Machine-Maker:
The BlockDispenseEvent was incorrectly called in two specific situations, when bonemealing an invalid location, like a solid block or when trying to use flint & steel in an incorrect way. This doesn't match up with the rest of the calls of BlockDispenseEvent in which something always happens, and in other cases where something *might* happen, that logic is checked before calling the event.


Machine-Maker:
Rebased for 1.20.4

Something that probably needs to be worked out is when exactly BlockDispenseEvent should be fired for a "failed dispense". A lot of items fallback to just being spit out if their "special action" fails, like glowstone being dispensed but the block in front isn't a respawn anchor. In this situation, dispenser with glowstone inside pointing at a non-respawn anchor block, the glowstone will be spat out. But if the block in front is a full respawn anchor, the dispense will "fail", not spitting out glowstone or refilling the anchor. Does BlockDispenseEvent fire in that case? Nothing happened, nothing changed, no item ejected, nothing. Unsure.

PR #8658, opened 2022-12-12
Properly track the changed item from dispense events

Machine-Maker:
There are several places where the potentially changed item from the BlockDispenseEvent wasn't used in later dispenser logic (in a way that had an effect).

Example: If you changed the spawn egg in the BlockDispenseEvent, the spawned mob didn't change, and bucket dispenser behavior has some issues like for mob buckets, the nbt would be transferred over to the spawned entity from the original stack, not the stack set in the event.


Machine-Maker:
Fixed this issue in a few more places, armor stands, shulker boxes, fire charges, and some more projectiles. 

There is a question about the Flint & Steel which needs to be resolved, how to handle its behavior if someone changes the flint & steel in the dispense event.

Machine-Maker:
Rebased for 1.20.4

lynxplay (approved):

PR #8659, opened 2022-12-12
[ci skip] add missing final newline

Machine-Maker:
Intellij keeps making this change automatically, I think its missing a newline? But github usually shows those visually, so idk what it is.


lynxplay (approved):
Files without a final new line look ugly anyway.

electronicboy (approved):

PR #8665, opened 2022-12-13
Fix CreativeCategory api

### LINKED ISSUE #8770: getCreativeCategory() doesn't work correctly
### Expected behavior

Returns the correct CreativeCategory from a Material

### Observed/Actual behavior

Allways returns BUILDING_BLOCKS

### Steps/models to reproduce

get the CreativeCategory from a Material and check if its the right one

### Plugin and Datapack List

FFA(The problematic Plugin), LuckPerms, Multiverse-Core, Plan, WebConsole, Worldedit

### Paper version

paper-1.19.3-371

### Other

_No response_

Machine-Maker:
Probably in the future this should be expanded to get all itemstacks in a specific category. It's not perfect system since ItemStacks are in categories now, so also a method to check if an ItemStack is in a specific category would be nice.

_____
Fixes #8770

### DIFF
--- a/patches/server/0980-Fix-CreativeCategory-api.patch
+++ b/patches/server/0980-Fix-CreativeCategory-api.patch
@@ -1,0 +1,225 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Tue, 13 Dec 2022 12:58:27 -0800
+Subject: [PATCH] Fix CreativeCategory api
+
+
+diff --git a/src/main/java/io/papermc/paper/inventory/PaperCreativeCategory.java b/src/main/java/io/papermc/paper/inventory/PaperCreativeCategory.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..7dd68ca04370e9402629caa138dc0f1d5d12483d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/inventory/PaperCreativeCategory.java
+@@ -0,0 +1,47 @@
++package io.papermc.paper.inventory;
++
++import com.google.common.base.Enums;
++import com.google.common.collect.ImmutableMap;
++import java.util.List;
++import java.util.Map;
++import net.minecraft.network.chat.contents.TranslatableContents;
++import net.minecraft.world.item.CreativeModeTab;
++import net.minecraft.world.item.CreativeModeTabs;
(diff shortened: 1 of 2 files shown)


Owen1212055 (changes requested):
This doesn't seem to work? 
Tested and it always returns an empty set.

Machine-Maker:
It'd help if I actually pushed from my local repo...

Owen1212055 (approved):

AlmightyAnubis:
Would also be nice if the Materials in the Material Enum would be sortet the way the new search tab in the creativ inventory is sortet. So the wood is grouped etc.

Machine-Maker:
The material enum is most likely being deprecated either this update or next update, not really worth it.

Even so, I don't really think the order of fields is important at all. Also, some materials are in multiple categories, so not sure how that would factor into the field order.

AlmightyAnubis:
In the search bar, more or less all items are contained. And they are sortet in the new way, which is much more convinient. When I list items, I would really like this way of order over the old one.
image

electronicboy (approved):

patches/server/0980-Fix-CreativeCategory-api.patch
```diff
@@ -0,0 +1,225 @@
...
++import net.minecraft.world.item.CreativeModeTab;
++import net.minecraft.world.item.CreativeModeTabs;
++import org.bukkit.inventory.CreativeCategory;
++import org.checkerframework.checker.nullness.qual.NonNull;
++import org.checkerframework.framework.qual.DefaultQualifier;
++
++@DefaultQualifier(NonNull.class)
++public final class PaperCreativeCategory {
++
++    public static final List<CreativeModeTab> BLOCK_TABS = CreativeModeTabs.allTabs().stream().filter(tab -> tab.getType() == CreativeModeTab.Type.CATEGORY).toList();
```
Warriorrrr, line 29: CreativeModeTabs#streamAllTabs can be used here

Warriorrrr (approved):

KodingDev:
Hey, any updates on this? Can't really fallback to the existing implementation since it's just a constant for all items.

NonSwag:
> Hey, any updates on this? Can't really fallback to the existing implementation since it's just a constant for all items.

Iirc the creative categories are deprecated since they no longer exist on the server

KodingDev:
Gotcha, any way we could semi mimic it with tags? That's about the only half-baked solution that's really worked for us

PR #8666, opened 2022-12-13
Fix custom worlds

### LINKED ISSUE #8664: WorldCreator creates world as super flat, when specified as normal.
### Expected behavior

World should be normal type - plains, mountains, deserts, rivers, etc.

### Observed/Actual behavior

World is super flat.

### Steps/models to reproduce

I've been using this code to create a world weekly for a few years. I started having this problem after the update to 1.19.3 this morning.

WorldCreator creator = new WorldCreator("sampleWorld").type(WorldType.NORMAL).environment(Environment.NORMAL).generateStructures(true);
this.plugin.sampleWorld = this.plugin.getServer().createWorld(creator);

I fully deleted my world and re-created it with this, but it's super flat mode instead of normal. My only WILD GUESS is that it might be related to my default world's mode, which is super flat.

### Plugin and Datapack List

Lots, but I have not changed anything in the transition from 1.19.2 to 1.19.3. I only updated the server JAR.

### Paper version

This server is running Paper version git-Paper-325 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 0148283)

### Other

It should be unrelated, but I have specified minecraft:normal in server.properties.

Clarifying - my default world is super flat because it was created that way 3 years ago. My current SETTING in server.properties is minecraft:normal. And the code above specifies normal to the world creator. I expected the new world would be normal, but it is instead super flat.

I'm stuck in a big way because my players need this additional world for resources.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/8664

Tested and it works

---
Download the paperclip jar for this pull request: [paper-8666.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/474865055.zip)

### DIFF
--- a/patches/server/0947-Fix-custom-worlds.patch
+++ b/patches/server/0947-Fix-custom-worlds.patch
@@ -1,0 +1,74 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Tue, 13 Dec 2022 14:54:53 -0800
+Subject: [PATCH] Fix custom worlds
+
+
+diff --git a/src/main/java/net/minecraft/world/level/levelgen/WorldGenSettings.java b/src/main/java/net/minecraft/world/level/levelgen/WorldGenSettings.java
+index 42c569561afdce599b72cfa455435e87e250885f..e0fb7cbce04bb47971d546ae88cdcb11f4d6de09 100644
+--- a/src/main/java/net/minecraft/world/level/levelgen/WorldGenSettings.java
++++ b/src/main/java/net/minecraft/world/level/levelgen/WorldGenSettings.java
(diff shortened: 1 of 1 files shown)


BigScary:
Hmmm.... yes it's sort of fixed, but chunk generation is incredibly slow, so slow that players are just stuck hovering in the air for a minute or more while the chunk they have teleproted into waits to generate.

Machine-Maker:
Ok, well this should have no effect on this. Chunk generation is always slow, much slower than on flat worlds. But if the world now loads correct, as in its a normal world not a flat world, then this can be fixed when that PR is merged.

Machine-Maker:
Upstream did exactly this in https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/2ed3e3e6e50cb606709522d4476f98e9edc5e79c#nms-patches/net/minecraft/world/level/storage/WorldDataServer.patch

PR #8669, opened 2022-12-14
Fix crop modifier problem with Sugar Cane and Cactus (Issue #8544)

bridgelol:


jpenilla:
Just removing the modifier isn't an option, https://github.com/PaperMC/Paper/pull/8564 has a more proper solution to the linked issue

Also when creating PRs in the future please read the information in CONTRIBUTING.md about paper comments and minimal diff

PR #8673, opened 2022-12-14
Improve some logging and errors

Machine-Maker:
Previously, if you had a file inside the root `data/` folder of a datapack, it was ignored. Now it just assumes they are all directories and so loading a datapack with that spits out about 50 stacktraces. Just need to log the error less and earlier.

Also changes the legacy mats to a warning not an error. We get too many "reports" of an error because it was error in the logs.

Do we have any other patches that either remove bad logged errors or any other place we want to change this stuff? I feel like we have more, but I just can't remember them.

### DIFF
--- a/patches/server/0947-Improve-logging-and-errors.patch
+++ b/patches/server/0947-Improve-logging-and-errors.patch
@@ -1,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 14 Dec 2022 15:52:11 -0800
+Subject: [PATCH] Improve logging and errors
+
+
+diff --git a/src/main/java/net/minecraft/server/packs/PathPackResources.java b/src/main/java/net/minecraft/server/packs/PathPackResources.java
+index 0232c29d96e1021a9f5a9678996993dc55fe7254..8ad8ad1189d7cdb58caaa39c482d32685afa3f9a 100644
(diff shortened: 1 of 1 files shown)


patches/server/0948-Improve-logging-and-errors.patch
```diff
@@ -0,0 +1,44 @@
...
+index 0232c29d96e1021a9f5a9678996993dc55fe7254..748c959c1e112279c361889be0c70f8ba81999eb 100644
+--- a/src/main/java/net/minecraft/server/packs/PathPackResources.java
++++ b/src/main/java/net/minecraft/server/packs/PathPackResources.java
+@@ -105,6 +105,12 @@ public class PathPackResources extends AbstractPackResources {
+         try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(path)) {
+             for(Path path2 : directoryStream) {
+                 String string = path2.getFileName().toString();
++                // Paper start
++                if (!Files.isDirectory(path2)) {
++                    LOGGER.error("Invalid directory entry: {} in {}. This will cause way more errors in vanilla.", string, this.root, new java.nio.file.NotDirectoryException(string));
```
electronicboy: > This will cause way more errors in vanilla.

not fond of the wording of that, otherwise, lgtm

electronicboy (approved):

patches/server/0948-Improve-logging-and-errors.patch
Machine-Maker: k, just removed that.

PR #8675, opened 2022-12-15
Correctly mark pseudo inventories as not creatable

Machine-Maker:
Fixes an NPE when the server tries to find an inventory creator instance based on the InventoryType. Now it throws an IllegalArgumentException like other un-creatable inventories.

I think this is the correct solution, not to add inventory creator instances to the CraftInventoryCreator but I could be wrong. The chiseled bookshelf is a tile entity which technically would be easy to add one for.


Owen1212055 (approved):

Owen1212055:
Fixed by upstream in https://github.com/PaperMC/Paper/commit/7af4cd3647c9c89f25d13ce9e041814e2baca50d

PR #8677, opened 2022-12-15
Fix several issues with EntityBreedEvent

Machine-Maker:
Upstream did not account for different hands when storing the breed item for later use in the event. Also they only stored a reference to the stack, not a copy so if the stack changed after love mode was started, the breed item in the event also changed. Also in several places, the breed item was stored after it was decreased by one to consume the item.


Owen1212055 (approved):

Warriorrrr (commented):
It might be nice to add an ItemStack parameter to setInLove to pass the copy through, DoNotUse could then be added to the old signature in case of updates.

Machine-Maker:
> It might be nice to add an ItemStack parameter to setInLove to pass the copy through, DoNotUse could then be added to the old signature in case of updates.

Yeah, I like this. Implemented.

lynxplay (approved):

PR #8695, opened 2022-12-17
Avoid respawn desync of player metadata

### LINKED ISSUE #8689: Skin layer Dissapears after death
### Expected behavior

Alright so I was playing on my Paper MC 1.19.3 server when I died my skin layers Dissapeared and my Internet was still good but my skin layer disappeared 
I tried everything nothing worked I hope you fix this issue

### Observed/Actual behavior

My skin layer Dissapeared after death 

### Steps/models to reproduce

Install a paper software and run the server without any plugins or Datapacks
Then try to die in your paper mc 
server 
See if you lose your Skin layer 
Same steps I did 7 times still happening in 1.19.3

### Plugin and Datapack List

No datapacks and plugins were installed 

### Paper version

Paper mc version 1.19.3

### Other

That was Only the issue and stupid Bug Costed my Friendship with my friend please fix this As soon as possible

Lulu13022002:
Closes #8689 
Avoid to desync the player metadata on respawn and world teleport
Before the 1.19.3 the metadata was copied with the respawn packet but now it's not always the case (notably at the death)
And since upstream doesn't recreate the player instance, the metadata aren't dirty anymore here and so aren't resend to the client later. This emulate the behavior of like if the player was recreated with the default values to avoid a total reset at the respawn packet.

### DIFF
--- a/patches/server/0950-Avoid-respawn-desync-of-player-metadata.patch
+++ b/patches/server/0950-Avoid-respawn-desync-of-player-metadata.patch
@@ -3,6 +3,9 @@ From: Lulu13022002 <41980282+Lulu13022002@users.noreply.github.com>
 Date: Sat, 17 Dec 2022 13:04:14 +0100
 Subject: [PATCH] Avoid respawn desync of player metadata
 
+== AT ==
+public net.minecraft.world.entity.player.Player DATA_PLAYER_ABSORPTION_ID
+public net.minecraft.world.entity.player.Player DATA_SCORE_ID
 
 diff --git a/src/main/java/net/minecraft/network/syncher/SynchedEntityData.java b/src/main/java/net/minecraft/network/syncher/SynchedEntityData.java
 index 37e193f57938047c8b886ed7d2816411392f94b4..f7d1425d102959f7a2f1da5585e2d3d2032a5abb 100644
@@ -25,10 +28,10 @@ index 37e193f57938047c8b886ed7d2816411392f94b4..f7d1425d102959f7a2f1da5585e2d3d2
      public boolean isDirty() {
          return this.isDirty;
 diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
-index c0b0a7fdb75266a7064d54bda6441953184ecc64..96f4be56db7d02e4d6c342b5c0dbea7958f9f564 100644
+index c0b0a7fdb75266a7064d54bda6441953184ecc64..14960c8e40f34047664ef7f2ddcb214bc65034f9 100644
 --- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
 +++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
-@@ -1776,31 +1776,35 @@ public class ServerPlayer extends Player {
+@@ -1776,11 +1776,14 @@ public class ServerPlayer extends Player {
          if (alive) {
              this.getInventory().replaceWith(oldPlayer.getInventory());
              this.setHealth(oldPlayer.getHealth());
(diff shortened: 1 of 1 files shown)


patches/server/0950-Avoid-respawn-desync-of-player-metadata.patch
```diff
@@ -0,0 +1,86 @@
...
+             this.getInventory().replaceWith(oldPlayer.getInventory());
+             this.setHealth(oldPlayer.getHealth());
++            this.getEntityData().markDirtyIfNotDefault(LivingEntity.DATA_HEALTH_ID); // Paper
++            this.getEntityData().markDirtyIfNotDefault(Player.DATA_PLAYER_ABSORPTION_ID); // Paper
+             this.foodData = oldPlayer.foodData;
+             this.experienceLevel = oldPlayer.experienceLevel;
+             this.totalExperience = oldPlayer.totalExperience;
+             this.experienceProgress = oldPlayer.experienceProgress;
+-            this.setScore(oldPlayer.getScore());
++            this.getEntityData().markDirtyIfNotDefault(Player.DATA_SCORE_ID); // Paper
```
Machine-Maker, line 45: I think the old setXXX(getXXX()) stuff should stay, just after it you mark it as dirty if its not default. In case it ever changes upstream and we don't catch it here

patches/server/0950-Avoid-respawn-desync-of-player-metadata.patch
```diff
@@ -0,0 +1,86 @@
...
+     private static final Map<Pose, EntityDimensions> POSES = ImmutableMap.<Pose, EntityDimensions>builder().put(Pose.STANDING, Player.STANDING_DIMENSIONS).put(Pose.SLEEPING, Player.SLEEPING_DIMENSIONS).put(Pose.FALL_FLYING, EntityDimensions.scalable(0.6F, 0.6F)).put(Pose.SWIMMING, EntityDimensions.scalable(0.6F, 0.6F)).put(Pose.SPIN_ATTACK, EntityDimensions.scalable(0.6F, 0.6F)).put(Pose.CROUCHING, EntityDimensions.scalable(0.6F, 1.5F)).put(Pose.DYING, EntityDimensions.fixed(0.2F, 0.2F)).build();
+     private static final int FLY_ACHIEVEMENT_SPEED = 25;
+-    private static final EntityDataAccessor<Float> DATA_PLAYER_ABSORPTION_ID = SynchedEntityData.defineId(Player.class, EntityDataSerializers.FLOAT);
+-    private static final EntityDataAccessor<Integer> DATA_SCORE_ID = SynchedEntityData.defineId(Player.class, EntityDataSerializers.INT);
++    public static final EntityDataAccessor<Float> DATA_PLAYER_ABSORPTION_ID = SynchedEntityData.defineId(Player.class, EntityDataSerializers.FLOAT); // Paper
++    public static final EntityDataAccessor<Integer> DATA_SCORE_ID = SynchedEntityData.defineId(Player.class, EntityDataSerializers.INT); // Paper
```
Machine-Maker: We now recommend using ATs again for this because the ATs are per-patch and much nicer to look at (instead of them all being in one BIG HUGE patch).

Machine-Maker (changes requested):

patches/server/0950-Avoid-respawn-desync-of-player-metadata.patch
Lulu13022002: That's should be good

Owen1212055:
Do the other synced values not need to be resent? I would think that instead what we should do here is mark all of the sync values as dirty?

Lulu13022002:
The respawn packet will already do that for us in case of a world teleport. For the death all values are reset in vanilla except those i have marked as dirty here

Owen1212055 (approved):
I'd put in the description that this is caused by spigot reusing the server player instance and that this should be removed in the future.

Lulu13022002:
Upstream has pushed a fix for that to update all metadata when not default, this solve the skin desync and those metadata desync (clientside) but since upstream has also a lot of desync (serverside) that happens when a player died (due to some missing metadata reset in the ServerPlayer#reset to again simulate the Player instance recreated) that brings back those desync into the client. Then there no desync at all but it's not anymore vanilla and also that send another packet when the player change of dimension with portal for no reason. So if we want to keep the upstream patch we would need to reset all the metadata on death serverside otherwise some metadata will not be reinitialized properly. Those desync are also present in this patch but aren't resend to the client so ideally we would recreate the instance for a proper fix.

Owen1212055:
Closing as upstream has fixed this and just resends everything.

PR #8703, opened 2022-12-20
Add copy constructor to MerchantRecipe

Joo200:
I want to create multiple Merchant with MerchantRecipes based on player permissions. The idea is: Store a List of MerchantRecipes and create a new Merchant for every player with custom uses.

It's not possible to clone MerchantRecipes so modifying the stored recipes will modify the results for other players too.


Machine-Maker:
I thought Cloneable sucked. Wouldn't we be better off with a copy constructor instead? I know the api is full of "Cloneable"s but do we want to keep going down that path?

JOO200:
ItemStack has already Cloneable that's why I used Cloneable here.

I don't see that advantage to use a copy instructor here. Cloneables are not that great but to have "some copy ctors" and "some cloneables" is not that great.

Owen1212055:
In general, I agree that moving away from cloneable is probably better. 
I'd just make a copy constructor as MM mentioned. 👍

JOO200:
I'll change it to a copy constructor next week

JOO200:
Changed to copy ctor

Owen1212055 (approved):

lynxplay (approved):

kennytv (approved):

PR #8707, opened 2022-12-21
Fix spawn-animals and spawn-monsters settings

Machine-Maker:
There are a few things this fixes. Craftbukkit attempted to move the logic regarding the `spawn-monsters`, `spawn-npcs` and `spawn-animals` server.properties settings to the logic for calling the entity spawn events. This broke quite a bit with regards to vanilla behavior.

1. The `spawn-monsters` setting should only affect natural spawns, not any other method of spawning (summon command, spawn eggs).
2. The `spawn-animals` setting should only affect instances of `nms.Animal` and `nms.WaterAnimal`, not `nms.AbstractGolem`.
3. Fixes an issue with entity loading due to chunk loading not respecting the per-world nature of `spawn-animals`

### DIFF
--- a/patches/server/1053-Fix-spawn-animals-and-spawn-monsters-settings.patch
+++ b/patches/server/1053-Fix-spawn-animals-and-spawn-monsters-settings.patch
@@ -1,0 +1,73 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 21 Dec 2022 10:38:41 -0800
+Subject: [PATCH] Fix spawn-animals and spawn-monsters settings
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ChunkMap.java b/src/main/java/net/minecraft/server/level/ChunkMap.java
+index d9cd497bc1b654030ff1a597f038b6a881df9f6b..b8ad61d7a0fb4bda02a561b284fcbf19ee3b0e0c 100644
+--- a/src/main/java/net/minecraft/server/level/ChunkMap.java
++++ b/src/main/java/net/minecraft/server/level/ChunkMap.java
+@@ -649,18 +649,12 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+         if (!nbt.isEmpty()) {
+             // CraftBukkit start - these are spawned serialized (DefinedStructure) and we don't call an add event below at the moment due to ordering complexities
+             world.addWorldGenChunkEntities(EntityType.loadEntitiesRecursive(nbt, world).filter((entity) -> {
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

Machine-Maker:
Rebased for 1.19.4

patches/server/0969-Fix-spawn-animals-and-spawn-monsters-settings.patch
```diff
@@ -0,0 +1,73 @@
...
+index 5a5ff40df37db9cbd53c584ed26a3ce4888b29c0..c1811c3ab960f8587a3630b96bbad4523c436042 100644
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -784,6 +784,11 @@ public class ServerLevel extends Level implements WorldGenLevel {
+     private boolean shouldDiscardEntity(Entity entity) {
+         return !this.server.isSpawningAnimals() && (entity instanceof Animal || entity instanceof WaterAnimal) ? true : !this.server.areNpcsEnabled() && entity instanceof Npc;
+     }
++    // Paper start - per world version of above method
++    public boolean shouldDiscardEntityInWorld(Entity entity) {
++        return !this.getChunkSource().spawnFriendlies && (entity instanceof Animal || entity instanceof WaterAnimal) ? true : !this.server.areNpcsEnabled() && entity instanceof Npc; // TODO per-world npc
```
Warriorrrr: This seems like it could also be done like this, but good either way.

```suggestion
+        return (!this.getChunkSource().spawnFriendlies && (entity instanceof Animal || entity instanceof WaterAnimal)) || (!this.server.areNpcsEnabled() && entity instanceof Npc); // TODO per-world npc
```

Warriorrrr (approved):

patches/server/0969-Fix-spawn-animals-and-spawn-monsters-settings.patch
Machine-Maker: I did it that way to copy more exactly what was done above since its copied directly from there. I think, for that reason, its fine to leave as is

Machine-Maker:
https://bugs.mojang.com/browse/MC-379 There is a super old confirmed report saying the `spawn-animals` and `spawn-npcs` should function the same way as `spawn-monsters` in only preventing the natural generation of monsters, not spawn eggs, commands, etc.

This would be a relatively large behavioral change that is > 10 years old where spawn eggs for monsters on servers `spawn-monsters: false` would suddenly start working.

Machine-Maker:
1.21.2 has removed `spawn-animals` and `spawn-npcs`, so this is blocked until that is merged. This pr will likely condense down into only fixing the `spawn-monsters` inconsistencies with vanilla.

PR #8710, opened 2022-12-22
Fix OfflinePlayer getPlayerProfile return type

Machine-Maker:
---
Download the paperclip jar for this pull request: [paper-8710.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/485802101.zip)


electronicboy (approved):

LOOHP:
Does this cause `getPlayerProfile()`compiled with Spigot to fail when calling this method with a `NoSuchMethodError`?

Machine-Maker:
No it shouldn’t. That was addressed in the Commodore.java file which rewrites plugins when they are loaded.

Issue #4787, opened 2020-11-19
ItemFrameMeta to SPAWN invisible item frames


Rincewind34:
**Is your feature request related to a problem? Please describe.**
I would like to create the ITEM, that can "spawn" an invisible item frame introduced in 1.16.

**Describe the solution you'd like**
A new subclass of ItemMeta with a new method

**Describe alternatives you've considered**
Current work around: meta.getPersistentDataContainer().set(new NamespacedKey(plugin, "invisible"), PersistentDataType.BYTE, (byte) 1);

kennytv:
https://papermc.io/javadocs/paper/1.16/org/bukkit/entity/ItemFrame.html#setVisible-boolean-

kennytv:
Whoops sorry, misread it. 🤔

Rincewind34:
The title was a little bit misleading. I changed it ;)

lynxplay:
Closing this as it will be implemented via properly API lands. Exposing meta API for this seems like more trouble than what this feature is worth to maintain.

I hope you can understand, we are going to work hard on the property API so this should be possible sooner than later :)

codeHusky:
This didn't end up getting implemented in the DataComponent PR, is there still a current plan to implement this or was the ability to affect ItemFrame visibility / fixed state on an ItemMeta level (as well as the other CraftMetaEntityTag items)? @Owen1212055

electronicboy:
The issue is that there is no current means of representing an entity tag in the API, and the desire to deal with this while mojang is heavily churning how they represent data in the server is ultra low

codeHusky:
Gotcha, makes sense, glad it's on the radar at least. Will work around the API gap in the meantime

Issue #4708, opened 2020-10-31
Implement a way to read/write itemmeta without expensive cloning


ghost:
**Is your feature request related to a problem? Please describe.**
Itemstack.getItemmeta() even when you only want something like displayname clones the entire meta and if the item is a written book, runs heavy regex on every page taking excess of 100ms in some cases

**Describe the solution you'd like**
perhaps getitemmeta(boolean snapshot) or a new interface called liveItemmeta etc. Something that wouldn't break plugins already using it this way. Perhaps _parts_ of the meta could be initialized when the for example getlore() methods are called etc

**Describe alternatives you've considered**
There isn't really any other solution other than using NMS to read NBT an item's NBT tags.

**Additional context**

electronicboy:
ItemMeta is contractually a snapshot, there is no way to implement ItemMeta without cloning the entire existing data structure; This would need to be an entirely separate API for this specific purpose

MiniDigger:
I mean, couldn't you say the same about block states? yet paper still allows you not take a snapshot

electronicboy:
That API works fundamentally differently to how ItemMeta works

Issue #7747, opened 2022-04-18
1.18.2 Paper Server Wipes NBT Data in Villager Trades


bryanmcgrane:
### Expected behavior

I believe I found an issue with Paper 1.18.2. Upgrading from a vanilla server to Paper totally broke my quests, so I had to roll back to vanilla. I have a hunch it's a bug in the way Paper reads and stores NBT data.

So on my server, I have custom items that are used in quests. These items can look like this:
/give @p copper_ore{display:{Name:'{"text":"Super Copper"}',Lore:['{"text":"It\'s amazing"}']},Enchantments:[{}]} 1

Notice - this copper ore has Enchantments:[{}] - this is to give it a glint without actually giving it any other enchantments.

Now, when players receive these items through the quests, they'll trade them into villagers that look like this:
/summon villager ~ ~ ~ {Offers:{Recipes:[{buy:{id:"minecraft:copper_ore",Count:1b,tag:{display:{Name:'{"text":"Super Copper"}',Lore:['{"text":"It\'s amazing"}']},Enchantments:[{}]}},sell:{id:"minecraft:iron_axe",Count:1b}}]}}

It seems when a player opens a dialog with this villager and attempts to trade, sometimes the Enchantments:[{}] tag information will be removed from either the trading item or the villager trade itself. This means the quest is completely broken and / or the player loses their quest item.

Is this a known issue? I am happy to provide as much detail as required, as I am unable to use Paper right now, and I need to for the chunk loading optimizations.

### Observed/Actual behavior

What I observed is the NBT data is altered when trading with a villager.

In vanilla, this does not occur.

### Steps/models to reproduce

Give yourself and item:
```
/give @p copper_ore{display:{Name:'{"text":"Super Copper"}',Lore:['{"text":"It\'s amazing"}']},Enchantments:[{}]} 1
```

Create a villager:
```
/summon villager ~ ~ ~ {Offers:{Recipes:[{buy:{id:"minecraft:copper_ore",Count:1b,tag:{display:{Name:'{"text":"Super Copper"}',Lore:['{"text":"It\'s amazing"}']},Enchantments:[{}]}},sell:{id:"minecraft:iron_axe",Count:1b}}]}}
```

Attempt to trade with this villager several times.

### Plugin and Datapack List

None

### Paper version

I did not capture the exact version, as I am now rolled back: 1.18.2

Machine-Maker:
[...]

PR #7335, opened 2022-01-14
Fix StoredEnchants reordering inconsistently

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/6437

Tested using the reproduction steps outlined in above issue.

I think its better to not reorder stored enchants, as the order does have an effect on anvil usage. It's also a smaller diff, less to maintain, and will change 0 existing items, only newly created ones as LinkedHashSets keep the order from the EnchantmentMap but don't change it.

PR #8711, opened 2022-12-22
Paper Item Property API

### LINKED ISSUE #8757: Ability to get entity data from spawner eggs and spawner block (and eventually other ItemMeta and BlockState containing entities/entity tags)

I'm writing a plugin to scan my world for cheated/unusual stuff. I intend to scan everything, like blocks, entities, inventories,...
I try my best to look into entity/tileentity inventories, inventories in items (shulker box),... 
I know that spawner blocks and spawner eggs can have custom nbt that would be copied into the spawned entity (like pre-filled custom name, inventory content,... ).
I'm not able to find API that provides the entity data (in `SpawnEggMeta` for the item meta of spawner eggs, and `CreatureSpawner` block state for the spawner block), except the `EntityType`.

- 1st idea: provide a method to get an instance of Entity that correspond to the type and data of the spawner, and is not actually spawned in the world (like if it was already dead or despawned instantly), so it is possible to use the bukkit entity api to access all the desired data.
- Another solution is to provide a method to actually trigger the spawning of the entity (this is the same as above, but we are not returning a dead/not-spawned entity instance). The plugin dev would also have the choice to kill/delete the entity or not, after.
- Alfway solution: the method would have a boolean parameter to determine if the entity is spawned into the world or not.

- Access the nbt data directly
- don't scan my world that much, there is a low chance that someone thought about hiding a cheated item in a shulker-box inside a mule inventory that is stored in a spawner egg... 

### Other

In addition to solve the initial issue, the presented solution would also provide a way to programmatically spawn an entity based on a spawner egg item or a spawner block.

EDIT: it would also be nice to have this functionality extended for:
- Other items that spawn entities: ItemFrame (no specific ItemMeta interface), ArmorStand (`ArmorStandMeta`) and Aquatic mob buckets (`TropicalFishBucketMeta`, `AxolotlBucketMeta`) that can all have a EntityTag tag in their NBT.
- BlockState:
    - Beehive (and possibly all `EntityBlockStorage`) with the possibility to list the entities without having to release them (with releaseEntities() method)

Owen1212055:
Very much WIP, but is the coming of future sponge-inspired ways to manipulate items/entity data.

The current approach to edit items has poorly aged, and in general, creates a lot of issues due to it acting as its own implementation. This PR aims to prevent the API from interacting with items unless necessary, preventing unnecessary parsing and in general useless reparsing for each item.

ItemMeta is still very important as it allows plugin developers to have a more friendly way of editing items. In this PR, it is proposed that we add a new property API that allows developers to ALSO directly edit values themselves. This area of the API would be marked as experimental and in general, would be subject to changes if Mojang changes nbt tags. 

(WIP WIP WIP)
```java
ItemStack itemstack = ItemStack.of(Material.STONE);
itemstack.getPropertyHolder().set(ItemProperties.DAMAGE, 0);

// Same thing as (but more performant since it directly edits the item)
ItemMeta meta = itemstack.getItemMeta(); // Clones the item nbt compound to a meta wrapper
meta.setDamage(0); // modifies that wrapped meta
itemstack.setItemMeta(meta); // overrides the wrapped meta
```

Using the property system, you can directly modify a value without needing to do any sort of cloning, it's a 1:1 operation. ItemMeta has also been improved in this, as now it will instead act as a **wrapper** that uses this item property api internally. This means that the behavior of the property/meta api will be consistent. For legacy reasons item meta will still need to be cloned, so it will have a small overhead as it clones the underlying NBT tag. After that, however, it is still much faster and prevents parsing the entire item.

Future goals:
- [ ] Figure out what to do about things that need api only itemstacks.
- [ ] Use this same property system to allow for entity nbt manipulation. (For example, making a zombie a baby in spawn egg)
- [ ] Thoroughly test
- [ ] Finalize design points
- [ ] Always use a compound based stack
- [ ] ItemStack interface maybe?

Fixes:
#4708 
#5071 
#6719 
#6689
#6437 
#6414
#7747
#7335

To fix:
https://github.com/PaperMC/Paper/issues/4787

---
Download the paperclip jar for this pull request: [paper-8711.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1140490339.zip)


patches/api/0423-Paper-Item-API.patch
```diff
@@ -0,0 +1,653 @@
...
+-import org.jetbrains.annotations.NotNull;
+-import org.jetbrains.annotations.Nullable;
+ 
+ /**
+  * A ClassLoader for plugins, to allow shared classes across multiple plugins
+@@ -66,6 +67,24 @@ public final class PluginClassLoader extends URLClassLoader { // Spigot
+         this.logger = com.destroystokyo.paper.utils.PaperPluginLogger.getLogger(description); // Paper - Register logger early
+ 
+         try {
++
```
olijeffers0n: what is this lol?

patches/api/0423-Paper-Item-API.patch
```diff
@@ -0,0 +1,653 @@
...
++    static <K, V> PropertyKey<MultiMapValue<K, V>> ofMultiMap(String key, TypeToken<K> keyToken, TypeToken<V> valueToken) {
++        return new PropertyKeyImpl<>(key, TypeFactory.parameterizedClass(Multimap.class, keyToken.getType(), valueToken.getType()), TypeFactory.parameterizedClass(MultiMapValue.class, keyToken.getType(), valueToken.getType()));
++    }
++
++    static <K, V> PropertyKey<MultiMapValue<K, V>> ofMultiMap(String key, Class<K> keyToken, Class<V> valueToken) {
++        return new PropertyKeyImpl<>(key, TypeFactory.parameterizedClass(Multimap.class, keyToken, valueToken), TypeFactory.parameterizedClass(MultiMapValue.class, keyToken, valueToken));
++    }
++
++    String getKey();
```
yannicklamprecht: Any reason why you used `getKey()` instead of `key()`. Same question for the other two.
With that we can reduce the size of the PropertyKeyImpl.

patches/api/0423-Paper-Item-API.patch
```diff
@@ -0,0 +1,653 @@
...
++import org.bukkit.attribute.AttributeModifier;
++
++import java.lang.reflect.Type;
++import java.util.List;
++import java.util.Map;
++import java.util.Set;
++
++public interface PropertyKey<T extends Value<?>> {
++
++    static <V> PropertyKey<SetValue<V>> ofSet(String key, TypeToken<V> valueToken) {
```
yannicklamprecht: The issue I see here currently, is that any dev is able to define arbitrary nbt entries.
They could probably create an invalid map itemstack with it.
Instead of 

```java
 PropertyKey<Value<Integer>> MAP_ID = PropertyKey.of("map", Integer.class);
```
they could for example use that.

```java
 PropertyKey<Value<String>> MAP_ID_CODE = PropertyKey.of("map", String.class);
 ```

The game / client expects the `map` to be an Integer but it gets a String instead.
As with an interface we cannot ensure package-private static methods, it should either be moved to another class.

A Bukkitvalues equivalent would be neat as well.

File: patches/api/0423-Paper-Item-API.patch (hunk 1 above)
Comment (yannicklamprecht) [reply patches/api/0423-Paper-Item-API.patch]: Guess that is not intended as the only changes here are new whitespace and a import organize.

File: patches/server/0950-Paper-Item-API.patch [hunk 4]
```diff
@@ -0,0 +1,4333 @@
...
++        super(propertyHolder);
++    }
++
++    @Override
++    public boolean hasAuthor() {
++        return MetaBook.this.propertyHolder.hasProperty(ItemProperties.AUTHOR);
++    }
++
++    @Override
++    public boolean hasTitle() {
```
yannicklamprecht: Most of the simpler methods could be moved into API as we can operate on.

```java
ItemStack#getPropertyHolder()..hasProperty(ItemProperties.AUTHOR);
```

Then we're able to fully unit test that on the API side as we just inject a dummy PropertyHolder.

patches/server/0950-Paper-Item-API.patch
```diff
@@ -0,0 +1,4333 @@
...
++    List<PropertyKey<?>> CRAFT_REPAIRABLE = List.of(
++        REPAIR_COST
++    );
++
++    // CraftMetaPotion
++//    PropertyHolder.Immutable DEFAULT_STATE = PropertyHolder.Immutable.builder()
++//        .of(DEFAULT_POTION_TYPE, Value.of(DEFAULT_POTION_TYPE, new PotionData(PotionType.UNCRAFTABLE, false, false)))
++//        .build();
++
++    List<PropertyKey<?>> CRAFT_META_POTION = List.of(
```
yannicklamprecht: I see that this is for validating which property can be used on which type of ItemStack.
Do we have a possibility to populate these kind of List dynamically using server implementation?

The dirty idea is to generate a default itemstack for each Materials once on startup, iterate over the nbt keys and check if a PropertyKey is applicable for that stack.

patches/api/0423-Paper-Item-API.patch
FawksX: key() is used for kyori, pretty sure paper still use get prefixes in API

patches/api/0423-Paper-Item-API.patch
MiniDigger: I would just argue that creation of such keys should be blocked somehow? only the api itself is allowed to create keys

Owen1212055: @MiniDigger 

I discussed it a bit in the paper contributing thread but I disagree with this. In general I think that in general devs being able to add their own keys does no harm, but it won’t be officially supported. You’ll need to register your own adapter for that key, which will require NMS. So it’s technically blocked normally but is still doable with access to sources.

patches/api/0423-Paper-Item-API.patch
```diff
@@ -0,0 +1,653 @@
...
++    PropertyKey<Value<Boolean>> RESOLVED = PropertyKey.of("resolved", Boolean.class);
++
++    // UnsignedBookMeta
++    PropertyKey<ListValue<String>> PAGES_WRITTEN = PropertyKey.ofList("pages", String.class);
++
++    // BlockStateMeta
++    PropertyKey<StoredPropertyHolder> BLOCK_ENTITY_TAG = PropertyKey.ofPropertyHolder("BlockEntityTag");
++
++    // AxolotlBucket
++    PropertyKey<Value<Axolotl.Variant>> VARIANT = PropertyKey.of("Variant", Axolotl.Variant.class);
```
kashike: This should be a more specific name, like `AXOLOTL_VARIANT`.

kashike (changes requested):

UntouchedOdin0:
Is this still being worked on at all? It looks really promising

Owen1212055:
For those wanting to play around, this now adds support for most ItemMeta stuff. Some isn't implemented yet...
Note: API item stacks are not supported nor tested yet.

Owen1212055:
No longer wrapped in values

patches/api/0419-Paper-Item-API.patch
```diff
@@ -0,0 +1,446 @@
...
++        return new PropertyKeyImpl<>(key, token);
++    }
++
++    static <T> PropertyKey<T> of(String key, TypeToken<T> token) {
++        return new PropertyKeyImpl<>(key, token.getType());
++    }
++
++    static <T> PropertyKey<T> of(String key, Type type) {
++        return new PropertyKeyImpl<>(key, type);
++    }
```
Machine-Maker: no public key creation

patches/api/0419-Paper-Item-API.patch
```diff
@@ -0,0 +1,446 @@
...
++++ b/src/main/java/io/papermc/paper/inventory/item/properties/TropicalFishVariant.java
+@@ -0,0 +1,9 @@
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.DyeColor;
++import org.bukkit.entity.TropicalFish;
++import org.jetbrains.annotations.ApiStatus;
++
++@ApiStatus.Experimental
++public record TropicalFishVariant(DyeColor patternColor, DyeColor bodyColor, TropicalFish.Pattern type) {
```
Machine-Maker: I think we settled on not exposing records directly to the API. This should be an interface with a package-private record impl (same of all of these). Using interfaces improves documentation and allows for easier, better changes without breaking API.

patches/server/0966-Paper-Item-API.patch
```diff
@@ -0,0 +1,5048 @@
...
++import org.bukkit.DyeColor;
++import org.bukkit.Material;
++import org.bukkit.block.banner.Pattern;
++import org.bukkit.inventory.meta.BannerMeta;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++import java.util.List;
++
++public class MetaBanner extends PaperMeta implements BannerMeta {
```
Machine-Maker: I don't like this naming scheme. I suggest sticking to what craftbukkit did, but with paper, so this would be `PaperMetaBanner`

patches/server/0966-Paper-Item-API.patch
```diff
@@ -0,0 +1,5048 @@
...
++
++        Material stateMaterial = (this.material != Material.SHIELD) ? this.material : shieldToBannerHack(holder); // Only actually used for jigsaws
++        if (holder != null) {
++            if (this.material == Material.SHIELD) {
++                holder.set(PaperMetaProperties.IDENTIFIER, "minecraft:banner");
++            } else if (this.material == Material.BEE_NEST || this.material == Material.BEEHIVE) {
++                holder.set(PaperMetaProperties.IDENTIFIER, "minecraft:beehive");
++            } else if (MaterialSetTag.SHULKER_BOXES.isTagged(this.material)) {
++                holder.set(PaperMetaProperties.IDENTIFIER, "minecraft:shulker_box");
++            }
```
Machine-Maker: As far as I can tell, these fixes are no longer needed. Vanilla always includes the block entity id in the itemstack nbt data now (as of a 1.19 snapshot)

Machine-Maker (changes requested):
Can you separate out the impls of the various meta types into a different patch? It makes review the diff easier. All except PaperMeta really.

patches/server/0966-Paper-Item-API.patch
```diff
@@ -0,0 +1,3586 @@
...
++            Material.LEATHER_BOOTS), MetaLeatherArmor::new);
++        register(List.of(Material.POTION,
++            Material.SPLASH_POTION,
++            Material.LINGERING_POTION,
++            Material.TIPPED_ARROW), MetaPotion::new);
++        register(Material.FILLED_MAP, MetaMap::new);
++        register(Material.FIREWORK_ROCKET, MetaFirework::new);
++        register(Material.FIREWORK_STAR, MetaFireworkEffect::new);
++        register(Material.ENCHANTED_BOOK, MetaEnchantmentStorage::new);
++        register(List.of(
```
Lulu13022002: This could use the item tags or is it not enough flexible ?

patches/server/0966-Paper-Item-API.patch
```diff
@@ -0,0 +1,3586 @@
...
++
++    @Override
++    public Position getType(Tag tag) {
++        BlockPos pos = NbtUtils.readBlockPos((CompoundTag) tag);
++        return Position.block(pos.getX(), pos.getY(), pos.getZ());
++    }
++
++    @Override
++    public Tag write(Position type) {
++        return NbtUtils.writeBlockPos(new BlockPos(type.blockZ(), type.blockY(), type.blockZ()));
```
Lulu13022002: wrong serialized coordinates: Z Y Z ?

Owen1212055:
Now uses SNBT for bukkit item serialization. Note that in the future we will probably move the count/material field out. Although, TBD.

```yml
test:
  ==: org.bukkit.inventory.ItemStack
  v: 3337
  type: PLAYER_HEAD
  meta:
    ==: ItemMeta
    meta-type: PAPER_META
    snbt: |-
      {
          Count: 1b,
          id: "minecraft:player_head",
          tag: {
              CustomModelData: 1,
              Enchantments: [
                  {
                      id: "minecraft:respiration",
                      lvl: 1s
                  }
              ],
              HideFlags: 2,
              SkullOwner: {
                  Id: [I; 1065279268, -1351007849, -1353122767, -455512511],
                  Name: "Owen1212055",
                  Properties: {
                      textures: [
                          {
                              Signature: "tlahGxBeyXBBe7FZ247LVUEzozL98moRaj1UFkAt5RSk73RvyAa/bv+JmHJi7orRtzoSKxF0+Zfx7t1jr2jrwIqptnuv1poxRrrdNR+rio9ni06D0NNxW/xJH2IflPRvpaRg2QdX9+ci8RnL1Wj6xyljFzu/8WK7FggqxkxHCrGf5OOTTWADMXphWS8hepn0+j+EXyQk3XVBT5H95CVY07Y3EpDymJ4/kMRoRRT6TqzQ0wYw2vxpHS6kzFnmqE12gV2SSQzCVxVcVmtikMmn80AzmJ3uU0PHZYz3oWU7H8OUoEwDQor+MHzJrteVFlMQhFV669QIOCKVwf0aAz5D4LvdZ8tW1ZNmBDUWtmAYizrbjKh7eGmxhfQL0U0XaBZPGjQPcpo/jrALqvOIqPrwdE+DiU060BfZuYi0uHMg/AdaVeyTHLSX3YlWfI+1UMbgPdZ0QxgE0AhMFw4RwuBGQTSUAWlhLOQJdYCI/mGtaO3uIFhiHhg9+r5VWk0gqJl+nVkvgi1f4uBr35JOX7xH9OhuVTaUB+Uen1gp05fbYjzDDkwgSdNv7RPTDC7ME8rwbIq3R8lflRUGRG98G0Q152kUjLpjcmKMwPupUHVNnM4HIpY6iMMIQZnj+hdWX5F6aDOEF7s8MibdoYhoHAFNgO9zdTha/bRJ6SIOrOFhsHA=",
                              Value: "ewogICJ0aW1lc3RhbXAiIDogMTY3OTk1MzExODA1NywKICAicHJvZmlsZUlkIiA6ICIzZjdlZGYyNGFmNzk0MTk3YWY1OGZjMzFlNGQ5NmU0MSIsCiAgInByb2ZpbGVOYW1lIiA6ICJPd2VuMTIxMjA1NSIsCiAgInNpZ25hdHVyZVJlcXVpcmVkIiA6IHRydWUsCiAgInRleHR1cmVzIiA6IHsKICAgICJTS0lOIiA6IHsKICAgICAgInVybCIgOiAiaHR0cDovL3RleHR1cmVzLm1pbmVjcmFmdC5uZXQvdGV4dHVyZS83OWY2OTc5NmU3N2U2ZmM0YWM0MWU0YTEyZmE1Mjc1YjcyNzlkMzZlM2I5NDI1NWVlYmI0NzgzNGZhNjJjYTZlIgogICAgfSwKICAgICJDQVBFIiA6IHsKICAgICAgInVybCIgOiAiaHR0cDovL3RleHR1cmVzLm1pbmVjcmFmdC5uZXQvdGV4dHVyZS9lN2RmZWExNmRjODNjOTdkZjAxYTEyZmFiYmQxMjE2MzU5YzBjZDBlYTQyZjk5OTliNmU5N2M1ODQ5NjNlOTgwIgogICAgfQogIH0KfQ=="
                          }
                      ]
                  }
              },
              Unbreakable: 1b,
              display: {
                  Lore: [
                      '{"text":"real"}'
                  ],
                  Name: '{"text":"GET REAL!"}'
              }
          }
      }
```

patches/api/0423-Paper-Item-API.patch
Owen1212055: Decided against in the end

patches/server/0950-Paper-Item-API.patch
Owen1212055: That's true, although that will most likely be far future.

patches/server/0950-Paper-Item-API.patch
Owen1212055: Not exactly, that was for something that was outdated. Generally, every property key can be applied to any stack. There is no "compatibility" with material types.

Owen1212055:
This now passes all tests.

Feel free to use the download jar to test and play around. Mostly, try to see how this works with plugins using the item meta api.

aerulion:
When testing with this build and our custom item plugin, I'm experiencing some "desync" of player head textures in creative mode. I need to do some further investigating what's exactly causing the issue, since we are doing a lot of item modifications during the InventoryCreativeEvent to provide custom lore text etc. for items taken out the creative inventory. But when I click a player head, the skull texture gets invisible, until I relog for example. That does not happen without our plugin, and does not happen with current "normal" paper builds.
But as I said, I need to first track down what exactly is causing the problem, which I probably won't have time for today. I will let you know as soon as I figure it out.

Owen1212055:
There was some issues with skullmeta, should be fixed, so would love to see how it works now.

Owen1212055:
The issue @aerulion describes should be resolved. This was an issue with meta migration logic.

aerulion:
The skull texture issue is fixed now, thank you!

But I noticed something else, that doesn't quite work:
Setting the custom color of a potion will result in the wrong nbt tag set. Also, the resulting nbt tags differ between this build and the current "normal" paper build.

The red item was generated using this build, the green one using the current normal paper build.
IWO7zVr

Code used to get the item:
```
    final @NotNull ItemStack itemStack = new ItemStack(Material.POTION);
    if (itemStack.getItemMeta() instanceof final @NotNull PotionMeta potionMeta) {
      potionMeta.setColor(Color.fromRGB(0x25a55f));
      itemStack.setItemMeta(potionMeta);
    }
    event.getPlayer().getInventory().addItem(itemStack);

```

Owen1212055:
@aerulion Thanks for the report! It should now correctly set the correct tag. Additionally, I also fixed an issue with the PublicBukkitValues being written when not needed. 

It seems the ``Potion`` tag is no longer written, and has the same behavior as a potion item with no tag. This will most likely be some of the (many) differences that this recode will bring. Because the old ItemMeta system wrote a lot of things when not needed.

Using /give command to give potion
image

What old ItemMeta would have given
image

aerulion:
I can't really say that it's fixed, the color tag is now no longer set in the display compound, but it's still set as `color` instead of `CustomPotionColor`.
I guess the empty effect tag not getting set is completely fine, as in vanilla / single player the give command results in an item without any tags too.

Owen1212055:
Didn't even realize it was a different property, oops.

aerulion:
The custom potion colors work correctly now, thanks!

Owen1212055:
This PR will not include entity property stuff initially most likely.

patches/server/0975-Paper-Item-API.patch
```diff
@@ -0,0 +1,4382 @@
...
++    public int getDamage() {
++        return this.propertyHolder.getOrDefault(DAMAGE, 0);
++    }
++
++    @Override
++    public void setDamage(int damage) {
++        this.propertyHolder.set(DAMAGE, damage);
++    }
++
++    // REpair meta
```
Phoenix616: **E**

Phoenix616:
Right now there seems to be no way to check if a certain item can hold certain data nor any validation on setting the data. Is that planned or intentionally not included and if so why not?

The thinking here is that with the ItemMeta system it is possible to instantly know which values can be set on a specific item and which can't which makes it not only easy to work with for beginners but also ensures that non-valid, potentially client-breaking data isn't set.

Owen1212055:
This is an intentional part of the property system. 
The system is not intended to be used for beginners. In fact, these keys are not going to be most likely backwards compatible, and will represent the most current internal representation of how items are stored. 

ItemMeta will be continued to be offered as a backwards compatible and more beginner friendly API to use instead. This property api allows for more fine tuning, and also allows us to move our meta implementations to not directly use NMS.

electronicboy:
Vanilla does not have a source of truth for what item stacks support what NBT tags, so there are no real means for adding validation

Owen1212055:
Support has been added for inlined armor trims (kinda)

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
+@@ -0,0 +1,23 @@
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.enchantments.Enchantment;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents a serialized enchantment with its level.
++ */
++public interface EnchantmentPair {
```
Machine-Maker: I think instead of a `List<EnchantmentPair>`, some other custom type should be created to interact with enchantments. While a "list" is what vanilla uses to store enchantments, it is not very API friendly. I think the list should be wrapped in a Map implementation.

You can preserve the list structure behind the scenes, but lazilly create a map using the first instance of an enchantment in the list (which is what vanilla does when looking for an enchantment).

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.attribute.Attribute;
++import org.bukkit.attribute.AttributeModifier;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents a serialized attribute that is grouped with a modifier.
++ */
++public interface AttributePair {
```
Machine-Maker: I think instead of a `List<AttributePair>`, the attributes should be exposed via a custom MultiMap implementation.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++import java.util.Set;
++
++@ApiStatus.Experimental
++public interface ItemProperties {
++
++    // COMMON
++    PropertyKey<List<EnchantmentPair>> ENCHANTMENTS = PropertyKey.of("Enchantments", TypeFactory.parameterizedClass(List.class, EnchantmentPair.class));
++    PropertyKey<Boolean> UNBREAKABLE = PropertyKey.of("Unbreakable", Boolean.class);
++    PropertyKey<List<Namespaced>> CAN_DESTROY_BLOCK_LIST = PropertyKey.of("CanDestroy", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
++    PropertyKey<List<Namespaced>> CAN_PLACE_ON_BLOCK_LIST = PropertyKey.of("CanPlaceOn", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
```
Machine-Maker: I would hold off on adding these 2 entirely, just don't add them at all, like comment them out until the BlockProperty API is merged.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
+
+diff --git a/build.gradle.kts b/build.gradle.kts
+index 149f9088fe806467656e8b1c4157df60fda69ba7..a90074480a3506f439cf8f38ee4b196053bcc96d 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -36,6 +36,7 @@ dependencies {
+     api("com.googlecode.json-simple:json-simple:1.1.1") {
+         isTransitive = false // includes junit
+     }
++    api("io.leangen.geantyref:geantyref:1.3.13") // Paper
```
Machine-Maker: I don't think its appropriate to expose this as an `api` dependency, rather just implementation. And no API method should return `TypeToken`, but rather the java `Type` instead. That way plugins can use whatever type token library (like gson) they want to.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<Boolean> UNBREAKABLE = PropertyKey.of("Unbreakable", Boolean.class);
++    PropertyKey<List<Namespaced>> CAN_DESTROY_BLOCK_LIST = PropertyKey.of("CanDestroy", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
++    PropertyKey<List<Namespaced>> CAN_PLACE_ON_BLOCK_LIST = PropertyKey.of("CanPlaceOn", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
++    PropertyKey<Set<ItemFlag>> HIDE_FLAGS = PropertyKey.of("HideFlags", TypeFactory.parameterizedClass(Set.class, ItemFlag.class));
++    PropertyKey<Integer> REPAIR_COST = PropertyKey.of("RepairCost", Integer.class);
++    PropertyKey<Integer> CUSTOM_MODEL_DATA = PropertyKey.of("CustomModelData", Integer.class);
++    PropertyKey<Integer> DAMAGE = PropertyKey.of("Damage", Integer.class);
++    PropertyKey<List<AttributePair>> ATTRIBUTES = PropertyKey.of("AttributeModifiers", TypeFactory.parameterizedClass(List.class, AttributePair.class));
++    PropertyKey<MutablePropertyView> BLOCK_STATE = PropertyKey.of("BlockStateTag", MutablePropertyView.class);
```
Machine-Maker: This also shouldn't be added until the BlockProperty API is merged with types that can better support what this actually does.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<MutablePropertyView> ENTITY_TAG = PropertyKey.of("EntityTag", MutablePropertyView.class);
++
++    // DISPLAY PROPERTIES
++    PropertyKey<MutablePropertyView> DISPLAY = PropertyKey.of("display", MutablePropertyView.class);
```
Machine-Maker: Is it worth exposing this directly? afaik only the 4 things below can ever be in it. I guess its useful for copying over stuff to another stack, but it just feels excessive to expose it.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    /**
++     * Creates a property key with the provided key and class.
++     *
++     * @param key   string key
++     * @param token class type
++     * @param <T>   type
++     * @return property key
++     */
++    @NotNull
++    static <T> PropertyKey<T> of(@NotNull String key, @NotNull Class<T> token) {
```
Machine-Maker: Still hard disagree that this should be exposed in any way to the API. We have the PDC for storing custom information

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++        return new PropertyKeyImpl<>(key, type);
++    }
++
++    /**
++     * Gets they key of this propertykey.
++     *
++     * @return key
++     */
++    @NotNull
++    String getKey();
```
Machine-Maker: Not even sure this String should be exposed to the API.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
+index 0000000000000000000000000000000000000000..d73a2276a3e56905e9595f85f5532f06f8dc583d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/property/MutablePropertyView.java
+@@ -0,0 +1,56 @@
++package io.papermc.paper.property;
++
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public interface MutablePropertyView extends PropertyView, Cloneable {
```
Machine-Maker: I'm not sure the words `Mutable` and `View` belong together in this context. Doesn't a view usually imply that you can't change it, but the underlying data can change so you are able to read changes?

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++import org.bukkit.inventory.meta.trim.ArmorTrim;
++import org.bukkit.potion.PotionData;
++import org.bukkit.potion.PotionEffect;
++import org.jetbrains.annotations.ApiStatus;
++
++import java.util.List;
++import java.util.Set;
++
++@ApiStatus.Experimental
++public interface ItemProperties {
```
Machine-Maker: I don't think the benefit of less `public static final` is worth having a public interface that can be extended regardless of annotations.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    // UnsignedBookMeta
++    PropertyKey<List<String>> PAGES_WRITTEN = PropertyKey.of("pages", TypeFactory.parameterizedClass(List.class, String.class));
++
++    // BlockStateMeta
++    PropertyKey<MutablePropertyView> BLOCK_ENTITY_TAG = PropertyKey.of("BlockEntityTag", MutablePropertyView.class);
++    // Some odd entity tags included in item meta
++    PropertyKey<NamespacedKey> NOTE_BLOCK_SOUND = PropertyKey.of("note_block_sound", NamespacedKey.class);
++
++    // AxolotlBucket
++    PropertyKey<Axolotl.Variant> VARIANT = PropertyKey.of("Variant", Axolotl.Variant.class);
```
Machine-Maker: Missing properties for `HuntingCooldown`.

Technically, this is also missing properties for `NoAI`, `Silent`, `NoGravity`, `Glowing`, `Invulnerable`, and `Health`, but those could also be set via the entity tag. But then again, they are completely valid properties as far as the game is concerned.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<Component> DISPLAY_NAME = PropertyKey.of("Name", Component.class);
++    PropertyKey<List<? extends Component>> LORE = PropertyKey.of("Lore", TypeFactory.parameterizedClass(List.class, Component.class));
++    PropertyKey<Color> COLOR = PropertyKey.of("color", Color.class);
++    PropertyKey<Color> MAP_COLOR = PropertyKey.of("MapColor", Color.class);
++
++    // Armor Meta
++    PropertyKey<ArmorTrim> ARMOR_TRIM = PropertyKey.of("Trim", ArmorTrim.class);
++
++    // TropicalFishBucketMeta
++    PropertyKey<TropicalFishVariant> TROPICAL_FISH_BUCKET_VARIANT = PropertyKey.of("BucketVariantTag", TropicalFishVariant.class);
```
Machine-Maker: Technically, this is also missing properties for `NoAI`, `Silent`, `NoGravity`, `Glowing`, `Invulnerable`, and `Health`, but those could also be set via the entity tag. But then again, they are completely valid properties as far as the game is concerned.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    // AxolotlBucket
++    PropertyKey<Axolotl.Variant> VARIANT = PropertyKey.of("Variant", Axolotl.Variant.class);
++
++    // MusicInstrumentMeta
++    PropertyKey<MusicInstrument> INSTRUMENT = PropertyKey.of("instrument", MusicInstrument.class);
++
++    // BannerMeta
++    PropertyKey<DyeColor> SHIELD_BANNER_BASE = PropertyKey.of("Base", DyeColor.class);
++
++}
```
Machine-Maker: Ok, so some more missing properties are `NoAI`, `Silent`, `NoGravity`, `Glowing`, `Invulnerable`, and `Health` for the other mob bucket mobs like pufferfish, salmon, cod. 

Tadpole is also missing the above ones as well as `Age`.

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.potion.PotionEffectType;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Represents a serialized stew effect with a duration.
++ */
++public interface StewEffect {
```
Machine-Maker: I think similarly to enchantments and attributes, it'd be best to expose a custom MultiMap implementation of stew effects

patches/api/0421-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    /**
++     * Clears the property view.
++     */
++    void clear();
++}
+diff --git a/src/main/java/io/papermc/paper/property/PropertyKey.java b/src/main/java/io/papermc/paper/property/PropertyKey.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..4d02e565476a04cdccf78d700a14aa45586f7233
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/property/PropertyKey.java
```
Machine-Maker: I'm not sure its a good idea to generalize this much. I understand that both the item and later entity property APIs will be similar, but they shouldn't be interchangeable, like I shouldn't be able to put an item property on an entity and vice-versa. I think they should have specific types that aren't really related at all inheritance-wise.

Machine-Maker (changes requested):
Generally I think the property stuff should be more specifc to items.

`PropertyKey` -> `ItemPropertyKey`
`PropertyView` -> `ItemPropertyHolder`
And then ItemStack can implement `ItemPropertyHolder.

patches/api/0421-Paper-Item-API.patch
Owen1212055: I slightly disagree, I had this at some point but I really want to stick to the NMS implementation here. These properties are supposed to mirror and not provide much leeway. But am open to changing this, just not sure.

patches/api/0421-Paper-Item-API.patch
Owen1212055: We’d still need the impl for item meta, so I’d be made private

patches/api/0421-Paper-Item-API.patch
Owen1212055: It’s needed so you can actually write to the display tag. Since we aren’t letting people make their own.

patches/api/0421-Paper-Item-API.patch
Owen1212055: Yeah no this won’t be public.

patches/api/0421-Paper-Item-API.patch
Machine-Maker: Yeah ok, can just move them to the server or make them private or whatever.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
+index 0000000000000000000000000000000000000000..d73a2276a3e56905e9595f85f5532f06f8dc583d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/property/MutablePropertyView.java
+@@ -0,0 +1,56 @@
++package io.papermc.paper.property;
++
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public interface MutablePropertyView extends PropertyView, Cloneable {
```
Machine-Maker, line 335: This concept should not be introduced in the API. It doesn't make sense to have MutablePropertyView properties which can set any random key on them.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.attribute.Attribute;
++import org.bukkit.attribute.AttributeModifier;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents a serialized attribute that is grouped with a modifier.
++ */
++public interface AttributePair {
```
Machine-Maker, line 79: Need a static create method; record impl should be in the API (package-private)

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
+@@ -0,0 +1,23 @@
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.enchantments.Enchantment;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents a serialized enchantment with its level.
++ */
++public interface EnchantmentPair {
```
Machine-Maker, line 108: Need a static create method; record impl should be in the API (package-private)

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<Boolean> UNBREAKABLE = PropertyKey.of("Unbreakable", Boolean.class);
++    PropertyKey<List<Namespaced>> CAN_DESTROY_BLOCK_LIST = PropertyKey.of("CanDestroy", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
++    PropertyKey<List<Namespaced>> CAN_PLACE_ON_BLOCK_LIST = PropertyKey.of("CanPlaceOn", TypeFactory.parameterizedClass(List.class, Namespaced.class)); // TODO: BETTER REPRESENTATION
++    PropertyKey<Set<ItemFlag>> HIDE_FLAGS = PropertyKey.of("HideFlags", TypeFactory.parameterizedClass(Set.class, ItemFlag.class));
++    PropertyKey<Integer> REPAIR_COST = PropertyKey.of("RepairCost", Integer.class);
++    PropertyKey<Integer> CUSTOM_MODEL_DATA = PropertyKey.of("CustomModelData", Integer.class);
++    PropertyKey<Integer> DAMAGE = PropertyKey.of("Damage", Integer.class);
++    PropertyKey<List<AttributePair>> ATTRIBUTES = PropertyKey.of("AttributeModifiers", TypeFactory.parameterizedClass(List.class, AttributePair.class));
++    PropertyKey<MutablePropertyView> BLOCK_STATE = PropertyKey.of("BlockStateTag", MutablePropertyView.class);
```
Machine-Maker, line 169: This should not be `MutablePropertyView` but instead be `BlockPropertyHolder` (once the BlockProperty API is merged). This is just a collection of block state properties and values.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<MutablePropertyView> ENTITY_TAG = PropertyKey.of("EntityTag", MutablePropertyView.class);
```
Machine-Maker, line 170: I would hold off on exposing this, the type for this will be some `EntityPropertyHolder` or whatever for entities.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++
++    // DISPLAY PROPERTIES
++    PropertyKey<MutablePropertyView> DISPLAY = PropertyKey.of("display", MutablePropertyView.class);
```
Machine-Maker, line 173: This should not have a property. Instead, the specific parts of display should have their own properties which point to `display.Name`, `display.Lore`, etc.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<Integer> MAP_ID = PropertyKey.of("map", Integer.class);
++    PropertyKey<Byte> MAP_SCALE_DIRECTION = PropertyKey.of("map_scale_direction", Byte.class);
++    // Unused, really. Marks the map to be locked when crafted
++    PropertyKey<Boolean> MAP_TO_LOCK = PropertyKey.of("map_to_lock", Boolean.class);
++
++    // KnowledgeBookMeta
++    PropertyKey<List<NamespacedKey>> RECIPES = PropertyKey.of("Recipes", TypeFactory.parameterizedClass(List.class, NamespacedKey.class));
++
++    // FireworkMeta
++    PropertyKey<MutablePropertyView> FIREWORKS = PropertyKey.of("Fireworks", MutablePropertyView.class);
```
Machine-Maker, line 209: Same thing applies here, only have properties which point directly to specific attributes of fireworks

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    PropertyKey<List<Component>> PAGES = PropertyKey.of("pages", TypeFactory.parameterizedClass(List.class, Component.class));
++    PropertyKey<List<Component>> PAGES_FILTERED = PropertyKey.of("filtered_pages", TypeFactory.parameterizedClass(List.class, Component.class));
++    PropertyKey<BookMeta.Generation> GENERATION = PropertyKey.of("generation", BookMeta.Generation.class);
++    PropertyKey<Boolean> RESOLVED = PropertyKey.of("resolved", Boolean.class);
++
++    // UnsignedBookMeta
++    PropertyKey<List<String>> PAGES_WRITTEN = PropertyKey.of("pages", TypeFactory.parameterizedClass(List.class, String.class));
++
++    // BlockStateMeta
++    PropertyKey<MutablePropertyView> BLOCK_ENTITY_TAG = PropertyKey.of("BlockEntityTag", MutablePropertyView.class);
```
Machine-Maker, line 244: The type for this should be a `TileState`

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.potion.PotionEffectType;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Represents a serialized stew effect with a duration.
++ */
++public interface StewEffect {
```
Machine-Maker, line 273: Need a static create method; record impl should be in the API (package-private)

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++package io.papermc.paper.inventory.item.properties;
++
++import org.bukkit.DyeColor;
++import org.bukkit.entity.TropicalFish;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents a serialized tropical fish.
++ */
++public interface TropicalFishVariant {
```
Machine-Maker, line 304: Need a static create method; record impl should be in the API (package-private)

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++    /**
++     * Creates a property key with the provided key and class.
++     *
++     * @param key   string key
++     * @param token class type
++     * @param <T>   type
++     * @return property key
++     */
++    @NotNull
++    static <T> PropertyKey<T> of(@NotNull String key, @NotNull Class<T> token) {
```
Machine-Maker, line 417: No public creation

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++
++import java.lang.reflect.Type;
++
++/**
++ * Represents a key that holds a property.
++ *
++ * @param <T> type
++ */
++@ApiStatus.Experimental
++public interface PropertyKey<T> {
```
Machine-Maker, line 406: Probably needs to be item-specific. So `ItemPropertyKey`.

patches/api/0457-Paper-Item-API.patch
```diff
@@ -0,0 +1,582 @@
...
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Represents an object that can hold properties.
++ */
++@ApiStatus.Experimental
++public interface PropertyView {
```
Machine-Maker, line 508: item-specific

patches/server/1058-Paper-Item-API.patch
```diff
@@ -0,0 +1,4571 @@
...
++    @Override
++    default @NotNull String getDisplayName() {
++        return LegacyComponentSerializer.legacySection().serializeOr(this.displayName(), "");
++    }
++
++    @Override
++    default void setDisplayName(@Nullable String name) {
++        this.displayName(
++            LegacyComponentSerializer.legacySection().deserializeOr(name, Component.empty())
++                .decorationIfAbsent(TextDecoration.ITALIC, TextDecoration.State.FALSE)
```
Machine-Maker, line 300: Is there a reason we are changing this to use adventure's serializer instead of just sticking with CraftChatMessage? Just feels unneeded and you have to handle specific quirks ourselves, like the italic stuff here.

Machine-Maker (changes requested):
The server implementation of the properties feels needlessly complex. There doesn't have to be separate classes for every type that needs to be converted, like 90% of them are just calling 2 methods anyway.

I'm also not following the need for "accesses". 

For getting via an item property: An item property key be attached to some adapter/converter/whatever which takes has a method which takes in a CompoundTag and the key, goes to the path that key has, be in `display.Lore`, `CanPlaceOn`, whatever, and parse the value found into the correct API type and return it.

For setting via an item property: You just have a second method on that adapter which takes a CompoundTag, the key and the api type and converts it to nbt tags and puts that tag at the specified path in that compound tag.

CKATEPTb:
Any news?

Owen1212055:
This PR is officially blocked by snapshot changes as item components will cause this to be reimplemented. Most likely this PR will be repurposed completely.

Owen1212055:
Superceded by https://github.com/PaperMC/Paper/pull/10613

PR #8733, opened 2022-12-28
Fix collision between AABB and a dot

### LINKED ISSUE #8038: Markers spawned from spawn eggs spawn one block lower than they should
### Expected behavior

On vanilla servers, spawning a marker from a spawn egg spawns it directly on top of the block selected.

### Observed/Actual behavior

On a paper server, the marker spawns one block lower—on the bottom face of the block selected.
Note that this only happens when you right-click on the top face of a block—all other faces work as intended.

### Steps/models to reproduce

To reproduce easily, you can use a command like this:
`/give @p tropical_fish_spawn_egg{display:{Name:'{"text":"Game Start Button Marker","italic":false}'},EntityTag:{id:"minecraft:marker",Tags:["gameStartButton"]}} 1`
This gives you a spawn egg that will spawn a tagged marker for mapmaking purposes.

Next, put this command in a repeating command block to visualize the markers with particles.
`execute as @e[type=marker] at @s run particle minecraft:ash ~ ~ ~ 0.01 0.01 0.01 0 10 force`

Then, use the spawn egg on the top of a block. The marker will spawn inside that block—on its bottom face instead of its top face.

Try the same thing in vanilla and you'll see that the marker spawns on the top face of the block as expected.

### Plugin and Datapack List

Worldedit.

### Paper version

This server is running Paper version git-Paper-33 (MC: 1.19) (Implementing API version 1.19-R0.1-SNAPSHOT) (Git: 147a2b7)

### Other

_No response_

Lulu13022002:
Closes #8038 
Vanilla doesn't check if the target box is empty (a dot) only if the source box is empty to then fallback to the default max distance (-2 here) but paper do. But even if the marker has an empty box, the distance must still be calculated between where the marker has been default placed (one block above) and the top of the block (where the dot will collide) which is -1

### DIFF
--- a/patches/server/0746-Highly-optimise-single-and-multi-AABB-VoxelShapes-an.patch
+++ b/patches/server/0746-Highly-optimise-single-and-multi-AABB-VoxelShapes-an.patch
@@ -975,7 +975,7 @@ index 0000000000000000000000000000000000000000..a87f6380b2c387fb0cdd40d5087b5c93
 +}
 diff --git a/src/main/java/io/papermc/paper/voxel/AABBVoxelShape.java b/src/main/java/io/papermc/paper/voxel/AABBVoxelShape.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..d67a40e7be030142443680c89e1763fc9ecdfe0a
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
Can confirm that this does fix the issue.

Machine-Maker (approved):

PR #8744, opened 2022-12-30
Fix warden spawn reason from DEFAULT to NATURAL

### LINKED ISSUE #8283: Warden spawned by shrieker uses DEFAULT SpawnReason

Kenuux:
This fix is trivial. fixes #8283


patches/server/0958-Fix-warden-spawn-reason-from-DEFAULT-to-NATURAL.patch
```diff
@@ -0,0 +1,27 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/block/entity/SculkShriekerBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/SculkShriekerBlockEntity.java
+index 0554afa72ae6d2f01886916031c8b90b0d45eb52..a96f66e7a79078163214409c29709b0e331482cb 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/SculkShriekerBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/SculkShriekerBlockEntity.java
+@@ -36,6 +36,7 @@ import net.minecraft.world.level.gameevent.GameEvent;
+ import net.minecraft.world.level.gameevent.GameEventListener;
+ import net.minecraft.world.level.gameevent.vibrations.VibrationListener;
+ import net.minecraft.world.phys.Vec3;
++import org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason;
```
NoahvdAa: Use the fully qualified name instead of adding an import

NoahvdAa (commented):
👋 Welcome to Paper, congrats on making your first PR!

patches/server/0958-Fix-warden-spawn-reason-from-DEFAULT-to-NATURAL.patch
Kenuux: Done

Doc94:
Looks good, maybe this change can be included in [server/0939-Fixes-and-additions-to-the-SpawnReason-API.patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0939-Fixes-and-additions-to-the-SpawnReason-API.patch) ?

electronicboy:
Natural would imply that the mob was spawned by the natural mob spawning logic, that doesn't look to be the case here

The original issue appears to be WAI outside of the spawn reason shouldn't be default, but, I don't think that it should be "natural" either. There is maybe some wider issue around what all counts towards the limit, the years of butchering to the mob spawning and persistence logic really creates a lot of messes around how mob spawn limits work

Kenuux:
> Natural would imply that the mob was spawned by the natural mob spawning logic, that doesn't look to be the case here
> 
> The original issue appears to be WAI outside of the spawn reason shouldn't be default, but, I don't think that it should be "natural" either. There is maybe some wider issue around what all counts towards the limit, the years of butchering to the mob spawning and persistence logic really creates a lot of messes around how mob spawn limits work

Yeah, I thought so too. But then I saw that there used to be this pull request (https://github.com/PaperMC/Paper/pull/8296) that was closed by the author for no reason. Based on the comments, I assumed that we just take NATURAL now.

Machine-Maker (approved):

PR #8750, opened 2023-01-02
Make PlayerPurchaseEvent respect result item

fr3qu3ncy-dev:
When modifying a trade using setTrade() in PlayerPurchaseEvent or PlayerTradeEvent, the result item of that new trade is completely ignored. With this change, you can easily manipulate villager trading from within these events, because it sets the clicked item (which is the result item in MerchantResultSlot) to the new result item.

### DIFF
--- a/patches/server/0958-Make-PlayerTradeEvent-respect-result-item.patch
+++ b/patches/server/0958-Make-PlayerTradeEvent-respect-result-item.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: fr3qu3ncy <kilicbatuhan1205@gmail.com>
+Date: Sun, 1 Jan 2023 23:34:10 +0100
+Subject: [PATCH] Make PlayerTradeEvent respect result item
+
+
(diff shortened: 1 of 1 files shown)


Machine-Maker:
I think this won’t work for some inventory actions. Like shift clicking? Just mutating the stack object isn’t enough for that.

fr3qu3ncy-dev:
@Machine-Maker Also tested with shift clicking, working fine.

Machine-Maker:
@fr3qu3ncy-dev shift clicking did not work at all. I set the trade to a custom trade in PlayerTradeEvent, shift-clicked the result slot to move the result into my inventory, and the server actually hung in an infinite loop somewhere.

Machine-Maker (changes requested):
See my comment about the places this is broken.

fr3qu3ncy-dev:
@Machine-Maker What did you set the recipe to?
villager_working

Machine-Maker:
Set it to something that doesn’t consume the ingredients. I basically changed all trades to diamond for a dirt. Put in emeralds, and shift clicking out didn’t work.

PR #8752, opened 2023-01-02
don't warn about offline mode

NonSwag:
Me and many other users of paper are very annoyed by the warning when starting our servers in offline mode
we all are aware of the possible consequences and therefore using velocity in online mode
the message is very useless with a server behind velocity because players can't exploit this without knowing the security key
This patch will not print the warning when velocity is enabled and set to online

### DIFF
--- a/patches/server/0957-don-t-warn-about-offline-mode.patch
+++ b/patches/server/0957-don-t-warn-about-offline-mode.patch
@@ -1,0 +1,35 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: NonSwag <mrminecraft00@gmail.com>
+Date: Mon, 2 Jan 2023 10:35:23 +0100
+Subject: [PATCH] don't warn about offline mode
+
+
+diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+index 51b3db0b6c2cede95b584268e035c0fb36d38094..95d5c4ab5610d3b5b4cc7ce1fbe6a75031888c5b 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -1,5 +1,6 @@
+ package net.minecraft.server.dedicated;
+ 
++import com.destroystokyo.paper.PaperConfig;
+ import com.google.common.collect.Lists;
+ import com.mojang.authlib.GameProfile;
+ import com.mojang.datafixers.DataFixer;
+@@ -19,6 +20,8 @@ import java.util.Locale;
+ import java.util.Optional;
+ import java.util.function.BooleanSupplier;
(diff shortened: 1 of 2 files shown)


electronicboy:
I could see downgrading the messages from a warn in such cases, but, given
that this is the only indicator that the server is even configured for a
proxy properly in the logs, I'm inclined to say no

kennytv:
I'd make it a different log message in that case, informing you that you are running offline mode with a proxy configured and that you still might want to take measures to prevent direct connections to the backend server

electronicboy:
Yea, that's kinda what I was thinking, best of both worlds

NonSwag:
Paper does already prevent players from connecting directly to the server if velocity is enabled
so I see no point in any kind of information about that

electronicboy:
I want something in the logs to still indicate that a server has been
configured for a proxy, given that I can hardly go a week without using the
absence of that line in the logs in order to spot an issue with some
servers setup, it's just something which noting in the logs is useful.

It doesn't need to be the full warning, just generally a note that the
server is configured for modern forwarding

NonSwag:
there you go

patches/server/0957-don-t-warn-about-offline-mode.patch
```diff
@@ -0,0 +1,35 @@
...
+
+
+diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+index 51b3db0b6c2cede95b584268e035c0fb36d38094..95d5c4ab5610d3b5b4cc7ce1fbe6a75031888c5b 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -1,5 +1,6 @@
+ package net.minecraft.server.dedicated;
+ 
++import com.destroystokyo.paper.PaperConfig;
```
Doc94, line 14: This can be better a inline-import.

patches/server/0957-don-t-warn-about-offline-mode.patch
```diff
@@ -0,0 +1,35 @@
...
+ import com.google.common.collect.Lists;
+ import com.mojang.authlib.GameProfile;
+ import com.mojang.datafixers.DataFixer;
+@@ -19,6 +20,8 @@ import java.util.Locale;
+ import java.util.Optional;
+ import java.util.function.BooleanSupplier;
+ import javax.annotation.Nullable;
++
++import io.papermc.paper.configuration.GlobalConfiguration;
```
Doc94, line 23: This can be better a inline-import.

Doc94:
Maybe all this can be a unique patch and not two?

patches/server/0957-don-t-warn-about-offline-mode.patch
NonSwag: okay, but why?

Doc94: well this or add the // Paper and remove the new line.. in the most of cases its better a patch with less diffs for this its better the inline import.

PR #4611, opened 2020-10-04
Fix item locations dropped from campfires

BillyGalbreath:
Fixes #4259 by not flooring the blockposition among other weirdness

PR #8754, opened 2023-01-02
Fix campfire recipes not always outputting full result

Machine-Maker:
[This fix](https://github.com/PaperMC/Paper/pull/4611) did not include the while(notEmpty) look to drop all of the result stack. If you try to create a recipe with an output stack greater than 10, it might not produce all of it.

### DIFF
--- a/patches/server/0470-Fix-item-locations-dropped-from-campfires.patch
+++ b/patches/server/0470-Fix-item-locations-dropped-from-campfires.patch
@@ -4,21 +4,25 @@ Date: Sat, 3 Oct 2020 20:32:25 -0500
 Subject: [PATCH] Fix item locations dropped from campfires
 
 Fixes #4259 by not flooring the blockposition among other weirdness
+Vanilla Issue: MC-267622
 
 diff --git a/src/main/java/net/minecraft/world/level/block/entity/CampfireBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/CampfireBlockEntity.java
-index c4cf6abf0a962794ddbb4d7a691406054062ffee..24e2063db933bfbc8fc1f34edb8106ae4d7c633c 100644
+index c4cf6abf0a962794ddbb4d7a691406054062ffee..f706c787f7608f7440a5f5e05e7e9c4cb582368c 100644
 --- a/src/main/java/net/minecraft/world/level/block/entity/CampfireBlockEntity.java
 +++ b/src/main/java/net/minecraft/world/level/block/entity/CampfireBlockEntity.java
-@@ -82,7 +82,11 @@ public class CampfireBlockEntity extends BlockEntity implements Clearable {
+@@ -82,7 +82,14 @@ public class CampfireBlockEntity extends BlockEntity implements Clearable {
                          result = blockCookEvent.getResult();
                          itemstack1 = CraftItemStack.asNMSCopy(result);
                          // CraftBukkit end
 -                        Containers.dropItemStack(world, (double) pos.getX(), (double) pos.getY(), (double) pos.getZ(), itemstack1);
-+                    // Paper start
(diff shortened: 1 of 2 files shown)


patches/server/0492-Fix-item-locations-dropped-from-campfires.patch
```diff
@@ -6,19 +6,21 @@ Subject: [PATCH] Fix item locations dropped from campfires
...
 -                        Containers.dropItemStack(world, (double) pos.getX(), (double) pos.getY(), (double) pos.getZ(), itemstack1);
-+                    // Paper start
-+                    net.minecraft.world.entity.item.ItemEntity droppedItem = new net.minecraft.world.entity.item.ItemEntity(world, pos.getX() + 0.5D, pos.getY() + 0.5D, pos.getZ() + 0.5D, itemstack1.split(world.random.nextInt(21) + 10));
-+                    droppedItem.setDeltaMovement(world.random.nextGaussian() * 0.05D, world.random.nextGaussian() * 0.05D + 0.2D, world.random.nextGaussian() * 0.05D);
-+                    world.addFreshEntity(droppedItem);
-+                    // Paper end
++                        // Paper start
++                        while (!itemstack1.isEmpty()) {
++                            net.minecraft.world.entity.item.ItemEntity droppedItem = new net.minecraft.world.entity.item.ItemEntity(world, pos.getX() + 0.5D, pos.getY() + 0.5D, pos.getZ() + 0.5D, itemstack1.split(world.random.nextInt(21) + 10));
++                            droppedItem.setDeltaMovement(world.random.nextGaussian() * 0.05D, world.random.nextGaussian() * 0.05D + 0.2D, world.random.nextGaussian() * 0.05D);
```
Lulu13022002: The velocity in recent version has changed here if it doesn't cause trouble with the original issue it should stay like vanilla

Machine-Maker: Yeah updated it to use the new setDeltaMovement info. Do you have any idea where that big long number comes from? I couldn't find a fraction or something nicer than it to change it to.

MWHunter: My guess is 0.11485F

Lulu13022002: In order to make the game more predictable some random events has been changed to use a triangle distribution instead of the Gaussian distribution to avoid extreme case (outside of -1 and 1). This transform the following calculation:
gauss(mean=0, deviation=1) * scale + base = triangle(mode, derivation)
The important part is:
derivation = 2.297 * scale.
mode = base
That been said i don't know why they take this number to calculate their derivation from the mean :/ probably to keep some consistency with the old behavior. If you apply those formulas to the code we get
float j = 0.05F;
world.random.triangle(0.0D, j * 2.297D)
But since mojang has saved 0.05 as a float in memory there's some rounding issue that appears (on top of how Java represent decimal number) 
so instead of 0.11485000000000001 we gets 0.11485000171139836 (those errors doesn't appear for the others triangle calculation).
If you want to keep the vanilla behavior (with this error) we needs to keep the float.
The exact value would be 0.11485 without any error.

Lulu13022002: I found where the spread constant was, as for the error this seems to be the usual binary representation error caused by a float multiply by a double. The constant is RandomSource.GAUSSIAN_SPREAD_FACTOR (a double). So you can simplify this by:
double deviation = 0.05F * GAUSSIAN_SPREAD_FACTOR;
in the end this will produce the same number.

Machine-Maker: Ok, updated to that. 

It's been so long, do we care that the x, y, z coords for the ItemEntity don't follow the exact calculations from `Containers#dropItemStack`? It's missing the stuff involving the item entity width.

Owen1212055 (approved):

PR #8755, opened 2023-01-02
add config for velocity secret key file

### LINKED ISSUE #6332: Add an environment variable for velocity secret

Misterio77:
Adds a `proxies.velocity.secret-file` option, making it possible to specify a file containing the velocity secret file instead of including it in the yml. Similarly to https://github.com/PaperMC/Velocity/pull/712.

Should work nicely, here's a couple of questions:
- Should I bump the config version to 29?
- Should we somehow deprecate the previous option? For now, I've added a log warning for it.

Fixes https://github.com/PaperMC/Paper/issues/6332

### DIFF
--- a/patches/server/0957-add-config-for-velocity-secret-key-file.patch
+++ b/patches/server/0957-add-config-for-velocity-secret-key-file.patch
@@ -1,0 +1,96 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gabriel Fontes <hi@m7.rs>
+Date: Mon, 2 Jan 2023 17:32:38 -0300
+Subject: [PATCH] add config for velocity secret key file
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java b/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
+index c4934979b1ed85bfc4f8d9e6f8848b2beaad95c3..6703468b0c0547502aee8c788d77a624c41d56e1 100644
+--- a/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
++++ b/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
+@@ -4,6 +4,7 @@ import io.papermc.paper.configuration.GlobalConfiguration;
+ import com.google.common.net.InetAddresses;
+ import com.mojang.authlib.GameProfile;
+ import com.mojang.authlib.properties.Property;
++import java.io.IOException;
+ import java.net.InetAddress;
+ import java.security.InvalidKeyException;
+ import java.security.MessageDigest;
+@@ -33,12 +34,12 @@ public class VelocityProxy {
+ 
+         try {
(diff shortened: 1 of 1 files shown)


tomcraft:
You should edit the patch adding Velocity support rather than adding a new one.

electronicboy:
And config file stuff should stay in the config patch

If we're going to force people to move over, we should be handling that as
a migration, but, I'm not sure we should force people to use an external
file

patches/server/0957-add-config-for-velocity-secret-key-file.patch
```diff
@@ -0,0 +1,96 @@
...
++            public String secretFile = "";
++            public String getSecret() throws IOException {
++                if (!this.secret.isEmpty()) {
++                    return this.secret;
++                } else if (!this.secretFile.isEmpty()) {
++                    return String.join("", Files.readAllLines(Path.of(this.secretFile)));
++                } else {
++                    return "";
++                }
++            }
```
Machine-Maker, line 61: I'm not a fan of adding unneeded methods to the config class. Preferably it's just fields. Perhaps we should have a custom type to handle secret in text or secret in file?

patches/server/0957-add-config-for-velocity-secret-key-file.patch
```diff
@@ -0,0 +1,96 @@
...
+
+diff --git a/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java b/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
+index c4934979b1ed85bfc4f8d9e6f8848b2beaad95c3..6703468b0c0547502aee8c788d77a624c41d56e1 100644
+--- a/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
++++ b/src/main/java/com/destroystokyo/paper/proxy/VelocityProxy.java
+@@ -4,6 +4,7 @@ import io.papermc.paper.configuration.GlobalConfiguration;
+ import com.google.common.net.InetAddresses;
+ import com.mojang.authlib.GameProfile;
+ import com.mojang.authlib.properties.Property;
++import java.io.IOException;
```
Machine-Maker, line 15: Inline the imports (same with other imports into nms classes)

Machine-Maker (changes requested):

robbelouwet:
Hi there, what's the status on this?

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/10127

PR #8766, opened 2023-01-06
Use pub static field for deprecated name

Machine-Maker:
The undeprecated, newly named item flag should be part of the enum with the old, now-deprecated name being the public static final field. I tested this with a spigot plugin compiled to use `ItemFlag.HIDE_POTION_EFFECTS` and it worked fine without any bytecode rewriting. I assume that's because enum entries are pretty much equivalent to public static final fields. So just switching the names around shouldn't cause any ABI/API problems.


Machine-Maker:
We could also add a bytecode rewrite for the Enum#valueOf method, but I'm not sure how important that is. All of the other enums which have changed values don't have one (except for DyeColor).

PR #8767, opened 2023-01-07
Fix some explosion damage not scaling with difficulty

### LINKED ISSUE #8735: End crystal and TNT explosions do same damage regardless of difficulty
### Expected behavior

https://bugs.mojang.com/browse/MC-258535

### Observed/Actual behavior

https://bugs.mojang.com/browse/MC-258535

### Steps/models to reproduce

https://bugs.mojang.com/browse/MC-258535

### Plugin and Datapack List

https://bugs.mojang.com/browse/MC-258535

### Paper version

#365

### Other

_No response_

Machine-Maker:
Fixes MC-258535
Fixes https://github.com/PaperMC/Paper/issues/8735

----
If we wanted to restrict the fix to those specific cases, I could add the override to `IndirectEntityDamageSource` which is the only `EntityDamageSource` type which has the `setScalesWithDifficulty()` method called.

### DIFF
--- a/patches/server/0902-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0902-Fix-a-bunch-of-vanilla-bugs.patch
@@ -40,6 +40,10 @@ https://bugs.mojang.com/browse/MC-257875
   by: Jake Potrebic <jake.m.potrebic@gmail.com>
   Fixes server not consuming fire charges when igniting a creeper
 
+https://bugs.mojang.com/browse/MC-258535
+  by: Jake Potrebic <jake.m.potrebic@gmail.com>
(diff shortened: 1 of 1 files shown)


Chris6ix:
Misspelled your name ;)

Lulu13022002:
We should fix this in a general case to avoid issue with the new DamageOrigin API for custom origin

Machine-Maker:
> We should fix this in a general case to avoid issue with the new DamageOrigin API for custom origin

Ok, that's what this PR does then. It checks if the field scalesWithDifficulty has been set, and if it has, overrides the check in EntityDamageSource.

kennytv (approved):

PR #8775, opened 2023-01-09

### LINKED ISSUE #8157: Goat Horn API

There is seemingly no class or method in the API to get the Type of a Goat Horn 

Create a new subclass of ItemMeta which exposes a getter and setter for the goat horn "instrument" type

Janky NMS or NBT hacks? idk

### Other

_No response_

byquanton:
Removed 0956-Return-0-by-default-for-FireworkMeta Patch because it was fixed upstream.


Machine-Maker:
You can just delete the patch, no need to move it to the removed folder. That folder is for other types of "removed" patches, like patches that only applied to previous versions, or patches that don't make sense on a codebase level anymore. Not simple patches like this which upstream fixed.

byquanton:
Should I create a second commit or amend the commit?

olijeffers0n:
Approve Workflow?

Owen1212055:
Should be noted for future use that usually only the development team is responsible for pulling upstream. 👍

electronicboy (approved):

electronicboy:
This PR was created improperly, i.e. the patches haven't been rebuilt, which will cause issues; given that this PR was created from master, that means that we can't easily fix this on our side

PR #8781, opened 2023-01-11
Expand Pose API

### LINKED ISSUE #7016: Allow paper API to set player pose

So, there is an api call to show the player position. But there is little to non control over player pose. I was trying to make a plugin heavily dependent on player swimming pose, so i googled it. After a short reaserch i found that there is no way to force a player pose other than sending a fake block packet. There is a setSwimming function but it does not work. There is a function in craftplayer to set pose. I used it in the past but even that set player position for 1 tick. There is curently no way to force player position.

I would like to see some kind of api call that woudl set player position (just use setPose() from craftplayer). Optional i woudl like to see some kind of api call that woudl disable pose checking on every tick. Overall i think that it woudl make working with player possition mutch easier.

I used a packet modification that woudl send a swimming packet for other players on the server. However this solution did not change client hitbox serverside.

### Other

I woudl love some feedback on this. I coudl to try to implement this as PR. However i do not want to wase my time developing a feature that woudl not be implemented.

SoSeDiK:
Adds a way to set entity's pose.
Fixed pose can be changed/reset only via another API call.
(my take on #7318 since it was closed))

Resolves #7016.

### DIFF
--- a/patches/api/0430-Expand-Pose-API.patch
+++ b/patches/api/0430-Expand-Pose-API.patch
@@ -1,0 +1,53 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SoSeDiK <mrsosedik@gmail.com>
+Date: Wed, 11 Jan 2023 20:59:02 +0200
+Subject: [PATCH] Expand Pose API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Entity.java b/src/main/java/org/bukkit/entity/Entity.java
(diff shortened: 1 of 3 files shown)


MWHunter:
Other players will see the pose but the client will revert this immediately on the next tick.  This seems hacky and PoseChangeEvent might be a better thing.  There should also be warnings so people don't expect too much from poses.

SoSeDiK:
Yep, poses work for entities and other players will see the pose correctly (and having it fixed won't let the player reset it next tick).
Only the player with the pose applied won't be able to see it client-side. I left a note about this in javadocs,

kennytv (approved):

PR #8782, opened 2023-01-11
Feature/player listing api

CoreyShupe:
These patches aim to expose the new player listing features present in 1.19.3; the implementation is not complete as it does not account for certain states.

For the implementation to be "complete" there needs to be discussion on potential API behavioral changes to what currently exist; along with more API to cover what was changed or removed.

This PR simply exposes the new API as-is and readies for the following changes.

### DIFF
--- a/patches/api/0427-Add-Listing-API-for-Player.patch
+++ b/patches/api/0427-Add-Listing-API-for-Player.patch
@@ -5,16 +5,16 @@ Subject: [PATCH] Add Listing API for Player
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index de960716478477ce199526b8f860cfafa1541ee9..d3dcfdcc5a879ea5842306cdc811277900825171 100644
+index de960716478477ce199526b8f860cfafa1541ee9..1e3fb072bcf7fab5ba9c1bd9f9412cd7b135f173 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
-@@ -1483,6 +1483,43 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+@@ -1483,6 +1483,32 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      @org.jetbrains.annotations.ApiStatus.Experimental // Paper
      public boolean canSee(@NotNull Entity entity);
  
-+    // paper start
++    // Paper start
 +    /**
-+     * Determines if the {@code other} player is listed for {@code this}.
++     * Returns whether the {@code other} player is listed for {@code this}.
 +     *
 +     * @param other The other {@link Player} to check for listing.
 +     * @return True if the {@code other} player is listed for {@code this}.
@@ -22,32 +22,21 @@ index de960716478477ce199526b8f860cfafa1541ee9..d3dcfdcc5a879ea5842306cdc8112779
 +    public boolean isListed(@NotNull Player other);
 +
 +    /**
-+     * Sets the {@code other} player as listed or unlisted.
-+     *
-+     * @param plugin The {@link Plugin} which wants to mutate the {@code other} player's listing.
-+     * @param other The other {@link Player} to set listed or unlisted.
-+     * @param listed Whether the {@code other} player should be listed or de-listed.
-+     */
-+    public void setListed(@NotNull Plugin plugin, @NotNull Player other, boolean listed);
-+
-+    /**
 +     * De-lists the {@code other} player.
 +     *
-+     * @param plugin The {@link Plugin} that is de-listing the {@code other} player.
 +     * @param other The other {@link Player} to de-list.
 +     * @return True if the {@code other} player was listed.
 +     */
-+    public boolean deListPlayer(@NotNull Plugin plugin, @NotNull Player other);
++    public boolean delistPlayer(@NotNull Player other);
 +
 +    /**
 +     * Lists the {@code other} player. This method is equivalent to {@code setListed(other, true)}.
 +     *
-+     * @param plugin The {@link Plugin} that is listing the {@code other} player.
 +     * @param other The other {@link Player} to list.
 +     * @return True if the {@code other} player was not listed.
 +     */
-+    public boolean listPlayer(@NotNull Plugin plugin, @NotNull Player other);
-+    // paper end
++    public boolean listPlayer(@NotNull Player other);
++    // Paper end
 +
      /**
       * Checks to see if this player is currently flying or not.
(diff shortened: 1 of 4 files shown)


patches/api/0427-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,54 @@
...
++    public void setListed(@NotNull Plugin plugin, @NotNull Player other, boolean listed);
++
++    /**
++     * De-lists the {@code other} player.
++     *
++     * @param plugin The {@link Plugin} that is de-listing the {@code other} player.
++     * @param other The other {@link Player} to de-list.
++     * @return True if the {@code other} player was listed.
++     */
++    public boolean deListPlayer(@NotNull Plugin plugin, @NotNull Player other);
```
Doc94: not a big fan for this name... maybe can be a unique method with a bool param for add/remove?

CoreyShupe: https://github.com/PaperMC/Paper/pull/8782/files#diff-7f2b831be83681583b338f6fb1dd559ffd5eec2e05ef3016ac7c66975a8e8b31R31

patches/api/0427-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,54 @@
...
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index de960716478477ce199526b8f860cfafa1541ee9..d3dcfdcc5a879ea5842306cdc811277900825171 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -1483,6 +1483,43 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+     @org.jetbrains.annotations.ApiStatus.Experimental // Paper
+     public boolean canSee(@NotNull Entity entity);
+ 
++    // paper start
```
kennytv: Uppercase P in start and end comments

patches/api/0427-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,54 @@
...
++    /**
++     * Determines if the {@code other} player is listed for {@code this}.
```
kennytv: Returns whether...

patches/api/0427-Add-Listing-API-for-Player.patch
kennytv: delistPlayer(

patches/api/0427-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,54 @@
...
++    public boolean isListed(@NotNull Player other);
++
++    /**
++     * Sets the {@code other} player as listed or unlisted.
++     *
++     * @param plugin The {@link Plugin} which wants to mutate the {@code other} player's listing.
++     * @param other The other {@link Player} to set listed or unlisted.
++     * @param listed Whether the {@code other} player should be listed or de-listed.
++     */
++    public void setListed(@NotNull Plugin plugin, @NotNull Player other, boolean listed);
```
kennytv: I don't think we need this one if we already have list/unlist

patches/server/0957-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,168 @@
...
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
+index 4b754f6eae683248d7fe11d6d6cb168d5dd696a2..23a6f73a724ec6b9288f681112e5185d23f8ea35 100644
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
++++ b/src/main/java/net/minecraft/server/players/PlayerList.java
+@@ -331,7 +331,7 @@ public abstract class PlayerList {
+         // CraftBukkit end
+ 
+         // CraftBukkit start - sendAll above replaced with this loop
+-        ClientboundPlayerInfoUpdatePacket packet = ClientboundPlayerInfoUpdatePacket.createPlayerInitializing(List.of(player));
++        ClientboundPlayerInfoUpdatePacket packet = ClientboundPlayerInfoUpdatePacket.createPlayerInitializing(List.of(player), player); // Paper
```
kennytv: This doesn't look right, you need to check visibility for other players, not the player itself

kennytv (changes requested):

dawon:
Maybe a suggestion to incorporate this as another parameter to hide/showPlayer overload, so you can hide the player as an entity while retaining said player in the tab list...

patches/server/0957-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,168 @@
...
++    }
++
++    public ClientboundPlayerInfoUpdatePacket(EnumSet<ClientboundPlayerInfoUpdatePacket.Action> actions, ClientboundPlayerInfoUpdatePacket.Entry entry) {
++        this.actions = actions;
++        this.entries = List.of(entry);
++    }
++    // Paper end
+ 
+-    public static ClientboundPlayerInfoUpdatePacket createPlayerInitializing(Collection<ServerPlayer> players) {
++    public static ClientboundPlayerInfoUpdatePacket createPlayerInitializing(Collection<ServerPlayer> players, ServerPlayer forPlayer) { // Paper
```
montlikadani: Consider creating a different method with the ServerPlayer parameter and deprecate this method or just leave it as is. With this change many things will be broken.

Edit: or make it nullable
In "many things" I mean plugins or other resources using this method will throw a runtime exception because of a missing parameter.

CoreyShupe:
> Maybe a suggestion to incorporate this as another parameter to hide/showPlayer overload, so you can hide the player as an entity while retaining said player in the tab list...

I don't think this is necessarily a good idea; the showPlayer/hidePlayer is going to get complicated rather quickly; new API for specifically this is probably better for right now. We'll be looking to improve and expand on this as it gets built. For now new API is good here since it leaves the showPlayer hidePlayer logic alone.

There are missing possible "states" but considering how to implement that will further complicate the system and will need to be further discussed; probably more than this PR will require.

As for the rest of the comments they should be resolved.

patches/api/0427-Add-Listing-API-for-Player.patch
```diff
@@ -0,0 +1,43 @@
...
++    /**
++     * De-lists the {@code other} player.
++     *
++     * @param other The other {@link Player} to de-list.
++     * @return True if the {@code other} player was listed.
++     */
++    public boolean delistPlayer(@NotNull Player other);
++
++    /**
++     * Lists the {@code other} player. This method is equivalent to {@code setListed(other, true)}.
```
Lulu13022002: This do a reference to a method that doesn't seems to exist anymore

kennytv (approved):

PR #8790, opened 2023-01-15
Don't clone Location in PreCreatureSpawnEvent

alex6777:
A new Location object is already created every time this event is called.  Micro opt but it's in a hot path 

PreCreatureSpawnEvent


Owen1212055 (approved):
The location created here seems to always be created from an NMS blockpos anyways.

kennytv (approved):

lynxplay (approved):
For that one plugin out there that manually calls that event and soon breaks because another one is mutating its location: Please clone your locations before passing it :+1:

LGTM

PR #8793, opened 2023-01-15
Add ItemStack#hoverName API

MelnCat:
Currently, ItemStack#displayName will return the item's display name but with brackets added and with a hover event added. This pull request adds a method named "hoverName" that will return the item's display name exactly how it looks when hovering over the item (just the name, without brackets or hover events). The displayName method actually internally calls the NMS getHoverName method to apply the brackets and events to, so this pull request helps expose the hovername of the item.

### DIFF
--- a/patches/api/0427-Add-ItemStack-hoverName-API.patch
+++ b/patches/api/0427-Add-ItemStack-hoverName-API.patch
@@ -1,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MelnCat <melncatuwu@gmail.com>
+Date: Sun, 15 Jan 2023 15:32:50 -0800
+Subject: [PATCH] Add ItemStack#hoverName API
+
+
+diff --git a/src/main/java/org/bukkit/inventory/ItemFactory.java b/src/main/java/org/bukkit/inventory/ItemFactory.java
+index 40edff7c93b6bf75de81102326667135b9344666..4e9d0d55afaab4613ba274ac48de53d773a22b41 100644
+--- a/src/main/java/org/bukkit/inventory/ItemFactory.java
++++ b/src/main/java/org/bukkit/inventory/ItemFactory.java
+@@ -195,6 +195,15 @@ public interface ItemFactory {
+     @NotNull
+     net.kyori.adventure.text.Component displayName(@NotNull ItemStack itemStack);
+ 
++    /**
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Isn’t this exactly the same as the ItemMeta#displayName?

MelnCat:
It helps to have a method for getting the hover name rather than having to check hasDisplayName and needing to manually create a translatable component for the item. There's even already a method in NMS that allows getting the hoverName, so doing it manually is inelegant. Additionally, having an API method for hoverName allowa it to still have the intended behaviour if mojang ever changes the method

lynxplay:
Would probably be inclined to agree with MM here. This boils down to
```java
Optional.ofNullable(itemStack.getItemMeta())
            .map(ItemMeta::displayName)
            .orElse(Component.translatable(itemStack));
```
in the plugin code.

The fact that an NMS method exists for this seems like a weak argument for an API addition tbh 😓 
I'd rather leave as much of the "component building" to plugins as possible.

MelnCat:
Alright then

lynxplay:
Oh, I didn't want to sound rude here, paper has a lot of utility methods in the API, that was purely my own opinion.
Just because I don't 100% see a reason for a utility method like that added does not mean others won't. 

Sorry if my message sounded imposing like that, that was not my intention.
I am sure other team members have other opinions.

MelnCat:
I just agreed with your reasons

Phoenix616:
> Would probably be inclined to agree with MM here. This boils down to
> 
> ```java
> Optional.ofNullable(itemStack.getItemMeta())
>             .map(ItemMeta::displayName)
>             .orElse(Component.translatable(itemStack));
> ```
> 
> in the plugin code.
> 
> The fact that an NMS method exists for this seems like a weak argument for an API addition tbh 😓 I'd rather leave as much of the "component building" to plugins as possible.

For anyone looking at this PR at a later date: This is wrong, it's not (easily) possible to implement the functionality of `NMS ItemStack#getHoverName` right now as `Component.translatable(itemStack)` will only return a translatable component based on the raw item translation key without additional item information (the player of a skull or the title of the book).

An API method for this should not be called `hoverName` though as this is based on an internal name and only makes sense in the context of the item information being shown in the chat in a way where further information can be queried when hovering (while the returned component doesn't even have a hover event itself). This should be named more along the lines of `itemName`, `display(ed)Name` or just `name`.

PR #8795, opened 2023-01-17
[ci skip] Fix PiglinBarterEvent JavaDoc

tal5:
Currently `PiglinBarterEvent` is documented as `can be triggered by a piglin picking up an item that's on its bartering list.`, but it's fired when the barter is complete, right before the items are dropped.
I went with the generic `Called when X` format most events use, let me know if it should include anything else.


Machine-Maker:
I'd maybe keep the the first line of the javadoc there.

tal5:
> I'd maybe keep the the first line of the javadoc there.

Re-added

Machine-Maker (approved):

lynxplay (approved):

jpenilla (approved):

PR #8798, opened 2023-01-18
Use a proper thread pool in ServerLoginPacketListenerImpl and MCUtil

### LINKED ISSUE #8797: ServerLoginPacketListenerImpl uses Executors.newCachedThreadPool with unbounded queue
### Timings or Profile link

N/A

### Description of issue

Discussed internally- no timings link or plugin list. This is an issue created so I can reference it in a near-future PR.

Commit [0ed4b91](https://github.com/PaperMC/Paper/commit/0ed4b9148b72a7fb702333cf1221be6aeecff7ca) reverts ThreadPoolExecutor to `Executors#newCachedThreadPool` which was changed to solve an issue of hanging logins. The problem with both, as discussed, is that both have the unfortunate side-effect of using one thread in the pool when combined with an unbounded queue. In the case of ServerLoginPacketListenerImpl this means that ALL logins are sent to the same thread regardless of how many threads the pool has available.

Discussions lead to a few different solutions, but after looking at the benefits and drawbacks of each solution the easiest fix is to use a fixed thread pool (a new ThreadExecutor where the core pool size is equal to the maximum pool size).

Other resolutions discussed were using a bounded queue and a "[scaling thread pool](https://medium.com/@uditharosha/java-scale-first-executorservice-4245a63222df)" but ultimately the overhead for a fixed thread pool is minimal, is the easiest to maintain, and is likely the best solution for an authentication pool which is expected to live for the lifetime of the server anyway.

Further, discussions of a bounded vs unbounded queue led us to consider potential attack vectors. A bounded queue for the pool would see a successful attack by way of flooding the authentication service with fake logins and an unbounded pool would see a successful attack by way of flooding the authentication service with fake logins. The former would lead to `RejectedExecutionExeption`s and the latter would fill the heap until the server ran out of memory.

After deliberation, it was decided that a bounded queue would have less impact on the server despite a successful L7 attack on the authentication service. There are other solutions to L7 attacks on Minecraft servers and Paper's due diligence on attacks is best-effort while still keeping the original goals of performance, stability, and bugfixes in mind. Rejected logins are a better failure case than a crashed server even though the latter is more difficult.

A semi-arbitrary number of 1024 for the queue size seems reasonable.

### Plugin and Datapack List

N/A

### Server config files

N/A

### Paper version

1.19.3 380, [0ed4b91](https://github.com/PaperMC/Paper/commit/0ed4b9148b72a7fb702333cf1221be6aeecff7ca)

### Other

_No response_

egg82:
Fixes #8797 

An arbitrary-ish number of 1024 was chosen for the bounded queue size.
The asyncExecutor in MCUtil was left unbounded.

### DIFF
--- a/patches/server/0111-Cache-user-authenticator-threads.patch
+++ b/patches/server/0111-Cache-user-authenticator-threads.patch
@@ -5,59 +5,209 @@ Subject: [PATCH] Cache user authenticator threads
 
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
-index a6e90e70730e106d1cac01abf7a41df8df787d89..453a9dd794c39be5e7221b05fabb71b802691003 100644
+index a6e90e70730e106d1cac01abf7a41df8df787d89..b6103cfe2a5cce6feba51a2771cd0fa2dea97e43 100644
 --- a/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
 @@ -112,6 +112,8 @@ public class ServerLoginPacketListenerImpl implements ServerLoginPacketListener,
  
      }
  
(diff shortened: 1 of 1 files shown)


NoahvdAa (changes requested):
Try to avoid reformatting these large blocks to keep the diff small :)

This should probably also go into [the original patch](https://github.com/PaperMC/Paper/blob/0ed4b9148b72a7fb702333cf1221be6aeecff7ca/patches/server/0111-Cache-user-authenticator-threads.patch).

lynxplay:
Are you still planning on moving forward with this ? I don't recall the dhrama discussion about it but I'd close this otherwise.

lynxplay:
Closing this given it has gone rather stale :sweat:

PR #8799, opened 2023-01-19
Add AntiXray layered obfuscation mode

DrexHD:
**Changes**
* Removes code duplication from `obfuscateLayer`
* Adds a 3rd engine mode (layered obfuscation), this mode works the same as engine mode 2, but instead of randomizing every block, only randomizes blocks [per chunk layer](https://github.com/DrexHD/AntiXray/blob/1.19.3/stable/media/enginemode-3.png). This helps a lot with chunk packet compression and can reduce the network load when joining by a factor of ~2 (from very basic testing)

**Testing**
The tests have been run on a plain new paper server by joining three times in the same spot in spawn chunks. After every test, the server has been rebooted, and these results have shown to be reproducible. 
Results: [imgur](https://imgur.com/a/CmJDzcq)

### DIFF
--- a/patches/server/0344-Anti-Xray.patch
+++ b/patches/server/0344-Anti-Xray.patch
@@ -199,10 +199,10 @@ index 0000000000000000000000000000000000000000..bd86dc2ad2f87969da4add06de2a629f
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java b/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..1b70024bfa2828fb12ecadf11b4d93291da28509
+index 0000000000000000000000000000000000000000..e6d5504e9eae41b8326d823affc520be2e928e7b
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/antixray/ChunkPacketBlockControllerAntiXray.java
-@@ -0,0 +1,479 @@
+@@ -0,0 +1,671 @@
 +package com.destroystokyo.paper.antixray;
 +
 +import io.papermc.paper.configuration.WorldConfiguration;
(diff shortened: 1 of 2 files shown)


stonar96:
>* Removes code duplication from `obfuscateLayer`

The code duplication is very much intended. You are touching the hottest code in Anti-Xray and with your changes you are introducing lots of branches into that code.

Also your changes can be implemented without even touching the `obfuscateLayer` method and avoiding the additional `engineMode == EngineMode.OBFUSCATE_LAYER ? ` condition in the hot code. There's a reason why the `random` object is passed to this method.

I like the idea. The implementation of it should be a matter of changing like 2 or 3 lines tho.

stonar96:
This will do it (not tested):

```
LayeredIntSupplier random = numberOfBlocks == 1 ? (() -> 0) : engineMode == EngineMode.OBFUSCATE_LAYER ? new LayeredIntSupplier() {
    private int state;
    private int next;

    {
        while ((state = ThreadLocalRandom.current().nextInt()) == 0) ;
    }

    @Override
    public void nextLayer() {
        // https://en.wikipedia.org/wiki/Xorshift
        state ^= state << 13;
        state ^= state >>> 17;
        state ^= state << 5;
        // https://www.pcg-random.org/posts/bounded-rands.html
        next = (int) ((Integer.toUnsignedLong(state) * numberOfBlocks) >>> 32);
    }

    @Override
    public int getAsInt() {
        return next;
    }
} : new LayeredIntSupplier() {
    private int state;

    {
        while ((state = ThreadLocalRandom.current().nextInt()) == 0) ;
    }

    @Override
    public int getAsInt() {
        // https://en.wikipedia.org/wiki/Xorshift
        state ^= state << 13;
        state ^= state >>> 17;
        state ^= state << 5;
        // https://www.pcg-random.org/posts/bounded-rands.html
        return (int) ((Integer.toUnsignedLong(state) * numberOfBlocks) >>> 32);
    }
};

@FunctionalInterface
private static interface LayeredIntSupplier extends IntSupplier {
    default void nextLayer() {

    }
}
```

Then call `random.nextLayer()` before each `obfuscateLayer(...)`.

patches/server/0344-Anti-Xray.patch
```diff
@@ -960,6 +995,19 @@ index 0000000000000000000000000000000000000000..80a2dfb266ae1221680a7b24fee2f7e2
...
+diff --git a/src/main/java/io/papermc/paper/configuration/type/EngineMode.java b/src/main/java/io/papermc/paper/configuration/type/EngineMode.java
+index 99e90636051fa0c770ee2eafb7f0d29c8195f9ae..5d6657939595220a4cf6f1bf3ccfca3619b5bce1 100644
+--- a/src/main/java/io/papermc/paper/configuration/type/EngineMode.java
++++ b/src/main/java/io/papermc/paper/configuration/type/EngineMode.java
+@@ -5,7 +5,7 @@ import org.spongepowered.configurate.serialize.ScalarSerializer;
+ 
+ public enum EngineMode {
+ 
+-    HIDE(1, "hide ores"), OBFUSCATE(2, "obfuscate");
++    HIDE(1, "hide ores"), OBFUSCATE(2, "obfuscate"), OBFUSCATE_LAYER(3, "obfuscate_layer");
```
stonar96: This goes into 0004-Paper-config-files.patch. Also to be consistent with `"hide ores"`, I'd call it `"obfuscate layer"` instead of `"obfuscate_layer"`, this parameter is a `description` after all.

patches/server/0344-Anti-Xray.patch
```diff
@@ -494,6 +518,7 @@ index 0000000000000000000000000000000000000000..4f3670b2bdb8b1b252e9f074a6af56a0
 +
 +                    // Abuse the obfuscateLayer method to read the blocks of the first layer of the current chunk section
 +                    bitStorageWriter.setBits(0);
++                    random.nextLayer();
```
stonar96: This one can be removed, it doesn't matter tho. I think it feels more correct when it's removed.

stonar96 (changes requested):
Everything else looks good to me. Thanks!

stonar96 (approved):

kennytv (approved):

PR #8800, opened 2023-01-20
Add Player#sendEquipmentChange(Map) API

tal5:
Adds `Player#sendEquipmentChange(LivingEntity, Map<EquipementSlot, ItemStack>)`, this is useful as the current method only changes one slot, but the underlaying packet takes a list of pairs, which means using the existing method you might need to send 6 packets when it can be done in one.

### DIFF
--- a/patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
+++ b/patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
@@ -1,0 +1,40 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aya <31237389+tal5@users.noreply.github.com>
+Date: Fri, 20 Jan 2023 13:49:35 +0000
+Subject: [PATCH] Add Player#sendEquipmentChange(Map) API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index de960716478477ce199526b8f860cfafa1541ee9..e98acbb96bfe226f929c3d24be24d0e5ae80e83d 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -657,6 +657,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+     public void sendMultiBlockChange(@NotNull java.util.Map<Location, BlockData> blockChanges, boolean suppressLightUpdates);
+     // Paper end
+ 
++    // Paper start
+     /**
+      * Send the equipment change of an entity. This fakes the equipment change
+      * of an entity for a user. This will not actually change the inventory of
+@@ -666,7 +667,20 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      * @param slot The slot of the spoofed equipment change
+      * @param item The ItemStack to display for the player
+      */
+-    public void sendEquipmentChange(@NotNull LivingEntity entity, @NotNull EquipmentSlot slot, @NotNull ItemStack item);
++    default void sendEquipmentChange(@NotNull LivingEntity entity, @NotNull EquipmentSlot slot, @NotNull ItemStack item) {
++        this.sendEquipmentChange(entity, java.util.Map.of(slot, item));
++    };
++
++    /**
++     * Send an equipment change for an entity. This fakes the equipment change
++     * of an entity for a user. This will not actually change the inventory of
++     * the specified entity in any way.
++     *
++     * @param entity The entity that the player will see the change for
++     * @param equipmentChanges A map of slots to the items they will be changed to
++     */
++    void sendEquipmentChange(@NotNull LivingEntity entity, @NotNull java.util.Map<EquipmentSlot, ItemStack> equipmentChanges);
++    // Paper end
+ 
+     // Paper start
+     /**
--- a/patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
+++ b/patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
@@ -5,22 +5,15 @@ Subject: [PATCH] Add Player#sendEquipmentChange(Map) API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index de960716478477ce199526b8f860cfafa1541ee9..e98acbb96bfe226f929c3d24be24d0e5ae80e83d 100644
+index de960716478477ce199526b8f860cfafa1541ee9..10232a4568c026aa66229f55733078c32d1c71eb 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
-@@ -657,6 +657,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
-     public void sendMultiBlockChange(@NotNull java.util.Map<Location, BlockData> blockChanges, boolean suppressLightUpdates);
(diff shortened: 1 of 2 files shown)


patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
```diff
@@ -0,0 +1,40 @@
...
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index de960716478477ce199526b8f860cfafa1541ee9..e98acbb96bfe226f929c3d24be24d0e5ae80e83d 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -657,6 +657,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+     public void sendMultiBlockChange(@NotNull java.util.Map<Location, BlockData> blockChanges, boolean suppressLightUpdates);
+     // Paper end
+ 
++    // Paper start
```
lynxplay: Move this paper block down to the actual start of the diff, e.g. right above the change to a default method.

patches/server/0958-Add-Player-sendEquipmentChange-Map-API.patch
```diff
@@ -0,0 +1,39 @@
...
+-                new Pair<>(CraftEquipmentSlot.getNMS(slot), CraftItemStack.asNMSCopy(item))
+-        );
++        List<Pair<net.minecraft.world.entity.EquipmentSlot, net.minecraft.world.item.ItemStack>> equipment = new ArrayList<>(equipmentChanges.size());
++        for (Map.Entry<EquipmentSlot, ItemStack> entry : equipmentChanges.entrySet()) {
++            equipment.add(new Pair<>(CraftEquipmentSlot.getNMS(entry.getKey()), CraftItemStack.asNMSCopy(entry.getValue())));
++        }
+ 
+         this.getHandle().connection.send(new ClientboundSetEquipmentPacket(entity.getEntityId(), equipment));
+     }
++    // Paper end
```
lynxplay: Move this up to the actual end of the diff. (right below the changed "equipment" list creation.

lynxplay (changes requested):
Looks pretty good to me!
Also welcome to your first paper PR!

A few comments, mainly in regards to the // Paper comments.
Beyond that, looks good from a first look over it, I'll test it later today :)

tal5:
My bad, Paper comments should be fixed.

lynxplay (dismissed):
Alright, tested it in general and it looks good, with one small and rather unfortunate missing precondition.

Previously the logic ensured that both slot and item were not null.
Now the only thing ensured is the fact that the map is not null.
While the logic works fine for null values (`CraftItemStack.asNMSCopy` returns ItemStack.EMPTY for null values) it does not work fine for null keys, which map implementations like `HashMap` allow. 

Imo this would hence benefit from ensuring that the map does *not* contain `null` has a key, which should be easily checkable with a `map.containsKey(null)` precondition check as otherwise `CraftEquipmentSlot.getNMS(entry.getKey())` errors when calling `ordinal()`.

Beyond that LGTM tho!

tal5:
> Alright, tested it in general and it looks good, with one small and rather unfortunate missing precondition.
> 
> Previously the logic ensured that both slot and item were not null. Now the only thing ensured is the fact that the map is not null. While the logic works fine for null values (`CraftItemStack.asNMSCopy` returns ItemStack.EMPTY for null values) it does not work fine for null keys, which map implementations like `HashMap` allow.
> 
> Imo this would hence benefit from ensuring that the map does _not_ contain `null` has a key, which should be easily checkable with a `map.containsKey(null)` precondition check as otherwise `CraftEquipmentSlot.getNMS(entry.getKey())` errors when calling `ordinal()`.
> 
> Beyond that LGTM tho!

`Map.of` already does that, it doesn't allow null values and uses `Objects.requireNonNull` on both the key and value - it's not the exact same exception but I think it should be fine?

lynxplay:
I am not talking about the default method call, rather about the newly added one.

Beyond that, the messages in predictions have proper messages, where as Map.of does not nor does an ordinal call against a null value deep in the method

SirYwell:
Note that `containsKey(null)` is not safe to call as the method is allowed to throw an NPE if the map does not allow null keys.

lynxplay:
> Note that `containsKey(null)` is not safe to call as the method is allowed to throw an NPE if the map does not allow null keys.

TIL thank you XD

tal5:
I've added a null check in the loop as that seems like the best way to do it, let me know if there's a better solution

electronicboy (approved):

patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
```diff
@@ -0,0 +1,33 @@
...
++
++    /**
++     * Send an equipment change for an entity. This fakes the equipment change
++     * of an entity for a user. This will not actually change the inventory of
++     * the specified entity in any way.
++     *
++     * @param entity The entity that the player will see the change for
++     * @param equipmentChanges A map of slots to the items they will be changed to
++     */
++    void sendEquipmentChange(@NotNull LivingEntity entity, @NotNull java.util.Map<EquipmentSlot, ItemStack> equipmentChanges);
```
Machine-Maker, line 29: Use nullability annotations on the key and value type of this map. (Put `@NotNull` on both).

patches/server/0958-Add-Player-sendEquipmentChange-Map-API.patch
```diff
@@ -0,0 +1,37 @@
...
++        Preconditions.checkNotNull(equipmentChanges, "equipmentChanges must not be null");
+ 
+         if (this.getHandle().connection == null) return;
+ 
+-        List<Pair<net.minecraft.world.entity.EquipmentSlot, net.minecraft.world.item.ItemStack>> equipment = Arrays.asList(
+-                new Pair<>(CraftEquipmentSlot.getNMS(slot), CraftItemStack.asNMSCopy(item))
+-        );
++        List<Pair<net.minecraft.world.entity.EquipmentSlot, net.minecraft.world.item.ItemStack>> equipment = new ArrayList<>(equipmentChanges.size());
++        for (Map.Entry<EquipmentSlot, ItemStack> entry : equipmentChanges.entrySet()) {
++            Preconditions.checkNotNull(entry.getKey(), "EquipmentSlot key must not be null");
```
Machine-Maker, line 31: Also make sure the ItemStack (value) is not null.

Machine-Maker (changes requested):

tal5:
Both should be fixed

Machine-Maker (approved):

patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
```diff
@@ -0,0 +1,33 @@
...
++
++    /**
++     * Send an equipment change for an entity. This fakes the equipment change
++     * of an entity for a user. This will not actually change the inventory of
++     * the specified entity in any way.
++     *
++     * @param entity The entity that the player will see the change for
++     * @param equipmentChanges A map of slots to the items they will be changed to
++     */
++    void sendEquipmentChange(@NotNull LivingEntity entity, @NotNull java.util.Map<@NotNull EquipmentSlot, @NotNull ItemStack> equipmentChanges);
```
kennytv: Remove the annotations from the generic types again and add a note in the jd comment that values should be non-null

kennytv (approved):

patches/api/0427-Add-Player-sendEquipmentChange-Map-API.patch
tal5: Should be done

lynxplay (approved):

kennytv:
Thanks!

PR #8801, opened 2023-01-20
Spawner Minecart API

Lama06:
Adds all not deprecated methods from CreatureSpawner to SpawnerMinecart
Adds the SpawnerHolder interface for the common methods of CreatureSpawner and SpawnerMinecart (simmilar to how CommandBlockHolder is the superinterface for CommandBlock and CommandMinecart)

This is my first pull request ever, so feedback is highly appreciated :)


patches/api/0180-Mob-Spawner-API-Enhancements.patch
```diff
@@ -4,16 +4,190 @@ Date: Fri, 19 Apr 2019 12:41:19 -0500
...
+new file mode 100644
+index 0000000000000000000000000000000000000000..f8c6975509655b59c739aa0e634e0f3c2e6a72b2
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/SpawnerHolder.java
+@@ -0,0 +1,202 @@
++package io.papermc.paper;
++
++import org.bukkit.entity.EntityType;
++import org.jetbrains.annotations.NotNull;
++
```
MiniDigger, line 21: misses a class comment like "represents something that can handle spawner information" or something
(I bet you can come up with something better than me, lol)

patches/server/0312-Mob-Spawner-API-Enhancements.patch
```diff
@@ -101,3 +101,131 @@ index e8f888b9e37454d9863fbaf0f0c3dc982d582f1e..95b01ddddb1ba90da495927099147e77
...
++    public void setSpawnedItem(org.bukkit.inventory.ItemStack itemStack) {
++        com.google.common.base.Preconditions.checkArgument(itemStack != null && !itemStack.getType().isAir(),
++            "spawners cannot spawn air");
++        net.minecraft.world.item.ItemStack item = org.bukkit.craftbukkit.inventory.CraftItemStack.asNMSCopy(itemStack);
++        net.minecraft.nbt.CompoundTag entity = new net.minecraft.nbt.CompoundTag();
++        entity.putString("id", net.minecraft.core.registries.BuiltInRegistries.ENTITY_TYPE.getKey(net.minecraft.world.entity.EntityType.ITEM).toString());
++        entity.put("Item", item.save(new net.minecraft.nbt.CompoundTag()));
++        getHandle().getSpawner().setNextSpawnData(getHandle().getLevel(), getHandle().blockPosition(),
++            new net.minecraft.world.level.SpawnData(entity, java.util.Optional.empty()));
++    }
```
MiniDigger, line 302: misses a paper end comment here I believe

MiniDigger (approved):
provisionally approving, just found two minor things, but welcome, nice work on your first PR!

patches/api/0180-Mob-Spawner-API-Enhancements.patch
```diff
@@ -4,16 +4,190 @@ Date: Fri, 19 Apr 2019 12:41:19 -0500
...
-+++ b/src/main/java/org/bukkit/block/CreatureSpawner.java
-@@ -199,4 +199,30 @@ public interface CreatureSpawner extends TileState {
-      * @see #getSpawnRange()
-      */
-     public void setSpawnRange(int spawnRange);
+diff --git a/src/main/java/io/papermc/paper/SpawnerHolder.java b/src/main/java/io/papermc/paper/SpawnerHolder.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..f8c6975509655b59c739aa0e634e0f3c2e6a72b2
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/SpawnerHolder.java
```
Machine-Maker: Don't put this in the root package

patches/api/0180-Mob-Spawner-API-Enhancements.patch
```diff
@@ -37,5 +211,206 @@ index cb447a4ad5a9dce7c98999a5d7fcd6111fc9b10e..5bbae759ce39d42886994e500fd9454e
...
+-    public int getSpawnRange();
+-
+-    /**
+-     * Set the new spawn range.
+-     * <br>
+-     *
+-     * @param spawnRange the new spawn range
+-     * @see #getSpawnRange()
+-     */
+-    public void setSpawnRange(int spawnRange);
```
Machine-Maker: I don't think this should be removed. I imagine the javadocs are the same, and this just creates more diff.

patches/server/0312-Mob-Spawner-API-Enhancements.patch
Machine-Maker: Instead of adding all of these to the Craft class, make a separate interface with a bunch of default methods like PaperSpawnerHolder. Then only implement the one method you need to in both CraftMinecrartMobSpawner and CraftCreatureSpawner. [Take a look at what I did for something almost identical to this for command blocks and their minecarts](https://github.com/PaperMC/Paper/blob/d5f1ffc2e6928a6728f91ab2ad2e020b105eab7f/patches/server/0682-More-CommandBlock-API.patch).

Machine-Maker (changes requested):

patches/api/0180-Mob-Spawner-API-Enhancements.patch
Lama06: Having created the PaperSpawnerHolder interface, I now have to remove these methods because there is a compile error otherwiese. For some reason the PaperSpawnerHolder interface cant implement methods from the CreatureSpawner interface, even though the methods have the same signature because it only provides default implementations for the methods of the SpawnerHolder interface. It is hard to explain, I hope you understand what I mean

Machine-Maker: That doesn't make any sense. There shouldn't be a compile error if the method in the CreatureSpawner interface is just overriding one in the SpawnerHolder interface. In fact I just did that right now, and there was no compile error. You need to make sure that the methods you added to the SpawnerHolder interface exactly match the ones in the CreatureSpawner interface

API Patch: https://pastes.dev/2OcV0QHNRT
Server Patch: https://pastes.dev/5AEhAfFEde

Lama06: As you have suggested I have made sure that every signature from SpawnerHolder is exactly the same as in CreatureSpawnerHolder but it still gives me a compile error if I dont remove the overriding methods from CreatureSpawner.
Please try to compile this example, I think it will make it clear what I mean: https://gist.github.com/Lama06/b05e4f5c6226f3b7309a8a5348bd25d5

Machine-Maker: Well that's cause you did the exact same thing with the methods in CraftCreatureSpawner. Don't delete the already implemented methods. Just like you shouldn't delete the methods in CreatureSpawner, to reduce diff, don't delete the methods in CraftCreatureSpawner.

Lama06: I have done what you have suggested and I now no longer delete the existing methods in CraftCreatureSpawner and CreatureSpawner.
Thanks for your help!

patches/server/0312-Mob-Spawner-API-Enhancements.patch
```diff
@@ -69,35 +116,189 @@ index 6ba97a0b4f2cb15d5435657c8e8f5c71c6fee3db..c5a117308f051c20b81818ad91e0ca40
...
++
++    @Override
++    public EntityType getSpawnedType() {
++        var spawnData = getHandle().getSpawner().nextSpawnData;
++        if (spawnData == null) {
++            return EntityType.PIG;
++        }
++
++        var type = net.minecraft.world.entity.EntityType.by(spawnData.getEntityToSpawn());
```
Machine-Maker, line 201: Don't use `var`

patches/server/0312-Mob-Spawner-API-Enhancements.patch
```diff
@@ -8,6 +8,53 @@ public net.minecraft.world.level.BaseSpawner isNearPlayer(Lnet/minecraft/world/l
...
++    @Override
++    default void resetTimer() {
++        getSpawnerHandle().delay(getLevel(), getBlockPosition());
++    }
++
++    @Override
++    default void setSpawnedItem(ItemStack itemStack) {
++        Preconditions.checkArgument(itemStack != null && !itemStack.getType().isAir(), "spawners cannot spawn air");
++        var item = CraftItemStack.asNMSCopy(itemStack);
++        var entity = new CompoundTag();
```
Machine-Maker, line 52: Don't use var, especially since you can add imports to this file.

patches/server/0312-Mob-Spawner-API-Enhancements.patch
```diff
@@ -69,35 +116,189 @@ index 6ba97a0b4f2cb15d5435657c8e8f5c71c6fee3db..c5a117308f051c20b81818ad91e0ca40
...
++    public MinecartSpawner getHandle() {
++        return (MinecartSpawner) super.getHandle();
++    }
++
++    @Override
++    public EntityType getSpawnedType() {
++        var spawnData = getHandle().getSpawner().nextSpawnData;
++        if (spawnData == null) {
++            return EntityType.PIG;
++        }
```
Machine-Maker, line 199: Why can't this method also be in PaperSpawnerHolder?

patches/server/0312-Mob-Spawner-API-Enhancements.patch
```diff
@@ -69,35 +116,189 @@ index 6ba97a0b4f2cb15d5435657c8e8f5c71c6fee3db..c5a117308f051c20b81818ad91e0ca40
...
++    }
++
++    @Override
++    public net.minecraft.world.level.Level getLevel() {
++        return getHandle().getLevel();
++    }
++
++    @Override
++    public net.minecraft.core.BlockPos getBlockPosition() {
++        return getHandle().blockPosition();
```
Machine-Maker, line 301: All of these should be in PaperSpawnerHolder

Machine-Maker (changes requested):

patches/server/0312-Mob-Spawner-API-Enhancements.patch
Lama06: If I move these methods to PaperSpawnerHolder, should I also delete the already exisiting methods in CraftCreatureSpawner? 
Because if I delete them, that will make the diff longer. 
And if I dont remove them, the methods in CraftCreatureSpawner will override the shared implementation in PaperSpawnerHolder. Because of that, the implementation in PaperSpawnerHolder will effectively only be used in CraftMinecartMobSpawner.

PR #8802, opened 2023-01-21
End Poem API

Lama06:
Adds a method that allows plugins to show the end poem to the player.
By using a resource pack plugin developers can change the content of the poem to something different.


Machine-Maker (commented):
"Poem" feels like the wrong word to describe it. I've always seen it (and called it) the end credits. I know it starts with that story, but I don't know if that qualifies as a poem?

imDaniX:
> "Poem" feels like the wrong word to describe it. I've always seen it (and called it) the end credits. I know it starts with that story, but I don't know if that qualifies as a poem?

I'd say it's actually both, as it starts with the poem, then proceeds to credits. Anyway, the author of the text calls it poem, and wikis (both [Wikipedia](https://en.wikipedia.org/wiki/Julian_Gough) and [MCWiki](https://minecraft.fandom.com/wiki/End_Poem)) uses this word as well.

Machine-Maker:
Ok, then I'd call it the `winScreen` then. If as the OP wants to, a resource pack changes it, its not a poem anymore. So call it something that can't change, its the screen that shows when you win the game.

Lama06:
I renamed the method to showWinScreen and changed its documentation accordingly.
Please let me now if the Javadoc comment sounds strange as I am not a native speaker :)

olijeffers0n:
Maybe:

Shows the player the win screen that would normally be displayed after the ender dragon is killed and the player exits the end for the first time. In Vanilla, the win screen starts with a poem and then continues with the credits but it’s content can be changed by using a resource pack.

Machine-Maker (approved):
Probably don't use "var" for the `packet` local var. No particular need for it here. But otherwise, looks good.

kennytv (approved):

lynxplay (approved):

lynxplay:
Renamed the commit on master to the chosen name (win screen) as well :+1:

PR #8804, opened 2023-01-22
Bell API

Lama06:
Adds methods to ring a bell and to query its state


patches/server/0959-Bell-API.patch
```diff
@@ -0,0 +1,62 @@
...
++    @Override
++    public int getHasBeenRingingFor() {
++        return getTileEntity().ticks;
++    }
++
++    @Override
++    public org.bukkit.block.BlockFace getRingingDirection() {
++        if (!isRinging()) return null;
++
++        return switch (getTileEntity().clickDirection) {
```
Lulu13022002: You can use the CraftBlock.notchToBlockFace method to avoid that. Also it's a little bit strange that you have mapped the up/down direction since you explicitly blocked that in the ring method

patches/server/0959-Bell-API.patch
```diff
@@ -0,0 +1,62 @@
...
+         super(world, tileEntity);
+     }
++
++    // Paper start - Bell API
++    @Override
++    public void ring(org.bukkit.block.BlockFace direction) {
++        if (!direction.isCartesian()) throw new IllegalArgumentException("A bell can only ring in a cartesian direction");
++        if (direction.getModY() != 0) throw new IllegalArgumentException("A bell cannot ring up or down");
++        requirePlaced();
++        getTileEntity().onHit(net.minecraft.core.Direction.fromNormal(direction.getModX(), direction.getModY(), direction.getModZ()));
```
Lulu13022002: Since you have checked that the block face is cartesian, you can directly use the CraftBlock.blockFaceToNotch

Lama06: Thanks for your feedback, I now use these methods

patches/api/0429-Bell-API.patch
```diff
@@ -0,0 +1,67 @@
...
++
++    /**
++     * Returns the direction the bell is ringing in or null if it is not ringing.
++     * @return The direction the bell is ringing in
++     */
++    @org.jetbrains.annotations.Nullable BlockFace getRingingDirection();
++
++    /**
++     * Returns if the bell is currently resonating.
++     * The resonation sound starts playing 5 ticks after the bell is hit and ends 40 ticks later.
```
Machine-Maker: This also requires nearby raiders.

patches/api/0429-Bell-API.patch
```diff
@@ -0,0 +1,67 @@
...
++     * @return Whether the bell is currently ringing
++     */
++    boolean isRinging();
++
++    /**
++     * Returns the number of ticks this bell has been ringing for or 0 if it is not ringing.
++     * A bell stops ringing after 50 ticks.
++     * @return The number of ticks this bell has been ringing for
++     */
++    int getHasBeenRingingFor();
```
Machine-Maker:
```diff
-    int getHasBeenRingingFor();
+    int getTicksRinging();
```
or something, should have some mention of a time or ticks in the method name.

patches/api/0429-Bell-API.patch
```diff
@@ -0,0 +1,67 @@
...
++     * @return Whether the bell is currently resonating
++     */
++    boolean isResonating();
++
++    /**
++     * Returns the number of ticks this bell has been resonating for or 0 if it is not resonating.
++     * A bell stops resonating after 40 ticks.
++     * @return The number of ticks this bell has been resonating for
++     */
++    int getHasBeenResonatingFor();
```
Machine-Maker: same concern as comment on `getHasBeenRingingFor`

patches/api/0429-Bell-API.patch
```diff
@@ -0,0 +1,67 @@
...
+  */
+-public interface Bell extends TileState { }
++public interface Bell extends TileState {
++    // Paper start - Bell API
++    /**
++     * Rings the bell in the direction that it is facing.
++     */
++    default void ring() {
++        ring(((org.bukkit.block.data.type.Bell) getBlockData()).getFacing());
++    }
```
Machine-Maker: I don't think this should be a default method, it just is an unneeded conversion to API types when you can just do it in the impl without any API types created.

Machine-Maker (changes requested):

patches/api/0429-Bell-API.patch
Lama06: Thanks for all your feedback! I have done everything you suggested now.

Machine-Maker (approved):

MiniDigger (approved):

patches/server/0962-Bell-API.patch
```diff
@@ -0,0 +1,61 @@
...
++    }
++
++    @Override
++    public boolean isResonating() {
++        return getTileEntity().resonating;
++    }
++
++    @Override
++    public int getTicksResonating() {
++        return getTileEntity().resonationTicks;
```
lynxplay: I have a bit of an issue with this method.
`resonationTicks` is not reset to 0 until the next time the bell is rang.
A non resonating bell will end up with a 40 tick val on this, even if it is no longer resonating.

patches/api/0418-Bell-API.patch
```diff
@@ -0,0 +1,65 @@
...
++     * A bell stops ringing after 50 ticks.
++     * @return The number of ticks this bell has been ringing for
++     */
++    int getTicksRinging();
++
++    /**
++     * Returns the direction the bell is ringing in or null if it is not ringing.
++     * @return The direction the bell is ringing in
++     */
++    @org.jetbrains.annotations.Nullable BlockFace getRingingDirection();
```
lynxplay: This method needs a bit more javadocs.
The ringing direction is not really the direction a player rang the bell in, it is the relative block face the bell was hit on.

If I stand infront of a bell that is `facing=west` I can ring it in, as a player, in the directions EAST and WEST.
When I look in the EAST direction and I "ring" the bell, this method returns "WEST" as I hit it while looking east.

If this is a linguistic thing or not (a west wind for example is a wind that blows east etc, so a bell rining from the west to the east could be interpreted as such) does not matter I think, more documentation is required here to ensure we don't run into confusion about the returned value.

patches/api/0418-Bell-API.patch
```diff
@@ -0,0 +1,65 @@
...
+ /**
+  * Represents a captured state of Bell.
+  */
+-public interface Bell extends TileState { }
++public interface Bell extends TileState {
++    // Paper start - Bell API
++    /**
++     * Rings the bell in the direction that it is facing.
++     */
++    void ring();
```
lynxplay: This method also requires more documentation imo detailing something that, at least `ring` does imply.
The lack of sound when calling this method.

The same goes for the below method, neither will cause a sound to be played.
While this is "understandable" for developers deep in the server code, it should be documented to avoid confusion.

lynxplay (changes requested):

patches/server/0962-Bell-API.patch
Lama06: I now check whether the bell is still resonating and return 0 if it is not

patches/api/0418-Bell-API.patch
Lama06: I have now added javadocs to this method and Bell#ring(BlockFace) to make clear what is meant by the direction a bell is ringing in.
If my javadocs sound weird please correct me, I am not a native English speaker

patches/api/0418-Bell-API.patch
Lama06: I now mention this in the javadocs of this method.
Thanks for all your feedback!

lynxplay:
Upstream seems to have opened a PR that covers most of these usecases.
I am really sorry we were not able to merge it in time, concerning that Choco owns the PR in question (https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/1160/overview for those with stash access 🤡), I presume it will land in upstream at some point, making our methods potentially double if we merged this.

If you are okay with it, I'd like to keep the PR around, as it adds a method that upstream does not add and would hence work even after the currently proposed upstream PR is merged into upstream and hence paper.

Owen1212055:
Implemented by upstream https://github.com/PaperMC/Paper/commit/29b17a892d11697dce9ee35d8eab593bb4748fb1 … thanks for your PR regardless! 😁

PR #8807, opened 2023-01-24
Remove CraftItemStack#setAmount null assignment

JRoy:
This creates a problem with Paper's item serialization api where deserialized items, which are internally created as a CraftItemStack, will be completely lost if #setAmount(0) is invoked (since the underlying handle is set to null), while a regular Bukkit ItemStack
simply sets the amount field to zero, retaining the item's data.

Vanilla treats items with zero amounts the same as items with less than zero amounts, so this code doesn't create a problem with operations on the vanilla ItemStack.

Also switches the item/entity serialization apis over to dataconverter rather than dfu

the following code will not add an item to an inventory prior to this patch but now works properly
```java
    final ItemStack first = new ItemStack(Material.DIAMOND_HELMET);
    ItemStack stack = Bukkit.getUnsafe().deserializeItem(Bukkit.getUnsafe().serializeItem(first));

    ItemStack clone = stack.clone();
    clone.setAmount(0);

    clone.setAmount(stack.getAmount());
    player.getInventory().addItem(clone);
```

cc @Spottedleaf


LaxWasHere (approved):
👍

electronicboy (approved):

PR #8812, opened 2023-01-24
Modify offline mode warning to include Velocity line

mja00:
This adds a check to the offline-mode warning to see if Velocity is enabled. If it is enabled then it warns the user to make sure to secure their server and links to the Velocity documentation on that.

Attempt 2 since I had no idea about the branch name requirement.

### DIFF
--- a/patches/server/0880-Add-Velocity-IP-Forwarding-Support.patch
+++ b/patches/server/0880-Add-Velocity-IP-Forwarding-Support.patch
@@ -95,10 +95,10 @@ index 0000000000000000000000000000000000000000..c4934979b1ed85bfc4f8d9e6f8848b2b
 +    }
 +}
 diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
-index 51b3db0b6c2cede95b584268e035c0fb36d38094..42ec72c7a04d121cacbe69b544372e6c50f0e55b 100644
+index 51b3db0b6c2cede95b584268e035c0fb36d38094..a7e133f3495e9132a5fdae2c24f225e7b026295a 100644
 --- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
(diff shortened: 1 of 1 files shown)


NoahvdAa (changes requested):
This should go into the [Velocity support patch](https://github.com/PaperMC/Paper/blob/18fabc08d52ecbddd993d17fb06c51db27941014/patches/server/0882-Add-Velocity-IP-Forwarding-Support.patch)

mja00:
Time to figure out how to modify patches now. 👍🏻

mja00:
I think I did it correctly lol

MiniDigger (approved):
that dos indeed look correct.
the branch requirement is so that the team can properly modify your branch, github prohibits that on master.

one thought I had is extracting the two variables (link and software) out so that the text isn't duplicated, but prolly not worth it so approving as it

NoahvdAa (approved):
lgtm

Peter-Crawley:
I think if velocity forwarding + velocity online mode is enabled the warning should be hidden entirely

electronicboy:
This message feels weird given that velocity forwarding is much more secure than bungees; I also wouldn't mind if velocity online mode was an info message instead of a warn

Peter-Crawley:
tbh the existing message for bungeecord is kinda bad too, might be a good idea to clean that up a bit as well
i dont know if anyone wants suggestions for the wording or not though

mja00:
If someone else figures out the wording I'd gladly swap it to that (I am not a wordsmith). Otherwise I can change up the if statement so we're not duplicating that one line and swap it to INFO if wanted.

mja00:
This latest change should be what Mini was suggesting I believe. Definitely makes it cleaner. I'm not too sure how to go about swapping to using INFO if online-mode for velocity is enabled. It seems like we'd be duplicating a lot of lines for that. Unless we just change that message entirely.

Machine-Maker (approved):

Owen1212055 (approved):

PR #8816, opened 2023-01-25
Add projectile hit simulation API

SamB440:
This adds API to force a projectile to hit a provided entity. Example usage could be if you have a player disguised as another entity, you could simulate an arrow colliding with the (fake) entity hitbox. 

Simply teleporting projectiles does not suffice - it is buggy both visually and physically (they can become stuck in your hitbox), and you have to edit the velocity to force it to hit the entity which is not great for preserving damage.

I edited the existing "More Projectile API" patch, I wasn't sure whether I should make a new patch or not.

### DIFF
--- a/patches/api/0371-More-Projectile-API.patch
+++ b/patches/api/0371-More-Projectile-API.patch
@@ -178,10 +178,10 @@ index d1b37530319f6d37ee37f62080289c1e45848bc8..e94c7e279356c510f60508b26277d489
 +    // Paper end
  }
 diff --git a/src/main/java/org/bukkit/entity/Projectile.java b/src/main/java/org/bukkit/entity/Projectile.java
-index a523fca4baab447181ef91df67fa69b24e010149..34c5037682a78644c925733078b92ef505b706dc 100644
+index a523fca4baab447181ef91df67fa69b24e010149..278eefa43c5e5810a95b9e2871017124bb85faa9 100644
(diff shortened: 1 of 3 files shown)


kennytv (approved):

lynxplay (commented):
Both of these methods should probably mention that they result in calls to `ProjectileHitEvent`.

As most API methods do not call methods, I feel like it would be justified to specifically document the exception from that default.

Machine-Maker (commented):
How does a piercing arrow behave with this?

SamB440:
> How does a piercing arrow behave with this?

If you call hitEntity on your fake hitbox then it will simply continue on, potentially hitting the specified entity twice if it hits their actual hitbox. If you didn't want this to happen, it's easy enough to track the arrow and cancel the second ProjectileHitEvent. 

However, I do not think this is how I'd expect it to behave, I would expect it to have hit with the entity I provided then continue passing through that entity. Perhaps the provided entity should be added to the `piercingIgnoreEntityIds` list if it is a piercing arrow. Just unsure where that would go and what the opinions of others is.

SamB440:
I've decided to expose Projectile#canHitEntity which I think is best as it gives people the choice of whether they want to follow piercing arrow rules. Let me know what you think.

Owen1212055 (approved):

PR #8820, opened 2023-01-27
Fix force-opening enchantment tables

### LINKED ISSUE #8818: Force opening a enchanting table inventory not working
### Expected behavior

The Enchanting table should open with the force parameter like the workbench and similar blocks through the direct method instead of the need to use following way at the enchanting table:
```java
player.openInventory(Bukkit.createInventory(null, InventoryType.ENCHANTING));
```

### Observed/Actual behavior

Inventory only opening, when player is standing on an enchanting table.

### Steps/models to reproduce

1. Create a command (or something similar), which is executing following code (Force set to true):
 ```java
player.openEnchanting(null, true);
 ```
2. Execute the command (...)

### Plugin and Datapack List

Plugins (2): LuckPerms, Enchanting-Inventory

### Paper version

[00:43:45 INFO]: This server is running Paper version git-Paper-384 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 18fabc0)
Previous version: git-Paper-381 (MC: 1.19.3)

### Other

_No response_

Machine-Maker:
Actually force open an enchanting menu at the location if force is true.

---
Fixes https://github.com/PaperMC/Paper/issues/8818

### DIFF
--- a/patches/server/0960-Fix-force-opening-enchantment-tables.patch
+++ b/patches/server/0960-Fix-force-opening-enchantment-tables.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Thu, 26 Jan 2023 16:19:26 -0800
+Subject: [PATCH] Fix force-opening enchantment tables
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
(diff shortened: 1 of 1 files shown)


lynxplay (approved):

lynxplay:
Could instead write
```java
if (menuProvider == null) {
    if (!force) return null;
    
    menuProvider = new net.minecraft.world.SimpleMenuProvider((syncId, inventory, player) -> {
        return new net.minecraft.world.inventory.EnchantmentMenu(syncId, inventory, net.minecraft.world.inventory.ContainerLevelAccess.create(this.getHandle().level, pos));
    }, Component.translatable("container.enchant"));
}
```

To avoid the doubled if-else check.

kennytv (approved):

PR #8822, opened 2023-01-27
Add Entity Body Yaw Api

### LINKED ISSUE #8571: Add API for NMS LivingEntity yBodyRot

Making a shooter game and need the body rotation to calculate more specific hitboxes. Can also be used in other situations and I can't think of any reason not to add it.

Any way to get the body rotation of living entities (and set it)

Using NMS

### Other

_No response_

TheTuso:
Closes #8571

### DIFF
--- a/patches/api/0429-Add-Entity-Body-Yaw-API.patch
+++ b/patches/api/0429-Add-Entity-Body-Yaw-API.patch
@@ -1,0 +1,32 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheTuso <piotrekpasztor@gmail.com>
+Date: Thu, 2 Feb 2023 16:40:11 +0100
+Subject: [PATCH] Add Entity Body Yaw API
+
+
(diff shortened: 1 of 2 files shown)


kennytv (approved):

patches/api/0429-Add-Entity-Body-Yaw-Api.patch
```diff
@@ -0,0 +1,30 @@
...
+      * @param amount the amount of damage to do
+      */
+     void damageItemStack(org.bukkit.inventory.@NotNull EquipmentSlot slot, int amount);
++
++    /**
++     * Gets entity body yaw
++     *
++     * @return entity body yaw
++     */
++    float getBodyYaw();
```
lynxplay: The restrictions a yaw value follows are rather loosely defined (or non existent).
https://jd.papermc.io/paper/1.19/org/bukkit/Location.html#getYaw() already partially lines out the values yaw values represent.

I think this method would benefit from a `@see` reference to the location#getYaw method as well as definitions on the expected value range.

PR #8824, opened 2023-01-30
Fix `Tick::addTo`/`::between` causing SOE

emilyy-dev:
Calling `Instant::plus`/`::minus` with `Tick` as `TemporalUnit` causes the `Instant` to delegate the operation to the `TU::addTo`, which calls the same method in the `Instant`, calling each other recursively resulting in a StackOverflowError. Same process happens with `Instant::until` and `TU::between`.
In theory this can happen with any `Temporal` implementation but I would make the argument that `Instant` is by far the most commonly used.

This pull request fixes that by performing the operations with standard and well-known API elements (ChronoUnit and Duration), which should be supported even by foreign implementations.

I also added unit tests to ensure proper functionality (they failed when I introduced them before adding the fix, you'll have to take my word for it :p)

### DIFF
--- a/patches/api/0413-Add-Tick-TemporalUnit.patch
+++ b/patches/api/0413-Add-Tick-TemporalUnit.patch
@@ -6,16 +6,17 @@ Subject: [PATCH] Add Tick TemporalUnit
 
 diff --git a/src/main/java/io/papermc/paper/util/Tick.java b/src/main/java/io/papermc/paper/util/Tick.java
 new file mode 100644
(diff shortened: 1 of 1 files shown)


lynxplay (approved):
LGTM, merger might want to remove the now outdated `// following ChronoUnit#addTo` comment as this implementation does no longer follow the mentioned implementation.

emilyy-dev:
Ah, I thought the comment was about the warning suppression (which still applies)

lynxplay:
> Ah, I thought the comment was about the warning suppression (which still applies)

it may :kekw:, yea might just stay then, even if it only describes half the method now.

kennytv (approved):

PR #8834, opened 2023-02-09
Fix chunk loading/sending when the rate is set to 1.0

### LINKED ISSUE #6616: Setting target-player-chunk-send-rate to 1.0 disables the limit
### Expected behavior

When setting target-player-chunk-send-rate to 1.0, I expect the chunk send rate to be 1.

### Observed/Actual behavior

The chunk send rate is disabled (unlimited).

### Steps/models to reproduce

Start with the latest build (# 264 at time of writing), set target-player-chunk-send-rate to 1.0, and join. The chunks load at high speed (much much faster than setting target-player-chunk-send-rate to 2.0).

### Plugin list

None

### Paper version

This server is running Paper version git-Paper-264 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: ce7516d)

### Agreements

### Other

MartijnMuijsers:
Fixes #6616.

This allows the following `paper-global.yml` configuration values:
* `global-max-concurrent-loads` can be set to 1.0 and higher
  (previously could only be set to strictly higher than 1.0)
* `target-player-chunk-send-rate` and `global-max-chunk-send-rate` can be set to 10<sup>-3</sup> (1 chunk per 1000 seconds, a value that is safe and lower than anyone could ever need) and higher
  (previously could only be set to strictly higher than 1.0)

### DIFF
--- a/patches/server/0016-Rewrite-chunk-system.patch
+++ b/patches/server/0016-Rewrite-chunk-system.patch
@@ -1309,7 +1309,7 @@ index 0000000000000000000000000000000000000000..99f49b5625cf51d6c97640553cf5c420
 +}
 diff --git a/src/main/java/io/papermc/paper/chunk/PlayerChunkLoader.java b/src/main/java/io/papermc/paper/chunk/PlayerChunkLoader.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..0b060183429f4c72ec767075538477b4302bbf0d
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

MartijnMuijsers:
In addition, the rate used to be in chunks/tick, so that the minimum was 20 chunks/second, which is quite high for a minimum settable bound. Now, the rate is in chunks/second, so that the minimum is 1 chunk/second, which I also expect to be low enough to cover all environments.

PR #8839, opened 2023-02-10
Call EntityDeathEvent for EnderDragon#kill

### LINKED ISSUE #8836: /kill on EnderDragon doesnt fire EntityDeathEvent
### Expected behavior

when i /kill an entity EntityDeathEvent should fire

### Observed/Actual behavior

/killed an ender dragon and the event never fired

### Steps/models to reproduce

summon enderdragon

/kill ender dragon

nottice that EntityDeathEvent does not fire

### Plugin and Datapack List

none besides the one im developing 

### Paper version

 paper-1.19.3-388

### Other

_No response_

alfiejfs:
Fixes #8836.

### DIFF
--- a/patches/server/0961-Call-EntityDeathEvent-for-EnderDragon-kill.patch
+++ b/patches/server/0961-Call-EntityDeathEvent-for-EnderDragon-kill.patch
@@ -1,0 +1,24 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Alfie-Joe Smith <alfie@alfiesmith.net>
+Date: Fri, 10 Feb 2023 06:30:11 +0000
+Subject: [PATCH] Call EntityDeathEvent for EnderDragon#kill
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index c7caaebfb4b9f28cbe700d88fdcf232a500e8ca7..3c3a99a87094515b37a441ce3d9a3b6ef9fe3e57 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
++++ b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -616,6 +616,13 @@ public class EnderDragon extends Mob implements Enemy {
+ 
+     @Override
+     public void kill() {
++        // Paper start
++        org.bukkit.event.entity.EntityDeathEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callEntityDeathEvent(this, drops);  // Paper - call event
++        if(event.isCancelled()) {
++            return;
++        }
++        // Paper end
++
+         this.remove(Entity.RemovalReason.KILLED);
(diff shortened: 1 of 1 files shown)


patches/server/0961-Call-EntityDeathEvent-for-EnderDragon-kill.patch
```diff
@@ -0,0 +1,24 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index c7caaebfb4b9f28cbe700d88fdcf232a500e8ca7..3c3a99a87094515b37a441ce3d9a3b6ef9fe3e57 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
++++ b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -616,6 +616,13 @@ public class EnderDragon extends Mob implements Enemy {
+ 
+     @Override
+     public void kill() {
++        // Paper start
++        org.bukkit.event.entity.EntityDeathEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callEntityDeathEvent(this, drops);  // Paper - call event
```
Machine-Maker, line 16: I'm not sure we should pass in the `drops` field here. It looks like the ender dragon should drop nothing here, so passing in a list that might have something in it, might not is misleading.

Also, the callEntityDeathEvent method inside CraftEventFactory has some logic for playing death sounds. I think you need to set the `silentDeath` field on LivingEntity to true before calling the event for EnderDragon so it doesn't play the death sound by default (which would be a behavior change).

Machine-Maker (changes requested):

patches/server/0961-Call-EntityDeathEvent-for-EnderDragon-kill.patch
alfiejfs: Will fix this in a few hours. I passed in the drops as a convention (the same is done for armor stand, but I suppose that has drops) - surely it is always empty for the dragon?

Will change the silentDeath field also.

Machine-Maker: Yeah, as far as I can tell it is always going to be empty there, but we should just pass an empty list regardless, so its clear its always empty. If you pass the drops list, it might look like its supposed to have drops.

EDIT: As I'm typing this, I'm realizing that datapacks might be able to change the loot table of the dragon so it does have drops, but it looks like the kill command bypasses those in vanilla anyways. So my point still stands.

cunhar:
how can we get this blessed by the merge gods?

lynxplay:
I can look at it later today specifically in regards to MMs comment on drops👍

lynxplay:
Well, the merge gods cannot give a blessing to a PR that has not addressed the issues outlined :+1:

alfiejfs:
Apologies, have been very busy. Might get to look at it today!

lynxplay:
No worries! Take your time :) My comments were just for [cunhar](https://github.com/cunhar) :sweat_smile:

kennytv:
Superseded by https://github.com/PaperMC/Paper/pull/9495

PR #8840, opened 2023-02-11
Cleanup flying fall damage API

tal5:
Currently the flying fall damage API uses a `TriState`, but treats it like a `boolean`.
Looking at the original PR, it was a `boolean`, was changed into a `TriState` so that it can be saved to a file if it's set (and be ignored if its `UNSET`), and then saving the file was removed and it remained a `TriState` without any actual reason to be one.

This PR adds new `Player#flyingFallDamage` and `Player#flyingFallDamage(boolean)` methods that better reflect the way it works internally.
The new methods are `default` methods which use the now-deprecated `TriState` ones to avoid API breaks.

### DIFF
--- a/patches/api/0424-Flying-Fall-Damage-API.patch
+++ b/patches/api/0424-Flying-Fall-Damage-API.patch
@@ -5,26 +5,48 @@ Subject: [PATCH] Flying Fall Damage API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index b27d7414f34f1d49c56dbc33d6d23bc822adf721..b06f759b6188d87cf406072b6d7ef8266512ce50 100644
+index b27d7414f34f1d49c56dbc33d6d23bc822adf721..d5268e73ab50c577fed89b46f20ec5985089e023 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
-@@ -1416,6 +1416,23 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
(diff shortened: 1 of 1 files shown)


Owen1212055:
Having a Tri state is intentional, as it’s overriding/forcing behavior. I don’t understand the need for this.

tal5:
> Having a Tri state is intentional, as it’s overriding/forcing behavior. I don’t understand the need for this.

Well, there's no specific need, it's just confusing currently - the only place it's used in uses `toBooleanOrElse(false)`, I.e. setting it to `UNSET` is the same as setting it to `FALSE`.
I made this PR because I've gotten confused myself as to what does the `TriState` do before looking into the source and finding out it's basically a `boolean`

electronicboy:
I think that the way the tristate is used is a bit weird, but, maybe that's
fog brain; but, this partially seems like a mixture between a documentation
issue and maybe the tristates application being weird

tal5:
(See discussion on Discord @ https://discord.com/channels/289587909051416579/555462289851940864/1072662941029113956)

tal5:
> I think that the way the tristate is used is a bit weird, but, maybe that's fog brain; but, this partially seems like a mixture between a documentation issue and maybe the tristates application being weird

Well, if there's a way to make the `TriState` make sense that'd be better (or maybe I'm just misunderstanding something), but as far as I can see it's just a `true`/`false` situation
Also sorry about that, GitHub should really have a confirmation lol

Owen1212055:
No, I think this application makes every sense
The Tri state acts as an override, see frictional api, see ice ticking api, see bee “stinging” api

It represents an override, where TRUE forces it as true, UNSET reverts to vanilla behavior, and FALSE forces it to false.

I’m not sure where the confusion is coming from here.

tal5:
I don't 100% agree, but I see your point - will close this PR then

PR #8842, opened 2023-02-11
Call PlayerReadyArrowEvent for when items in the offhand are used

### LINKED ISSUE #8841: PlayerReadyArrowEvent does not fire when players hold the arrow in their offhand
### Expected behavior

That the event PlayerReadyArrowEvent fires when the player holds an arrow in their offhand, which is prioritised over other arrows in the inventory.

### Observed/Actual behavior

The event, well, does not fire.

### Steps/models to reproduce

1. Have a plugin listening to the event.
2. Hold an arrow in the hotbar, confirm that it fires.
3. Hold an arrow in your offhand, confirm that it doesn't fire despite the arrow being used when firing.

### Plugin and Datapack List

All plugins for a test server and confirmed elsewhere but just to confirm:
- AdvancedTeleport
- ATHENA
- ClearLag
- CratesPlus
- HeadsPlus
- LuckPerms
- Multiverse-Core
- PlaceholderAPI
- Spoofer
- Vault
- ViaVersion

Datapacks:
- vanilla
- file/bukkit

### Paper version

This server is running Paper version git-Paper-396 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: bd77b78)
Previous version: git-Paper-379 (MC: 1.19.3)

### Other

_No response_

Thatsmusic99:
Fixes #8841 

Combines the existing item predicates with the tryReadyArrow method.


patches/server/0210-PlayerReadyArrowEvent.patch
```diff
@@ -28,12 +28,22 @@ index e088e5174a4666130cd1f66c751e7ab83d9f7a65..8aa5f8bacb868a3b04af6e696f9eed89
      @Override
      public ItemStack getProjectile(ItemStack stack) {
          if (!(stack.getItem() instanceof ProjectileWeaponItem)) {
-@@ -2242,7 +2253,7 @@ public abstract class Player extends LivingEntity {
+             return ItemStack.EMPTY;
+         } else {
+-            Predicate<ItemStack> predicate = ((ProjectileWeaponItem) stack.getItem()).getSupportedHeldProjectiles();
++            Predicate<ItemStack> predicate = (item) -> ((ProjectileWeaponItem) stack.getItem()).getSupportedHeldProjectiles().test(item) && tryReadyArrow(stack, item); // Paper
```
Machine-Maker: You should combine these predicates with the instance `and(Predicate)` method instead.

patches/server/0210-PlayerReadyArrowEvent.patch
```diff
@@ -28,12 +28,22 @@ index e088e5174a4666130cd1f66c751e7ab83d9f7a65..8aa5f8bacb868a3b04af6e696f9eed89
...
+             ItemStack itemstack1 = ProjectileWeaponItem.getHeldProjectile(this, predicate);
+ 
+             if (!itemstack1.isEmpty()) {
+                 return itemstack1;
+             } else {
+-                predicate = ((ProjectileWeaponItem) stack.getItem()).getAllSupportedProjectiles();
++                predicate = (item) -> ((ProjectileWeaponItem) stack.getItem()).getAllSupportedProjectiles().test(item) && tryReadyArrow(stack, item); // Paper
```
Machine-Maker: see above

Machine-Maker (changes requested):
Looks good, just 1 comment, and I think you added some other random commits to the branch, like "Added exception printing when player data fails to load".

Thatsmusic99:
Crap, thought I avoided that, I had my fork originally used for debugging. Let me review that.

patches/server/0210-PlayerReadyArrowEvent.patch
Thatsmusic99: Updated using this method.

Thatsmusic99:
Should've removed the erroneous patch now, did notice the hashes in some of the other hash files changing, not sure how big of a deal that is.

patches/server/0210-PlayerReadyArrowEvent.patch
```diff
@@ -28,12 +28,23 @@ index e088e5174a4666130cd1f66c751e7ab83d9f7a65..8aa5f8bacb868a3b04af6e696f9eed89
...
+             } else {
+-                predicate = ((ProjectileWeaponItem) stack.getItem()).getAllSupportedProjectiles();
++                predicate = ((ProjectileWeaponItem) stack.getItem()).getAllSupportedProjectiles().and(item -> tryReadyArrow(stack, item)); // Paper
+ 
                  for (int i = 0; i < this.inventory.getContainerSize(); ++i) {
                      ItemStack itemstack2 = this.inventory.getItem(i);
  
 -                    if (predicate.test(itemstack2)) {
-+                    if (predicate.test(itemstack2) && tryReadyArrow(stack, itemstack2)) { // Paper
++                    if (predicate.test(itemstack2)) {
```
Machine-Maker: Looks like you added a space at the end of this line causing it to show up in the diff when it should be an "unchanged" line as you removed the diff that was originally added.

Machine-Maker:
> did notice the hashes in some of the other hash files changing, not sure how big of a deal that is.

nah, that's fine. can just ignore that.

patches/server/0210-PlayerReadyArrowEvent.patch
Thatsmusic99: Removed now, came from the paper comment that was originally there.

Machine-Maker (approved):

Doc94:
maybe im confuse this is not related to https://github.com/PaperMC/Paper/pull/7714 ? i feel when make that PR test this... or was for the not-bow things...

Machine-Maker:
Well, in that PR, you don't check each hand individually, I think its better do do that. That PR just checks the result of checking both hands. Also this PR handles it in a much cleaner way, just adding the check into the predicate once. I also don't there there's really a need for a separate event just to fix naming issues. Or if there is, it should be handled separately. Also it looks like there are some other issues with that PR.

Doc94:
> Well, in that PR, you don't check each hand individually, I think its better do do that. That PR just checks the result of checking both hands. Also this PR handles it in a much cleaner way, just adding the check into the predicate once. I also don't there there's really a need for a separate event just to fix naming issues. Or if there is, it should be handled separately. Also it looks like there are some other issues with that PR.

okay, originally that PR was only for try fix this and also include crossbow with fireworks, but get the recommendation for make a new event for avoid issues with the name.. also yes this PR handle better the thing about check the offhand... then maybe can close my original PR and wait for the merge of this to make a new PR for only the thing about the crossbows/fireworks?

Thatsmusic99:
I have tested and confirmed that with this PR the event fires with crossbows, and when using fireworks. The only potential issue I see with the current fix is that it doesn't fire when preparing an arrow in creative mode because, well, said item would be null. Have no clue whether that is intentional or not.

lynxplay:
That was the case before as well, generally there isn't a consistent arrow item that the player is preparing so I don't think we necessarily have to fire and event here :+1:

lynxplay (approved):

kennytv (approved):

PR #8848, opened 2023-02-13
convert API/server tests to mockito

Machine-Maker:
This is gonna solve a lot of future classloading order headaches. Also this is way nicer than manually building invocation handlers. At some point we should think about updating junit too, but not as needed as I feel this is.


patches/api/0003-Test-changes.patch
```diff
@@ -0,0 +1,478 @@
...
++                    // Paper start - Allow use of TYPE_USE annotations
++                    boolean warn = true;
++                    if (isWellAnnotated(method.visibleTypeAnnotations)) {
++                        warn = false;
++                    } else if (method.invisibleTypeAnnotations != null) {
++                        dance: for (final org.objectweb.asm.tree.TypeAnnotationNode invisibleTypeAnnotation : method.invisibleTypeAnnotations) {
++                            final org.objectweb.asm.TypeReference ref = new org.objectweb.asm.TypeReference(invisibleTypeAnnotation.typeRef);
++                            if (ref.getSort() == org.objectweb.asm.TypeReference.METHOD_RETURN && java.util.Arrays.binarySearch(ACCEPTED_ANNOTATIONS, invisibleTypeAnnotation.desc) >= 0) {
++                                warn = false;
++                                break dance; // cha cha real smooth
```
MiniDigger, line 181: lol

MiniDigger (approved):
I really like this, this is long overdue. 

junit5 can wait till we have source in repo I guess, patching that would be kinda messy.

PR #8855, opened 2023-02-15
Add TeleportFlags

Owen1212055:
Abstracts relative teleport flags and instead makes a generic TeleportFlag option. This has the benefit of being able to easily add new flags in the future.

This adds a new flag, which allows you to keep inventories open when teleporting players (vanilla behavior). 

These are breaking changes to the teleport api, however, it's still marked as experimental so I find this a fair change.

### DIFF
--- a/patches/api/0375-More-Teleport-API.patch
+++ b/patches/api/0375-More-Teleport-API.patch
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
 +     */
 +    EYES;
 +}
-diff --git a/src/main/java/io/papermc/paper/entity/RelativeTeleportFlag.java b/src/main/java/io/papermc/paper/entity/RelativeTeleportFlag.java
+diff --git a/src/main/java/io/papermc/paper/entity/TeleportFlag.java b/src/main/java/io/papermc/paper/entity/TeleportFlag.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..0426ee8bd71142b6f933a479c0f2e5ef6043147d
+index 0000000000000000000000000000000000000000..61bf11978a712be805afb34506a4171ed1972777
 --- /dev/null
-+++ b/src/main/java/io/papermc/paper/entity/RelativeTeleportFlag.java
-@@ -0,0 +1,34 @@
++++ b/src/main/java/io/papermc/paper/entity/TeleportFlag.java
+@@ -0,0 +1,83 @@
 +package io.papermc.paper.entity;
 +
 +import org.bukkit.Location;
 +import org.bukkit.event.player.PlayerTeleportEvent;
++import org.jetbrains.annotations.ApiStatus;
 +
 +/**
-+ * Represents coordinates in a teleportation that should be handled relatively.
++ * Represents a flag that can be set on teleportation that may
++ * slightly modify the behavior.
 + *
-+ * @see org.bukkit.entity.Player#teleport(Location, PlayerTeleportEvent.TeleportCause, boolean, boolean, RelativeTeleportFlag...)
++ * @see EntityStateFlag
++ * @see RelativeTeleportFlag
 + */
-+@org.jetbrains.annotations.ApiStatus.Experimental
-+public enum RelativeTeleportFlag {
++@ApiStatus.Experimental
++public sealed interface TeleportFlag permits TeleportFlag.EntityStateFlag, TeleportFlag.RelativeTeleportFlag {
++
 +    /**
-+     * Represents the player's X coordinate
++     * Note: These flags only work on {@link org.bukkit.entity.Player} entities.
++     * <p>
++     * Represents coordinates in a teleportation that should be handled relatively.
++     * <p>
++     * Coordinates of the location that the client should handle as relative teleportation
++     * Relative teleportation flags are only used client side, and cause the player to not lose velocity in that
++     * specific coordinate. The location of the teleportation will not change.
++     *
++     * @see org.bukkit.entity.Player#teleport(Location, PlayerTeleportEvent.TeleportCause, TeleportFlag...)
 +     */
-+    X,
++    @org.jetbrains.annotations.ApiStatus.Experimental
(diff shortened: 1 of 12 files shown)


emilyy-dev (commented):
I like this API and its setting, mostly minor things/corrections but the entry names in EntityStateFlag I find non descriptive

emilyy-dev:
okay github decided to eat and throw away my comments, thanks a lot

patches/api/0387-More-Teleport-API.patch
```diff
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
...
++     * Represents coordinates in a teleportation that should be handled relatively.
++     * <p>
++     * Coordinates of the location that the client should handle as relative teleportation
++     * Relative teleportation flags are only used client side, and cause the player to not lose velocity in that
++     * specific coordinate. The location of the teleportation will not change.
++     *
++     * @see org.bukkit.entity.Player#teleport(Location, PlayerTeleportEvent.TeleportCause, TeleportFlag...)
 +     */
-+    YAW,
++    @org.jetbrains.annotations.ApiStatus.Experimental
```
emilyy-dev:
```suggestion
+    @ApiStatus.Experimental
```
`ApiStatus` is imported

patches/api/0387-More-Teleport-API.patch
```diff
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
...
++    }
++
 +    /**
-+     * Represents the player's pitch
++     * Represents flags that effect the entity's state on
++     * teleportation.
 +     */
-+    PITCH;
++    @ApiStatus.Experimental
++    enum EntityStateFlag implements TeleportFlag {
```
emilyy-dev: Adding this comment here because it involves the three ESF entries.
Maybe it's just me but I find the names do not communicate what effect they actually perform (or not); `IGNORE_OPEN_INVENTORY`, ignore what about it? Perhaps something like `RETAIN_...` is a bit more self-descriptive and communicative, or some other wording; `IGNORE_PASSENGERS`, again, what about them? `RETAIN_...`/`KEEP_...` conveys more sense; same deal with `IGNORE_DISMOUNT`

patches/api/0387-More-Teleport-API.patch
```diff
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
...
++        /**
++         * If all passengers should not be required to be removed prior to teleportation.
++         *
++         * Note
++         * Teleporting to a different world with ignorePassengers to true while the entity has entities riding it
```
emilyy-dev: "with ignorePassengers to true" doesn't feel right now that they are (or, would be) a flag set than boolean parameters, "with this flag included" or "when this flag is set" sounds more fitting, (I assume this was copy/pasted as-is though)

patches/api/0387-More-Teleport-API.patch
```diff
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
...
++         * will cause this teleportation to return false and not occur.
++         */
++        IGNORE_PASSENGERS,
++        /**
++         * If the entity should not be dismounted if they are riding another entity.
++         *
++         * Note:
++         * Teleporting to a different world with dismount to false while this entity is riding another entity will
```
emilyy-dev: Same thing as above, "with dismount to false" sounds even more confusing, implying it can explicitly be set to false (?), "without this flag" or "when this flag is excluded" or however else

emilyy-dev (commented):
beep boop

patches/api/0387-More-Teleport-API.patch
Owen1212055: Covering your previous comment too, 100% agree. Kinda just moved those docs, should defo be adjusted due to them being flags now.

patches/api/0387-More-Teleport-API.patch
Owen1212055: Yeah, it's a bit tricky I think. I do like the whole retain prefix.

"RETAIN_PASSENGERS", "RETAIN_OPEN_INVENTORY", "RETAIN_VEHICLE"

How do these sound?

emilyy-dev: Yeah, I think that makes it more obvious what does/doesn't happen without recourse to the docs, (at least not to check the details about cross-world teleportation, that can't exactly be included in the name.)

kennytv (approved):

Machine-Maker (dismissed):
I'm not a fan of the current nesting/naming of the types. I think most people don't import inner classes (and I'd argue its good to have your intellij set to not do that), so doing `TeleportFlag.RelativeTeleportFlag.YAW` is very verbose. If you are going to have both enums nested in the common interface, I wouldn't duplicate the `TeleportFlag` part so instead it would be `TeleportFlag.Relative.YAW` or `TeleportFlag.EntityState.IGNORE_PASSENGERS`. Or just don't have the enums nested.

Owen1212055:
Resolved, opinons?

patches/api/0375-More-Teleport-API.patch
```diff
@@ -35,51 +35,100 @@ index 0000000000000000000000000000000000000000..7713a5b7e06da185685f18e79eae2c97
...
++     * Represents flags that effect the entity's state on
++     * teleportation.
 +     */
-+    PITCH;
++    @ApiStatus.Experimental
++    enum EntityState implements TeleportFlag {
++        /**
++         * If all passengers should not be required to be removed prior to teleportation.
++         * <p>
++         * Note
```
emilyy-dev:
```suggestion
+         * Note:
```
You missed a spot :p
Other than that, I'm satisfied with the changes 👍🏻

emilyy-dev:
Shouldn't this API also expand to the teleportAsync methods?

Owen1212055:
> Shouldn't this API also expand to the teleportAsync methods?

Once this API is finalized, most likely.

PR #8860, opened 2023-02-18
Update the flag when a captured block state is outdated

### LINKED ISSUE #8858: Blocks do not update when planting trees
### Expected behavior

wood normal display
image

### Observed/Actual behavior

There are three logs that are not updated.
It only shows up when you right click on the sapling, or re-enter the game.
image

### Steps/models to reproduce

1. Place 2x2 dark oak saplings
2. Ripen with bone meal
3. you can see

### Plugin and Datapack List

not any

### Paper version

This server is running Paper version git-Paper-404 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: f9dc371)
Previous version: git-Purpur-1916 (MC: 1.19.3)

### Other

_No response_

Lulu13022002:
When a tree grows there's multiple update that happens in the same position but since the captured block state list doesn't support that it just override the last one with the new one but without updating the update flag.
Closes https://github.com/PaperMC/Paper/issues/8858


Machine-Maker (approved):
Looks right. Seems a similar thing is done in the blockstatelist populator and in capturedBlockStates.

kennytv (approved):

PR #8861, opened 2023-02-18
Configure mokitor Server mock as stubOnly

lynxplay:
A recent patch moved the internal unit tests to mokito, allowing deep mocking to easily setup a mocked server instance. While this change is useful, the server's Server#getItemFactory methods is one of the hottest paths during unit testing, being called numerous times by material tests.
As mokito mocks keep track of each invocation to allow for verifications of invocations down the line, the server mock allocates a huge amount of memory to keep track of all invocations, ultimately leading to an OOM exception.

The previous solution solved this by increasing the tests memory to 2 GB, however as of right now simply configuring the server mock as "stubOnly", properly prevents the overflow of invocation records as none of the unit test code relies on invocation verification.

### DIFF
--- a/patches/server/0004-Test-changes.patch
+++ b/patches/server/0004-Test-changes.patch
@@ -18,7 +18,7 @@ index 064a0c5b2a4b379ceec84652ff1e004d57419115..218a0db30ac8e4403da4ee50ca0a67ca
      runtimeOnly("mysql:mysql-connector-java:8.0.29")
 diff --git a/src/test/java/io/papermc/paper/testing/DummyServer.java b/src/test/java/io/papermc/paper/testing/DummyServer.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..286790061c9d81c872108ef63e1a1aba2fbec809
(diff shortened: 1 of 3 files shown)


Owen1212055 (approved):

patches/server/0004-Test-changes.patch
```diff
@@ -52,7 +52,7 @@ index 0000000000000000000000000000000000000000..286790061c9d81c872108ef63e1a1aba
 +            return;
 +        }
 +
-+        final Server dummyServer = mock(Server.class);
++        final Server dummyServer = mock(Server.class, org.mockito.Mockito.withSettings().stubOnly());
```
Machine-Maker: Import it here, this is a new file Paper made.

Machine-Maker (changes requested):

patches/server/0004-Test-changes.patch
lynxplay: Done :+1:

Machine-Maker (approved):

MiniDigger (approved):

kashike (approved):

PR #7650, opened 2022-03-22
Restore vanilla entity drops behavior

Machine-Maker:
Supersedes https://github.com/PaperMC/Paper/pull/4882
Fixes https://github.com/PaperMC/Paper/issues/4743

Instead of just tracking the itemstacks, this tracks with it, the action to take with that itemstack to apply the correct logic on dropping the item instead of generalizing it for all dropped items like CB does.

There are some ABI breaks **(in NMS)** with various methods, specifically the call death event ones. but, ugh, its just super ugly if we wanna keep that.

PR #8862, opened 2023-02-18
Fix fox drops on death

Lulu13022002:
When a fox died the item in its mouth doesn't have a 100% chance to drops and can even be damaged using the summon command (and having a great luck). This is due to them relying on the generic mob logic before the fox loot.
Also closes MC-153010

### DIFF
--- a/patches/server/0246-Improve-death-events.patch
+++ b/patches/server/0246-Improve-death-events.patch
@@ -234,10 +234,31 @@ index 9e0555b1e98bba6a9305c3996c69347e47d73204..bd70046b4fef307b6b8b626a65f0b4a3
      // CraftBukkit start
      public int getExpReward() {
 diff --git a/src/main/java/net/minecraft/world/entity/Mob.java b/src/main/java/net/minecraft/world/entity/Mob.java
-index 4ae0f36276592e37aeb5f881b713efa76d086f8e..c4b05c7e339a2ec01ee3f5e4dec0b3493788d8d2 100644
+index 4ae0f36276592e37aeb5f881b713efa76d086f8e..236b10b86cdf6e0a723d8c9f199dde9cc983198e 100644
 --- a/src/main/java/net/minecraft/world/entity/Mob.java
 +++ b/src/main/java/net/minecraft/world/entity/Mob.java
-@@ -1075,7 +1075,13 @@ public abstract class Mob extends LivingEntity implements Targeting {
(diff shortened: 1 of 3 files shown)


Machine-Maker (changes requested):
This unfixes https://bugs.mojang.com/browse/MC-147598. Baby foxes should still drop held items when killed, but the dropCustomDeathLoot function is only called for adult ageable mobs.

Lulu13022002:
Should be good now, all the fox now always drop their bitten item (without being damaged for the summon-ed one).

patches/server/0255-Improve-death-events.patch
```diff
@@ -220,10 +220,31 @@ index 7c0e1963383a1a2862930cf77844e5c8c80e70e3..4c1992bcffd629ef53f14c5a0146eab2
...
+index 984a33d5e1f790a9c78ba57f2dc21fb072a44b3d..5524a26f6401731872b11c0360fc6eb6cc2e042a 100644
 --- a/src/main/java/net/minecraft/world/entity/Mob.java
 +++ b/src/main/java/net/minecraft/world/entity/Mob.java
-@@ -1065,7 +1065,13 @@ public abstract class Mob extends LivingEntity implements Targeting {
+@@ -1047,6 +1047,12 @@ public abstract class Mob extends LivingEntity implements Targeting {
+ 
+     }
+ 
++    // Paper start
++    protected boolean shouldLootFrom(EquipmentSlot slot) { // method to avoid to fallback into the global mob loot logic (i.e fox)
```
Machine-Maker: I think this boolean should be inverted the whole way through. To avoid a double flip of the value that just adds confusing. It should be `shouldSkipLoot`, and then adjust everything else accordingly.

Machine-Maker (changes requested):

Machine-Maker (approved):

lynxplay (approved):
Foxes now drop their items in their main hand 100% of the time woop.
Leaving to machine to merge for a final look again if needed after rebase.

Incluuu:
This broke the old loot from players. Now when a player dies, items are scattered up to 10 squares.

Warriorrrr:
It sounds like you are describing what #7650 fixed, items being scattered far is how it works in vanilla.

Lulu13022002:
Yep see the related issue: https://github.com/PaperMC/Paper/issues/10152

PR #8864, opened 2023-02-19
Fix plugin updater logic, allow null update directory

Owen1212055:


patches/server/0013-Paper-Plugins.patch
```diff
@@ -483,7 +486,37 @@ index 0000000000000000000000000000000000000000..39e503f52684e928c61c6d27bf61727a
...
 +        this.pluginDirectory = ((File) minecraftOptionSet.valueOf("plugins")).toPath();
-+        this.updateDirectory = this.pluginDirectory.resolve(YamlConfiguration.loadConfiguration(configFileLocationBukkit).getString("settings.update-folder", "update"));
++
++        String updateDirectory = YamlConfiguration.loadConfiguration(configFileLocationBukkit).getString("settings.update-folder", "update");
++        if (updateDirectory.isBlank()) {
++            this.updateDirectory = null;
++        } else {
++            Path resolvedUpdateDirectory = this.pluginDirectory.resolve(updateDirectory);
++            if (!Files.isDirectory(resolvedUpdateDirectory)) {
++                this.updateDirectory = null;
```
MiniDigger, line 496: Quietly disabling the update mechanism here is a bit meh, should log a warning since obviously the admin did something wrong

Owen1212055: The issue with this is that it includes if the directory is not there. That is the current behavior for that, at least. (It's always been this behavior for non-directory configured update directories)

patches/server/0013-Paper-Plugins.patch
```diff
@@ -4712,14 +4745,15 @@ index 0000000000000000000000000000000000000000..cfc34bbb32d5cf19dbb100003a4aae62
 +     */
 +    private Path checkUpdate(Path file) throws Exception {
 +        PluginInitializerManager pluginSystem = PluginInitializerManager.instance();
-+        if (!Files.isDirectory(pluginSystem.pluginUpdatePath())) {
++        Path updateDirectory = pluginSystem.pluginUpdatePath();
++        if (updateDirectory == null || !Files.isDirectory(updateDirectory)) {
```
MiniDigger, line 4749: Technically we don't need to check if it's a dir here anymore, right? Guess it doesn't matter

Owen1212055: Wanted to check here *just* incase this is changed while the server is still running.

patches/server/0013-Paper-Plugins.patch
MiniDigger: Oh right, I was thinking about the case where it's a file, not a folder, but yours is obviously more likely.
Is there any reason we shouldn't create the folder? 
Would be out of scope here I guess but generally help with discovery of this feature.

MiniDigger (approved):

patches/server/0013-Paper-Plugins.patch
Owen1212055: Reason for not making a file is because that at least has been the default logic for a while... and you know.. people go crazy over stuff like that.

MiniDigger: Right, people are obsessed with their layouts and stuff, still something we could keep in mind for the future, discoverability for stuff like this warrents a change like this

PR #8873, opened 2023-02-20
Redefine Cyclic Dependencies

Owen1212055:
Spigot Plugins:
- If dependency is in ``loadbefore`` and in ``depend``/``softdepend``, ignore the load order induced by dependency and just make it load before. This means that it will always load before now. 
- If plugin has a 1-1 dependency loop (PluginX -> PluginZ, PluginZ -> PluginX), an effectively "random" load order will be chosen.
- Remove the class loading warning for spigot plugins, we have no reason to care about it.

This should remove *most* cyclic dependency issues, as so many of these were introduced as a hack to get rid of the class loading warnings. 

Paper Plugins:
- Dependencies NO LONGER affect load order. This means that you can have "cyclic dependencies", however, you cannot have cyclic load order.
- New ``loadafter`` field in paper-plugin.yml

---
Download the paperclip jar for this pull request: [paper-8873.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/565080593.zip)


mbax:
> * If dependency is in `loadbefore` and in `depend`/`softdepend`, ignore the load order induced by dependency and just make it load before. This means that it will always load before now.

IMO this should be changed so that `depend` takes priority, as it's the most "this must happen" of them, refusing to load the plugin if the declared dependency isn't present. Proposed priority: `depend`, `loadbefore`, `softdepend`

Owen1212055:
Paper plugins will now load slightly differently under the legacy loader, this is done to help preserve legacy behavior with spigot plugins.

Owen1212055:
I've setup the build jar tag, once that finishes building if people are willing to test it would be much appreicated.

Mostly, those who had to add the ``-Dpaper.useLegacyPluginLoading=true`` tag. Please test to see if it's still screaming at you to the extent it did before (example plugins like GriefPrevention).

WillFP:
Brilliant PR, can confirm this fixes all the problems I've encountered!

(Relevant `/pl`: Actions, Boosters, eco, EcoArmor, EcoBits, EcoBosses, EcoCrates, EcoEnchants, EcoItems, EcoJobs, EcoPets, EcoShop, EcoSkills, Essentials, EssentialsSpawn, HolographicDisplays, LuckPerms, PlaceholderAPI, PlugMan, ProtocolLib, Reforges, spark, StatTrackers, Talismans, Vault)

WillFP:
Yep, this works all good now - fixes _many_ plugins breaking altogether since the new plugin loader was introduced.
This is a very high priority fix as otherwise thousands of server owners are going to have many plugins break when they update to versions with the new plugin loader.

Good work Owen!

kashike (approved):

Bobcat00:
> I'm just sad that there's no way to win here.

Perhaps the way to win was to have left it alone in the first place.

kennytv:
Perhaps the real treasure was the friendship we made along the way

Bobcat00:
Sounds like it worked for reasons no one understood, which is the definition of a kludge. But many times, kludges should be left alone. (Like the time my employer made changes to the payroll direct deposit schedule for no good reason.)

electronicboy:
it worked because for the most part the issue was invisible until it caused stupid issues for devs due to stuff like that breaking the load order of plugins in unexpected ways, often in a manner that was hard to diagnose in other servers, especially if you're unaware of how broken that system was

Even spigot took a "recent" stab at trying to resolve issues with this system

MiniDigger:
> Sounds like it worked for reasons no one understood, which is the definition of a kludge. But many times, kludges should be left alone. (Like the time my employer made changes to the payroll direct deposit schedule for no good reason.)

that mentality is cancerous, since if you can't understand it, you cant ever touch this, leading to stagnation.

reworking this was much needed, we knew it wasn't gonna be easy, we know more about the stuff devs do now, it made the paper plugin system more robust now.

stuff like this is how we can move the ecosystem out of the deadlock it has been for the past years

Strahilchu:
This crashed my server

MiniDigger:
please provide a log

mercurialmusic:
Also crashed for me. Had to add the new flag in startup, which worked. Log: https://gist.github.com/mercurialmusic/decf93df0402d18e74045a25c39d1ed9
On git-Paper-412 (MC: 1.19.3). Will also report to CMI. Not sure what Heroes is though... I don't have a plugin with that name...

Owen1212055:
> Also crashed for me. Had to add the new flag in startup, which worked. Log: https://gist.github.com/mercurialmusic/decf93df0402d18e74045a25c39d1ed9
> On git-Paper-412 (MC: 1.19.3). Will also report to CMI. Not sure what Heroes is though... I don't have a plugin with that name...

Generally, this PR does not resolve large loops like that as those are still what we want to stop from loading due to the unexpected issues that those can cause. It’s most likely tricky for CMI due to it being a “monolith” plugin.

Bobcat00:
In the above referenced issue from malonnnn:
> I do not use CMI on my server, nor heroes, despite those being in the chain

So is Paper considering all the softdepends and loadbefores listed in the plugins, even if the referenced plugins aren't being used on the server? If so, why?

mercurialmusic:
This ended up messing so many plugins. Votifier, DiscordSRV, not even sure what else yet, even with the flag enabled. I just rolled it back and everything magically started working again.

Owen1212055:
> This ended up messing so many plugins. Votifier, DiscordSRV, not even sure what else yet, even with the flag enabled. I just rolled it back and everything magically started working again.

Because alot of these plugins were doing unsupported things with the plugin system. That is totally independent from the plugin loading system changes.

Owen1212055:
> In the above referenced issue from malonnnn:
> 
> > I do not use CMI on my server, nor heroes, despite those being in the chain
> 
> So is Paper considering all the softdepends and loadbefores listed in the plugins, even if the referenced plugins aren't being used on the server? If so, why?

See https://github.com/PaperMC/Paper/pull/8890, thank you for the pointer here.

Strahilchu:
I'll just use an older version until you guys figure out what you're doing.

Owen1212055:
> I'll just use an older version until you guys figure out what you're doing.

It’s rather hard to account for plugins who are abusing the plugin system using reflection. 😄

PR #8881, opened 2023-02-21
Properly re-add providers when reloading

Owen1212055:
Paper plugins will explode when reloading, but, I added a little warning.


patches/server/0013-Paper-Plugins.patch
```diff
@@ -534,6 +540,43 @@ index 0000000000000000000000000000000000000000..9926eb59b83abffffb578356f148f045
...
++                hasPaperPlugin = true;
++                break;
++            }
++        }
++
++        if (hasPaperPlugin) {
++            LOGGER.error("======== WARNING ========");
++            LOGGER.error("You are reloading while having Paper plugins installed on your server.");
++            LOGGER.error("Paper plugins do NOT support being reloaded. This will cause some unexpected issues.");
++            LOGGER.error("=========================");
```
kennytv: Should be warnings

kennytv (approved):

patches/server/0013-Paper-Plugins.patch
Machine-Maker: ADDRESSED

Owen1212055:
LGTM

PR #8884, opened 2023-02-21
Do not send plugin headers when they are not required

Jakubk15:


Owen1212055 (changes requested):
Please merge this into the already existing paper plugin patch.

PR #8886, opened 2023-02-21
Add missing deprecation

Lulu13022002:
Add missing deprecation around the resource pack api
And also deprecate the spigot silenceable lightning stuff. The lightning sound has been moved clientside and this will just
create some side effect like not spreading more fire in normal+ diff or doesn't power the lightning rod. The only change that could be intended is the sculk sensor that will be unable to listen to that vibration for silent lightning.

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -2133,7 +2133,7 @@ index f3afe67f0832cb828d25be3654518ff73a80b0e1..598abaa82c634178043a29f6caa6ac52
 +    // Paper end
  }
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 3d84a75a6999129353a9ead48f9db25bc80f685b..a5010cb46aa5ef69c559ef9fc717f90365c05807 100644
+index 3d84a75a6999129353a9ead48f9db25bc80f685b..e5a54a2f35d83f8c7a5f2b6512b86481aa7f9f51 100644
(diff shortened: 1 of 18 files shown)


Machine-Maker (commented):
So the lightning silent stuff.... I still see that isSilent field on the entity being used on the server. Not for the sound exactly, but a silent strike won't spawn fire. So there is still some use here for that.

patches/api/0005-Adventure.patch
```diff
@@ -2471,15 +2471,23 @@ index 3d84a75a6999129353a9ead48f9db25bc80f685b..a5010cb46aa5ef69c559ef9fc717f903
       * @param url The URL from which the client will download the resource
       *     pack. The string must contain only US-ASCII characters and should
       *     be encoded as per RFC 1738.
-@@ -1358,6 +1570,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+@@ -1327,6 +1539,7 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      * @throws IllegalArgumentException Thrown if the hash is not 20 bytes
+      *     long.
+      */
++    @Deprecated // Paper
+     public void setResourcePack(@NotNull String url, @Nullable byte[] hash);
```
lynxplay, line 2483: Can you also fix the faulty javadocs of the old setResourcePack methods that do not FQN the Component ?

Given we don't import them, the javadocs link to the wrong method rn.

Lulu13022002: Yes, sure

Machine-Maker (approved):

PR #8889, opened 2023-02-21
Do not send plugin headers when it is not required.

Jakubk15:
This PR removes the redundant plugin headers while they are not required. Imagine a situation when you have no Paper plugins and some Spigot plugins. It will look like this, which is pretty ugly:

```
Paper plugins:
Bukkit plugins:
- plugin 1
- plugin 2
```

The Paper plugins header is redundant here.


AVuTuVA:
And if a person does not have plugins at all? It will just ignore it...

Jakubk15:
I tested it. It will print Server Plugins(0):

Machine-Maker (approved):

kennytv (approved):

MiniDigger (approved):

Owen1212055 (approved):

PR #8896, opened 2023-02-22
Additional ExplosiveMinecart API

SoSeDiK:
Adds methods to:
- `#isIgnited()` - check if the Minecart is ignited (its fuse is primed)
- `#ignite()` - prime the fuse
  differs from `#setFuseTicks()` by following vanilla behavior (4 seconds before the explosion, notifying the players to play the animation, playing sound if not silent); `#setFuseTicks` can be used additionally after to change the ticks before the explosion)
- `#explode()` and `#explode(power)` - to immediately explode the Minecart (parameterless method follows vanilla behavior)

### DIFF
--- a/patches/api/0304-Missing-Entity-Behavior-API.patch
+++ b/patches/api/0304-Missing-Entity-Behavior-API.patch
@@ -6,6 +6,7 @@ Subject: [PATCH] Missing Entity Behavior API
 Co-authored-by: Nassim Jahnke <nassim@njahnke.dev>
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 Co-authored-by: William Blake Galbreath <blake.galbreath@gmail.com>
+Co-authored-by: SoSeDiK <mrsosedik@gmail.com>
 
 diff --git a/src/main/java/com/destroystokyo/paper/entity/ai/VanillaGoal.java b/src/main/java/com/destroystokyo/paper/entity/ai/VanillaGoal.java
 index dddbb661265aa23f88d93d0681f418f40a872351..998f629852e1103767e005405d1f39c2251ecd28 100644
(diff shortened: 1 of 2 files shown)


olijeffers0n (approved):

patches/api/0304-Missing-Entity-Behavior-API.patch
```diff
@@ -1176,16 +1177,16 @@ index 7cc1d9a966454af70b7c25735fe474fe3eb97eb4..ad2eaee347cd2864aef30d2af48c1be6
...
 +++ b/src/main/java/org/bukkit/entity/minecart/ExplosiveMinecart.java
-@@ -6,4 +6,19 @@ import org.bukkit.entity.Minecart;
+@@ -6,4 +6,44 @@ import org.bukkit.entity.Minecart;
   * Represents a Minecart with TNT inside it that can explode when triggered.
   */
  public interface ExplosiveMinecart extends Minecart {
 +    // Paper start - Entity API
 +    /**
-+     * Set the number of ticks until the minecart explodes after being primed.
++     * Set the number of ticks until the minecart explodes after being primed
```
Machine-Maker: Why? We usually have complete sentences in the paragraph potion of the doc, and fragments for each param and return.

patches/api/0304-Missing-Entity-Behavior-API.patch
```diff
@@ -1197,7 +1198,32 @@ index a4411daca0d8a770b0d15e08847b82df8a3ec27f..50464aa8c38cd47fc049ee022e25cb49
...
 +     */
 +    int getFuseTicks();
-+    // Paper end
++
++    /**
++     * Checks whether the fuse is primed
++     *
++     * @return whether the fuse is primed
++     */
++    boolean isPrimed();
```
Machine-Maker: Method should be isIgnited() as it is on a Creeper. 

(Another PR should add this method to PrimedTNT)

patches/api/0304-Missing-Entity-Behavior-API.patch
```diff
@@ -1197,7 +1198,32 @@ index a4411daca0d8a770b0d15e08847b82df8a3ec27f..50464aa8c38cd47fc049ee022e25cb49
...
++
++    /**
++     * Primes the fuse
++     */
++    void primeFuse();
```
Machine-Maker: Method should be ignite() as it is on creeper.

patches/server/0647-Missing-Entity-Behavior-API.patch
```diff
@@ -818,6 +820,26 @@ index 9569068c0a6e1eb934f063634f5d3efe674aa33d..dfd0b5e018194343ca40629db6f70c60
...
++    }
++
++    @Override
++    public void explode() {
++        explode(this.getHandle().getDeltaMovement().horizontalDistanceSqr());
++    }
++
++    @Override
++    public void explode(double power) {
++        this.getHandle().explode(power);
```
Machine-Maker, line 842: You need to have a check that it’s nonnegative as its square root is taken inside that method.

Machine-Maker (changes requested):

patches/api/0304-Missing-Entity-Behavior-API.patch
SoSeDiK: Javadocs seem to differ from place to place, there's no standard that I'm aware of? Even in the present patch, the setter has a dot, and the getter doesn't, and since I usually don't place a dot if there's only one sentence, I removed it.

But as per request placed dots instead! :)

patches/api/0304-Missing-Entity-Behavior-API.patch
SoSeDiK: I chose the method namings based on Mojang mappings.
Updated to be on pair with Creeper instead.

patches/server/0647-Missing-Entity-Behavior-API.patch
SoSeDiK: Nice catch! Added a check for that

patches/api/0304-Missing-Entity-Behavior-API.patch
SoSeDiK: And sure, I'll make a PR adding the method to PrimedTNT later

Machine-Maker (approved):

kennytv (approved):

PR #8866, opened 2023-02-19
Small paper plugin fixes

Owen1212055:
Some fixes that can be merged whenever. 
Aren't really critical.

PR #8901, opened 2023-02-22
Close library classloader and improve PluginDescriptionFile

lynxplay:
Readds changes already proposed in a previous PR which were lost, presumably during the fixup process.

Specifically, this commit aims to readd the two previously proposed improvements:

1) Close the library classloader when a paper plugins classloader
   is closed.

2) Pass soft- and normal dependencies as well as the load-before plugins
   specified by a paper plugin to the dummy PluginDescriptionFile
   implementation spigot plugins expect via Plugin#getDescription().

See: #8866


Owen1212055 (approved):
This was originally done by me but was reverted when I messed up at some point.
LGTM

kennytv (approved):

Machine-Maker (approved):

PR #8907, opened 2023-02-23
Fix SpawnEggMeta#get/setSpawnedType

Machine-Maker:
This is one way to fix the wrongly (imo) deprecated get/setSpawnedType methods on SpawnEggMeta. Spawn eggs set a custom entity type be spawned that isn't the type associated with the item. This lets plugins get and set that custom type. This allows spawn eggs for entities that don't otherwise have a spawn egg item (like item frames).


kennytv (approved):

kennytv:
As per contrib discussion, make two new methods (and ideally mark the old two for removal)

PR #8910, opened 2023-02-24
Add allowsWrapping property to paper plugin meta to bypass return class restriction

### LINKED ISSUE #8892: Allow proxying main plugin class

Currently, the paper plugin loader prohibits returning a class that is not inherited (or is) the main class of plugin.

The part:
```java
// Don't allow plugins to load plugins other than the one defined in main. This restriction might not be necessary.
if (!plugin.getClass().isAssignableFrom(Class.forName(PaperPluginParent.this.description.getMainClass(), true, plugin.getClass().getClassLoader()))) {
  throw new IllegalArgumentException("Plugin provided must be the same type as main defined in plugin configuration!");
}
```

e.g. I want to create my own plugin interface, and then provide proxy class for its implementations. Say I have my own plugin interface named KotlinPlugin, and then I want people to be able to implement it, and use it as simple as JavaPlugin implementations, just with my own bootstrap. My bootstrap would wrap that implementation in a class that implements JavaPlugin and proxies calls to my own interface methods.
That way the restriction limits the flexibility of the new plugin system without a real reason (?).

Provide a property in the plugin config (e.g. allowWrappers) to bypass that restriction.

Or just remove that restriction at all.

### Other

_No response_

viktorhrekh:
Optionally allow plugin bootstrap to return plugin other than the declared as main.
It is useful to provide wrappers for own interfaces, and in general, makes the plugin system more flexible.
This PR basically adds the `allowsWrapping` property to the paper plugin meta and then in the PaperPluginParent checks if it is allowed, and bypasses the restriction if it is.

Resolves #8892.

Discussed it with a few maintainers, no one of them was against this idea, so going ahead to implement :)

### DIFF
--- a/patches/server/0962-Allow-plugin-bootstrap-to-provide-plugin-of-other-ty.patch
+++ b/patches/server/0962-Allow-plugin-bootstrap-to-provide-plugin-of-other-ty.patch
@@ -1,0 +1,75 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: viktorhrekh <viktorhrekh@viesoft.dev>
+Date: Fri, 24 Feb 2023 11:56:04 +0200
+Subject: [PATCH] Allow plugin bootstrap to provide plugin of other type than
+ the plugin main if allowed in meta
+
+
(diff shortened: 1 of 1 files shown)


viktorhrekh:
@electronicboy @Machine-Maker

Owen1212055:
In all honesty, remove the config but maybe add a little info message.

PR #8913, opened 2023-02-24
Expand PlayerDeathEvent API

dawon:
Allows to set death message on the death screen separately from the message sent to all Players

### DIFF
--- a/patches/api/0463-Expand-PlayerDeathEvent-API.patch
+++ b/patches/api/0463-Expand-PlayerDeathEvent-API.patch
@@ -1,0 +1,118 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jakub Zacek <dawon@dawon.eu>
+Date: Tue, 13 Feb 2024 18:58:54 +0100
+Subject: [PATCH] Expand PlayerDeathEvent API
+
+
+diff --git a/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java b/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
+index 66e9d65a8dd05bed05d0ab46ec64206a6dae0507..9be6b982a52f95c5dfd8be53f00168c3461598b0 100644
+--- a/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
++++ b/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
+@@ -17,24 +17,28 @@ public class PlayerDeathEvent extends EntityDeathEvent {
+     private boolean keepLevel = false;
+     private boolean keepInventory = false;
+     private boolean doExpDrop; // Paper - shouldDropExperience API
++    // Paper start - Expand PlayerDeathEvent API
++    private net.kyori.adventure.text.Component deathScreenMessage;
++    private boolean displayDeathMessage;
++    // Paper end - Expand PlayerDeathEvent API
+     // Paper start - adventure
+     @org.jetbrains.annotations.ApiStatus.Internal
+-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final @Nullable net.kyori.adventure.text.Component deathMessage) {
+-        this(player, drops, droppedExp, 0, deathMessage);
++    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final @Nullable net.kyori.adventure.text.Component deathMessage, boolean displayDeathMessage) { // Paper - Expand PlayerDeathEvent API
++        this(player, drops, droppedExp, 0, deathMessage, displayDeathMessage); // Paper - Expand PlayerDeathEvent API
+     }
+ 
+     @org.jetbrains.annotations.ApiStatus.Internal
+-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final @Nullable net.kyori.adventure.text.Component deathMessage) {
+-        this(player, drops, droppedExp, newExp, 0, 0, deathMessage);
++    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final @Nullable net.kyori.adventure.text.Component deathMessage, boolean displayDeathMessage) { // Paper - Expand PlayerDeathEvent API
++        this(player, drops, droppedExp, newExp, 0, 0, deathMessage, displayDeathMessage); // Paper - Expand PlayerDeathEvent API
+     }
+ 
+     @org.jetbrains.annotations.ApiStatus.Internal
+-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable net.kyori.adventure.text.Component deathMessage) {
++    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable net.kyori.adventure.text.Component deathMessage, boolean displayDeathMessage) { // Paper - Expand PlayerDeathEvent API
+         // Paper start - shouldDropExperience API
+-        this(player, drops, droppedExp, newExp, newTotalExp, newLevel, deathMessage, true);
++        this(player, drops, droppedExp, newExp, newTotalExp, newLevel, deathMessage, true, displayDeathMessage); // Paper - Expand PlayerDeathEvent API
+     }
+     @org.jetbrains.annotations.ApiStatus.Internal
+-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable net.kyori.adventure.text.Component deathMessage, final boolean doExpDrop) {
++    public PlayerDeathEvent(final @NotNull Player player, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable net.kyori.adventure.text.Component deathMessage, final boolean doExpDrop, boolean displayDeathMessage) { // Paper - death screen api
+         // Paper end - shouldDropExperience API
+         super(player, drops, droppedExp);
+         this.newExp = newExp;
+@@ -42,9 +46,12 @@ public class PlayerDeathEvent extends EntityDeathEvent {
+         this.newLevel = newLevel;
+         this.deathMessage = deathMessage;
+         this.doExpDrop = doExpDrop; // Paper - shouldDropExperience API
++        this.deathScreenMessage = deathMessage; // Paper - Expand PlayerDeathEvent API
++        this.displayDeathMessage = displayDeathMessage; // Paper - Expand PlayerDeathEvent API
+     }
+     // Paper end - adventure
+ 
++    @Deprecated // Paper
+     public PlayerDeathEvent(@NotNull final Player player, @NotNull final List<ItemStack> drops, final int droppedExp, @Nullable final String deathMessage) {
+         this(player, drops, droppedExp, 0, deathMessage);
+     }
+@@ -69,6 +76,8 @@ public class PlayerDeathEvent extends EntityDeathEvent {
+         this.newLevel = newLevel;
+         this.deathMessage = net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserializeOrNull(deathMessage); // Paper
+         this.doExpDrop = doExpDrop; // Paper - shouldDropExperience API
++        this.deathScreenMessage = this.deathMessage; // Paper - Expand PlayerDeathEvent API
++        this.displayDeathMessage = true; // Paper - Expand PlayerDeathEvent API
+     }
+ 
+     @Deprecated // Paper
+@@ -160,6 +169,49 @@ public class PlayerDeathEvent extends EntityDeathEvent {
+         return this.deathMessage;
+     }
+     // Paper end - adventure
++    // Paper start - Expand PlayerDeathEvent API
++    /**
++     * Set the death message that will appear on the death screen of the dying player.
++     * <p>
++     * If set to null, no message will be shown to the dying player.
++     * <p>
++     * If the message exceeds 256 characters it will be truncated.
++     *
++     * @param deathScreenMessage Message to appear on the death screen to the dying player.
++     */
++    public void deathScreenMessage(@Nullable net.kyori.adventure.text.Component deathScreenMessage) {
++        this.deathScreenMessage = deathScreenMessage;
++    }
++
++    /**
++     * Get the death message that will appear on the death screen of the dying player.
++     * By default, it is the same value as {@link #deathMessage()}.
++     *
++     * @return Message to appear on the death screen to the dying player.
++     */
++    public @Nullable net.kyori.adventure.text.Component deathScreenMessage() {
++        return this.deathScreenMessage;
++    }
++
++    /**
++     * Returns whether the death message in this event should be displayed according to
++     * `showDeathMessages` game rule.
++     *
++     * @return whether the death message in this event should be displayed
++     */
++    public boolean shouldDisplayDeathMessage() {
++        return displayDeathMessage;
++    }
++
++    /**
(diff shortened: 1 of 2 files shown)


patches/server/0962-Add-player-death-screen-message-API.patch
```diff
@@ -0,0 +1,53 @@
...
+ 
+         net.kyori.adventure.text.Component deathMessage = event.deathMessage() != null ? event.deathMessage() : net.kyori.adventure.text.Component.empty(); // Paper - Adventure
+ 
+-        if (deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty() && flag) { // Paper - Adventure // TODO: allow plugins to override?
+-            Component ichatbasecomponent = PaperAdventure.asVanilla(deathMessage); // Paper - Adventure
+-
+-            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), ichatbasecomponent), PacketSendListener.exceptionallySend(() -> {
++        // Paper start - death screen message api
++        net.kyori.adventure.text.Component deathScreenMessage = event.deathScreenMessage() != null ? event.deathScreenMessage() : deathMessage;
++        if ((true || deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty()) && flag) { // Paper - Adventure
```
XanderWander, line 21: ``

Xander-de-Keijzer: Why is this true here?

Xander-de-Keijzer:
```
boolean hasDeathMessage = deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty();
boolean hasDeathScreenMessage = deathScreenMessage != null && deathScreenMessage != net.kyori.adventure.text.Component.empty();

if ((hasDeathMessage || hasDeathScreenMessage) && flag) {
    Component ichatbasecomponent = hasDeathScreenMessage ? PaperAdventure.asVanilla(deathScreenMessage) : PaperAdventure.asVanilla(deathMessage);
```
    
Maybe something like this?

olijeffers0n: > Why is this true here?

To minimise the Diff normally

Xander-de-Keijzer: > ```
> boolean hasDeathMessage = deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty();
> boolean hasDeathScreenMessage = deathScreenMessage != null && deathScreenMessage != net.kyori.adventure.text.Component.empty();
> 
> if ((hasDeathMessage || hasDeathScreenMessage) && flag) {
>     Component ichatbasecomponent = hasDeathScreenMessage ? PaperAdventure.asVanilla(deathScreenMessage) : PaperAdventure.asVanilla(deathMessage);
> ```
> 
> Maybe something like this?

nvm im stupid lol

dawon: Yup it is there to minimize the diff. I have moved this if bit below... :-)

patches/api/0418-Add-player-death-screen-message-API.patch
```diff
@@ -0,0 +1,63 @@
...
+         this.doExpDrop = doExpDrop;
++        this.adventure$deathScreenMessage = null;
+     }
+     // Paper end
+ 
+@@ -67,6 +69,7 @@ public class PlayerDeathEvent extends EntityDeathEvent {
+         this.newLevel = newLevel;
+         this.deathMessage = deathMessage;
+         this.adventure$deathMessage = deathMessage != null ? net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(deathMessage) : null; // Paper
++        adventure$deathScreenMessage = null; // Paper
```
lynxplay: Should this really be null ? Especially when the death message is initialised to its correct value here.
I feel like this should be equal to the death message.

patches/server/0962-Add-player-death-screen-message-API.patch
lynxplay: This diff is rather large for how little it needs to do especiall due to the change in control flow that i would personally not suggest.

It might be worth simply evaluating the deathScreenMessage before the branching control flow at the top, similar to how `net.kyori.adventure.text.Component deathMessage` is computed.

To also include `flag` or well the gamerule, it could be computed as follows:
`net.kyori.adventure.text.Component deathScreenMessage = event.deathScreenMessage() != null && flag ? event.deathScreenMessage() : net.kyori.adventure.text.Component.empty(); // Paper - death screen api`

Leaving it empty if either no death message is supposed to be displayed or a plugin set the component to null.
From there on out, we can literally just replace the usages both inside the following if block and in the else block with the variable.

This should get us to a diff such as
```diff
diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
index 289429eb4..b31d359b4 100644
--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
+++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
@@ -922,13 +922,14 @@ public class ServerPlayer extends Player {
         }

         net.kyori.adventure.text.Component deathMessage = event.deathMessage() != null ? event.deathMessage() : net.kyori.adventure.text.Component.empty(); // Paper - Adventure
+        var deathScreenMessage = PaperAdventure.asVanilla(event.deathScreenMessage() != null && flag ? event.deathScreenMessage() : net.kyori.adventure.text.Component.empty()); // Paper - death screen api

         if (deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty() && flag) { // Paper - Adventure // TODO: allow plugins to override?
             Component ichatbasecomponent = PaperAdventure.asVanilla(deathMessage); // Paper - Adventure

-            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), ichatbasecomponent), PacketSendListener.exceptionallySend(() -> {
+            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), deathScreenMessage), PacketSendListener.exceptionallySend(() -> {
                 boolean flag1 = true;
-                String s = ichatbasecomponent.getString(256);
+                String s = deathScreenMessage.getString(256);
                 MutableComponent ichatmutablecomponent = Component.translatable("death.attack.message_too_long", Component.literal(s).withStyle(ChatFormatting.YELLOW));
                 MutableComponent ichatmutablecomponent1 = Component.translatable("death.attack.even_more_magic", this.getDisplayName()).withStyle((chatmodifier) -> {
                     return chatmodifier.withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, ichatmutablecomponent));
@@ -948,7 +949,7 @@ public class ServerPlayer extends Player {
                 this.server.getPlayerList().broadcastSystemMessage(ichatbasecomponent, false);
             }
         } else {
-            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), CommonComponents.EMPTY));
+            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), deathScreenMessage));
         }
```
which achieves the same thing.

lynxplay (changes requested):

patches/api/0418-Add-player-death-screen-message-API.patch
dawon: You are right, this is probably a more intuitive approach for the developers, thank you, fixed!

patches/server/0962-Add-player-death-screen-message-API.patch
dawon: And I thought my diff was already small enough... :D Thank you, rebased on master and fixed!

patches/server/0966-Add-player-death-screen-message-API.patch
```diff
@@ -0,0 +1,37 @@
...
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -922,13 +922,14 @@ public class ServerPlayer extends Player {
+         }
+ 
+         net.kyori.adventure.text.Component deathMessage = event.deathMessage() != null ? event.deathMessage() : net.kyori.adventure.text.Component.empty(); // Paper - Adventure
++        Component deathScreenMessage = PaperAdventure.asVanilla(event.deathScreenMessage() != null && flag ? event.deathScreenMessage() : net.kyori.adventure.text.Component.empty()); // Paper - death screen api
+ 
+-        if (deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty() && flag) { // Paper - Adventure // TODO: allow plugins to override?
++        if (deathMessage != null && deathMessage != net.kyori.adventure.text.Component.empty() && flag) { // Paper - Adventure
```
lynxplay: Why the removal here ? I don't think that comment related to the death screen api more to the `flag` or well gamerule.

dawon: Shall I also add an api to turn on/off display death messages as well then?

Machine-Maker: Probably not in this PR. Or actually you probably can. Just some boolean to override that behavior initial value is the value of that `flag` variable.

dawon: I rebranded the PR & commits names and added API to get/set whether the death message should be shown.

patches/api/0419-Expand-PlayerDeathEvent-API.patch
```diff
@@ -0,0 +1,111 @@
...
++     *
++     * @return Message to appear on the death screen to the dying player.
++     */
++    public @Nullable net.kyori.adventure.text.Component deathScreenMessage() {
++        return this.adventure$deathScreenMessage;
++    }
++
++    /**
++     * Returns whether the death message in this event should be shown according to
++     * `showDeathMessages` game rule.
```
Machine-Maker: Clarify that the value is **initially** set by the gamerule to show death messages. Cause plugins can change it and there's no guarantee that its the same value when a listener is called. Also clarify what `shown` means. There's a method to set the deathScreenMessage, I feel it could easily be confused. Maybe mention something about controlling the `{@link #deathMessage()}` being sent to other players. This controls both the message being sent to players and the message shown on screen to the player who died.

patches/api/0419-Expand-PlayerDeathEvent-API.patch
```diff
@@ -0,0 +1,111 @@
...
++     *
++     * @return whether the death message in this event should be shown
++     */
++    public boolean shouldShowDeathMessage() {
```
Machine-Maker: I would use the verb `send` or `display` here instead of `show`. I think `show` too closely associates the message with the `deathScreenMessage`.

patches/server/0966-Expand-PlayerDeathEvent-API.patch
```diff
@@ -0,0 +1,62 @@
...
++                String s = deathScreenMessage.getString(256); // Paper - death screen api
+                 MutableComponent ichatmutablecomponent = Component.translatable("death.attack.message_too_long", Component.literal(s).withStyle(ChatFormatting.YELLOW));
+                 MutableComponent ichatmutablecomponent1 = Component.translatable("death.attack.even_more_magic", this.getDisplayName()).withStyle((chatmodifier) -> {
+                     return chatmodifier.withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, ichatmutablecomponent));
+@@ -948,7 +949,7 @@ public class ServerPlayer extends Player {
+                 this.server.getPlayerList().broadcastSystemMessage(ichatbasecomponent, false);
+             }
+         } else {
+-            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), CommonComponents.EMPTY));
++            this.connection.send(new ClientboundPlayerCombatKillPacket(this.getCombatTracker(), deathScreenMessage)); // Paper - death screen api
```
Machine-Maker: This diff shouldn't be needed. If the death message shouldn't be sent/shown, then an empty component should always be passed to the `ClientboundPlayerCombatKillPacket`.

patches/server/0966-Expand-PlayerDeathEvent-API.patch
```diff
@@ -0,0 +1,62 @@
...
+-        org.bukkit.event.entity.PlayerDeathEvent event = CraftEventFactory.callPlayerDeathEvent(this, loot, PaperAdventure.asAdventure(defaultMessage), defaultMessage.getString(), keepInventory); // Paper - Adventure
++        org.bukkit.event.entity.PlayerDeathEvent event = CraftEventFactory.callPlayerDeathEvent(this, loot, PaperAdventure.asAdventure(defaultMessage), defaultMessage.getString(), keepInventory, flag); // Paper - Adventure
+         // Paper start - cancellable death event
+         if (event.isCancelled()) {
+             // make compatible with plugins that might have already set the health in an event listener
+@@ -922,13 +922,14 @@ public class ServerPlayer extends Player {
+         }
+ 
+         net.kyori.adventure.text.Component deathMessage = event.deathMessage() != null ? event.deathMessage() : net.kyori.adventure.text.Component.empty(); // Paper - Adventure
++        Component deathScreenMessage = PaperAdventure.asVanilla(event.deathScreenMessage() != null && event.shouldShowDeathMessage() ? event.deathScreenMessage() : net.kyori.adventure.text.Component.empty()); // Paper - death screen api
```
Machine-Maker: This can just be a local var above the `ClientboundPlayerCombatKillPacket` constructor. It doesn't have to exist outside the scope of the following if statement. Just a simple
```java
event.deathScreenMessage() != null ? PaperAdventure.asVanilla(event.deathScreenMessage()) : ichatbasecomponent
```

Machine-Maker (changes requested):

patches/server/0966-Expand-PlayerDeathEvent-API.patch
dawon: Well what about case, when you don't want to set the deathMessage to other players  but want to show the death message to the player dying? Thinking of it, I am afraid the diff cannot be as simple as that, because this would miss the check for the length of the message...

Machine-Maker: You're billing the shouldShowDeathMessages method as a way to give per-death control over the gamerule, it should follow the rules of that gamerule. And that gamerules is either all or nothing. If you want to add further control over things, then what that method is should be rethought or repurposed.

Machine-Maker: > because this would miss the check for the length of the message...

the length check is inside the if statement that's below it. It's not skipping it.

dawon: I was more talking about the case, when you set the deathMessage null, deathScreenMessage not null and shouldShowDeathMessage to true. In this case it should only show the death message to the player dying.

Machine-Maker: Wouldn't this then be a change in behavior for existing plugins? Say I already have a listener that sets the death message to null. Now, the player who died suddenly does get a message on their screen when they didn't before.

dawon: You're right, this would break the original behavior. My original implementation would not break it, but as lynxplay pointed out it would be counterintuitive and would make a case when you don't want to show the message to the dying player but to everyone else yes more complicated (you would have to send empty component instead of null).

Ok, another possible solution is to deprecate current death message methods (that would set message for both the one dying and everyone else) and introduce a new pair of functions to set/get the message for everyone and the one dying separately. I can't think of a cleaner solution... What do you think?

dawon: Should I go with the abovementioned solution? Deprecating the old message and introducing a pair of new ones?

dawon:
For now rebased on master

patches/api/0463-Expand-PlayerDeathEvent-API.patch
```diff
@@ -0,0 +1,118 @@
...
++    public boolean shouldDisplayDeathMessage() {
++        return displayDeathMessage;
++    }
++
++    /**
++     * Sets whether the death message in this event should be shown.
++     *
++     * @param displayDeathMessage whether the death message in this event should be shown.
++     */
++    public void setDisplayDeathMessage(boolean displayDeathMessage) {
```
lynxplay, line 112: This name is kind of misleading. `setDisplayDeathMessage` sounds like it sets the message, when in reality it only sets if the message is shown.

dawon: Would setShouldDisplayDeathMessage be better? :D Or should I also omit the get/set prefixes?

dawon:
Replaced with #12221.

PR #8914, opened 2023-02-24
Log message if plugin bootstrap provided external plugin instead of throwing an exception

viktorhrekh:
Makes the plugin bootstrap more flexible about wrapping your own plugin interfaces into JavaPlugin implementation


viktorhrekh:
@Owen1212055, created another PR as it seems easier and cleaner :)

patches/server/0962-Log-warning-instead-of-exception-if-plugin-bootstrap.patch
```diff
@@ -0,0 +1,22 @@
...
+--- a/src/main/java/io/papermc/paper/plugin/provider/type/paper/PaperPluginParent.java
++++ b/src/main/java/io/papermc/paper/plugin/provider/type/paper/PaperPluginParent.java
+@@ -172,9 +172,8 @@ public class PaperPluginParent {
+                     plugin = bootstrap.createPlugin(PaperPluginParent.this.context);
+                 }
+ 
+-                // Don't allow plugins to load plugins other than the one defined in main. This restriction might not be necessary.
+                 if (!plugin.getClass().isAssignableFrom(Class.forName(PaperPluginParent.this.description.getMainClass(), true, plugin.getClass().getClassLoader()))) {
+-                    throw new IllegalArgumentException("Plugin provided must be the same type as main defined in plugin configuration!");
++                    logger.warning("Plugin " + getMeta().getName() + " uses bootstrap that is returning another plugin.");
```
Machine-Maker: Does this even have to be a warning? I would just log an info that contains the type of the provided plugin and something about it not matching the configured main class.

Machine-Maker (changes requested):

lynxplay (approved):

lynxplay:
LGTM, generally should go into the original paper plugins patch instead of its own patch.
You can do that yourself or the developer or maintainer merging this can do it for you :+1:

viktorhrekh:
@lynxplay @Machine-Maker

Machine-Maker (approved):

PR #8917, opened 2023-02-25
[ci skip] Grammar correction

Chris6ix:
Basically #8898 and #8899, but correct.
The hyphen of "High-performance" should just be added to "About" :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ you will most likely use this for WSL), `homebrew` (macOS / Linux), and more:
   - [Adoptium](https://adoptium.net/) has builds for most operating systems.
   - Paper requires JDK 17 to build, however, makes use of Gradle's
     [Toolchains](https://docs.gradle.org/current/userguide/toolchains.html)
-    feature to allow building with only JRE 8 or later installed. (Gradle will
+    feature to allow building with only JRE 11 or later installed. (Gradle will
     automatically provision JDK 17 for compilation if it cannot find an existing
     install).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,11 +4,11 @@ PaperMC is happy you're willing to contribute to our projects. We are usually
 very lenient with all submitted PRs, but there are still some guidelines you
 can follow to make the approval process go more smoothly.
 
-## Use a Personal Fork and not Organization
+## Use a Personal Fork and not an Organization
 
 Paper will routinely modify your PR, whether it's a quick rebase or to take care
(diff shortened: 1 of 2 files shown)


Jakubk15 (approved):

lynxplay (approved):

CONTRIBUTING.md
```diff
@@ -29,7 +29,7 @@ you will most likely use this for WSL), `homebrew` (macOS / Linux), and more:
 - `git` (package `git` everywhere);
 - A Java 17 or later JDK (packages vary, use Google/DuckDuckGo/etc.).
   - [Adoptium](https://adoptium.net/) has builds for most operating systems.
-  - Paper requires JDK 17 to build, however makes use of Gradle's
+  - Paper requires JDK 17 to build, however, makes use of Gradle's
```
Machine-Maker, line 32: Something about this still reads wrong to me. I feel like there needs to be an `it` before `makes use of` to be correct. But please correct me if I'm wrong.

Machine-Maker (changes requested):

CONTRIBUTING.md
olijeffers0n: It is fine as it is

Chris6ix: To me, that sounds wrong. `it` would refer to `Paper`, but there is no need to refer to it again, `Paper` is already named, and the connection to `makes use of` is clear.

I double-checked with my [Language checker](https://languagetool.org/?pk_campaign=addon2-popup-logo), and it seems to agree with me.

Machine-Maker (approved):

CONTRIBUTING.md
```diff
@@ -29,7 +29,7 @@ you will most likely use this for WSL), `homebrew` (macOS / Linux), and more:
...
     [Toolchains](https://docs.gradle.org/current/userguide/toolchains.html)
     feature to allow building with only JRE 8 or later installed. (Gradle will
```
jpenilla: Can you update this to 11?

jpenilla (changes requested):

MiniDigger (approved):

CONTRIBUTING.md
Chris6ix: > Can you update this to 11?

Only this specific one, or the ones above and below too?

Chris6ix: Wait, are you talking about 17 or 8?

jpenilla: 8 -> 11
paperweight needs a minimum JRE 11 runtime now instead of 8

Chris6ix: Done :)

jpenilla (approved):

PR #8920, opened 2023-02-26
Registry Modification API

Machine-Maker:
Adds two new "events" that PluginBootstrap can register to which enable modification and addition to both built-in and data-driven registries. This opens up tons of new possibilities for plugins to customize things with API that was previously only possible with nms/reflection.

**Data-Driven**: Types which are loaded purely from datapacks (custom and vanilla)
**Built-In**: Loaded via source code, never read from a file

There are two events plugins can listen to, one is fired when an "entry" is about to be added to a registry, either a built-in or data-driven. This allows modification of the "entry" before it's registered. The second event is fired right before the registry is frozen (to prevent new additions). This is used to add completely new entries. 

### Demo
There are two "demonstration" apis added, one for ChatType and one for GameEvent. Just allows some very simply customization, more just to show how it works for both built-in and data-driven types.

---

```[tasklist]
### TODO
- [x] https://github.com/PaperMC/Paper/pull/8642
- [ ] https://github.com/PaperMC/Paper/pull/9009
- [ ] https://github.com/PaperMC/Paper/pull/9233
- [x] better naming (not attached to any of them)
- [x] add specific "event" types instead to allow for more context to be added to listeners
- [x] ~~add more registry api, like getOrThrow(), hasKey(), helper util methods like that.~~ will be handled in different pr
- [x] Either remove, or fully flesh out the two "demo" API this adds (added GameEvent)
- [x] Add registry keys for other already-existing registries from Registry.java
- [x] Move Paper plugin changes to Paper Plugins patch
- [x] Add exception handling to event fires
- [x] ~~Discuss any "removal" or "cancel addition" API~~ left for future possible API
- [x] Should static final RegistryKey be in the RegistryKey or should we separate them? Like BuiltInRegistries could have keys and instances and DataDrivenRegistries would just have keys. **No, they should all be in one file like vanilla has now with Registries.java**
- [ ] Discuss having the "builders" under a separate API policy as the types they apply to change very frequently.
- [x] move registry modification patch into Paper Registry API patch
- [ ] https://github.com/PaperMC/Paper/pull/9629
- [ ] https://github.com/PaperMC/Paper/pull/10154
```


patches/api/0418-Registry-Modification-API.patch
```diff
@@ -0,0 +1,436 @@
...
++import io.papermc.paper.registry.RegistryBuilder;
++import org.bukkit.Keyed;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++@ApiStatus.Experimental
++@ApiStatus.NonExtendable
++public interface ModifyRegistryEvent<T extends Keyed, B extends RegistryBuilder<T>> extends RegistryEvent<T, B> {
++
++    @NotNull B currentBuilder();
```
jpenilla: currentBuilder kind of implies it's not mutable. just builder or getBuilder would be better imo.

patches/api/0418-Registry-Modification-API.patch
```diff
@@ -0,0 +1,436 @@
...
++package io.papermc.paper.registry.event;
++
++import io.papermc.paper.registry.RegistryBuilder;
++import org.bukkit.Keyed;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++@ApiStatus.Experimental
++@ApiStatus.NonExtendable
++public interface ModifyRegistryEvent<T extends Keyed, B extends RegistryBuilder<T>> extends RegistryEvent<T, B> {
```
jpenilla: ModifyRegistryEntryEvent

patches/api/0418-Registry-Modification-API.patch
```diff
@@ -0,0 +1,436 @@
...
++import io.papermc.paper.registry.RegistryView;
++import java.util.function.Consumer;
++import net.kyori.adventure.key.Key;
++import net.kyori.adventure.key.Keyed;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++@ApiStatus.NonExtendable
++@ApiStatus.Experimental
++public interface ModifiableRegistryView<T extends Keyed, B extends RegistryBuilder<T>> extends RegistryView<T> {
```
jpenilla: I don't like Modifiable since it implies more than registration (editing existing entries, etc)

Hard to come up with a better name here though

patches/api/0418-Registry-Modification-API.patch
```diff
@@ -0,0 +1,436 @@
...
++import net.kyori.adventure.key.Key;
++import net.kyori.adventure.key.Keyed;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++@ApiStatus.Experimental
++@ApiStatus.NonExtendable
++public interface RegistryView<T extends Keyed> extends Iterable<T> {
```
jpenilla: I know we discussed something like this, but I thought we decided it wasn't needed. What's the benefit of RegistryView and its subtypes instead of just subtyping registry?

Machine-Maker: Because we need a different impl for the `get()` methods. The current impl, just delegates to the registry get methods (which will throw because its not frozen yet). I think this is better than creating different `Registry` instances for different points in the startup. That way the only `Registry` instances created are the final "immutable" ones.

patches/api/0418-Registry-Modification-API.patch
Machine-Maker: I too don't like Modifiable. but everything else is worse. I don't like `AdditionSupportingRegistryView` or `RegisterableRegistryView`. `MutableRegistryView` might work.

Maybe the register method shouldn't be part of the view, and just a separate thing on the event.

patches/api/0418-Registry-Modification-API.patch
jpenilla: couldn't we just have the `get` implementation behave differently before freezing?

Machine-Maker: We could, but the static fields in Registry are filled when this happens, I don't want to support people doing `Registry.GAME_EVENT.get()` inside a listener for the game event registry. Plus, any methods we (or upstream) adds to Registry like anything tag related would also exist in the listeners then. Tags are loaded way later. 

Like a `Registry#hasKey()` would be confusing. `Registry.GAME_EVENT.hasKey()`'s returned value shouldn't ever change. It should throw if its not ready, or always return the same value. The "view" is kinda like a view of it being built, its separate.

We could call it `PartialRegistry` or something separate. Something to clarify its an incomplete picture. But it shouldn't extend Registry.

jpenilla: The name is fine, not permitting accessing the registry outside the provided context is a decent enough reason for this

patches/api/0418-Registry-Modification-API.patch
kennytv: I'd say Writable > Mutable > the rest and I think it's better as part of the view

patches/server/0962-Registry-Modification-API.patch
```diff
@@ -0,0 +1,1239 @@
...
+index 6e6d9573115eeedef4cb24beac43f1bc96eebb49..befe662974bd3d6b4d0a2d69f9bfef46ec93986b 100644
+--- a/src/main/java/net/minecraft/core/MappedRegistry.java
++++ b/src/main/java/net/minecraft/core/MappedRegistry.java
+@@ -80,6 +80,14 @@ public class MappedRegistry<T> implements WritableRegistry<T> {
+             return MappedRegistry.this.getTags().map(Pair::getSecond);
+         }
+     };
++    // Paper start
++    @Nullable
++    private Map<ResourceLocation, T> beforeFrozen = new HashMap<>(2048);
```
kennytv: 2000 is a weird initial size here, as a greta majority of registries have under 100 entries (with the notable exception of items, blocks, sounds, and block states)

Machine-Maker: I chose that because that's what it is for all the other maps in MappedRegistry, so do you want all of them changed?

patches/server/0964-Support-registry-mod-API-with-GameEvent.patch
```diff
@@ -0,0 +1,132 @@
...
++        super(key);
++        this.nms = nms;
++    }
++
++    @Override
++    public int getRange() {
++        return this.nms.getNotificationRadius();
++    }
++
++    public static final class Builder implements GameEvent.Builder, PaperRegistryBuilder<net.minecraft.world.level.gameevent.GameEvent, GameEvent> {
```
Lulu13022002: Maybe include the vibration level inside the builder and the backed object ? The map seems to be static so i think it would be a good idea since there's no way otherwise. By default all the new game event will fallback with a vibration of zero.

patches/server/0964-Support-registry-mod-API-with-GameEvent.patch
```diff
@@ -0,0 +1,132 @@
...
++            }
++        }
++
++        @Override
++        public NamespacedKey getKey() {
++            return this.key;
++        }
++
++        @Override
++        public int range() {
```
Lulu13022002: Is there a way to do that without duplicating all the getter into the builder, this can become really complex on some other objects like Biome or DimensionType. I was thinking of something like that to have only the setters on the builder and then the getter on the final object:
```Java
 public void beforeRegister(final RegistryAdditionEvent<GameEvent, GameEvent.Builder> event) {
            if (event(.getOldObject()).key().equals(TO_MODIFY)) {
                event.builder().range(event.getOldObject().getRange() * 2);
            }
        }
```

Machine-Maker: I don't think I'm a fan of that. I certainly don't like actually constructing the API type here. Sometimes not even the NMS type is constructed yet, and so that would require making API type impls not use nms types. 

I did consider having the builder also extend the api type, but I don't really like that either. It just feels too limiting if there are then methods added to the API type that aren't simple getters.

Machine-Maker:
Updated for 1.19.4 and moved the vibration level to the builder so that it can be changed. Also added a test to make sure the match the map that was previously used.

patches/server/0964-Support-registry-mod-API-with-GameEvent.patch
Lulu13022002: Ah yes nevermind

patches/server/0962-Support-registry-mod-API-with-GameEvent.patch
```diff
@@ -0,0 +1,310 @@
...
+@@ -16,4 +16,13 @@ public class GameEventTest extends AbstractTestingBase {
+             Assert.assertNotNull("Bukkit should not be null " + nms, bukkit);
+         }
+     }
++    // Paper start
++    @Test
++    public void testVibrationLevels() {
++        for (final net.minecraft.world.level.gameevent.GameEvent nms : BuiltInRegistries.GAME_EVENT) {
++            final int expectedLevel = net.minecraft.world.level.gameevent.vibrations.VibrationListener.VIBRATION_FREQUENCY_FOR_EVENT.getInt(nms);
++            Assert.assertEquals(nms.builtInRegistryHolder().key().location() + " doesn't have the expected vibration level: expectedLevel", expectedLevel, nms.vibrationLevel);
```
Lulu13022002: the expectedLevel is not interpreted here since it's a string

Machine-Maker:
For the GameEvent builder, I decided to not make the vibration level configurable there. They are really 2 different systems, the vibration stuff and game events. Not all game events are listened to for vibrations, so I think a better solution is to use the game event bukkit events to change the level of certain game events. 

I also reverted the change that changed all the static final fields in GameEvent to Reference because we are now going with code generation to generate a file full of keys for GameEvent for use with the registry mod api.

patches/api/0246-Registry-Modification-API.patch
```diff
@@ -0,0 +1,703 @@
...
++/**
++ * The {@link #INSTANCE} provides access to registries given a {@link RegistryKey}.
++ * <p>
++ * Replacement for {@link org.bukkit.Bukkit#getRegistry(Class)}.
++ */
++@ApiStatus.NonExtendable
++@ApiStatus.Experimental
++public interface RegistryAccess {
++
++    RegistryAccess INSTANCE = ServiceLoader.load(RegistryAccess.class).iterator().next();
```
Owen1212055: There's a nice Services.load() method that adventure offers so you don't have to use any iterator funnies.

Additionally, are we sure we want this as a publicly exposed field instead of an access method?

patches/api/0246-Registry-Modification-API.patch
```diff
@@ -0,0 +1,703 @@
...
++     * to be frozen. Useful for adding new entries to registries
++     * that support custom entries.
++     *
++     * @param registryKey the registry key
++     * @param listener an instance of the listener
++     * @param <T> object type
++     * @param <B> object builder type
++     */
++    <T extends Keyed, B extends RegistryBuilder<T>> void registerPreFreezeListener(@NotNull RegistryKey<T> registryKey, @NotNull RegistryPreFreezeListener<T, B> listener);
++}
```
Owen1212055: It may just be me, but I find the usage of "register" in the context of registries despite being used for the listener system be a bit confusing. 

Although it would deviate from PluginManger's registerEvent naming scheme.. I think it may be better to reduce confusion.
Maybe something like:
- ``newRegisterListener``, ``newRegistrationListener``, ``onRegistration``
- ``onFinalization``, ``onPreFreeze``, ``onRegistryFinalization``

I am not sure, the point is, I think that we should perhaps not use the term register here as it may be a bit confusing. Although, just seeing what others might think here.

patches/server/0546-Registry-Modification-API.patch
```diff
@@ -0,0 +1,1281 @@
...
+             loader.getSecond().load(resourceManager, registryInfoLookup);
+             // Paper start
+             if (loader.getFirst().key() == Registries.CHAT_TYPE) {
+-                Registry.register((Registry<ChatType>) loader.getFirst(), ChatType.RAW, new ChatType(new net.minecraft.network.chat.ChatTypeDecoration("%s", List.of(net.minecraft.network.chat.ChatTypeDecoration.Parameter.CONTENT), net.minecraft.network.chat.Style.EMPTY), new net.minecraft.network.chat.ChatTypeDecoration("%s", List.of(net.minecraft.network.chat.ChatTypeDecoration.Parameter.CONTENT), net.minecraft.network.chat.Style.EMPTY))); // CraftBukkit
++
++                final net.minecraft.network.chat.ChatType rawType = new net.minecraft.network.chat.ChatType(new net.minecraft.network.chat.ChatTypeDecoration("%s", java.util.List.of(net.minecraft.network.chat.ChatTypeDecoration.Parameter.CONTENT), net.minecraft.network.chat.Style.EMPTY), new net.minecraft.network.chat.ChatTypeDecoration("%s", java.util.List.of(net.minecraft.network.chat.ChatTypeDecoration.Parameter.CONTENT), net.minecraft.network.chat.Style.EMPTY));
```
Owen1212055: Paper comment? Optimally, we should just add this ourselves using the paperweight PR you have that allows us to add to the vanilla datapack.

patches/server/0546-Registry-Modification-API.patch
```diff
@@ -0,0 +1,1281 @@
...
++    }
++    // Paper end
+ 
+     public MappedRegistry(ResourceKey<? extends Registry<T>> key, Lifecycle lifecycle) {
+         this(key, lifecycle, false);
+@@ -166,6 +174,7 @@ public class MappedRegistry<T> implements WritableRegistry<T> {
+         }
+ 
+         this.lifecycles.put(object, lifecycle);
++        Objects.requireNonNull(this.beforeFrozen).put(resourceKey.location(), object); // Paper
```
Owen1212055: Is this require non null really needed here? As I don't think this should even occur if it's false due to frozen being true?

patches/server/0546-Registry-Modification-API.patch
```diff
@@ -0,0 +1,1281 @@
...
+--- a/src/main/java/net/minecraft/core/MappedRegistry.java
++++ b/src/main/java/net/minecraft/core/MappedRegistry.java
+@@ -80,6 +80,14 @@ public class MappedRegistry<T> implements WritableRegistry<T> {
+             return MappedRegistry.this.getTags().map(Pair::getSecond);
+         }
+     };
++    // Paper start
++    @Nullable
++    private Map<ResourceLocation, T> beforeFrozen = new HashMap<>(2048);
++    public Map<ResourceLocation, T> beforeFrozenView() {
```
Owen1212055: Additionally, why do we need this before frozen map? Can't we just transform the this.byLocation map to unwrap the holder?

patches/api/0246-Registry-Modification-API.patch
Machine-Maker: Yeah, we can't use adventures. I already tried it. You can't change the classloader, and the classloader it uses will fail for tests, you need the thread context classloader for it to load the service for tests

Machine-Maker: > are we sure we want this as a publicly exposed field instead of an access method?

yeah, idc either way. It just can't be on the Bukkit singleton, or well I guess it can, just not part of the Server interface.

patches/server/0546-Registry-Modification-API.patch
Machine-Maker: You can't unwrap the holders, they don't have their values set yet, they get set when the registry is frozen.

Owen1212055: Ah wow, I see here.

electronicboy (approved):
approving, a few comments from others but it generally looks to be semantics, (and I really just wanna get that method change in..)

Browsit:
Just pointing out https://github.com/PaperMC/Paper/pull/10154 in case the TODO item gets overlooked

patches/api/0479-Registry-Modification-API.patch
```diff
@@ -0,0 +1,407 @@
...
++     * @return a registry view
++     */
++    @NotNull RegistryView<T> registry();
++
++    /**
++     * Get the name of the event.
++     *
++     * @return the event name
++     */
++    default @NotNull String eventName() {
```
kennytv: I'd remove this before someone starts relying on internal class names, don't really see much use for it outside of debugging

patches/server/1051-Registry-Modification-API.patch
```diff
@@ -0,0 +1,966 @@
...
++        final ResourceKey<M> key,
++        final @Nullable M oldNms,
++        final B builder,
++        final RegistrationInfo registrationInfo,
++        final RegisterMethod<M, R> registerMethod
++    ) {
++        @Subst("namespace:key") final ResourceLocation beingAdded = key.location();
++        @SuppressWarnings("PatternValidation") final TypedKey<T> typedKey = TypedKey.create(entry.apiKey(), Key.key(beingAdded.getNamespace(), beingAdded.getPath()));
++        final RegistryAdditionEventImpl<T, B> event = entry.createAdditionEvent(typedKey, builder, registryView);
++        LifecycleEventRunner.INSTANCE.callEvent(this.additionHooks.getHook(entry.apiKey()), event);
```
56738: After removing the addition handler from the test plugin (only listening to preFreeze), an NPE is thrown because `getHook` returns null.

```
[LifecycleEventRunner] Could not run 'RegistryKeyImpl[key=minecraft:game_event] / pre-freeze' lifecycle event handler from Paper-Test-Plugin v1.0.0-SNAPSHOT
java.lang.NullPointerException: Cannot invoke "io.papermc.paper.plugin.lifecycle.event.types.AbstractLifecycleEventType.forEachHandler(java.util.function.Consumer, java.util.function.Predicate)" because "lifecycleEventType" is null
	at io.papermc.paper.plugin.lifecycle.event.LifecycleEventRunner.callEvent(LifecycleEventRunner.java:63) ~[main/:?]
	at io.papermc.paper.plugin.lifecycle.event.LifecycleEventRunner.callEvent(LifecycleEventRunner.java:58) ~[main/:?]
	at io.papermc.paper.registry.PaperRegistryListenerManager.registerWithListeners(PaperRegistryListenerManager.java:127) ~[main/:?]
	at io.papermc.paper.registry.PaperRegistryListenerManager.registerWithListeners(PaperRegistryListenerManager.java:111) ~[main/:?]
	at io.papermc.paper.registry.WritableCraftRegistry.register(WritableCraftRegistry.java:46) ~[main/:?]
	at io.papermc.paper.registry.event.PaperWritableRegistry.register(PaperWritableRegistry.java:21) ~[main/:?]
	at test-plugin-1.0.0-SNAPSHOT.jar/io.papermc.testplugin.TestPluginBootstrap.lambda$bootstrap$1(TestPluginBootstrap.java:32) ~[test-plugin-1.0.0-SNAPSHOT.jar:?]
```

Machine-Maker: Nice catch, fixed.

kennytv (approved):

Amejonah1200:
🥳❔

lynxplay:
Merged into preliminary 1.21 registry dev branch

PR #8936, opened 2023-03-04
Properly handle BlockBreakEvent#isDropItems

### LINKED ISSUE #8931: Statistic Statistic.MINE_BLOCK is not incremented when drops are turned off via BlockBreakEvent#setDropItems(false)
### Expected behavior

The MINE_BLOCK statistic should be incremented regardless of whether it drops items or not

### Observed/Actual behavior

The statistic is not incremented when drops are turned off for the block.

### Steps/models to reproduce

I created a simple plugin that turns off drops on BlockBreakEvents and messages the player their current MINE_BLOCK statistic

### Plugin and Datapack List

Plugins: Only the custom made one described as above to test the behaviour
Datapacks: vanilla, bukkit

### Paper version

This server is running Paper version git-Paper-431 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: e574412)
Previous version: git-Paper-327 (MC: 1.19.3)

### Other

The statistic is also not incremented in creative mode (where also blocks do not drop items upon breaking them). Not sure if this is intended behaviour or also subject to be patched.

Machine-Maker:
Setting whether a block break dropped items controlled far more than just whether blocks dropped, like stat increases food consumption, turtle egg count decreases, ice to water conversions and beehive releases

I put an `@DoNotUse` annotation on the old method which will catch any other calls (including super.playerDestroy calls added in future versions).

Fixes https://github.com/PaperMC/Paper/issues/8931

### DIFF
--- a/patches/server/1006-Properly-handle-BlockBreakEvent-isDropItems.patch
+++ b/patches/server/1006-Properly-handle-BlockBreakEvent-isDropItems.patch
@@ -1,0 +1,165 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 4 Mar 2023 10:52:52 -0800
+Subject: [PATCH] Properly handle BlockBreakEvent#isDropItems
+
+Setting whether a block break dropped items controlled
+far more than just whether blocks dropped, like stat increases
+food consumption, turtle egg count decreases, ice to water
+conversions and beehive releases
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
+index baf3e79489e310f443788bc917c553ae7ea86c89..9a17193c9fe26c9c0390906b4c1f9d619ccba249 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
+@@ -438,8 +438,8 @@ public class ServerPlayerGameMode {
+                     isCorrectTool = flag1; // Paper
+ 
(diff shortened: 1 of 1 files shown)


patches/server/0964-Properly-handle-BlockBreakEvent-isDropItems.patch
```diff
@@ -0,0 +1,106 @@
...
++    @io.papermc.paper.annotation.DoNotUse // Paper - method below allows better control of item drops
+     public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack stack) {
++        // Paper start
++        this.playerDestroy(world, player, pos, state, blockEntity, stack, true);
++    }
++    public void playerDestroy(Level world, Player player, BlockPos pos, BlockState state, @Nullable BlockEntity blockEntity, ItemStack stack, boolean includeDrops) {
++        // Paper end
+         player.awardStat(Stats.BLOCK_MINED.get(this));
+         player.causeFoodExhaustion(0.005F, org.bukkit.event.entity.EntityExhaustionEvent.ExhaustionReason.BLOCK_MINED); // CraftBukkit - EntityExhaustionEvent
++        if (includeDrops) // Paper
```
lynxplay: Wouldn't we want to include the call to `Block.dropResources` in the diff here to ensure failure on change ?
If statement without brackets just seems fun :sweat_smile:

Machine-Maker: the diff is close enough to conflict. If the line below changes, the diff will conflict because when it inserts the line above it, it will expect the line below it to be what it is now.

lynxplay: Okay yea, wasn't sure specifically when it breaks :+1:

lynxplay (approved):

kennytv (approved):

PR #8420, opened 2022-10-01
Fix MC-123848

BillyGalbreath:

PR #8940, opened 2023-03-05
fix 2 vanilla issues

Machine-Maker:
Fix sheeps eating grass outside ticking chunks
Fix itemframe drops dropping in correct location

Replacement for https://github.com/PaperMC/Paper/pull/8420 and https://github.com/PaperMC/Paper/pull/8424

### DIFF
--- a/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
@@ -40,6 +40,14 @@ https://bugs.mojang.com/browse/MC-224454
 https://bugs.mojang.com/browse/MC-263999
   Fix mobs breaking doors not spawning block break particles
 
+https://bugs.mojang.com/browse/MC-210802
+  by: BillyGalbreath <Blake.Galbreath@Gmail.com>
+  Fixes sheep eating blocks outside of ticking range
+
+https://bugs.mojang.com/browse/MC-123848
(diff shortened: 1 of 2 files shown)


Owen1212055 (approved):

patches/server/0899-Fix-a-bunch-of-vanilla-bugs.patch
```diff
@@ -44,6 +44,14 @@ https://bugs.mojang.com/browse/MC-258535
   by: Jake Potrebic <jake.m.potrebic@gmail.com>
   Fixes certain explosion damage not scaling with difficulty
 
+https://bugs.mojang.com/browse/MC-210802
+  by: BillyGalbreath <Blake.Galbreath@Gmail.com>
```
Owen1212055: Also, diff capitalization on emails?

Machine-Maker: Just copied them right from the original patch files.

lynxplay:
For documentation purposes, `shouldTickBlocksAt` does not work here, spawn chunks, even if not random ticked, remain at FullStatus ENTITY_TICKING and hence `shouldTickBlocksAt` returns true for blocks that are not random ticked.

Machine-Maker:
Ok, changed the check in EatBlockGoal to check if any player was near enough for spawning, the same check used for random ticking in ServerChunkCache

PR #8941, opened 2023-03-05
Add PlayerTeleportEndExitEvent

Doc94:
Based in a conversation in discord i think currently not exists a good form to cancel the teleport of player when going to the end portal in the END to go to overworld like other portals, exists methods but only allow change the destination and not cancel before all happen.

The PR add a new method extented from PlayerTeleportEvent for handle this thing and allow cancel.

This PR has a things im not pretty sure.

- Thanks to mojang this method not allow change the destination because the destination in first place is the respawn position, and this is handled in another place when the packet for respawn is send because yes this send a packet for make all the respawn logic then mess with that is... well "hard"

### DIFF
--- a/patches/api/0419-Add-PlayerTeleportEndExitEvent.patch
+++ b/patches/api/0419-Add-PlayerTeleportEndExitEvent.patch
@@ -1,0 +1,38 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 5 Mar 2023 15:07:13 -0300
+Subject: [PATCH] Add PlayerTeleportEndExitEvent
+
+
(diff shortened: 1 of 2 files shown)


Machine-Maker:
There are already ways to cancel the teleport before it even happens. Using `EntityInsideBlockEvent` and checking for end portals as being the block the player is in. That even is cancellable. If that isn't specific enough, a better solution is to make `EntityPortalEnterEvent` cancellable (but I don't think its really needed as the other event should cover everything already).

Also I don't think its a good idea to add yet another teleport event for something as specific as this.

Doc94:
> There are already ways to cancel the teleport before it even happens. Using `EntityInsideBlockEvent` and checking for end portals as being the block the player is in. That even is cancellable. If that isn't specific enough, a better solution is to make `EntityPortalEnterEvent` cancellable (but I don't think its really needed as the other event should cover everything already).
> 
> Also I don't think its a good idea to add yet another teleport event for something as specific as this.

yeah i prefer wait if is necesary allow cancel the portalevent, then im closing this.

PR #8942, opened 2023-03-05
Allow entity effect modification off the main thread for worldgen

### LINKED ISSUE #8381: Structure loading adding entity effects async
### Stack trace

Hello
I have recently been running a Server and after a few days, when people we're exploring a little faster, the server crashed multiple times when a person was at a certain chunk and would continue to crash the server when logging in. I could fix this with teleporting them out which would fix it most of the time. I didnt really know where else to post this but here and I would like to know what could cause this. Thanks in advance.

Pastebin: https://pastebin.com/Vp2AmSKz

### Plugin and Datapack List

Plugins:
Chunky
Slimefun (Including Addons)

Datapacks:
Dungeons Now Loading
Incendium
Terralith

_No response_

### Paper version

Version: git-Paper-142 (Git ef0e5a6)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/8381

There are probably more options for a check to only conditionally run the AsyncCatcher.

Another option I came up with was checking if the `chunkStatus` on `Entity` was null, as that means it hasn't been added to the ChunkEntitySlices yet.

The `valid` field is set to true when the entity has had its "start tracking" callback called.

Either way, there is a completely valid use for adding effects off the main thread... when the entity object only exists on a separate thread like it can during worldgen.


Machine-Maker:
I did just have an idea for an alternative solution, remove the async catcher from the nms method, and just have it directly on the CraftLivingEntity impl method, so only the API can't do it "not on the main thread".

Owen1212055:
I agree with your alternate impl here

Owen1212055 (approved):
Approved, but I think it may be better to go with your alternate implementation of instead doing the check on the craft entity instance.

Machine-Maker:
Switched to commenting out spigot's check and adding a separate one in the CraftLivingEntity impl

Owen1212055 (approved):

electronicboy (approved):

PR #8944, opened 2023-03-05
Correctly handle ArmorStand invisibility

### LINKED ISSUE #8943: Armor stand is invulnerable when set invisible
### Expected behavior

An armor stand to be destroyable regardless of gamemode when it is invisible, but only destroyable in creative mode when it is invulnerable.

### Observed/Actual behavior

Didn't test other cases than these two, but
1. When set invisible, the armor stand is immediately invulnerable to being destroyed, regardless of the player's gamemode
2. When set invisible AND invulnerable, the armor stand is only invulnerable to being destroyed after a server restart (before the restart, the stand can be destroyed in creative mode as expected)

### Steps/models to reproduce

1. Spawn an armor stand and call `setInvisible(true)` -> observe armor stand to be unbreakable in all gamemodes
2. Spawn an armor stand, call `setInvisible(true)` and `setInvulnerable(true)` -> observe armor stand to be unbreakable in survival, but breakable in creative mode
3. Spawn the armor stand as in step 2 and restart the server -> observe armor stand to be unbreakable in all gamemodes

### Plugin and Datapack List

Test plugin to spawn armor stands as described in the steps above:

```java
@Override
public void onEnable() {
        World world = getServer().getWorld("world");
        ArmorStand invisible = world.spawn(new Location(world, 0, 120, 0), ArmorStand.class, CreatureSpawnEvent.SpawnReason.CUSTOM);
        invisible.setAI(false);
        invisible.setGravity(false);
        invisible.setInvisible(true);

        ArmorStand invulnerable = world.spawn(new Location(world, 2, 120, 0), ArmorStand.class, CreatureSpawnEvent.SpawnReason.CUSTOM);
        invulnerable.setAI(false);
        invulnerable.setGravity(false);
        invulnerable.setInvulnerable(true);

        ArmorStand both = world.spawn(new Location(world, 4, 120, 0), ArmorStand.class, CreatureSpawnEvent.SpawnReason.CUSTOM);
        both.setAI(false);
        both.setGravity(false);
        both.setInvisible(true);
        both.setInvulnerable(true);
    }, 20 * 5L);
}
```

### Paper version

This server is running Paper version git-Paper-442 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 5cc78f2)

### Other

 If possible, in the future, I'd like there to be an option (eg. `setUnbreakable(boolean)`) to set whether the armor stand can be broken at all, regardless of gamemode (aka. only with `/kill`).

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/8943

ArmorStand has a separate field that tracks invisibility which wasn't getting set from the api calls to setInvisible()

### DIFF
--- a/patches/server/0967-Correctly-handle-ArmorStand-invisibility.patch
+++ b/patches/server/0967-Correctly-handle-ArmorStand-invisibility.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 5 Mar 2023 14:38:21 -0800
+Subject: [PATCH] Correctly handle ArmorStand invisibility
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftArmorStand.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftArmorStand.java
+index f80cafe3544c7e6c3c29073ba6539783adf6666c..17d08d2c90047ebfc4098b087be091b83111a40f 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftArmorStand.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftArmorStand.java
+@@ -158,6 +158,14 @@ public class CraftArmorStand extends CraftLivingEntity implements ArmorStand {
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
Kinda weird that the craft method doesn't just call this method instead, but, looks good. (Since it seems only armorstand overrides this anyways)

kashike (approved):

PR #8945, opened 2023-03-06
Fix spawners checking max nearby entities with correct type

Machine-Maker:
Fixes [MC-259321](https://bugs.mojang.com/browse/MC-259321)

Previously, spawners would take into account type inheritance when checking for nearby mobs. For example, a regular spider spawner wouldn't spawn mobs if 6 cave spiders were nearby. Similar issues arose with mobs that have different mobs subclass from them. This PR changes it to be a direct check on the type of mob.

### DIFF
--- a/patches/server/0856-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0856-Fix-a-bunch-of-vanilla-bugs.patch
@@ -395,6 +395,42 @@ index ca6a2b9840c9ade87ec8effab01d4f184fe876b7..43129ecefcc8beccbcf2978f262b1ce8
                      entity.level().gameEvent(entity, GameEvent.EQUIP, entity.position());
                      stack.shrink(1);
                  }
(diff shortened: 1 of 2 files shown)


Owen1212055:
I’m a little iffy about this one, could it be argued that because of how long this has been around that it could be considered expected behavior? 

Cause at least, now you’d have twice or even more the amount of mobs in some cases.

Machine-Maker:
> I’m a little iffy about this one, could it be argued that because of how long this has been around that it could be considered expected behavior?

Mojang gave it a priority and a confirmed label. It's very unpredictable, and not super common really. The only natural case I can think of is if you somehow get a spider spawner near a cave spider spawner, and that's pretty rare. You don't generally have husks around a zombie spawner or strays around a skeleton spawner.

Owen1212055:
Fixed by mojang, waiting to see mojang's fix to pull the fix.

Owen1212055 (approved):

PR #8952, opened 2023-03-09
Add FeatureFlag API

### LINKED ISSUE #8950: API to check if a material is disabled

Minecraft 1.19.3 added Feature Flag system, which allows some items to be disabled. I think there should be a way for server plugins to check if a material is disabled.

An API to check if a bukkit Material is disabled or not in the specified world would be helpful. It needs a world since worlds can have different datapacks.

I'm currently using NMS codes to find out if a block is disabled or not (`net.minecraft.world.level.block.Block#isEnabled`)

### Other

_No response_

Machine-Maker:
Closes https://github.com/PaperMC/Paper/issues/8950

My initial thoughts on API for checking if a Material or EntityType is enabled with a `FeatureDependant` interface.

Adds tests to check on updates when this changes.

### DIFF
--- a/patches/api/0270-Add-basic-Datapack-API.patch
+++ b/patches/api/0270-Add-basic-Datapack-API.patch
@@ -70,7 +70,7 @@ index 0000000000000000000000000000000000000000..58f78d5e91beacaf710f62461cf869f7
 +
 +}
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index b558fa73dbcf3747690933e6aadf7061a0de2630..8a22c242a93b0e16e0bca583d0918bab695248b1 100644
+index b558fa73dbcf3747690933e6aadf7061a0de2630..be68351555bde59a4e55bf1bad261e9f6bc9f704 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
 @@ -328,9 +328,11 @@ public final class Bukkit {
@@ -81,7 +81,7 @@ index b558fa73dbcf3747690933e6aadf7061a0de2630..8a22c242a93b0e16e0bca583d0918bab
       * @return the manager
       */
      @NotNull
-+    @Deprecated(forRemoval = true)
++    @Deprecated(forRemoval = true, since = "1.20")
      public static DataPackManager getDataPackManager() {
          return server.getDataPackManager();
      }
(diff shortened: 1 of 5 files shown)


maestro-denery:
How about using `@Deprecated` and `@APIStatus.ScheduledForRemoval()` for feature flags enabling updates? Also I think there should be a list of experimental feature flags available (all feature flags excluding the vanilla one) so people won't have to explicitly use fields which will be removed later to check whether a Material / EntityType is experimental or not.

Machine-Maker:
> How about using @Deprecated and @APIStatus.ScheduledForRemoval() for feature flags enabling updates?

Deprecated seems inappropriate for this, and scheduledforremoval might work for UPDATE_1_20, but not bundle as we have no clue when that's gonna be removed. I think Experimental is the most accurate marker annotation.

patches/server/0968-Add-FeatureFlag-API.patch
```diff
@@ -0,0 +1,138 @@
...
++}
+diff --git a/src/main/java/net/minecraft/server/Main.java b/src/main/java/net/minecraft/server/Main.java
+index 781b72fd88149642c9fceaecfbfe7546273fb749..25726d007d8f0875bf86faa835717017edda0415 100644
+--- a/src/main/java/net/minecraft/server/Main.java
++++ b/src/main/java/net/minecraft/server/Main.java
+@@ -111,6 +111,7 @@ public class Main {
+                 JvmProfiler.INSTANCE.start(Environment.SERVER);
+             }
+ 
++            io.papermc.paper.world.flag.PaperFeatureFlagProviderImpl.INSTANCE.getClass(); // Paper - init FeatureFlagProvider
```
lynxplay: Is this the preferred way we want to inject server singletons into the API ?
I know this is a somewhat long standing issue between services, the server interface and this one.

kennytv (approved):

Owen1212055 (commented):
Generally looks good, but think we should wait for the enum PR to allow us to have the "isEnabled" method be implemented manually to prevent the need for instance for the "FeatureElement" conversion thing.

qixils:
Sorry, what exactly is blocking this? I don't see a "enum PR" from a glance, unless you mean the property API tagged "for: future" which I quite hope not... This is preventing me from being able to determine what items/entities are spawnable which is quite unfortunate for my use cases. I worked around it in the meantime with NMS but obviously this is not an ideal solution for cross-version compatibility among other issues. I'm hoping this could just be merged now (ideally to 1.19.4 while it's still supported ❤️) and improved later since it "generally looks good"?

Machine-Maker:
The enum PR was the PR previous thought to be slated by upstream for 1.20. It replaces a ton of the enums and deprecates Material.

qixils:
Ah, upstream, I can see the hesitation then. Unfortunate but I can live with NMS for the moment

RedstoneFuture:
Would it be an option to extend the API with a method that outputs a permitted `EntityType[]` and `Material[]` for the corresponding feature-flag?

This could be helpful e.g. for tab-completions in which only the enabled blocks are to be displayed. Otherwise, you have to check all game objects individually each time.

Machine-Maker:
Ok, revamped this and took over as much as possible from upstream's feature flag API. Added tests to ensure nothing is missed as well.

lynxplay (approved):

PR #8953, opened 2023-03-10
NFC: fix indentation

brickmonster:
Fix patch indentation that has become out of sync from the surrounding code over time.
No functional changes, just less confusing.

### DIFF
--- a/patches/server/0325-Prevent-consuming-the-wrong-itemstack.patch
+++ b/patches/server/0325-Prevent-consuming-the-wrong-itemstack.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Prevent consuming the wrong itemstack
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
-index 69fbbd2745008e2d9caf6a30dd0779339e1c685b..5eaedecd2e53bcdc30defbf56d6a70a0c49f03ec 100644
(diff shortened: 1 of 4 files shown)


patches/server/0325-Prevent-consuming-the-wrong-itemstack.patch
```diff
@@ -28,18 +28,18 @@ index cb53fc775df387e8bbb099517793daea5ea15e4b..ba4e0bac019b68da166b277028015999
...
                      this.stopUsingItem();
--                // Paper start - if the replacement is anything but the default, update the client inventory
--                if (this instanceof ServerPlayer && !com.google.common.base.Objects.equal(defaultReplacement, itemstack)) {
-+                // Paper start
-+                if (this instanceof ServerPlayer) {
-                     ((ServerPlayer) this).getBukkitEntity().updateInventory();
-                 }
-                 // Paper end
+-                    // Paper start - if the replacement is anything but the default, update the client inventory
+-                    if (this instanceof ServerPlayer && !com.google.common.base.Objects.equal(defaultReplacement, itemstack)) {
```
brickmonster, line 40: Is the original patch correct here? It seems to undo something intentional from an earlier patch.

brickmonster: Patch 0072 added:
` && !com.google.common.base.Objects.equal(defaultReplacement, itemstack)`
and patch 325 reverts it.

I can fix it in a new pull request or change this request to include the fix if desired.

brickmonster:
Rebased

MiniDigger:
fyi, this indentation is on purpose to reduce diff

PR #8955, opened 2023-03-10
Fix flat bedrock setting

Machine-Maker:
So I decided to go ahead and do this anyway, if only to demonstrate that this should NOT be the fix we decide to go with. This fix does work, and will only affect noise settings files for the `overworld` and `nether` from the built-in datapack, but it'd be much easier to change **1** line in each of the 5 json files inside the jar instead.


Machine-Maker:
Requires https://github.com/PaperMC/paperweight/pull/190 to be properly fixed

PR #8956, opened 2023-03-11
Upgrade to snakeyaml 2.0

yannicklamprecht:
- Probably better for 1.19.4 or 1.20


lynxplay:
Closing this for now. While this upgrade certainly is useful to prevent insecure deserialization, we cannot bump an api exposed dependency if it includes breaking changes for plugins that might depend on the spigot-api and hence develop against potentially now removed methods present in 1.33.

PR #8958, opened 2023-03-12
[ci skip] add workflow for auto-managing issue board

Machine-Maker:
This adds a workflow which will run on any label changes to issues or prs and update the associated projects accordingly. The new projects are useful for categorizing things, but they are missing some automation, such as auto-moving between statuses (columns) when labels are changed.

Right now, this only updates the Issues board ([found here](https://github.com/orgs/PaperMC/projects/5/views/2)). Other boards can be added once this is tested on the org project (I only tested it on user projects, but it should be the same).

~~This will require a new secret to be added named `PROJECT_ACTIONS_TOKEN` as the default access token available in actions does not have the required permissions to change user/org projects.~~
Changed it to generate the token in the action using an org-only github app.

The `label-to-column-map` can be changed per any other suggestions, it's just what I came up with after a quick look through of all the issue labels I thought mattered.

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -8,13 +8,12 @@ on:
       - closed
       - reopened
 
-concurrency:
-  group: update-projects-${{ github.event_name }}
-  cancel-in-progress: true
-
 jobs:
   bugs:
     if: "github.event_name == 'issues' && contains(github.event.*.labels.*.name, 'type: bug')"
+    concurrency:
+      group: update-bugs-project-${{ github.event.issue.number }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Machine-Maker:
This should now handle pretty much any combination of actions in any sequence that you can do on GitHub. Labeling and closing, closing then labeling, re-opening without labels, re-opening with labels. The concurrency systems works to make sure only the latest state is used for computing where an issue/pr should go.

lynxplay (approved):
Tested on https://github.com/lynxplay-edu/Paper/issues/1, requires a normal access token as github fine grained access tokens do not seem to be usable with graphql yet.

kennytv (approved):

PR #8995, opened 2023-03-17
Fix entity damage event triggers

Machine-Maker:
Fixes a couple places where the original damage and modified damage were passed in the reverse order to the advancement triggers.

Thanks to @Lulu13022002 for pointing this out.


patches/server/0962-Fix-advancement-triggers-for-entity-damage.patch
```diff
@@ -0,0 +1,46 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Nassim Jahnke <nassim@njahnke.dev>
```
kennytv: You can change the author if you want

kennytv (approved):

Issue #8982, opened 2023-03-16
Ability to hide tile entity content


masmc05:
### Is your feature request related to a problem?

Yes, we want to improve loot tables of vanilla structures, but because player receives full information about the chest, it can see loot from far away, making it very easy to seek good loot and ignore already looted chests

### Describe the solution you'd like.

![image](https://user-images.githubusercontent.com/63639746/225608016-a4d8c8f7-3990-45ea-850e-1b41b5587732.png)
this is a method in BlockEntity, and it can be used to easily add such functionality, hiding vanilla loot table and the items from chest, maybe other data too from other blocks

### Describe alternatives you've considered.

Create a bukkit snapshot of the block using that nbt, and run an event, and allow plugins to modify the data that players will see

MiniDigger:
the patch in question: https://github.com/PaperMC/Paper/blob/master/patches/server/0847-Sanitize-Sent-BlockEntity-NBT.patch#L41
should prolly use a config option under the obfuscation section, just like for item meta

Owen1212055:
Wait a minute... when is this actually the case? Container entities don't send their loot tables over the network under any circumstances.

masmc05:
oh, feels like it, even checked with a mixin that client didn't receive it, sorry that didn't check that before that
```java
@Mixin(ClientboundLevelChunkPacketData.class)
public class PacketMixin {
    @Shadow @Final
    public List<ClientboundLevelChunkPacketData.BlockEntityInfo> blockEntitiesData;

    @Inject(method = "getBlockEntitiesTags", at = @At("HEAD"))
    public void handle(ClientboundLevelChunkPacketData.BlockEntityTagOutput blockEntityTagOutput, int i, int j, CallbackInfo ci) {
        for (var blockEntityInfo : this.blockEntitiesData) {
            if (blockEntityInfo.type == BlockEntityType.CHEST) {
                var nbt = Optional.ofNullable(blockEntityInfo.tag).map(Tag::getAsString).orElse("null");
                Minecraft.getInstance().gui.getChat().addMessage(Component.literal(nbt));
            }
        }

    }
}
```
seems like every check always returns null

PR #8998, opened 2023-03-18
fix for #8982

GOjayson:
If it is totally wrong please let me know, i'm new and trying to help.

### DIFF
--- a/patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
+++ b/patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
@@ -29,6 +29,7 @@ index f2d1ef0e02524544352c757da406d3f7224961d3..d573ee9dc67370d76d0ab63906d03c7d
 +import com.destroystokyo.paper.PaperConfig;
  import com.mojang.logging.LogUtils;
  import javax.annotation.Nullable;
++import java.awt.print.Paper;
 +
 +import io.papermc.paper.configuration.PaperConfigurations;
  import net.minecraft.CrashReportCategory;
@@ -47,8 +48,6 @@ index f2d1ef0e02524544352c757da406d3f7224961d3..d573ee9dc67370d76d0ab63906d03c7d
  import co.aikar.timings.MinecraftTimings; // Paper
  import co.aikar.timings.Timing; // Paper
  
-+import java.awt.print.Paper;
-+
  public abstract class BlockEntity {
  
      public Timing tickTimer = MinecraftTimings.getTileEntityTimings(this); // Paper
--- a/patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
+++ b/patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
@@ -1,0 +1,85 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jayson <jayson2001@hotmail.be>
+Date: Sun, 19 Mar 2023 12:44:43 +0100
+Subject: [PATCH] Added config to Hide loottables and PDC
+
+
+diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+index 4532f3a0d74feae0a1249b53e1bfbc18a8808b32..bcdcc4b411fc794760589a6e0b8791801daf2c1c 100644
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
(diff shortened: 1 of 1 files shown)


lynxplay (changes requested):
Welcome to your first paper PR! 

As mini mentioned, this should most likely be behind a configuration option.
We also definitely still want to remove the PublicBukkitValues from the tag before sending.

To add configuration to paper (see minis comment on the issue) you want to be editing https://github.com/PaperMC/Paper/blob/master/patches/server/0005-Paper-config-files.patch. I'd argue we would want boolean toggles for PDC and LootTables separately.

GOjayson:
Okay thank you for the directions!

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
```diff
@@ -0,0 +1,85 @@
...
++        // Check if LootTable visibility is enabled and remove LootTable and LootTableSeed tags if necessary
++        boolean hideLootTables = this.level.paperConfig().entities.hideLootTables;
++        if (hideLootTables && tag.contains("LootTable")) {
+             tag.remove("LootTable");
+             tag.remove("LootTableSeed");
+         }
++
++        // Check if PDC functionality is enabled and remove PDC-related tags if necessary
++        boolean hidePdc = this.level.paperConfig().entities.hidePDC;
++        if (hidePdc && tag.contains("PlayerData")) {
```
Lulu13022002, line 72: This is the wrong tag for the pdc, idk from what this come but the pdc tag is the "PublicBukkitValues" even if the name is not really good.

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
```diff
@@ -0,0 +1,85 @@
...
++import org.bukkit.Bukkit;
++import org.bukkit.craftbukkit.CraftServer;
+ import org.slf4j.Logger;
+ 
+ // CraftBukkit start
+@@ -25,6 +30,8 @@ import org.spigotmc.CustomTimingsHandler; // Spigot
+ import co.aikar.timings.MinecraftTimings; // Paper
+ import co.aikar.timings.Timing; // Paper
+ 
++import java.awt.print.Paper;
```
Lulu13022002: Clean your import a little bit to reduce the diff

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
```diff
@@ -0,0 +1,85 @@
...
+diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+index 4532f3a0d74feae0a1249b53e1bfbc18a8808b32..bcdcc4b411fc794760589a6e0b8791801daf2c1c 100644
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+@@ -100,6 +100,10 @@ public class WorldConfiguration extends ConfigurationPart {
+ 
+     public class Entities extends ConfigurationPart {
+         public boolean entitiesTargetWithFollowRange = false;
++
++        public boolean hideLootTables = false;
```
Lulu13022002, line 16: I don't think this is the right config section. should go below the anticheat.obfuscation.tile-entities/block-entities or somethings like that.

GOjayson:
Hmm, I might have done the patching wrong.

Lulu13022002 (commented):
It seems that you have now two patches, when you add a new change (to an already created patch) you need to amend the commit (while the rebase) to avoid creating a duplicate patch but for now you can just drop the first one

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
GOjayson: Ohw wow, yea found it a bit confusing. Thanks.

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
GOjayson: Will clean up, Thanks

patches/server/0967-Added-config-to-Hide-loottables-and-PDC.patch
GOjayson: Yea i've just put them somewhere in the config, i'm not farmiliar yet with all config things. Thank you for correcting me, will change it.

GOjayson:
Okay, finallly. I got a headache trying not to push the wrong things. I hope this is better. 

I don't know if the Configuration part is fully right now. I've put it into Anticheat and added the bools to an enity configurationpart.

GOjayson:
Ok should all be fixed now.

GOjayson:
No... nvm. I don't know what happend exactly but my branch is a mess. I'm going to close this PR and open a new one.

PR #9002, opened 2023-03-18
Tag Modification API

Machine-Maker:
You can create `TagUpdate`s on the Registry instance, which you can then send to a single player or apply to the whole server to affect all players.

```java
// createTagUpdate will throw UOE on all non-CraftRegistry impls
TagUpdate<?> update = Registry.STRUCTURE.createTagUpdate(map -> { 
    // map is a mutable Map<Key, Set<Structure>>
    map.put(Key.key("eye_of_ender_located"), Collections.emptySet()); // will prevent eye of enders from finding strongholds
});

Bukkit.getServer().applyTagUpdate(update);
```

```[tasklist]
- [ ] https://github.com/PaperMC/Paper/pull/8920
```


Machine-Maker:
Added a method to edit a single registry's tags directly. Also added improved javadocs listing the 2 registries this works with and that its best to group changes to multiple registries.

PR #9005, opened 2023-03-19
Check any loaded plugins for dependencies

Blast-MC:
This change allows spigot plugins with dependencies to be loaded at runtime and should not affect paper plugins. I know plugin loading at runtime is not something you guys want to support, but this simple change shouldn't hurt your system while still allowing those who accept the risks to continue doing so.


Owen1212055:
Were dependencies not originally checked when loading plugins during runtime? 🤔

Additionally this may break bootstrappers due to them having dependency validation at this stage. Probably not good to introduce bukkit api here. (Since it may not even exist at certain points)

Blast-MC:
Before, it only checked if the plugin existed, not if it was enabled. This can do the same if wanted using `PaperPluginManagerImpl#getPlugin`.

Owen1212055:
Closing in favor of https://github.com/PaperMC/Paper/pull/9068
Please feel free to test your environments with the above PR.

PR #9006, opened 2023-03-19
Add server reload event for tracking when is the server reloading

TheMeinerLP:
Helps plugin developer to save data between reloads only

### DIFF
--- a/patches/api/0419-Add-server-reload-event-for-tracking-when-is-the-ser.patch
+++ b/patches/api/0419-Add-server-reload-event-for-tracking-when-is-the-ser.patch
@@ -1,0 +1,56 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheMeinerLP <p.glanz@madfix.me>
+Date: Sun, 19 Mar 2023 15:00:56 +0100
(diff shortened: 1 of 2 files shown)


molor:
Same as `ServerLoadEvent` w/ `LoadType.RELOAD`?

TheMeinerLP:
Not really. The ServerLoadEvent is called after. But this one is called before. To tigger a save logic inside of plugins for like databases or disk save

e-im:
Just `Server#isStopping` in `onDisable`?

Machine-Maker (changes requested):
I think this event is redundant. As it was pointed out, the onDisable method in JavaPlugin is called on reloads and Server#isStopping returns false if it's a reload. Is there some usecase that that fact and the ServerLoadEvent doesn't cover?

TheMeinerLP:
I have some cases that is not coverd. Also i miss a ServerStopEvent thats called before all Players kicked

emilyy-dev:
During shutdown, Plugin#onDisable is called before players are kicked, and players aren't kicked on reload anyway

Machine-Maker:
@TheMeinerLP can you be specific about what cases you believe aren't covered by existing events/onDisable? Otherwise I'm gonna close this as redundant.

PR #9009, opened 2023-03-19
Switch to adventure Key/Keyed for Registry

Machine-Maker:
I believe this is all we have to do, to migrate away from `org.bukkit.Keyed` when dealing with all things "Registry". The `T` generic in Registry changes to extend `net.kyori.adventure.key.Keyed`, which means just a single method has to be rewritten with Commodore for the new return type of `T get(NamespacedKey)`. Along with maintaining ABI compat, this shouldn't break any compiles either, as `org.bukkit.Keyed` extends `net.kyori.adventure.key.Keyed` and `NamespacedKey` extends `Key`.

Once this is finalized, most of this will go into the server/api adventure patches.

This isn't deprecating anything, like `NamespacedKey` or the other 500 methods that use `NamespacedKey`, it just means we won't have to keep using `NamespacedKey` when `Key` is much better.

I tested this by compiling a plugin against the non-modified API and made sure the Commodore changes work as expected.

### DIFF
--- a/patches/api/0418-Switch-to-adventure-Key-Keyed.patch
+++ b/patches/api/0418-Switch-to-adventure-Key-Keyed.patch
@@ -1,0 +1,135 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 19 Mar 2023 15:39:30 -0700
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Updated to remove the default method with `NamespacedKey`. I just realized that the parameter could also be rewritten to adventure `Key` since `NamespaceKey` implements `Key`. 


This PR should allow us to not have to extend bukkit Keyed for any registries that *we* add which is useful

PR #9021, opened 2023-03-21
Fix spigot ServerPlayer#teleportTo param forward

### LINKED ISSUE #9019: spreadplayers command doesnt spread players
### Expected behavior

If i use the /spreadplayers command in chat or with a command block it should teleport entities in a given range (for example '/spreadplayers ~ ~ 0 1000 false @a' should spread all player entities in a 1000 block radius)

### Observed/Actual behavior

The command teleports entities to location 0 0 0

### Steps/models to reproduce

using the command '/spreadplayers ~ ~ 0 1000 false @a' or with any other parameters results in the entities teleporting to 0 0 0 always

### Plugin and Datapack List

tested with and without plugins and datapacks results in the same behaviour every time

### Paper version

This server is running Paper version git-Paper-466 (MC: 1.19.4) (Implementing API version 1.19.4-R0.1-SNAPSHOT) (Git: 7af4cd3)
Previous version: git-Paper-448 (MC: 1.19.3)

### Other

i first went to the @paper-help channel on discord to confirm this issue with other people and user @pug#1070 also tested and reported the same issue

lynxplay:
During the spigot 1.19.4 update, the overload of
ServerPlayer#teleportTo which includes a TeleportCause is incorrectly called by its predecessor with the value 0 for x,y,z,yaw and pitch.

This causes mojang code still calling said methods to basically teleport the player to said coords.

Resolves: #9019

### DIFF
--- a/patches/server/0969-Fix-spigot-ServerPlayer-teleportTo-param-forward.patch
+++ b/patches/server/0969-Fix-spigot-ServerPlayer-teleportTo-param-forward.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bjarne Koll <lynxplay101@gmail.com>
+Date: Tue, 21 Mar 2023 15:25:50 +0100
+Subject: [PATCH] Fix spigot ServerPlayer#teleportTo param forward
+
+During the spigot 1.19.4 update, the overload of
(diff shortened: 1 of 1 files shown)


kennytv (approved):

lynxplay:
Fixed by upstream

lynxplay:
Fixed by upstream

PR #9024, opened 2023-03-21
New event verbose command

TheRealRyGuy:
Saw a message from Machine Maker on Discord bringing this up, figured it'd be a fun random thing to do

Essentially just adds a /paper dumpevents command. Running this once will begin logging events, and running it again will save the events in order with timestamp and event class fields in JSON to a file in the debug directory. Attached below is an example output

Two issues that aren't too big of a deal with this

Does not handle inheritance of event classes well, Field#getDeclaredFields only retrieves minimal fields, could use a superclass check working all the way up to org.bukkit.Event
Could further serialize non primitive event fields just for a bit more info, stringified versions of them are typically fine though

[event-verbose-2023-03-21_18.03.40.txt](https://github.com/PaperMC/Paper/files/11034393/event-verbose-2023-03-21_18.03.40.txt)


Machine-Maker:
Part of what I had described was logging to be able to see event class fields before and after each listener for an event. That way if an an event had 50 listeners, you could try and find what listener changed what.

TheRealRyGuy:
Most QoL way to do that would be with an EventExecutor and listen for a specific event - not sure if there's a way to register one internally though. My thought process was that this was just for all of the "what event runs when xyz" questions pop up. 

Will exclude tick start and end events, not sure what you mean by "changeable" values though.

Machine-Maker:
In the SimplePluginManager#fireEvent and PaperEventManager#callEvent, can't you just get the values of the event before and after each call?

"changeable" is hard to define. Generally I think events that have all final fields with no mutable things exposed don't need to be included. These events are just events used to inform that something happened, and a listener can't actually change anything about the event. The `VehicleUpdateEvent` is one of those. You can't change anythinga bout the event, so it doesn't need to be logged. Although I'm not sure there's a good way of knowing this automatically.

TheRealRyGuy:
Now ignores all final fields and can take specific exclusions, right now just tick starting and ending

Worth keeping this as its own system? Can just make a new command for watching specific events, can also just nuke this feature entirely for what you were more thinking of

Machine-Maker:
I would exclude final collections (Collection, Map) and ItemStack from the final check. It's possible to have a final one of those that can be mutated via a getter. 

I suppose both are useful. Seeing what events fire between two points, and a deeper look at what changes for each listener for those events. Might be best for the system I imagined to specify a single event to do that for.

TheRealRyGuy:
Added in the general premise for what you were looking for through `/paper watchevent [event]`, attached an example output.
Also moved the event verbose to paper's event manager to ignore bad event calls. The final check also now excludes `org.bukkit.inventory.ItemStack` as well as collections and maps 

[event-AsyncPlayerChatEvent-2023-03-21_20.43.03.txt](https://github.com/PaperMC/Paper/files/11035219/event-AsyncPlayerChatEvent-2023-03-21_20.43.03.txt)

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++import java.io.IOException;
++import java.io.PrintStream;
++import java.io.StringWriter;
++import java.lang.reflect.Field;
++import java.nio.charset.StandardCharsets;
++import java.nio.file.Files;
++import java.nio.file.Path;
++import java.time.LocalDateTime;
++import java.time.format.DateTimeFormatter;
++import java.util.*;
```
Machine-Maker: No `*` imports

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventVerboseCommand implements PaperSubcommand {
++
++    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH.mm.ss");
++    @Nullable private Path currentFile;
++    //have to be static to static call from paperpluginmanagerimpl
++    private static boolean isRunning = false;
++    private static final LinkedList<Wrapper> toDump = Lists.newLinkedList();
```
Machine-Maker: Is there a reason this is a LinkedList?

Also just use the `new ` constructor instead of guava's method.

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++    //ignore unnecessary events
++    private static final List<Class<? extends Event>> exclude = List.of(ServerTickStartEvent.class,
++        ServerTickEndEvent.class);
++
++
++    @Override
++    public boolean execute(CommandSender sender, String subCommand, String[] args) {
++        if(!isRunning) {
```
Machine-Maker: There's a lot of formatting oddities like the missing space between `if` and `(!isRunning)`

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++            isRunning = true;
++            Path parent = Path.of("debug");
++            Path path = parent.resolve("event-verbose-" + FORMATTER.format(LocalDateTime.now()) + ".txt");
++            try {
++                Files.createDirectories(parent);
++                currentFile = Files.createFile(path);
++                sender.sendMessage(text("Now storing event verbose, run this command again to save your file", GREEN));
```
Machine-Maker: `Now storing event verbose` doesn't make sense. Probably best to be more descriptive as to what it's actually doing.

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++         * Properly serialize our Wrapper into a JsonObject, so we can log it properly
++         * @implNote Is null if the event has only final fields
++         */
++        private JsonObject serialize() {
++            JsonObject parent = new JsonObject();
++            parent.addProperty("timestamp", timestamp);
++            JsonObject obj = new JsonObject();
++            obj.addProperty("event", event.getEventName());
++            try {
++                Field[] fields = event.getClass().getDeclaredFields();
```
Machine-Maker: This will ignore important fields like the Block or Entity in question of BlockEvent or EntityEvent since the field for that is stored on the superclass. You need to collect important fields from all the types in the hierarchy.

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++                //ignore update events / anything unchangeable except for ItemStacks, Collections and Maps
++                if(Stream.of(fields).allMatch(f -> Modifier.isFinal(f.getModifiers()) && !f.getType().equals(ItemStack.class) &&
++                    !Map.class.isAssignableFrom(f.getType()) && !Collection.class.isAssignableFrom(f.getType())))
++                    return null;
++                for (Field f : fields) {
++                    if(f.getName().equalsIgnoreCase("handlers"))
```
Machine-Maker: This is a bad way to ignore the handler list fields. You should just ignore "static" fields

patches/server/0969-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,204 @@
...
++                        continue;
++                    f.setAccessible(true);
++                    Object field = f.get(event);
++                    obj.addProperty(f.getName(), field == null ? "null" : field.toString());
++                }
++            } catch(SecurityException | IllegalAccessException ex) {
++                MinecraftServer.LOGGER.warn("Failure to serialize event verbose", ex);
```
Machine-Maker: Include the event name or object in the log.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++
++import static net.kyori.adventure.text.Component.text;
++import static net.kyori.adventure.text.format.NamedTextColor.GREEN;
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventWatcherCommand implements PaperSubcommand {
```
Machine-Maker: This is missing the tab complete logic. See `DumpListenersCommand` for an example of tab completion for event names.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++
++    public static void tryLog(Event event, @Nullable Listener listener) { //listener should only be null on the initial event call
++        Optional<Watcher> watcher = watching.stream().filter(w -> w.eventPath.equals(event.getClass().getSimpleName())).findFirst();
++        watcher.ifPresent(value -> value.log(event, listener));
++    }
++
++    @Override
++    public boolean execute(CommandSender sender, String subCommand, String[] args) {
++        if(args.length < 1) {
++            sender.sendMessage(text("Invalid Usage! /paper watchevent [event path]", RED));
```
Machine-Maker: `event path` is not the best name. Why not just `event name` as it's the simple class name of each event.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++            try {
++                w.dump();
++                sender.sendMessage(text("Successfully dumped event information to " + w.file, GREEN));
++            }catch(IOException ex) {
++                sender.sendMessage(text("Failed to dump event verbose! See the console for more info.", RED));
++                MinecraftServer.LOGGER.warn("Error occurred while creating event verbose files", ex);
++            }
++            watching.remove(w);
++        }else {
++            Path parent = Path.of("debug");
```
Machine-Maker: Make this a static final field instead of creating it each time as it's always the same. Same goes for the other command.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++            Path path = parent.resolve("event-" + args[0] + "-" + FORMATTER.format(LocalDateTime.now()) + ".txt");
```
Machine-Maker: I would have separate subfolders for each type of verbose mode instead of shoving everything into the debug folder.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventWatcherCommand implements PaperSubcommand {
++
++    /**
++     * The key is the equivalent of {@link Class#getSimpleName()} on the class,
++     * and value is the file we're trying to log to
++     */
++    private static final List<Watcher> watching = Collections.synchronizedList(new ArrayList<>()); //only sync in the rare case that multiple people are debugging, should be unnecessary
```
Machine-Maker: This should be a `Map<Class<? extends Event>, smth>` instead of a List. twice you are doing streams to filter to find the one that matches the supplied class.

patches/server/0970-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,220 @@
...
++            JsonObject obj = new JsonObject();
++            obj.addProperty("event", event.getEventName());
++            try {
++                for (Field f : event.getClass().getDeclaredFields()) {
++                    if(f.getName().equalsIgnoreCase("handlers"))
++                        continue;
++                    f.setAccessible(true);
++                    Object field = f.get(event);
++                    obj.addProperty(f.getName(), field == null ? "null" : field.toString());
++                }
```
Machine-Maker: Same thing here as for the other command. Since its almost identifal logic, it should be abstracted out. There's a lot of similarity between these events and therefore a ton of duplicated logic.

Machine-Maker (changes requested):

TheRealRyGuy:
All should be done unless I missed anything 
`dumpevents` now saves to the `event-debug` folder, whereas `watchevent` will log into `event-debug/[event name]/` dir

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,196 @@
...
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventWatcherCommand implements PaperSubcommand {
++
++    /**
++     * The key is the equivalent of {@link Class#getName()} on the class.
++     * The left side of the pair value is what file we're going to write to, and
++     */
++    private static final Map<String, List<Wrapper>> watching = new ConcurrentHashMap<>(); //only concurrent in the rare case that multiple people are debugging, should be unnecessary
```
Machine-Maker: Instead of this pattern, I might make both commands a singleton, in that it has a private constructor and just one instance in a final static field. Then this field can be just an instance field

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,196 @@
...
++    public static void tryLog(Event event, @Nullable Listener listener) { //listener should only be null on the initial event call
++        List<Wrapper> wrappers = watching.get(event.getClass().getName());
++        if (wrappers != null)
++            wrappers.add(new Wrapper(event, listener));
++    }
++
++    @Override
++    public boolean execute(CommandSender sender, String subCommand, String[] args) {
++        if (args.length < 1) {
++            sender.sendMessage(text("Invalid Usage! /paper watchevent [event name]", RED));
```
Machine-Maker: Generally `<event name>` is the syntax for required arguments while `[event name]` means its optional.

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,196 @@
...
+diff --git a/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java b/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java
+index aa44d4685de3caee4131449bead7a084868ff976..9d8d5ae2a8655c49d0d1bcc6e7cd6a94eef9d39c 100644
+--- a/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java
++++ b/src/main/java/io/papermc/paper/command/subcommands/DumpListenersCommand.java
+@@ -143,7 +143,7 @@ public final class DumpListenersCommand implements PaperSubcommand {
+     }
+ 
+     @SuppressWarnings("unchecked")
+-    private static Set<String> eventClassNames() {
++    public static Set<String> eventClassNames() {
```
Machine-Maker: This change should be made the patch which adds the method.

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,196 @@
...
+         EventVerboseCommand.tryLogEvent(event);
++        EventWatcherCommand.tryLog(event, null);
+ 
+         HandlerList handlers = event.getHandlers();
+         RegisteredListener[] listeners = handlers.getRegisteredListeners();
+@@ -54,6 +56,7 @@ class PaperEventManager {
+ 
+             try {
+                 registration.callEvent(event);
++                EventWatcherCommand.tryLog(event, registration.getListener());
```
electronicboy: Should also log the event executor, as that often has info about the actual method being called

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -202,3 +178,75 @@ index 7ce9ebba8ce304d1f3f21d4f15ee5f3560d7700b..dfb91cb8e19e0159e0eba9367ca73961
...
++        JsonObject obj = new JsonObject();
++        obj.addProperty("event", event.getEventName());
++        try {
++            List<Field> fields = new ArrayList<>(Arrays.asList(event.getClass().getDeclaredFields()));
++            if (!shouldSerialize(fields)) return null; //check current event class fields if we should serialize
++            Class<?> clazz = event.getClass().getSuperclass();
++            while (clazz != null) {
++                fields.addAll(Arrays.asList(clazz.getDeclaredFields()));
++                clazz = clazz.getSuperclass();
++            }
```
Machine-Maker: The `shouldSerialize` logic is only applied to the fields on the event class, not all fields in the event type hierarchy. You should collect all the fields while filtering out specific ones.

Also, your shouldSerialize does seem to be incorrect on its face. If *any* of the fields shouldn't serialized, you skip the whole event.

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,196 @@
...
++            out.print(stringWriter);
++        }
++        sender.sendMessage(text("Successfully dumped event information to " + file, GREEN));
++        watching.remove(event);
++    }
++
++    /**
++     * EventWrapper subclass to apply listener information to our Json object
++     */
++    private static final class Wrapper extends EventWrapper {
```
Machine-Maker: Does this also need its own override for shouldSerialize()? I think both wrapper should have the same logic for that, so you can put it in the superclass.

Machine-Maker (changes requested):

patches/server/0968-New-event-verbose-logging-command.patch
TheRealRyGuy: > EDIT: Another thing, is I feel some events should be excluded by default. Like the tick events, ones that are just spammed a ton. Can have something that does include those, but most of the time they are just gonna clog up the system. We can also exclude events who don't have "changeable" values.

shouldSerialize was intentionally implemented because of that ^, figured it'd just be a redundant computation in watch command. Only was checking the classes fields for events like `VehicleUpdateEvent`, because I'd assume the inheritance would be irrelevant. 
Will make it global and check all fields with proper logic either way in a bit

Machine-Maker: I guess that kind of only applies to the verbose functionality that logs the event after every listener is called.

TheRealRyGuy:
Above handles everything regarding `watchevent`, I've spent 15 mins tryna edit event-verbose patch using verbatim contributing instructions and it's just either merge conflicts or modifying the wrong patch, so I'll fix those up later

TheRealRyGuy:
Am wholeheartedly struggling to fix the patch desync but I figured I'd at least leave the full code up w/ changes to be reviewed

TheRealRyGuy:
Is now properly ready for another round of merge god reviews

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,264 @@
...
++import java.time.LocalDateTime;
++import java.time.format.DateTimeFormatter;
++import java.util.ArrayList;
++import java.util.Arrays;
++import java.util.List;
++
++/**
++ * A simple event wrapper with a timestamp to allow for QoL JSon serialization
++ */
++public class EventWrapper {
```
Owen1212055: LoggedEvent

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,199 @@
...
++        JsonWriter jsonWriter = new JsonWriter(stringWriter);
++        jsonWriter.setIndent(" ");
++        jsonWriter.setLenient(false);
++        Streams.write(array, jsonWriter);
++
++        String[] unformatted = event.split("\\.");
++        String eventName = unformatted[unformatted.length - 1];
++
++        Path parent = PARENT.resolve(eventName);
++        Path path = parent.resolve("event-" + eventName + "-" +
```
Owen1212055: this is redundant, as the main parent directory already says the event name

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,264 @@
...
++public class EventVerboseCommand implements PaperSubcommand {
++
++    @Nullable
++    private static EventVerboseCommand instance; //instance is declared registering in PaperCommand
++    private final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH.mm.ss");
++
++    private final List<Wrapper> WRAPPERS = new ArrayList<>();
++    private final Path PARENT = Path.of("event-debug");
++    //ignore unnecessary events
++    private final List<Class<? extends Event>> EXCLUDE = List.of(ServerTickStartEvent.class,
```
Owen1212055: Use a set (although it may not matter currently due to its hyper-optimized implemention... but future planning)

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,199 @@
...
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventWatcherCommand implements PaperSubcommand {
++
++    private static EventWatcherCommand instance; //will be null if called before command reg
++    /**
++     * The key is the equivalent of {@link Class#getName()} of your event class, and the value is our list of {@link Wrapper}'s
++     */
++    private final Map<String, List<Wrapper>> watching = new ConcurrentHashMap<>(); //only concurrent in the rare case that multiple people are debugging, should be unnecessary
```
Owen1212055: Use Map keyed by event class

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,264 @@
...
++                obj.addProperty(f.getName(), field == null ? "null" : field.toString());
++            }
++        } catch (SecurityException | IllegalAccessException ex) {
++            MinecraftServer.LOGGER.warn("Failure to serialize event " + event.getEventName(), ex);
++        }
++        parent.add("info", obj);
++        return parent;
++    }
++
++    public boolean shouldSerialize(List<Field> fields) {
```
Owen1212055, line 255: Exclude the name field in the event class, it causes duplicate names to be logged.

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,264 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/command/PaperCommand.java b/src/main/java/io/papermc/paper/command/PaperCommand.java
+index 7ba60b4b4f29a42c58d98aafc5ea0fa3214f554c..4aaa903c390593507f46d2a5b25d581e446a791b 100644
+--- a/src/main/java/io/papermc/paper/command/PaperCommand.java
++++ b/src/main/java/io/papermc/paper/command/PaperCommand.java
+@@ -45,6 +45,7 @@ public final class PaperCommand extends Command {
+         commands.put(Set.of("dumpitem"), new DumpItemCommand());
+         commands.put(Set.of("mobcaps", "playermobcaps"), new MobcapsCommand());
+         commands.put(Set.of("dumplisteners"), new DumpListenersCommand());
++        commands.put(Set.of("dumpevents"), EventVerboseCommand.get());
```
Owen1212055: This dump event name is rather misleading, it's logging, capturing, or whatever.
I think a better name might be good, and also make the class name match the command.

patches/server/0969-New-specific-event-watching-command.patch
```diff
@@ -0,0 +1,199 @@
...
++        return switch (args.length) {
++            case 0 -> new ArrayList<>(DumpListenersCommand.eventClassNames());
++            case 1 -> DumpListenersCommand.eventClassNames().stream()
++                .filter(clazz -> clazz.toLowerCase(Locale.ROOT).contains(args[0].toLowerCase(Locale.ROOT)))
++                .toList();
++            default -> Collections.emptyList();
++        };
++    }
++
++    private void dumpToFile(CommandSender sender, String event, List<Wrapper> wrappers) throws IOException {
```
Owen1212055: In general, events can quickly accumulate. I think it may be better to instead actively log to a file async, similar to the latest.log that is auto generated. Have it log an event instead of having to store all of these event calls in memory.

patches/server/0968-New-event-verbose-logging-command.patch
```diff
@@ -0,0 +1,264 @@
...
++import java.util.List;
++import java.util.Map;
++import java.util.Collection;
++
++import static net.kyori.adventure.text.Component.text;
++import static net.kyori.adventure.text.format.NamedTextColor.GREEN;
++import static net.kyori.adventure.text.format.NamedTextColor.RED;
++
++@DefaultQualifier(NonNull.class)
++public class EventVerboseCommand implements PaperSubcommand {
```
Owen1212055: Alot of what I said about the previous command applies here too.

Owen1212055 (changes requested):
Generally, look at your field names, some of them are UPPERCASE_EXAMPLE when not needed.

patches/server/0968-New-event-verbose-logging-command.patch
TheRealRyGuy: thoughts on `/logevent [event]` and `/logevents`?

patches/server/0969-New-specific-event-watching-command.patch
TheRealRyGuy: Is latest.log not just sl4j? Would appreciate a point towards an async internal task method

Issue #6456, opened 2021-08-21
Mobs despawn when going through portal due to CB ticking all worlds


Krinios:
### Expected behavior

I expected a wither skeleton to come through the nether portal 

### Observed/Actual behavior

So I went to the farm and a wither skeleton spawned, so I saw it went into the nether portal and I followed it after going through the nether portal the wither skeleton was not found idk where it went

### Steps/models to reproduce

this is the wither skeleton farm that I'm using
https://www.youtube.com/watch?v=VL7HkV4y3Uo&t=684s

### Plugin list

No plugins I made a new server and just dragged the world folder into to test the farm

### Paper version

1.17.1 #196 latest

### Agreements

- [X] I am running the latest version of Paper available from https://papermc.io/downloads.
- [X] I have searched for and ensured there isn't already an open issue regarding this.
- [X] My version of Minecraft is supported by Paper.

imDaniX:
> So I went to the farm and a wither skeleton spawned, so I saw it went into the nether portal and I followed it after going through the nether portal the wither skeleton was not found idk where it went

Did you test with anyone on the other end? I think skeletons just despawn.

e-im:
Yes, it is more than likely that the skeletons just despawned. Please ensure you are close enough to the skeletons, or try with another user around. In addition, check your `despawn-ranges` in paper.yml. They are in blocks.

e-im:
To recap conversation on discord apparently the wither skeletons will despawn when they go through the portal, behaviour that does not persist on vanilla or spigot. I haven't looked at this yet at all, but re-opening for the author until that is done.

ghost:
It is also happening on spigot, however functions as intended on vanilla.

e-im:
caused by https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/server/level/WorldServer.patch?at=9ed7e4fbe4c0cd2076a52d65a9ea8ae810d0e176#123-132 / https://github.com/Byteflux/CraftBukkit/commit/671a2893f42a66ce1e966369ce2212c4df9953eb. I haven't fully evaluated the possible ramifications of reverting this. but seems safeish? Keeping behaviour from 1.2.5 is a bit...

Alex1337F:
Any updates on this?

NBS-Paste:
I'm using a chunk loader plugin and they still despawn.

SRK7Kyros:
[...]

PR #9025, opened 2023-03-21
Don't tick empty worlds

electronicboy:
restores vanilla behavior, will improve commit message if we decide to merge this

This build is intended to debug if anything goes sideways when disabling this check which was added many moons ago

Cross refs: #6456 and various other issues

---
Download the paperclip jar for this pull request: [paper-9025.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1293198167.zip)


Machine-Maker:
Does anyone remember what the "issues on worlds with no players" the craftbukkit comment is referencing? It's from so long ago, no clue if it's still relevant or not.

electronicboy:
From my understanding, generally leaking resources, i.e. entity instances
not being removed properly

Owen1212055:
Perhaps this could be locked under a config option?

electronicboy:
I mean, that would allow it to be merged I guess, can't really get feedback on this otherwise

Owen1212055 (approved):

kennytv (approved):

PR #9030, opened 2023-03-22
Expose Path to jar file in PluginProviderContext

emilyy-dev:


Owen1212055 (approved):

patches/api/0008-Paper-Plugins.patch
```diff
@@ -290,6 +290,14 @@ index 0000000000000000000000000000000000000000..8be416f01055d94b474ecc425fc0bec6
 +    @NotNull
 +    ComponentLogger getLogger();
 +
++    /**
++     * Provides the path to the jar file of the plugin.
++     *
++     * @return the location of the plugin's jar file
++     */
++    @NotNull
++    Path getJarLocation();
```
Machine-Maker: I feel like a file type-agnostic method name would be better like getPath, getSource, getPluginPath, smth like that. Idk how this would ever happen, but in somehow it made sense in the future for this not to return the path to a jar, this would be named poorly.

emilyy-dev: Yeah, I did think about that, but I couldn't settle on a name that felt right
* getPluginLocation/getPluginPath I wouldn't be surprised it it was easily confused with the plugin's data directory
* getPath feels lacking in context

Maybe getPluginSource? But.. idk

Machine-Maker: I think getPluginSource is fine. And in the docs it can elaborate.

emilyy-dev:
Rebased on master, renamed method & reworded javadoc

Machine-Maker (approved):

electronicboy (approved):

PR #9031, opened 2023-03-23
Added abillity to hide LootTables and PDC

GOjayson:
Added this for #8982

Have closed a PR I had made before because i messed up the Branch in some way.


patches/server/0969-Added-config-option-to-hide-Loottables-and-PDC.patch
```diff
@@ -0,0 +1,46 @@
...
+     public CompoundTag sanitizeSentNbt(CompoundTag tag) {
+-        tag.remove("PublicBukkitValues");
++
++        if(level.paperConfig().anticheat.obfuscation.entitydata.hideLootTables && tag.contains("LootTable")){
++            tag.remove("LootTable");
++            tag.remove("LootTableSeed");
++        };
++        if(level.paperConfig().anticheat.obfuscation.entitydata.hidePDC && tag.contains("PublicBukkitValues") ){
++            tag.remove("PublicBukkitValues");
++        };
```
Machine-Maker: Do you need the extra contains() checks? remove should still succeed if it doesn't have that tag key. It's just one less lookup

Machine-Maker (approved):
Approved, but the merger has to put stuff in the right patch files and fix the formatting on the conditional blocks. Also verify that `hidePDC` doesn't turn into `hide-p-d-c` in the config file.

patches/server/0969-Added-config-option-to-hide-Loottables-and-PDC.patch
GOjayson: I just added this just to be sure. I will test it with the rest.

GOjayson:
Ah yes, it changed to hide-p-d-c

image

as for the contains() they are not needed. Will update the patch now.

Machine-Maker (approved):
Merger must move the diff for the `WorldConfiguration` class into the `Paper Config` patch and fix the formatting on the conditionals in the other hunk.

Owen1212055:
Had to change to pdc, setting to PDC results in 
image

Owen1212055 (approved):

Owen1212055 (changes requested):
Looking again, it doesn't seem that there are any cases where block entities will actually send their loot table information.

Certain block entities have to run their serialization logic to pass extra data to the client, and this is what causes PDC to be sent with it. So, I don't actually see why we should make this configurable, as this is unreliable behavior that should most likely be fixed by moving the core serialization behavior.

It would be better to move the PDC serialization logic into one of the different block entity methods instead.

But this is out of the scope of this PR, it's more that the issue this is based off is misleading, and if it's actually worth exposing this as a config option (since this is actually a small side effect of an implementation detail)

GOjayson:
> Looking again, it doesn't seem that there are any cases where block entities will actually send their loot table information.
> 
> Certain block entities have to run their serialization logic to pass extra data to the client, and this is what causes PDC to be sent with it. So, I don't actually see why we should make this configurable, as this is unreliable behavior that should most likely be fixed by moving the core serialization behavior.
> 
> It would be better to move the PDC serialization logic into one of the different block entity methods instead.
> 
> But this is out of the scope of this PR, it's more that the issue this is based off is misleading, and if it's actually worth exposing this as a config option (since this is actually a small side effect of an implementation detail)

So I should undo the PDC part? leave it as it was before?

Owen1212055:
I am not sure, I feel like it may actually be best to leave this not configurable. Because to me, this is a bug.

GOjayson:
Closing PR since issue is closed for being wrong.

PR #9034, opened 2023-03-24
Added a config option for ticking markers

piratecody:
Addresses [#8118](https://github.com/PaperMC/Paper/issues/8118) by adding a configuration option that allows the ticking of marker entities to be enabled and disabled. By default, marker ticking is disabled.

I had an old PR for this, but I didn't modify the patches correctly & the logic in one of the checks was wrong.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1500,13 +1500,15 @@ index 0000000000000000000000000000000000000000..56e9155ebe850bcf759f5ae08e838f1a
 +import org.spongepowered.configurate.objectmapping.meta.Required;
 +import org.spongepowered.configurate.objectmapping.meta.Setting;
 +
-+@SuppressWarnings({"FieldCanBeLocal", "FieldMayBeFinal", "NotNullFieldNotInitialized", "InnerClassMayBeStatic"})
++@SuppressWarnings({ "FieldCanBeLocal", "FieldMayBeFinal", "NotNullFieldNotInitialized", "InnerClassMayBeStatic" })
 +public class WorldConfiguration extends ConfigurationPart {
 +    private static final Logger LOGGER = LogUtils.getClassLogger();
-+    static final int CURRENT_VERSION = 30; // (when you change the version, change the comment, so it conflicts on rebases): rename filter bad nbt from spawn eggs
++    static final int CURRENT_VERSION = 30; // (when you change the version, change the comment, so it conflicts on
++                                           // rebases): rename filter bad nbt from spawn eggs
 +
 +    private transient final SpigotWorldConfig spigotConfig;
 +    private transient final ResourceLocation worldKey;
++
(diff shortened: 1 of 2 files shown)


MiniDigger (dismissed):
please revert all the unrelated whitespace changes, makes it really hard to review

patches/server/0005-Paper-config-files.patch
```diff
@@ -1574,6 +1574,12 @@ index 0000000000000000000000000000000000000000..51cf0014c4229fc8671804d885b63819
 +            public boolean tick = true;
 +        }
 +
++        public Markers markers;
++
++        public class Markers extends ConfigurationPart {
++            public boolean tick = false;
```
Machine-Maker: This should default to true to keep vanilla behavior. Having them not tick by default breaks them working as passengers.

patches/server/0839-Don-t-tick-markers.patch
```diff
@@ -3,46 +3,47 @@ From: Noah van der Aa <ndvdaa@gmail.com>
...
-index 66b1ef69fe48340b5ccebd845b39f898515ff117..7ae780c4a38eb088f876e5bd03a0dba5836787fb 100644
+index 66b1ef69fe48340b5ccebd845b39f898515ff117..5051aea4f71c992d1884e4462435775af8beee4f 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
 @@ -2497,6 +2497,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
          }
  
          public void onTickingStart(Entity entity) {
-+            if (entity instanceof net.minecraft.world.entity.Marker) return; // Paper - Don't tick markers
++            if (entity instanceof net.minecraft.world.entity.Marker && !paperConfig().entities.markers.tick) return; // Paper - Don't tick markers by default
```
Machine-Maker: The paper commands should be changed to all be smth like `configurable marker ticking` instead of specify `Don't tick markers by default`. Since there is a configuration option, specifying what the default is just more annoying to change if we ever decide to change the default (like I suggested in the other comment).

Machine-Maker (changes requested):

patches/server/0005-Paper-config-files.patch
piratecody: I made this false to keep Paper's behavior consistent with how it has been since the patch was originally applied (Jan 2022ish?). Doesn't make much difference to me, though - if it's better to have the vanilla behavior by default then I'll change it rq

Machine-Maker: Yeah, I think we should stick to the vanilla default here. It should've been that from the beginning.

Machine-Maker (approved):

piratecody:
@MiniDigger Any updates?

Owen1212055 (approved):

PR #9042, opened 2023-03-25
Use ApiStatus.Internal instead of Deprecated

Machine-Maker:
This should finally rid the API once and for all of deprecations that aren't actually deprecations (assuming I got them all, feel free to chime in with ones I missed).

Will be merged into the Fix Spigot annotations patch when ready.

### DIFF
--- a/patches/api/0172-Fix-Spigot-annotation-mistakes.patch
+++ b/patches/api/0172-Fix-Spigot-annotation-mistakes.patch
@@ -6,8 +6,39 @@ Subject: [PATCH] Fix Spigot annotation mistakes
 while some of these may of been true, they are extreme cases and cause
 a ton of noise to plugin developers.
 
+Use ApiStatus.Internal instead of Deprecated for actual internal API
+that continues to have use (internally).
+
 These do not help plugin developers if they bring moise noise than value.
 
+diff --git a/src/main/java/org/bukkit/Art.java b/src/main/java/org/bukkit/Art.java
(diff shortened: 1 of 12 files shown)


Machine-Maker:
Rebased. Things of note here:
* Most of the diff is changing `Deprecated` to `Internal` for still-used magic values.
* A couple annotation fixes as well, like undeprecating `TextDisplay#get/setBackgroundColor`. I think upstream forgot to remove the annotations from that [here](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/bceda6abd980a97681c2cd8bef4e0af68ba4af6d). 
* Deprecates 2 legacy materialdata enums (there always seems to be more of this API to deprecate).

patches/api/0422-Use-ApiStatus.Internal-instead-of-Deprecated.patch
```diff
@@ -0,0 +1,700 @@
...
+ 
+     /**
+      * Sets the text background color.
+      *
+      * @param color new background color
+-     * @deprecated API subject to change
+      */
+-    @Deprecated
++    @Deprecated // Paper - is stable API
+     void setBackgroundColor(@Nullable Color color);
```
Warriorrrr: This annotation doesn't seem to be commented out like the one above is.

Machine-Maker: good catch, corrected.

lynxplay (approved):

PR #9048, opened 2023-03-26
Added the ability to disable non-player markers

### LINKED ISSUE #8464: remove the green markers from maps (if it possible)

it killing fps / server perfomances & get player higher pings when someone tried to place item frames with normal maps as many as & hold the map & also getting some glitches <how is it works> -> https://www.youtube.com/watch?v=W12zKDvHsQI&ab_channel=FitMC (i can't post my own video cuz of due limits of file size)

i hope this one should be fix able or i need to find other way to prevent em (but seems like it doesn't has some kind of plugins to prevent it)

remove green markers from map

### Other

_No response_

poonkje112:
Fix #8464

### DIFF
--- a/patches/server/0968-Added-the-ability-to-disable-non-player-map-markers.patch
+++ b/patches/server/0968-Added-the-ability-to-disable-non-player-map-markers.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: poonkje112 <aaron.knoop@live.nl>
+Date: Sun, 26 Mar 2023 04:17:53 +0200
+Subject: [PATCH] Added the ability to disable non-player map markers
+
+
+diff --git a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+index 50348a5eec1e2ffabfc8405d708461376a162913..27846b631b5ef2b3fee29e9d17963094e3ba0c5e 100644
+--- a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
++++ b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+@@ -583,6 +583,8 @@ public class MapItemSavedData extends SavedData {
+                 // If this cursor is for a player check visibility with vanish system
+                 org.bukkit.entity.Player other = org.bukkit.Bukkit.getPlayerExact(name); // Spigot
+                 if (other == null || player.canSee(other)) {
++                    if(!io.papermc.paper.configuration.GlobalConfiguration.get().misc.showMapMarkers && mapIcon.getType() != MapDecoration.Type.PLAYER) return; // Paper - Configurable map markers
++
+                     icons.add(mapIcon);
+                 }
+             });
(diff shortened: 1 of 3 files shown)


patches/server/0969-Added-the-ability-to-disable-non-player-map-markers.patch
```diff
@@ -0,0 +1,49 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+index 8d442c5a498ecf288a0cc0c54889c6e2fda849ce..772cc473de64076dcb11b84f48fa08ab1d7ef167 100644
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -286,5 +286,6 @@ public class GlobalConfiguration extends ConfigurationPart {
+         public boolean lagCompensateBlockBreaking = true;
+         public boolean useDimensionTypeForCustomSpawners = false;
+         public boolean strictAdvancementDimensionCheck = false;
++        public boolean showMapMarkers = true;
```
Machine-Maker: This bit of diff should go in the patch which adds this file. We try to keep all configuration options in the same patch, where the use of those options can be its own patch. It should be one of the very first server patches.

patches/server/0969-Added-the-ability-to-disable-non-player-map-markers.patch
```diff
@@ -0,0 +1,49 @@
...
+ import net.minecraft.nbt.CompoundTag;
+ import net.minecraft.nbt.ListTag;
+@@ -44,6 +45,10 @@ import org.bukkit.craftbukkit.map.CraftMapView;
+ import org.bukkit.craftbukkit.util.CraftChatMessage;
+ // CraftBukkit end
+ 
++// Paper start
++import io.papermc.paper.configuration.GlobalConfiguration;
++// Paper end
++
```
Machine-Maker: Remove all this diff. Use a fully qualified import for `GlobalConfiguration` and remove the newline added.

Machine-Maker (changes requested):

patches/server/0969-Added-the-ability-to-disable-non-player-map-markers.patch
poonkje112: I totally missed that part in the CONTRIBUTING.MD
It should be resolved now.

patches/server/0969-Added-the-ability-to-disable-non-player-map-markers.patch
poonkje112: Thanks for the feedback, I applied the changes requested!

patches/server/0968-Added-the-ability-to-disable-non-player-map-markers.patch
```diff
@@ -0,0 +1,19 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+index 50348a5eec1e2ffabfc8405d708461376a162913..27846b631b5ef2b3fee29e9d17963094e3ba0c5e 100644
+--- a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
++++ b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+@@ -583,6 +583,8 @@ public class MapItemSavedData extends SavedData {
+                 // If this cursor is for a player check visibility with vanish system
+                 org.bukkit.entity.Player other = org.bukkit.Bukkit.getPlayerExact(name); // Spigot
+                 if (other == null || player.canSee(other)) {
++                    if(!io.papermc.paper.configuration.GlobalConfiguration.get().misc.showMapMarkers && mapIcon.getType() != MapDecoration.Type.PLAYER) return; // Paper - Configurable map markers
```
Warriorrrr, line 15: This will also need to be added to the CraftMapRenderer class

Warriorrrr (changes requested):

PR #9062, opened 2023-03-28
Add Structure check API

Machine-Maker:
As far as I can tell, structures can overlap in generation, so instead of just returning the nullable structure at the location, this just asks for a structure to check if it exists.


electronicboy (approved):

Machine-Maker:
I added a note about this method potentially causing chunk loads/generation if the position isn't loaded and added a helper method to check if a Position is loaded or not.

lynxplay (approved):

PR #9064, opened 2023-03-28
Expose more native methods for PDC

### LINKED ISSUE #8925: Expose more native map methods to PersistentDataContainer

Will reduce boilerplate code & reduce multiple map calls where unnecessary.

The methods are self-explanatory, since they would be modifying the underlying map, just like the existing PersistentDataContainer set & get methods.

PersistentDataContainer::computeIfAbsent
PersistentDataContainer::compute
PersistentDataContainer::computeIfPresent

Having to access the underlying map numerous times, instead of being able to use the methods provided by the underlying map.

### Other

_No response_

TheRealRyGuy:
ref #8925 

PDC now inherits `Iterable<NamespacedKey>`, and adds these methods 
- #setIfAbsent(NamespacedKey, PersistentDataType, Object)
- #computeIfAbsent(NamespacedKey, PersistentDataType, Function)
- #computeIfPresent(NamespacedKey, PersistentDataType, BiFunction)
- #compute(NamespacedKey, PersistentDataType, BiFunction)
- #getOrThrow(NamespacedKey, PersistentDataType, Supplier)
- #getOrThrow(NamespacedKey, PersistentDataType)
- #replace(NamespacedKey, PersistentDataType, Object, Object)
- #replace(NamespacedKey, PersistentDataType, Object)

_i suck at java docs and i apologize_

### DIFF
--- a/patches/api/0420-More-native-map-PDC-methods-use-adventure-Key.patch
+++ b/patches/api/0420-More-native-map-PDC-methods-use-adventure-Key.patch
@@ -1,0 +1,235 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: RyGuy <git@ryguy.me>
+Date: Tue, 28 Mar 2023 01:27:28 -0400
+Subject: [PATCH] More native map PDC methods, use adventure Key
+
+This commit has PDC inherit Iterable<NamespacedKey>, replaces all NamespacedKey uses with adventure's key,
+and implements these methods
+- #setIfAbsent(NamespacedKey, PersistentDataType, Object)
+- #computeIfAbsent(NamespacedKey, PersistentDataType, Function)
+- #computeIfPresent(NamespacedKey, PersistentDataType, BiFunction)
+- #compute(NamespacedKey, PersistentDataType, BiFunction)
+- #getOrThrow(NamespacedKey, PersistentDataType, Supplier)
+- #getOrThrow(NamespacedKey, PersistentDataType)
+- #replace(NamespacedKey, PersistentDataType, Object, Object)
+- #replace(NamespacedKey, PersistentDataType, Object)
+
+diff --git a/src/main/java/org/bukkit/persistence/PersistentDataContainer.java b/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
+index 57609b7793122e135fa0c3b926500849379637b2..cc38aec898f408b157e8e32aa63cb71a445cbfcb 100644
+--- a/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
++++ b/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
+@@ -1,15 +1,21 @@
+ package org.bukkit.persistence;
+ 
+-import java.util.Set;
+ import org.bukkit.NamespacedKey;
+ import org.jetbrains.annotations.NotNull;
+ import org.jetbrains.annotations.Nullable;
+ 
++import java.util.Iterator;
++import java.util.NoSuchElementException;
++import java.util.Set;
++import java.util.function.BiFunction;
++import java.util.function.Function;
++import java.util.function.Supplier;
++
+ /**
+  * This interface represents a map like object, capable of storing custom tags
+  * in it.
+  */
+-public interface PersistentDataContainer {
++public interface PersistentDataContainer extends Iterable<net.kyori.adventure.key.Key> {
+ 
+     /**
+      * Stores a metadata value on the {@link PersistentDataHolder} instance.
+@@ -116,7 +122,7 @@ public interface PersistentDataContainer {
+     <T, Z> Z getOrDefault(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type, @NotNull Z defaultValue);
+ 
+     /**
+-     * Get a set of keys present on this {@link PersistentDataContainer}
++     * Get a set of {@link NamespacedKey}s present on this {@link PersistentDataContainer}
+      * instance.
+      *
+      * Any changes made to the returned set will not be reflected on the
+@@ -197,5 +203,181 @@ public interface PersistentDataContainer {
+     default void readFromBytes(byte @NotNull [] bytes) throws java.io.IOException {
+         this.readFromBytes(bytes, true);
+     }
++
++    /**
++     * @return The iterator bound to the set from {@link #keySet()}
++     */
++    @Override
++    default @NotNull Iterator<net.kyori.adventure.key.Key> iterator() {
++        return keySet().iterator();
++    }
++
++    /**
++     * Will set a metadata value on the {@link PersistentDataHolder} instance if the value is not currently set
++     *
++     * @param key   the key this value will be stored under
++     * @param type  the type this tag uses
++     * @param value the value stored in the tag
++     * @param <T>   the generic java type of the tag value
++     * @param <Z>   the generic type of the object to store
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws NullPointerException     if the value is null. Removing a tag should
++     *                                  be done using {@link #remove(NamespacedKey)}
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> void setIfAbsent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
++
++    /**
++     * Attempts to set a metadata value to your {@link PersistentDataHolder} instance if the supplied key isn't bound,
++     * or the current value is null. If your {@param mappingFunction} returns null, it will not set the value
++     *
++     * @param key             the key this value will be stored under
++     * @param type            the type this tag uses
++     * @param mappingFunction the function to compute your value
++     * @param <T>             the generic java type of the tag value
++     * @param <Z>             the generic type of the object to store
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @NotNull Z computeIfAbsent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                                     @NotNull Function<net.kyori.adventure.key.Key, Z> mappingFunction);
++
++    /**
++     * Attempts to remap a current metadata value on your {@link PersistentDataHolder} instance if the value
++     * is currently present and the mapping function returns a non-null value
++     * The key is removed if your remapped value is null
++     *
++     * @param key               the key this value will be stored under
++     * @param type              the type this tag uses
++     * @param remappingFunction the function to compute your value
++     * @param <T>               the generic java type of the tag value
++     * @param <Z>               the generic type of the object to store
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @Nullable Z computeIfPresent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                                      @NotNull BiFunction<net.kyori.adventure.key.Key, Z, Z> remappingFunction);
++
++    /**
++     * Attempts to remap an {@link PersistentDataHolder} metadata value using a supplied mapping function
++     * Your key is removed if your remapped value is null
++     *
++     * @param key               the key this value will be stored under
++     * @param type              the type this tag uses
++     * @param remappingFunction the function to compute your value
++     * @param <T>               the generic java type of the tag value
++     * @param <Z>               the generic type of the object to store
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @Nullable Z compute(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                             @NotNull BiFunction<net.kyori.adventure.key.Key, Z, Z> remappingFunction);
++
++    /**
++     * Attempts to retrieve a value from the metadata of your {@link PersistentDataHolder} instance. If there is
++     * no value, it will throw an exception
++     *
++     * @param key               the key this value will be stored under
++     * @param type              the type this tag uses
++     * @param exceptionSupplier supplies the exception you want to throw
++     * @param <T>               the generic java type of the tag value
++     * @param <Z>               the generic type of the object to store
++     * @param <X>               the exception type you want to throw
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance, if present
++     *
++     * @throws NullPointerException if the key is null
++     * @throws NullPointerException if the type is null
++     * @throws X                    If there was no metadata element bound to your key and type
++     */
++    <T, Z, X extends Throwable> @NotNull Z getOrThrow(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                                                     @NotNull Supplier<X> exceptionSupplier) throws X;
++
++    /**
++     * Attempts to retrieve a value from the metadata of your {@link PersistentDataHolder} instance. If there is
++     * no value, it will throw an exception
++     *
++     * @param key  the key this value will be stored under
++     * @param type the type this tag uses
++     * @param <T>  the generic java type of the tag value
++     * @param <Z>  the generic type of the object to store
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance, if present
++     *
++     * @throws NullPointerException   if the key is null
++     * @throws NullPointerException   if the type is null
++     * @throws NoSuchElementException If there was no metadata element bound to your key and type
++     */
++    default <T, Z> @NotNull Z getOrThrow(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type) {
++        return getOrThrow(key, type, () -> new NoSuchElementException("Key " + key.asString() + " not present in PDC " +
++            this.getClass().getName()));
++    }
++
++    /**
++     * Replaces the metadata value of your specified key and type if it is currently bound to your `oldValue`
++     *
++     * @param key      the key this value will be stored under
++     * @param type     the type this tag uses
++     * @param oldValue expected value bound to your specified key
++     * @param newValue value to bind to your specified key
++     * @param <T>      the generic java type of the tag value
++     * @param <Z>      the generic type of the object to store
++     *
++     * @return True if the mapped value was changed, false otherwise
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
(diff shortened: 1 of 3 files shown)


Machine-Maker (changes requested):
There is a ton of random diff in here to already existing lines. Like all the javadocs for existing methods seem to have been reformatted.

Remove that so it makes it easier to see what has changed.

TheRealRyGuy:
Diff should be readable now, didn't realize I committed white space last night
tests are apparently failing for missing annotations that aren't missing, will have time later for that

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,281 @@
...
++import java.util.function.BiFunction;
++import java.util.function.Function;
++import java.util.function.Supplier;
++
+ /**
+  * This interface represents a map like object, capable of storing custom tags
+  * in it.
+  */
+-public interface PersistentDataContainer {
++public interface PersistentDataContainer extends Iterable<NamespacedKey> {
```
Machine-Maker: Don't use `NamespacedKey`, use `Key` from adventure.

Machine-Maker (changes requested):
All the new `NamespacedKey` should just be adventure `Key`. Also, while I like having default methods on interfaces for simple things, I'm not sure how I feel about just putting the full implementations of these methods on the interface. I think all the new methods with the exception of the `getOrThrow(Key, PDT)` should be on the Craft impl type.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
TheRealRyGuy: IDE is screaming at me because this would require changing `#getKeys` to use Key
Wasn't sure if this should be done because it can be considered an API break?

TheRealRyGuy:
Moved all impls to craftpdc

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
Machine-Maker: yeah, don't change the return type of getKeys(), but what we can do is add a new method, `keySet()` (also following Map method names) that returns a `Set<Key>`. And then use that to get the iterator.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     * @throws NullPointerException     if the value is null. Removing a tag should
++     *                                  be done using {@link #remove(NamespacedKey)}
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> void setIfAbsent(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
++
++    /**
++     * Attempts to set a metadata value to your {@link PersistentDataHolder} instance if the supplied key isn't bound,
++     * or the current value is null. If your {@param mappingFunction} is null, it will not set the value
```
Machine-Maker: This makes it sound like if the parameter passed is null, not if the return value of the mappingFunction is null.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     * @param <Z>   the generic type of the object to store
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws NullPointerException     if the value is null. Removing a tag should
++     *                                  be done using {@link #remove(NamespacedKey)}
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> void setIfAbsent(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
```
Machine-Maker: Use adventure `Key`.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> Z computeIfAbsent(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type,
++                                     @NotNull Function<NamespacedKey, Z> mappingFunction);
```
Machine-Maker: Use adventure `Key` in both places here.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> Z computeIfPresent(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type,
++                                      @NotNull BiFunction<NamespacedKey, Z, Z> remappingFunction);
```
Machine-Maker: Use adventure `Key` in both places here.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> Z compute(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type,
++                             @NotNull BiFunction<NamespacedKey, Z, Z> remappingFunction);
```
Machine-Maker: Use adventure `Key` in both places

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     * @param <T>  the generic java type of the tag value
++     * @param <Z>  the generic type of the object to store
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance, if present
++     *
++     * @throws NullPointerException   if the key is null
++     * @throws NullPointerException   if the type is null
++     * @throws NoSuchElementException If there was no metadata element bound to your key and type
++     */
++    default <T, Z> Z getOrThrow(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type) {
```
Machine-Maker: Use adventure `Key`

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++        return getOrThrow(key, type, () -> new NoSuchElementException("Key " + key.getKey() + " not present in PDC " +
```
Machine-Maker: Use `key.asString()` to include the whole key

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     * @param <Z>      the generic type of the object to store
++     *
++     * @return True if the mapped value was changed, false otherwise
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> boolean replace(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type,
```
Machine-Maker: Use adventure `Key`.

patches/api/0418-Expose-more-native-map-methods-for-PDC.patch
```diff
@@ -0,0 +1,217 @@
...
++     * @param <Z>   the generic type of the object to store
++     *
++     * @return The previous value bound to your key, or null if no value exists
++     * 
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> Z replace(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
```
Machine-Maker: Use adventure `Key`.

Machine-Maker (changes requested):

TheRealRyGuy:
All review changes applied - so many adventure key changes were requested I just ported all of PDC to use adventure keys, which also involved changing DirtyCraftPDC to support 
 `getKeys` and `keySet` methods don't feel like they belong in the same class to me, but I clarified documentation on both to emphasize the difference

patches/api/0418-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,290 @@
...
+      * @throws NullPointerException if the key is null
+      * @throws NullPointerException if the type is null
+      * @throws NullPointerException if the value is null. Removing a tag should
+-     * be done using {@link #remove(NamespacedKey)}
++     * be done using {@link #remove(net.kyori.adventure.key.Key)}
+      * @throws IllegalArgumentException if no suitable adapter will be found for
+      * the {@link PersistentDataType#getPrimitiveType()}
+      */
+-    <T, Z> void set(@NotNull NamespacedKey key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
++    <T, Z> void set(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
```
Machine-Maker: Yeah, there was a reason I didn't include changing this. This is an ABI break, plugins compiled against previous versions of the API will not run with this change.

Only new methods can use Key unless we also add bytecode rewriting (which we will do at some point as well, but its not needed now).

Machine-Maker (changes requested):
There was a reason I didn't request changes to all the other methods that already use NamespacedKey... See other comment.

patches/api/0418-More-native-map-PDC-methods-use-adventure-Key.patch
TheRealRyGuy: image
When it changing only what you specified

Machine-Maker: move the method impl to the Craft impl, and then can probably just make duplicates of all the methods as private methods, that just accept strings from the NamespacedKeys that's all they are done with anyways, the impl just calls toString on the keys.

So you'd have
```java
@Override
public <T, Z> Z get(NamespacedKey key, PersistentDataType<T, Z> type) {
	// Paper start
	return this.get(key.toString(), type);
}
private <T, Z> Z get(String key, PersistentDataType<T, Z> type) {
	// Paper end
	// method impl
}
```

TheRealRyGuy:
Should be the way you wanted it now

lynxplay:
> could this also expose more the list tag, the possibility to make a pdc type that is a list of other tags?
> 
> for now we only have a list/array of pdc containers, which is a list of compounds, but we don't have like list/array of strings, also not sure but i think nms can have a list of lists of something (even of some more recursive lists), and this also is impossible right now with the api
> 
> so a new pdc type which can wrap any pdc type and make a list of that type would be useful 

A list type is a somewhat complex other topic that should not be added into this PR in my opinion. It needs a lot more design choices made.

patches/server/0971-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,225 @@
...
++    }
++
++    private <T, Z> void set(String key, PersistentDataType<T, Z> type, Z value) {
++        //Paper end
+         Validate.notNull(key, "The provided key for the custom value was null");
+         Validate.notNull(type, "The provided type for the custom value was null");
+         Validate.notNull(value, "The provided value for the custom value was null");
+ 
+-        this.customDataTags.put(key.toString(), this.registry.wrap(type.getPrimitiveType(), type.toPrimitive(value, adapterContext)));
++        this.customDataTags.put(key, this.registry.wrap(type.getPrimitiveType(), type.toPrimitive(value, adapterContext)));
```
Machine-Maker: `// Paper`

patches/server/0971-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,225 @@
...
++        return has(key.toString(), type);
++    }
++
++    private <T, Z> boolean has(String key, PersistentDataType<T, Z> type) {
++        //Paper end
+         Validate.notNull(key, "The provided key for the custom value was null");
+         Validate.notNull(type, "The provided type for the custom value was null");
+ 
+-        Tag value = this.customDataTags.get(key.toString());
++        Tag value = this.customDataTags.get(key);
```
Machine-Maker: `// Paper`

patches/server/0971-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,225 @@
...
++        return get(key.toString(), type);
++    }
++
++    private <T, Z> Z get(String key, PersistentDataType<T, Z> type) {
++        //Paper end
+         Validate.notNull(key, "The provided key for the custom value was null");
+         Validate.notNull(type, "The provided type for the custom value was null");
+ 
+-        Tag value = this.customDataTags.get(key.toString());
++        Tag value = this.customDataTags.get(key);
```
Machine-Maker: `// Paper`

patches/server/0971-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,225 @@
...
++        //Paper start
++        remove(key.toString());
++    }
++
++    private void remove(String key) {
++        //Paper end
+         Validate.notNull(key, "The provided key for the custom value was null");
+ 
+-        this.customDataTags.remove(key.toString());
++        this.customDataTags.remove(key);
```
Machine-Maker: `// Paper`

patches/server/0971-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,225 @@
...
+ 
+     @Override
+@@ -189,5 +218,115 @@ public class CraftPersistentDataContainer implements PersistentDataContainer {
+             this.putAll(compound);
+         }
+     }
++
++    @Override
++    public <T, Z> void setIfAbsent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value) {
++        Z z = this.get(key.toString(), type);
```
Machine-Maker: For adventure keys, I think we prefer the `asString` method instead of Object's toString. Because different implementations of Key could do whatever they want with toString, but asString's contract specifies what it has to return.

Machine-Maker (changes requested):
Just some paper comments, and switch to use Key#asString instead of Key#toString.

TheRealRyGuy:
rebased, fixed tests, all requested changes should be fixed

TheRealRyGuy:
rebased to 1.20

TheRealRyGuy:
Rebased again, condensed commit history - should be ready to go

patches/server/0980-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,227 @@
...
+         Preconditions.checkArgument(value != null, "The provided value cannot be null");
++        //Paper start
++        set(key.toString(), type, value);
++    }
++
++    private <T, Z> void set(String key, PersistentDataType<T, Z> type, Z value) {
++        //Paper end
++        Preconditions.checkArgument(key != null, "The provided key for the custom value was null");
++        Preconditions.checkArgument(type != null, "The provided type for the custom value was null");
++        Preconditions.checkArgument(value != null, "The provided value for the custom value was null");
```
Machine-Maker: So Some of these Precondition checks are duplicated now. I think I'd be for not having those checks on the private methods, as those checks should be done in whatever method calls that method.

patches/server/0980-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,227 @@
...
+         Preconditions.checkArgument(key != null, "The NamespacedKey key cannot be null");
+         Preconditions.checkArgument(type != null, "The provided type cannot be null");
++        //Paper start
++        return has(key.toString(), type);
++    }
++
++    private <T, Z> boolean has(String key, PersistentDataType<T, Z> type) {
++        //Paper end
++        Preconditions.checkArgument(key != null, "The provided key for the custom value was null");
++        Preconditions.checkArgument(type != null, "The provided type for the custom value was null");
```
Machine-Maker: Same here with these preconditions and the ones above in the public `has` method.

Machine-Maker (changes requested):

TheRealRyGuy:
did not see those rebasing, fixed

patches/api/0420-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,235 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @NotNull Z computeIfAbsent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                                     @NotNull Function<net.kyori.adventure.key.Key, Z> mappingFunction);
```
Machine-Maker, line 103: The `mappingFunction` type should be `Function<? super Key, ? extends Z>`.

patches/api/0420-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,235 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @Nullable Z computeIfPresent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                                      @NotNull BiFunction<net.kyori.adventure.key.Key, Z, Z> remappingFunction);
```
Machine-Maker, line 124: The `remappingFunction` type should be `BiFunction<? super Key, ? super Z, ? extends Z>`.

Look at the signature of the Map methods, they all follow this pattern.

patches/api/0420-More-native-map-PDC-methods-use-adventure-Key.patch
```diff
@@ -0,0 +1,235 @@
...
++     *
++     * @return The currently existing metadata within your {@link PersistentDataHolder} instance
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> @Nullable Z compute(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type,
++                             @NotNull BiFunction<net.kyori.adventure.key.Key, Z, Z> remappingFunction);
```
Machine-Maker, line 144: The `remappingFunction` type should be `BiFunction<? super Key, ? super Z, ? extends Z>`.

Machine-Maker (changes requested):

PR #9069, opened 2023-03-28
Deprecate ChatColor

Machine-Maker:
All API that uses it is deprecated/obsolete

### DIFF
--- a/patches/server/0188-Call-PaperServerListPingEvent-for-legacy-pings.patch
+++ b/patches/server/0188-Call-PaperServerListPingEvent-for-legacy-pings.patch
@@ -6,16 +6,17 @@ Subject: [PATCH] Call PaperServerListPingEvent for legacy pings
 
 diff --git a/src/main/java/com/destroystokyo/paper/network/PaperLegacyStatusClient.java b/src/main/java/com/destroystokyo/paper/network/PaperLegacyStatusClient.java
 new file mode 100644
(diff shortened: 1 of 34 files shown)


yannicklamprecht (approved):

JRoy (approved):
Will PR ChatColors2 once this is merged 👍

4drian3d (approved):

patches/server/0188-Call-PaperServerListPingEvent-for-legacy-pings.patch
```diff
@@ -69,12 +70,12 @@ index 0000000000000000000000000000000000000000..74c012fd40491f1d870fbc1aa8c318a2
 +    }
 +
 +    public static String getMotd(PaperServerListPingEvent event) {
-+        return getFirstLine(event.getMotd());
++        return getFirstLine(PlainTextComponentSerializer.plainText().serialize(event.motd()));
```
kennytv: This looks like an api behavior break (if intentional)?

Machine-Maker: nah, you right. I was just replacing uses of deprecated stuff, but this is a valid use of getMotd() so I'll just suppress that use.

kennytv (approved):

Owen1212055 (approved):

PR #9075, opened 2023-03-30
Make debug dump file names consistent

powercasgamer:
before: 
image

after:
image

PR #9076, opened 2023-03-30
add hyphen to syncloadinfo command file

TheRealRyGuy:
the world deserves more hyphens
image


jpenilla:
Thanks for your contribution, but this is already being handled in #9075

PR #9080, opened 2023-04-01
Resend more data on profile refresh

AlexProgrammerDE:
This PR resolves issues with game features resetting to default state on respawn/profile change. Currently, breaking are potion effects and the XP bar, both are now being resent to make the client know about them again. Because the #teleport method also updates abilities and scaled health, I'm updating those as well, just in case.
It may be worth looking into other things that need to be resent in the future.
All of these additions are taken from the teleport respawn code that is used when the respawn packet is sent. There are some other packets like setting view distance or some specific metadata/attributes/permission things for the player, but I think it's a bit unnecessary to mess around with all of those right now. Think I've caught the worst ones in this PR.

### DIFF
--- a/patches/server/0184-Player.setPlayerProfile-API.patch
+++ b/patches/server/0184-Player.setPlayerProfile-API.patch
@@ -55,7 +55,7 @@ index e7442952ef1f03969949014492a7ddc6d0796ba5..69a1852905dd4724c30ac8ab88c14251
  
      public Server getServer() {
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
-index f19b8258e30133f3111ca27e9082b7f3a4ce3877..dd1e5b2ec23b90dff11fce2de0caea10342d3168 100644
+index 61baea9efa7c1b35c437f6f0a4387c7611f401c3..9099604b5bffc92ca46c096d2d6d7d857ef22c38 100644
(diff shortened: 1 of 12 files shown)


kennytv (approved):

patches/server/0184-Player.setPlayerProfile-API.patch
```diff
@@ -166,6 +166,16 @@ index f19b8258e30133f3111ca27e9082b7f3a4ce3877..53074702d0d606ce0e26c4cf435c3b71
...
 +
++        // Resend their XP and effects because the respawn packet resets it
++        connection.send(new net.minecraft.network.protocol.game.ClientboundSetExperiencePacket(handle.experienceProgress, handle.totalExperience, handle.experienceLevel));
++        for (net.minecraft.world.effect.MobEffectInstance mobEffect : handle.getActiveEffects()) {
++            connection.send(new net.minecraft.network.protocol.game.ClientboundUpdateMobEffectPacket(handle.getId(), mobEffect));
++        }
++
++        // Resend other player information
++        handle.onUpdateAbilities();
++        updateScaledHealth();
```
Owen1212055: We shouldn't need to update scaled health, it should already be updated in the player list update method.

AlexProgrammerDE: Alright, also removed onUpdateAbilities, seems like it is already called in the same method.

AlexProgrammerDE:
Should I pull latest changes from master? Seems like there are currently some merge conflicts.

Owen1212055 (approved):

PR #9084, opened 2023-04-02
Added Support for Unix Domain Socket.

0xIO32:
Added Support for Unix Domain Socket.

I already have a Pull Request on Velocity to add also support for this:
https://github.com/PaperMC/Velocity/pull/991

PR #9086, opened 2023-04-02
Fixed Unix Domain Socket Patch

0xIO32:
It Fixes that an exception is thrown when using a Unix Domain Socket.

There was already a Pull Request for this, but i made a new one, because it has added a new patch instead of updating an existing: https://github.com/PaperMC/Paper/pull/9084


Owen1212055 (changes requested):
Please wrap your diff in paper comments, and use fully qualified names to avoid imports.

Warriorrrr:
@MarkusTieger Any chance you could revisit this PR and address Owen's comment? If you don't have time to update it we could update/supersede this ourselves.

MarkusTieger:
> @MarkusTieger Any chance you could revisit this PR and address Owen's comment? If you don't have time to update it we could update/supersede this ourselves.

Sorry, i completely forgot about this pr. I will update it tomorrow and address the comments...

MarkusTieger:
If you need something to test:
https://github.com/MarkusTieger/Velocity/tree/dev/3.0.0
(from an older velocity pr, that wasn't merge because it broke the api. I updated it and it still works fine. Velocity 5.0.0 does support uds, but not 1.20.1)

patches/server/0608-Add-Unix-domain-socket-support.patch
```diff
@@ -139,3 +140,40 @@ index 50c79be12cf8778a2ce451a8c9bb2cdc277d6830..8393c1a2d15d5ad255c9808b0d0edd6a
...
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
+index 3dbad6f9ea70db07a9bfa199b28def6a67353d80..684da653376549e20bbda04f6f4e9e62130f6269 100644
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
++++ b/src/main/java/net/minecraft/server/players/PlayerList.java
+@@ -688,7 +688,12 @@ public abstract class PlayerList {
+ 
+         ServerPlayer entity = new ServerPlayer(this.server, this.server.getLevel(Level.OVERWORLD), gameprofile);
+         Player player = entity.getBukkitEntity();
+-        PlayerLoginEvent event = new PlayerLoginEvent(player, loginlistener.connection.hostname, ((java.net.InetSocketAddress) socketaddress).getAddress(), ((java.net.InetSocketAddress) loginlistener.connection.channel.remoteAddress()).getAddress());
++        java.net.InetAddress rawAddress = java.net.InetAddress.getLoopbackAddress();
```
Warriorrrr: Some paper comments are still needed here

patches/server/0608-Add-Unix-domain-socket-support.patch
```diff
@@ -139,3 +140,40 @@ index 50c79be12cf8778a2ce451a8c9bb2cdc277d6830..8393c1a2d15d5ad255c9808b0d0edd6a
...
++++ b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
+@@ -331,7 +331,13 @@ public class ServerLoginPacketListenerImpl implements ServerLoginPacketListener,
+         public void fireEvents() throws Exception {
+                         String playerName = ServerLoginPacketListenerImpl.this.gameProfile.getName();
+                         java.net.InetAddress address = ((java.net.InetSocketAddress) ServerLoginPacketListenerImpl.this.connection.getRemoteAddress()).getAddress();
+-                        java.net.InetAddress rawAddress = ((java.net.InetSocketAddress) connection.channel.remoteAddress()).getAddress(); // Paper
++                        // Paper start
++                        java.net.SocketAddress rawSocketAddress = connection.channel.remoteAddress();
++                        java.net.InetAddress rawAddress = java.net.InetAddress.getLoopbackAddress();
++                        if(rawSocketAddress != null && rawSocketAddress instanceof java.net.InetSocketAddress rawInetSocketAddress) {
```
Warriorrrr: Missing space between the if and (, same for the diff below

Warriorrrr (commented):
Still missing some paper comments, will give this a test later today

patches/server/0609-Add-Unix-domain-socket-support.patch
```diff
@@ -139,3 +140,42 @@ index 50c79be12cf8778a2ce451a8c9bb2cdc277d6830..8393c1a2d15d5ad255c9808b0d0edd6a
...
++++ b/src/main/java/net/minecraft/server/network/ServerLoginPacketListenerImpl.java
+@@ -331,7 +331,13 @@ public class ServerLoginPacketListenerImpl implements ServerLoginPacketListener,
+         public void fireEvents() throws Exception {
+                         String playerName = ServerLoginPacketListenerImpl.this.gameProfile.getName();
+                         java.net.InetAddress address = ((java.net.InetSocketAddress) ServerLoginPacketListenerImpl.this.connection.getRemoteAddress()).getAddress();
+-                        java.net.InetAddress rawAddress = ((java.net.InetSocketAddress) connection.channel.remoteAddress()).getAddress(); // Paper
++                        // Paper start
++                        java.net.SocketAddress rawSocketAddress = connection.channel.remoteAddress();
++                        java.net.InetAddress rawAddress = java.net.InetAddress.getLoopbackAddress();
++                        if (rawSocketAddress != null && rawSocketAddress instanceof java.net.InetSocketAddress rawInetSocketAddress) {
```
Warriorrrr: Null check here can be removed here and in the other spot, the instanceof takes care of it

electronicboy (dismissed):

lynxplay:
After some further discussion, this should properly address the API as well.
imo best to pass the entire SocketAddress to the PlayerLoginEvent and AsyncPlayerPreLoginEvent, deprecated the existing
methods yielded InetAddress and having them do the fallback, e.g. the API would have the 
```java
if (this.socketAddr instanceOf InetSocketAddress inet) return inet.getAddress();
return getLoopback()
```

and adding a new method that yields the plain socket addr.
Additionally, mark the constructors you change with `@ApiStatus.Internal`

MarkusTieger:
Would be great, if someone could rebase it, because there are conflicts and i never actually dealt with them in **patches**.
(Let's say i tryed, and after that, the applyPatches task failed)

MarkusTieger:
tbh I don't want to deal with this pr anymore.

PR #9087, opened 2023-04-02
Port airplane's use array for gamerule storage patch

Warriorrrr:
This PR ports over airplane's use array for gamerule storage patch, which improves gamerule lookups (which happen very frequently during ticking). Permission from Paul to have this ported into paper was given of course.


patches/server/0973-Use-array-for-gamerule-storage.patch
```diff
@@ -0,0 +1,66 @@
...
+     }
+ 
+     public CompoundTag createTag() {
+@@ -187,6 +204,10 @@ public class GameRules {
+     }
+ 
+     public static final class Key<T extends GameRules.Value<T>> {
++        // Paper start
++        private static int lastGameRuleIndex = 0;
++        public final int gameRuleIndex = lastGameRuleIndex++;
```
kennytv, line 62: Why does this exist?

electronicboy (approved):

PR #9094, opened 2023-04-03
Fix block place logic

Lulu13022002:
Fix some broken block place logic:
- The dispensed shulker will not play the place sound (due to its logic been removed)
- The hanging sign text editor cannot be cancelled using the BlockPlaceEvent (not consistent with the other regular signs)
- The jukebox logic hasn't been synced with the RecordItem


kennytv (approved):

Owen1212055 (approved):

Owen1212055:
Snuck in some paper comment fixes, thank you tho!

PR #9096, opened 2023-04-03
Mob experience reward API

iliabhz:
You can get the amount of experience the mob will drop from the API

### DIFF
--- a/patches/api/0418-Mob-experience-reward-API.patch
+++ b/patches/api/0418-Mob-experience-reward-API.patch
@@ -1,0 +1,24 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: GodOfPro <1387ilia@gmail.com>
+Date: Mon, 3 Apr 2023 22:57:24 +0430
+Subject: [PATCH] Mob experience reward API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Mob.java b/src/main/java/org/bukkit/entity/Mob.java
+index d53de07b40d9cd63a9553d4b0ad442d526530b92..cb67b2e81e59b2b1244e5e80fb41cd44cdb5b5ca 100644
+--- a/src/main/java/org/bukkit/entity/Mob.java
++++ b/src/main/java/org/bukkit/entity/Mob.java
+@@ -178,4 +178,13 @@ public interface Mob extends LivingEntity, Lootable {
+       */
+     public void setLeftHanded(boolean leftHanded);
+     // Paper end
++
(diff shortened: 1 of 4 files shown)


patches/api/0418-Mob-experience-reward-API.patch
```diff
@@ -0,0 +1,24 @@
...
+--- a/src/main/java/org/bukkit/entity/Mob.java
++++ b/src/main/java/org/bukkit/entity/Mob.java
+@@ -178,4 +178,13 @@ public interface Mob extends LivingEntity, Lootable {
+       */
+     public void setLeftHanded(boolean leftHanded);
+     // Paper end
++
++    // Paper start
++    /**
++     * Gets the amount of experience the mob will drop
```
Machine-Maker, line 18: This isn't true. It is not returning the amount of xp the mob **will** drop, as a lot of that has to do with randomized values. This is generating one potential value for the xp to drop, and there's a good chance that calling it multiple times on an entity will return different values. This needs to be clarified int the docs

Machine-Maker (changes requested):
I think from an API perpective, `getExperienceReward` is an extremely misleading method name. I think it should be like `generateExperienceReward` or `getPossibleExperienceReward`. Something that hints and the random nature of the value returned.

patches/api/0418-Mob-experience-reward-API.patch
iliabhz: I changed the method name to getPossibleExperienceReward and stated that the value is randomized

Machine-Maker: You should merge your patches together. You can use `git commit --amend` to merge any staged changes into the last commit.

iliabhz: can you explain how to this properly ? i am not too sure how to do it

Machine-Maker: So in the -Server/ and -API/ directories, you can do `git reset --soft HEAD~1` which basically undoes the last commit (which should be the one you did to implement my suggestions). The `--soft` part doesn't remove the changes, it just removes the commit so those changes are now back in the set of staged changes. Then, you can do `git commit --amend --no-edit` to add those changes to the new latest commit (which should be your first commit since you just deleted the second one). 

Its not a huge deal, and if you can't figure it out, its fine. Whoever merges this PR can handle it.

iliabhz: Alright thanks. i will try to fix it

PR #9097, opened 2023-04-03
Add component support to bukkit's BossBar

Machine-Maker:
Bukkit's BossBar interface is still very much supported, and I think it probably will continue to be supported, but it was missing methods relating to Component.


MiniDigger (approved):

Owen1212055:
As prompted by some of the conversations from paper contributing, I think we should really first discuss if we want to upkeep two totally separate api or refer one to another.

Because at least, the current adventure bossbar impl apparently can be cleaned up due now being able to provide your own platform implementation… 

For me, this is equivalent to having two entity types in the api, like two ItemStack types. This is honestly confusing in the api, and I think that we should make some kind of adventure “view” of a bukkit bossbar.

MiniDigger:
maybe https://github.com/KyoriPowered/adventure/pull/894 can help

PR #9038, opened 2023-03-24
Fix PaperPluginMeta#getDisplayName

TheRealRyGuy:
fixes #9027, below is the only change
```java
@Override
    public @NotNull String getDisplayName() {
        return this.displayName + " v" + this.getVersion();
    }
```

PR #9098, opened 2023-04-04
Fix inconsistencies with plugin names

TheLukeGuy:
Hello! I recently noticed that -- since the introduction of Paper plugins -- there are a number of inconsistencies (and issues) with the different types of plugin names and where they're used. Below is a table showing the results of tests I conducted on multiple versions of Paper. The name of the test plugin is `Very_Fancy_Plugin`.

Test Results

The big issues with this currently (specifically with Paper plugins) are:
1. `PluginDescriptionFile#getName()` and `PaperPluginMeta#getName()` return different strings when plugin names contain underscores. This is very unintuitive and -- in the case of `PluginDescriptionFile#getName()` -- incorrect according to the [documentation](https://jd.papermc.io/paper/1.19/io/papermc/paper/plugin/configuration/PluginMeta.html#getName()):
    > - Will only contain alphanumeric characters, underscores, hyphens, and periods: [a-zA-Z0-9_\\-\\.].
2. `PluginProviderContext#getDataDirectory()` returns a directory with the version in the name, while `JavaPlugin#getDataFolder()` returns one without the version. Not only is this inconsistency unintuitive, but it's also problematic to include the version in the directory name. This is a bug caused by #9038.
3. Both of the methods mentioned in <span>#</span>2 can return directories with spaces in their names, which can be a pain to work with.
4. Plugins with underscores in their names will be listed with their original names (e.g., `Very_Fancy_Plugin`) in `/plugins`, despite those underscores being replaced with spaces (e.g., `Very Fancy Plugin`) in `/version [plugin name]`.
5. Nag messages use all kinds of name variations, sometimes containing original plugin names; other times containing original plugin names with underscores replaced with spaces; other times appending the version; etc.

**Everything below here is now irrelevant. See [this comment](https://github.com/PaperMC/Paper/pull/9098#issuecomment-1496450556) instead.**

---

~~My solution uses the following naming conventions:~~
- ~~Plugin **name** = The original name of the plugin as defined in `[paper-]plugin.yml`, e.g., `Very_Fancy_Plugin`.~~
- ~~Plugin **display name** = A human-readable name for the plugin with no defined behavior, e.g., `Very Fancy Plugin` or `Very_Fancy_Plugin`.~~
- ~~Plugin **display name + version** = The plugin's display name + version, e.g., `Very Fancy Plugin v1.0.0` or `Very_Fancy_Plugin v1.0.0`.~~

~~Specifically, my solution is to:~~
- ~~Rename the current `PluginMeta#getDisplayName` method (and all its usages) to `getDisplayNameWithVersion`.~~
- ~~Add a new `PluginMeta#getDisplayName` method that returns just the display name.~~
- ~~Modify `PluginDescriptionFile#getFullName` to use `PluginMeta#getDisplayNameWithVersion`.~~
- ~~Modify the `/plugins` and `/version` commands to use the plugin's display name.~~
- ~~Change plugin data directories to use the plugin's name in all places. This matches the behavior of Spigot plugins.~~
- ~~Change nag messages to use the plugin's display name + version in (I believe) all places.~~

~~This shouldn't cause any API breakages that render plugins uncompilable/unusable, however, naturally, this could break any plugins that depend on a specific implementation of the aforementioned methods. All of these changes will affect Paper plugins, but Spigot plugins should only be affected by the changes to `PluginMeta#getDisplayName` (as well as the addition of `PluginMeta#getDisplayNameWithVersion`).~~

~~Things I'm unsure of are:~~
~~1. Is there a better name for `PluginMeta#getDisplayNameWithVersion`? It could alternatively be `getDisplayNameAndVersion` or `getFullDisplayName` or something entirely different.~~
~~2. Should plugin logger names be changed to use display names? They're technically "displayed" to users, but it could be considered unconventional to use spaces.~~
~~3. Should `/version` tab completion be changed to use display names in some way? I think this would be a little weird considering how tab completion works, but it would be more consistent.~~
~~4. Is this even the right approach? 😬~~

---

Thanks! :)

---
Download the paperclip jar for this pull request: [paper-9098.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/649665440.zip)

### DIFF
--- a/patches/server/0404-Wait-for-Async-Tasks-during-shutdown.patch
+++ b/patches/server/0404-Wait-for-Async-Tasks-during-shutdown.patch
@@ -22,7 +22,7 @@ index e189de6d2aa94e9bbb20f1477ee2e34431adb324..4a58843f7ce2dd9e50f9daf3065d056a
          // CraftBukkit end
          if (this.getConnection() != null) {
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
-index 4d51834b6e7bc9058579ce5c6b88f5e9e1c3155d..c92b61c0228d112c73253a5c8b6654312f2b6e95 100644
+index 4d51834b6e7bc9058579ce5c6b88f5e9e1c3155d..670ad9f79e2430e956ef204c7e7bf0c651810c06 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
 @@ -1002,6 +1002,31 @@ public final class CraftServer implements Server {
@@ -46,8 +46,8 @@ index 4d51834b6e7bc9058579ce5c6b88f5e9e1c3155d..c92b61c0228d112c73253a5c8b665431
 +            Plugin plugin = worker.getOwner();
 +            getLogger().log(Level.SEVERE, String.format(
 +                "Nag author(s): '%s' of '%s' about the following: %s",
-+                plugin.getDescription().getAuthors(),
-+                plugin.getDescription().getFullName(),
++                plugin.getPluginMeta().getAuthors(),
++                plugin.getPluginMeta().getDisplayName(),
 +                "This plugin is not properly shutting down its async tasks when it is being shut down. This task may throw errors during the final shutdown logs and might not complete before process dies."
 +            ));
 +        }
--- a/patches/server/0404-Wait-for-Async-Tasks-during-shutdown.patch
+++ b/patches/server/0404-Wait-for-Async-Tasks-during-shutdown.patch
@@ -22,10 +22,10 @@ index e189de6d2aa94e9bbb20f1477ee2e34431adb324..4a58843f7ce2dd9e50f9daf3065d056a
          // CraftBukkit end
          if (this.getConnection() != null) {
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
-index bec471a617780e4fa88d4c559c42175370687d7b..9e278a740cf7cb7a593d76f01d6aa590407b087b 100644
+index 4d51834b6e7bc9058579ce5c6b88f5e9e1c3155d..c92b61c0228d112c73253a5c8b6654312f2b6e95 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
-@@ -1002,6 +1002,35 @@ public final class CraftServer implements Server {
+@@ -1002,6 +1002,31 @@ public final class CraftServer implements Server {
          org.spigotmc.WatchdogThread.hasStarted = true; // Paper - Disable watchdog early timeout on reload
      }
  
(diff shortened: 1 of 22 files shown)


Owen1212055:
Thank you for your very detailed PR!
Because it changes directory names, we most likely want to get a fix for this out ASAP.
 
Instead of introducing another method, can we try to get the paper plugin implemention to match the bukkit implementation as close as possible? Instead of having to introduce some new method. 

So, whatever the current plugin description file implementation is, consider that the standard.

TheLukeGuy:
Sure! I just assumed that the underscore to space thing was intentional, which formed the basis of the changes I made. Just to be clear, you want that gone entirely so it's consistent with Bukkit plugins? Even in `getDisplayName`?

Owen1212055:
Yes, ideally we can just make it all consistent.

patches/server/0659-Add-System.out-err-catcher.patch
```diff
@@ -89,7 +89,7 @@ index 0000000000000000000000000000000000000000..76d0d00cd6742991e3f3ec827a75ee87
 +                    String.format("Nag author(s): '%s' of '%s' about their usage of System.out/err.print. "
 +                            + "Please use your plugin's logger instead (JavaPlugin#getLogger).",
 +                        plugin.getDescription().getAuthors(),
-+                        plugin.getName())
++                        plugin.getDescription().getFullName())
```
Owen1212055: Ideally this should call the meta method instead of description

patches/server/0404-Wait-for-Async-Tasks-during-shutdown.patch
```diff
@@ -44,14 +44,10 @@ index bec471a617780e4fa88d4c559c42175370687d7b..9e278a740cf7cb7a593d76f01d6aa590
...
-+            String author = "<NoAuthorGiven>";
-+            if (plugin.getDescription().getAuthors().size() > 0) {
-+                author = plugin.getDescription().getAuthors().get(0);
-+            }
 +            getLogger().log(Level.SEVERE, String.format(
-+                "Nag author: '%s' of '%s' about the following: %s",
-+                author,
-+                plugin.getDescription().getName(),
++                "Nag author(s): '%s' of '%s' about the following: %s",
++                plugin.getDescription().getAuthors(),
```
Owen1212055: Same for these too

TheLukeGuy:
Sounds good! Here's the updated table of names:

Test Results

Owen1212055 (approved):
I think this is good. Thank you so much for the in-depth testing here!

TheLukeGuy:
Rebased on PaperMC/master

TheLukeGuy:
Sure! I believe that testing a lot of the behavior here (like the `PluginDescriptionFile`'s name + display name + full name, as well as the plugin's data directory) would require creating a test instance of `JavaPlugin` which I don't think could be done without adding an extra constructor to `JavaPlugin` that takes a `ConfiguredPluginClassLoader` or something. Should I do that, or should the tests just be specific to `PaperPluginMeta`? Or am I missing something else?

Owen1212055:
Just make the test specific to ``PaperPluginMeta``. Make public methods to modify fields, but make sure to mark them with a @TestOnly annotation.

TheLukeGuy:
Is that good?

Tests passed: 2 of 2 tests – 92ms

I can't think of any other tests that would work without a `JavaPlugin`.

Owen1212055:
Perfect, can we also have a test that accounts for the whole underscore logic?

TheLukeGuy:
Sorry, not totally sure what you mean by the underscore logic? I think the current tests would cover underscore issues as accurately as it can without creating a plugin, but is there something else specifically that should be tested?

Owen1212055:
Nevermind, I see what you mean. Looks good to me! 😄

Owen1212055 (approved):

TheLukeGuy:
Sounds good, thanks! :D

Machine-Maker (approved):

Owen1212055:
Thank you for your contribution! 😄

Issue #8911, opened 2023-02-24
Plugin chat prompts are no longer being cancelled from being visible in chat


Sikatsu:
### Expected behavior

When a plugin asks me to type a value within the chat, the message should be cancelled from being visible to everyone.

### Observed/Actual behavior

The messages are no longer being cancelled and are visible to every player on the server.

### Steps/models to reproduce

You can try this with many different plugins, as they are all broken in this regard:

- Auction house
- DisplayShops
- HyperDrive
- HeadDatabase
- PremiumVanish/SuperVanish
- etc

The issue is there after updating from paper build 399 to 420.
I have tested this on my survival, creative AND nearly empty test server with just HDB and the two different paper builds.

### Plugin and Datapack List

Plugins (28): AntiPopup, ArmorStandEditor, ArtMap, CMI, CMILib, ChatControlRed, CommandPanels, CoreProtect, ExecuteEverywhere, HeadDatabase, LuckPerms, Multiverse-Core, MyCommand, PlaceholderAPI, PlayerParticles, PlotSquared, PremiumVanish, ProtocolLib, ServerUtils, ShowItem, SupremeTags, TAB-Bridge, VCLink2-Bukkit, Vault, VillagerCraftCore, WorldEdit, WorldGuard and eGlow

### Paper version

[11:15:16 INFO]: This server is running Paper version git-Paper-420 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 81d7ff6)
You are running the latest version

### Other

![image](https://user-images.githubusercontent.com/5664135/221153329-75d5b43b-ca5b-4080-87a2-d5ed19c5c265.png)

electronicboy:
Hard to say around premium/closed source plugins, given that the only real change has been in the plugin loader, this would generally scream more towards a plugin issue, primarily around their dependency on event order;

Sikatsu:
I see. Wouldn't it be suspicious this happens to all plugins though? None of the ones I mentioned are open source?

For now i'll stay on the old version.

Sikatsu:
It indeed runs behind a waterfall proxy. Though nothing was changed or updated on that regard. All that I have actually changed is the paper version on the backend servers.

I'll take a look at the website you provided.

Sikatsu:
Would installing HuskChat also solve my issue?

[...]

PR #9099, opened 2023-04-04
Fix ignoreCancelled having no effect on events registered by direct EventExecutor

harrydevane:
We recently updated our Paper version and noticed that `ignoreCancelled` was no longer having an effect on our listeners.

This only impacts events registered by direct `EventExecutor`. Events registered using the `registerEvents` method are not impacted.

This is potentially also the cause of https://github.com/PaperMC/Paper/issues/8911 as it looks like the plugins mentioned are registering events this way (see https://github.com/PaperMC/Paper/issues/8911#issuecomment-1482589353.


Owen1212055 (approved):

Owen1212055:
Bahahah, thank you.

Machine-Maker (approved):

electronicboy (approved):

PR #9102, opened 2023-04-07
Add mob spawner ranges changes and API

DavidTs93:
- Add vertical range to mob spawners
- Add Paper settings for default mob spawner ranges (in `paper-global.yml`)
- Add API for mob spawner's vertical range
- Change the existing mob spawner range API's details to represent the change

### DIFF
--- a/patches/api/0417-Add-mob-spawner-ranges-API.patch
+++ b/patches/api/0417-Add-mob-spawner-ranges-API.patch
@@ -1,0 +1,71 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DavidTs93 <david.ts93@gmail.com>
+Date: Fri, 7 Apr 2023 19:28:25 +0300
+Subject: [PATCH] Add mob spawner ranges API
+
+
+diff --git a/src/main/java/org/bukkit/block/CreatureSpawner.java b/src/main/java/org/bukkit/block/CreatureSpawner.java
+index 5bbae759ce39d42886994e500fd9454ec328f804..1e79a6068297fc96d5eca215a802a1345b70aa5a 100644
+--- a/src/main/java/org/bukkit/block/CreatureSpawner.java
++++ b/src/main/java/org/bukkit/block/CreatureSpawner.java
+@@ -174,33 +174,52 @@ public interface CreatureSpawner extends TileState {
+      * in order for this spawner to be active.
+      */
+     public void setRequiredPlayerRange(int requiredPlayerRange);
+-
++    
++    // Paper start
+     /**
+-     * Get the radius around which the spawner will attempt to spawn mobs in.
++     * Get the horizontal radius around which the spawner will attempt to spawn mobs in.
+      * <br>
+      * This area is square, includes the block the spawner is in, and is
+      * centered on the spawner's x,z coordinates - not the spawner itself.
+      * <br>
+-     * It is 2 blocks high, centered on the spawner's y-coordinate (its bottom);
++     * It is {@link #getSpawnRangeVertical()} + 1 blocks high, centered on the spawner's y coordinate (its bottom);
+      * thus allowing mobs to spawn as high as its top surface and as low
+-     * as 1 block below its bottom surface.
++     * as {@link #getSpawnRangeVertical()} blocks below its bottom surface.
+      * <br>
+      * Default value is 4.
+      *
+-     * @return the spawn range
++     * @return the horizontal spawn range
+      */
+     public int getSpawnRange();
+ 
+     /**
+-     * Set the new spawn range.
++     * Set the new horizontal spawn range.
+      * <br>
+      *
+-     * @param spawnRange the new spawn range
++     * @param spawnRange the new horizontal spawn range
+      * @see #getSpawnRange()
+      */
+     public void setSpawnRange(int spawnRange);
+ 
+-    // Paper start
++    /**
++     * Get the vertical radius around which the spawner will attempt to spawn mobs in.
++     * <br>
++     * Default value is 1.
++     *
++     * @return the vertical spawn range
++     * @see #getSpawnRange()
++     */
++    public int getSpawnRangeVertical();
++    
++    /**
++     * Set the new vertical spawn range.
++     * <br>
++     *
++     * @param spawnRangeVertical the new vertical spawn range
++     * @see #getSpawnRangeVertical()
++     */
++    public void setSpawnRangeVertical(int spawnRangeVertical);
++
+     /**
+      * Check if spawner is activated (a player is close enough)
+      *
--- a/patches/api/0417-Add-mob-spawner-ranges-API.patch
+++ b/patches/api/0417-Add-mob-spawner-ranges-API.patch
@@ -5,16 +5,12 @@ Subject: [PATCH] Add mob spawner ranges API
 
 
 diff --git a/src/main/java/org/bukkit/block/CreatureSpawner.java b/src/main/java/org/bukkit/block/CreatureSpawner.java
-index 5bbae759ce39d42886994e500fd9454ec328f804..1e79a6068297fc96d5eca215a802a1345b70aa5a 100644
+index 5bbae759ce39d42886994e500fd9454ec328f804..dbe8e405d261b867c972b83eb923ba836d1d2e1a 100644
 --- a/src/main/java/org/bukkit/block/CreatureSpawner.java
 +++ b/src/main/java/org/bukkit/block/CreatureSpawner.java
-@@ -174,33 +174,52 @@ public interface CreatureSpawner extends TileState {
-      * in order for this spawner to be active.
-      */
+@@ -176,31 +176,50 @@ public interface CreatureSpawner extends TileState {
      public void setRequiredPlayerRange(int requiredPlayerRange);
--
-+    
-+    // Paper start
+ 
      /**
 -     * Get the radius around which the spawner will attempt to spawn mobs in.
 +     * Get the horizontal radius around which the spawner will attempt to spawn mobs in.
@@ -46,7 +42,7 @@ index 5bbae759ce39d42886994e500fd9454ec328f804..1e79a6068297fc96d5eca215a802a134
       */
      public void setSpawnRange(int spawnRange);
  
--    // Paper start
+     // Paper start
 +    /**
 +     * Get the vertical radius around which the spawner will attempt to spawn mobs in.
 +     * <br>
--- a/patches/api/0417-Add-mob-spawner-ranges-API.patch
+++ b/patches/api/0417-Add-mob-spawner-ranges-API.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add mob spawner ranges API
 
 
 diff --git a/src/main/java/org/bukkit/block/CreatureSpawner.java b/src/main/java/org/bukkit/block/CreatureSpawner.java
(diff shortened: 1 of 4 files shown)


patches/api/0417-Add-mob-spawner-ranges-API.patch
```diff
@@ -0,0 +1,71 @@
...
++     * Set the new horizontal spawn range.
+      * <br>
+      *
+-     * @param spawnRange the new spawn range
++     * @param spawnRange the new horizontal spawn range
+      * @see #getSpawnRange()
+      */
+     public void setSpawnRange(int spawnRange);
+ 
+-    // Paper start
```
Machine-Maker: I would keep the comment here. You don't need to include Javadoc changes inside paper start/end comments.

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,108 @@
...
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -276,6 +276,14 @@ public class GlobalConfiguration extends ConfigurationPart {
+                 executor.setMaximumPoolSize(_chatExecutorMaxSize);
+             }
+         }
++
++        public MobSpawnerRange mobSpawnerRange;
++
++        public class MobSpawnerRange extends ConfigurationPart {
```
Machine-Maker: Should this make some mention of the fact that it's the default range? So like DefaultMobSpawnerRange?

Otherwise it might be more likely to be misinterpreted as overriding any other changes plugins made.

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,108 @@
...
++        }
++        // Paper end
+ 
+         this.displayEntity = null;
+     }
+@@ -313,6 +322,7 @@ public abstract class BaseSpawner {
+         nbt.putShort("MaxNearbyEntities", (short) this.maxNearbyEntities);
+         nbt.putShort("RequiredPlayerRange", (short) this.requiredPlayerRange);
+         nbt.putShort("SpawnRange", (short) this.spawnRange);
++        nbt.putShort("SpawnRangeVertical", (short) this.spawnRangeVertical); // Paper
```
molor: I think `Paper.SpawnRangeVertical` should be used instead

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,108 @@
...
+ 
+                         if (k >= this.maxNearbyEntities) {
+                             this.delay(world, pos);
+@@ -290,6 +293,12 @@ public abstract class BaseSpawner {
+         if (nbt.contains("SpawnRange", 99)) {
+             this.spawnRange = nbt.getShort("SpawnRange");
+         }
++    
++        // Paper start
++        if (nbt.contains("SpawnRangeVertical", 99)) {
```
Machine-Maker: Prefix the key with "Paper." so there is no chance if a key conflict in the future. Also it'd be good to make a private static final String field with the key.

Machine-Maker (changes requested):
Thanks for the contribution!

Just have a couple minor suggestions.

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
DavidTs93: What if someone changes it via a plugin?

DavidTs93: Sorry, I'm new to this whole business, guess you're right :D
I'll add it

DavidTs93:
> Thanks for the contribution!
> 
> Just have a couple minor suggestions.

I agree with these, but how do I make those changes and add them to this PR? I'm new to this whole thing...

Machine-Maker:
> I agree with these, but how do I make those changes and add them to this PR? I'm new to this whole thing...

So what you want to do is edit the patches by ammending to the commits you created in `Paper-Server` and `Paper-API`. Since your patches are the newest patches, just doing `git commit --amend --no-edit` once you've staged changes inside the `-Server` and `-API` directories will merge them into the commit. Then you can `./gradlew rebuildPatches` to rebuild the patches.

DavidTs93:
> > I agree with these, but how do I make those changes and add them to this PR? I'm new to this whole thing...
> 
> So what you want to do is edit the patches by ammending to the commits you created in `Paper-Server` and `Paper-API`. Since your patches are the newest patches, just doing `git commit --amend --no-edit` once you've staged changes inside the `-Server` and `-API` directories will merge them into the commit. Then you can `./gradlew rebuildPatches` to rebuild the patches.

I've added them - hopefully I did it correctly 😁

patches/api/0417-Add-mob-spawner-ranges-API.patch
DavidTs93, line 1: This doesn't fix any other problem, just the first fix (maybe there'll be more so first) so added "#1", didn't know it would associate with something else... Sorry

DavidTs93:
@Machine-Maker do I need to do something now or am I just waiting for this to be approved?

patches/api/0417-Add-mob-spawner-ranges-API.patch
```diff
@@ -0,0 +1,67 @@
...
+     // Paper start
++    /**
++     * Get the vertical radius around which the spawner will attempt to spawn mobs in.
++     * <br>
++     * Default value is 1.
++     *
++     * @return the vertical spawn range
++     * @see #getSpawnRange()
++     */
++    public int getSpawnRangeVertical();
```
Machine-Maker: I think "VerticalSpawnRange" sounds better than SpawnRangeVertical. Its more correct English. Same for all the mentions of those 3 words. Best to stick to the more grammatically correct version.

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,117 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/BaseSpawner.java b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+index 31ac0e5ca26c7bdfa9b710d0bb78d846ddf6863e..0e084a9516f6220065c8f2be175f74be1a036277 100644
+--- a/src/main/java/net/minecraft/world/level/BaseSpawner.java
++++ b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+@@ -4,6 +4,8 @@ import com.mojang.logging.LogUtils;
+ import java.util.Optional;
+ import java.util.function.Function;
+ import javax.annotation.Nullable;
++
++import io.papermc.paper.configuration.GlobalConfiguration;
```
Machine-Maker: Don't add this import, instead use the fully qualified class name on the line that needs it. Generally we don't add imports to files we don't create cause it can cause extra merge conflicts when rebasing on upstream changes.

patches/server/0977-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,117 @@
...
+ import net.minecraft.core.BlockPos;
+ import net.minecraft.core.particles.ParticleTypes;
+ import net.minecraft.core.registries.BuiltInRegistries;
+@@ -27,7 +29,7 @@ import net.minecraft.world.phys.AABB;
+ import org.slf4j.Logger;
+ 
+ public abstract class BaseSpawner {
+-
++    private static final String SPAWN_RANGE_VERTICAL_TAG = "Paper.SpawnRangeVertical";
```
Machine-Maker: Make this just a line addition instead of a line replacement (and add a `// Paper` comment).

Machine-Maker (changes requested):

DavidTs93:
@Machine-Maker everything changed, anything else needs to be done before this can be merged?
Also thx for all the help!

patches/server/1015-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,119 @@
...
+ 
+                         if (k >= this.maxNearbyEntities) {
+                             this.delay(world, pos);
+@@ -284,6 +287,12 @@ public abstract class BaseSpawner {
+         if (nbt.contains("SpawnRange", 99)) {
+             this.spawnRange = nbt.getShort("SpawnRange");
+         }
++    
++        // Paper start
++        if (nbt.contains(SPAWN_RANGE_VERTICAL_TAG, 99)) {
```
Lulu13022002, line 84: You can use the Tag.TAG_ANY_NUMERIC constant

patches/server/1015-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,119 @@
...
+         public boolean loadPermissionsYmlBeforePlugins = true;
+diff --git a/src/main/java/net/minecraft/world/level/BaseSpawner.java b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+index 633500aefd515df5dadda3802b94079f75a03fa0..c557a0d51699c9ecfeee3fa2b3564910f0483032 100644
+--- a/src/main/java/net/minecraft/world/level/BaseSpawner.java
++++ b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+@@ -4,6 +4,7 @@ import com.mojang.logging.LogUtils;
+ import java.util.Optional;
+ import java.util.function.Function;
+ import javax.annotation.Nullable;
++
```
Lulu13022002, line 38: This newline is not needed

patches/api/0429-Add-mob-spawner-ranges-API.patch
```diff
@@ -0,0 +1,67 @@
...
++     * Default value is 1.
++     *
++     * @return the vertical spawn range
++     * @see #getSpawnRange()
++     */
++    public int getVerticalSpawnRange();
++    
++    /**
++     * Set the new vertical spawn range.
++     * <br>
```
Lulu13022002, line 58: This newline is not needed

patches/server/1015-Add-mob-spawner-ranges-changes-and-API.patch
```diff
@@ -0,0 +1,119 @@
...
++        public class DefaultMobSpawnerValues extends ConfigurationPart {
++            public Range range;
++            
++            public class Range extends ConfigurationPart {
++                public int horizontal = 4;
++                public int vertical = 1;
++            }
++        }
++
```
Lulu13022002, line 26: This should be in the config patch ([0005-Paper-config-files.patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0005-Paper-config-files.patch))

Machine-Maker (commented):
I'm not sold on the usefulness or virtue of having the global defaults. They won't apply retroactively to any already-generated spawners because the NBT is already set to 4. I think adding a default for the height is better, but they are already controllable via a plugin.

DavidTs93:
@kennytv why was this closed?

lynxplay:
The master branch was deleted as we have now been on the post paper hardfork layout for 4 full months and this PR hasn't been updated for it.

PR #9103, opened 2023-04-07
Fix demo flag not enabling demo mode

### LINKED ISSUE #9046: Demo mode does not work
### Expected behavior

When started with the `demo` parameter, the server should start in demo mode like vanilla.

### Observed/Actual behavior

The server does not start in demo mode.

CraftBukkit replaces the main() method and sets the demo options correctly (`Main`, L260). However, MinecraftServer#setDemo is commented out and never called, so the server does not run in demo mode (Line 304).

### Steps/models to reproduce

1. Start server with `demo`
2. No demo.

### Plugin and Datapack List

Not relevant

### Paper version

1.19.4 470

### Other

Im a bit unsure if there is a reason behind doing this. It being commented out seems somewhat intentional, but the demo mode really doesn't do much.

Warriorrrr:
Closes #9046

### DIFF
--- a/patches/server/0974-Fix-demo-flag-not-enabling-demo-mode.patch
+++ b/patches/server/0974-Fix-demo-flag-not-enabling-demo-mode.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Warrior <50800980+Warriorrrr@users.noreply.github.com>
+Date: Fri, 7 Apr 2023 20:11:17 +0200
+Subject: [PATCH] Fix demo flag not enabling demo mode
+
+https://github.com/PaperMC/Paper/issues/9046
+
+diff --git a/src/main/java/net/minecraft/server/Main.java b/src/main/java/net/minecraft/server/Main.java
+index c6fa6bcd66d61359124a8426b919493c6ec43f06..cdbf10339a8e8846a8c364de483a0ccd95cd225a 100644
+--- a/src/main/java/net/minecraft/server/Main.java
++++ b/src/main/java/net/minecraft/server/Main.java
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

patches/server/0977-Fix-demo-flag-not-enabling-demo-mode.patch
```diff
@@ -0,0 +1,20 @@
...
+index c6fa6bcd66d61359124a8426b919493c6ec43f06..cdbf10339a8e8846a8c364de483a0ccd95cd225a 100644
+--- a/src/main/java/net/minecraft/server/Main.java
++++ b/src/main/java/net/minecraft/server/Main.java
+@@ -301,7 +301,7 @@ public class Main {
+                 /*
+                 dedicatedserver1.setSingleplayerProfile(optionset.has(optionspec8) ? new GameProfile((UUID) null, (String) optionset.valueOf(optionspec8)) : null);
+                 dedicatedserver1.setPort((Integer) optionset.valueOf(optionspec11));
+-                dedicatedserver1.setDemo(optionset.has(optionspec2));
++                */dedicatedserver1.setDemo(optionset.has("demo"));/* // Paper - Restore setting the demo mode
+                 dedicatedserver1.setId((String) optionset.valueOf(optionspec12));
```
Lulu13022002: The serverId / singleplayer profile is also affected

electronicboy (approved):

patches/server/0977-Fix-demo-flag-not-enabling-demo-mode.patch
Owen1212055: Will pull this fix for now. If people want these options, this fix can prolly be extended.

PR #9108, opened 2023-04-09
Fix NPE due to Nullable Color in CraftMetaColorableArmor

Moulberry:
CraftMetaLeatherArmor prevents this.color from being set to null by substituting DEFAULT_LEATHER_COLOR CraftMetaColorableArmor does not do this, causing various NPEs when leather armor does not have a color NBT tag

This fix copies the setColor method from CraftMetaLeatherArmor and uses it in CraftMetaColorableArmor


Warriorrrr:
Already fixed by upstream over here: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/50ee87fbcb037e4c9bf184d65959089f169ae449

electronicboy:
pulled in upstreams fixed, I think that there was some notion that this/their fix maybe borked something, but, idk

PR #9110, opened 2023-04-09
Optimize hoppers by not trying to merge full items.

DungeonDev:
This can skip many very expensive call to ItemStack.tagMatches. Makes canMergeItems return false for merging into ItemStacks that are already full.

Noticed this due to a stray item going into a double chest otherwise filled with shulkers of NBT-heavy items, which made the hopper pushing items into that chest start creating a noticeable amount of tick ms usage by itself. In testing on my local machine I've been able to make 6 hoppers and double chest cause about 50ms of tick usage by just tagMatches using shulkers of written bocks.

I don't think there is any reason at all to try and merge into items that are already full (please do tell if there is), the original code checks if the item being merged into is greater then its maximum stack size, which makes me believe it might be a logical error from Mojang, using a greater then instead of equal or greater then?

### DIFF
--- a/patches/server/0963-Optimize-Hoppers.patch
+++ b/patches/server/0963-Optimize-Hoppers.patch
@@ -68,7 +68,7 @@ index 585d1d1f4b1b212295da36e31ae2670b0d2b06c3..1b248db497500aa6bd346b306dcb908a
          }
  
 diff --git a/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/HopperBlockEntity.java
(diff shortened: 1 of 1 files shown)


kennytv (approved):

Machine-Maker (approved):
Looks good to me. Should the patch really re-work the logic chain there instead of just changing that one operator to >=? I feel like it makes the diff more confusing.

DungeonDev:
The original code was an unreadable mess of nested ternary operators - the logic chain is equivalent (except the stacksize check of course) but more readable. I was a bit conflicted between the code readability and diff readability myself as well. Should I change the PR to only change that one operator instead?

PR #9111, opened 2023-04-10
Fix exception when using World#generateTree near crops

Redned235:
When generating trees near crops, the server will check whether the crop block can survive, and as part of that check in CropBlock#canSurvive, a lookup to the light engine is done. This adds a change into the BlockStateListPopulator to get the brightness from the world itself, since the light engine is not accessible in DummyGeneratorAccess.

Exception from testing prior to fix:
https://paste.gg/p/anonymous/0a29a4634cfd4fb5b412f20b3de8aaf9

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -55,4 +55,5 @@ TheFruxz <cedricspitzer@outlook.de>
 Kieran Wallbanks <kieran.wallbanks@gmail.com>
 Denery <dorofeevij@gmail.com>
 Jakubk15 <jakubk15@protonmail.com>
+Redned <redned235@gmail.com>
 ```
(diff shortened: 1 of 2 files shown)


Owen1212055:
There should be an already existing patch that deals with similar issues with this class. Could just be merged into this.

Machine-Maker:
This is using the block populator (which extends the DummyGeneratorAccess). I think it should be overriden there and delegate to getting the actual light data.

Redned235:
> This is using the block populator (which extends the DummyGeneratorAccess). I think it should be overriden there and delegate to getting the actual light data.

Should I merge that into the patch Owen mentioned or keep this change as a separate one?

Redned235:
Updated.

Owen1212055 (approved):

kennytv (approved):

Issue #8661, opened 2022-12-13
Boat#getStatus() throws NPE if called before the entity is ticked


Machine-Maker:
### Expected behavior

No error

### Observed/Actual behavior

throws an error

### Steps/models to reproduce

Spawn a boat entity and call getStatus (or just toString) on it right after.

Error is because the boat status on the nms entity isn't set until the first call of tick(). that doesn't happen until later.

### Plugin and Datapack List

none

### Paper version

1.19.3 323

### Other

Not sure what the best fix is. There is a getStatus() method that is called at the top of tick() to determine the status, but idk if it safe to call that before the entity is ticking. Could just pick a status to return as default if its null and document that.

Doc94:
i dont notice this when added the method... maybe can be force a status (with `HANDLE#getStatus` ) when the instance is created or this is the same issue?
the other thing can be add a `#hasStatus` and add docs for check this before to get the status? but this only happen in the first instance...

Warriorrrr:
Resolved by #9115

PR #9115, opened 2023-04-10
Fix NPE on Boat#getStatus

LemonCaramel:
Tested in the three situations

TestPlugin.java
```java
@Override
public void onEnable() {

    /*
     * Get status at the same time as entity spawn
     */
    this.getServer().getCommandMap().register("boat", new Command("boat") {
        @Override
        public boolean execute(@NotNull CommandSender sender, @NotNull String commandLabel, @NotNull String[] args) {
            if (sender instanceof Player player) {
                final var loc = player.getLocation();
                player.sendRawMessage( loc.getWorld().spawn(loc, Boat.class).getStatus().name() );
            }
            return false;
        }
    });
}

@EventHandler
void onEntityAddToWorld(final EntityAddToWorldEvent event) {
    if (event.getEntity() instanceof Boat boat) {
        this.logger.info("EntityAddToWorldEvent: " + boat.getStatus().name());
    }
}

/*
 * Ride on the boat and reconnect the server
 */
@EventHandler
void onEntityMount(final EntityMountEvent event) {
    if (event.getMount() instanceof Boat boat) {
        this.logger.info("EntityMountEvent: " + boat.getStatus().name());
    }
}
```

### DIFF
--- a/patches/api/0429-Fix-NPE-on-Boat-getStatus.patch
+++ b/patches/api/0429-Fix-NPE-on-Boat-getStatus.patch
@@ -1,0 +1,18 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: LemonCaramel <admin@caramel.moe>
+Date: Tue, 11 Apr 2023 04:04:41 +0900
+Subject: [PATCH] Fix NPE on Boat getStatus
+
+
+diff --git a/src/main/java/org/bukkit/entity/Boat.java b/src/main/java/org/bukkit/entity/Boat.java
+index f7548098bcdd033d9c530fdc584fc5538c635ca1..2ac685fb1817f3ce06ebe6391cc863712d68367c 100644
+--- a/src/main/java/org/bukkit/entity/Boat.java
++++ b/src/main/java/org/bukkit/entity/Boat.java
+@@ -169,6 +169,7 @@ public interface Boat extends Vehicle {
+      */
+     public enum Status {
+ 
++        NOT_IN_WORLD, // Paper
+         IN_WATER,
+         UNDER_WATER,
+         UNDER_FLOWING_WATER,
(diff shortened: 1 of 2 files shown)


Machine-Maker:
My only concern, which I mentioned in the issue I opened about this a while ago https://github.com/PaperMC/Paper/issues/8661, is that it's not immediately clear to me that it's safe to call getStatus when the boat hasn't been "in the world". It calls a lot of logic about the current position of the boat and world state which could totally not be written to work when the boat entity isn't in the world. 

An alternative solution would be to add a new "status" like "UNKNOWN" or "NOT_IN_WORLD" or something that it defaults to before it's been ticked. I am not advocating for either solution, it is just something that has to be considered (if you haven't already).

LemonCaramel:
I want to get the always correct status returned. so I want to avoid adding a new status.
_(I want to keep this way until find a fatal problem)_


I try several tests _(change the dimension, spawn very far away, and more...)_, i have not encountered any errors or incorrect results.
```java
final Boat boat = loc.getWorld().spawn(loc, Boat.class, ett -> {
    LOGGER.info("Before: " + ett.getStatus().name());
});

LOGGER.info("After: " + boat.getStatus().name());

Bukkit.getScheduler().runTaskLater(TestPlugin.this, () -> {
    LOGGER.info("after tick(): " + boat.getStatus().name());
}, 2L);
```
 Is there anything else I should test?

talwat:
I'm also experiencing this issue, so I'm glad there's a PR for it.

electronicboy (approved):

patches/server/0973-Fix-NPE-on-Boat-getStatus.patch
```diff
@@ -0,0 +1,32 @@
...
++        // Paper start
++        final var handle = this.getHandle();
++        if (handle.status == null) {
++            if (handle.valid && !handle.updatingSectionStatus) {
++                handle.status = handle.getStatus();
++            } else {
++                return Status.NOT_IN_WORLD;
++            }
++        }
++        if (true) return CraftBoat.boatStatusFromNms(handle.status);
```
Owen1212055: Is this if (True) needed here? As isn't this just doing the same thing?

LemonCaramel: > Is this if (True) needed here? As isn't this just doing the same thing?

I just didn't want to call getHandle twice.

Warriorrrr: getHandle is just a cast so avoiding calling it twice isn't too important

kennytv (approved):
Removed the setting of the status field

PR #9116, opened 2023-04-10
Handle location of player if was moved for not valid world

### LINKED ISSUE #9081: Add `PlayerSpawnInUnknownWorldEvent`

Yea, our players sometimes gets stuck in blocks (in main overworld at same position where they were in the other world) and then die due to suffocation :/

This happens only if the server was restarted and the other world's folder (created by plugin or vanilla nether/the end) was deleted.

Add `PlayerSpawnInUnknownWorldEvent` (or something with better name) that gets called if player joining the game and the server can't find the world this player was in before.

That event contains `#setSpawnLocation(Location)` and nullable `#getSpawnLocation()`. By default that `Location` is `null` to keep existing behavior. If no one plugin changed this `Location`, then this player will spawn in overworld at same position (current behavior), otherwise the server will use that `Location` to spawn this player at.

idk. I'm tried `PlayerSpawnLocationEvent` but it gets called with default world/overworld in this case, so plugins cannot determine exactly which world the player was in before.

### Other

_No response_

Doc94:
This clos e #9081 like the issue said when the world of spawn change because the world is unloaded or not exists keep the previous location causing issues when the player spawn, this PR modify two patchs related to handle the world spawn for add a conditional for call a method used to handle location for "new" players in world.

### DIFF
--- a/patches/server/0384-Move-player-to-spawn-point-if-spawn-in-unloaded-worl.patch
+++ b/patches/server/0384-Move-player-to-spawn-point-if-spawn-in-unloaded-worl.patch
@@ -6,8 +6,46 @@ Subject: [PATCH] Move player to spawn point if spawn in unloaded world
 The code following this has better support for null worlds to move
 them back to the world spawn.
 
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
+index a0832d8e8b9e91073c15e5a87953d6ccee544ae0..89dd1b7d3687df12d84b82666ac5c0201cb11ff8 100644
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
++++ b/src/main/java/net/minecraft/server/players/PlayerList.java
+@@ -202,13 +202,19 @@ public abstract class PlayerList {
+             s = bukkit.contains("lastKnownName", 8) ? bukkit.getString("lastKnownName") : s;
+         }
+         // CraftBukkit end
++        boolean needResetSpawnPos = false; // Paper
+ 
+         if (nbttagcompound != null) {
+             DataResult<ResourceKey<Level>> dataresult = DimensionType.parseLegacy(new Dynamic(NbtOps.INSTANCE, nbttagcompound.get("Dimension"))); // CraftBukkit - decompile error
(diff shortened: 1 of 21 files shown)


Doc94:
NOTE: I prefer only edit the patchs... but maybe is better merge the two patchs in one because the two are related to the spawn player. (but dont know how correctly make this and not destroy git in the process)

also... if can add the label for get a JAR for allow the user in the ticket test, i make basic test using vanilla worlds (me in the end, stop server, delete end and start server again)

Machine-Maker (changes requested):
This only handles when the dimension isn't parsed successfully from the playerdata. You need to add something where the server checks if a world exists with the parsed key. Where the logger warns about spawning in an unknown world

patches/server/0384-Move-player-to-spawn-point-if-spawn-in-unloaded-worl.patch
```diff
@@ -6,8 +6,46 @@ Subject: [PATCH] Move player to spawn point if spawn in unloaded world
...
+             Logger logger = PlayerList.LOGGER;
+ 
+             Objects.requireNonNull(logger);
+-            resourcekey = (ResourceKey) dataresult.resultOrPartial(logger::error).orElse(Level.OVERWORLD);
++            // Paper start - detect when cannot find the dimension
++            //resourcekey = (ResourceKey) dataresult.resultOrPartial(logger::error).orElse(Level.OVERWORLD);
++            Optional<ResourceKey<Level>> optionalResourceKey = dataresult.resultOrPartial(logger::error);
++            needResetSpawnPos = optionalResourceKey.isEmpty();
++            resourcekey = optionalResourceKey.orElse(Level.OVERWORLD);
++            // Paper stop
```
Warriorrrr, line 30:
```suggestion
+            // Paper end
```

Warriorrrr (commented):
needResetSpawnPos should also be set to true if the world isn't found like MM said. https://github.com/PaperMC/Paper/blob/73af2d4cea597e65dd13dafab6325ea5709c9067/patches/server/0435-Spawn-player-in-correct-world-on-login.patch also defaults the level key to overworld which will also need to be accounted for.

Machine-Maker:
Superseded by #9887

PR #9119, opened 2023-04-11
Remove javadoc that doesn't apply to Paper

Machine-Maker:
Paper has a patch that restores ExactChoice working for shapeless recipes.


kennytv (approved):

MiniDigger (approved):
maybe a link to the patch as reference? idk how we would have a link there tho, so maybe in your commit?
in any case, lgtm

Owen1212055:
Oops

PR #9124, opened 2023-04-11
Added wither vanilla invulnerability API

booky10:
This method is called in vanilla directly after spawning the wither.

I'm not sure about the wording of the javadoc.

### DIFF
--- a/patches/api/0300-Missing-Entity-Behavior-API.patch
+++ b/patches/api/0300-Missing-Entity-Behavior-API.patch
@@ -1111,10 +1111,30 @@ index 3794db8867b53f3b3735ad82fdd8765a26df2bfb..efaa45f41bc1dc8df6665c55b4e5ade3
       * Increases the anger level of this warden.
       *
 diff --git a/src/main/java/org/bukkit/entity/Wither.java b/src/main/java/org/bukkit/entity/Wither.java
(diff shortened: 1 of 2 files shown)


patches/api/0301-Missing-Entity-Behavior-API.patch
```diff
@@ -1147,6 +1148,14 @@ index b86f0196e6eb8070830f63a94f732522c2a6c2f1..a1b42ae35dda2da90ba00a2d6666514f
...
 +     */
 +    void setCanTravelThroughPortals(boolean value);
++
++    /**
++     * Makes the wither invulnerable for 11 seconds and
++     * sets the health to one third of the max health.
++     * <br>
++     * This is called in vanilla directly after spawning the wither.
++     */
++    void makeInvulnerable();
```
Machine-Maker: I'm not sure about the name of this method when it also changes the current health of the wither. I get that its called that in mojang mappings, but it doesn't seem descriptive enough.

olijeffers0n: I’m with Machine here, maybe something that links it to the game “event” like “enterInvunlerabilty“

booky10: Maybe call it `enterInvulnerabilityPhase`?

Owen1212055 (approved):

Owen1212055 (approved):

kennytv (approved):

PR #9125, opened 2023-04-11
Undo making JavaPlugin#logger field public

emilyy-dev:
During the development of paper plugins, [api/0072 Add workaround for plugins modifying the parent of the plugin logger](https://github.com/PaperMC/Paper/blob/e338793603e7b57ab439a1b925e3092cb623f616/patches/api/0072-Add-workaround-for-plugins-modifying-the-parent-of-t.patch) was modified to make the `Logger logger` field public so `PaperPluginClassLoader` could set it like `PluginClassLoader` does (the patch originally made the field package-private). This has caused some issues (most noticeably) for devs using Kotlin depending on an API version post paper plugins and trying to run the plugin on older versions (because the kt compiler seeing `plugin.logger` decided to use the field instead of the getter as one would assume) and has been brought a few times already.

MiniDigger suggested the patch could do without making the field public, and I decided to PR that. To change this I decided to DI the logger in the `JavaPlugin#init` method instead of making it package-private or public.

Additionally I added a Commodore fix for the (hopefully few) plugins that were compiled using the field access rather than the getter, mapping getfield to an invokevirtual call (though I did consider not doing this altogether since it would be misuse of the API (?)).

Thanks Owen :3

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -57,4 +57,5 @@ Denery <dorofeevij@gmail.com>
 Jakubk15 <jakubk15@protonmail.com>
 Redned <redned235@gmail.com>
 Luke Chambers <consolelogluke@gmail.com>
+Emily <emilia.lopezf.1999@gmail.com>
 ```
--- a/patches/api/0072-Add-workaround-for-plugins-modifying-the-parent-of-t.patch
+++ b/patches/api/0072-Add-workaround-for-plugins-modifying-the-parent-of-t.patch
@@ -67,33 +67,33 @@ index 0000000000000000000000000000000000000000..087ee57fe5485bc760fadd45a176d4d9
 +
 +}
 diff --git a/src/main/java/org/bukkit/plugin/java/JavaPlugin.java b/src/main/java/org/bukkit/plugin/java/JavaPlugin.java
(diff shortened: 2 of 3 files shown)


emilyy-dev:
After some careful thinking and considering that the Commodore fix:
* is barely useful unless one accesses by `super.logger` inside the plugin class or by `plugin.logger` where `plugin`'s static type is a `JavaPlugin` (**not**  a subclass) at compile-time. `javac` will emit a `getfield me/coolest/MyPlugin logger` if accessed by `this.logger` or `plugin.logger` where `plugin`'s static type is `me.coolest.MyPlugin` (subclassing `JavaPlugin`); attempting to shoehorn a fix to cover all possible cases is impossible in a stateless manner
* does not work on Paper plugins (the PaperPluginClassLoader does not Commodore-process classes when loading them as right now there is no reason to), so the fix only works on Bukkit plugins

I think I'll be removing the Commodore "fix" altogether, bringing the consequence of breaking existing plugins happily (ab)using public access of the logger field.
Alternatively, the field can be kept public and nothing be done about it, the PR closed without merging.

Owen1212055 (approved):

electronicboy (approved):

emilyy-dev:
Rebased on master

PR #9127, opened 2023-04-12
Use player scoreboard for determining collision rule when pushing entities

Redned235:
The vanilla code in EntitySelector#pushable uses the Entity#getTeam method which returns the Level scoreboard. This means the wrong collision option is returned here  when an API scoreboard is being used. Although no actual pushing occurs here as the client takes authority for that, the server still emits footstep sounds and treats it as if collision should be happening when it should not be.

Demonstration of the issue (turn sound on):

https://user-images.githubusercontent.com/29153871/231344227-2ed04480-a01c-4106-8a33-fdbfe49dd720.mp4

### DIFF
--- a/patches/server/0380-Don-t-run-entity-collision-code-if-not-needed.patch
+++ b/patches/server/0380-Don-t-run-entity-collision-code-if-not-needed.patch
@@ -12,7 +12,7 @@ The entity's current team collision rule causes them to NEVER collide.
 Co-authored-by: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
 
 diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
-index e6e4856db7d983460fdf898583915ff2e4a43d14..25e1dc97e3260e7f58b15a1759b8f9531c6a6f20 100644
+index 1619942f973cd20bcfce67e48762c69ffdc1e3e0..cdcb063c1a014cc0028a4eec5ed34a38dd43a697 100644
 --- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
 +++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
 @@ -3328,10 +3328,24 @@ public abstract class LivingEntity extends Entity implements Attackable {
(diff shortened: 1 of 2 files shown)


electronicboy:
collisions get weird, as especially player on player is pretty much 100% server sided, but, there is the wider caveat that this is a potential behavioral change for non player entities? I do wonder if something like this should be configurable

Redned235:
> collisions get weird, as especially player on player is pretty much 100% server sided, but, there is the wider caveat that this is a potential behavioral change for non player entities? I do wonder if something like this should be configurable

I feel this would be more of a bug since in the event a vanilla scoreboard was used, the pushing behavior on the server would be properly in sync. Not entirely sure the ramifications of this, but if this could cause problems in certain situations I can add a config option.

Owen1212055:
If we are overriding the getScoreBoard method, shouldn't we just modify the getTeam behavior to instead use .getScoreBoard() instead of ``this.level.getScoreboard()``. Since, is it possible this issue could occur in other places due to this team desync?

Redned235:
> If we are overriding the getScoreBoard method, shouldn't we just modify the getTeam behavior to instead use .getScoreBoard() instead of `this.level.getScoreboard()`. Since, is it possible this issue could occur in other places due to this team desync?

I was thinking about doing this but noticed that other places like the /team command also would be affected by that. As of now none of the vanilla commands can interact with API scoreboards or teams and I thought going this route would be too big of a behavioral change.

Owen1212055 (approved):

Warriorrrr (approved):

patches/server/0380-Don-t-run-entity-collision-code-if-not-needed.patch
```diff
@@ -23,7 +23,7 @@ index e6e4856db7d983460fdf898583915ff2e4a43d14..25e1dc97e3260e7f58b15a1759b8f953
 +            if (!this.isPushable()) {
 +                return;
 +            }
-+            net.minecraft.world.scores.Team team = this.getTeam();
```
lynxplay, line 26: I'd probably argue that, if we are adding a new patch for this, this change should be in the added patch as well.

Fixing it in two places makes dropping a patch harder / knowing what the patch/fix entails.
Not too sure on it tho, I hope we can get some PRarathon started this weekend and I'll get back on this :+1:

PR #9137, opened 2023-04-15
Add custom per entity tracking range API

bergerkiller:
Adds an additional CUSTOM tracking range mode and a field in Entity that, if set to a positive value, enables a custom tracking range for that entity to be used. This range can be higher or lower than the default entity tracking range of that entity type.

This change clashes slightly with SpottedLeaf's tracking range optimizations from 2020. The AreaMap system looks pretty slow as it seems to be re-calculating the same chunk slots for many of the same tracking range of entity types. I purposefully did not touch any of this and stuck to a more simpler approach of adding a CUSTOM one with an 'unbounded' range (decided on 100 chunks - 1600 blocks) - bounded mostly by the player send distance.

It would be good to hear SpottedLeaf's input on this one before pulling. I feel a lot of this is in need of a rewrite.

A future addition people might request is to set this per player as well, but I didn't want to strain performance any more than really needed here.

### DIFF
--- a/patches/api/0440-Add-custom-per-entity-tracking-range-API.patch
+++ b/patches/api/0440-Add-custom-per-entity-tracking-range-API.patch
@@ -1,0 +1,36 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Irmo van den Berge <irmo.vandenberge@ziggo.nl>
+Date: Sat, 15 Apr 2023 02:43:25 +0200
+Subject: [PATCH] Add custom per-entity tracking range API
+
+Signed-off-by: Irmo van den Berge <irmo.vandenberge@ziggo.nl>
+
+diff --git a/src/main/java/org/bukkit/entity/Entity.java b/src/main/java/org/bukkit/entity/Entity.java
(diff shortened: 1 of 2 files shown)


bergerkiller:
Should this be made persistent with bukkit entity NBT btw?

olijeffers0n:
As a plugin developer, I would somewhat expect this to be persistent in implementation

bergerkiller:
I've added NBT saving/loading of custom tracking range, making it a persistent property

bergerkiller:
Any update on this?

electronicboy:
This is essentially delegated to leaf who afaik is somewhat busy atm

bergerkiller:
It's been a while and 1.20.1 seems done, so I'm bumping this again in case theres time / I need to rebase the commits.

xMvds:
Is there already more info about when this commit will be in paper?

electronicboy:
We don't offer etas, this is something I'd prefer somebody with knowledge of this system to look into

bergerkiller:
Rebased everything on latest paper, as there were some conflicts

PR #9141, opened 2023-04-16
Add support for io_uring

skbeh:
This only works when not using unix domain socket
and compression is disabled because
the netty io_uring transport does not support
unix domain socket and the client can not decode
compressed packets sent with io_uring.

### DIFF
--- a/patches/server/0003-Build-system-changes.patch
+++ b/patches/server/0003-Build-system-changes.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Build system changes
 
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index e090175d2fbe8eba664500feafc29c0567bb0c87..d77f15c58f41f1e00c75e0a022919bb7dacc9926 100644
+index e090175d2fbe8eba664500feafc29c0567bb0c87..410c562d70242d327337f79b55468be00080d992 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -9,10 +9,9 @@ plugins {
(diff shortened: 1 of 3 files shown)


clrxbl:
arm64 support should also be added

electronicboy:
having to disable compression is a huge red flag that something is wrong somewhere, and we shouldn't really be glossing over that

Also, this *really* needs to be configurable, I don't think that we should be pushing unstable linux features and netty incubator projects, especially without any form of sane opt-out

skbeh:
@electronicboy I added the option to make it configurable.

skbeh:
@olijeffers0n I think it should be opt-in now, because it is still relatively experimental.

skbeh:
Also bump netty so that the versions of dependencies match.

toinouH:
About #paper-contrib discussion on netty io_uring here is my opinion:
- I like the idea to disable it by default.
- I think the security concerns (and the experimental natur of netty-io_uring) could be mentioned with a WARN message when enabling this mode on every startup.
-We might also need a WARN or INFO message about the compression issue.

- I also think (as mentioned if I remember) you don't really need to maintain it so it's "free performance".

Sytm:
I am just gonna leave [this Google Blog post](https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html?m=1) here regarding the security state of io_uring as a resource for further discussion if it is still desired to add this into paper.

For me this gives me more "free vulnerabilities" than "free performance" vibes, but I digress.

xymb-endcrystalme:
I ported that patch to my Folia fork and the total throughput of clientbound packets on 4 Netty threads increased from about 120k to about 500k per second.

io_uring is definitely worth implementing in Paper & Velocity. 👍

kennytv:
Ofc throughput is going to increase if you disable compression as well (at the cost of clients and your networking bill), but another problem is that io uring is that, in a funny twist, many big companies have completely blocked io_uring due to being prone to tons of security vulnerabilities

xymb-endcrystalme:
Compression had nothing to do with it. My tests were with compression disabled for epool.

PR #9143, opened 2023-04-16
Fix players are invisible after using setPlayerProfile

### LINKED ISSUE #9061: setPlayerProfile makes player go invisible to others
### Expected behavior

Player not be invisible after running setPlayerProfile 

### Observed/Actual behavior

Player goes invisible after running setPlayerProfile with this error in the other player's client logs

`Server attempted to add player prior to sending player info (Player id f3c4dfb9-1c7b-40ac-81fd-462538538523)`

### Steps/models to reproduce

PlayerProfile profile = Bukkit.createProfile("Simon");
profile.complete();
player.setPlayerProfile(profile);

Other clients that were online won't be able to see player, shows up in tab though

### Plugin and Datapack List

Bukkit Plugins:
- FastAsyncWorldEdit, ProfileTest

### Paper version

1.19.4 - 477

### Other

_No response_

montlikadani:
Fixes #9061 

I am not sure this is the proper solution to fix the issue, but after changing an existing online player's profile using #9061 reproduction code the player appeared to others.


Owen1212055:
It's probably best that we document that the player's uuid won't be swapped, only their name and gameprofile properties..

Machine-Maker:
Shouldn't the profile you set have the same UUID? I'm not quite understanding what the issue is I guess. I would assume as an api consumer setting the game profile to have a different UUID would do some wacky stuff.

Owen1212055:
Well, it's previously supported setting a separate uuid. Should we enforce an exception or something if there is a uuid mismatch?

Machine-Maker:
I feel like an exception is better if those are the options. But there probably is a use case for changing the UUID, so isn't it on plugins to ensure that stuff is the same?

montlikadani:
> But there probably is a use case for changing the UUID, so isn't it on plugins to ensure that stuff is the same?

Yes, we're only changing the UUID to the new profile's id during the update player info packet is being sent and changing back to the original to allow for plugins to retrieve the same player with the changed profile properties.

montlikadani:
Or I thought another solution would be to not change the entire gameprofile on player, but only the skin and other properties, with the exception of the name and UUID.

Owen1212055:
I think an exception is better here. Especially with the cryptographically signed messages, it's rather tricky since that is dependent on the uuid of the player entity. And at least currently, it doesn't seem as easy to change the uuid anymore.

montlikadani:
In this case the `Server#createProfile(String)` method should be marked as `@Obsolete` or `@Deprecated` because the UUID must match every time with the current player to avoid the exception if it not matches.

So basically I added this condition into `setPlayerProfile` impl in which if the UUID not matches it will throws this exception
```java
if (!self.getUUID().equals(gameProfile.getId())) {
    throw new IllegalStateException("The UUID of player does not matches");
}
```
Let me know this is what you want.

Just a side note, the warning log message which coming from the client `Server attempted to add player prior to sending player info` can not be caught via server.

montlikadani:
Updated to 1.20 and resolved conflicts

Machine-Maker (approved):
I updated this PR to not expose the "fake" UUID to the API in the show entity event. Basically it sets and resets the "fake" UUID right before and after sending the player info update packet.

lynxplay (approved):

PR #9144, opened 2023-04-16
Allow non persistent Entities in Raw Byte Serialization

Joo200:
This will allow to serialize non persist entities with the current implementation.

In my case I want to store entities not with the default implementation. So I want to set persistent to false and handle it on my own (EntitiesLoadEvent and EntitiesUnloadEvent). Using serializeEntity and deserializeEntity looks promising but it doesn't return the nbt correctly when it's not persistent.


Owen1212055:
This looks like a dupe of https://github.com/PaperMC/Paper/pull/8492

Joo200:
> This looks like a dupe of #8492

Yes, it is. I didn't know #8492 is currently active and will be merged soon. This should be a much smaller workaround but I don't mind if you merge #8492 instead.

Owen1212055:
Closing in favor of said PR, in general, it's unlikely a temporary workaround (that changes behavior) is the best long-term decision here

Thank you for the PR regardless. 😄

PR #9147, opened 2023-04-17
Fixed large packet handling

lukalt:
Fixed handling of too large packets. This patch seems to got broken at some version update in the past.

### DIFF
--- a/patches/server/0286-Handle-Large-Packets-disconnecting-client.patch
+++ b/patches/server/0286-Handle-Large-Packets-disconnecting-client.patch
@@ -7,16 +7,19 @@ If a players inventory is too big to send in a single packet,
 split the inventory set into multiple packets instead.
 
 diff --git a/src/main/java/net/minecraft/network/Connection.java b/src/main/java/net/minecraft/network/Connection.java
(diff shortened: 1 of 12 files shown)


electronicboy (approved):

Owen1212055:
image
Does max size need to be adjusted?

electronicboy:
was just looking deeper; yes, max size needs to be adjusted, that if statement down below actually needs to be moved up

lukalt:
@Owen1212055 You're right. I actually did not see that the throwing of the PacketTooLargeException was moved further but down. I have updated the constant and removed the second check as it is obsolete now.

electronicboy:
The second check was *the* check, this actually won't work right now, that if statement needs to be moved up

electronicboy:
So, basically;
That check at the bottom is what was actually doing the size check, I was wrong in saying that this won't work, it will, just not for packets which are actually "skippable", which, is probably not an issue, but something we should probably restore the correctness of

I think that we should restore papers check, and remove mojangs

lukalt:
Thanks! Looks good to me @electronicboy

PR #9149, opened 2023-04-18
Add option to flush region files on save

Redned235:
Solves #9142. 

I added a config option for this in the event the existing behavior is preferred, however I'm not entirely if this is the best option given vanilla by default will call this same method to flush the region files. Feedback welcome!

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1448,10 +1448,10 @@ index 0000000000000000000000000000000000000000..1bb16fc7598cd53e822d84b69d6a9727
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..ee430563b99e48cd722c2897e4729936457d3017
+index 0000000000000000000000000000000000000000..a33de97340f14219291c4175e9194914cdf441db
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
-@@ -0,0 +1,481 @@
+@@ -0,0 +1,482 @@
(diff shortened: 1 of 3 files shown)


electronicboy (approved):

BillyGalbreath:
Ok, I can confirm this fixes the problem as reported (fixes `/save-all flush`).

However, there still is no FileWatcher activity on the autosave interval, until I open another terminal and `ls` the region directory again. Possibly this is one of the 3 spots I mentioned above?

Redned235:
> About to give this a run to see if it fixes the problem.
> 
> Any reason you only added the extra flush to this one spot and not the other 3 spots where `RegionFileIOThread.flush();` are present? ([here](https://github.com/PaperMC/Paper/blob/master/patches/server/0018-Rewrite-chunk-system.patch#L3719), [here](https://github.com/PaperMC/Paper/blob/master/patches/server/0018-Rewrite-chunk-system.patch#L5638), and [here](https://github.com/PaperMC/Paper/blob/master/patches/server/0018-Rewrite-chunk-system.patch#L17175))

For the first two, the region files are already calling the RegionFile#close method which flushes to disk as well (first one also has no world context available as it’s static). The third method is unused within the codebase so I didn’t see it necessary to add this code there. 

I’ll double check the auto saving since I’m not sure if it flushes by default in vanilla.

Redned235:
After checking vanilla, it does not appear the auto saving flushes there either.

BillyGalbreath:
Interesting, I didn't know vanilla had auto saving now. I thought that was a Bukkit thing.

I wonder if we should add an option to flush on autosaves since we're here.

Redned235:
Yea an option for flushing on auto save wouldn't be a bad idea. On second thought I'm wondering if it's best to remove the current flushing option and have it always flush the region files when `flush` is true (i.e. in `/save-all flush`), since this is consistent with vanilla & Spigot. Then we can have a different option for flushing on auto saves. Curious on what others think about this.

Owen1212055 (approved):

PR #9152, opened 2023-04-19
add AttemptItemMergeEvent

alaajel:
*updated version of* #9151  

Basically this event is called before ItemMergeEvent which can allow to see which items are trying to merge,
unlike ItemMergeEvent which thrown on successful merged items, this event is called before checking if can be merged

## example:
throwing Dead bush with Red

```java
@EventHandler
    public void onAttemptItemMerge(AttemptItemMergeEvent e) {
        Item source = e.getEntity();
        Item target = e.getTarget();
        this.getSLF4JLogger().info("Item {} {}  is trying to merge with {} {}",
                source.getItemStack().getType(), source.getUniqueId(), target.getItemStack().getType(), target.getUniqueId());
    }
```
console:
```
[10:13:27 INFO]: [paperweight-test-plugin] Item DEAD_BUSH a474ec48-be9d-4079-85d2-a5c69d8aed03  is trying to merge with RED_SAND 28cbb557-ca5b-49b7-8524-fe2a77eb4181
[10:13:27 INFO]: [paperweight-test-plugin] Item RED_SAND 28cbb557-ca5b-49b7-8524-fe2a77eb4181  is trying to merge with DEAD_BUSH a474ec48-be9d-4079-85d2-a5c69d8aed03
[10:13:29 INFO]: [paperweight-test-plugin] Item DEAD_BUSH a474ec48-be9d-4079-85d2-a5c69d8aed03  is trying to merge with RED_SAND 28cbb557-ca5b-49b7-8524-fe2a77eb4181
[10:13:29 INFO]: [paperweight-test-plugin] Item RED_SAND 28cbb557-ca5b-49b7-8524-fe2a77eb4181  is trying to merge with DEAD_BUSH a474ec48-be9d-4079-85d2-a5c69d8aed03
```

### DIFF
--- a/patches/api/0416-Add-AttemptItemMergeEvent.patch
+++ b/patches/api/0416-Add-AttemptItemMergeEvent.patch
@@ -7,24 +7,22 @@ Basically this event fires before ItemMergeEvent which allows see what item sour
 
 diff --git a/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java b/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..eb1246cfb85c62f5a6014c155206ba248a45e968
+index 0000000000000000000000000000000000000000..84299e1255ab9dab387706a42641d0cd77110f4a
 --- /dev/null
 +++ b/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java
-@@ -0,0 +1,58 @@
+@@ -0,0 +1,48 @@
 +package org.bukkit.event.entity;
 +
 +import org.bukkit.entity.Item;
-+import org.bukkit.event.Cancellable;
 +import org.bukkit.event.HandlerList;
 +import org.jetbrains.annotations.NotNull;
 +
 +/**
 + * thrown when are Items trying to merge
 + */
-+public class AttemptItemMergeEvent extends EntityEvent implements Cancellable {
++public class AttemptItemMergeEvent extends EntityEvent {
 +
 +    private static final HandlerList handlers = new HandlerList();
-+    private boolean cancelled;
 +    private final Item target;
 +
 +    public AttemptItemMergeEvent(@NotNull Item item, @NotNull Item target) {
@@ -32,15 +30,7 @@ index 0000000000000000000000000000000000000000..eb1246cfb85c62f5a6014c155206ba24
 +        this.target = target;
 +    }
 +
-+    @Override
-+    public boolean isCancelled() {
-+        return cancelled;
-+    }
 +
-+    @Override
-+    public void setCancelled(boolean cancelled) {
-+        this.cancelled = cancelled;
-+    }
 +
 +    @NotNull
 +    @Override
--- a/patches/api/0416-Add-AttemptItemMergeEvent.patch
+++ b/patches/api/0416-Add-AttemptItemMergeEvent.patch
@@ -5,17 +5,18 @@ Subject: [PATCH] Add AttemptItemMergeEvent
 
 Basically this event fires before ItemMergeEvent which allows see what item source item want to merge with and can be canceld and class is exact copy of ItemMergeEvent
 
-diff --git a/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java b/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java
+diff --git a/src/main/java/io/papermc/paper/event/entity/AttemptItemMergeEvent.java b/src/main/java/io/papermc/paper/event/entity/AttemptItemMergeEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..eb1246cfb85c62f5a6014c155206ba248a45e968
+index 0000000000000000000000000000000000000000..9ca917a1a4e4bf3ea3250bbcaf87650ce56c8008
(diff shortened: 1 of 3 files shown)


lynxplay:
I am rather confused here, the ItemMergeEvent provides both the item entities involved in the merge as per:
https://jd.papermc.io/paper/1.19/org/bukkit/event/entity/ItemMergeEvent.html#getEntity()
https://jd.papermc.io/paper/1.19/org/bukkit/event/entity/ItemMergeEvent.html#getTarget()

Beyond that, what is the point of calling this event before even testing for mergeability ?
Items that will never merge do not "attempt" to merge either, so the event name seems off here.

ham1255:
1: ItemMergeEvent only fires when item successfully merges, not when tries to merge.

2: Reason i called it AttemptItemMergeEvent is because is tryToMerge method from game's code 
- if name is not sutible what do you suggest here?

converting to draft for now

lynxplay:
Ah, I misread your PR then. Exposing an event called when invalid items "attempt" or well, whatever they do, to merge seems interesting. I presume use-cases for this is manually merging non-mergable items ?

It would be a rather hot event when called concerning it would fire every tick for items that won't merge.

patches/api/0416-Add-AttemptItemMergeEvent.patch
```diff
@@ -0,0 +1,71 @@
...
+
+Basically this event fires before ItemMergeEvent which allows see what item source item want to merge with and can be canceld and class is exact copy of ItemMergeEvent
+
+diff --git a/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java b/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..eb1246cfb85c62f5a6014c155206ba248a45e968
+--- /dev/null
++++ b/src/main/java/org/bukkit/event/entity/AttemptItemMergeEvent.java
+@@ -0,0 +1,58 @@
++package org.bukkit.event.entity;
```
Machine-Maker: New events should go in a paper package not bukkit.

patches/server/0975-Implement-event-calling-for-AttemptItemMergeEvent.patch
```diff
@@ -0,0 +1,52 @@
...
+                 ItemEntity.merge(this, itemstack, other, itemstack1);
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index 819c9c020f4d5a1373f68850134960d24b8fc308..5ef16cc2566f9735bcfe6ceee1031d2e2e61079f 100644
+--- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -174,6 +174,7 @@ import org.bukkit.event.entity.FoodLevelChangeEvent;
+ import org.bukkit.event.entity.HorseJumpEvent;
+ import org.bukkit.event.entity.ItemDespawnEvent;
+ import org.bukkit.event.entity.ItemMergeEvent;
++import org.bukkit.event.entity.AttemptItemMergeEvent; // Paper - AttemptItemMergeEvent
```
Machine-Maker: Don't add this import, just use the fully qualified name directly.

patches/server/0975-Implement-event-calling-for-AttemptItemMergeEvent.patch
```diff
@@ -0,0 +1,52 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/item/ItemEntity.java b/src/main/java/net/minecraft/world/entity/item/ItemEntity.java
+index d47b3ac633e7936d30abfda6fc46c2c7412d76fe..2d39763929e0cf6f4009ba629753ae04057a4881 100644
+--- a/src/main/java/net/minecraft/world/entity/item/ItemEntity.java
++++ b/src/main/java/net/minecraft/world/entity/item/ItemEntity.java
+@@ -289,7 +289,7 @@ public class ItemEntity extends Entity implements TraceableEntity {
+     private void tryToMerge(ItemEntity other) {
+         ItemStack itemstack = this.getItem();
+         ItemStack itemstack1 = other.getItem();
+-
++        if (org.bukkit.craftbukkit.event.CraftEventFactory.callAttemptItemMergeEvent(other, this).isCancelled()) return; // Paper - AttemptItemMergeEvent
```
Machine-Maker: Change this to just be a line addition instead of a line replacement

Machine-Maker (changes requested):
I'm still not sold on the need for this event. You are incorrect in saying that `ItemMergeEvent` is only called after a merge. It is cancellable, and cancelling it stops the merge from happening.

ham1255:
> I'm still not sold on the need for this event. You are incorrect in saying that `ItemMergeEvent` is only called after a merge. It is cancellable, and cancelling it stops the merge from happening.

reason i said it that way looking at merge method that where event resides. but eh i agree its incorrect to say it that way.

also reason i added the event is doing merging myself *in a plugin* rather using server code, so i can change items data etc rather than relying on some runnable

electronicboy:
part of my question here is why is this event firing for items which will never be mergable? really not fond of adding spammy as all heck events for niche use-cases

PR #9155, opened 2023-04-19
[ci skip] add missing logger's javadoc

powercasgamer:


patches/api/0005-Adventure.patch
```diff
@@ -62,6 +62,9 @@ index 5fccd83117c2c153951d102756acfd4862b8c51b..acf12dfbecb9360c0921f70ed31650fe
 +        "https://jd.advntr.dev/text-serializer-gson/$adventureVersion/",
 +        "https://jd.advntr.dev/text-serializer-legacy/$adventureVersion/",
 +        "https://jd.advntr.dev/text-serializer-plain/$adventureVersion/",
++        "https://jd.advntr.dev/text-logger-slf4j/$adventureVersion/",
++        "https://javadoc.io/doc/org.slf4j/slf4j-api/1.8.0-beta4",
++        "https://javadoc.io/doc/org.apache.logging.log4j/log4j-api/2.17.1/"
```
Machine-Maker: I'm not sure these should go here, but in an earlier patch like build system changes or smth. Just the adventure-specific one should be here I think.

patches/api/0002-Build-system-changes.patch
```diff
@@ -37,16 +37,19 @@ index 4b7f005519d8856fc35a94401636ae8d9bf41f10..bb6e3afe8e8bc1a3b598b085286ffcda
...
-@@ -65,7 +77,7 @@ tasks.withType<Javadoc> {
+@@ -65,8 +77,10 @@ tasks.withType<Javadoc> {
      options.links(
          "https://guava.dev/releases/31.1-jre/api/docs/",
          "https://javadoc.io/doc/org.yaml/snakeyaml/1.33/",
 -        "https://javadoc.io/doc/org.jetbrains/annotations-java5/23.0.0/",
 +        "https://javadoc.io/doc/org.jetbrains/annotations/23.0.0/", // Paper - we don't want Java 5 annotations
          "https://javadoc.io/doc/net.md-5/bungeecord-chat/1.16-R0.4/",
++        "https://javadoc.io/doc/org.apache.logging.log4j/log4j-api/2.17.1/",
++        "https://javadoc.io/doc/org.slf4j/slf4j-api/1.8.0-beta4",
```
Machine-Maker, line 48: Not a required change, but should we make `val` fields for the versions for these 2 dependencies? just to ensure they are consistent with the dependencies added in the dependency block.

PR #9129, opened 2023-04-13
Resolve Plugin Dependency Issues, Improve PluginLoading Compat, Small Loading Issues (part 2)

Owen1212055:
Redo of #9068

Fixes: https://github.com/PaperMC/Paper/issues/8971, https://github.com/PaperMC/Paper/issues/9057

Pass dependency tree when loading, this allows us to correctly resolve dependencies when loading while plugins already in the plugin manager.

Closes: https://github.com/PaperMC/Paper/issues/9114

---
Download the paperclip jar for this pull request: [paper-9129.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/660377927.zip)

PR #9160, opened 2023-04-22
Paper Plugins Dependency Format Update

Owen1212055:
Requires #9129

Updates the dependency declaration format to generally be more structured and avoid bootstrap flags everywhere. Format for the old format will be supported for a little while.

```yml
# BEFORE:
load-before:
  - name: RequiredPlugin
    bootstrap: false
load-after:
  - name: RegistryPlugin
    bootstrap: true
  - name: OtherPlugin
    bootstrap: false
dependencies:
  - name: OtherPlugin
    required: false
    bootstrap: false
  - name: RegistryPlugin
    required: true
    bootstrap: true
```

```yml
#AFTER:
dependencies:
  bootstrap:
    # Lets say that RegistryPlugin has some registry elements that this plugin requires.
    # We don't need this during runtime, so it's not required in the server section. However
    # can be added to both if needed
    RegistryPlugin:
      load: BEFORE
      required: true
      # (this is default)
      join-classpath: true
  server:
    # Add a required "RequiredPlugin" dependency, which this plugin must load BEFORE
    RequiredPlugin:
      load: AFTER
      required: true
      # This means that this plugin won't have access to classpath
      join-classpath: false
    # Add "OtherPlugin" dependency, which the plugin must load AFTER. Is not required. WILL join classpath (by default)
    OtherPlugin:
      load: BEFORE
      required: false
      join-classpath: true
    # Load order can be omitted to cause it to be ignored... or specified by load-order: OMIT
    SpecialDependency:
      required: true
      join-classpath: true
```

### DIFF
--- a/patches/api/0008-Paper-Plugins.patch
+++ b/patches/api/0008-Paper-Plugins.patch
@@ -2017,7 +2017,7 @@ index ee100b7ad89ce1eccef0c3bc55885cd78aadd1ec..f594913e6b94f77b26a4a758c447a42d
              throw new IllegalStateException("Cannot get plugin for " + clazz + " from a static initializer");
          }
 diff --git a/src/main/java/org/bukkit/plugin/java/JavaPluginLoader.java b/src/main/java/org/bukkit/plugin/java/JavaPluginLoader.java
-index 047c0304fd617cec990f80815b43916c6ef5a94c..ab04ffe4cd05315a2ee0f64c553b4c674740eb7f 100644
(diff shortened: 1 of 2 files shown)


imDaniX:
Just wanna say that `load-order` option name feels a bit unintuitive, like `load-order: AFTER` for a reader may mean either our plugin will load after the dependency, or it requires the dependency to be loaded after the plugin.
Something like `load-dependency: BEFORE` feels more natural to me, as there's clear indication of what's being loaded. Can't think of an inverted variant to match the current behavior.
imho imho imho

Leguan16:
or just `load: [BEFORE|AFTER]`?

Machine-Maker (approved):

PR #9162, opened 2023-04-22
Add speed argument to ProjectileSource#launchProjectile

### LINKED ISSUE #7708: Allow specifying force for Entity#launchProjectile

I want to be able to shoot arrows myself as if the player shot them, so it would be nice to be able to cancel the `EntityShootBowEvent` and fire it myself, that way I can shoot the arrow (or more arrows) then have the entity object after. The problem is though that theres no way to specify force on Entity#launchProjectile, which would be needed to actually make it seem like the player sent the arrow.

I would like a new overload that allows specifying force: `Entity#launchProjectile(Class<?> projectile, float force)`, force being the 0-1 number given from the EntityShootBowEvent.

once the arrow is spawned, I can multiply it’s velocity by the “force” then set that as the new velocity

### Other

_No response_

Wyrdix:
Fix #7708 
Add a way to change speed factor

### DIFF
--- a/patches/api/0416-Add-speed-argument-to-ProjectileSource-launchProject.patch
+++ b/patches/api/0416-Add-speed-argument-to-ProjectileSource-launchProject.patch
@@ -1,0 +1,53 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: wyrdix <wyrdix@github.com>
+Date: Sat, 22 Apr 2023 17:39:23 +0200
+Subject: [PATCH] Add speed argument to ProjectileSource#launchProjectile
+
+
+diff --git a/src/main/java/org/bukkit/projectiles/ProjectileSource.java b/src/main/java/org/bukkit/projectiles/ProjectileSource.java
+index 6cb28e65b82198b19f159f8e8334208f7f62d394..cca8a25fc56998fc6ba07b0329d03bf69ac24856 100644
+--- a/src/main/java/org/bukkit/projectiles/ProjectileSource.java
++++ b/src/main/java/org/bukkit/projectiles/ProjectileSource.java
+@@ -50,5 +50,42 @@ public interface ProjectileSource {
+      */
+     @NotNull
+     public <T extends Projectile> T launchProjectile(@NotNull Class<? extends T> projectile, @Nullable Vector velocity, @Nullable org.bukkit.util.Consumer<T> function);
++
++    /**
++     * Launches a {@link Projectile} from the ProjectileSource with an
++     * initial velocity, with the initial speed supplied
++     *
++     * <br>
++     * Note that when the function is run, the entity will not be actually in
++     * the world. Any operation involving such as teleporting the entity is undefined
++     * until after this function returns.
++     *
++     * @param <T> a projectile subclass
++     * @param projectile class of the projectile to launch
++     * @param velocity the velocity with which to launch
++     * @param speed the speed at which the projectile is launched
++     * @return the launched projectile
++     */
++    @NotNull
++    public <T extends Projectile> T launchProjectile(@NotNull Class<? extends T> projectile, @Nullable Vector velocity, float speed);
++
++    /**
++     * Launches a {@link Projectile} from the ProjectileSource with an
++     * initial velocity, with the initial speed supplied, with the supplied function run before the
++     * entity is added to the world.
++     * <br>
++     * Note that when the function is run, the entity will not be actually in
++     * the world. Any operation involving such as teleporting the entity is undefined
++     * until after this function returns.
++     *
++     * @param <T> a projectile subclass
++     * @param projectile class of the projectile to launch
++     * @param velocity the velocity with which to launch
++     * @param speed the speed at which the projectile is launched
++     * @param function the function to be run before the entity is spawned
++     * @return the launched projectile
++     */
++    @NotNull
++    public <T extends Projectile> T launchProjectile(@NotNull Class<? extends T> projectile, @Nullable Vector velocity, float speed, @Nullable org.bukkit.util.Consumer<T> function);
+     // Paper end
+ }
--- a/patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
+++ b/patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
@@ -1,0 +1,153 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: wyrdix <wyrdix@github.com>
+Date: Sat, 22 Apr 2023 17:39:23 +0200
+Subject: [PATCH] Add speed argument to ProjectileSource#launchProjectile
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
+index d43859f8aa7beed82dd3a146bb1086982cd0cda7..e5eb4cdb69f761709ecf8a45183ba50f472bef8c 100644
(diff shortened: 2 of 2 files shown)


patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,153 @@
...
++        // Paper start - initial speed
++        return this.launchProjectile(projectile, velocity, speed, null);
+     }
+ 
+     @Override
+     @SuppressWarnings("unchecked")
+-    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity, org.bukkit.util.Consumer<T> function) {
++    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity, float speed, org.bukkit.util.Consumer<T> function) {
+         // Paper end - launchProjectile consumer
++        // Paper end - initial speed
```
Doc94: two ends?

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,153 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
+@@ -83,8 +83,11 @@ import org.bukkit.potion.PotionEffect;
+ import org.bukkit.potion.PotionEffectType;
+ import org.bukkit.potion.PotionType;
+ import org.bukkit.util.BlockIterator;
++import org.bukkit.util.Consumer;
+ import org.bukkit.util.RayTraceResult;
+ import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
```
Doc94: - The server patchs dont need the annotation for Null or NotNull.
- Missing // Paper for the imports..  maybe is better a inline import

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,153 @@
...
++    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity, float speed, org.bukkit.util.Consumer<T> function) {
+         // Paper end - launchProjectile consumer
+         Validate.isTrue(this.getBlock().getType() == Material.DISPENSER, "Block is no longer dispenser");
+         // Copied from BlockDispenser.dispense()
+@@ -148,7 +162,7 @@ public class CraftBlockProjectileSource implements BlockProjectileSource {
+                 b *= 1.25F;
+             }
+             // Copied from DispenseBehaviorProjectile
+-            ((net.minecraft.world.entity.projectile.Projectile) launch).shoot((double) enumdirection.getStepX(), (double) ((float) enumdirection.getStepY() + 0.1F), (double) enumdirection.getStepZ(), b, a);
++            ((net.minecraft.world.entity.projectile.Projectile) launch).shoot((double) enumdirection.getStepX(), (double) ((float) enumdirection.getStepY() + 0.1F), (double) enumdirection.getStepZ(), b*speed, a);
```
Doc94:
```suggestion
+            ((net.minecraft.world.entity.projectile.Projectile) launch).shoot((double) enumdirection.getStepX(), (double) ((float) enumdirection.getStepY() + 0.1F), (double) enumdirection.getStepZ(), b * speed, a);
```

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
Wyrdix: I didn't know how to manage this, I tried to fit the contributing rules as much as possible.
I have another start above so it seemed right to put another end

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,153 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/projectiles/CraftBlockProjectileSource.java
++++ b/src/main/java/org/bukkit/craftbukkit/projectiles/CraftBlockProjectileSource.java
+@@ -36,7 +36,10 @@ import org.bukkit.inventory.ItemStack;
+ import org.bukkit.potion.PotionData;
+ import org.bukkit.potion.PotionType;
+ import org.bukkit.projectiles.BlockProjectileSource;
++import org.bukkit.util.Consumer;
+ import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
```
Machine-Maker, line 113: Don't add this imports. You don't need to even do the nullability annotations at all, and just add the fully qualified name for the Consumer.

Machine-Maker (changes requested):
We have a patch that adds the Consumer parameter to the launchProjectile methods, this should go into that patch and that patch should be changed to be smth like expanding launchProjectile API.

Wyrdix:
I've just pushed what I knew how to fix but could you please explain how to modify a patch.
Do I need to do it by hand like those typos that I've fixed ?

Leguan16:
You can learn about how to modify patches [here](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#modifying-patches).

patches/api/0416-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,53 @@
...
++     *
++     * <br>
++     * Note that when the function is run, the entity will not be actually in
++     * the world. Any operation involving such as teleporting the entity is undefined
++     * until after this function returns.
++     *
++     * @param <T> a projectile subclass
++     * @param projectile class of the projectile to launch
++     * @param velocity the velocity with which to launch
++     * @param speed the speed at which the projectile is launched
```
Lulu13022002, line 28: This should be clear that it's the factor/percent where 100% (1F) is the vanilla speed (even if it's not true for the trident launched by a living entity but that's out of scope of this pr).

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,150 @@
...
+ import org.bukkit.util.Vector;
+ 
+ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
+     private CraftEntityEquipment equipment;
+@@ -502,13 +505,25 @@ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
+     @Override
+     public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity) {
+         // Paper start - launchProjectile consumer
+-        return this.launchProjectile(projectile, velocity, null);
++        return this.launchProjectile(projectile, velocity, 1f,null);
```
Lulu13022002, line 26: Missing space between 1F and null

patches/server/0975-Add-speed-argument-to-ProjectileSource-launchProject.patch
```diff
@@ -0,0 +1,150 @@
...
+     public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity) {
+         // Paper start - launchProjectile consumer
+-        return this.launchProjectile(projectile, velocity, null);
++        return this.launchProjectile(projectile, velocity, 1f, null);
+     }
+ 
+     @Override
+-    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity, org.bukkit.util.Consumer<T> function) {
++    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile, Vector velocity, Consumer<T> function) {
++        return this.launchProjectile(projectile, velocity, 1f, null);
```
Lulu13022002, line 128: The callback function should be redirected here.

PR #6418, opened 2021-08-18
Correct Remaining Items For Non-Recipe Crafting

tahmid-23:
Fixes #1822 (kind of)

from what I understand, the client has a RecipeManager as well which expects the item stack to follow the doubling behavior. that being said I think it's fine to deviate here since stacks should never be greater than 64. however, this will cause a slight flicker sometimes, doubling the stack and until the server later sets the amount it should really be at. the same flickering occurs when we set the remaining item.

PR #9163, opened 2023-04-22
Fix  #1822

### LINKED ISSUE #1822: Items in crafting table get duplicated if result slot is set manually
If there is no recipe for the items in the crafting grid, taking items out of the result slot of the crafting inventory should leave the items in the crafting grid alone.

After taking an item out of the result slot, the items in the crafting grid are first decremented by one and then doubled.
https://i.imgur.com/vZUlwIN.png
https://i.imgur.com/k0hneCD.png

Put a stack of 2-64 items into a crafting table
Set the result slot using either `.setResult()` or `.setItem()`
Take the item of the result slot while leaving the crafting grid alone

### Plugin list:
Only the plugin I'm using to test this

### Paper build number:
504

okx-code:
This PR is slightly different to the previous attempts #6418 and #7264 in that it more closely reflects the expected behaviour in the original bug #1822

The pull requests mentioned will still subtract an item from the ingredients when you take an item out of the result, even if the code just sets the result on PrepareItemCraftEvent. This pull request will leave the ingredients unaffected.

### DIFF
--- a/patches/server/0975-if-there-is-no-recipe-the-ingredients-should-be-unaf.patch
+++ b/patches/server/0975-if-there-is-no-recipe-the-ingredients-should-be-unaf.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: okx-code <okx@okx.sh>
+Date: Sat, 22 Apr 2023 23:19:56 +0100
+Subject: [PATCH] if there is no recipe, the ingredients should be unaffected
+ when taking an item out of the result slot
+
+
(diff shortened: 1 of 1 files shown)


kennytv (dismissed):

Machine-Maker (changes requested):
This doesn't seem to actually fix the issue?

I tested it with just this listener in a test plugin
```java
@EventHandler
public void onEvent(PrepareItemCraftEvent event) {
    event.getInventory().setResult(new ItemStack(Material.STRING, 20));
}
```

I think the problem is, that the itemstack in the craftslots is decremented inside the ResultSlot#take logic.

kennytv:
If you want to take up the PR again, we can reopen it

PR #9176, opened 2023-05-01
Add PlayerGetRespawnLocationEvent

Cryptite:
There is a non-trivial amount of work done to determine a safe location to respawn a player. This event simply provides for plugin developers to provide their own, known respawn location event to save on any performance implications with determining a respawn location.

### DIFF
--- a/patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
+++ b/patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
@@ -6,13 +6,12 @@ Subject: [PATCH] Add PlayerPreRespawnLocationEvent
 
 diff --git a/src/main/java/io/papermc/paper/event/player/PlayerPreRespawnLocationEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerPreRespawnLocationEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..cf81752c60fb1cd5a649c50acea759036ba9d094
+index 0000000000000000000000000000000000000000..bae7b9d61a061f9a4cc812c1b9107b2695dd5374
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/player/PlayerPreRespawnLocationEvent.java
-@@ -0,0 +1,68 @@
+@@ -0,0 +1,65 @@
 +package io.papermc.paper.event.player;
 +
-+import org.apache.commons.lang3.Validate;
 +import org.bukkit.Location;
 +import org.bukkit.entity.Player;
 +import org.bukkit.event.HandlerList;
@@ -25,7 +24,7 @@ index 0000000000000000000000000000000000000000..cf81752c60fb1cd5a649c50acea75903
 + * Called when a respawn event tries to determine the location of a respawn. This is called before {@link PlayerRespawnEvent}
 + */
 +public class PlayerPreRespawnLocationEvent extends PlayerEvent {
-+    private static final HandlerList handlers = new HandlerList();
++    private static final HandlerList HANDLER_LIST = new HandlerList();
 +    private final PlayerRespawnEvent.RespawnReason respawnReason;
 +    private Location respawnLocation;
 +
@@ -41,20 +40,17 @@ index 0000000000000000000000000000000000000000..cf81752c60fb1cd5a649c50acea75903
 +     */
 +    @Nullable
 +    public Location getRespawnLocation() {
-+        return this.respawnLocation;
++        return this.respawnLocation != null ? this.respawnLocation.clone() : null;
 +    }
 +
 +    /**
 +     * Sets the new respawn location.
 +     *
-+     * @param respawnLocation The exact location for the respawn.<br/>
++     * @param respawnLocation The exact location for the respawn. Can be set to null to invalidate any previously called/modified events<br/>
 +     * <b>Note:</b> If this is provided, <b>no vanilla logic that calculates "safe" respawn locations will be done.</b> It is up to you to ensure you are
 +     * providing a good respawn location for a Player.
 +     */
-+    public void setRespawnLocation(@NotNull Location respawnLocation) {
-+        Validate.notNull(respawnLocation, "Respawn location can not be null");
-+        Validate.notNull(respawnLocation.getWorld(), "Respawn world can not be null");
-+
++    public void setRespawnLocation(@Nullable Location respawnLocation) {
 +        this.respawnLocation = respawnLocation;
 +    }
 +
@@ -63,6 +59,7 @@ index 0000000000000000000000000000000000000000..cf81752c60fb1cd5a649c50acea75903
 +     *
 +     * @return the reason the event was called.
 +     */
++    @NotNull
 +    public PlayerRespawnEvent.RespawnReason getRespawnReason() {
 +        return respawnReason;
 +    }
@@ -70,11 +67,11 @@ index 0000000000000000000000000000000000000000..cf81752c60fb1cd5a649c50acea75903
 +    @NotNull
 +    @Override
 +    public HandlerList getHandlers() {
-+        return handlers;
++        return HANDLER_LIST;
 +    }
 +
 +    @NotNull
 +    public static HandlerList getHandlerList() {
-+        return handlers;
++        return HANDLER_LIST;
 +    }
 +}
--- a/patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
+++ b/patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
@@ -1,15 +1,15 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Cryptite <cryptite@gmail.com>
 Date: Mon, 1 May 2023 15:23:34 -0500
-Subject: [PATCH] Add PlayerGetRespawnLocationEvent
+Subject: [PATCH] Add PlayerPreRespawnLocationEvent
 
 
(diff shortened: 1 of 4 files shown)


molor:
Hi! Will this resolve https://github.com/PaperMC/Paper/issues/9081?
Also see [this](https://github.com/PaperMC/Paper/pull/9152#discussion_r1176839017)

Cryptite:
> Hi! Will this resolve #9081? Also see [this](https://github.com/PaperMC/Paper/pull/9152#discussion_r1176839017)

Moved! Thanks. I'm not sure this patch is really meant to solve that issue which sounds more like plugin funkyness. If this inadvertently fixes that somehow, then huzzah.

Cryptite:
I understand this could arguably called PlayerPreRespawnEvent... ish. Open to renaming suggestions if need be.

Leguan16:
i think thats a good name for it.

patches/server/0977-Add-PlayerGetRespawnLocationEvent.patch
```diff
@@ -0,0 +1,102 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
+index 4d837c1530a3031a4c2a5a39d87bd013d60e14a6..2f07456c0ff0346b621d89bbf8d84d299daba00b 100644
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
++++ b/src/main/java/net/minecraft/server/players/PlayerList.java
+@@ -835,49 +835,57 @@ public abstract class PlayerList {
+ 
+         // CraftBukkit start - fire PlayerRespawnEvent
+         if (location == null) {
+-            // boolean isBedSpawn = false; // Paper - moved up
```
Owen1212055: This has a rather difficult to read diff, can you possibly try to reduce the diff/prevent changing indentation here?

Cryptite: Fixed by not intending. Kind of atrocious but gets the job done? Open to alternative suggestions

Cryptite:
Decided to go ahead and rename it to Pre

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,69 @@
...
+Subject: [PATCH] Add PlayerPreRespawnLocationEvent
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerPreRespawnLocationEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerPreRespawnLocationEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..b321cc19d6ca57050233339ced6c84d9f793f1b6
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerPreRespawnLocationEvent.java
+@@ -0,0 +1,57 @@
++package com.destroystokyo.paper.event.player;
```
Lulu13022002: This is not the right package for the new events, should be io.papermc.paper.event

patches/server/0977-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,35 @@
...
++            if (location == null) {
++            // Paper end
+             // boolean isBedSpawn = false; // Paper - moved up
+             ServerLevel worldserver1 = this.server.getLevel(entityplayer.getRespawnDimension());
+             if (worldserver1 != null) {
+@@ -876,8 +884,8 @@ public abstract class PlayerList {
+                 blockposition = entityplayer1.getSpawnPoint(worldserver1);
+                 location = new Location(worldserver1.getWorld(), (double) ((float) blockposition.getX() + 0.5F), (double) ((float) blockposition.getY() + 0.1F), (double) ((float) blockposition.getZ() + 0.5F), worldserver1.levelData.getSpawnAngle(), 0.0F); // Paper - use world spawn angle
+             }
++            }
```
Lulu13022002: Missing a paper comment

Lulu13022002 (commented):
Does it makes sense to include the respawn reason here for more context ?

Cryptite:
> Does it makes sense to include the respawn reason here for more context ?

Sure why not

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,80 @@
...
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.event.player.PlayerRespawnEvent;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Called when a respawn event tries to determine the location of a respawn. This is called before {@link PlayerRespawnEvent}
++ */
++public class PlayerPreRespawnLocationEvent extends PlayerEvent {
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker: Use the standard for static final fields, aka all caps (e.g. `HANDLER_LIST`). Ignore bukkit's dum way of just doing `handlers`.

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,80 @@
...
++    }
++
++    /**
++     * Gets the current respawn location
++     *
++     * @return Location current respawn location
++     */
++    @Nullable
++    public Location getRespawnLocation() {
++        return this.respawnLocation;
```
Machine-Maker: This should be cloned (if it exists) because there is a setter.

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,80 @@
...
++
++    /**
++     * Sets the new respawn location.
++     *
++     * @param respawnLocation The exact location for the respawn.<br/>
++     * <b>Note:</b> If this is provided, <b>no vanilla logic that calculates "safe" respawn locations will be done.</b> It is up to you to ensure you are
++     * providing a good respawn location for a Player.
++     */
++    public void setRespawnLocation(@NotNull Location respawnLocation) {
++        Validate.notNull(respawnLocation, "Respawn location can not be null");
```
Machine-Maker: Should setting null be allowed as a way for a plugin to "undo" a previous listener?

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,80 @@
...
++
++        this.respawnLocation = respawnLocation;
++    }
++
++    /**
++     * Gets the reason this respawn event was called.
++     *
++     * @return the reason the event was called.
++     */
++    public PlayerRespawnEvent.RespawnReason getRespawnReason() {
```
Machine-Maker, line 63: Missing `@NotNull` annotation.

Machine-Maker (changes requested):

patches/api/0417-Add-PlayerPreRespawnLocationEvent.patch
Cryptite: Fair point

Lulu13022002:
This event is called in that weird place where some changes are not persisted, i think this should be stated somewhere in the documentation like the respawn event

patches/server/0977-Add-PlayerPreRespawnLocationEvent.patch
```diff
@@ -0,0 +1,35 @@
...
+--- a/src/main/java/net/minecraft/server/players/PlayerList.java
++++ b/src/main/java/net/minecraft/server/players/PlayerList.java
+@@ -835,6 +835,14 @@ public abstract class PlayerList {
+ 
+         // CraftBukkit start - fire PlayerRespawnEvent
+         if (location == null) {
++            // Paper start
++            Player respawnPlayer = entityplayer1.getBukkitEntity();
++            io.papermc.paper.event.player.PlayerPreRespawnLocationEvent preRespawnEvent = new io.papermc.paper.event.player.PlayerPreRespawnLocationEvent(respawnPlayer, reason);
++            preRespawnEvent.callEvent();
```
lynxplay, line 18: I think
```patch
diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
index 4d837c153..f9040dafe 100644
--- a/src/main/java/net/minecraft/server/players/PlayerList.java
+++ b/src/main/java/net/minecraft/server/players/PlayerList.java
@@ -833,6 +833,14 @@ public abstract class PlayerList {
         boolean isLocAltered = false; // Paper - Fix SPIGOT-5989
         // Paper end
 
+        // Paper start
+        if (location ==  null) {
+            Player respawnPlayer = entityplayer1.getBukkitEntity();
+            io.papermc.paper.event.player.PlayerPreRespawnLocationEvent preRespawnEvent = new io.papermc.paper.event.player.PlayerPreRespawnLocationEvent(respawnPlayer, reason);
+            preRespawnEvent.callEvent();
+            location = preRespawnEvent.getRespawnLocation();
+        }
+        // Paper stop
         // CraftBukkit start - fire PlayerRespawnEvent
         if (location == null) {
             // boolean isBedSpawn = false; // Paper - moved up
```

seems like a potentially cleaner patch. In general, I am unsure if the name of the event really makes sense yet, nor do I see much use in the event beyond the mentioned performance improvement that is gained by not having to compute a safe respawn location.

PR #9177, opened 2023-05-01
Add PlayerShieldDisableEvent

Cryptite:
Called whenever a player's shield is being disabled. The cooldown (default 100 ticks) is also provided to optionally be changed


patches/api/0417-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,83 @@
...
++    private boolean cancelled = false;
++
++    public PlayerShieldDisableEvent(@NotNull final Player player, @Nullable final Entity damager, int cooldown) {
++        super(player);
++        this.damager = damager;
++        this.cooldown = cooldown;
++    }
++
++    /**
++     * @return Return the Entity that caused the shield disable attempt
```
Owen1212055: Can you document what might make it be null?

kennytv (approved):

patches/api/0417-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,84 @@
...
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Thrown whenever a Player's shield is disabled
++ */
++public class PlayerShieldDisableEvent extends PlayerEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker: Use java standards for static finals, `HANDLER_LIST` preferably.

Machine-Maker (changes requested):
As I was looking over the impl for this event, I realized its REALLY close to another event, the `PlayerItemCooldownEvent`. 

Would it be better to make this event a sub-event of that event, adding the extra context for the damager? The cooldown event is called right inside addCooldown, so its not really that big of a refactor I don't think.

Cryptite:
> As I was looking over the impl for this event, I realized its REALLY close to another event, the `PlayerItemCooldownEvent`.
> 
> Would it be better to make this event a sub-event of that event, adding the extra context for the damager? The cooldown event is called right inside addCooldown, so its not really that big of a refactor I don't think.

Less code best code!

patches/api/0417-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,55 @@
...
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * Thrown whenever a Player's shield is disabled
++ */
++public class PlayerShieldDisableEvent extends PlayerItemCooldownEvent implements Cancellable {
++    private static final HandlerList HANDLER_LIST = new HandlerList();
```
Machine-Maker: My idea actually involved not having a handler list on this event. That way a listener for PlayerItemCooldownEvent would also catch this one.

patches/server/0977-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,64 @@
...
+@@ -1468,7 +1474,12 @@ public abstract class Player extends LivingEntity {
+         }
+ 
+         if (this.random.nextFloat() < f) {
+-            this.getCooldowns().addCooldown(Items.SHIELD, 100);
++            // Paper start
++            org.bukkit.entity.Entity finalAttacker = attacker != null ? attacker.getBukkitEntity() : null;
++            io.papermc.paper.event.player.PlayerShieldDisableEvent shieldDisableEvent = new io.papermc.paper.event.player.PlayerShieldDisableEvent((org.bukkit.entity.Player) getBukkitEntity(), finalAttacker, 100);
++            if (!shieldDisableEvent.callEvent()) return;
++            this.getCooldowns().addCooldown(Items.SHIELD, shieldDisableEvent.getCooldown());
```
Machine-Maker: same situation here, need to not fire the event inside addCooldown.

patches/server/0977-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,64 @@
...
++++ b/src/main/java/net/minecraft/world/entity/Mob.java
+@@ -1686,7 +1686,11 @@ public abstract class Mob extends LivingEntity implements Targeting {
+             float f = 0.25F + (float) EnchantmentHelper.getBlockEfficiency(this) * 0.05F;
+ 
+             if (this.random.nextFloat() < f) {
+-                player.getCooldowns().addCooldown(Items.SHIELD, 100);
++                // Paper start
++                io.papermc.paper.event.player.PlayerShieldDisableEvent shieldDisableEvent = new io.papermc.paper.event.player.PlayerShieldDisableEvent((org.bukkit.entity.Player) player.getBukkitEntity(), getBukkitEntity(), 100);
++                if (!shieldDisableEvent.callEvent()) return;
++                player.getCooldowns().addCooldown(Items.SHIELD, shieldDisableEvent.getCooldown());
```
Machine-Maker: You need some way to bypass the `PlayerItemCooldownEvent` that is fired inside addCooldown because we just fired it.

Machine-Maker (changes requested):

Machine-Maker (changes requested):
I should've been more explicit with what I meant by making it a sub event. Not just re-using the class, but replacing the PlayerItemCooldownEvent call with this one.

patches/server/1053-Add-PlayerShieldDisableEvent.patch
```diff
@@ -0,0 +1,108 @@
...
+@@ -1470,7 +1477,16 @@ public abstract class Player extends LivingEntity {
+         }
+ 
+         if (this.random.nextFloat() < f) {
+-            this.getCooldowns().addCooldown(Items.SHIELD, 100);
++            // Paper start - add player shield disable event
++            final org.bukkit.entity.Entity finalAttacker = attacker != null ? attacker.getBukkitEntity() : null;
++            if (finalAttacker != null) {
++                final io.papermc.paper.event.player.PlayerShieldDisableEvent shieldDisableEvent = new io.papermc.paper.event.player.PlayerShieldDisableEvent((org.bukkit.entity.Player) getBukkitEntity(), finalAttacker, 100);
++                if (!shieldDisableEvent.callEvent()) return;
```
Owen1212055: @lynxplay Seeing as though you may have reimplemented this stuff, is there a reason why `return` is being done here? (Also applies above)

lynxplay:
Rebased this once more.
I have dropped the extends for the item cooldown event.

In general, issue with that even being a child of item cooldown event is that plugins listening for the cooldown event itself might cancel a shield disable, leading to the extra "effect" of a disabled shield damage event, (no stopUse item call and no game event) which might not be intended.

Machine-Maker (approved):

PR #9179, opened 2023-05-02
add PlayerCreateMapEvent

### LINKED ISSUE #9073: Event Request: MapCreateEvent

I wanted to implement something in a plugin that would basically stop the player from creating a map in a certain area if they didn't have permission to do so. I thought "this should be simple, just cancel the MapCreate event, right?", but I found that there is actually no event like that. This means that I would have to go the complicated route by canceling the PlayerInteractEvent, and before that, I would need to do a lot of filtering to make sure that the PlayerInteractEvent is the player creating a map and not the player opening a chest, for example. I also thought canceling the MapInitialiseEvent would work, but apparently, the MapInitialiseEvent is called every time the map is opened and not just when it is created. So this means there is no event in the Bukkit, Spigot, and Paper API that deals specifically with a player initially creating a map.

Can a new event be implemented. Something like "MapCreateEvent", to allow devs to manipulate the way players interact with maps in a simple manner.

Discussed in the first paragraph...

### Other

_No response_

chmodsayshello:
This small pull request adds an event that gets fired whenever a player creates a map...

Merging this PR would result in #9073 being resolved

### DIFF
--- a/patches/api/0417-add-PlayerCreateMapEvent.patch
+++ b/patches/api/0417-add-PlayerCreateMapEvent.patch
@@ -1,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: chmodsayshello <chmodsayshello@hotmail.com>
+Date: Tue, 2 May 2023 18:11:50 +0200
+Subject: [PATCH] add PlayerCreateMapEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..ea6a9b7d98161bfaf41c397d30b23af1ba2dffa0
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java
+@@ -0,0 +1,53 @@
++package io.papermc.paper.event.player;
++
++import org.bukkit.entity.Player;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player generates a new map.
++ * This happens when right-clicking on a non filled map
++ * <p>
++ * Cancelling it will result in the map not getting generated.
++ * That way the item won't convert into a filled map...
++ */
++public class PlayerCreateMapEvent extends PlayerEvent implements Cancellable {
++    private boolean cancelled = false;
++    private final ItemStack mapStack;
++    private static final HandlerList handlers = new HandlerList();
++    public PlayerCreateMapEvent(Player player, ItemStack mapStack){
++        super(player);
++        this.mapStack = mapStack;
++    }
++
++    /**
++     * Gets the ItemStack of the Map.
++     *
++     * @return ItemStack of the newly created map
++     */
++    public ItemStack getMapStack(){
++        return mapStack;
++    }
++
++    @Override
++    public boolean isCancelled() {
++        return cancelled;
++    }
++
++    @Override
++    public void setCancelled(boolean cancel) {
++        this.cancelled = cancel;
++    }
++
++    @Override
++    public @NotNull HandlerList getHandlers() {
++        return handlers;
++    }
++
++    public static HandlerList getHandlerList(){
++        return handlers;
++    }
++}
--- a/patches/api/0417-add-PlayerCreateMapEvent.patch
+++ b/patches/api/0417-add-PlayerCreateMapEvent.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] add PlayerCreateMapEvent
 
 diff --git a/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..ea6a9b7d98161bfaf41c397d30b23af1ba2dffa0
+index 0000000000000000000000000000000000000000..7a9a77f4c7f02369b29b1d996abd168d30d6db5f
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/player/PlayerCreateMapEvent.java
 @@ -0,0 +1,53 @@
@@ -29,7 +29,7 @@ index 0000000000000000000000000000000000000000..ea6a9b7d98161bfaf41c397d30b23af1
 +public class PlayerCreateMapEvent extends PlayerEvent implements Cancellable {
 +    private boolean cancelled = false;
 +    private final ItemStack mapStack;
-+    private static final HandlerList handlers = new HandlerList();
++    private static final HandlerList HANDLER_LIST = new HandlerList();
 +    public PlayerCreateMapEvent(Player player, ItemStack mapStack){
 +        super(player);
 +        this.mapStack = mapStack;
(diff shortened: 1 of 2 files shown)


chmodsayshello:
force pushed in order to change misleading commit message

patches/server/0978-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,41 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/item/EmptyMapItem.java b/src/main/java/net/minecraft/world/item/EmptyMapItem.java
+index e92bb09dd3218d5a13e6251bddd0812acadda2be..8b07c86f22299e9c89e70603abf77ef78209ac31 100644
+--- a/src/main/java/net/minecraft/world/item/EmptyMapItem.java
++++ b/src/main/java/net/minecraft/world/item/EmptyMapItem.java
+@@ -6,6 +6,7 @@ import net.minecraft.world.InteractionHand;
+ import net.minecraft.world.InteractionResultHolder;
+ import net.minecraft.world.entity.player.Player;
+ import net.minecraft.world.level.Level;
++import io.papermc.paper.event.player.PlayerCreateMapEvent; //Paper
```
kennytv: Inline the import

patches/server/0978-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,41 @@
...
+             ItemStack itemStack2 = MapItem.create(world, user.getBlockX(), user.getBlockZ(), (byte)0, true, false);
++            //Paper start
++            PlayerCreateMapEvent playerCreateMapEvent = new PlayerCreateMapEvent((org.bukkit.entity.Player) user.getBukkitEntity(), itemStack2.asBukkitCopy());
++            world.getCraftServer().getPluginManager().callEvent(playerCreateMapEvent);
++            if(playerCreateMapEvent.isCancelled()){
++                return InteractionResultHolder.consume(itemStack);
++            }else if (!user.getAbilities().instabuild) { //moved down here so an item gets removed only if the event wasn't cancelled!
++                itemStack.shrink(1);
++            }
++            //Paper end
```
kennytv: Fix spacing after comments and brackets and before brackets

kennytv (changes requested):

patches/server/0978-add-PlayerCreateMapEvent.patch
chmodsayshello: Sorry, I might be stupid here, but I don't see whats wrong there

tal5: Use the fully qualified name in-line instead of importing (e.g. `new io.papermc.paper.event.player.PlayerCreateMapEvent()` instead of `new PlayerCreateMapEvent()`)

chmodsayshello: aaagh, thanks, my bad, let me do that really quick

patches/api/0417-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,65 @@
...
++ * Called when a player generates a new map.
++ * This happens when right-clicking on a non filled map
++ * <p>
++ * Cancelling it will result in the map not getting generated.
++ * That way the item won't convert into a filled map...
++ */
++public class PlayerCreateMapEvent extends PlayerEvent implements Cancellable {
++    private boolean cancelled = false;
++    private final ItemStack mapStack;
++    private static final HandlerList handlers = new HandlerList();
```
tal5: Should use the java standard for static final fields, I.e. `HANDLER_LIST` probably

chmodsayshello: At first, I did that, however, I noticed almost every class within that package uses calls it `handlers` instead

tal5: https://github.com/PaperMC/Paper/pull/9176#discussion_r1186716577 - that's just bukkit's weird practice

patches/server/0978-add-PlayerCreateMapEvent.patch
chmodsayshello: Should be better now, but for time being I'll leave this open

patches/server/0978-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,33 @@
...
+         if (world.isClientSide) {
+             return InteractionResultHolder.success(itemStack);
+         } else {
+-            if (!user.getAbilities().instabuild) {
+-                itemStack.shrink(1);
+-            }
+-
+             user.awardStat(Stats.ITEM_USED.get(this));
+             user.level.playSound((Player)null, user, SoundEvents.UI_CARTOGRAPHY_TABLE_TAKE_RESULT, user.getSoundSource(), 1.0F, 1.0F);
+             ItemStack itemStack2 = MapItem.create(world, user.getBlockX(), user.getBlockZ(), (byte)0, true, false);
```
chmodsayshello, line 21: Data is saved even if event is cancelled...
Thanks to @tal5

Phoenix616:
Would probably also be nice to include the original item stack here seeing as it's easily accessible? Then you could cancel the event depending on that too instead of having to try to get the correct empty map item from the player's inventory manually.

patches/api/0417-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,65 @@
...
++        super(player);
++        this.mapStack = mapStack;
++    }
++
++    /**
++     * Gets the ItemStack of the Map.
++     *
++     * @return ItemStack of the newly created map
++     */
++    public ItemStack getMapStack(){
```
Warriorrrr, line 43: Missing newline between fields and constructor, also missing space between the () and { in two spots

patches/api/0417-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,65 @@
...
++ * This happens when right-clicking on a non filled map
++ * <p>
++ * Cancelling it will result in the map not getting generated.
++ * That way the item won't convert into a filled map...
++ */
++public class PlayerCreateMapEvent extends PlayerEvent implements Cancellable {
++    private boolean cancelled = false;
++    private final ItemStack mapStack;
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    public PlayerCreateMapEvent(Player player, ItemStack mapStack){
```
Warriorrrr, line 33: It would be nice to expose the original itemstack here as phoenix commented

patches/server/0978-add-PlayerCreateMapEvent.patch
```diff
@@ -0,0 +1,33 @@
...
++            // Paper start
++            io.papermc.paper.event.player.PlayerCreateMapEvent playerCreateMapEvent = new io.papermc.paper.event.player.PlayerCreateMapEvent((org.bukkit.entity.Player) user.getBukkitEntity(), itemStack2.asBukkitCopy());
++            world.getCraftServer().getPluginManager().callEvent(playerCreateMapEvent);
++            if (playerCreateMapEvent.isCancelled()) {
++                return InteractionResultHolder.consume(itemStack);
++            } else if (!user.getAbilities().instabuild) { // Paper - moved down a couple of lines so an item gets removed only if the event wasn't cancelled!
++                itemStack.shrink(1);
++            }
++            // Paper end
```
Warriorrrr, line 30: I think it would be best to move all this logic to before the instabuild check on line 20, so that the event is called before data is created & saved. We could create a map item manually and add the next map id (without incrementing the counter incase of cancellation) to the tag, then replace that MapItem#create with MapItem#createAndStoreSavedData and pass in the itemstack from the event.

yannicklamprecht:
Isn't it more a PlayerInitializeMapEvent?
Creating is basically crafting for me.

Phoenix616:
Well, using "initialize' for that event might be confusing seeing as the map loading by the server on startup is called that too and describes a completely different behaviour.

"Create" sounds correct for me as a new map (as in the MapView, not as in an ItemStack with the kinda arbitrary type "map") is created on the server. Don't see it getting confused with crafting either, seeing as it would just use "CraftEvent" like the others if that was what the event refers too.

PR #9180, opened 2023-05-02
Restore MC-2025 patch

electronicboy:
This patch was originally dropped as it was incorrectly suspected for causing another bug, that bug was subsequently fixed, however, this patch was long forgotten about by most.

---
Download the paperclip jar for this pull request: [paper-9180.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/676366679.zip)

### DIFF
--- a/patches/server/0978-MC-2025-Save-and-load-entity-AABB-to-prevent-wobble.patch
+++ b/patches/server/0978-MC-2025-Save-and-load-entity-AABB-to-prevent-wobble.patch
@@ -1,0 +1,57 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Zach Brown <zach.brown@destroystokyo.com>
+Date: Tue, 4 Sep 2018 19:07:57 -0400
+Subject: [PATCH] MC-2025: Save and load entity AABB to prevent wobble
+
+What follows is a summarized analysis provided on the Mojira and associated subreddit by various MC community members
+who have investigated this issue and found a solution. This work is largely a result of their efforts.
+
+The underlying cause of MC-2025 is that sometimes an AABB ends up slightly smaller than the desired width. If this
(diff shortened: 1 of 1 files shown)


Spottedleaf:
These days an epsilon (1e-7 or ~2^-23) is required for intersection to take place, which is significantly larger than the reported wobble of ~2^-40

electronicboy:
Yea, I mean, I'm not convinved it's the wobble that's truly the issue these days, so, this is probably the wrong solution now; I just recalled this existing as one of many attempts at a solution issue we were seemingly facing on a server and debated on if it was worth looking into restoring it or finding another solution

patches/server/0978-MC-2025-Save-and-load-entity-AABB-to-prevent-wobble.patch
```diff
@@ -0,0 +1,57 @@
...
+index 4705d7066207250c03a5f98eef61554c901f2e35..c47dee4c76fc1767e9cc0f29946b5c2a99eedcbd 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -2303,6 +2303,12 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
+             if (freezeLocked) {
+                 nbt.putBoolean("Paper.FreezeLock", true);
+             }
++            // Paper start - MC-2025 fix - Save entity AABB and load it, floating point issues recalculating AABB can result in wobble
++            AABB boundingBox = this.getBoundingBox();
++            nbt.put("Paper.AAAB", this.newDoubleList(
```
olijeffers0n, line 35: Typo?

Owen1212055: Yeah so this was spotted so apparently this patch has never done anything anyways, lol.

PR #9183, opened 2023-05-04
fix secure profile with proxy online mode

NonSwag:
This change fixes the problem with the new **unverified chat** toast if the server is in offline-mode behind an online-mode proxy.


NonSwag:
Someone wants to review this ?

lynxplay (changes requested):
This, unless I am missing something, can just use io.papermc.paper.configuration.GlobalConfiguration.Proxies#isProxyOnlineMode to avoid this code duplication.

NonSwag:
Thanks for the tip, done!

Warriorrrr (commented):
Shouldn't this be `enforceSecureProfile && (onlineMode || proxyOnlineMode)`, right now if the enforce-secure-profile option is disabled in the properties but proxy online mode is used it'll be enforced, or is this intended?

NonSwag:
missed that sorry

MiniDigger:
please merge your patches by amending your original commit instead of creating a new one

NonSwag:
I don't know how to merge commits after they are already pushed
I've deleted the first file, so the merge would only consist out of 17e7bb7

NonSwag:
Anything else that I can improve, or does someone want to accept/close the PR?

electronicboy:
Has this been properly tested on velocity and bungee/waterfall?

electronicboy:
It's not just a value which is passed to the client which is my concern,
there's some internal logic which is backed off that

NonSwag:
I've checked everything related to that option.
It changes just the ServerData which is **only** sent to the user and the chat signing.
image

NonSwag:
> Anything else that I can improve, or does someone want to accept/close the PR?

what about now ?

NonSwag:
Hello?

Warriorrrr:
The patch as it is now won't apply since you deleted the first patch that it depended on, you need to undo the deletion of the first patch and then amend the second one into it like mini said.

PR #9185, opened 2023-05-05
Undeprecate Server#getMap

NoahvdAa:
The only way to get a map is via its id, which is not a magic value. This implementation isn't going to be changing anytime soon, so it should be undeprecated.


Machine-Maker (approved):

lynxplay (approved):

MiniDigger (approved):

yannicklamprecht (approved):

Brokkonaut:
What about MapMeta.hasMapId() and MapMeta.getMapId() ? I think they should be undeprecated too.

Owen1212055:
Removed the deprecation on those, I have not removed the setMapId method deprecation as I argue that requires further discussion.

Owen1212055 (approved):

PR #9197, opened 2023-05-08
Make ExperienceOrb entities call EntitySpawnEvent

Golfing7:
The ExperienceOrb entity currently has no universal 'spawn' event. This patch adds the event prior to the ExperienceOrbMergeEvent call so new experience orbs do not affect old ones.

### DIFF
--- a/patches/server/0989-ExperienceOrb-should-call-EntitySpawnEvent.patch
+++ b/patches/server/0989-ExperienceOrb-should-call-EntitySpawnEvent.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Golfing8 <atroo@me.com>
+Date: Mon, 8 May 2023 09:18:17 -0400
+Subject: [PATCH] ExperienceOrb should call EntitySpawnEvent
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index 9b9a9606e5be94b394606f2306e7b7a5ed781e1d..acb38f927c2418c91adf91c7b790ea6869835537 100644
+--- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -693,7 +693,10 @@ public class CraftEventFactory {
+         } else if (entity instanceof net.minecraft.world.entity.ExperienceOrb) {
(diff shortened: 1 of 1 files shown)


patches/server/0979-ExperienceOrb-should-call-EntitySpawnEvent.patch
```diff
@@ -0,0 +1,22 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -689,7 +689,10 @@ public class CraftEventFactory {
+         } else if (entity instanceof net.minecraft.world.entity.ExperienceOrb) {
+             net.minecraft.world.entity.ExperienceOrb xp = (net.minecraft.world.entity.ExperienceOrb) entity;
+             double radius = world.spigotConfig.expMerge;
+-            if (radius > 0) {
++            // Paper start - Call EntitySpawnEvent for ExperienceOrb entities.
++            event = CraftEventFactory.callEntitySpawnEvent(entity);
++            if (radius > 0 && !event.isCancelled() && !entity.isRemoved()) {
++                // Paper end
```
Leguan16: I think this is redundant here? Since the line below says `// Paper start` again

Golfing7: Yeah, I was unsure whether or not to merge the two Paper blocks or not since they have different purposes.

Leguan16: Maybe something like this would be good? 
```diff
-             if (radius > 0) {
+            // Paper start - Call EntitySpawnEvent for ExperienceOrb entities.
+            event = CraftEventFactory.callEntitySpawnEvent(entity);
+            if (radius > 0 && !event.isCancelled() && !entity.isRemoved()) {
-             // Paper start - Maximum exp value when merging - Whole section has been tweaked, see comments for specifics
+            // Maximum exp value when merging - Whole section has been tweaked, see comments for specifics
              final int maxValue = world.paperConfig().entities.behavior.experienceMergeMaxValue;
              final boolean mergeUnconditionally = world.paperConfig().entities.behavior.experienceMergeMaxValue <= 0;
 ```
 In the end I would wait for a maintainer to tell you how you should do it. Interesting PR. Looking forward to it being merged!

Owen1212055 (approved):

patches/server/0979-ExperienceOrb-should-call-EntitySpawnEvent.patch
Machine-Maker: Leave it as a separate block. I've come to really hate combining paper comment blocks cause they you have to look at the blame to know what patch adds what cause you can have like 5 patches adding methods inside one block.

Machine-Maker (approved):

PR #9209, opened 2023-05-14
Add player whitelist events

SageSphinx63920:
This pr add events when a player is being removed from the whitelist or added.


patches/api/0416-Add-whitelist-events.patch
```diff
@@ -0,0 +1,124 @@
...
++import org.jetbrains.annotations.NotNull;
++
++import java.util.UUID;
++
++/**
++ * Called when a player is being added to the whitelist
++ */
++public class PlayerWhitelistAddEvent extends Event implements Cancellable {
++
++    private static final HandlerList handlers = new HandlerList();
```
Warriorrrr: Static final fields should be uppercase, i.e. HANDLER_LIST. Same for the other event class.

patches/server/0979-Add-whitelist-events.patch
```diff
@@ -0,0 +1,31 @@
...
++++ b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+@@ -30,4 +30,20 @@ public class UserWhiteList extends StoredUserList<GameProfile, UserWhiteListEntr
+     protected String getKeyForUser(GameProfile gameProfile) {
+         return gameProfile.getId().toString();
+     }
++
++    //Paper start - Add whitelist events
++    @Override
++    public void add(UserWhiteListEntry entry) {
++        if(! new io.papermc.paper.event.player.PlayerWhitelistAddEvent(entry.getUser().getId()).callEvent()) return;
```
Warriorrrr: `if(! new` should be `if (!new`

SageSphinx63920: Done

patches/api/0416-Add-whitelist-events.patch
```diff
@@ -0,0 +1,124 @@
...
++        return HANDLER_LIST;
++    }
++}
+diff --git a/src/main/java/io/papermc/paper/event/player/PlayerWhitelistRemoveEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerWhitelistRemoveEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..1c8c392646498ab09e0721e1d00ba8e3c7c47992
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/player/PlayerWhitelistRemoveEvent.java
+@@ -0,0 +1,53 @@
++package io.papermc.paper.event.player;
```
lynxplay: I'd argue this package is not really the best package for the event.
The player package usually holds events caused by in-game players.

This does not reference such a player but rather just their resolved uuid.
The event.server package might be better ?

patches/server/0979-Add-whitelist-events.patch
```diff
@@ -0,0 +1,31 @@
...
++++ b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+@@ -30,4 +30,20 @@ public class UserWhiteList extends StoredUserList<GameProfile, UserWhiteListEntr
+     protected String getKeyForUser(GameProfile gameProfile) {
+         return gameProfile.getId().toString();
+     }
++
++    //Paper start - Add whitelist events
++    @Override
++    public void add(UserWhiteListEntry entry) {
++        if (!new io.papermc.paper.event.player.PlayerWhitelistAddEvent(entry.getUser().getId()).callEvent()) return;
```
lynxplay: This becomes kinda ugly for server owners imo.
Calling `/whitelist add <player>` will still result in a success message to the server admins even if the event was cancelled.

Obviously, this method sadly returns void so we cannot use pre-existing logic to provide feedback to the call if the removal succeeded, however imo this should be explored.

patches/api/0416-Add-whitelist-events.patch
SageSphinx63920: I talked about this before creating the event in this package. I could also call Bukkit.getPlayer which would give me the player. The event is only triggered for a player but calling the method would add some extra server load

patches/server/0979-Add-whitelist-events.patch
SageSphinx63920: So i may change the return to boolean but this would be some deeper change into the server code. Is this what you mean?

lynxplay: Nah, we cannot change that methods return type without breaking things.
We could however, create a delegating method along the lines of 
```java
public boolean tryAdd(UserWhiteListEntry entry) {
    if (event.call().isCancelled()) return false;
    add(entry);
    return true;
}
```

and then update the Whitelist command to use said method.
Just an idea, I'd appreciate feedback for other people on this too. The current logic e.g. does also only log IO errors when editing the whitelist and still results in a success message, so its already not pretty. Hence this whole suggestion might be overkill.

patches/api/0416-Add-whitelist-events.patch
```diff
@@ -0,0 +1,124 @@
...
++        this.who = who;
++    }
++
++    /**
++     * Gets the UUID of the player being added to the whitelist
++     *
++     * @return the UUID of the player being added to the whitelist
++     */
++    @NotNull
++    public UUID getUUID() {
```
NoahvdAa: The rest of the whitelist API, e.g. `Server#getWhitelistedPlayers` uses `OfflinePlayer`s. Those should probably also be used here?

NoahvdAa (commented):
👋 Congrats on making your first PR, welcome to Paper! :D

patches/api/0416-Add-whitelist-events.patch
SageSphinx63920: The method i used for triggering the event only return a gameprofile and i was unsure if i could use Bukkit.getPlayer. I would let the plugins handle how they use this uuid because creating an offline player would do extra server load

patches/server/0979-Add-whitelist-events.patch
SageSphinx63920: This could be done in the WhitelistCommand#L58

SageSphinx63920:
Thx! I've done some patching before but first time on paper

patches/api/0416-Add-whitelist-events.patch
Leguan16: I think getting the Player should be up to the plugins really.

SageSphinx63920: Yeah that's my thought

patches/api/0416-Add-whitelist-events.patch
Warriorrrr: Getting an offline player via the `CraftServer#getOfflinePlayer(GameProfile)` method won't cause any unnecessary server load since it doesn't do any lookups, you're likely confusing it with the String method that does.

patches/server/0979-Add-whitelist-events.patch
SageSphinx63920: Any more opinions here?

SageSphinx63920:
So we do have 2 opinions here: One is getting the player is up to the plugin, which would be useful for plugins that add custom behavior like an own whitelist system and the other saying we should pass an OfflinePlayer because other whitelist related API does it. Which one is the best now?

emilyy-dev:
One doesn't really dismiss the other, a whitelist plugin might be using their own system but providing the OfflinePlayer/PlayerProfile is still valuable data.

patches/api/0416-Add-whitelist-events.patch
SageSphinx63920: So i should use this `if (!new io.papermc.paper.event.player.PlayerWhitelistRemoveEvent(org.bukkit.Bukkit.getOfflinePlayer(profile.getId())).callEvent()) return;`? This creates an OfflinePlayer used by the event then

SageSphinx63920:
Is using `org.bukkit.Bukkit.getOfflinePlayer(profile.getId())` the way i should do this?

lynxplay:
> Is using `org.bukkit.Bukkit.getOfflinePlayer(profile.getId())` the way i should do this?

yes :+1:

patches/server/0979-Add-whitelist-events.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/src/main/java/net/minecraft/server/players/UserWhiteList.java b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+index a6c70decad1a496595944a5dd2aeeb387757fef3..50506c4de595623e0503a69b41675757e0814500 100644
+--- a/src/main/java/net/minecraft/server/players/UserWhiteList.java
++++ b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+@@ -30,4 +30,20 @@ public class UserWhiteList extends StoredUserList<GameProfile, UserWhiteListEntr
+     protected String getKeyForUser(GameProfile gameProfile) {
+         return gameProfile.getId().toString();
+     }
++
++    //Paper start - Add whitelist events
```
kennytv: Space after the slashes on start and end

kennytv (approved):
Just small two things. I don't think changing the whitelist command output is necessary for now

kennytv (approved):
Just small two things. I don't think changing the whitelist command output is necessary for now

patches/api/0416-Add-whitelist-events.patch
```diff
@@ -0,0 +1,122 @@
...
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player is being added to the whitelist
++ */
++public class PlayerWhitelistAddEvent extends Event implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private boolean cancel = false;
++    @NotNull private final OfflinePlayer who;
```
kennytv: Rename the player field in both events to `player`

Owen1212055:
I would say to provide the PlayerProfile the default, but then have a utility in the event to provide an OfflinePlayer if needed. That way it will only be retrieved if needed, and later calls will use the cached version.

SageSphinx63920:
Do you mean the vanilla used player profile or the paper implementation? And should I only make a getter for the profile or also one that converts the profile into an offline player

Owen1212055:
In the event, pass a paper ``PlayerProfile``, where then there would be a utility method in the event that uses the provided player profile to fetch an OfflinePlayer. 

So you'd have both

SageSphinx63920:
@Owen1212055 done

SageSphinx63920:
Is this ready to merge then?

patches/api/0416-Add-whitelist-events.patch
```diff
@@ -0,0 +1,146 @@
...
++import org.bukkit.OfflinePlayer;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player is being added to the whitelist
++ */
++public class PlayerWhitelistAddEvent extends Event implements Cancellable {
```
Machine-Maker: Should this maybe be a ServerEvent?

Machine-Maker (commented):
Is it worth adding a super event WhitelistEvent, that has the getPlayer and getPlayerProfile methods on it to reduce duplication?

Leguan16:
> maybe a single event and an action?
that's what i initially thought. Don't really see a reason for having 2 separate events. Just means unnecessary code on both sides (paper and plugin). Having a `WhitelistChangeEvent` with an `ADD` and `REMOVE` enum constant would do it imo.

SageSphinx63920:
I decided to make this use a single event for both updates

patches/api/0416-Add-whitelist-events.patch
SageSphinx63920: I think updates regarding the player should be a PlayerEvent

electronicboy (approved):
ClassNamesLikeThis, otherwise, preliminary approval

SageSphinx63920:
Isn't the event already called `PlayerWhitelistUpdateEvent`

electronicboy:
The enum

SageSphinx63920:
So i should  call it `WhitelistStatus`?

patches/server/0979-Add-whitelist-events.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/src/main/java/net/minecraft/server/players/UserWhiteList.java b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+index a6c70decad1a496595944a5dd2aeeb387757fef3..f97fda57c84d52ea1c6e9485f289cfb0b16e46da 100644
+--- a/src/main/java/net/minecraft/server/players/UserWhiteList.java
++++ b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+@@ -2,6 +2,8 @@ package net.minecraft.server.players;
+ 
+ import com.google.gson.JsonObject;
+ import com.mojang.authlib.GameProfile;
++import io.papermc.paper.event.player.PlayerWhitelistUpdateEvent;
++
```
tal5: Don't think this import should be here?

SageSphinx63920: Yeah IntelliJ auto imported. I used FQN in the calls

electronicboy:
Yes

patches/server/0979-Add-whitelist-events.patch
```diff
@@ -0,0 +1,39 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/players/UserWhiteList.java b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+index a6c70decad1a496595944a5dd2aeeb387757fef3..ff52e64a070740bde0c2aaa6561aae72a58654e3 100644
+--- a/src/main/java/net/minecraft/server/players/UserWhiteList.java
++++ b/src/main/java/net/minecraft/server/players/UserWhiteList.java
+@@ -2,6 +2,7 @@ package net.minecraft.server.players;
+ 
+ import com.google.gson.JsonObject;
+ import com.mojang.authlib.GameProfile;
++
```
tal5: There still seems to be a bit of diff here, should remove that

patches/api/0416-Add-whitelist-events.patch
Machine-Maker: Well it still isn't using the PlayerEvent, and I'm not sure that it's applicable because the Player object might not exist, for an offline player.

SageSphinx63920: There are other events in the player package not using the player Event, but I do agree the player object might not exist at this point. I can move them to the server package if this is the solution

Leguan16: Maybe just call it something like `WhitelistModifyEvent` then?

electronicboy: It can't extend PlayerEvent because nullability, server would maybe be a better place because it's more concerned about the state of a system within the server; don't think it's really a profile event, but, maybe?

SageSphinx63920: I think calling it WhitelistStateUpdateEvent would be a good name in the Server package

SageSphinx63920:
Any changes i should do?

SageSphinx63920:
Will this be merged before the 1.20 release?

Leguan16:
just keep it up to date and it will be merged eventually.

Owen1212055 (approved):
Cleaned up a bit. Now uses conversion methods for GameProfiles.

PR #9211, opened 2023-05-14
Allow LEFT_CLICK_AIR in 3.0->4.5 range of entity

Moulberry:
In Creative, the block and entity reach is the same (4.5), so the code works as is.
In Adventure, the event always fires, so the code works as is.
In Survival, the block reach is 4.5 and the entity reach is 3.0. Therefore, the existing code will not fire a LEFT_CLICK_AIR if an entity is between 3.0 and 4.5 blocks away.

To be honest, this entire method of performing a raycast to decide whether or not to fire a LEFT_CLICK_AIR is rather error-prone, even with this fix. As of 1.19.4, the vanilla client will send the block/entity interaction packets before the swing, so using that information to decide whether or not to fire LEFT_CLICK_AIR would (in my opinion) be more reliable.

---
Download the paperclip jar for this pull request: [paper-9211.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/806719986.zip)


patches/server/0979-Allow-LEFT_CLICK_AIR-in-3.0-4.5-range-of-entity.patch
```diff
@@ -0,0 +1,32 @@
...
+decide whether or not to fire LEFT_CLICK_AIR would (in my opinion) be
+more reliable.
+
+diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+index aa287d7f37f38d938d195114408cb6dbda59063d..6bff5bb0c80e6ac55fbb207a5cfa356f5395278f 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -2613,7 +2613,7 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+         // SPIGOT-5607: Only call interact event if no block or entity is being clicked. Use bukkit ray trace method, because it handles blocks and entities at the same time
+         org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.1, entity -> entity != this.player.getBukkitEntity() && this.player.getBukkitEntity().canSee(entity));
```
Owen1212055: Can this fix also be extended to this ray? Make the width 0, not 0.1.

patches/server/0979-Allow-LEFT_CLICK_AIR-in-3.0-4.5-range-of-entity.patch
```diff
@@ -0,0 +1,32 @@
...
+ 
+-        if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode
++        if (result == null || this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE || (player.gameMode.getGameModeForPlayer() != GameType.CREATIVE && result.getHitEntity() != null && origin.toVector().distanceSquared(result.getHitPosition()) > 3.0D*3.0D)) { // Paper - call PlayerInteractEvent when left-clicking on a block in adventure mode // Paper - call PlayerInteractEvent when entity distance is greater than 3.0
```
Owen1212055: Double-paper comments for this are a bit iffy. Maybe instead we could just squash this into the original patch, and make the logic a bit more readable here.

patches/server/0530-Fix-interact-event-not-being-called-sometimes.patch
```diff
@@ -0,0 +1,67 @@
...
+-        double d2 = this.player.getZ();
+-        Location origin = new Location(this.player.level.getWorld(), d0, d1, d2, f2, f1);
+-
+-        double d3 = player.gameMode.getGameModeForPlayer() == GameType.CREATIVE ? 5.0D : 4.5D;
+-        // SPIGOT-5607: Only call interact event if no block or entity is being clicked. Use bukkit ray trace method, because it handles blocks and entities at the same time
+-        org.bukkit.util.RayTraceResult result = this.player.level.getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.1, entity -> entity != this.player.getBukkitEntity() && this.player.getBukkitEntity().canSee(entity));
+-
+-        if (result == null) {
++
++        if (this.player.gameMode.getGameModeForPlayer() == GameType.ADVENTURE) {
```
Owen1212055: Avoid changing the indentation to reduce diff

3meraldK:
Tested on build-pr-jar, and works.

Moulberry:
Last commit changes Adventure mode bypass to specifically check if the hit type is block. The old patch was broken in this regard.

3meraldK:
Can you @Owen1212055 review this pull?

Owen1212055 (approved):
Tested locally, LGTM. Needs to be rebased however.

Moulberry:
PR is based once again

Moulberry:
I'm not going to continually rebase this PR.
Whenever a maintainer is ready to merge, it can be rebased.

kennytv (approved):

patches/server/0521-Fix-interact-event-not-being-called-sometimes.patch
```diff
@@ -0,0 +1,45 @@
...
++        org.bukkit.util.RayTraceResult result = this.player.level().getWorld().rayTrace(origin, origin.getDirection(), d3, org.bukkit.FluidCollisionMode.NEVER, false, 0.0, entity -> entity != this.player.getBukkitEntity() && this.player.getBukkitEntity().canSee(entity)); // Paper - change raySize from 0.1 to 0.0
+ 
+         if (result == null) {
+             CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_AIR, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);
+-        }
++        } else { // Paper start - Fix interact event not being called sometimes
++            GameType gameType = this.player.gameMode.getGameModeForPlayer();
++            if (gameType == GameType.ADVENTURE && result.getHitBlock() != null) {
++                CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_AIR, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);
```
Machine-Maker: I think this action should be `LEFT_CLICK_BLOCK` correct? If the ray trace hits a block, that means you interacted with a block?

Machine-Maker (changes requested):
Other than the one thing, this is good.

lynxplay (approved):
LGTM

Owen1212055 (approved):

Machine-Maker (approved):

PR #9212, opened 2023-05-14
Enum eradication

Machine-Maker:
Upstream has pushed a new branch containing the pr that gets rid of enums for things that shouldn’t be enums. This is Paper’s tracking of that branch.

---
Download the paperclip jar for this pull request: [paper-9212.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/695233502.zip)


Machine-Maker:
Gonna track this separately now that upstream has its own `experimental` branches.

PR #9216, opened 2023-05-18
Deprecate Material#isInteractable

Machine-Maker:
Due to an incorrect PerMaterialTest, all stairs were reporting as "interactable", when they are not. 

This PR, updates the check in Material and fixes the test to accurately test interactability.


Lulu13022002:
This behavior has been documented, so the javadoc should be updated if this behavior is no longer intended

Machine-Maker:
Oh I didn't see that it was documented... that kinda complicates things? Based on the convo I just had on discord, I'm not sure what this method's purpose is anymore. It doesn't cover things that a reasonable person would think, and then has false positives.

imDaniX:
The way I see it - the method should return `true` for blocks that are making a player raise their hand upon interacting with. And the rest of blocks seems to be this way (well, maybe aside of `MOVING_PISTON` mentioned in javadocs, since player can't do anything with it). So, as for me, stairs being interactable is rather an unexpected behavior.

Machine-Maker:
Changed this to just deprecate `Material#isInteractable`. The method is not really a useful way to learn anything about interactions with plenty of incorrect Material returning true, and plenty not returning true when it would be expected that it does.

Interactions are dependent upon world state when the interaction occurs.

Owen1212055 (approved):

lynxplay (approved):

PR #9223, opened 2023-05-22
Avoid duplicate death event call for armorstands

Lulu13022002:
This is a remake of https://github.com/PaperMC/Paper/pull/7385
This pr has been closed as duplicate but Noah hasn't fixed theses specific cases.
I have also fixed the fact that spigot doesn't emit the game event like in vanilla when an armorstand is killed in survival

### DIFF
--- a/patches/server/0253-Improve-death-events.patch
+++ b/patches/server/0253-Improve-death-events.patch
@@ -301,10 +301,10 @@ index ebe207e2c4d8b7e125fec8a5182fe4882c9339e3..8ac82d3efc0f6d8ff40226e4f8084435
      public void addAdditionalSaveData(CompoundTag nbt) {
          super.addAdditionalSaveData(nbt);
 diff --git a/src/main/java/net/minecraft/world/entity/decoration/ArmorStand.java b/src/main/java/net/minecraft/world/entity/decoration/ArmorStand.java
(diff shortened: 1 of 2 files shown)


patches/server/0255-Improve-death-events.patch
```diff
@@ -296,21 +296,49 @@ index e763421e263d9ed7a67106495fda4b74745a67fc..36f949f6a66edab8373439ff450ad34f
...
                                  this.gameEvent(GameEvent.ENTITY_DAMAGE, source.getEntity());
                                  this.lastHit = i;
                              } else {
 -                                this.brokenByPlayer(source);
 +                                org.bukkit.event.entity.EntityDeathEvent event = this.brokenByPlayer(source); // Paper
                                  this.showBreakingParticles();
+-                                this.discard(); // CraftBukkit - SPIGOT-4890: remain as this.discard() since above damagesource method will call death event
 +                                if (!event.isCancelled()) // Paper
-                                 this.discard(); // CraftBukkit - SPIGOT-4890: remain as this.discard() since above damagesource method will call death event
++                                this.kill(false); // Paper - we still need to kill to follow vanilla logic (emit game event etc...)
```
lynxplay: Concerning we touch this logic now, we can pretty up this logic to move the `this.kill(false)` line onto the same line as the if statement. Previously the discard line wasn't part of the diff and was hence not touched.

Machine-Maker (approved):

Issue #9045, opened 2023-03-25
Player#hideEntity/Player#hidePlayer and Entity#setVisibleByDefault on player should allow not removing player's name from player list


GliczDev:
### Is your feature request related to a problem?

I think it'd be very useful for only hiding players without removing them from player list

### Describe the solution you'd like.

Something like:
`player.hideEntity(plugin, player2, false)`
`player.hidePlayer(plugin, player2, false)`
`player2.setVisibleByDefault(false, false)`

### Describe alternatives you've considered.

\-

KabanFriends:
I needed this, it's quite annoying that bukkit API has extra packet sending to hide the player from tablist

J-ak-e:
I very much agree with the above mentioned suggestions.

JMBoy01:
Hi, I made some changes to the code and I think it works, but idk until i can test it... maybe anyone who can help?

XxGoldenbluexX:
I expected [Player.hideEntity()](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Player.html#hideEntity(org.bukkit.plugin.Plugin,org.bukkit.entity.Entity)) to keep player listed in tab. I tried working around this using [Player.listPlayer()](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Player.html#listPlayer(org.bukkit.entity.Player)), however you can't list hidden players. (I am getting IllegalStateException("Player cannot see other player")). If someone have a workaround until this is worked on, i would be glad to take it.

Hy3z:
> I expected [Player.hideEntity()](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Player.html#hideEntity(org.bukkit.plugin.Plugin,org.bukkit.entity.Entity)) to keep player listed in tab. I tried working around this using [Player.listPlayer()](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Player.html#listPlayer(org.bukkit.entity.Player)), however you can't list hidden players. (I am getting IllegalStateException("Player cannot see other player")). If someone have a workaround until this is worked on, i would be glad to take it.

[...]

PR #9229, opened 2023-05-26
Added boolean hideInPlayerList to hidePlayer function.

JMBoy01:
#9045 Made changes to the hidePlayer function. You can chose if the player also disappears from the player list.


lynxplay:
Presumably a partial duplicate of #8782, at least in what it aims to achive.

dawon:
> Presumably a partial duplicate of #8782, at least in what it aims to achive.
> 
> //edit
> Welcome to paper and thanks for your first PR :)

I think it is good add-on to the pr you mentioned. There is no way on said PR to hide player while retaining them in player list. You need to hide player and then reshow them in the list. This will allow to do it in one step.

But it would be good to coordinate these to prs so there are not conflicts...

JMBoy01:
Hello, is it normal that I'm still waiting on a review or isn't that going to happen since 1.20 came out in the mean time? I would just like to know.

kforbro:
> Hello, is it normal that I'm still waiting on a review or isn't that going to happen since 1.20 came out in the mean time? I would just like to know.

image

underscore11code:
Closing this as a duplicate of #8782. Thank you for your contributions.

PR #9231, opened 2023-05-28
Pathfinder hasPath should check if path is done.

romanalexander:
We should check if there is no path or the path is done navigating


kennytv:
The return value of `getCurrentPath` should also return null when this returns false

Machine-Maker (approved):

Owen1212055 (approved):

PR #9235, opened 2023-05-30
Avoid duplicate poi entries from the first section

Lulu13022002:
When the server search for nearby poi, it sometimes do some unneeded work on the center section (where the search start). This happens when the search repick the center section while searching for neighbor section. In this case duplicate entries are found on the returned list and this would cause some issues if a poi api would use theses functions. Vanilla features doesn't seems impacted except some performance implication.
To reproduce this: print the output of one affected function, place a poi block (lightning rod for example) and execute
a locate command in the same chunk section than the poi blocks, so really close.


Owen1212055 (approved):
Looking at these changes here, it makes sense.

Because the key isn't added to seen but is added to the queue when iterating over 
neighbors, it will cause it to be duplicated.

electronicboy (approved):

Machine-Maker (approved):

PR #9242, opened 2023-05-31
Several fixes and new api for experience merging/stacking

### LINKED ISSUE #9239: ExperienceOrb stacking results in XP loss on item repair
### Expected behavior

Stacked XP orbs (nbt value of `Count` > 1) give the correct amount of item repair.

### Observed/Actual behavior

Stacked orbs are treated as not stacked, resulting in the loss of `Count * Value` XP used for repairing items.

### Steps/models to reproduce

1. `/give @s minecraft:diamond_chestplate{Damage:100}`
2. `/enchant @s minecraft:mending 1`
3. Fly up and run `/summon minecraft:experience_orb ~ ~-5 ~ {Count: 2, Value: 5}`
4. Observe durability of diamond chestplate (F3 + H)
5. Collect the XP orb
6. Re-observe the durability, it should have increased by 10.

Repeat this on a vanilla server or singleplayer world, the durability will increase by 20.

### Plugin and Datapack List

none

### Paper version

f9f90791e43f396ed12be5e0e8867691fc87dbe7

### Other

This was broken by upstream in a "fix":

Original Report: https://hub.spigotmc.org/jira/browse/SPIGOT-6708
PR: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/926/overview

---

One solution is to just not change the `value` field on `ExperienceOrb`. But that causes the above bug. So an alternative is needed to expose whatever that issue is asking for. Maybe @Phoenix616 can chime in.

Could alternatively reset the value field after handling the repair logic. I don't really like this because orb.getValue() doesn't accurately track the value of the orb.

### LINKED ISSUE #9240: ExperienceOrbMergeEvent not called for all merges
### Expected behavior

ExperienceOrbMergeEvent should be fired for all XP orb merges.

### Observed/Actual behavior

the event is only fired for merges that occur as a result of one orb being added to the world.

### Steps/models to reproduce

0. plugin listening to ExperienceOrbMergeEvent
1. Disable all forms of entity spawning, gamerules, etc.
2. Spawn an xp orb: `/summon minecraft:experience_orb ~ ~-5 ~ {Value: 5}`
3. Spawn 39 other entities (orbs only merge if their entity id difference % 40 == 0)
4. Spawn another xp orb with the same value.
5. Verify they merged with `/execute as @e[type=experience_orb,distance=..10] run say hi` (should only print 1 "hi")

### Plugin and Datapack List

none

### Paper version

f9f90791e43f396ed12be5e0e8867691fc87dbe7

### Other

_No response_

### LINKED ISSUE #9241: ExperienceOrb count is not exposed in API

Cannot get or set the count of an experience orb.

Adding a get/set method pair for the count property

NMS/reflection

### Other

Related: https://github.com/PaperMC/Paper/issues/9239

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/9239
Resolves https://github.com/PaperMC/Paper/issues/9241
Fix (partially): https://github.com/PaperMC/Paper/issues/9240

Probably should be merged with https://github.com/PaperMC/Paper/pull/7382


electronicboy (approved):

Phoenix616 (changes requested):
This does not address the fact that the recursive method will assume that the experience orb still has the full value even when part of the experience of the orb was already used. (This call: `int j = Math.min(this.xpToDurability(this.value), itemstack.getDamageValue());` uses the orb's value to get the durability and not the left-over value. I think simply replacing `this.value` with `k` in that line should solve that)

Machine-Maker:
@Phoenix616 HMMM, this might be a vanilla bug? cause it does exactly that in vanilla. 
image

I will do some testing in vanilla right now.

Phoenix616:
@Machine-Maker Probably, yeah.

Machine-Maker:
Ok, added a fix for that issue and reported it to the mojira ([MC-263123](https://bugs.mojang.com/browse/MC-263123))

electronicboy (approved):

lynxplay (approved):

PR #9256, opened 2023-06-03
fix secure profile with proxy online mode

NonSwag:
This pr is same as the last one in #9183
I messed up with the commit, so i made this one


Warriorrrr (changes requested):
Seems like the code here is from an earlier state of the branch since https://github.com/PaperMC/Paper/pull/9183#pullrequestreview-1415895257 applies again.

NonSwag:
Oh, sorry, that's my fault.
I forgot to check the commit.

NonSwag:
Finally, rebased the code
totally forgot about this pr sorry

NonSwag:
bump

patches/server/0986-fix-secure-profile-with-proxy-online-mode.patch
```diff
@@ -0,0 +1,22 @@
...
+index a7e133f3495e9132a5fdae2c24f225e7b026295a..03f895c65d3d516bfdc40ef323794d2d567ca594 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -637,7 +637,10 @@ public class DedicatedServer extends MinecraftServer implements ServerInterface
+ 
+     @Override
+     public boolean enforceSecureProfile() {
+-        return this.getProperties().enforceSecureProfile && this.getProperties().onlineMode;
++        // Paper start - fix secure profile with proxy online mode
++        var globalConfiguration = io.papermc.paper.configuration.GlobalConfiguration.get();
```
Warriorrrr, line 17: This globalConfiguration variable isn't needed

Warriorrrr (commented):
Might be nice to put this behind a config option so that it isn't immediately enabled, proxies and chat signing is still a bit iffy

patches/server/0986-fix-secure-profile-with-proxy-online-mode.patch
NonSwag: What would you name the config entry for this?
And I am using this for a few months now and have no problems at all behind my velocity proxy (and without) and the chat is working as it should even with a custom chat format and (when modified) with modified messages.

NonSwag:
Failed to check out the right branch... again...
I'll close this and move it to a new pr... again...

PR #9202, opened 2023-05-11
ignoring moved too quickly message for teleports

NonSwag:
Almost all the time, a player gets teleported over a bigger distance, the player gets flagged for [moving too quickly](https://github.com/PaperMC/Paper/assets/54660361/35844ad7-5717-426f-826d-5b004cde70b3).
This change simply allows the movement after a teleport it.
_I could not find a way to exploit this, but I am no cheater, so I don't really know what's possible there._

PR #9257, opened 2023-06-03
ignoring moved too quickly message for teleports

NonSwag:
I messed up my local branch commit history, so this is a copy of #9202

### DIFF
--- a/patches/server/0986-ignoring-moved-too-quickly-message-for-teleports.patch
+++ b/patches/server/0986-ignoring-moved-too-quickly-message-for-teleports.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: NonSwag <mrminecraft00@gmail.com>
+Date: Sat, 3 Jun 2023 23:29:51 +0200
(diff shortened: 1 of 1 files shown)


Machine-Maker (commented):
I don't see the `justTeleported` field getting reset to false anywhere that's gonna happen every tick. I'm not sure its appropriate for use here, can you explain about your reasoning for using it?

NonSwag:
If I recall it correctly, it gets reset after the next movement packet comes in.
I've tested it with invalid movements and with teleports...
The teleports from the server did not get flagged as invalid movement, but the speed and clip cheats did.
But I will double-check for you and send the corresponding code.

Owen1212055 (commented):
This generally looks fine, but, I am a little iffy since this seems to be an "anti-cheat" feature here.. so this may be a little delicate to change here.

Someone a little more familiar with this area should probably give a second opinion (anti cheat developers or what not?)

NonSwag:
Is there any update about this?

Owen1212055:
The movement netcode is kinda delicate, hoping to get some other testing here.

NonSwag:
I think this is obsolete (for developers) now with the introduction of the new PlayerFailMoveEvent

PR #9258, opened 2023-06-04
Fix some issues with the retract event & sticky pistons

Machine-Maker:
Reported by @Lulu13022002 and [SPIGOT-2677](https://hub.spigotmc.org/jira/browse/SPIGOT-2677).

When cancelling the retract event for sticky pistons, the piston head did not stay extended but the "stickyied" blocks were left behind. This doesn't match the normal piston behavior of keeping the head extended.

Root of the issue is that sticky events are called later due to need to collect the blocks being moved. 

I didn't find any issues, and observers function correctly, not updating when the retract event is cancelled, but this could break some obscure redstone mechanic I'm not aware of.

### DIFF
--- a/patches/server/0680-Fix-sticky-pistons-and-BlockPistonRetractEvent.patch
+++ b/patches/server/0680-Fix-sticky-pistons-and-BlockPistonRetractEvent.patch
@@ -23,11 +23,16 @@ to WHY that was ever intended to be the case.
 Instead we opt to remove the check entirely so that the event fires for
 all piston types.
 
+Also fixes some issues with cancelling the event for sticky pistons. It
+required separate logic for restoring the pre-cancelled state of the piston
+head being extended as the sticky event is delayed from the non-sticky event
+due to the need to collect all the blocks that might be moved.
+
 Co-authored-by: Zach Brown <zach@zachbr.io>
 Co-authored-by: Madeline Miller <mnmiller1@me.com>
 
 diff --git a/src/main/java/net/minecraft/world/level/block/piston/PistonBaseBlock.java b/src/main/java/net/minecraft/world/level/block/piston/PistonBaseBlock.java
-index 0dbdcd443fe8a299119ea5ba3acb1a0412856184..3dfe79684f662ac7cae4583bfe03a633438b4df7 100644
+index 0dbdcd443fe8a299119ea5ba3acb1a0412856184..b9becbcc13569596fb18eb3c30a64e1df6f3fe8f 100644
 --- a/src/main/java/net/minecraft/world/level/block/piston/PistonBaseBlock.java
 +++ b/src/main/java/net/minecraft/world/level/block/piston/PistonBaseBlock.java
 @@ -159,15 +159,15 @@ public class PistonBaseBlock extends DirectionalBlock {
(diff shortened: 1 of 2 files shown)


patches/server/0768-Fix-sticky-pistons-and-BlockPistonRetractEvent.patch
```diff
@@ -69,17 +78,66 @@ index 24a822ade17849a083161216c184f02c30b5cb1f..f456ad8a74464414f69b616a48ee9a2c
...
+             world.setBlock(blockposition1, Blocks.AIR.defaultBlockState(), 20);
+         }
+ 
+@@ -379,6 +405,13 @@ public class PistonBaseBlock extends DirectionalBlock {
+             world.getCraftServer().getPluginManager().callEvent(event);
+ 
+             if (event.isCancelled()) {
++                // Paper start
++                if (!retract) {
++                    world.removeBlockEntity(pos);
```
Lulu13022002: Is there a reason to remove the tile earlier, the setBlock call will also remove this tile in the finalTick of the moving tile and since the blockstate has already changed it shouldn't trigger other updates

patches/server/0768-Fix-sticky-pistons-and-BlockPistonRetractEvent.patch
```diff
@@ -69,17 +78,66 @@ index 24a822ade17849a083161216c184f02c30b5cb1f..f456ad8a74464414f69b616a48ee9a2c
...
++                    world.setBlock(pos, originalPistonBlockState, Block.UPDATE_INVISIBLE | Block.UPDATE_KNOWN_SHAPE | 1024);
```
Lulu13022002: I though the UPDATE_INVISIBLE flag was only for the client and doesn't do anything on the serverside

patches/server/0768-Fix-sticky-pistons-and-BlockPistonRetractEvent.patch
```diff
@@ -69,17 +78,66 @@ index 24a822ade17849a083161216c184f02c30b5cb1f..f456ad8a74464414f69b616a48ee9a2c
...
 +                        // Paper end
                          world.removeBlock(pos.relative(enumdirection), false);
                      }
                  }
+@@ -320,10 +343,13 @@ public class PistonBaseBlock extends DirectionalBlock {
+         }
+     }
+ 
+-    private boolean moveBlocks(Level world, BlockPos pos, Direction dir, boolean retract) {
++    private boolean moveBlocks(Level world, BlockState originalPistonBlockState, BlockPos pos, Direction dir, boolean retract) { // Paper
```
Lulu13022002: Maybe a small comment about the fact that retract = extend here

Lulu13022002 (commented):
See minor comments below but this patch can creates a weird piston when i try to pull two slime blocks on the ground

Machine-Maker:
Rebased for 1.20.4. Tested by @Lulu13022002

Thorinwasher:
Found a bug: When the piston does not have a block which it's moving during retraction, the event PistonRetractEvent will show an inverted facing compared to when it does move a block.

Krakenied:
> Found a bug: When the piston does not move a block during retraction, the event PistonRetractEvent will show an opposite facing compared to when it does move a block.

These two should use to `enumdirection.getOpposite()` to fix that:
https://github.com/PaperMC/Paper/pull/9258/files#diff-e7381a22d3936796a90f4898d99ec21d6099cc8e49f45267c11841c942feb240R78
https://github.com/PaperMC/Paper/pull/9258/files#diff-e7381a22d3936796a90f4898d99ec21d6099cc8e49f45267c11841c942feb240R108

However I'm not sure if it is not actually the correct behavior.

Krakenied:
https://github.com/PaperMC/Paper/assets/46192742/f33fb12b-a480-43f4-a5ba-3323d32ffd25

Machine-Maker:
Ok, I think the correct solution is for getDirection to return the same value regardless of if its an extend or retract, just returning the direction the piston head is relative to the block itself.

Krakenied:
https://github.com/PaperMC/Paper/assets/46192742/f9ac2bb5-1cab-4231-bb15-99111a03113f

works correctly

WarnDa:
Waiting for this PR, working well on my tests ! :D

NoltoxGit:
I'm waiting this PR too :)

Krakenied:
> Ok, I think the correct solution is for getDirection to return the same value regardless of if its an extend or retract, just returning the direction the piston head is relative to the block itself.

However it will break plugins behavior on Spigot. Especially plugins tracking sticky pistons blocks movement as on Spigot the method (for sticky pistons) returns piston head direction for extend and the opposite for retract. It's not ideal though as well because currently on Spigot normal piston event getDirection method returns piston head direction for both extend and retract. There is one more inconsistency on Spigot too - the sticky piston doesn't fire retract event if there is no blocks attached to it.

PR #9261, opened 2023-06-05
Feature: FOV Scale

Moulberry:
Modifies the player's fov by a factor
The multiplier is affected by the client Accessibility option 'FOV Effects'

I'm not sure about the approach of exposing a variable in ClientboundPlayerAbilitiesPacket and then mutating it. Alternatively the onUpdateAbilities method could clone Abilities before constructing the packet. Let me know if that approach is preferred

### DIFF
--- a/patches/server/0986-FOV-Scale-Implementation.patch
+++ b/patches/server/0986-FOV-Scale-Implementation.patch
@@ -17,19 +17,9 @@ However, we want our FOV scale to work additively with the normal FOV scale caus
 Therefore, we replace movementSpeed with walkingSpeed, arriving at the final calculation:
 walkingSpeed = walkingSpeed / (fovScale * 2.0f - 1.0f)
 
-diff --git a/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java b/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
-index a20bcf5877106ae00c83cf376a0dd16016cd1c7c..f262347fa9de6f3f9d118a1ffa810f7292558509 100644
---- a/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
-+++ b/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
-@@ -14,7 +14,7 @@ public class ClientboundPlayerAbilitiesPacket implements Packet<ClientGamePacket
-     private final boolean canFly;
-     private final boolean instabuild;
-     private final float flyingSpeed;
--    private final float walkingSpeed;
-+    public float walkingSpeed; // Paper - expose for ServerPlayer#onUpdateAbilities
- 
-     public ClientboundPlayerAbilitiesPacket(Abilities abilities) {
-         this.invulnerable = abilities.invulnerable;
+== AT ==
+public-f net.minecraft.network.protocol.game.ClientboundPlayerAbilitiesPacket walkingSpeed
+
 diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
 index 98df2463bf41fc736aa6a2b6ddf89e5abde6eb39..d9397ebbc2e34bed77b87d7570f8924230e1965a 100644
 --- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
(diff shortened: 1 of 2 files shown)


patches/server/0986-FOV-Scale-Implementation.patch
```diff
@@ -0,0 +1,73 @@
...
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1877,7 +1877,11 @@ public class ServerPlayer extends Player {
+     @Override
+     public void onUpdateAbilities() {
+         if (this.connection != null) {
+-            this.connection.send(new ClientboundPlayerAbilitiesPacket(this.getAbilities()));
++            // Paper start - use fov scale
++            ClientboundPlayerAbilitiesPacket abilitiesPacket = new ClientboundPlayerAbilitiesPacket(this.getAbilities());
++            abilitiesPacket.walkingSpeed = abilitiesPacket.walkingSpeed / (this.getBukkitEntity().getFovScale() * 2 - 1);
```
MiniDigger: what is this formular based on? 
also need to mkae sure we dont divide by 0 here, fov scale of 0.5 would do that

patches/server/0986-FOV-Scale-Implementation.patch
```diff
@@ -0,0 +1,73 @@
...
+diff --git a/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java b/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
+index a20bcf5877106ae00c83cf376a0dd16016cd1c7c..f262347fa9de6f3f9d118a1ffa810f7292558509 100644
+--- a/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
++++ b/src/main/java/net/minecraft/network/protocol/game/ClientboundPlayerAbilitiesPacket.java
+@@ -14,7 +14,7 @@ public class ClientboundPlayerAbilitiesPacket implements Packet<ClientGamePacket
+     private final boolean canFly;
+     private final boolean instabuild;
+     private final float flyingSpeed;
+-    private final float walkingSpeed;
++    public float walkingSpeed; // Paper - expose for ServerPlayer#onUpdateAbilities
```
MiniDigger: cant we use an AT here? generally bit meh to mutate but I see why you did it this way and cant think of anything much nicer, maybe a new ctor?

MiniDigger (changes requested):

patches/server/0986-FOV-Scale-Implementation.patch
Moulberry: The client calculates the FOV scale as:
`(movementSpeed / walkingSpeed + 1.0f) / 2.0f`

The formula here is an algebraic rearrangement to calculate the FOV scale in terms of the walkingSpeed.
I'll add an explanation of this into the patch description.

Correct me if I'm wrong, but division by 0 should be OK because this is a float, so the result is well defined to be Infinity.
Putting Infinity into the client's FOV calculation produces 0.5 as expected.

patches/server/0986-FOV-Scale-Implementation.patch
Moulberry: I have updated the PR to use an AT.

As I mentioned in the PR description, an alternative method is to clone the net.minecraft.world.entity.player.Abilities struct, mutate it and then construct the packet from it. Would you prefer that to the AT?

lynxplay: I don't know with mojang moving packets to records more and more, I honestly think your later suggestion would fit better than an AT. Cloning the abilities struct, or honestly, even just mutating the Abilities instance before constructing the packet and mutating it back to its original state after the packet construction would probably be a tad more future proof when they inevitably touch that packet and make it a record (all its fields are final already so it would be rather easy for them)

patches/server/0986-FOV-Scale-Implementation.patch
```diff
@@ -0,0 +1,86 @@
...
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1877,7 +1877,20 @@ public class ServerPlayer extends Player {
+     @Override
+     public void onUpdateAbilities() {
+         if (this.connection != null) {
+-            this.connection.send(new ClientboundPlayerAbilitiesPacket(this.getAbilities()));
++            // Paper start - use fov scale
++            ClientboundPlayerAbilitiesPacket abilitiesPacket = new ClientboundPlayerAbilitiesPacket(this.getAbilities());
++            float fovScale = this.getBukkitEntity().getFovScale();
++            if (fovScale < -1E24f) {
```
lynxplay, line 39: Could technically reduce this to something like 
```java
if (Math.abs(fovScale) > 1E24F) {
    abilitiesPacket.walkingSpeed = Math.signum(fovScale * abilitiesPacket.walkingSpeed) * 1E-24F;
}
```
to avoid the duplicate code here.

kennytv:
Feels a bit too dank and like abusing some internal data handling to add this as API

PR #9269, opened 2023-06-07
1.20: Fix API compile issues

NoahvdAa:
Title. Makes the API compile again.


kennytv:
All the stuff in Material isn't actually experimental anymore

NoahvdAa:
Good catch, will update that.

PR #9270, opened 2023-06-07
Prevent some desync for poi and pistons

### LINKED ISSUE #5561: Pushing rails on slime blocks causes them to pop off
### Expected behavior

You can place rails on slime blocks (with slime block at the end of them too) and push them with a piston just fine.

### Observed/Actual behavior

Everything moves, but then all but last rail pops off.

### Steps/models to reproduce

Video from original reporter on Discord: https://cdn.discordapp.com/attachments/289587909051416579/837996727217815552/video.mp4

Minimal reproduction image:

You can put more rails. If you power the piston, all rails apart from the one that's against the slime block wall will get popped off.

### Plugin list

None

### Paper version

This server is running Paper version git-Paper-638 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)
Previous version: git-Paper-"4047cffca" (MC: 1.16.5)

### Agreements

### Additional info
Originally reported on Discord: https://discord.com/channels/289587909051416579/289587909051416579/837996731735212062

This is a regression caused by Piston Duping patch - disabling the fix causes it to work like in Vanilla.

Legitimate uses it breaks include flying machines.
Unsupported uses it breaks include perma block breaking exploits.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/5561
This fix two issues:
- poi desync when the BlockPhysicsEvent is cancelled
- the set air to prevent further updates doesn't have the UPDATE_MOVE_BY_PISTON flag and so the rail will drop when removed even with all the previous flags.

### DIFF
--- a/patches/server/0409-Fix-piston-physics-inconsistency-MC-188840.patch
+++ b/patches/server/0409-Fix-piston-physics-inconsistency-MC-188840.patch
@@ -59,7 +59,7 @@ index 4f6b7ed4809086811f0460544ba2ec0606f0f5da..9abae63e06c1dde9b8434d32bac87988
 +                }
 +                world.setBlockEntity(MovingPistonBlock.newMovingBlockEntity(blockposition3, iblockdata2, allowDesync ? list1.get(k) : iblockdata1, dir, retract, false));
 +                if (!allowDesync) {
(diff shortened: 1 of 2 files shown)


Machine-Maker (approved):
Confirmed this fixes the rail issue and doesn't re-enable tnt duping and the setting is still respected correctly.

Issue #8948, opened 2023-03-07
Replace Timings with Spark


kashike:
Based on an internal discussion in August of 2022, we are now entering the planning stages to remove the Timings functionality from the Paper server software, and to include [Spark](https://spark.lucko.me/) in the distribution instead. More information will be shared once plans have solidified, this issue is being opened now to begin public tracking of our plans.

Some of the reasons for this are as follows:
- Timings has been unmaintained for multiple years now, with only very few additional timings categories added since. Because of this, it is not ideal for actually identifying causes of lag in specific plugins or parts of the server outside of simple categories like synchronous chunk loads.
- Timings are difficult to read for people that are new to it, and there are no short and easy to digest tutorials on it either. Spark is a lot less confusing in terms of UI and has more beginner-friendly documentation.
- Enabling timings can sometimes have non-negligible hits on server performance, whereas Spark is lighter and also has an asynchronous profiling mode.

While all of this is still open for discussion, here are potential todos:
- [x] https://github.com/PaperMC/Paper/pull/8949 ?
- [x] Add runtime nags for uses of Timings API: https://github.com/PaperMC/Paper/commit/32875b429a0ad9d0b9c8de63f424d2f0fde545e5
- [ ] Figure out how exactly to bundle or include spark
- [ ] Make sure spark properly replaces the useful functionality timings had

hendoulmou9:
Figure out how exactly to bundle or include spark: for example, we can create a new timing using Spark's Spark.createTask() method, which takes a string as a parameter representing the name of the timing task. We then start the timing using the startTiming() method, and run some code. Finally, we stop the timing using the stopTiming() method.

Phoenix616:
I feel like one requirement for a replacement should be to have equally user-friendly names for entries like Timings has right now which is a major benefit of it over Spark: That someone who doesn't know anything about internals is able to quickly spot what is wrong. 

Having just the (sometimes confusing or even wrong) Mojang names of classes/methods doesn't really help people that are not developers and right now Timing names also provide additional context like the type of entity or the default tick counts that are currently not accessible via Spark or (easy) other means.

regulad:
[...]

PR #9276, opened 2023-06-08
Bundle Spark

TheRealRyGuy:
Fully aware this will need to be reapplied once 1.20 is stable, just figured I'd push now to show the diff in case this is wanted 

Handles another part of #8948 by downloading and registering Spark as a plugin on startup, has been minimally tested so is marked as a draft.


Owen1212055:
This is very similar to purpur's implementation, I am not sure if we are wanting to simply add a new plugin like this, but more merge it into our actual project properly.

TheRealRyGuy:
Didn't realize downstream already had a Spark patch - adding it as another plugin was most convenient, but API is very limited for replacing some Timings functionality from what I can see, so merging it may be best

PR #9288, opened 2023-06-09
Add sniffer egg api

Lulu13022002:
Call the BlockGrowEvent and BlockFadeEvent during hatch process for the sniffer eggs
Add the Egg spawn reason for the new baby sniffer
Call EntityDropItemEvent when the sniffer lay egg and drop its itemstack
Customize hatch time of the sniffer egg for the boosted or regular tick times


kennytv (approved):

lynxplay (changes requested):
Cancelling either of these events completely breaks the sniffer egg block as `onPlace` 
is no longer called and hence the tick method is never called again either, leaving the sniffer egg eternally stuck in its state.

We need to reschedule a tick here (either manually (probably tho to avoid block updates) or by updating the block to itself forcefully).

Lulu13022002:
You're right i have updated the patch, i need to reschedule the tick manually otherwise the particles/game event will shown when cancelled

Owen1212055:
Can you do a diff on change? (add paper comments where this is added, just incase this ever changes)

Summerfeeling:
It would be nice to be able to configure the hatch time in general (int baseDelay = hatchBoost(world, pos) ? 12000 : 24000;) either via an event or via the world configuration

Lulu13022002:
I don't really want an event that specific, now that the grow/fade are called anyway but i'm okay with a world config if you have some idea of its name/section in the file feel free. In fact i have modified those to test the events (because otherwise it's way too long) so i can understand the need.

Summerfeeling:
Maybe something like entities.sniffer.hatch-time and entities.sniffer.boosted-hatch-time? It would also be cool if you could reduce the eight minute cooldown after sniffing out a seed (entities.sniffer.dig-cooldown?)

patches/server/0859-Add-missing-BlockFadeEvents.patch
```diff
@@ -20,3 +20,34 @@ index 834c5e3fbff3819f3f72e95a1072d9b9e57f25b3..294d22b6b27e96b59c77527efcfefa94
...
+--- a/src/main/java/net/minecraft/world/level/block/SnifferEggBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/SnifferEggBlock.java
+@@ -54,12 +54,26 @@ public class SnifferEggBlock extends Block {
+         return this.getHatchLevel(state) == 2;
+     }
+ 
++    // Paper start
++    private void rescheduleTick(ServerLevel world, BlockPos pos) {
++        int baseDelay = hatchBoost(world, pos) ? 12000 : 24000;
++        world.scheduleTick(pos, this, (baseDelay / 3) + world.random.nextInt(300));
```
Machine-Maker: Either call this method in onPlace or put a comment there to conflict if something changes so this updates.

patches/server/0974-Call-BlockGrowEvent-for-missing-blocks.patch
```diff
@@ -0,0 +1,39 @@
...
+     public void tick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
+         if (!this.isReadyToHatch(state)) {
++            // Paper start
++            if (!org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockGrowEvent(world, pos, state.setValue(HATCH, Integer.valueOf(this.getHatchLevel(state) + 1)), 2)) {
++                rescheduleTick(world, pos);
++                return;
++            }
++            // Paper end
+             world.playSound((Player)null, pos, SoundEvents.SNIFFER_EGG_CRACK, SoundSource.BLOCKS, 0.7F, 0.9F + random.nextFloat() * 0.2F);
+-            world.setBlock(pos, state.setValue(HATCH, Integer.valueOf(this.getHatchLevel(state) + 1)), 2);
```
Machine-Maker: So this is more something I'm just noticing. All the other places the BlockGrowEvent is handled, the setBlock call is still there, it just happens after the event is handled. So are you sure we should be removing these setBlock calls? It also results in different update flags being used for the setBlock as the flag param of `handleBlockGrowEvent` isn't used anywhere.

Machine-Maker (changes requested):

patches/server/0974-Call-BlockGrowEvent-for-missing-blocks.patch
Lulu13022002: Not all the places (if you see the setBlock are not in the same position for the other cases) but yes i'm aware this is something everywhere in the server when the event require a blockstate that is forcibly placed using blockstate#update, the updates are always mangled because the blockstate almost never use the internal flag (same using the BlockStateListPopulator or the bukkit capture system). Owen started a cleanup of the BlockPlaceEvent logic around there for spigot (also add the support to tile entity in the tree). Then i might pr a patch that i wasn't planning to pr that soon but that will restore the right update flag and the right order of updates during the capture of bukkit (i redo the capture system here).

Lulu13022002:
I have added the config option, i will not allow to change the sniffer cooldown since this is done using the memory key. Sadly spigot doesn't cover this key yet but you can look at this pr: https://github.com/PaperMC/Paper/pull/5605/files#diff-95136ad09028dbf026de8d80d4a480d7a98c3cafa77dfc0f00b939654e20e916R181 (i have no idea why it's all void here)

patches/server/0974-Call-BlockGrowEvent-for-missing-blocks.patch
Machine-Maker: Ok, sounds good. I couldn't actually find a difference in behavior between the two flags, but I maybe was just trying the wrong things. Like they both broke nether portals and fired observers.

Summerfeeling:
Thanks a lot! I really appreciate your efforts

patches/server/0855-Add-missing-BlockFadeEvents.patch
```diff
@@ -20,3 +20,34 @@ index 834c5e3fbff3819f3f72e95a1072d9b9e57f25b3..294d22b6b27e96b59c77527efcfefa94
...
+index 1e115adfcfee518667559100d04050f5e71c8a23..8f08a977f7433b8237aff26c6d03d821a843c261 100644
+--- a/src/main/java/net/minecraft/world/level/block/SnifferEggBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/SnifferEggBlock.java
+@@ -54,12 +54,26 @@ public class SnifferEggBlock extends Block {
+         return this.getHatchLevel(state) == 2;
+     }
+ 
++    // Paper start
++    private void rescheduleTick(ServerLevel world, BlockPos pos) {
++        int baseDelay = hatchBoost(world, pos) ? 12000 : 24000;
```
Machine-Maker: Use the field constants here instead. So `12000 -> BOOSTED_HATCH_TIME_TICKS` and `24000 -> REGULAR_HATCH_TIME_TICKS`.

patches/server/0005-Paper-config-files.patch
```diff
@@ -1594,6 +1594,13 @@ index 0000000000000000000000000000000000000000..a33de97340f14219291c4175e9194914
 +            public boolean tick = true;
 +        }
 +
++        public Sniffer sniffer;
++
++        public class Sniffer extends ConfigurationPart {
++            public int hatchTime = 24000;
++            public int boostedHatchTime = 12000;
++        }
```
Machine-Maker: This is more optional, but I've really been becoming a fan of using vanilla constants whenever possible. So maybe use 2 ATs to make the 2 constants for these tick times in SnifferEggBlock public, and reference them here instead. It serves 2 purposes, if the time changes, it auto updates, and if the field is removed, then it conflicts/causes a compile error making someone look at the changes

patches/server/0855-Add-missing-BlockFadeEvents.patch
```diff
@@ -20,3 +20,34 @@ index 834c5e3fbff3819f3f72e95a1072d9b9e57f25b3..294d22b6b27e96b59c77527efcfefa94
...
++        world.scheduleTick(pos, this, (baseDelay / 3) + world.random.nextInt(300));
```
Machine-Maker: Use the constant field here too, `RANDOM_HATCH_OFFSET_TICKS`.

Machine-Maker (changes requested):

patches/server/0005-Paper-config-files.patch
```diff
@@ -1601,6 +1604,13 @@ index 0000000000000000000000000000000000000000..a33de97340f14219291c4175e9194914
 +            public boolean tick = true;
 +        }
 +
++        public Sniffer sniffer;
++
++        public class Sniffer extends ConfigurationPart {
++            public int hatchTime = SnifferEggBlock.REGULAR_HATCH_TIME_TICKS;
```
lynxplay: Should be IntOr.Default

lynxplay (changes requested):

Machine-Maker (approved):

lynxplay (approved):

PR #9289, opened 2023-06-09
Add back the source uuid on exp orb for animals breeding

Lulu13022002:
The source uuid is null when two animals breed


lynxplay (changes requested):
The turtles spawn them on their own, should probably also be addressed.

lynxplay (approved):
I lied, turtles don't get the right to be the source of exp *yet*
(javadocs specify the child entity to be the source of exp during breeding, needs expansion for fertialisation, further work is needed, not in this PR tho)

Machine-Maker (approved):

Issue #8747, opened 2022-12-31
Packet Rate kick reason


jacklollz2:
### Is your feature request related to a problem?
When players get kicked from a packet rate override limit, it doesn't tell me what limiter triggered the kick.
### Describe the solution you'd like.
I would like the console to output which packet rate override trigger occurred.
### Describe alternatives you've considered.
Plugin use, but it's not as efficient as the built in packet limiter.

Doc94:
What is the message when are kicked?

jacklollz2:
> What is the message when are kicked?

No, I am looking for the console to output which override was triggered.
For example, if I create an override for ServerboundSetCreativeModeSlotPacket and make it so that players get kicked by this packet, it just says "Kicked for exceeding packet rate limit", but I want to know in the console exactly which override triggered.

Doc94:
okay i think this is related to this patch https://github.com/PaperMC/Paper/blob/d5f1ffc2e6928a6728f91ab2ad2e020b105eab7f/patches/server/0713-Add-packet-limiter-config.patch

not sure if is better and another logger (info/warn) for tell in console what is the packet or add this info to the kick message (this broke the vanilla thing)

jacklollz2:
> okay i think this is related to this patch https://github.com/PaperMC/Paper/blob/d5f1ffc2e6928a6728f91ab2ad2e020b105eab7f/patches/server/0713-Add-packet-limiter-config.patch
> 
> not sure if is better and another logger (info/warn) for tell in console what is the packet or add this info to the kick message (this broke the vanilla thing)

It is no bug, I just want to see info on the trigger
I made it kick people intentionally, I just want to see the cause rather than vague info

NoahvdAa:
Resolved in b1696788d61ddc338bd2e61fac3b43ff60c432e1 (#9292)

PR #9292, opened 2023-06-09
ratelimited packet is logged to console on kick

bdotsamir:
First PR 🎉 Huge thanks to lynxplay for helping me out here

Hopefully fixes issue: #8747


EternalHuman:
I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL I WAS HERE xqcL

yannicklamprecht (approved):

lynxplay:
Welcome to paper :tada: thank you for your first PR.

Things we should discuss here are

- [ ] Should this be behind a configuration option ?
- [ ] This should probably also explicitly mention the user kicked

olijeffers0n:
I don't think this needs to be behind a config option as it is, arguably, very useful information. But yeah, the user / UUID in suspect should be logged IMO

cosmohdx:
I don‘t see any case this is not needed. In every case i have in my mind, it would be thankful to have the reason and the user logged.

electronicboy:
depends on how easy it is to trip, flooding logs is one of the many fun DDoS style attacks you can do, especially for servers running in janky wrapper processes; this is probably not a concern here, but it's not like log messages are 100% free

MachineBreaker:
This should be under a configuration option considering what electronicboy said

Owen1212055 (approved):
I see no huge reason for a config, especially since kicking the player already has some logging in itself. This just makes it a bit more descriptive.

electronicboy (approved):

PR #9296, opened 2023-06-10
Remove stacktrace on exception from reading server.properties

Machine-Maker:
1.20 added a full stacktrace to the error message when the server.properties doesn't exist. This stacktrace shouldn't be printed, really. The exception type+message should suffice.

If we want to be more specific, we could just skip the stacktrace if the exception is FileNotFoundException, but I feel like this is fine.


Owen1212055 (approved):

Owen1212055:
Don't we have a patch that does something similar? Would be nice to merge perhaps.

lynxplay:
Yea this would fit nicely into https://github.com/PaperMC/Paper/blob/master/patches/server/0920-Improve-logging-and-errors.patch

Machine-Maker:
Refactored to just check if server.properties exists at all before trying to read from it. If it doesn't, just return empty properties and log a regular line saying default values are being used.

electronicboy (approved):

PR #9297, opened 2023-06-10
Add PlayerPlaceBlockEvent/PostPlayerPlaceBlockEvent

Owen1212055:
In general, there are many flaws with the current block place event. Mainly, the server has to properly attempt to revert the block being placed, however, this had led to an unpredictable amount of issues each update and in general has seen to introduce some duping issues.

This PR suggests deprecating the current events and in favor of introducing events that more properly represent the current block-placing process. 

The **PlayerPlaceBlockEvent** allows you to cancel the placing of a block. This is a simple event and occurs at the highest point of block placement logic.

The **PostPlayerPlaceBlockEvent** allows you to get what blocks are placed due to a block placement, however, unlike the BlockPlaceEvent, you are unable to cancel this event. This is because the API cannot promise that the state will actually be resolved.

There isn't any win-win situation for this. In general, it is impossible to nicely revert block placements like this.

### DIFF
--- a/patches/api/0419-Add-PlayerPlaceBlockEvent-PostPlayerPlaceBlockEvent.patch
+++ b/patches/api/0419-Add-PlayerPlaceBlockEvent-PostPlayerPlaceBlockEvent.patch
@@ -1,0 +1,241 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Fri, 9 Jun 2023 21:35:22 -0400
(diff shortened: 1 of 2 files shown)


Malfrador:
Where in the new logic would the BlockCanBuildEvent fit in? To me it seems like the PlayerPlaceBlockEvent serves the same purpose, except that it properly uses Cancellable and not some weird setBuildable logic

Owen1212055:
> Where in the new logic would the BlockCanBuildEvent fit in? To me it seems like the PlayerPlaceBlockEvent serves the same purpose, except that it properly uses Cancellable and not some weird setBuildable logic

Very similar, but we can prolly end up deprecating that event too. In general that event is a block event, which can lead to the improper assumption that the block is already in the world (bad api, in general). And so really, it's better to just have our generic event there instead.

PR #9298, opened 2023-06-10
Player standing on position API

Owen1212055:
Allows plugins to more accurately detect what block the player is being supported by. Much better than just using -1 current entity position.

### DIFF
--- a/patches/api/0419-Player-standing-on-position-API.patch
+++ b/patches/api/0419-Player-standing-on-position-API.patch
@@ -1,0 +1,55 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Fri, 9 Jun 2023 22:49:45 -0400
+Subject: [PATCH] Player standing on position API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Entity.java b/src/main/java/org/bukkit/entity/Entity.java
+index a2a423d4e4c2702ba5967223cab0432dd7d04732..6fa7d251f54be9d0ef6b0ec0283efa927bae1a0b 100644
+--- a/src/main/java/org/bukkit/entity/Entity.java
++++ b/src/main/java/org/bukkit/entity/Entity.java
+@@ -954,4 +954,44 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
+      */
+     boolean wouldCollideUsing(@NotNull BoundingBox boundingBox);
+     // Paper End - Collision API
++    // Paper start - Player standing on position API
++    /**
++     * Gets the block that is currently being used to calculate movement affects (i.e friction) on the entity.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the ice block.
++     * - When flying on top of a flower located under a grass block this will return the position of the grass block (although this technically will not affect your movement).
++     *
++     * @return block position used for movement affects
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getBlockPosBelowThatAffectsMyMovement();
++
++    /**
++     * Gets the block that is currently under the entity.
++     * This will return {@link Entity#getSupportingPos()} as long as it is not null,
++     * else it will instead return the bottom most block at the entity's hitbox.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the slab.
++     * - When flying on top of a flower located under a grass block this will return the position of the corn flower.
++     *
++     * @return {@link Entity#getSupportingPos()} if not null or the entity's bottom most block
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getOnPos();
++
++    /**
++     * Gets the block that is currently supporting the entity.
++     * This takes into account the collisions of blocks under the entity.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the slab.
++     * - When flying on top of a flower located under a grass block this will return null, because no block is holding your location.
++     *
++     * @return block position currently supporting the entity, or null if a block is currently not supporting the entity
++     */
++    @Nullable
++    io.papermc.paper.math.BlockPosition getSupportingPos();
++    // Paper end
+ }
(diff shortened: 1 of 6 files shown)


patches/api/0419-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,55 @@
...
++     * Gets the block that is currently being used to calculate movement affects (i.e friction) on the entity.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the ice block.
++     * - When flying on top of a flower located under a grass block this will return the position of the grass block (although this technically will not affect your movement).
++     *
++     * @return block position used for movement affects
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getBlockPosBelowThatAffectsMyMovement();
```
lynxplay: :eyes: This really feels like a spigot method name :tm:

patches/api/0419-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,55 @@
...
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getOnPos();
++
++    /**
++     * Gets the block that is currently supporting the entity.
++     * This takes into account the collisions of blocks under the entity.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the slab.
```
Sytm: A lot of *sanding* is happening here

patches/api/0419-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,55 @@
...
++     * else it will instead return the bottom most block at the entity's hitbox.
++     * <p>
++     * Examples:
++     * - When sanding on a slab under an ice block, this will return the position of the slab.
++     * - When flying on top of a flower located under a grass block this will return the position of the corn flower.
++     *
++     * @return {@link Entity#getSupportingPos()} if not null or the entity's bottom most block
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getOnPos();
```
Machine-Maker: I think this would probably be more useful if it returned a `Block`. The javadocs even say `Gets the block...`. I think the usual usecase is gonna be to check the type. CraftBlock is just a wrapper so cheap to create.

Machine-Maker (changes requested):
The javadocs for these methods are very confusing I think.

patches/api/0419-Player-standing-on-position-API.patch
Owen1212055: Not exactly because it doesn't always represent a block actually. See the javadoc comments, where depending on the entity's position it returns the bottom of their hitbox.

Owen1212055: I am not sure actually, opinions?

patches/api/0419-Player-standing-on-position-API.patch
Owen1212055: Fixed hopefully

molor:
I like the new names of the methods (full "position" instead of "pos").

patches/api/0419-Player-standing-on-position-API.patch
Machine-Maker: These values always apply to the world the player is in correct? If that is the case, Block or Location seems more appropriate. Giving extra context in a returned value is fine. The problem with Location usually shows up when it is a parameter rather than a return value.

Owen1212055: Hmm okay, I think Location is fine then. I would just rather stay to more primitive types.

patches/api/0458-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,55 @@
...
++     * - When standing on a slab under an ice block, this will return the position of the ice block.
++     * - When flying on top of a flower located under a grass block this will return the position of the grass block (although this technically will not affect your movement).
++     *
++     * @return block position used for movement affects
++     */
++    @NotNull
++    io.papermc.paper.math.BlockPosition getMovementAffectingBlock();
++
++    /**
++     * Gets the position of the the that is currently under the entity.
```
granny:
```suggestion
+     * Gets the position of the block that is currently under the entity.
```

lynxplay (changes requested):
Does not function during PlayerMoveEvent.
Caused by *something* clearing the supported block optional temporarily in the logic flow.

Needs further investigation before merging as onMove is a way too useful place to call this, merging this without that working is a dead end.

notTamion (changes requested):
Perhaps mention methods in the Player#isOnGround deprecation message

patches/api/0459-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,41 @@
...
+@@ -1107,4 +1107,30 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
+      */
+     @NotNull String getScoreboardEntryName();
+     // Paper end - entity scoreboard name
++    // Paper start - Player standing on position API
++    /**
++     * Gets the block that is currently being used to calculate movement effects (i.e friction) on the entity.
++     * <p>
++     * Examples:
++     * - When standing on a slab under an ice block, this will return the position of the ice block.
```
FireInstall, line 20: This reads wierdly, like if the player stands on the ice and under the ice is a slab.

```suggestion
+     * - When standing on a slab on top of an ice block, this will return the position of the ice block.
```

patches/api/0459-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,41 @@
...
++     */
++    @NotNull
++    org.bukkit.block.Block getMovementAffectingBlock();
++
++    /**
++     * Gets the block that is currently supporting the entity.
++     * This takes into account the collisions of blocks under the entity.
++     * <p>
++     * Examples:
++     * - When standing on a slab under an ice block, this will return the position of the slab.
```
FireInstall, line 33: same here - alternatively maybe `When standing on a slab with an ice block  under it `

patches/api/0459-Player-standing-on-position-API.patch
```diff
@@ -0,0 +1,41 @@
...
+     // Paper end - entity scoreboard name
++    // Paper start - Player standing on position API
++    /**
++     * Gets the block that is currently being used to calculate movement effects (i.e friction) on the entity.
++     * <p>
++     * Examples:
++     * - When standing on a slab under an ice block, this will return the position of the ice block.
++     * - When flying on top of a flower placed on a grass block, this will return the position of the grass block (although this technically will not affect your movement).
++     *
++     * @return block used for movement affects
```
rymiel, line 23: effects

PR #9300, opened 2023-06-11
Add event for player editing sign

### LINKED ISSUE #9293: Player Edit Sign Event

Cancelling PlayerInteractEvent might be limiting because there's a somewhat complicated logic for determining which side of sign the player clicked. (See nms `SignBlockEntity#isFacingFrontText`)

A event that triggers when a player tries to edit a sign. With a way to get which side of the sign is being clicked. Cancel the event to do nothing when clicked (don't open the sign edit screen)

Edit: After seeing @lynxplay's idea, this event maybe should also trigger after placing the sign. Cancelling the event would not open the sign edit screen when placed.

I mean, I can copy the nms logic in my plugin code but that's kinda bad

### Other

_No response_

by77er:
resolves https://github.com/PaperMC/Paper/issues/9293

### DIFF
--- a/patches/api/0420-Add-event-for-player-editing-sign.patch
+++ b/patches/api/0420-Add-event-for-player-editing-sign.patch
@@ -1,0 +1,81 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: by77er <by77er@gmail.com>
+Date: Sat, 10 Jun 2023 19:06:24 -0400
+Subject: [PATCH] Add event for player editing sign
+
+
+diff --git a/src/main/java/io/papermc/paper/event/player/PlayerOpenSignEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerOpenSignEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..4f81a6dbe6694d40f27445f5d17aed48ad75498a
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/player/PlayerOpenSignEvent.java
+@@ -0,0 +1,69 @@
++package io.papermc.paper.event.player;
++
++import org.bukkit.block.Sign;
++import org.bukkit.block.sign.Side;
++import org.bukkit.entity.Player;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player begins editing a sign's text.
++ * <p>
++ * Cancelling this event stops the sign editing menu from opening.
++ */
++public class PlayerOpenSignEvent extends PlayerEvent implements Cancellable {
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private boolean cancel = false;
++    @NotNull private final Sign sign;
++    @NotNull private final Side side;
++
++    public PlayerOpenSignEvent(@NotNull Player editor, @NotNull Sign sign, @NotNull Side side) {
++        super(editor);
++        this.sign = sign;
++        this.side = side;
++    }
++
++    /**
++     * Gets the sign that was clicked.
++     *
++     * @return {@link Sign} that was clicked
++     */
++    @NotNull
++    public Sign getSign() {
++        return sign;
++    }
++
++    /**
++     * Gets which side of the sign was clicked.
++     *
++     * @return {@link Side} that was clicked
++     * @see Sign#getSide(Side)
++     */
++    @NotNull
++    public Side getSide() {
++        return side;
++    }
++
++    @Override
++    public boolean isCancelled() {
++        return cancel;
++    }
++
++    @Override
++    public void setCancelled(boolean cancel) {
++        this.cancel = cancel;
++    }
++
++    @NotNull
++    @Override
++    public HandlerList getHandlers() {
++        return HANDLER_LIST;
++    }
++
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return HANDLER_LIST;
++    }
++}
--- a/patches/server/0006-MC-Dev-fixes.patch
+++ b/patches/server/0006-MC-Dev-fixes.patch
@@ -184,6 +184,31 @@ index 4d2d23ff118e9307d50ca2a5194b83450c91a752..ab6dc3449a1d3b7acf1d7bf5ac1c2422
          }));
          this.byName = Maps.newHashMap(builder.build()); // CraftBukkit
          RecipeManager.LOGGER.info("Loaded {} recipes", map1.size());
+diff --git a/src/main/java/net/minecraft/world/level/block/SignBlock.java b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+index 91bb294be2cd8ab3467a62006a5a2751e0bec4ba..5ab7df0042391cb621ed78a187dc20333e344186 100644
+--- a/src/main/java/net/minecraft/world/level/block/SignBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+@@ -76,18 +76,9 @@ public abstract class SignBlock extends BaseEntityBlock implements SimpleWaterlo
+     public InteractionResult use(BlockState state, Level world, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit) {
+         ItemStack itemStack = player.getItemInHand(hand);
+         Item item = itemStack.getItem();
+-        Item signBlockEntity = itemStack.getItem();
+-        SignApplicator var10000;
+-        if (signBlockEntity instanceof SignApplicator signApplicator) {
+-            var10000 = signApplicator;
+-        } else {
+-            var10000 = null;
+-        }
+-
+-        SignApplicator signApplicator2 = var10000;
++        SignApplicator signApplicator2 = item instanceof SignApplicator signApplicator ? signApplicator : null; // Paper - decompile fixes
+         boolean bl = signApplicator2 != null && player.mayBuild();
+-        BlockEntity bl2 = world.getBlockEntity(pos);
+-        if (bl2 instanceof SignBlockEntity signBlockEntity) {
++        if (world.getBlockEntity(pos) instanceof SignBlockEntity signBlockEntity) { // Paper - decompile fixes
+             if (!world.isClientSide) {
+                 boolean bl2 = signBlockEntity.isFacingFrontText(player);
+                 SignText signText = signBlockEntity.getText(bl2);
 diff --git a/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java b/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java
 index b445f90b70d5a74f5bdc1b081500da4742e1c093..afba733824d84d650e669eda028bf4d28cf006b4 100644
 --- a/src/main/java/net/minecraft/world/level/chunk/PalettedContainer.java
--- a/patches/server/0980-Add-event-for-player-editing-sign.patch
+++ b/patches/server/0980-Add-event-for-player-editing-sign.patch
@@ -1,0 +1,27 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: by77er <by77er@gmail.com>
+Date: Mon, 12 Jun 2023 12:56:46 -0400
+Subject: [PATCH] Add event for player editing sign
+
+
+diff --git a/src/main/java/net/minecraft/world/level/block/SignBlock.java b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+index 5ab7df0042391cb621ed78a187dc20333e344186..f7fda6fb4e908869310c783e68f7ad7025840592 100644
+--- a/src/main/java/net/minecraft/world/level/block/SignBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+@@ -144,6 +144,16 @@ public abstract class SignBlock extends BaseEntityBlock implements SimpleWaterlo
+     }
+ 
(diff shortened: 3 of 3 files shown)


patches/api/0419-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,72 @@
...
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player edits a sign
++ */
++public class PlayerEditSignEvent extends PlayerEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker: `HANDLER_LIST`

patches/api/0419-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,72 @@
...
++        this.side = side;
++    }
++
++    /**
++     * Get the sign that was clicked
++     *
++     * @return {@link Sign} that was clicked
++     */
++    @NotNull
++    public Sign getSign() { return sign; }
```
Machine-Maker: No need to make these methods in one line

patches/api/0419-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,72 @@
...
++    @NotNull
++    public Side getSide() { return side; }
++
++    public boolean isCancelled() {
++        return cancel;
++    }
++
++    public void setCancelled(boolean cancel) {
++        this.cancel = cancel;
++    }
```
Machine-Maker: Override annotations

patches/server/0977-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,53 @@
...
++        Item potentialApplicator = itemStack.getItem(); // Paper: renamed signBlockEntity -> potentialApplicator due to decompile error
+         SignApplicator var10000;
+-        if (signBlockEntity instanceof SignApplicator signApplicator) {
++        if (potentialApplicator instanceof SignApplicator signApplicator) { // Paper: renamed signBlockEntity -> potentialApplicator due to decompile error
+             var10000 = signApplicator;
+         } else {
+             var10000 = null;
+@@ -86,8 +86,8 @@ public abstract class SignBlock extends BaseEntityBlock implements SimpleWaterlo
+ 
+         SignApplicator signApplicator2 = var10000;
```
Machine-Maker: This whole block is super confusing, idk what happened to the decompiler here. Is it worth just deleting it and rewriting it to make actual sense? Looking for input here, not saying it should be done.

patches/server/0977-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,53 @@
...
+                 SignText signText = signBlockEntity.getText(bl2);
+@@ -153,8 +153,17 @@ public abstract class SignBlock extends BaseEntityBlock implements SimpleWaterlo
+     }
+ 
+     public void openTextEdit(Player player, SignBlockEntity blockEntity, boolean front) {
+-        blockEntity.setAllowedPlayerEditor(player.getUUID());
+-        player.openTextEdit(blockEntity, front);
++        // Paper start
++        var bukkitPlayer = (org.bukkit.entity.Player) player.getBukkitEntity();
++        var event = new com.destroystokyo.paper.event.player.PlayerEditSignEvent(
```
Machine-Maker: We aren't big fans of `var` even if it saves a fully qualified name.

patches/server/0977-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,53 @@
...
++            bukkitPlayer,
++            new org.bukkit.craftbukkit.block.CraftSign<>(bukkitPlayer.getWorld(), blockEntity),
```
Machine-Maker: This isn't correct, it could be a hanging sign. CraftSign doesn't cover that. You should instead use one of the methods on `CraftBlockStates` to get the blockstate instance.

Machine-Maker (dismissed):

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: In working on this I realized it might be better to return a `Block` in place of the `Sign`, since the event handler probably needs more context anyway. Thoughts?

Machine-Maker: Sign is fine as HangingSign extends it, but you need to create a CraftHangingSign if its a hanging sign instead of CraftSign for a regular sign.

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Done!

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Expanded.

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Done!

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: Replaced with fully qualified names.

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Added.

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: Alright, let me know if it looks good now.

molor:
Also wrong (outdated) package name

patches/server/0977-Add-event-for-player-editing-sign.patch
lynxplay: Yea I honestly wouldn't mind just yanking it.
I mean, its gonna get diffed on every update that touches it anyway with the amount of decompiler fixes it requires.

lynxplay: Both 
```patch
-        Item signBlockEntity = itemStack.getItem();
-        SignApplicator var10000;
-        if (signBlockEntity instanceof SignApplicator signApplicator) {
-            var10000 = signApplicator;
-        } else {
-            var10000 = null;
-        }
-
-        SignApplicator signApplicator2 = var10000;
+        SignApplicator signApplicator2 = item instanceof SignApplicator signApplicator ? signApplicator : null; // Paper - decompile fixes
```

and 
```patch
-        BlockEntity bl2 = world.getBlockEntity(pos);
-        if (bl2 instanceof SignBlockEntity signBlockEntity) {
+        if (world.getBlockEntity(pos) instanceof SignBlockEntity signBlockEntity) { // Paper - decompile fixes
```
would probably do it.

patches/api/0419-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,80 @@
...
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called when a player edits a sign
++ * <p>
++ * Cancelling this event stops the sign editing menu from appearing.
++ */
++public class PlayerEditSignEvent extends PlayerEvent implements Cancellable {
```
lynxplay: I wonder if this name properly communicates the fact that this event isn't really called when the edit happens but when the player opens the editing screen.

Maybe it'd be great to reiterate if we can find a more specific name here.

KabanFriends:
Hello, I'm the one who made the original suggestion #9293, and I have another idea for this. How about triggering the event for when player colors the sign text using dyes, inc sac, glow inc sac etc?

patches/api/0419-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,80 @@
...
++     * @return {@link Sign} that was clicked
++     */
++    @NotNull
++    public Sign getSign() {
++        return sign;
++    }
++
++    /**
++     * Get which side of the sign was clicked
++     * @return {@link Side} which was clicked
```
Lulu13022002: Missing a newline between the description and the return

patches/server/0977-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,123 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/block/SignBlock.java b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+index 91bb294be2cd8ab3467a62006a5a2751e0bec4ba..7909c3404b235c8259f9f9decd198e68e8ab9397 100644
+--- a/src/main/java/net/minecraft/world/level/block/SignBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/SignBlock.java
+@@ -3,6 +3,7 @@ package net.minecraft.world.level.block;
+ import java.util.Arrays;
+ import java.util.UUID;
+ import javax.annotation.Nullable;
++
```
Lulu13022002: Unnecessary diff

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: Added a commit that flattens the logic out in a more readable way. LMK if this is what you meant!

lynxplay: Might be a bit too much imo. The two diffs I posted above yank a bunch but leave the core logic that follows in tact.
Tho I'd like machines input on this too in regards to what he wanted :+1: 

Thanks for the work already tho!

by77er: Of course. Whatever works best!

bdotsamir (commented):
I remember looking at this file when I was going to try my hand at fixing the issue and thinking "damn these variable names are awful." thank you for cleaning them up, but isn't the goal to make the diff as small as possible? would a change like this warrant ignoring that goal?

edit since it didn't appear properly: I'm referring to the SignBlock class

by77er:
@bdotsamir Yeah, minimizing the diff makes sense to me. I think I'll push with @lynxplay's change. I'll hang on to the rewritten version just in case that's the way we decide to go. Still gotta decide on a good name for the event.

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Maybe just `PlayerOpenSignEvent`? Least clunky name I could think of, if a little vague.

patches/server/0977-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,61 @@
...
+-        Item signBlockEntity = itemStack.getItem();
+-        SignApplicator var10000;
+-        if (signBlockEntity instanceof SignApplicator signApplicator) {
+-            var10000 = signApplicator;
+-        } else {
+-            var10000 = null;
+-        }
+-
+-        SignApplicator signApplicator2 = var10000;
++        SignApplicator signApplicator2 = item instanceof SignApplicator signApplicator ? signApplicator : null; // Paper - decompile fixes
```
Lulu13022002: This should go into the right patch (MC-Dev fixes)

patches/api/0419-Add-event-for-player-editing-sign.patch
by77er: Added.

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: Removed.

patches/server/0977-Add-event-for-player-editing-sign.patch
by77er: Done. Good to know, thanks.

patches/server/0975-Add-event-for-player-editing-sign.patch
```diff
@@ -0,0 +1,32 @@
...
+-        player.openTextEdit(blockEntity, front);
++        // Paper start
++        org.bukkit.entity.Player bukkitPlayer = (org.bukkit.entity.Player) player.getBukkitEntity();
++        org.bukkit.block.Block bukkitBlock = org.bukkit.craftbukkit.block.CraftBlock.at(blockEntity.getLevel(), blockEntity.getBlockPos());
++        org.bukkit.craftbukkit.block.CraftSign<?> bukkitSign = (org.bukkit.craftbukkit.block.CraftSign<?>) org.bukkit.craftbukkit.block.CraftBlockStates.getBlockState(bukkitBlock);
++        io.papermc.paper.event.player.PlayerOpenSignEvent event = new io.papermc.paper.event.player.PlayerOpenSignEvent(
++            bukkitPlayer,
++            bukkitSign,
++            front ? org.bukkit.block.sign.Side.FRONT : org.bukkit.block.sign.Side.BACK);
++        if (event.callEvent()) {
```
lynxplay: Could technically reduce the diff here by changing this to a 
```java
if(!event.callEvent()) return
```

lynxplay (commented):
Generally looks good to me, I'll be testing it tomorrow to hopefully give it an approval :)
Thanks already for the great work!

patches/server/0975-Add-event-for-player-editing-sign.patch
by77er: True! Done.

molor:
(do not forget this PR please, we're waiting very much for it.. :c also thanks for all your work!)

kennytv (approved):

by77er:
Rebased and ready.

electronicboy (approved):

ammodev:
Using signSide.line(component) has no effect on the resulting gui, does it?

PR #9313, opened 2023-06-11
Use net.kyori.ansi for console logging

### LINKED ISSUE #7827: GUI console shows legacy formatting codes

rymiel:
Uses the new [ANSIComponentSerializer](https://github.com/KyoriPowered/adventure/tree/main/4/text-serializer-ansi) introduced in [Adventure 4.14.0](https://github.com/KyoriPowered/adventure/releases/tag/v4.14.0) to serialize components when logging them via the ComponentLogger, or when sending messages to the console.

This replaces the old solution which uses legacy jank and custom color conversions, with a new library [(net.kyori.ansi)](https://github.com/KyoriPowered/ansi) that handles the conversion and config.

Open questions:

 - The underlying net.kyori.ansi library tries to auto-detect the current terminal's color capabilities, but it can also be configured through some system properties, [documented here](https://docs.advntr.dev/serializer/ansi.html#usage). Should this also be documented on Paper's side?

 - This change removes the old `minecraftFormatting` log pattern converter, since it is no longer necessary, and is thus removed in the bundled default log4j config. However, anyone using a custom config will still probably have the converter in there. Should there be a noop shim to facilitate the old configs, and should it just be a noop or maybe even warn on use?

 - Is the new method fully compatible with existing setups? I can't answer this myself since I'm 100% sure of the details of the system here, but I do know the original (intentionally) broke a lot of legacy uses. This new method should hopefully work in exactly as many cases as the system it is replacing, but I can't be sure here.


jpenilla:
We should create a new log4j plugin to strip ansi codes for log files and the server gui

Ansi codes are going to be even more of a mess than the psuedo legacy formatting currently is there

jpenilla:
The `use-rgb-for named-text-colors` config option should probably be removed since it won't do anything anymore

rymiel:
> The `use-rgb-for named-text-colors` config option should probably be removed since it won't do anything anymore

It could technically be hooked up into the ANSI serializer ColorLevel stuff, but i guess the other option is to just remove it from the default config and forget it exists?

jpenilla:
I think keeping it around just needlessly complicates things, the new color support detection seems to be more robust in general, and people who have issues can use the system property.

rymiel:
I've removed the option from the config patch

jpenilla (approved):

Owen1212055:
Wouldn't we want to merge these into the patch that adds the TerminalConsoleCommandSender class?

emilyy-dev:
> Wouldn't we want to merge these into the patch that adds the TerminalConsoleCommandSender class?

isn't the patch file effectively renamed? same spot, different commit message, so different file name

Owen1212055:
Yeah, I guess tho I don't see why we shouldn't just merge this onto whatever patch we currently have for console logging logic?

rymiel:
Merged the patches, slightly changed the commit message cuz it wasn't that accurate anymore

jpenilla:
the log4j annotationProcessor dep needs to be moved up too

rymiel:
Moved up from where?

patches/server/0977-Fix-sendSignChange0-s-usage-of-SignText.patch
emilyy-dev, line 1: hm?

rymiel: no clue, but i think i got it back now

patches/server/0005-Paper-config-files.patch
```diff
@@ -604,7 +604,6 @@ index 0000000000000000000000000000000000000000..52b02cb1f02d1c65b840f38cfc8baee5
 +    public class Logging extends ConfigurationPart {
 +        public boolean logPlayerIpAddresses = true;
 +        public boolean deobfuscateStacktraces = true;
-+        public boolean useRgbForNamedTextColors = true;
```
Machine-Maker, line 607: This should be added to the "removed" config paths so its auto removed from configuration files.

Machine-Maker (dismissed):

PR #9322, opened 2023-06-12
Fix incorrect crafting result amount for fireworks

Redned235:
Mojang don't specifically call this code anywhere (since `FireworkRocketRecipe` overrides the `assemble` method), however using the `getResult()` method inside the Bukkit returns the wrong amount for fireworks as the count is not specified in the result item.


Machine-Maker:
Well, it is a `ComplexRecipe` and the javadoc says that getResult is pretty much undefined for these. But I think this is fine as it is always 3 (aka hardcoded).

Redned235:
> Can you clarify that in the patch description then? That generally this is only used in the bukkit api.

Added a comment to the patch.

electronicboy (approved):

PR #9326, opened 2023-06-12
Add getHighestMapId method

chmodsayshello:
This small pull request adds a method to get the highest Map ID currently generated via the bukkit api. Prior to this, plugin developers would need to keep track of this themselves, using some sort of external database...

### DIFF
--- a/patches/api/0419-add-getHighestMapId-method.patch
+++ b/patches/api/0419-add-getHighestMapId-method.patch
@@ -1,0 +1,45 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: chmodsayshello <chmodsayshello@hotmail.com>
+Date: Mon, 12 Jun 2023 18:33:43 +0200
+Subject: [PATCH] add getHighestMapId method
+
+
+diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
+index b0bc2df41506770e2854a287813f1c53f003eda1..410c965d4e663fc085c2fe00905b8a8802d85806 100644
+--- a/src/main/java/org/bukkit/Bukkit.java
++++ b/src/main/java/org/bukkit/Bukkit.java
+@@ -2483,4 +2483,17 @@ public final class Bukkit {
+     public static Server.Spigot spigot() {
+         return server.spigot();
+     }
++
++    // Paper start
++
++    /**
++     * Gets the highest map ID currently created
++     *
++     * @return the map ID (-1 if no maps were created yet)
++     */
++    @NotNull
++    public static int getHighestMapId() {
++        return server.getHighestMapId();
++    }
++    // Paper end
+ }
+diff --git a/src/main/java/org/bukkit/Server.java b/src/main/java/org/bukkit/Server.java
+index 1d1a1d087dabc9794e0062a064da2cced4062309..84dd347b8ab275e28341635a2626b19765fcaa1a 100644
+--- a/src/main/java/org/bukkit/Server.java
++++ b/src/main/java/org/bukkit/Server.java
+@@ -2161,5 +2161,12 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
+      * @return the potion brewer
+      */
+     @NotNull org.bukkit.potion.PotionBrewer getPotionBrewer();
++
++    /**
++     * Gets the highest map ID currently created
++     *
++     * @return the map ID (-1 if no maps were created yet)
++     */
++    int getHighestMapId();
+     // Paper end
+ }
--- a/patches/server/0978-add-getHighestMapId-method.patch
+++ b/patches/server/0978-add-getHighestMapId-method.patch
@@ -1,0 +1,37 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: chmodsayshello <chmodsayshello@hotmail.com>
+Date: Mon, 12 Jun 2023 18:56:48 +0200
+Subject: [PATCH] add getHighestMapId method
+
+
+diff --git a/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java b/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java
+index 9b2948b5150c8f039ca667a50765109721b93947..9f69db46239d3e8055d5793a6a9ba62d321b5e4f 100644
+--- a/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java
++++ b/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java
+@@ -40,4 +40,10 @@ public class MapIndex extends SavedData {
+         this.setDirty();
+         return i;
+     }
++
++    // Paper start - get value without writing
++    public int getCurrentValueForMap() {
++        return this.usedAuxIds.getInt("map");
++    }
++    // Paper end
+ }
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+index ffa27c9c02dc4d12411fc089de3af8e8e12ba06e..2a32560d0e09dc9a93e17b7d13b612ef4a4bd803 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -2971,5 +2971,11 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+ 
++    @Override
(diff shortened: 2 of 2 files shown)


patches/server/0978-add-getHighestMapID-method.patch
```diff
@@ -0,0 +1,45 @@
...
+ }
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+index ffa27c9c02dc4d12411fc089de3af8e8e12ba06e..790271eaad20c8276fcae597b5ff79a81eaf0dfb 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -105,6 +105,7 @@ import net.minecraft.world.level.levelgen.WorldDimensions;
+ import net.minecraft.world.level.levelgen.WorldOptions;
+ import net.minecraft.world.level.material.Fluid;
+ import net.minecraft.world.level.saveddata.maps.MapDecoration;
++import net.minecraft.world.level.saveddata.maps.MapIndex;
```
Owen1212055: Inline this import plesae

patches/server/0978-add-getHighestMapID-method.patch
```diff
@@ -0,0 +1,45 @@
...
+@@ -2971,5 +2972,11 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+ 
++    @Override
++    public int getHighestMapID(){
++        // Paper - all map IDs are meant for the overworld, this line is more or less copied from NMS
++        // Paper - for additional info, view getFreeMapId() in ServerLevel.java and getFreeAuxValueForMap() in MapIndex.java
++        return ((MapIndex) this.getServer().overworld().getDataStorage().computeIfAbsent(MapIndex::load, MapIndex::new, "idcounts")).getCurrentValueForMap();
++    }
```
Owen1212055: Properly surround this in paper comments

Owen1212055 (changes requested):

patches/api/0419-add-getHighestMapID-method.patch
```diff
@@ -0,0 +1,23 @@
...
+      * @return the potion brewer
+      */
+     @NotNull org.bukkit.potion.PotionBrewer getPotionBrewer();
++
++    /**
++     * Gets the highest map ID currently created
++     *
++     * @return the map ID (-1 if no maps were created yet)
++     */
++    int getHighestMapID();
```
Redned235: This is an incredibly small nitpick, but typically the `Id` naming convention is used in the Bukkit API, rather than the fully uppercase `ID`.

patches/server/0978-add-getHighestMapID-method.patch
chmodsayshello: I put this in an already existing paper block... Thats why here, it only shows paper end.

In my opinion doing something like:
```
// Paper End
// Paper Start
<code>
// Paper End
```
is silly! Now, should I still do so?

patches/server/0978-add-getHighestMapID-method.patch
```diff
@@ -0,0 +1,45 @@
...
+ import net.minecraft.world.level.storage.PlayerDataStorage;
+@@ -2971,5 +2972,11 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+ 
++    @Override
++    public int getHighestMapID(){
++        // Paper - all map IDs are meant for the overworld, this line is more or less copied from NMS
++        // Paper - for additional info, view getFreeMapId() in ServerLevel.java and getFreeAuxValueForMap() in MapIndex.java
++        return ((MapIndex) this.getServer().overworld().getDataStorage().computeIfAbsent(MapIndex::load, MapIndex::new, "idcounts")).getCurrentValueForMap();
```
Lulu13022002: The cast doesn't seems needed, also take care, the test fail because you need to mirror all the stuff into Bukkit

patches/server/0978-add-getHighestMapID-method.patch
Owen1212055: You're fine then

patches/server/0978-add-getHighestMapID-method.patch
chmodsayshello: Basicly I copied this part from NMS and just changed it to use the method I added... Gonna test removing it later, thanks!

chmodsayshello: You mean like providing a way to call this from the main Bukkit class?

Lulu13022002: Yes you can copy the method name into the Bukkit class (with a static method) and then add the impl that just redirect over the server with server.getHighestMapID();

chmodsayshello: ummm, sure, I will add it, but if I may ask, what is the use of that/when would anyone use the static method instead of the one in the server object?

electronicboy: Because 99% of people just use the bukkit singletons methods, as a level of "that's how it's been for years"

chmodsayshello: added to the core Bukkit class

patches/api/0419-add-getHighestMapID-method.patch
chmodsayshello: done, thanks ;)

patches/server/0978-add-getHighestMapID-method.patch
chmodsayshello: cast removed!

patches/server/0978-add-getHighestMapID-method.patch
chmodsayshello: inlined

patches/server/0978-add-getHighestMapID-method.patch
Lulu13022002: Good news, the test has been skipped so now you can do whatever you want and decide if this method should stay in the Bukkit class or not without this restriction. Also small nitpick but the parentheses of the cast remains.

chmodsayshello: I guess I will just leave it there then

patches/api/0419-add-getHighestMapId-method.patch
```diff
@@ -0,0 +1,45 @@
...
++
++    // Paper start
++
++    /**
++     * Gets the highest map ID currently created
++     *
++     * @return the map ID (-1 if no maps were created yet)
++     */
++    @NotNull
++    public static int getHighestMapId() {
```
Warriorrrr, line 24: This annotation is redundant, an int is never null.

patches/server/0978-add-getHighestMapId-method.patch
```diff
@@ -0,0 +1,37 @@
...
+--- a/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java
++++ b/src/main/java/net/minecraft/world/level/saveddata/maps/MapIndex.java
+@@ -40,4 +40,10 @@ public class MapIndex extends SavedData {
+         this.setDirty();
+         return i;
+     }
++
++    // Paper start - get value without writing
++    public int getCurrentValueForMap() {
++        return this.usedAuxIds.getInt("map");
```
Warriorrrr, line 18: You could possibly use an [AT](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#access-transformers) here instead of a new method.

patches/server/0978-add-getHighestMapId-method.patch
```diff
@@ -0,0 +1,37 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -2971,5 +2971,11 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+ 
++    @Override
++    public int getHighestMapId(){
++        // Paper - all map IDs are meant for the overworld, this line is more or less copied from NMS
++        // Paper - for additional info, view getFreeMapId() in ServerLevel.java and getFreeAuxValueForMap() in MapIndex.java
++        return (this.getServer().overworld().getDataStorage().computeIfAbsent(net.minecraft.world.level.saveddata.maps.MapIndex::load, net.minecraft.world.level.saveddata.maps.MapIndex::new, "idcounts")).getCurrentValueForMap();
```
Warriorrrr, line 34: The extra parentheses around it aren't needed, probably a leftover from the cast.

Warriorrrr (commented):
It might be worth mentioning in the javadoc that the method can do file IO in some cases, but other than some trivial things this looks pretty good.

patches/server/0978-add-getHighestMapId-method.patch
kennytv: Nah it's fine for a single thing

PR #9330, opened 2023-06-12
Add Adventure InventoryView title methods

### LINKED ISSUE #9189: Add adventure setTitle (and originalTitle?) methods to inventory view
Spigot's implementation isn't... *nice*, so we'll have to see if we just carbon copy it with components or change it up

TheRealRyGuy:
Adds in `Compontent` methods for the new title methods in `InventoryView` - wasn't 100% sure on the [custom inventoryholder rebased patch changes](https://github.com/PaperMC/Paper/compare/master...TheRealRyGuy:Paper:feat/invtitle?expand=1#diff-f324938383e32d7ae067b48d303501fffb5096b1ef52d413b0861c3fb950bb57), but everything else should be alright
Partially fixes #9189, however, I just carbon copied Spigot's implementation, unsure if Paper wanted a better implementation

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -3957,10 +3957,10 @@ index 5adbe0514129abf3cfbc4b29a213f522359fe2e1..72ebc29db42d08d1d0361dba462fc8a5
  
      /**
 diff --git a/src/main/java/org/bukkit/inventory/InventoryView.java b/src/main/java/org/bukkit/inventory/InventoryView.java
-index e12996492c1558fed9fab30de9f8018e0ed7fac3..002acfbdce1db10f7ba1b6a013e678f504ac6e69 100644
+index e12996492c1558fed9fab30de9f8018e0ed7fac3..f67dc7a242cf7ecf5815ed1ae455442fbe09e62c 100644
 --- a/src/main/java/org/bukkit/inventory/InventoryView.java
 +++ b/src/main/java/org/bukkit/inventory/InventoryView.java
-@@ -447,12 +447,26 @@ public abstract class InventoryView {
+@@ -447,12 +447,51 @@ public abstract class InventoryView {
          return getPlayer().setWindowProperty(prop, value);
      }
  
@@ -3973,6 +3973,31 @@ index e12996492c1558fed9fab30de9f8018e0ed7fac3..002acfbdce1db10f7ba1b6a013e678f5
      @NotNull
 +    public /*abstract*/ net.kyori.adventure.text.Component title() {
 +        return net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(this.getTitle());
++    }
++
++    /**
++     * Get the original title of this inventory window, before any changes were
++     * made using {@link #title(net.kyori.adventure.text.Component)}.
++     *
++     * @return the original title
++     */
++    public /*abstract*/ net.kyori.adventure.text.@NotNull Component originalTitle() {
++        return net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(this.getOriginalTitle());
++    }
++
++    /**
++     * Sets the title of this inventory window to the specified title if the
++     * inventory window supports it.
++     * <p>
++     * Note if the inventory does not support titles that can be changed (ie, it
++     * is not creatable or viewed by a player), then this method will throw an
++     * exception.
++     *
++     * @param title The new title.
++     * @throws IllegalArgumentException if you cannot set the title of this inventory
++     */
++    public /*abstract*/ void title(net.kyori.adventure.text.@NotNull Component title) {
++        this.setTitle(net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().serialize(title));
 +    }
 +    // Paper end
 +
@@ -3987,6 +4012,25 @@ index e12996492c1558fed9fab30de9f8018e0ed7fac3..002acfbdce1db10f7ba1b6a013e678f5
      public abstract String getTitle();
  
      /**
+@@ -460,7 +499,9 @@ public abstract class InventoryView {
+      * made using {@link #setTitle(String)}.
+      *
+      * @return the original title
++     * @deprecated in favour of {@link #originalTitle()}
+      */
++    @Deprecated // Paper
+     @NotNull
+     public abstract String getOriginalTitle();
+ 
+@@ -473,6 +514,8 @@ public abstract class InventoryView {
+      * exception.
+      *
+      * @param title The new title.
++     * @deprecated in favour of {@link #title(net.kyori.adventure.text.Component)}
+      */
++    @Deprecated // Paper
+     public abstract void setTitle(@NotNull String title);
+ }
 diff --git a/src/main/java/org/bukkit/inventory/ItemFactory.java b/src/main/java/org/bukkit/inventory/ItemFactory.java
 index a66bec33fffd4e11d859efd8d17d274a5fa75f69..9578d6b0fa54feac75fa9a0727d878bd13b9e19e 100644
 --- a/src/main/java/org/bukkit/inventory/ItemFactory.java
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -4196,20 +4196,59 @@ diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java 
 index bfaa48cfda08c2716f61032b71d38285a9f715e8..1b526fff47e5a3d9a5325c73966dd2bf3dbbdc19 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
-@@ -72,6 +72,13 @@ public class CraftContainer extends AbstractContainerMenu {
+@@ -49,8 +49,8 @@ public class CraftContainer extends AbstractContainerMenu {
+     public CraftContainer(final Inventory inventory, final Player player, int id) {
+         this(new InventoryView() {
+ 
+-            private final String originalTitle = (inventory instanceof CraftInventoryCustom) ? ((CraftInventoryCustom.MinecraftInventory) ((CraftInventory) inventory).getInventory()).getTitle() : inventory.getType().getDefaultTitle();
+-            private String title = this.originalTitle;
++            private final net.kyori.adventure.text.Component originalTitle = (inventory instanceof CraftInventoryCustom) ? ((CraftInventoryCustom.MinecraftInventory) ((CraftInventory) inventory).getInventory()).title() : inventory.getType().defaultTitle(); // Paper - Adventure
++            private net.kyori.adventure.text.Component title = this.originalTitle; // Paper - Adventure
+ 
+             @Override
+             public Inventory getTopInventory() {
+@@ -72,20 +72,37 @@ public class CraftContainer extends AbstractContainerMenu {
                  return inventory.getType();
              }
  
-+            // Paper start
++            // Paper start - Adventure
 +            @Override
 +            public net.kyori.adventure.text.Component title() {
-+                return inventory instanceof CraftInventoryCustom ? ((CraftInventoryCustom.MinecraftInventory) ((CraftInventory) inventory).getInventory()).title() : net.kyori.adventure.text.Component.text(inventory.getType().getDefaultTitle());
++                return this.title;
 +            }
-+            // Paper end
++
++            @Override
++            public net.kyori.adventure.text.Component originalTitle() {
++                return this.originalTitle;
++            }
++
++            @Override
++            public void title(final net.kyori.adventure.text.Component title) {
++                CraftInventoryView.sendInventoryTitleChange(this, title);
++                this.title = title;
++            }
++            // Paper end - Adventure
 +
              @Override
              public String getTitle() {
-                 return title;
+-                return title;
++                return net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().serialize(this.title); // Paper - Adventure
+             }
+ 
+             @Override
+             public String getOriginalTitle() {
+-                return originalTitle;
++                return net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().serialize(this.originalTitle); // Paper - Adventure
+             }
+ 
+             @Override
+             public void setTitle(String title) {
+-                CraftInventoryView.sendInventoryTitleChange(this, title);
+-                this.title = title;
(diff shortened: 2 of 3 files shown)


imDaniX:
Just wanna point out that this will currently break existing custom `InventoryView` implementations, which will surely affect some plugins compatibility.

MiniDigger:
yeah, this needs to be done in a backwards compatible way. sadly plugins implement parts of the bukkit api themselves...

TheRealRyGuy:
In that case the component methods should just serialize back and forth and store as strings? Almost feels unnecessary but keeps compat

electronicboy:
What you’d do is add default methods which defer to the legacy methods, add add the MustOverride annotation to hint people into replacing them

imDaniX:
> Upstream just added 2 new methods to InventoryView, getOriginalTitle and setTitle, so I'm pretty sure extending this is very much unsupported API.

Yeah, I guess so, but it's actually one of rare occurrences where it mentions possibility of implementing it in the javadoc:

> Note: If you implement this interface but fail to satisfy the expected contracts of certain methods, there's no guarantee that the game will work as it should.

So I'm 50/50 about this. Upstream adding `getOriginalTitle` and `setTitle` as abstract might've been a simple oversight.
I'd say if Paper goes this way anyway, the note should just be removed, like with any other Bukkit interface.

patches/server/0010-Adventure.patch
```diff
@@ -4218,23 +4244,78 @@ index 6486a76466691f958349a4706d7c9caff9cb8f64..f3ebaefd949ae73afad3dcb69b8d9c63
...
++    private final net.kyori.adventure.text.Component originalTitle;
++    private net.kyori.adventure.text.Component title;
+ 
+     public CraftInventoryView(HumanEntity player, Inventory viewing, AbstractContainerMenu container) {
+         // TODO: Should we make sure it really IS a CraftHumanEntity first? And a CraftInventory?
+         this.player = (CraftHumanEntity) player;
+         this.viewing = (CraftInventory) viewing;
+         this.container = container;
+-        this.originalTitle = CraftChatMessage.fromComponent(container.getTitle());
++        this.originalTitle = io.papermc.paper.adventure.PaperAdventure.asAdventure(container.getTitle());
```
Machine-Maker: Add `// Paper` comments on each of these 3 line changes.

patches/api/0005-Adventure.patch
```diff
@@ -3834,36 +3834,82 @@ index 5adbe0514129abf3cfbc4b29a213f522359fe2e1..72ebc29db42d08d1d0361dba462fc8a5
...
      /**
+      * Sets the title of this inventory window to the specified title if the
+@@ -474,5 +515,6 @@ public abstract class InventoryView {
+      *
+      * @param title The new title.
+      */
+-    public abstract void setTitle(@NotNull String title);
++    public abstract void title(@NotNull net.kyori.adventure.text.Component title);
++    // Paper end - Adventure
+ }
```
Machine-Maker: The `Paper` comments need to be re-done here. Generally we want `// Paper start/end` and end comments around lines that start with `+/-` to wrap the diff. As you can see here, the "start" comment is then followed by 0 line changes for several lines. The end comment is preceded by only 1 line change, and then no changes for more lines.

patches/server/0010-Adventure.patch
```diff
@@ -4114,23 +4114,49 @@ index 3c3d5a527604d6f14abb1815816153747fcc0839..1b9a08098199f305606289724ee047e8
...
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
-@@ -72,6 +72,13 @@ public class CraftContainer extends AbstractContainerMenu {
+@@ -49,8 +49,8 @@ public class CraftContainer extends AbstractContainerMenu {
+     public CraftContainer(final Inventory inventory, final Player player, int id) {
+         this(new InventoryView() {
+ 
+-            private final String originalTitle = (inventory instanceof CraftInventoryCustom) ? ((CraftInventoryCustom.MinecraftInventory) ((CraftInventory) inventory).getInventory()).getTitle() : inventory.getType().getDefaultTitle();
+-            private String title = this.originalTitle;
++            private final net.kyori.adventure.text.Component originalTitle = inventory instanceof CraftInventoryCustom ? ((CraftInventoryCustom.MinecraftInventory) inventory).title() : inventory.getType().defaultTitle(); // Paper
++            private net.kyori.adventure.text.Component title = this.originalTitle;
```
Machine-Maker: `// Paper`

patches/server/0010-Adventure.patch
```diff
@@ -4114,23 +4114,49 @@ index 3c3d5a527604d6f14abb1815816153747fcc0839..1b9a08098199f305606289724ee047e8
...
++                return this.originalTitle;
+             }
+ 
+             @Override
+-            public void setTitle(String title) {
++            public void title(net.kyori.adventure.text.Component title) {
+                 CraftInventoryView.sendInventoryTitleChange(this, title);
+                 this.title = title;
+             }
++            //Paper end - Adventure
```
Machine-Maker: Same situation here. Use the `// Paper` comments more strictly instead of wrapping lots of unchanged lines.

patches/server/0010-Adventure.patch
```diff
@@ -4218,23 +4244,78 @@ index 6486a76466691f958349a4706d7c9caff9cb8f64..f3ebaefd949ae73afad3dcb69b8d9c63
...
+         return this.originalTitle;
+     }
+ 
+     @Override
+-    public void setTitle(String title) {
++    public void title(net.kyori.adventure.text.Component title) {
+         CraftInventoryView.sendInventoryTitleChange(this, title);
+         this.title = title;
+     }
++    // Paper end - Adventure
```
Machine-Maker: Same here

patches/server/0010-Adventure.patch
```diff
@@ -4218,23 +4244,78 @@ index 6486a76466691f958349a4706d7c9caff9cb8f64..f3ebaefd949ae73afad3dcb69b8d9c63
...
++    public static void sendInventoryTitleChange(InventoryView view, net.kyori.adventure.text.Component title) { //Paper - Adventure
+         Preconditions.checkArgument(view != null, "InventoryView cannot be null");
+         Preconditions.checkArgument(title != null, "Title cannot be null");
+         Preconditions.checkArgument(view.getPlayer() instanceof Player, "NPCs are not currently supported for this function");
+@@ -106,7 +108,12 @@ public class CraftInventoryView extends InventoryView {
+         final ServerPlayer entityPlayer = (ServerPlayer) ((CraftHumanEntity) view.getPlayer()).getHandle();
+         final int containerId = entityPlayer.containerMenu.containerId;
+         final MenuType<?> windowType = CraftContainer.getNotchInventoryType(view.getTopInventory());
+-        entityPlayer.connection.send(new ClientboundOpenScreenPacket(containerId, windowType, CraftChatMessage.fromString(title)[0]));
++        entityPlayer.connection.send(new ClientboundOpenScreenPacket(containerId, windowType, io.papermc.paper.adventure.PaperAdventure.asVanilla(title)));
```
Machine-Maker: `// Paper`

Machine-Maker (changes requested):

TheRealRyGuy:
Paper comments look good to go hopefully? idk why i'm so bad at these

lynxplay:
The title changing generally behaves rather inconsistently when called in a InventoryClickEvent.
Here are a few run downs:
```java
    event.getView().title(Component.text(Instant.now().toString()));
}, 1);
``` 
causes the client to loose the ability to collect items of the same type via double click.

```java
event.getView().title(Component.text(Instant.now().toString()));
```
completely messes with the client server side sync, causing either desyncs in the cursor or simple inability to pick up/put down the item.

```java
event.setCancelled(true);
event.getView().title(Component.text(Instant.now().toString()));
```
works well, the only combination of states that properly works.

Generally, I think this method needs a bit of a larger discussion, I already touched on this when this PR was proposed to paper, imo this logic barely deserves to live in the API, at best in UnsafeValues, tho that is just my opinion.

broken1arrow:
> The title changing generally behaves rather inconsistently when called in a InventoryClickEvent. Here are a few run downs:
> 
> ```java
> Bukkit.getScheduler().runTaskLater(this, () -> {
>     event.getView().title(Component.text(Instant.now().toString()));
> }, 1);
> ```
> 
> causes the client to loose the ability to collect items of the same type via double click.
> 
> ```java
> event.getView().title(Component.text(Instant.now().toString()));
> ```
> 
> completely messes with the client server side sync, causing either desyncs in the cursor or simple inability to pick up/put down the item.
> 
> ```java
> event.setCancelled(true);
> event.getView().title(Component.text(Instant.now().toString()));
> ```
> 
> works well, the only combination of states that properly works.
> 
> Generally, I think this method needs a bit of a larger discussion, I already touched on this when this PR was proposed to paper, imo this logic barely deserves to live in the API, at best in UnsafeValues, tho that is just my opinion.

I have use my own nms way for this and my own experience is you have to send the update async for the title. I have not experiences any desync or item dupe with my way. So players can add or remove items without get stuck in a "loop". 

So what is it I do then, I not modify the actual inventory or the original title at all. What I do is trick the client to think the inventory have new title, but it is far from perfect and if you accentually detect wrong type or size it will be glitchy client side. What is happening is this items from players inventory ends up in wrong inventory and player can lose the items. 

I started a tread for a long time ago here and even show how I did it. I could show a video how my code works.

Grabsky:
> Generally, I think this method needs a bit of a larger discussion, I already touched on this when this PR was proposed to paper, imo this logic barely deserves to live in the API, at best in UnsafeValues, tho that is just my opinion.

Has there been discussion about that ever since?

Having access to Bukkit's `setTitle(String)` via the `InventoryView` class and Paper's `title(Component)` via `UnsafeValues` doesn't sound like a reasonable thing to do. Both methods should be in one place.

Despite all issues this can cause when called on `InventoryClickEvent`, which I do believe is also the case for `setTitle(String)` method, I think it still makes sense to add this for those who want to use it. A warning could be added to make sure people are aware of any potential side-effects that may come with using it.

<br />

> So what is it I do then, I not modify the actual inventory or the original title at all. What I do is trick the client to think the inventory have new title, but it is far from perfect and if you accentually detect wrong type or size it will be glitchy client side. What is happening is this items from players inventory ends up in wrong inventory and player can lose the items.

I've been doing that prior `InventoryTitle#setTitle(String)` was added and since I don't call it without cancelling the `InventoryClickEvent` first, I wasn't aware of any desyncs and other glitches that may happen. Either way, I think implementation of this PR should not be different from one introduced by CraftBukkit.

Machine-Maker:
> Despite all issues this can cause when called on InventoryClickEvent

We don't want to add more API that is inherently broken. I'm leaning towards deprecating the setTitle(String) method with the notice that it is just broken and shouldn't be relied upon.

>  I wasn't aware of any desyncs and other glitches that may happen. 

Yeah, they certainly exist as lynxplay pointed out. Double-clicking to collect items of the same type just doesn't work.

If we can show that cancelling + changing the title has some effect, I think separate API should be introduced to do just that. Force cancel the event + change the title so the 2 things **cannot** be separated. That can't be done with a method on InventoryView.

broken1arrow:
> > Generally, I think this method needs a bit of a larger discussion, I already touched on this when this PR was proposed to paper, imo this logic barely deserves to live in the API, at best in UnsafeValues, tho that is just my opinion.
> 
> Has there been discussion about that ever since?
> 
> Having access to Bukkit's `setTitle(String)` via the `InventoryView` class and Paper's `title(Component)` via `UnsafeValues` doesn't sound like a reasonable thing to do. Both methods should be in one place.
> 
> Despite all issues this can cause when called on `InventoryClickEvent`, which I do believe is also the case for `setTitle(String)` method, I think it still makes sense to add this for those who want to use it. A warning could be added to make sure people are aware of any potential side-effects that may come with using it.
> 
> 
> > So what is it I do then, I not modify the actual inventory or the original title at all. What I do is trick the client to think the inventory have new title, but it is far from perfect and if you accentually detect wrong type or size it will be glitchy client side. What is happening is this items from players inventory ends up in wrong inventory and player can lose the items.
> 
> I've been doing that prior `InventoryTitle#setTitle(String)` was added and since I don't call it without cancelling the `InventoryClickEvent` first, I wasn't aware of any desyncs and other glitches that may happen. Either way, I think implementation of this PR should not be different from one introduced by CraftBukkit.

I talking about if you implement that function self and what I seen does this api use same logic and methods as I do. You need to make sure you get right inventory and something I guess the set tittle in the api do.

When I implemented this I have to call this asynchronous or player get stuck in clicking on a item. What I experienced is you have to send a update package to player (after modify title) and that could in theory create issues. 

Just for you don't do it synchrony and at least I think you could dupe items in theory or item loss. Because the client and server side inventory is not syncing.

Cancel the event could reduce the problem, but not remove it. 

Only way I think is modify the original NMS code, so title is not hardwired to the inventory self. Only time I could say it is safe to change title is when you open inventory not after.

Grabsky:
> We don't want to add more API that is inherently broken. I'm leaning towards deprecating the setTitle(String) method with the notice that it is just broken and shouldn't be relied upon.

I mean, I understand that reliability of this method is questionable, is there anything that can be done on the server-side to fix it?

Also wouldn't call these methods "broken" - they do serve a purpose and can occasionally break when used in the wrong place, or at the wrong time. This applies to number of other methods within the Bukkit API and it's usually stated in javadocs.

> If we can show that cancelling + changing the title has no negative effect and is reliable, I think separate API should be introduced to do just that. Force cancel the event + change the title so the 2 things cannot be separated. That can't be done with a method on InventoryView.

Any idea of what this could be? Because when I'm thinking of `InventoryClickEvent#changeTitle(Component)` or similar, that doesn't really appeal to me. There are number of places such functionality may be desired to be used, including those that are not associated with any events - eg. animated titles using repeating task.

Machine-Maker:
> is there anything that can be done on the server-side to fix it?

I don't know, and I don't really care to try and hack a solution in to fix some of the issues. The game's protocol just flat out doesn't support changing titles without re-creating and re-sending the entire inventory open packet which makes the client create a whole new screen and change to it.

Based on my quick understanding of how double-click works, it is tracked on the client, and if the current screen sees 2 clicks within a certain time span, it then sends a packet telling the server to collect all items to the cursor. But that's not possible if you change the title because it is a new instance of the client's screen so it resets the double-click tracking.

> Also wouldn't call these methods "broken" - they do serve a purpose 

Having *any* effect doesn't mean they also aren't broken. 

> can occasionally break when used in the wrong place, or at the wrong time. This applies to number of other methods within the Bukkit API and it's usually stated in javadocs.

This is bad in the first place, we shouldn't add more API that this statement applies to. Just because API exists that this statement applies to, doesn't mean we should keep adding API in this manner.

lynxplay:
Since https://github.com/PaperMC/Paper/pull/11309, paper officially deprecated upstreams method (see the PR and this PR for reasaons)
I'll be closing this as we don't need to introduce adventure methods for deprecated upstream API.

PR #9331, opened 2023-06-12
Suppress Item Meta Validation Checks

### LINKED ISSUE #9314: Cannot load world with a player wearing armor with an invalid armor trim
### Stack trace

```
java.lang.IllegalArgumentException: pattern must not be null
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:145)
	at org.bukkit.inventory.meta.trim.ArmorTrim.<init>(ArmorTrim.java:27)
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftMetaArmor.<init>(CraftMetaArmor.java:74)
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftMetaColorableArmor.<init>(CraftMetaColorableArmor.java:32)
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftItemStack.getItemMeta(CraftItemStack.java:340)
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftItemStack.getItemMeta(CraftItemStack.java:287)
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftItemStack.asBukkitCopy(CraftItemStack.java:62)
	at net.minecraft.world.entity.LivingEntity.collectEquipmentChanges(LivingEntity.java:3114)
	at net.minecraft.world.entity.LivingEntity.detectEquipmentUpdates(LivingEntity.java:3076)
	at net.minecraft.world.entity.LivingEntity.tick(LivingEntity.java:3002)
	at net.minecraft.world.entity.player.Player.tick(Player.java:276)
	at net.minecraft.server.level.ServerPlayer.doTick(ServerPlayer.java:758)
	at net.minecraft.server.players.PlayerList.remove(PlayerList.java:602)
	at net.minecraft.server.players.PlayerList.remove(PlayerList.java:584)
	at net.minecraft.server.network.ServerGamePacketListenerImpl.onDisconnect(ServerGamePacketListenerImpl.java:2119)
	at net.minecraft.server.network.ServerGamePacketListenerImpl.onDisconnect(ServerGamePacketListenerImpl.java:2098)
	at net.minecraft.network.Connection.handleDisconnection(Connection.java:797)
	at net.minecraft.server.network.ServerConnectionListener.tick(ServerConnectionListener.java:254)
	at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1575)
	at net.minecraft.server.dedicated.DedicatedServer.tickChildren(DedicatedServer.java:447)
	at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1396)
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1173)
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:320)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

### Plugin and Datapack List

None

* Give yourself armor with an invalid trim:  
`/give @p minecraft:leather_chestplate{Trim:{material:"minecraft:abcd",pattern:"minecraft:abcd"}}`
* Equip the armor piece
* Restart the server after crash

### Paper version

> version
[12:23:29 INFO]: Checking version, please wait...
[12:23:29 INFO]: This server is running Paper version git-Paper-10 (MC: 1.20) (Implementing API version 1.20-R0.1-SNAPSHOT) (Git: f4dfdcb)
Previous version: git-Paper-8 (MC: 1.20)

### Other

If the trim is added using a datapack, the crash no longer occurs

Owen1212055:
Fixes https://github.com/PaperMC/Paper/issues/9314

The underlying issue is item meta, but getting rid of that implementation is future plans. 😄


patches/server/0978-Suppress-Item-Meta-Validation-Checks.patch
```diff
@@ -0,0 +1,42 @@
...
+@@ -68,10 +68,12 @@ public class CraftMetaArmor extends CraftMetaItem implements ArmorMeta {
+             CompoundTag trimCompound = tag.getCompound(TRIM.NBT);
+ 
+             if (trimCompound.contains(TRIM_MATERIAL.NBT, net.minecraft.nbt.Tag.TAG_STRING) && trimCompound.contains(TRIM_PATTERN.NBT, net.minecraft.nbt.Tag.TAG_STRING)) { // Paper - for now, ignore inline definitions of trim material & pattern
++                try { // Paper - Ignore invalid registries
+                 TrimMaterial trimMaterial = Registry.TRIM_MATERIAL.get(NamespacedKey.fromString(trimCompound.getString(TRIM_MATERIAL.NBT)));
+                 TrimPattern trimPattern = Registry.TRIM_PATTERN.get(NamespacedKey.fromString(trimCompound.getString(TRIM_PATTERN.NBT)));
+ 
+                 this.trim = new ArmorTrim(trimMaterial, trimPattern);
++                } catch (Exception ignored) {} // Paper
```
Machine-Maker: I'm not sure I really like this. We shouldn't catch "all" exceptions as that could hide stuff. But just handle the key being invalid for each or the mat or pattern being null.

Machine-Maker (changes requested):

patches/server/0978-Suppress-Item-Meta-Validation-Checks.patch
Owen1212055: I mean any errors that would be occurring here anyways shouldn’t happen anyways, as there is no reason why arbitrary NBT should throw exceptions like this. And these are really temporary fixes.

I mean I’m happy to more fine tune it, but I just don’t see the point.

Should I:
- Only catch certain exception types (NPE, illegal arg ex)
- Change logic to check if the key is null
- Leave it be

I’m open, curious what others think too.

Machine-Maker: I think you check if key is null, and then check if returned trimMaterial/Pattern is null. If any of them are null, you just don't create the ArmorTrim instance and set it to the `trim` field.

Machine-Maker (approved):

PR #9334, opened 2023-06-13
Don't enforce icanhasbukkit default if alias block exists

### LINKED ISSUE #9275: Commands.yml bug
### Expected behavior

I should be able to remove "icanhasbukkit" command from commands.yml

### Observed/Actual behavior

Everytime I remove that command from commands.yml file it's re-added after a server server. The only way is to completely disable the aliases thing by adding "[]" (aliases: []) but I don't want to disable it cause I'm using it for other aliases.

### Steps/models to reproduce

Remove   icanhasbukkit: - version $1- and add any other alias

### Plugin and Datapack List

-

### Paper version

I already reported to Pufferfish but they said to report it to upstream (Paper)

[03:35:34 INFO]: This server is running Pufferfish version git-Pufferfish-70 (MC: 1.19.4) (Implementing API version 1.19.4-R0.1-SNAPSHOT) (Git: 19c244b on HEAD)

### Other

_No response_

TheMeinerLP:
Fixes #9275 

### Test 1:
1. Start a fresh server
2. File get created
3. Modify file/delete entry and check if `icanhasbukkit` again there
### Test 2:
1. Start a already created server with that file
2. Delete it 
3. Restart server
4. Check if file again back

Both tests are functional


electronicboy:
Missing paper comments and should inline imports, but, this also just prevents us from being able to add new things in there in general, which idk if that’s a concern/issue or not

lynxplay:
Yea, preferably we'd just be yanking the `aliases` section from the defaults configuration loaded from the jar if it exists in the file on the disk. That way we still maintain the ability to add new defaults etc but `aliases: {}` would properly define it as an
empty section without it being overwritten down the line.

TheMeinerLP:
I hope thats now better.

lynxplay:
I mean, just `set("aliases", null)` on the defaults yaml configuration loaded from the jar if the one loaded from the disk contains a `aliases` block.

TheMeinerLP:
@lynxplay I added your changes so far

patches/server/0978-Add-a-check-if-the-file-exists-to-prevent-to-overrid.patch
```diff
@@ -0,0 +1,29 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+index ffa27c9c02dc4d12411fc089de3af8e8e12ba06e..467d2cfd1b48b7816b9a167376feea7a6afff8f7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -350,11 +350,14 @@ public final class CraftServer implements Server {
+         if (this.getCommandsConfigFile().isFile()) {
+             legacyAlias = null;
+         }
++
```
lynxplay: Useless diff.

patches/server/0978-Add-a-check-if-the-file-exists-to-prevent-to-overrid.patch
```diff
@@ -0,0 +1,29 @@
...
+         this.commandsConfiguration = YamlConfiguration.loadConfiguration(this.getCommandsConfigFile());
+-        this.commandsConfiguration.options().copyDefaults(true);
+-        this.commandsConfiguration.setDefaults(YamlConfiguration.loadConfiguration(new InputStreamReader(getClass().getClassLoader().getResourceAsStream("configurations/commands.yml"), Charsets.UTF_8)));
++        // Paper start - don't enforce icanhasbukkit default if alias block exists
++        final YamlConfiguration commandsDefaults = YamlConfiguration.loadConfiguration(new InputStreamReader(getClass().getClassLoader().getResourceAsStream("configurations/commands.yml"), Charsets.UTF_8));
++        if (this.commandsConfiguration.contains("aliases")) commandsDefaults.set("aliases", null);
++        this.commandsConfiguration.setDefaults(commandsDefaults);
++        // Paper stop - dont enforce icanhasbukkit default if alias block exists
+         this.saveCommandsConfig();
+-
```
lynxplay: Useless diff.

lynxplay (changes requested):

patches/server/0978-Add-a-check-if-the-file-exists-to-prevent-to-overrid.patch
```diff
@@ -0,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheMeinerLP <p.glanz@madfix.me>
+Date: Tue, 13 Jun 2023 16:10:59 +0200
+Subject: [PATCH] Add a check if the file exists to prevent to override the
```
lynxplay: Incorrect patch subject. This should describe what the patch does, not how.
`Add a check if the file exists to prevent to override the file each time` says absolutely nothing :sweat_smile:

TheMeinerLP: I will be change that. Thanks for notice that

patches/server/0978-Don-t-enforce-icanhasbukkit-default-if-alias-block-e.patch
```diff
@@ -0,0 +1,24 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+index ffa27c9c02dc4d12411fc089de3af8e8e12ba06e..30cad700d5278a2c5e79eecfd526c7ceb13aa5bd 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+@@ -351,8 +351,11 @@ public final class CraftServer implements Server {
+             legacyAlias = null;
+         }
+         this.commandsConfiguration = YamlConfiguration.loadConfiguration(this.getCommandsConfigFile());
+-        this.commandsConfiguration.options().copyDefaults(true);
```
lynxplay: This line cannot be removed, defaults still have to be copied, we just update the defaults to not include the icanhasbukkit block if aliases exists.

lynxplay (approved):

lynxplay:
Needs a rebase from a potential merger before merging if pulled.

TheMeinerLP:
Ready to merge. Rebased on master

electronicboy (approved):

PR #9339, opened 2023-06-14
OperatorInteractEvents - WIP

Machine-Maker:
This event provides a means for granular control over what "operator blocks" can be interacted by a player. Just an initial design, nothing is concrete like the event name, what type of event it is, or where it should be called. Only called in 2 example places for now.


molor:
I vote for `OperatorBlockInteractEvent` just because full words looks better Imo.

Machine-Maker:
Ok, went for `OperatorInteractEvent` with specific subclasses for Block and Entity since minecart command blocks also exist.

Landy005:
how to enable this? need give everyone in the server rights to change light blocks

PR #9347, opened 2023-06-15
Add clickable version on version command

TheMeinerLP:
Allows plugin developers to get more easy the version for there bug reports.
I hope minimessage is here the way to go.

Just for example:
image


patches/api/0418-Add-clickable-version-on-version-command.patch
```diff
@@ -0,0 +1,37 @@
...
+diff --git a/src/main/java/org/bukkit/command/defaults/VersionCommand.java b/src/main/java/org/bukkit/command/defaults/VersionCommand.java
+index e40f017f87d6b6b4770501b106c76dc69ec69abb..5c8274617d1419c53cc88b6ae8430097291fc141 100644
+--- a/src/main/java/org/bukkit/command/defaults/VersionCommand.java
++++ b/src/main/java/org/bukkit/command/defaults/VersionCommand.java
+@@ -18,6 +18,10 @@ import java.util.HashSet;
+ import java.util.List;
+ import java.util.Set;
+ import java.util.concurrent.locks.ReentrantLock;
++
++import net.kyori.adventure.text.minimessage.MiniMessage;
```
lynxplay: Inline imports.

patches/api/0418-Add-clickable-version-on-version-command.patch
```diff
@@ -0,0 +1,37 @@
...
+ import org.bukkit.Bukkit;
+ import org.bukkit.ChatColor;
+ import org.bukkit.command.CommandSender;
+@@ -89,7 +93,14 @@ public class VersionCommand extends BukkitCommand {
+ 
+     private void describeToSender(@NotNull Plugin plugin, @NotNull CommandSender sender) {
+         PluginDescriptionFile desc = plugin.getDescription();
+-        sender.sendMessage(ChatColor.GREEN + desc.getName() + ChatColor.WHITE + " version " + ChatColor.GREEN + desc.getVersion());
++        // Paper start - Helps plugins developer to get fast plugin versions in game
++        var tags = TagResolver.builder()
```
lynxplay: i don#t see the point here for mini message (beyond the fact that its usage here is less than idea, why are we explicitly creating a tag resolver instead of just passing two `Placeholder.component` tag resolvers directly to deserialise.

```java
Component.text()
    .append(Component.text(name, NamedTextColor.GREEN))
    .append(Component.text(" version "))
    .append(Component.text(version, NamedTextColor.GREEN).clickEvent(ClickEvent.copyToClipboard(
        version
    )));
```
does the same job.

lynxplay (changes requested):
This should be part of https://github.com/PaperMC/Paper/blob/master/patches/api/0014-Version-Command-2.0.patch

patches/api/0418-Add-clickable-version-on-version-command.patch
MiniDigger: I agree, minimessage only would make sense here if its a configurable message, else it just adds complexity

TheMeinerLP: I currently on it to implement it with the builder struct

patches/api/0418-Add-clickable-version-on-version-command.patch
MiniDigger: if using components, stuff with inlined imports will be really really ugly :/

lynxplay: True, I guess we can open an import block here.

lynxplay: Yea, the suggested patch this goes into also has an import. So moving that one, together with the adventure imports into an import block either at the top or bottom of the imports would also work.

Not that this command should see drastic changes across version updates anyway.

patches/api/0014-Version-Command-2.0.patch
```diff
@@ -108,7 +108,26 @@ index 04b4fb6859df0221f8f9f92c5a7ac2dda1073355..b50f614806f4634960d383e8a33f094c
              sendVersion(sender);
          } else {
              StringBuilder name = new StringBuilder();
-@@ -146,14 +156,14 @@ public class VersionCommand extends BukkitCommand {
+@@ -79,7 +89,17 @@ public class VersionCommand extends BukkitCommand {
+ 
+     private void describeToSender(@NotNull Plugin plugin, @NotNull CommandSender sender) {
+         PluginDescriptionFile desc = plugin.getDescription();
+-        sender.sendMessage(ChatColor.GREEN + desc.getName() + ChatColor.WHITE + " version " + ChatColor.GREEN + desc.getVersion());
```
lynxplay, line 118: Missing paper comments.

lynxplay (changes requested):
As mini said, this patch already has an import statement, so you could open a import block (e.g.
```java
// Paper start - version command 2.0
import .....
import ....
// Paper end - version command 2.0
```

Located at the end of the upstream import statements.

patches/api/0014-Version-Command-2.0.patch
```diff
@@ -73,18 +73,22 @@ index f68ef89f37057cf677a767026ab395f7a839a2f9..a6aa33b9574d0278e10927007a62290e
...
++import net.kyori.adventure.text.Component;
++import net.kyori.adventure.text.format.NamedTextColor;
++import net.kyori.adventure.text.event.ClickEvent;
++import net.kyori.adventure.text.format.TextDecoration;
++import net.kyori.adventure.text.serializer.plain.PlainComponentSerializer;
++import net.kyori.adventure.text.TextComponent;
++// Paper end - version command 2.0
  
  public class VersionCommand extends BukkitCommand {
 +    private VersionFetcher versionFetcher;
```
yannicklamprecht:
```suggestion
+    private VersionFetcher versionFetcher; // Paper - version command 2.0
```

TheMeinerLP: This was already before

lynxplay: Can still fix it 😓

yannicklamprecht (approved):

TheMeinerLP:
Something else from me needed ?

TheMeinerLP:
#push

Lulu13022002:
I think a tooltip on hover is needed otherwise it's kinda hidden especially since it's a normal sentence

TheMeinerLP:
> I think a tooltip on hover is needed otherwise it's kinda hidden especially since it's a normal sentence

I will add a tooltip. Thanks for that feedback

TheMeinerLP:
End result via test plugin: 
image

Machine-Maker (approved):

TheMeinerLP:
#push

patches/api/0014-Version-Command-2.0.patch
```diff
@@ -158,18 +182,18 @@ index 04b4fb6859df0221f8f9f92c5a7ac2dda1073355..b50f614806f4634960d383e8a33f094c
...
 -    private void setVersionMessage(@NotNull String msg) {
 +    // Paper start
-+    private void setVersionMessage(final @NotNull net.kyori.adventure.text.Component msg) {
++    private void setVersionMessage(final @NotNull Component msg) {
          lastCheck = System.currentTimeMillis();
 -        versionMessage = msg;
-+        final net.kyori.adventure.text.Component message = net.kyori.adventure.text.TextComponent.ofChildren(
-+            net.kyori.adventure.text.Component.text("This server is running " + Bukkit.getName() + " version " + Bukkit.getVersion() + " (Implementing API version " + Bukkit.getBukkitVersion() + ")", net.kyori.adventure.text.format.NamedTextColor.WHITE),
-+            net.kyori.adventure.text.Component.newline(),
++        final Component message = TextComponent.ofChildren(
```
lynxplay: ofChildren is deprecated for removal. This PR should probably just replace it with the alternative defined in the jd.

patches/api/0014-Version-Command-2.0.patch
```diff
@@ -158,18 +182,18 @@ index 04b4fb6859df0221f8f9f92c5a7ac2dda1073355..b50f614806f4634960d383e8a33f094c
...
++            Component.newline(),
 +            msg
 +        );
-+        this.versionMessage = net.kyori.adventure.text.Component.text()
++        this.versionMessage = Component.text()
 +            .append(message)
-+            .hoverEvent(net.kyori.adventure.text.Component.text("Click to copy to clipboard", net.kyori.adventure.text.format.NamedTextColor.WHITE))
-+            .clickEvent(net.kyori.adventure.text.event.ClickEvent.copyToClipboard(net.kyori.adventure.text.serializer.plain.PlainComponentSerializer.plain().serialize(message)))
++            .hoverEvent(Component.text("Click to copy to clipboard", NamedTextColor.WHITE))
++            .clickEvent(ClickEvent.copyToClipboard(PlainComponentSerializer.plain().serialize(message)))
```
lynxplay: PlainComponentSerializer is also deprecated for removal, should just be replaced with the  PlainTextComponentSerializer

lynxplay (changes requested):

lynxplay (approved):

PR #9349, opened 2023-06-15
Fix multiple spectator bugs

FriwiDev:
This is a small contribution from the PlayLegend server team with multiple issues resolved we had concerning player spectating. We started using this patch in production in 1.19.3. It is already in use for a long time and tested to be working flawlessly.

Fixes MC-261799: Spectator mode not teleporting viewers to targets in other worlds or viewers bug around (stuck) when a target changes world
Fixes MC-107113: Spectator mode viewers stuck when target is teleported more than viewing distance
Fixes MC-148993: Chunks go invisible when target moves too far over time while viewing (render POI on client is not updated)


NoahvdAa:
fwiw, the untracked issue is probably [MC-148993](https://bugs.mojang.com/browse/MC-148993)

FriwiDev:
Oh wow, you were fast on that one, I'm adding it :)

Issue #9316, opened 2023-06-12
Entity tab-complete doesn't ignore spectator target


molor:
### Expected behavior

Any vanilla command (for example, `data merge entity ...`) should always suggest a UUID of the entity you or your spectator target is looking at.

### Observed/Actual behavior

If your spectator target is `null`, the server will suggest you a target entity UUID.
BUT if your spectator target isn't `null`, then the server will always suggest you a spectator target UUID.

Moreover, your spectator target may also have other spectator target `(e.g. you (spectator) -> other_player (spectator) -> other_other_player (survival))`, so in case of fixing this we need to take this into account.

### Steps/models to reproduce

1. GameMode.SPECTATOR
2. Run `spectate other_player` command or try to attack *other_player*
3. Make *other_player* look at *other_entity*, for example at a random pig
4. Type `data merge entity ...`

### Plugin and Datapack List

Fresh install of Paper w/o any plugins or data packs.

### Paper version

`1.19.4-550`

### Other

Didn't tried 1.20 yet, but I think it might be even a vanilla issue. IIRC it is also present in 1.18.2.

PR #9350, opened 2023-06-15
Fix multiple spectator bugs

FriwiDev:
This is a small contribution from the PlayLegend server team with multiple issues resolved we had concerning player spectating. We started using this patch in production in 1.19.3. It is already in use for a long time and tested to be working flawlessly.

Fixes MC-261799: Spectator mode not teleporting viewers to targets in other worlds or viewers bug around (stuck) when a target changes world
Fixes MC-107113: Spectator mode viewers stuck when target is teleported more than viewing distance
Fixes MC-148993: Chunks go invisible when target moves too far over time while viewing (render POI on client is not updated)

### DIFF
--- a/patches/server/0977-Fix-multiple-spectator-bugs.patch
+++ b/patches/server/0977-Fix-multiple-spectator-bugs.patch
@@ -1,0 +1,53 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Fritz Windisch <friwidev@gmail.com>
+Date: Thu, 15 Jun 2023 17:10:22 +0200
+Subject: [PATCH] Fix multiple spectator bugs
+
+Fixes MC-261799: Spectator mode not teleporting viewers to targets in other worlds or viewers bug around (stuck) when a target changes world
+Fixes MC-107113: Spectator mode viewers stuck when target is teleported more than viewing distance
+Fixes untracked: Chunks go invisible when target moves too far over time while viewing (render POI on client is not updated)
+
+diff --git a/src/main/java/net/minecraft/server/level/ChunkMap.java b/src/main/java/net/minecraft/server/level/ChunkMap.java
+index 9209b598d7168b82574e4800056b8b9f84265dd0..2b37ccc8e2619677179b17a6c5bd758bd5f95150 100644
+--- a/src/main/java/net/minecraft/server/level/ChunkMap.java
++++ b/src/main/java/net/minecraft/server/level/ChunkMap.java
+@@ -1535,6 +1535,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+                 double d1 = vec3d_dx * vec3d_dx + vec3d_dz * vec3d_dz; // Paper
+                 double d2 = d0 * d0;
+                 boolean flag = d1 <= d2 && this.entity.broadcastToPlayer(player);
++                flag |= player.getCamera() == this.entity; // Paper - Make entities always visible for spectators, even when teleported far away (MC-107113)
+ 
+                 // CraftBukkit start - respect vanish API
+                 if (!player.getBukkitEntity().canSee(this.entity.getBukkitEntity())) {
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 9d46536f80b5b3e6641fd377c02166a431edfd77..bfe4d87835a4979fa03abac532b82f5f65dae5d7 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -732,8 +732,25 @@ public class ServerPlayer extends Player {
+ 
+         if (entity != this) {
+             if (entity.isAlive()) {
+-                this.absMoveTo(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot());
+-                this.serverLevel().getChunkSource().move(this);
++                // Paper start - Fix spectator on cross-world teleports (MC-261799)
++                if (entity.level() != this.level()) {
++                    // Teleport ourselves to our camera
++                    this.getBukkitEntity().teleport(entity.getBukkitEntity().getLocation(), TeleportCause.SPECTATE);
++                    // Update the tracker of the other dimension for our cross-dimension teleport
++                    ChunkMap.TrackedEntity tracker = ((ServerLevel) entity.level()).getChunkSource().chunkMap.entityMap.get(entity.getId());
++                    if (tracker != null) {
++                        tracker.updatePlayer(this);
++                    }
++                    // Advise the client to start spectating again
++                    this.connection.send(new ClientboundSetCameraPacket(entity));
++                }else {
++                    // Paper: We send the player an additional teleport packet here to indicate that the position of itself has been moved.
++                    // Without this packet, if a player travels a too far distance, chunks will start to become invisible for our spectator.
++                    this.connection.teleport(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot(), TeleportCause.SPECTATE);
++                    this.absMoveTo(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot());
++                    this.serverLevel().getChunkSource().move(this);
++                }
++                // Paper end
+                 if (this.wantsToStopRiding()) {
+                     this.setCamera(this);
+                 }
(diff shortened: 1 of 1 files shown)


nopjar:
> (maybe include #9316?)

Seems like a completely different issue. Should be a different patch imo.
Your issue has to do with tab completion, not spectating behavior itself.

FriwiDev:
This patch does not address any transitive spectator issues. It only fixes the issues mentioned above.

S3vro (approved):

nopjar (approved):

lennyrip (approved):

molor (approved):

IBims1ckoky (approved):

Owen1212055:
Review spam does not help.

patches/server/0977-Fix-multiple-spectator-bugs.patch
```diff
@@ -0,0 +1,53 @@
...
++                    ChunkMap.TrackedEntity tracker = ((ServerLevel) entity.level()).getChunkSource().chunkMap.entityMap.get(entity.getId());
++                    if (tracker != null) {
++                        tracker.updatePlayer(this);
++                    }
++                    // Advise the client to start spectating again
++                    this.connection.send(new ClientboundSetCameraPacket(entity));
++                }else {
++                    // Paper: We send the player an additional teleport packet here to indicate that the position of itself has been moved.
++                    // Without this packet, if a player travels a too far distance, chunks will start to become invisible for our spectator. (MC-148993)
++                    this.connection.teleport(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot(), TeleportCause.SPECTATE);
```
kennytv: This will result in teleport events being spammed constantly

kennytv (changes requested):

patches/server/0977-Fix-multiple-spectator-bugs.patch
FriwiDev: Fixed, sorry for this. I patched it by using internalTeleport instead. The case for cross-world teleporting should be fine because it is only called once per world change.

patches/server/0977-Fix-multiple-spectator-bugs.patch
```diff
@@ -0,0 +1,61 @@
...
+ import java.util.Iterator;
+ import java.util.List;
+ import java.util.Objects;
+@@ -732,8 +733,25 @@ public class ServerPlayer extends Player {
+ 
+         if (entity != this) {
+             if (entity.isAlive()) {
+-                this.absMoveTo(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot());
+-                this.serverLevel().getChunkSource().move(this);
++                // Paper start - Fix spectator on cross-world teleports (MC-261799)
```
Lulu13022002, line 40: This is not the right fix and doesn't work for me. The reason it doesn't work is that mojang explicitly prevent spectating across dimension and you can easily see it in the following line: ServerPlayer#teleportTo L2225~ (line doesn't match since i have modified the source). The way you would fix that (i think) would be either:
* by changing the ServerPlayer#setCamera saving the camera before the world tp and then resetting the camera back to what we expect here but then the spectator events doesn't really work.

* or by simply ignoring this restriction only for this specific call (all the teleportTo called with the bukkit spectate reason) normally this doesn't collide with the fact that plugin can change the teleport reason since they doesn't seems to use this method). I recommend this one.

You can then update the world where chunkSource#move is called.

patches/server/0977-Fix-multiple-spectator-bugs.patch
```diff
@@ -0,0 +1,61 @@
...
++                    ChunkMap.TrackedEntity tracker = ((ServerLevel) entity.level()).getChunkSource().chunkMap.entityMap.get(entity.getId());
++                    if (tracker != null) {
++                        tracker.updatePlayer(this);
++                    }
++                    // Advise the client to start spectating again
++                    this.connection.send(new ClientboundSetCameraPacket(entity));
++                }else {
++                    // Paper: We send the player an additional teleport packet here to indicate that the position of itself has been moved.
++                    // Without this packet, if a player travels a too far distance, chunks will start to become invisible for our spectator. (MC-148993)
++                    this.connection.internalTeleport(entity.getX(), entity.getY(), entity.getZ(), entity.getYRot(), entity.getXRot(), Collections.emptySet());
```
Lulu13022002: This really spam, i think this should be fixed in the client but if maintainer are ok, at least check if the player has moved in another chunk before sending another packet

patches/server/0977-Fix-multiple-spectator-bugs.patch
```diff
@@ -0,0 +1,61 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/level/ChunkMap.java b/src/main/java/net/minecraft/server/level/ChunkMap.java
+index 9209b598d7168b82574e4800056b8b9f84265dd0..2b37ccc8e2619677179b17a6c5bd758bd5f95150 100644
+--- a/src/main/java/net/minecraft/server/level/ChunkMap.java
++++ b/src/main/java/net/minecraft/server/level/ChunkMap.java
+@@ -1535,6 +1535,7 @@ public class ChunkMap extends ChunkStorage implements ChunkHolder.PlayerProvider
+                 double d1 = vec3d_dx * vec3d_dx + vec3d_dz * vec3d_dz; // Paper
+                 double d2 = d0 * d0;
+                 boolean flag = d1 <= d2 && this.entity.broadcastToPlayer(player);
++                flag |= player.getCamera() == this.entity; // Paper - Make entities always visible for spectators, even when teleported far away (MC-107113)
```
Lulu13022002, line 18: This looks to be a bug in the override of the broadcastToPlayer call above for the players. This issue only affect player so i assume this would be a better place.

patches/server/0977-Fix-multiple-spectator-bugs.patch
FriwiDev: This is also why we check before if it is a cross world spectating case, and if so we teleport the player after his spectator target and set the camera. It is working and tested to be working.

FriwiDev: You obviously can not start spectating while your target is in another world. But I see no issue with this to be honest and consider it more like a thing that should be implemented in the /spectate command if you wish to have this functionality. Then also all your events will continue working.

patches/server/0977-Fix-multiple-spectator-bugs.patch
FriwiDev: This could theoretically be adapted to only fire when walking over a chunk border, but can leave visual asymmetry of the world if the player has a relatively small render distance set.

patches/server/0977-Fix-multiple-spectator-bugs.patch
Lulu13022002: Hum, i think you shouldn't state that this fix MC-261799 since it does not and you can't fix this only for the spectate command since plugin could call this also.

Just to be clear your fix here is just when the entity that the player spectate changes of world ? and not the reverse way (MC-261799) ?

FriwiDev: It appears to be so. I guess it would be possible to fix the spectate issue though by teleporting after the spectate event has been accepted.

FriwiDev: This appears to already be fixed in craftbukkit (with the /spectate command). When setCamera is called the player is already teleported to his target, even across worlds. I still see the not being teleported while spectating as an issue that is closely related or part of this issue.

patches/server/0977-Fix-multiple-spectator-bugs.patch
FriwiDev: I now patched this by checking for chunk position inequality before teleporting the player with a packet.

patches/server/0977-Fix-multiple-spectator-bugs.patch
Lulu13022002: It's not related, the issue (MC-261799) is not about the player being teleported or not, it is always teleported using the spectate command but the camera isn't set. So either you fix this issue or you don't claim to fix it

Lulu13022002: I recommend you to take a look of the video Avoma did that explain the issue on the ticket. Also i would also say that your fix here works only for players (and not mobs since they are recreated) and i think might be broken once player instance are recreated in the future.

FriwiDev: Possibly, but how am I supposed to resolve the new instance in another world reliably? And more important: How would I detect the world change in the first place without jeopardizing performance too much?

FriwiDev: I mean scanning every world for an entity UUID is not really efficient...

patches/server/0977-Fix-multiple-spectator-bugs.patch
```diff
@@ -0,0 +1,64 @@
...
+                 if (!player.getBukkitEntity().canSee(this.entity.getBukkitEntity())) {
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 9d46536f80b5b3e6641fd377c02166a431edfd77..9943d0291ad05572f8fb0fb15d09eac18e942973 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -11,6 +11,7 @@ import com.mojang.serialization.Dynamic;
+ import java.net.InetSocketAddress;
+ import java.net.SocketAddress;
+ import java.util.Collection;
++import java.util.Collections;
```
Owen1212055, line 30: Please use fully qualified names to prevent this added import

PR #9359, opened 2023-06-17
Allow paper-server to incrementally compile

Floppy012:
With this PR and https://github.com/PaperMC/paperweight/pull/200 paper-server will make use of Gradle's incremental compilation feature.

## Why is this needed?

Currently, every time you want to change some code and start the `runDev` task, Gradle completely recompiles the server project taking a relatively long duration. This is especially annoying if you want to use the hot reload feature of an IDE while debugging paper (i.e. changing some lines of code and hot reloading the changes into the running VM). Waiting ~30 seconds for a single line of code to update is frustrating. And if you're unlucky the VM crashes because something went wrong while trying to reload 10k classes (because it just recompiled every file in the server project) and you have to restart the `runDev` process which triggers another recompile.

## Why is that?

Gradle disables incremental recompilation for several reasons. In this case, it is because the Log4j annotation processor is not marked as incremental:

```
Full recompilation is required because org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor is not incremental.
```

## What does this PR change?

This PR removes Log4j's annotation processor from the default dependencies and only "activates" it when a jar file is built. This allows the `compileJava` task to use incremental compilation (i.e. only recompiling the classes that were changed). This reduces the compile time for consecutive runs down to 2-3 seconds on my machine.

To still load Paper's Log4j plugins, the `io.papermc.paper` package gets added to the `packages` property of the log4j configuration when starting the server using the `runDev` task.

## Paper and Incremental compilation

Since paper patches a lot of classes from external libraries (and Minecraft) but still uses the original jar files as dependencies, incremental compilation would end up in an error (to my understanding the log4j annotation processor is the reason why these errors have not been seen yet). To fix that we need to modify the classpath and add the path of the already compiled classes as the first entry so the compiler loads our patched classes instead of the unpatched classes from the original dependencies. I've opened a PR for paperweight that does that: https://github.com/PaperMC/paperweight/pull/200 (more in-depth info: https://github.com/PaperMC/paperweight/issues/199). **Paperweight needs to be updated before this PR gets merged.**

(The part from paperweight could also be placed directly in paper-server's build config, let me know what you prefer)

## Things to test

1. [x] 1st Build runs cleanly
    - Normal builds (without incremental compilation and caches) should run as expected
2. [x] Subsequent builds without code changes should not perform any compilation
    - When not changing code all the tasks should complete with "UP-TO-DATE" / "FROM-CACHE"
3. [x] Subsequent builds with code changes should trigger incremental compilation on the `compileJava` task
    - Use the `--info` CLI flag to confirm.
4. [x] All the tasks creating server JAR-Files should include the Log4j annotation processor
    - Use the `--info` CLI flag to determine whether a full recompilation was triggered because of the annotation processor
5. [x] The generated server JAR should contain the `Log4jPlugins.dat` file
    - Test if `META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat` is present
6. [x] When running the server using `runDev` log4j should scan `io.papermc.paper` for plugins
    - Set `log4j2.debug` system property to `true` and check if log4j2 scans the package. It should find 3 plugins
       -  Search for `load 3 plugins from package io.papermc.paper`
    - The server log should look "normal" not contain any errors related to log4j (see below for examples)
        - `Uncrecognized format specifier ...`
        - `Unrecognized conversion specifier ...`
        - `Rewrite contains an invalid element or attribute ...`
 7. [x] When running a built jar file, log4j should not scan the `io.papermc.paper `  package for plugins
    - Set `log4j2.debug` system property to `true`. There should be no lines starting with `Checking to see if class io.papermc.paper`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
     java
     `maven-publish`
     id("com.github.johnrengelman.shadow") version "8.1.1" apply false
-    id("io.papermc.paperweight.core") version "1.5.5"
+    id("io.papermc.paperweight.core") version "1.5.6-SNAPSHOT"
 }
 
 allprojects {
--- a/patches/server/0134-Use-TerminalConsoleAppender-for-console-improvements.patch
+++ b/patches/server/0134-Use-TerminalConsoleAppender-for-console-improvements.patch
@@ -25,10 +25,10 @@ Other changes:
 Co-Authored-By: Emilia Kond <emilia@rymiel.space>
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 9ab4d3e462c02e4d23b45adb46965eb75eb2178e..a651a41b33cb671e0d0bd7b2f990b2a727d6b15c 100644
+index 9ab4d3e462c02e4d23b45adb46965eb75eb2178e..05a3fb9ee57444a1aeddc47e13c9b965b4f23448 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
-@@ -8,7 +8,20 @@ plugins {
+@@ -8,7 +8,13 @@ plugins {
  
  dependencies {
      implementation(project(":paper-api"))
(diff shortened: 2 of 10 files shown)


jpenilla:
As far as incremental compilation, this is superseded by #9428 which takes a less hacky approach.
Hot reload can be looked at separately later.

PR #9360, opened 2023-06-17
Move Folia region scheduler API to Paper for easy compat

kennytv:
Docs might need slight adjustment first to tell people they should only really use the schedulers when also targetting Folia, otherwise this keeps almost all logic of the actual Folia schedulers intact (excluding RegionScheduler) to stay consistent

### DIFF
--- a/patches/server/0976-Folia-scheduler-API.patch
+++ b/patches/server/0976-Folia-scheduler-API.patch
@@ -1145,10 +1145,22 @@ index 8547e7ff2f1f5b7701fb0f3c3010c14601a5f83e..fff7ad7a45f310783ac96b44575ad3db
          this.players.remove(entityplayer);
          this.playersByName.remove(entityplayer.getScoreboardName().toLowerCase(java.util.Locale.ROOT)); // Spigot
 diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
-index b38c4cbcf0405d82c7b6e018e80a3174e460c1a4..5689694451d179180fc80b52930d481bf7b2b7fd 100644
+index b38c4cbcf0405d82c7b6e018e80a3174e460c1a4..d2e65783d80586b70950ad2a5b432aa021e77412 100644
 --- a/src/main/java/net/minecraft/world/entity/Entity.java
 +++ b/src/main/java/net/minecraft/world/entity/Entity.java
-@@ -251,6 +251,12 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
+@@ -246,11 +246,23 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource {
+     public @Nullable Throwable addedToWorldStack; // Paper - entity debug
+     public CraftEntity getBukkitEntity() {
+         if (this.bukkitEntity == null) {
+-            this.bukkitEntity = CraftEntity.getEntity(this.level.getCraftServer(), this);
++            // Paper start - Folia schedulers
++            synchronized (this) {
++                if (this.bukkitEntity == null) {
++                    return this.bukkitEntity = CraftEntity.getEntity(this.level.getCraftServer(), this);
++                }
++            }
++            // Paper end - Folia schedulers
+         }
          return this.bukkitEntity;
      }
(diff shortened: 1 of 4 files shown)


olijeffers0n:
Handy, I might work on a docs page at some point to just outline how you should make plugins interoperable with Paper/Folia.

A lot of the ways people check for paper is with a `Class.forName` call, and this will break all of those plugins. Can we have a `Server/Bukkit#isFolia` or some method along those lines if we are going to blur the line between the two by pulling this patch into Paper anyways? Would alleviate a lot of the headache when this gets merged is all.

patches/server/0976-Folia-scheduler-API.patch
```diff
@@ -0,0 +1,1269 @@
...
++import java.util.concurrent.ThreadFactory;
++import java.util.concurrent.ThreadPoolExecutor;
++import java.util.concurrent.TimeUnit;
++import java.util.concurrent.atomic.AtomicInteger;
++import java.util.function.Consumer;
++import java.util.logging.Level;
++
++public final class FoliaAsyncScheduler implements AsyncScheduler {
++
++    private static final Logger LOGGER = LogUtils.getLogger();
```
Machine-Maker: I think we prefer the getClassLogger as it doesn't have a huge prefix.

stonar96:
Can we also add `Server#isOwnedByCurrentRegion(...)`?

kennytv:
The owned region methods have been added, minus the global region tick thread check. An isFolia method would be a bit silly, since these methods are pulled exactly so that you *don't* need to have any unnecessary Folia checks. Plugins can/should still check this on their own if they need additional Folia methods, whether through the server name, by checking whether methods/classes they actually need exist, or checking for `io.papermc.paper.threadedregions.RegionizedServer`

stonar96:
> The owned region methods have been added, minus the global region tick thread check. An isFolia method would be a bit silly, since these methods are pulled exactly so that you _don't_ need to have any unnecessary Folia checks. Plugins can/should still check this on their own if they need additional Folia methods, whether through the server name, by checking whether methods/classes they actually need exist, or checking for `io.papermc.paper.threadedregions.RegionizedServer`

First of all thanks for adding the `isOwnedByCurrentRegion` methods. Regarding `isFolia` I only partially agree, since (I think) I have a valid usecase: Let's assume you have a `BukkitRunnable` in your Paper plugin that does something for each player on the main thread. Now you would have to schedule a task for each player to make it compatible with Folia (afaik). In this case I'd prefer to be able to check if `isFolia` and if `false`, still use a single `BukkitRunnable`.

imDaniX:
I'd say `isFolia` should just be a part of [PaperLib](https://github.com/PaperMC/PaperLib), not Paper itself.

stonar96:
Agree, didn't think about that.

olijeffers0n:
Yeah, I mentioned it because of cat's comment on a PaperLib PR from the other day:
> PaperLib is concerned about paper vs spigot, idk if there is any intent to expand this onto paper vs spigot vs folia

So it seemed like maybe this was the more appropriate place

Malfrador:
I don't really want to use PaperLib for a Folia check.  I don't care about Spigot support, and needing to use a lib with many spigot-related features would be unnecessary. 
There certainly are situations in which I would like to check if the server is running Folia or not, even if the Scheduler methods are in Paper. For example I might want to run different logic that only makes sense if there are multiple Folia regions. 

Checking for the server implementation name doesn't work with forks, and needing to check for the existence of RegionizedServer is just an awkward solution. So I really would like an isFolia method as well.
Folia and Paper being worked on by the same people is a different situation than Spigot/Paper, so there shouldn't be any workarounds like checking for the existence of a specific class needed.

electronicboy:
I would much rather have a feature flag system as opposed to some asinine "isX" set of methods, having methods to check what platform you're on would generally suggest a flaw in how the API is working or in how you're using it, you'd still need to handle aspects of those methods existing or not unless you'reonly catering towards the latest version of paper, etc

F3F5:
i love this pull

ghost (approved):
like

kennytv:
Feature flags do sound better, but out of scope for this pr, so here I go

PR #9361, opened 2023-06-17
Add method to remove all active potion effects

Machine-Maker:
Adds a new API method to `LivingEntity` called `removeAllActivePotionEffects` which calls the same logic the milk bucket does. The method name was chosen because of the existing method called `getActivePotionEffects`.


jpenilla:
I'd prefer clear over removeAll for the method name

Machine-Maker:
Done, changed to use `clear` over `removeAll`.

electronicboy (approved):

PR #9364, opened 2023-06-18
Add Owner UUID api for AreaEffectCloud

Machine-Maker:
Can maybe be merged into Owen's big "missing entity API" patch?

I followed the name convention for other "get owner uuid" methods with `getOwnerUniqueId`, but things are less clear for "set owner uuid" method. There is "setOwner(UUID)" in some places, but it feels weird to not use the same pattern for a get/set pair.

### DIFF
--- a/patches/api/0299-Missing-Entity-API.patch
+++ b/patches/api/0299-Missing-Entity-API.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
 Date: Fri, 28 May 2021 21:06:59 -0400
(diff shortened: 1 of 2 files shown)


Owen1212055:
Yes. I also think that patch should be renamed to missing entity api rather than missing entity behavior api.

Sytm:
> but it feels weird to not use the same pattern for a get/set pair.

If the names are identical of the getter and setter, it would have the added benefit that they could be accessed from [Kotlin with synthetic properties](https://kotlinlang.org/docs/java-interop.html#getters-and-setters)

Machine-Maker:
Ok, keeping those names so the get/set have the same name. And I merged it into the `Missing Entity API` patch (renamed from `Missing Entity Behavior API`).

Owen1212055 (approved):

PR #9365, opened 2023-06-18
Improved AreaEffectCloud API

Machine-Maker:
Open to different naming for everything here, I couldn't think of a straightforward easy name to any methods/fields that wasn't like a full sentence long.

----
This new API allows more control over how AreaEffectClouds are used. An "empty" cloud can be forced to still calculate affected entities and the AreaEffectCloudApplyEvent has finer control over what counts as a "use" for decreasing the "radius" of the cloud.


Machine-Maker:
Am just re-looking at this... I wonder if its best for a whole separate event to check if an area effect cloud should apply if it's got an empty effects list instead of a field on the cloud that you mark? A whole event seems like a lot for this as there really isn't much else that's configurable.

PR #9366, opened 2023-06-18
Don't tab-complete namespaced commands if `send-namespaced` is false

EpicPlayerA10:
Tab-complete packet is supposed to tab-complete args for commands, but for some reason it also can suggest commands like in version 1.12.2 or lower.

Let's say that player sends `ServerboundCommandSuggestionPacket` packet (yarn name: `RequestCommandCompletionsC2SPacket`) with `command` parameter set to `/`. Without this patch, the server would send all commands including namespaced commands when `send-namespaced` is false. Some hacked clients may use this bug to show namespaces (and therefore plugins names).

**TLDR:**
This patch prevents server from sending namespaced commands list when `send-namespaced` is false and player requests to tab-complete only commands.

---
Download the paperclip jar for this pull request: [paper-9366.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/867731183.zip)


patches/server/0979-Don-t-tab-complete-namespaced-commands-if-send-names.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+index 2a609e43370e68943c580083f7f7d8c9b0972955..6c1d164561b9b7b4d905832fb65e9b34d21b4de7 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -918,6 +918,11 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                         ParseResults<CommandSourceStack> parseresults = this.server.getCommands().getDispatcher().parse(stringreader, this.player.createCommandSourceStack());
+ 
+                         this.server.getCommands().getDispatcher().getCompletionSuggestions(parseresults).thenAccept((suggestions) -> {
++                            // Paper start - Don't tab-complete namespaced commands if send-namespaced is false
++                            if (!org.spigotmc.SpigotConfig.sendNamespaced && !command.contains(" ")) {
```
Machine-Maker: I'm not 100% sure on how the brig `Suggestions` type works, but could you instead check the string range in the `Suggestions` object to see if it begins at 0? That might be a better way of knowing if its completing the first node.

Machine-Maker (commented):
A totally separate way of going about this, and one that might be better as it provides other ways of knowing if a command is a "namespaced" duplicate, would be to create a new impl of the CommandDispatcher that can filter on the "namespaced" commands.

patches/server/0979-Don-t-tab-complete-namespaced-commands-if-send-names.patch
EpicPlayerA10: I don't know if I understood correctly. What do you mean by checking the string range? What string range?

EpicPlayerA10:
> A totally separate way of going about this, and one that might be better as it provides other ways of knowing if a command is a "namespaced" duplicate, would be to create a new impl of the CommandDispatcher that can filter on the "namespaced" commands.

I don't want to complicate stuff. There would be two dispatchers with similar commands. I want to stick with much simpler and still efficient solution.

patches/server/0979-Don-t-tab-complete-namespaced-commands-if-send-names.patch
Machine-Maker: The Suggestions object in that completable future contains a range that the suggestions is for.

EpicPlayerA10: Done

Machine-Maker (approved):

Warriorrrr (approved):

NoahvdAa (approved):

PR #9367, opened 2023-06-18
Add setting to disable noteblock updates

Boy0000:
Adds an "unsupported" setting to prevent noteblocks from updating.
Mostly to avoid the need for listening to BlockPhysicsEvent in plugins that try and add custom blocks via NoteBlocks
This should remove the need for that.

Just a nice feature I saw Purpur had, and figured I would open a PR.
Unsure if this is even wanted and if it should go in unsupportedSettings

https://github.com/PaperMC/Paper/assets/62521371/413210c4-6bdb-4f9d-a668-a581cdcdf1fa


electronicboy:
Configuration changes should go into the configuration patch

kennytv:
I think that's also fine to be outside the unsupported section

PR #9368, opened 2023-06-18
Add option to disable NoteBlock and Tripwire updates

Boy0000:
Superseeds #9367
Closing that as it was made from an Organization and not a Personal fork.
This also includes tripwires

Moved config stuff to Config patch, and out of "unsupported" and into "blockUpdates"
Figured a section for it was best as there are multiple options now and perhaps more down the line

In addition to preventing noteblock state updates this also prevents tripwires.
Noteblock example: (Config is outdated)

https://github.com/PaperMC/Paper/assets/62521371/85faef6d-4ee9-4907-b739-a4c1f8f0ca87

tripwire example:

https://github.com/PaperMC/Paper/assets/62521371/10977f17-08f2-4125-9d72-5c73a93ff615

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -440,10 +440,10 @@ index 0000000000000000000000000000000000000000..9ef6712c70fcd8912a79f3f61e351aac
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..8633fdfa7e8aeac7414c232006fc298b3bfe1b58
+index 0000000000000000000000000000000000000000..58cdb70c0b9de45ca28811416d871099eefec2ce
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,276 @@
+@@ -0,0 +1,283 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
@@ -718,6 +718,13 @@ index 0000000000000000000000000000000000000000..8633fdfa7e8aeac7414c232006fc298b
 +        public boolean lagCompensateBlockBreaking = true;
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
++    }
++
++    public BlockUpdates blockUpdates;
++
++    public class BlockUpdates extends ConfigurationPart {
++        public boolean disableNoteblockUpdates = false;
++        public boolean disableTripwireUpdates = false;
 +    }
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/InnerClassFieldDiscoverer.java b/src/main/java/io/papermc/paper/configuration/InnerClassFieldDiscoverer.java
--- a/patches/server/0019-Rewrite-chunk-system.patch
+++ b/patches/server/0019-Rewrite-chunk-system.patch
@@ -15673,7 +15673,7 @@ index 0000000000000000000000000000000000000000..962d3cae6340fc11607b59355e291629
 +
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-index 8633fdfa7e8aeac7414c232006fc298b3bfe1b58..eb12227b009b1fb766a5f9e338e5d2394d498376 100644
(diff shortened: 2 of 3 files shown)


kennytv (approved):

Machine-Maker:
I'm curious why include tripwires? Does it have anything to do with custom blocks? I'm not too familiar with how custom blocks are implemented, but shouldn't mushroom blocks be included in this? Or have noteblocks taken over as the "go to"?

Boy0000:
Tripwires are used for plants and crops, as shown in the second clip.
Kinda flashes to examples used for custom plants in our pack
image

patches/server/0005-Paper-config-files.patch
```diff
@@ -719,6 +719,13 @@ index 0000000000000000000000000000000000000000..8633fdfa7e8aeac7414c232006fc298b
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
 +    }
++
++    public BlockUpdates blockUpdates;
++
++    public class BlockUpdates extends ConfigurationPart {
++        public boolean disableNoteblockUpdates = false;
++        public boolean disableTripwireUpdates = false;
++    }
```
Machine-Maker, line 728: ~~Since they are unsupported settings, This section should be in the unsupported section.~~

Should this go in the WorldConfiguration? Is there any benefit you see as a user of custom blocks to having it per world?

Machine-Maker (changes requested):

patches/server/0005-Paper-config-files.patch
Boy0000: I thought about making it per world, but I don't see any real benefit as the plugins that would implement this most likely doesnt distinguish between worlds for the feature.

The resourcepack would have to be changed between each world then. Since the blockstate files dictate the "model" shown, a world where this would be disabled would have very funky looking tripwires/noteblocks
Might be benefits to making it per-world though that im not seeing

Machine-Maker: Ignore the bit about unsupported. They are fine in their own section. 

> Might be benefits to making it per-world though that im not seeing

can be moved there if needed in the future then.

Machine-Maker (approved):

Malfrador:
Would it be reasonably possible to have this as a list of blocks to disable instead?

I currently disable a bunch of chorus blocks as well as mushroom block states, and there are quite a[ few other blocks](https://theepicblock.github.io/PolyMc/blockstates/) that can be used for this purpose as well that each have their own advantages and disadvantages. 
Having a list that you can just add blocks to would also make this more future proof imo.

Boy0000:
The main reason Noteblock and Tripwires are used over all the ones that link mentions is WorldGen.
Sucar Canes, Mushroom Stems etc, generate throughout the world, whilst Noteblocks and tripwires dont.
The PR here focuses therefore on the 2 most commonly used ones across all plugins and servers, and the ones with least amount of issues.

Also @Machine-Maker not sure but do I need to rebase this again since path.800 has already been used?

EpicPlayerA10:
@Boy0000 Could you also add an ability to disable updates for chorus plant? Chorus plants are widely used as transparent blocks in plugins that are using a resourcepack to create custom blocks and items.

electronicboy:
We can rebase on merge if needs be

electronicboy:
I don't think that there is a sane way to implement such a feature as a list of blocks given that you need to actually insert checks into various places, to actually have a list would probably require the logic to prevent updates to be shoved in a place and effectivrly prevent all block state mutations for the same block type, which would probably have some hefty side effects, and probably cause some wider issues with other logic in the server which we couldn't really support

Machine-Maker:
Yeah, I agree with cat. I don't think there's a good way to make it fully dynamic like this without adding code into incredibly hot paths already. We can handle the main few with specific settings, and any others will have to be handled manually via the physics event in plugins.

With that being said, is there any other very commonly used block used for general purpose custom blocks other than note blocks and tripwire?

Boy0000:
These are the most common once. The rest kinda require altering world gen. Chorus plant is probably the main one that a few plugins use because noteblock doesnt support transparency. Some plugins use leaves for this aswell which have alot morw states but requires a lot of additional behaviour

EpicPlayerA10:
> Some plugins use leaves for this aswell which have alot morw states but requires a lot of additional behaviour

Yeah, I know but that would be awesome if you also add this option to disable block updates for chorus plants 🥹

kennytv:
Anything past these two is out of scope for the original intent and required effort in terms of this pr

welove69:
I know this shouldn't be discussed here
Please fixInfinite String Farm，I also don't want the server to have any vulnerabilities to copy items

https://github.com/PurpurMC/Purpur/discussions/1368

electronicboy:
Please don’t post on unrelated issues, we know it’s a bug, it’s not trivial to solve and I do not have working hardware to be able to look at it much.

PR #9377, opened 2023-06-19
Only tick item frames if players can see it

Redned235:
In the event that an item frame cannot be seen by any players, ticking the item frame every tick is unnecessary. This can be a very hot section of the entity tracker when lots of item frames are present on a server, so this reduces the logic which speeds it up.

---
Download the paperclip jar for this pull request: [paper-9377.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/771539235.zip)


kennytv (approved):

Owen1212055 (approved):

frqnny:
Quick questions. What does the build-pr-jar tag mean, and what's holding this PR from being merged? I'm willing to see how this PR affects my server's performance since it would greatly help us, so if a performance comparison helps I'd be willing to run spark profiles of the server to compare. I tracked down this exact method to be the culprit that really lags us the most. 
Were all item frames ticking regardless of whether the player had them inside their simulation distance?

electronicboy:
What's holding it is that we want people to test it before we pull it, hence the build pr jar label, which allows you to download a copy from the link added to the first post

frqnny:
This is in our dev server so it's a little less pronounced but we still have thousands of maps on item frames.

before: https://spark.lucko.me/kwGBCvRSHb
ItemWorldMap.getMapId is 13% of our tick.

after: https://spark.lucko.me/WSCzCNrT3T
ItemWorldMap.getMapId is now only 5% of our tick.
Even though this spark has a higher MSPT by 1, this method greatly reduced our lag coming from maps on item frames. Considering that I only ran spark for less than a minute and flew around, I think this PR is doing what it's supposed to: optimize the server pretty well. 

for note I have to fly around a bit for a few minutes to let the image maps generate first since it's pretty laggy when they do and then after that we get continous lag from this method visable in the sparks profile. 

I think that in our Production server it'll REALLY cause MSPT to fall because 1. it's a way better machine 2. thousands of more item frames that can  go by without getting changes sent. 


another note, this was with item-frame-cursor-update-interval of 35. That's pretty high, and our Production Server has them at 3. For note, /paper entity list tells me this was tested with 6k item frames ticking. Our production server though can hit 12k item frames ticking. Please merge soon.

electronicboy:
I don't have the means to test this, but, but it generally looked fine, I'll fight my machine tomorrow and try to merge this if there's no objections

electronicboy:
tried rebasing by my setup is being weird and refusing to do so

sss-ryun:
I suggest checking if the entity is contained in the _trackedPlayers_ instead, assuming _trackedPlayers_ has the entities seen by the players. This change is good. I'm new to contributing to Paper so I will check later if my idea checks out

PR #9379, opened 2023-06-20
Fix missing item types in SlotType for armor change event

Machine-Maker:
Also added 2 parameterized tests to make sure this doesn't happen again.


Owen1212055 (approved):
Kinda strange api, may be worth making it use item tags or something in the future. But, lgtm.

Issue #9303, opened 2023-06-11
Add sign side to Player#sendSignChange


KabanFriends:
### Is your feature request related to a problem?

Player#sendSignChange is currently missing sign side

### Describe the solution you'd like.

Add sign side parameter to Player#sendSignChange

### Describe alternatives you've considered.

Directly send packets via NMS 💀

Owen1212055:
Whoever wants to take a stab at this, feel free. 😄

Infynyty:
Hey, I'd like to contribute, but I'd be my first time here, so please bear with me :)
Would it be better to add new methods that include the sign side as an argument in the `Player` interface to keep backwards compatibility or should the current methods be modified? In that case should deprecated methods also be updated?

Owen1212055:
> Hey, I'd like to contribute, but I'd be my first time here, so please bear with me :)
> Would it be better to add new methods that include the sign side as an argument in the `Player` interface to keep backwards compatibility or should the current methods be modified? In that case should deprecated methods also be updated?

You would want to add new methods, and most likely deprecate the old ones.

yannicklamprecht:
It doesn't just use the sendBlockUpdate instead.

yannicklamprecht:
> er, I don't think that this could ever work as what is hoped here? the entire sign TE is sent, unless I'm missing something here
> 
> all I could see is basically adding a 2nd param for the backsides components

The TE is sent everytime. The other sendSignChange methods do instantiate a TE in impl.

yannicklamprecht:
> Yea, but you can't just send _a_ side is what I mean, it sends the TE update packet; not really sure adding a Side param makes too much sense as it won't work as expected if you're expecting it to just update a side as seen by the client

Ah missunderstood you. Yes, the opposite site will be send empty, right.

Infynyty:
> The long term goal should be to deprecate the sendSignChange method in favour of the sendBlockUpdate which can handle multiple blockentity types.

That seems like a cleaner solution. Adding new methods now that waxing and sign sides have been added would create a lot of different parameters. Passing a `Sign` as an argument would be shorter, but then it is really just `sendBlockUpdate` with extra steps. 
So, should the old `sendSignChange` methods be deprecated and refer to  `sendBlockUpdate` instead or is there still a case to update the methods?

Machine-Maker:
[...]

PR #9382, opened 2023-06-22
Mark Player.sendSignChange as deprecated

Infynyty:
As discussed in #9303 this pull request marks the Player.sendSignChange methods as deprecated in favor of Player.sendBlockUpdate.

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -2111,7 +2111,7 @@ index f3afe67f0832cb828d25be3654518ff73a80b0e1..598abaa82c634178043a29f6caa6ac52
 +    // Paper end
  }
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 7591d6de16694467b0fa5ab7ee746f3efdafad82..72b014ebf1a88bf0bf583f570b339ecc1d7a8ef1 100644
+index 7591d6de16694467b0fa5ab7ee746f3efdafad82..513e48de04a78268d16c924717d64896a23b437f 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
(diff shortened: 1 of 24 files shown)


Machine-Maker (changes requested):
The methods should be deprecated in adventure, and the deprecation messages on both the String and Component methods should all point to sendBlockUpdate. Also, I think it should expand more on how to exactly replicate behavior, like say to create sign tile states via `Material.OAK_SIGN.createBlockData().createTileState()` or whatever it is exactly.

KabanFriends:
Is there even an API to create arbitary TileStates?

KabanFriends:
> The methods should be deprecated in adventure, and the deprecation messages on both the String and Component methods should all point to sendBlockUpdate. Also, I think it should expand more on how to exactly replicate behavior, like say to create sign tile states via `Material.OAK_SIGN.createBlockData().createTileState()` or whatever it is exactly.

Is there even an API to create arbitrary TileStates?
Also, names mentioned in the docs seems to be wrong. It should be `sendBlockChange` instead

Machine-Maker:
> Is there even an API to create arbitrary TileStates?

Yes, recently added by upstream, `BlockData#createBlockState`.

yannicklamprecht:
> > The methods should be deprecated in adventure, and the deprecation messages on both the String and Component methods should all point to sendBlockUpdate. Also, I think it should expand more on how to exactly replicate behavior, like say to create sign tile states via `Material.OAK_SIGN.createBlockData().createTileState()` or whatever it is exactly.
> 
> Is there even an API to create arbitrary TileStates?
> Also, names mentioned in the docs seems to be wrong. It should be `sendBlockChange` instead of `sendBlockUpdate`.


Have fun with the examples.
https://gist.github.com/yannicklamprecht/f42caeade70ffec33ee1e0c15bcabfef

Owen1212055 (approved):

yannicklamprecht (approved):

Machine-Maker (approved):

PR #9388, opened 2023-06-23
Add Sign#getInteractableSideFor

### LINKED ISSUE #9423: Player Interact Sign Event for Dyeing and Glowing

Using the existing logic of PlayerInteractEvent makes it incredibly complicated to determine which side of a sign the player interacted with, when using dyes or ink sacs. 

Issue #9293 / PR https://github.com/PaperMC/Paper/pull/9300 addressed a similar issue for the purposes of editing signs, but did not address this use case, and the original issue reporter even noted this use case in the PR before merging: https://github.com/PaperMC/Paper/pull/9300#issuecomment-1586080271

A event that triggers when a player tries to dye a sign, or use an ink sac on it, with a way to get which side of the sign is being clicked.

As before with #9293, the only alternative I see is via NMS.

### Other

_No response_

Machine-Maker:
Adds API to get the specific Side of a sign that an entity, position, or coord pair is facing. Not 100% on the method name, so open to other options.

### DIFF
--- a/patches/api/0420-Add-Sign-getInteractableSideFor.patch
+++ b/patches/api/0420-Add-Sign-getInteractableSideFor.patch
@@ -1,0 +1,46 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Fri, 23 Jun 2023 12:16:35 -0700
+Subject: [PATCH] Add Sign#getInteractableSideFor
+
+
+diff --git a/src/main/java/org/bukkit/block/Sign.java b/src/main/java/org/bukkit/block/Sign.java
+index e6b87ad5b2099800a61e7f585a5404e9e672228a..2f90bbb1542580da5a45be881fe2143e9da96eca 100644
+--- a/src/main/java/org/bukkit/block/Sign.java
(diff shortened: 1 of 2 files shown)


lynxplay (approved):

Machine-Maker:
Renamed method to `computeVisibleSide`.

molor:
(why does adding all the new features that users most want (look at reactions) always take so long?)

electronicboy:
Welcome to the joys of a hobby project, I currently do not have hardware/eyes to work on stuff right now either

kennytv:
Why "compute"? Prefixing every method that computes something with compute would be silly and isn't too relevant to the user here (as opposed to something like getComputedBiome or computeIfAbsent). If it's to let users know they should put it into a local var instead of re-calling it multiple times, you should note that to the javadocs instead

electronicboy:
It was the nicer name that we could come up out of the alternatives to getting this information

kennytv (approved):
Whatever name it may be, pre-approving anyways

me4502:
Suggested in Discord, but how does `findSide` feel as a name? IMO it's less verbose and still gets the point across. It also still implies it's not a simple getter, but doesn't use the word compute

Machine-Maker:
Hopefully the final rename to `getInteractableSideFor`. Its longer, but more accurate.

PR #9391, opened 2023-06-24
Fix possible NPE on painting creation

### LINKED ISSUE #9374: Can't spawn a painting via world.spawnEntity()
### Expected behavior

Should be able to spawn a painting, like in 1.19.4

### Observed/Actual behavior

[08:42:05 WARN]: java.lang.IllegalArgumentException: Cannot spawn an entity for org.bukkit.entity.Painting
[08:42:05 WARN]:        at org.bukkit.craftbukkit.v1_20_R1.CraftRegionAccessor.createEntity(CraftRegionAccessor.java:994)
[08:42:05 WARN]:        at org.bukkit.craftbukkit.v1_20_R1.CraftRegionAccessor.spawn(CraftRegionAccessor.java:543)
[08:42:05 WARN]:        at org.bukkit.craftbukkit.v1_20_R1.CraftRegionAccessor.spawn(CraftRegionAccessor.java:539)
[08:42:05 WARN]:        at org.bukkit.craftbukkit.v1_20_R1.CraftRegionAccessor.spawn(CraftRegionAccessor.java:525)
[08:42:05 WARN]:        at org.bukkit.craftbukkit.v1_20_R1.CraftRegionAccessor.spawnEntity(CraftRegionAccessor.java:433)

### Steps/models to reproduce

Call this code  
Entity entity = location.getWorld().spawnEntity(location, EntityType.PAINTING));

### Plugin and Datapack List

Custom plugin

### Paper version

This server is running Paper version git-Paper-17 (MC: 1.20) (Implementing API version 1.20-R0.1-SNAPSHOT) (Git: c287e92)
Previous version: git-Paper-"fa8fa1c" (MC: 1.19.4)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/9374

This should go back to the original behavior of spawning the painting in, if only for a few seconds until it breaks because it can't exist at that location.

Can be tested by trying to spawn a painting at the same location as a ladder.

### DIFF
--- a/patches/server/0987-Fix-possible-NPE-on-painting-creation.patch
+++ b/patches/server/0987-Fix-possible-NPE-on-painting-creation.patch
@@ -1,0 +1,28 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 24 Jun 2023 09:42:53 -0700
+Subject: [PATCH] Fix possible NPE on painting creation
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java b/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
+index b305941aaf50b417bb054a9327c67a0891f2fe9e..995b3e31c5725f3d8879df349351eae3bcdfdddf 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Machine-Maker:
So there's probably a better way to fix this, like updating cb's checks to use the correct logic to check if a painting is valid, and then fallback to SOUTH.

Machine-Maker:
Added a missing collision check for ItemFrames and Paintings when being created. Should hopefully help with trying to guess the correct direction. I think the other bit of diff handling `null` from the `Painting.create()` method is still useful, as I'm not 100% sure we covered all bases when checking for collisions. Like we don't have the entity context.

Owen1212055 (approved):

PR #5460, opened 2021-04-05
Implement mobspawning back off mechanic

kickash32:
Closes #5076

Tested with a plugin which cancelled all mobspawns spawns using PreCreatureSpawnEvent, using 1 player.
Observed a 75% reduction in execution time of mobspawns (from 54% of tick to 4%).

TODO
- Add missing Paper comments
- Configurable bleed
- Change bleed algorithm?
- Improve CreatureSpawnEvent cancelling as well

PR #9400, opened 2023-06-26
Improve cancelling PreCreatureSpawnEvent with per player mob spawns

### LINKED ISSUE #5076: mob spawn cancellation back-off
[Not feeling well, but, wanted to make this issue while I remembered, please bear with me]

A common issue with mob spawning in modern versions of the server is how expensive and heavy-handed it is, plugins that clear mobs in the world often induce these types of spikes

In vanilla, the seeming reliance on mob spawning is that when you log in, you'll quickly (within a hanful of ticks to my recollection of the logic/experience) hit the spawn limit, it seems that the intent is that you'll stick to the mob limits pretty closely and deal with mobs slowly spawning back in due to the nature of things like the hard/soft despawn ranges removing mobs slow enough for the server to spawn the back in a small number each tick as they die off, this is fine;

However, there comes an issue due to this with when mobs spawns are canceled, e.g. such as with worldguard, in that the server will consistently try to meet the spawn targets, this is especially an issue with aspects like per player mob spawning, in which it's pretty easy for a plugin preventing spawns in a region for a handful of players to spend a good few ms trying to spawn mobs (I've personally seen this off the top of my head peek to 7-10ms with just bats prevented on some servers)

While, solving this on a global level is generally unviable, imho, there is no sane way to guarantee mob spawns without some form of granularity to ensure that spawning stays sane (the way the vanilla logic works has me worried that a global/chunk based attempt would introduce more caveats than worth investigating), it would potentially be ideal for per-player mob spawning to have a per-player back-off, easiest way I can see to do this is to artificially increase the number of mobs it sees for a player, if the server has had 3 mobs spawns, act as if we have 3 extra mobs in the chunk when checking the limits to allow us to back off here, and slowly bleed those off

This is just a theory as to what can be done to improve the performance of servers using things like protection plugins, there may be other solutions or the idea here may be so nuanced to implement, this does not solve a few vanilla issues with how spawning e.g. cases where there are few viable places for mobs to spawn can also induce similar performance penalties

### LINKED ISSUE #9393: TPS drops down to 15 with default paper spawning settings, high amount of events
### Timings or Profile link

https://timings.aikar.co/dev/?id=cf3a557c91e04708ae62c2a72095144d

### Description of issue

I got extreme lags with the default mob spawning settings, i lowered the monsters and it got better.

I thought this is a myworlds issue, but the plugin developer suggested, that i should report this here because "myworlds just uses papers builtin entity pre-spawn event so theres nothing I can do about it".

### Plugin and Datapack List

MyWorlds, BKCommonLib

### Server config files

default generated

### Paper version

This server is running Paper version git-Paper-46 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 206dd3a)
You are 2 version(s) behind
Previous version: git-Paper-45 (MC: 1.20.1)

### Other

_No response_

Machine-Maker:
Re-implementation of https://github.com/PaperMC/Paper/pull/5460

Fixes https://github.com/PaperMC/Paper/issues/9393

---- 
Tracks an additional count of mobs per category for each category. The number is incremented on an aborted or cancelled PreCreatureSpawnEvent. That number is added to the true mob count per player when the game requests the mob count for a player. This "backoff number" is then decremented by 1 each tick for each player for each mob category.

A rudimentary test I did involved a plain server with 1 player not moving for 60 seconds. With the PreCreatureSpawnEvent set to be cancelled and aborted WITHOUT these changes, it fired ~48000 times during that time period. With the changes, it fired ~5000 times over the same time period. I included the test plugin I used to do this in the PR. Will be removed before merging.

The algorithm used to decrease the "backoff number" is linear and simple now, unclear if it should be customizable further, or default to a non-linear implementation.

---
Download the paperclip jar for this pull request: [paper-9400.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/873225710.zip)


dandud100:
Thank you, for this fast fix. I hope this gets merged soon.

patches/server/0172-PreCreatureSpawnEvent.patch
```diff
@@ -165,12 +173,26 @@ index b2bb9bbd3af414c50ec3f8e3e171a679e95ef75e..5338e0e1a67925da0c386735a545bb31
...
+-            return false;
++            return PreSpawnStatus.FAIL; // Paper
+         } else if (entitytypes.canSummon() && NaturalSpawner.canSpawnMobAt(world, structureAccessor, chunkGenerator, group, spawnEntry, pos)) {
+             SpawnPlacements.Type entitypositiontypes_surface = SpawnPlacements.getPlacementType(entitytypes);
+ 
+-            return !NaturalSpawner.isSpawnPositionOk(entitypositiontypes_surface, world, pos, entitytypes) ? false : (!SpawnPlacements.checkSpawnRules(entitytypes, world, MobSpawnType.NATURAL, pos, world.random) ? false : world.noCollision(entitytypes.getAABB((double) pos.getX() + 0.5D, (double) pos.getY(), (double) pos.getZ() + 0.5D)));
++            return !NaturalSpawner.isSpawnPositionOk(entitypositiontypes_surface, world, pos, entitytypes) ? PreSpawnStatus.FAIL : (!SpawnPlacements.checkSpawnRules(entitytypes, world, MobSpawnType.NATURAL, pos, world.random) ? PreSpawnStatus.FAIL : world.noCollision(entitytypes.getAABB((double) pos.getX() + 0.5D, (double) pos.getY(), (double) pos.getZ() + 0.5D)) ? PreSpawnStatus.SUCCESS : PreSpawnStatus.FAIL); // Paper
```
kennytv: Should just put the boolean into an extra local var first, then return the spawn status, to keep this easier to read (especially on updates)

LucidAPs:
Will this be merged?

patches/server/0172-PreCreatureSpawnEvent.patch
Machine-Maker: Good idea, done.

dandud100:
Any updates on this? Isn't it ready to merge already?

LucidAPs:
Please do not let this die in pull request queue.

kennytv (approved):

dandud100:
Fixes also #5076

dandud100:
Can you provide a snapshot build with this patch and without the test plugin to run a smoother server until this gets merged? @Machine-Maker

dandud100:
That was fast! Will the action to build the paperclip.jar get started automatically?

Machine-Maker:
> That was fast! Will the action to build the paperclip.jar get started automatically?
Huh, it should've automatically started when I added the label.

dandud100:
>> Huh, it should've automatically started when I added the label.

Is there a way to trigger it manually?

PR #9406, opened 2023-06-27
Add entity tracker check Y

ruViolence:
Added the option to enable Y coordinate checking in the entity tracker, which reduces the server load and the amount of packet sending.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1588,6 +1588,7 @@ index 0000000000000000000000000000000000000000..d47c57afafc01e25b965f1844938b251
 +
 +    public class Entities extends ConfigurationPart {
 +        public boolean entitiesTargetWithFollowRange = false;
++        public boolean trackerCheckY = false;
 +        public MobEffects mobEffects;
 +
 +        public class MobEffects extends ConfigurationPart {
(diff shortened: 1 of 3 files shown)


MachineBreaker:
Can't the configuration name be a little more... explainable?

ruViolence:
> Can't the configuration name be a little more... explainable?

My English isn't good enough, that's why it's such a dumb name. I can change it to something else if there are any suggestions, well, or the contributors can change it.

> I've never really been comfortable with tying this to the same range limits as the others, like, you look down into an area where there's 0 mobs, then you drop down and they all come into view, whereas without the care for Y, they'd be there; ofc, there comes some expectations for people to not go crazy, but, I've heard of servers pushing entity tracking down stupidly low that this could be an actual concern

Well, this feature is disabled by default. Cheaters can walk around the world and find deep underground minecart's, players and stuff like that, so someone might need this feature. After all, the height of the world is quite large in the new versions, so why would the player see everything that is deep underground where he often does not have the opportunity to fall down there.

Machine-Maker:
Perhaps its worth somehow making the Y distance configurable independent of the other tracking distances?

ruViolence:
Okay, made it more configurable. Also did a rebase somehow, I'm still new to git 😝

MachineBreaker (approved):

TheRealDardan:
A comment, with the configuration in the last commit this can also be used to help balance servers with custom long range combat by preventing people on pillars from raining hell down from y 300, so the benefits arent limited to performance/packet things. So , great change.

kennytv (approved):
d0 would benefit of an obfuscation helper, otherwise good

PR #9408, opened 2023-06-27
Do SavedData IO Saving Async

Cryptite:
This work stems from a discovery of some fairly heavy main-thread IO writing going on with SavedData coming, at least in the below screenshot, from DimensionDataStorage saving: 
image

We should be able to safely perform async IO writes of this data when doing typical incremental saves. There was some debate about which executor to use, so open to discussion on that and more.

---
Download the paperclip jar for this pull request: [paper-9408.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1178586640.zip)


electronicboy:
I'm still in a general conflict with this sorta thing, the lack of means to ensure that we get behind all data saves is generally a nuisance, while 99.999999999999...% of the time this is not going to cause issues, there is still a small technical risk of mangling the data on the disk

electronicboy:
https://github.com/electronicboy/Paper/commit/4e672d574c5eeacf58971c6125bc8e83945b7457

Not 100% on it, but I generally feel that that would be more closer to making me happy here

MinecraftCitiesNetwork:
> [electronicboy@4e672d5](https://github.com/electronicboy/Paper/commit/4e672d574c5eeacf58971c6125bc8e83945b7457)
> 
> Not 100% on it, but I generally feel that that would be more closer to making me happy here

We've just started using this patch to solve an issue we reported in #paper-help, world saving no longer lags for us! :)

StKillReal:
I've been trying to fix world save lag spikes for a month. This patch fixed the issue, thanks

Warriorrrr:
Closing in favour of #10171 which should be getting merged soon, thank you for your contribution!

PR #9409, opened 2023-06-27
Add canDespawn to Warden API

Cryptite:
Adds the ability to get/set whether the Warden can ever despawn (dig down after idling for 1200 ticks)


Doc94:
And working with the Digging class its not a option? rather than modify the cooldown?

Cryptite:
> And working with the Digging class its not a option? rather than modify the cooldown?

It didn't seem like the better way to go. Since the Digging class is part of its `makeBrain` method, that would require somehow defining that the warden spawning behavior is not added to the Brain Memory at creation time at all. 

With the proposed API you can optionally change the despawning behavior with a Warden at any time

kennytv:
"can despawn" is a bit misleading since it definitely can still despawn, it just doesn't despawn by digging away

patches/server/0982-Add-Warden-setCanDespawn.patch
```diff
@@ -0,0 +1,60 @@
...
+     public static void setDigCooldown(LivingEntity warden) {
+         if (warden.getBrain().hasMemoryValue(MemoryModuleType.DIG_COOLDOWN)) {
+-            warden.getBrain().setMemoryWithExpiry(MemoryModuleType.DIG_COOLDOWN, Unit.INSTANCE, 1200L);
++            // Paper start - Wardens despawn based off an expiring DIG_COOLDOWN memory. So, the
++            // only way to sanely prevent them despawning is to reset their cooldown memory to
++            // a non-expiring one.
++            if (warden instanceof Warden nmsWarden && !nmsWarden.canDespawn) {
++                warden.getBrain().setMemory(MemoryModuleType.DIG_COOLDOWN, Unit.INSTANCE);
++            } else {
++                warden.getBrain().setMemoryWithExpiry(MemoryModuleType.DIG_COOLDOWN, Unit.INSTANCE, 1200L);
```
kennytv, line 34: Remove the indentation here

patches/server/0982-Add-Warden-setCanDespawn.patch
```diff
@@ -0,0 +1,60 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftWarden.java
+@@ -92,4 +92,16 @@ public class CraftWarden extends CraftMonster implements org.bukkit.entity.Warde
+             case ANGRY -> AngerLevel.ANGRY;
+         };
+     }
++
++    // Paper start
++    @Override
++    public void setCanDespawn(boolean canDespawn) {
++        this.getHandle().canDespawn = canDespawn;
```
kennytv, line 52: Maybe this should also remove the dig cooldown?

PR #9413, opened 2023-06-28
Dont resend blocks on interactions

Owen1212055:
In general, the client now has an acknowledgment system that will cause block changes made by the client to be reverted correctly.

Essentially: 
1. The client enters a "prediction" stage, where any block changes made will have its old blockstate captured (this is referred to as "server state"). 
2. If you update blocks during this stage, the client will update this captured server state as long as they're still currently predicting. 
3. After prediction is done (via an ack packet) all captured blockstates are reverted to their captured server state. 

This means that if the server *actually updated* a block and send a block update packet, it's correctly set, while if a block wasn't updated on the server but WAS updated on the client (server state wasn't updated), that change will be reverted.

It should be noted that this system does not yet support block entities, so those still need to be resynced when needed.

I discovered this when noticing that blocks broken outside of the player's valid interaction distance are still properly reverted, even though the server doesn't send any block updates, only an ack packet.

---
Download the paperclip jar for this pull request: [paper-9413.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1088613506.zip)

### DIFF
--- a/patches/server/0176-Extend-Player-Interact-cancellation.patch
+++ b/patches/server/0176-Extend-Player-Interact-cancellation.patch
@@ -16,19 +16,7 @@ diff --git a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java 
 index 16468dfd657501f6fd5eefa4c32682e24eaba22d..7adf7d1248680e591638298f13f5ae88788f5e4c 100644
 --- a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
 +++ b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
-@@ -186,6 +186,11 @@ public class ServerPlayerGameMode {
-                 PlayerInteractEvent event = CraftEventFactory.callPlayerInteractEvent(this.player, Action.LEFT_CLICK_BLOCK, pos, direction, this.player.getInventory().getSelected(), InteractionHand.MAIN_HAND);
-                 if (event.isCancelled()) {
-                     // Let the client know the block still exists
-+                    // Paper start - brute force neighbor blocks for any attached blocks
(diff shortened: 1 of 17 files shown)


patches/server/0982-Dont-resend-blocks-on-interactions.patch
```diff
@@ -0,0 +1,149 @@
...
+-                    this.player.connection.send(new ClientboundBlockUpdatePacket(this.level, pos));
++                    // Paper - Don't resync blocks
+                     // Update any tile entity data for this block
+                     capturedBlockEntity = true; // Paper - send block entity after predicting
+                     return;
+@@ -214,7 +214,7 @@ public class ServerPlayerGameMode {
+                 // Spigot start - handle debug stick left click for non-creative
+                 if (this.player.getMainHandItem().is(net.minecraft.world.item.Items.DEBUG_STICK)
+                         && ((net.minecraft.world.item.DebugStickItem) net.minecraft.world.item.Items.DEBUG_STICK).handleInteraction(this.player, this.level.getBlockState(pos), this.level, pos, false, this.player.getMainHandItem())) {
+-                    this.player.connection.send(new ClientboundBlockUpdatePacket(this.level, pos));
```
electronicboy: If you're gonna add comments like this, please comment the line instead of drop + comment

Owen1212055:
Rebased.

It should be noted here that in many cases, blocks are still manually resent. This is mostly in cases where it's obvious there is a desync.
Screenshot 2023-11-11 at 6 25 15 PM

electronicboy (approved):
Approving on a code perspective and my general understanding of the underlying mechanism, will prefer for others to test, etc

kennytv (approved):
I'm sure bug reports will tell us whether Owen broke everything again

PR #9419, opened 2023-06-29
Fix default permission for xp command

Machine-Maker:
Should've actually checked that the one special case cb had for the `/experience` command was actually correct. Turns out its been wrong for a long time. CB un-redirects vanilla command nodes, so when checking perms for the `/xp` command node, it uses the `/experience` command node for the literal string (`experience`) and so the permission should be `minecraft.command.experience`. Don't think `minecraft.command.xp` has worked at all recently.

----

Depending on how much we care about permission descriptions being more specific than "Lets you use /somecommand", these could just be completely autogenerated in the future. Using the vanilla perm system to figure out what the default should be, OP or TRUE.


kennytv (approved):

dandud100:
/tm and it's alias cmds should also be disabled I think

PR #9431, opened 2023-07-01
fix bed placement bug

outlineutils:
fix MC-123619

### DIFF
--- a/patches/server/0984-Fix-bed-placement-incorrect-hitbox-check.patch
+++ b/patches/server/0984-Fix-bed-placement-incorrect-hitbox-check.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: JNI <admin@hydraclient.com>
+Date: Fri, 30 Jun 2023 22:16:16 -0400
+Subject: [PATCH] Fix-bed-placement-incorrect-hitbox-check
+
+
+diff --git a/src/main/java/net/minecraft/world/level/block/BedBlock.java b/src/main/java/net/minecraft/world/level/block/BedBlock.java
+index d40500f9a807cab0b2fb6fa9032f33f4fb74c895..43a8d758d795748fe1f705d1ce864b46c9d28c2e 100644
+--- a/src/main/java/net/minecraft/world/level/block/BedBlock.java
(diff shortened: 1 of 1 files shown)


MiniDigger:
this is a pretty big behavior change, what is your reasoning for wanting this fixed?

DerEchtePilz:
> this is a pretty big behavior change, what is your reasoning for wanting this fixed?

I could imagine they want it to be fixed because it is a bug and it exists a long time.

catgirljenni:
People can place beds inside peoples faces in nether pvp when they are in a hole, forcing them into the crouching animation so they can't even pearl out, while dealing massive damage (1 bed kills in most cases). In older game versions this bug was present but wasn't an issue since you couldn't place beds unless you had two supporting blocks beneath it, but since that was removed this bug is now very easy to exploit. Since there's no practical use for this mechanic outside of cheesing pvp, and it's been acknowledged as a bug by mojang, I don't think there is any negative to fixing it.

Machine-Maker:
The mojira issue talks about all "multi block" blocks being affected, but this fix only seems to be for bed blocks? Perhaps it should be expanded to all blocks. 

I am for fixing this in general, it is a confirmed mojang bug. But because it is an old one without any we may have reason to make it configurable.

DerEchtePilz:
> Would also be in favour of hiding this behind a configuration option. Especially with doors, the mechanic of "boosting" yourself by jumping > placing the door > landing partially in the door comes to mind.

I personally wouldn't do that. If Mojang fixed that bug, that "mechanic" would also not exist anymore.
But still, I let the Paper team decide, I've got no power anyway, I just wanted to express my thoughts.

catgirljenni:
> Especially with doors, the mechanic of "boosting" yourself by jumping > placing the door > landing partially in the door comes to mind.

This is why I only fixed it for beds

Owen1212055:
I’m a bit iffy about this, in general this is an imperfect solution and doesn’t cover all blocks like this. May be a better thing to just have as a plugin, or just wait for vanilla to resolve such.

underscore11code:
Due to this not covering all cases of the original bug, I'm going to close this. Feel free to submit a new PR which fixes all multiblocks (due to the age, putting the fix behind a config option would also be good). Thank you for your contribution.

PR #9436, opened 2023-07-02
Fix breaking tripwires with shears

DungeonDev:
A bug in vanilla causes tripwires broken by shears to go into a weird state, see https://bugs.mojang.com/browse/MC-129055
In this state, destroying the tripwire will not work, it will instead disconnect from tripwire hooks for 10 ticks before connecting again. Destroying the tripwire during this 10 tick window will actually destroy the tripwire.

Abusing this, and using water to repeatedly destroy the tripwire, an infinite amount of string can be easily duplicated in a fast manner, as the tripwire will drop one string when "destroyed" by water.

It seems tripwire in this bugged state was previously accepted as a feature, see #7259 and #7261
This PR fixes the issue by moving a line of code down one line into an empty if-statement by Mojang...
If bugged "disarmed" tripwire is something we'd like to keep around, a more thorough solution is needed, as the current "unbreakable" behavior is quite unintuitive.

---
Download the paperclip jar for this pull request: [paper-9436.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/861343771.zip)

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -440,10 +440,10 @@ index 0000000000000000000000000000000000000000..9ef6712c70fcd8912a79f3f61e351aac
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..5cfd0bd7c09aacd7313575e30cf0a171c597e2bd
+index 0000000000000000000000000000000000000000..fc513278f2e410551f70b11f95138f960941cb34
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,291 @@
+@@ -0,0 +1,294 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
@@ -597,6 +597,9 @@ index 0000000000000000000000000000000000000000..5cfd0bd7c09aacd7313575e30cf0a171
 +            ZLIB,
 +            NONE
 +        }
++
++        @Comment("This setting controls if shears should be able to leave behind disarmed tripwire.")
++        public boolean allowShearsDisarmedTripwire = false;
 +    }
 +
 +    public Commands commands;
--- a/patches/server/0019-Rewrite-chunk-system.patch
+++ b/patches/server/0019-Rewrite-chunk-system.patch
@@ -15687,7 +15687,7 @@ index 0000000000000000000000000000000000000000..962d3cae6340fc11607b59355e291629
 +
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-index 5cfd0bd7c09aacd7313575e30cf0a171c597e2bd..115c44d6f064049270f1fae07683da1da974b08e 100644
+index fc513278f2e410551f70b11f95138f960941cb34..4274b43069d9aca4d6018b96e5446dcdd2eb225f 100644
 --- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 @@ -116,21 +116,6 @@ public class GlobalConfiguration extends ConfigurationPart {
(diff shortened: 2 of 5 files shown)


kennytv (changes requested):
I'm not sure what the intention behind this patch is. It doesn't fix duplication for existing contraptions, it just disables disarming tripwires? ~~Disarming without breaking has basically been accepted as a feature at this point is the problem, so we don't want to rip it out entirely~~

kennytv:
I'd maybe add a config option for it since it goes beyond just a dupe, and have the fix enabled by default

welove69:
That's great，Would very much like to fix this bug

DungeonDev:
> Disarming without breaking has basically been accepted as a feature at this point is the problem, so we don't want to rip it out entirely

Would something like this be better? We keep disarmed wires around, but make them actually break on subsequent block breaks. The code is a bit ugly by setting a block in onRemove, but it already did that previously, just hidden further away in TripWireHookBlock. This has the by-effect of making all tripwires broken by shears being disarmed on the first break, not just ones currently in a tripwire circuit - making it a bit more consistent.

kennytv:
Having thought about it a little more, I think the proper feature fix you had before is the best option as long as you add a config option to revert it in the `unsupported` section (see `0005-Paper-config-files.patch`). After the initial attempt at fixing the dupe without changing mechanics, we had hoped Mojang would possibly fix either soon enough, but given that's not the case and there's no sign of it either unfortunately, having the proper fix is a lot easier to maintain

patches/server/0005-Paper-config-files.patch
```diff
@@ -589,6 +589,8 @@ index 0000000000000000000000000000000000000000..58cdb70c0b9de45ca28811416d871099
 +        public boolean allowHeadlessPistons = false;
 +        @Comment("This setting controls if grindstones should be able to output overstacked items (such as cursed books).")
 +        public boolean allowGrindstoneOverstacking = false;
++        @Comment("This setting controls if shears should be able to create disarmed tripwire, allowing string duplication.")
```
kennytv:
```suggestion
+        @Comment("This setting controls if shears should be able to leave behind disarmed tripwire.")
```

kennytv (commented):
Thanks, just a small request. Could you also rebase the pr?

DungeonDev:
Rebased.

welove69:
Great
item dupe bug should be fixed

welove69:
Can this PR be merged, looks great
Please do not let this die in pull request queue.

welove69:
Fix breaking tripwires with shears, fixing string dupe

https://github.com/PaperMC/Paper/assets/135555687/85355234-4103-4347-b540-baa226a77493

Works perfectly,even offers a switch option

welove69:
allow-shears-disarmed-tripwire: true
https://github.com/PaperMC/Paper/assets/135555687/940c04e0-18fa-41e3-8823-82d40a8be309

welove69:
allow-shears-disarmed-tripwire: true

https://github.com/PaperMC/Paper/assets/135555687/fc40215e-1793-446f-abd3-ffe056c8c2ee

welove69:
allow-shears-disarmed-tripwire: true

https://github.com/PaperMC/Paper/assets/135555687/335739c3-ed81-4d56-a0c6-51cf87d3e53d

Huge-mistake:
Test everything works as expected.

Huge-mistake:
Test everything works as expected.
Can the code approve the merge.
꒰^ↀωↀ^꒱❤

welove69:
> 香草中的错误导致被剪断的绊线进入奇怪的状态，请参阅 https://bugs.mojang.com/browse/MC-129055 在这种状态下，破坏绊线将不起作用，而是在再次连接之前断开绊线钩 10 个刻度。在这 10 个刻度窗口期间破坏绊线实际上会破坏绊线。
> 
> 滥用这一点，用水反复破坏绊线，可以快速复制无限量的绳子，因为绊线在被水“破坏”时会掉落一根绳子。
> 
> 似乎这种错误状态下的绊线以前被接受为一项功能，请参阅 #7259 和 #7261 此 PR 通过将一行代码向下移动到 Mojang 的空 if 语句中来解决此问题......如果我们想保留错误的“解除武装”绊线，则需要更彻底的解决方案，因为当前的“牢不可破”行为非常不直观。
> 
> 下载此拉取请求的回形针罐：[paper-9436.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/802283327.zip)

Can confirm that it works fine.
fix https://github.com/PaperMC/Paper/issues/9542

CompeyDev:
Since it seems unclear in this PR, has tripwire duping been patched? If so, is there a configuration option to opt into allowing it as discussed in #9542?

electronicboy:
Some variation of tripwire duping remains, this should patch them, however, it has wider behavioral changes for the solution, and so, this specific variant has a configuration option
Do note that we generally do not add configuration options for bug fixes

Huge-mistake:
> Since it seems unclear in this PR, has tripwire duping been patched? If so, is there a configuration option to opt into allowing it as discussed in #9542?
> 
> I can't seem to find in the docs, so I'm guessing this PR was abandoned and superseded by something else perhaps (since [certain users](https://www.reddit.com/r/Minecraft/comments/yy1pjz/papermc_duping_disabled/) report being unable to make use of this duplication glitch even though this PR might not have been merged).`allow-shears-disarmed-tripwire`

TNT isn't it?

CompeyDev:
> Some variation of tripwire duping remains, this should patch them, however, it has wider behavioral changes for the solution, and so, this specific variant has a configuration option Do note that we generally do not add configuration options for bug fixes

Why should only one variation of tripwire duping include a configuration option over the other kinds?

welove69:
> I'd maybe add a config option for it since it goes beyond just a dupe, and have the fix enabled by default

The option is optional, and six months have passed, Can it be merged

PR #9443, opened 2023-07-03
Fix mob breaking doors not spawning particles

Machine-Maker:
This is a vanilla issue reported as [MC-263999](https://bugs.mojang.com/browse/MC-263999)

Note that the mojira issue isn't marked as confirmed, but I'm pretty sure its an issue and I did just open that ticket. Can wait to merge if you want to wait for a response on that ticket.


Owen1212055 (approved):
Idk if we need confirmation here, defo seems unintentional.

Machine-Maker:
Well, it was confirmed either way, :joy: .

PR #9446, opened 2023-07-03
Call BlockGrowEvent for more blocks

Lulu13022002:
Call the BlockGrowEvent for the following situation:
* the mangrove propagule and when bee grow crops on their way
* sweet berry bush, cave vines (including the plant), kelp, twisting vines, weeping vines, sapling, grass/seagrass, bigdripleaf (stem+leaf, i didn't include the small dripleaf for now since there's some issue on spigot) using bonemeal
* cactus/sugar cane for each age grow and not only when a new block grow

Generally those events are weird because they state to be called only for natural way/world condition but spigot call them using bone meal so...

I have also added the source block to the BlockGrowEvent for consistency and fixed some cancellability issues.


patches/server/0969-Call-BlockGrowEvent-BlockSpreadEvent-for-missing-blo.patch
```diff
@@ -0,0 +1,80 @@
...
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/animal/Bee.java b/src/main/java/net/minecraft/world/entity/animal/Bee.java
+index f6062bf8c888baeb7b421150a2c64bf1af1a312b..b4210902c9fe2a018ef94cf79b50a358b5f1fd56 100644
+--- a/src/main/java/net/minecraft/world/entity/animal/Bee.java
++++ b/src/main/java/net/minecraft/world/entity/animal/Bee.java
+@@ -1243,8 +1243,8 @@ public class Bee extends Animal implements NeutralMob, FlyingAnimal {
+                         }
+ 
+                         if (iblockdata1 != null&& !org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(Bee.this, blockposition, iblockdata1).isCancelled()) { // CraftBukkit
```
Machine-Maker: I think you are missing some event calls here. A few lines above, the `performBonemeal` method is called for cave vines. That means that the grow event below won't be fired because that local var `iblockdata1` will still be null.

patches/api/0421-Add-source-block-to-the-BlockGrowEvent.patch
```diff
@@ -0,0 +1,100 @@
...
+     public BlockFormEvent(@NotNull final Block block, @NotNull final BlockState newState) {
+-        super(block, newState);
++        this(block, null, newState); // Paper
+     }
+ 
++    // Paper start
++    public BlockFormEvent(@NotNull final Block block, @org.jetbrains.annotations.Nullable final Block source, @NotNull final BlockState newState) {
++        super(block, source, newState);
++    }
++    // Paper end
```
Machine-Maker: I think you can just avoid this. Just keep the source Block field in both the Grow and Spread event, overriding the getSource method in the spread event to return it's own field.

I think its a smaller diff as well.

patches/api/0421-Add-source-block-to-the-BlockGrowEvent.patch
```diff
@@ -0,0 +1,100 @@
...
++    private final Block source; // Paper
+     private boolean cancelled = false;
+ 
++    @Deprecated // Paper
+     public BlockGrowEvent(@NotNull final Block block, @NotNull final BlockState newState) {
++        this(block, null, newState); // Paper
++    }
++
++    // Paper start
++    public BlockGrowEvent(@NotNull final Block block, @org.jetbrains.annotations.Nullable final Block source, @NotNull final BlockState newState) { // Paper
```
Machine-Maker: Add `@Internal` here. I think its good we do that for new event constructors to further emphasize that they can change whenever.

patches/api/0421-Add-source-block-to-the-BlockGrowEvent.patch
```diff
@@ -0,0 +1,100 @@
...
++        super(block, source, newState); // Paper
+     }
+ 
+     /**
+@@ -37,7 +34,7 @@ public class BlockSpreadEvent extends BlockFormEvent {
+      */
+     @NotNull
+     public Block getSource() {
+-        return source;
++        return super.getSource(); // Paper
```
Machine-Maker: See what I said above, I think you can leave this, and just override getSource() to return the field in this event instead of passing it up the inheritance chain.

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
```diff
@@ -0,0 +1,176 @@
...
+index 0003fb51ae3a6575575e10b4c86719f3061e2577..c02f7b0dbf8aaafa76da99a76f1ff87e1a1f0274 100644
+--- a/src/main/java/net/minecraft/world/level/block/CactusBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/CactusBlock.java
+@@ -59,13 +59,13 @@ public class CactusBlock extends Block {
+ 
+                 int modifier = world.spigotConfig.cactusModifier; // Spigot - SPIGOT-7159: Better modifier resolution
+                 if (j >= 15 || (modifier != 100 && random.nextFloat() < (modifier / (100.0f * 16)))) { // Spigot - SPIGOT-7159: Better modifier resolution
+-                    CraftEventFactory.handleBlockGrowEvent(world, blockposition1, this.defaultBlockState()); // CraftBukkit
++                    if (!CraftEventFactory.handleBlockGrowEvent(world, blockposition1, pos, this.defaultBlockState(), 3)) return; // CraftBukkit // Paper - properly cancel
+                     BlockState iblockdata1 = (BlockState) state.setValue(CactusBlock.AGE, 0);
```
Machine-Maker, line 98: I think that the `BlockState` argument to `handleBlockGrowEvent` should be the `iblockdata1` that is below instead of the returned getDefaultBlockState. They should always be the same now, but what if cactus somehow get another blockstate property or whatever.

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
```diff
@@ -0,0 +1,176 @@
...
+             if (i >= 3) {
+                 BlockPos blockPos = pos.above();
++                if (!org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockGrowEvent(world, blockPos, pos, copyWaterloggedFrom(world, pos, this.defaultBlockState().setValue(AGE, Integer.valueOf(i)).setValue(HALF, DoubleBlockHalf.UPPER)), 2)) return; // Paper
+                 world.setBlock(blockPos, copyWaterloggedFrom(world, pos, this.defaultBlockState().setValue(AGE, Integer.valueOf(i)).setValue(HALF, DoubleBlockHalf.UPPER)), 3);
+             }
++            // Paper start
++            else {
++                org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockGrowEvent(world, pos, state.setValue(AGE, Integer.valueOf(i)), 2);
++            }
++            // Paper end
```
Machine-Maker: Just confirming that removing the double setBlock isn't gonna break any obscure mechanic right? something that relies on updating something somewhere maybe?

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
```diff
@@ -0,0 +1,176 @@
...
++++ b/src/main/java/net/minecraft/world/entity/animal/Bee.java
+@@ -1247,8 +1247,8 @@ public class Bee extends Animal implements NeutralMob, FlyingAnimal {
+                         }
+ 
+                         if (iblockdata1 != null && CraftEventFactory.callEntityChangeBlockEvent(Bee.this, blockposition, iblockdata1)) { // CraftBukkit
++                            if (!org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockGrowEvent(Bee.this.level(), blockposition, iblockdata1, 3)) continue; // Paper
+                             Bee.this.level().levelEvent(2005, blockposition, 0);
+-                            Bee.this.level().setBlockAndUpdate(blockposition, iblockdata1);
+                             Bee.this.incrementNumCropsGrownSincePollination();
+                         }
```
Machine-Maker, line 21: Having the event here misses 2 instances of bee-caused growth. Cave vines don't go into this conditional block since `iblockdata1` will still be null.

Machine-Maker (changes requested):

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
Lulu13022002: what do you mean?

Machine-Maker: Like in vanilla, if `i >= 3`, there will be 2 consecutive setBlock calls right? the one being deleted here, and the one below the first grow event handler. Removing one of them isn't going to break anything right?

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
Lulu13022002: the blockstate below is the sourceblock but the grow event called when a new block grow always have the new block data instead which is the defaultBlockState see here: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/world/level/block/BlockCactus.patch#17 the event call replace the setblock

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
Lulu13022002: the event call is the setblock, the flags are just mangled

patches/server/0966-Call-BlockGrowEvent-for-missing-blocks.patch
Lulu13022002: i don't think this is needed cave vines just get berries here i think otherwise need to be called when using bonemeal for players too

Machine-Maker: Yeah, I think its worth calling for berries. It is growing, but yeah, you can put it in the performBonemeal method.

patches/api/0421-Add-source-block-to-the-BlockGrowEvent.patch
```diff
@@ -0,0 +1,53 @@
...
+     /**
+      * Gets the state of the block where it will form or spread to.
+@@ -41,6 +50,20 @@ public class BlockGrowEvent extends BlockEvent implements Cancellable {
+         return newState;
+     }
+ 
++    // Paper start
++    /**
++     * Gets the source block involved in this event.
++     * The block can be null when the block grow without
```
Machine-Maker:
```suggestion
+     * The block can be null when the block grows without
```

patches/api/0421-Add-source-block-to-the-BlockGrowEvent.patch
```diff
@@ -0,0 +1,53 @@
...
++     * creating a new block but instead modifying its properties.
```
Machine-Maker:
```suggestion
+     * creating a new block but ageing itself.
```

Machine-Maker (changes requested):

PR #9460, opened 2023-07-08
Array backed synched entity data

Warriorrrr:
Improves the speed of synched entity data lookups by using an array instead of a map. The difference probably isn't too drastic but given how many places it's called from it's still nice. 

This patch is originally from lithium and can be found [here](https://github.com/CaffeineMC/lithium-fabric/blob/ae1bd56556248aceba51f1d27e1317b0958448b1/src/main/java/me/jellysquid/mods/lithium/mixin/entity/data_tracker/use_arrays/DataTrackerMixin.java).

### DIFF
--- a/patches/server/0983-Array-backed-synched-entity-data.patch
+++ b/patches/server/0983-Array-backed-synched-entity-data.patch
@@ -1,0 +1,56 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: jellysquid3 <jellysquid3@users.noreply.github.com>
+Date: Sat, 8 Jul 2023 21:38:05 +0200
+Subject: [PATCH] Array backed synched entity data
+
+Original code by jellysquid3 in Lithium, licensed under the GNU Lesser General Public License v3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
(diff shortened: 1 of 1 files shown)


patches/server/0983-Array-backed-synched-entity-data.patch
```diff
@@ -0,0 +1,55 @@
...
+     public <T> boolean hasItem(EntityDataAccessor<T> key) {
+@@ -130,7 +144,14 @@ public class SynchedEntityData {
+ 
+         return datawatcher_item;
+         */
+-        return (SynchedEntityData.DataItem) this.itemsById.get(key.getId());
++        // Paper start - array backed synched entity data
++        final int id = key.getId();
++
++        if (id < 0 || id >= this.itemsArray.length)
```
kennytv: Add curly braces

patches/server/0983-Array-backed-synched-entity-data.patch
```diff
@@ -0,0 +1,55 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Warrior <50800980+Warriorrrr@users.noreply.github.com>
```
kennytv: Author should be the original code author with a link to the license in the patch description

kennytv (changes requested):

kennytv (approved):

PR #9461, opened 2023-07-09
 Add Entity Coordinate and Rotation API

TonytheMacaroni:
Added methods to get an entity's yaw/pitch directly.

### DIFF
--- a/patches/api/0406-Add-Entity-Body-Yaw-API.patch
+++ b/patches/api/0406-Add-Entity-Body-Yaw-API.patch
@@ -4,6 +4,54 @@ Date: Thu, 2 Feb 2023 16:40:11 +0100
 Subject: [PATCH] Add Entity Body Yaw API
 
 
+diff --git a/src/main/java/org/bukkit/entity/Entity.java b/src/main/java/org/bukkit/entity/Entity.java
+index a2a423d4e4c2702ba5967223cab0432dd7d04732..7b944a4fde359e169da9188fa63cf937f5329091 100644
+--- a/src/main/java/org/bukkit/entity/Entity.java
++++ b/src/main/java/org/bukkit/entity/Entity.java
+@@ -931,6 +931,43 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
+      * @return true if in powdered snow.
+      */
+     boolean isInPowderedSnow();
++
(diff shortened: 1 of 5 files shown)


frqnny:
I disagree with this PR. It’s already exposed through Location- why add these two random helpers here? Where is the setter methods that should go along with it? These methods should remain in Location unless some new API gets made to wrap around BlockPos

TonytheMacaroni:
Seems unnecessary to restrain getting an entity's pitch/yaw to Location, especially for times when you only need its rotation and not its position. As for setters, I didn't add them as there's already [Entity#setRotation](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Entity.html#setRotation(float,float)), which fulfills a similar purpose.

Sytm:
I don't know what the ultimate goal of the Position API is, but what if the entity would implement the FinePosition interface, just as an idea

Owen1212055:
Uhh a #position() method would be better there. I'd consider that abuse of the position API.

Sytm:
Yeah, I almost guessed that it would be abuse. #position() isn't really required because Location implements Position already (except if you would want to move away from Location in the future)

TonytheMacaroni:
> How about we just add methods to expose getX()/getY()/getZ() too?

Added. Also moved the methods into the body yaw patch, since they're somewhat related.

kennytv (approved):
While kind of cluttery, getting individual values of an entity's location is done often enough that I think it's worth exposing directly

patches/server/0929-Add-Entity-Body-Yaw-API.patch
```diff
@@ -4,6 +4,42 @@ Date: Thu, 2 Feb 2023 16:40:41 +0100
...
++    }
++
++    @Override
++    public float getPitch() {
++        return this.entity.getXRot();
++    }
++
++    @Override
++    public float getYaw() {
++        return this.entity.getYRot();
```
Lulu13022002: This will not return the bukkit yaw and will not match #getLocation, you can see the bukkit implementation of #getLocation to behave the same for living entities

TonytheMacaroni: Should correctly use #getBukkitYaw now.

PR #9466, opened 2023-07-10
SculkCatalyst bloom API

### LINKED ISSUE #9198: Method for Sculk Catalyst spreading Sculk

As far as I've looked and asked, there isn't yet an API method for making a sculk catalyst spread sculk. Usually this happens when a mob dies and tries to drop XP, which is handled by a NMS method (attached).
nms

It would be nice to have a method in the SculkCatalyst block data class that would take in a location and experience points, and attampt to spread sculk from a catalyst. Something like SculkCatalyst#growSculk(Location at, int exp);

You can probably do it the boring way, by picking random blocks in an area and seeing if the material matches Tag.SCULK_REPLACEABLE, and replacing afterwards, but that's not gonna look natural, is it?

### Other

image

Nacioszeczek:
Resolves PaperMC/Paper#9198


patches/api/0421-SculkCatalyst-bloom-API.patch
```diff
@@ -0,0 +1,23 @@
...
+  */
+ public interface SculkCatalyst extends TileState {
++
++    /**
++     * Bloom at the specified location as if an entity that drops experience just died there.
++     *
++     * @param location location to bloom at
++     * @param charge charge to bloom with, normally the amount of experience dropped from the dead entity
++     */
++    void bloom(org.bukkit.Location location, int charge);
```
Owen1212055: Missing paper comments here

patches/server/0984-SculkCatalyst-bloom-API.patch
```diff
@@ -0,0 +1,35 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/block/CraftSculkCatalyst.java
++++ b/src/main/java/org/bukkit/craftbukkit/block/CraftSculkCatalyst.java
+@@ -9,4 +9,22 @@ public class CraftSculkCatalyst extends CraftBlockEntityState<SculkCatalystBlock
+     public CraftSculkCatalyst(World world, SculkCatalystBlockEntity tileEntity) {
+         super(world, tileEntity);
+     }
++
++    // Paper start - SculkCatalyst bloom API
++    @Override
++    public void bloom(org.bukkit.Location location, int charge) {
```
Owen1212055: In general, you can use the Position type instead of a location. This is because we don't want to have a world be a possible parameter, as it will always use the sculk's world anyway.

patches/server/0984-SculkCatalyst-bloom-API.patch
```diff
@@ -0,0 +1,35 @@
...
++        requirePlaced();
++
++        getTileEntity().getListener().bloom(
++            world.getHandle(),
++            getTileEntity().getBlockPos(),
++            getTileEntity().getBlockState(),
++            world.getHandle().getRandom()
++        );
++        getTileEntity().getListener().getSculkSpreader().addCursors(
++            new net.minecraft.core.BlockPos(position.blockX(), (int) (position.blockY() + 0.5D), position.blockZ()),
```
Machine-Maker: Should we be doing this position mutation with the API method? I feel like its to offset from the emitter position within the game, but the API can specify whatever position it wants.

Machine-Maker (changes requested):

patches/server/0984-SculkCatalyst-bloom-API.patch
Nacioszeczek: The +0.5 is copied directly from game's code. I assume it's for cases like the entity standing on e.g. a slab, if someone just passes an entity's Position imho should still add that half.

Unless I don't know what the emitterPos is

Nacioszeczek: ~~Or we could just have an 'overload' for LivingEntity~~ Even if someone specifies a block position that 0.5 ~~won't change anything since it's cast back to int~~ (nvm negative values), for an entity's position it may change things

Machine-Maker (approved):

lynxplay (approved):

PR #9470, opened 2023-07-10
improve checking handled tags in itemmeta

### LINKED ISSUE #9469: NBT data loss on item
### Expected behavior

I have a datapack that allows for applying armor trims to items. When an armor trim is applied I expect the NBT data associated with the trim to remain on the item while moving between inventories, dropped on the ground, etc.

### Observed/Actual behavior

When armor trim is applied if a player moves the item between inventory slots too quickly the NBT data for the trim simply vanishes into the void.

### Steps/models to reproduce

**Steps I took to replicate:**
- booted fresh 1.20.1-69 server on local
- installed datapack to allow for tools to have armor trim applied (nothing else, no plugins, no config changes, etc)
- restarted server
- gave myself op, changed gamemode to creative, gave myself needed items for recipe
- apply template to tool (pickaxe primarily used for testing)
- quickly randomly move tool around inventory/rapidly click on it (if its slow enough it doesnt seem to happen)
- NBT data for trim is lost.

### Plugin and Datapack List

Server Plugins (0):
There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [file/dp-trimmable-tools-mc1.20-v1.0.4.zip (world)]

### Paper version

This server is running Paper version git-Paper-69 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 6b325cd)

### Other

**Notes:**
- I was not able to replicate this on a vanilla local server, or singleplayer
- I also tested it in survival and experienced the same NBT data loss
- all tests were done with a vanilla client (resource packs only)
- other (vanilla) NBT data (for enchants, names, etc) is not lost

[2023-07-10-2.log](https://github.com/PaperMC/Paper/files/12008605/2023-07-10-2.log) (log is second file because it was restarted after adding datapack)
[dp-trimmable-tools-mc1.20-v1.0.4.zip](https://github.com/PaperMC/Paper/files/12008606/dp-trimmable-tools-mc1.20-v1.0.4.zip)

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/9469

----

Issue here is how CB was checking if an nbttag key was handled or not. It didn't check for handled keys specific to that item meta, it checked if any item meta used that key or not. This PR changes it to a type specific check. Also adds a test to make sure all subtypes of CraftMetaItem are present in that map.

### DIFF
--- a/patches/server/0961-Deprecate-ItemStack-setType.patch
+++ b/patches/server/0961-Deprecate-ItemStack-setType.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Deprecate ItemStack#setType
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
-index 2ba771efa61e109804f3141e95f77613ac952ed1..39e46f3f4b51fce9bc492fa498aa7427b27b78d1 100644
+index 2ba771efa61e109804f3141e95f77613ac952ed1..6b794776c1e7a00b7a00a1379cf559be182996cd 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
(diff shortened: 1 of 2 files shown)


Machine-Maker:
Rebased for 1.20.4

kennytv (approved):

PR #9475, opened 2023-07-12
Redirect `ServerPlayer#createEndPlatform` to `ServerLevel#makeObsidianPlatform`

nostalfinals:
It seems that Mojang wrote the same function at 2 different places, which is very annoying.
This patch redirects `ServerPlayer#createEndPlatform` to `ServerLevel#makeObsidianPlatform` to fix that problem.

### DIFF
--- a/patches/server/0985-Redirect-ServerPlayer-createEndPlatform-to-ServerLev.patch
+++ b/patches/server/0985-Redirect-ServerPlayer-createEndPlatform-to-ServerLev.patch
@@ -1,0 +1,69 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: nostalfinals <yuu8583@proton.me>
+Date: Wed, 12 Jul 2023 16:34:25 +0800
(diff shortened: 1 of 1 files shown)


TheFruxz:
Im not into this and my assumption might be wrong, but does not the one thing create it from the player (like someone jump through the end portal) and the other one from the server (like generating a new end world), so it can end up to be changed in the future?

MiniDigger:
while code duplication is obviously meh, whats the reason you care here? this looks to me like more diff for us to maintain without any gain?

Lulu13022002:
Those are similar however they are not the same the #createEndPlatform will create the platform one block below the centerPos. But the #makeObsidianPlatform creates the platform two blocks below the centerPos so this is a vanilla change for players (platform will spawn one block below than vanilla and so the teleport will be one block above the platform)

Machine-Maker:
I'm not really a fan of adding diff for this, its not "fixing" anything, and it looks like there are behavioral differences.

PR #9480, opened 2023-07-13
Fix DEFAULT getEntitySpawnReason for /summon

Warriorrrr:
Fixes an issue where summoned entities wouldn't have the spawn reason of COMMAND, but instead defaulted to default/natural/spawner inside the Entity#load method before the command reason could be set.

### DIFF
--- a/patches/server/0300-Entity-getEntitySpawnReason.patch
+++ b/patches/server/0300-Entity-getEntitySpawnReason.patch
@@ -9,6 +9,18 @@ Pre existing entities will return NATURAL if it was a non
 persistenting Living Entity, SPAWNER for spawners,
 or DEFAULT since data was not stored.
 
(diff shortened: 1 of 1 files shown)


kennytv (approved):

Machine-Maker:
Does `Player#respawnEnttiyOnShoulder0` suffer from the exact same problem? The nbt loading of an entity before then adding it to the world with a custom spawn reason? I think the entity's spawn reason isn't overriden with `SHOULDER_ENTITY` like it should.

Warriorrrr:
The entity's original SpawnReason will be set so it won't be overridden with the SHOULDER_ENTITY one, I think keeping the original would be more in line with what the javadoc message says.

lynxplay (approved):

PR #9488, opened 2023-07-14
Fix main thread priority being lowered

Warriorrrr:
This PR fixes vanilla's thread priority of 8 being overridden with 7 (regular priority of 5 + 2). I'm not too sure how much of a difference this makes or if this line is still even needed, but a higher priority seems better than a lower one.

---
Download the paperclip jar for this pull request: [paper-9488.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/863108412.zip)


Owen1212055 (approved):
Apparently, this has no effect on most operating systems, only Windows.

lynxplay (approved):

NoahvdAa (approved):

Machine-Maker (approved):

PR #9491, opened 2023-07-14
Fix MC-65198

elmital:
Closing accidentaly the previous PR (#9132) during the rebase however I've found a cleaner way to fix the bug.

### DIFF
--- a/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
@@ -48,6 +48,9 @@ https://bugs.mojang.com/browse/MC-123848
   by: BillyGalbreath <blake.galbreath@gmail.com>
   Fixes item frames dropping items above when pointing down
 
+https://bugs.mojang.com/browse/MC-65198
+  Fixes ITEM_CRAFTED statistic isn't always incremented correctly
+
 Co-authored-by: William Blake Galbreath <blake.galbreath@gmail.com>
 
 diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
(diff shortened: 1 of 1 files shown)


toinouH:
Maybe you should move your patch to that one instead https://github.com/PaperMC/Paper/blob/master/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
```diff
@@ -0,0 +1,56 @@
...
+diff --git a/src/main/java/net/minecraft/world/inventory/ResultSlot.java b/src/main/java/net/minecraft/world/inventory/ResultSlot.java
+index ddf87ef1979c4ae4f14cd6aa220f9146d5e20909..4a8529ebb1bfce015bd77a35076fcc3bee33fa90 100644
+--- a/src/main/java/net/minecraft/world/inventory/ResultSlot.java
++++ b/src/main/java/net/minecraft/world/inventory/ResultSlot.java
+@@ -45,7 +45,7 @@ public class ResultSlot extends Slot {
+     @Override
+     protected void checkTakeAchievements(ItemStack stack) {
+         if (this.removeCount > 0) {
+-            stack.onCraftedBy(this.player.level(), this.player, this.removeCount);
++            stack.onCraftedBy(this.player.level(), this.player, stack.getCount()); // Paper - fix MC-65198 for drop action in result slot
```
Machine-Maker: This removes the use of the `removeCount` field entirely. I cannot see a use case where it is actually needed, but I think caution should be taken if we just do this without considering what the field is actually used for, or why it exists.

For example, the `removeCount` field is absolutely required for the `FurnaceResultSlot` which is largely based on ResultSlot. However, it seems everywhere ResultSlot is used, the whole stack in the slot is removed always which would mean using stack is fine here. But we need to be sure.

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
```diff
@@ -0,0 +1,56 @@
...
+             public void onTake(Player player, ItemStack stack) {
+                 ItemCombinerMenu.this.onTake(player, stack);
+             }
++
++            // Paper start - fix MC-65198 shift clicking in result slot of a smiting table
++            @Override
++            protected void onQuickCraft(ItemStack stack, int amount) {
++                stack.onCraftedBy(player.level(), player, amount);
++            }
++            // Paper end
```
Machine-Maker: Does this mean it runs for items from Anvil's as well? AnvilMenu extends ItemCombinerMenu and uses the same logic for setting up slots. I don't think stats should be counted when an item is taken from an anvil.

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
```diff
@@ -0,0 +1,56 @@
...
+@@ -114,6 +114,13 @@ public class StonecutterMenu extends AbstractContainerMenu {
+                 super.onTake(player, stack);
+             }
+ 
++            // Paper start - fix MC-65198 shift clicking in the result slot of StoneCutter
++            @Override
++            protected void onQuickCraft(ItemStack stack, int amount) {
++                stack.onCraftedBy(level, playerInventory.player, amount);
++            }
++            // Paper end
```
Machine-Maker: I think this is missing some of the other logic, like awarding recipes perhaps?

Machine-Maker (changes requested):

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
elmital: I haven't really found an use of the field either excluding like you said in the `FurnaceResultSlot`. But without be certain I prefered to let it like this. The whole stuff is a little bit messy from my point of view for example the crafting statistic increase methods are called twice. In the past Mojang has already fix [issue](https://bugs.mojang.com/browse/MC-29515) with statistics incrementation so maybe the field should be a remainder of that? The incrementation is called a first time with a positive count value then after called with a 0 count value so the `removeCount` field check in `ResultSlot#CheckTakeAchievements(ItemStack)` was may be at first here to prevent the increase of the statistic twice idk. But since the `ItemStack#getCount()` is already set to 0 at the second call even without the `removeCount` check it should not increase the Statistic now. I didn't want to dive more into the stuff cause It could need a bigger refactor of the entire crafting stuff for the same result without known what Mojang planned for this part with all new job blocks.

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
elmital: Like I explain above similar methods are called twice so here the recipes awarding is already called later in the StoneCutter#quickMoveStack(Player, int) method.

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
elmital: You right I was pretty sure I've checked to avoid that but re-testing today with my changes the statistic is increased. I don't know how you prefer I change that. I could maybe simply add a check like this :
``` java 
@Override
protected void onQuickCraft(ItemStack stack, int amount) {
    if (ItemCombinerMenu.this instanceof SimthingMenu) stack.onCraftedBy(player.level(), player, amount);
}
```

Or maybe add a `onQuickCraft` method in `ItemCombinerMenu` and override it in SmithingMenu?

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
Machine-Maker: Yep, I see that now.

patches/server/0988-fix-MC-65198-ITEM_CRAFTED-statistic-isn-t-always-inc.patch
Machine-Maker: I think just doing the instanceof check in the existing method is fine.

elmital:
Requested changes applied to fix the anvil issue. PR rebased and move the MC bug fix to the existant patch file `0873-Fix-a-bunch-of-vanilla-bugs.patch`

PR #9493, opened 2023-07-15
Don't support hex in book title and author

Machine-Maker:
Book author and title are stored as strings, and not parsed as components. Therefore the jank hex legacy format is not supported, and that format should not be written or read. This creates a separate legacy serializer for use in the book meta impl.


electronicboy (approved):

Owen1212055 (approved):
Screenshot 2023-07-22 at 8 43 47 PM

PR #9509, opened 2023-07-18
Fix unbreakable flint and steel being consumed when igniting creepers

### LINKED ISSUE #9508: unbreakable Flint and Steel bug
### Expected behavior

in normal,if i use unbreakable Flint and Steel to ignite the crepper,the Steel will just be fine.

### Observed/Actual behavior

but now if i use unbreakable Flint and Steel to ignite the crepper,the Steel will just disappear
( if i use normal Flint and Steel won,t have this problem)

### Steps/models to reproduce

bug vidoe here
https://cdn.discordapp.com/attachments/289587909051416579/1130695367982321774/2023_07_18_10_58_08_473.mp4

### Plugin and Datapack List

[10:58:58 INFO]: 12345cat lost connection: Disconnected
[10:58:58 INFO]: 12345cat left the game
> plugins
[11:03:42 INFO]: Server Plugins (0):
> datapack list
[11:03:48 INFO]: There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
[11:03:48 INFO]: There are no more data packs available

### Paper version

> version
[11:04:02 INFO]: This server is running Paper version git-Paper-83 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: c793bd9)
Previous version: git-Paper-43 (MC: 1.20.1)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/9508

----

Mojang's (and originally ours since we fixed it before) fix for https://bugs.mojang.com/browse/MC-257875 didn't take into account the `Unbreakable` nbt tag for damageable items. We can bypass that by just directly checking if the itemstack's item type has durability. If it does not, we shrink by 1. If it does, you call the ItemStack#hurtAndBreak method which has it's own check for damageable items.

### DIFF
--- a/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
@@ -48,6 +48,9 @@ https://bugs.mojang.com/browse/MC-123848
   by: BillyGalbreath <blake.galbreath@gmail.com>
   Fixes item frames dropping items above when pointing down
 
+https://bugs.mojang.com/browse/MC-264285
+  Fix unbreakable flint and steel being consumed when igniting creepers
+
 Co-authored-by: William Blake Galbreath <blake.galbreath@gmail.com>
 
 diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

lynxplay (approved):

patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
```diff
@@ -48,6 +48,9 @@ https://bugs.mojang.com/browse/MC-123848
   by: BillyGalbreath <blake.galbreath@gmail.com>
   Fixes item frames dropping items above when pointing down
 
+https://bugs.mojang.com/browse/MC-264274
```
Lulu13022002: This is not the right fix for this bug

lynxplay: True lol, machine copying the wrong mojira id smh

patches/server/0873-Fix-a-bunch-of-vanilla-bugs.patch
```diff
@@ -248,6 +251,19 @@ index a86472cce8e8fcde16d761842fe443a619f6e305..b42c060a5d8d68b5773a8a5e38c59707
...
+diff --git a/src/main/java/net/minecraft/world/entity/monster/Creeper.java b/src/main/java/net/minecraft/world/entity/monster/Creeper.java
+index 7fe90ebc8eced53f72c7f935e40745075f02421b..38282337bf4adc296f84a91998603613ebbe6231 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/Creeper.java
++++ b/src/main/java/net/minecraft/world/entity/monster/Creeper.java
+@@ -248,7 +248,7 @@ public class Creeper extends Monster implements PowerableMob {
+             this.level().playSound(player, this.getX(), this.getY(), this.getZ(), soundeffect, this.getSoundSource(), 1.0F, this.random.nextFloat() * 0.4F + 0.8F);
+             if (!this.level().isClientSide) {
+                 this.ignite();
+-                if (!itemstack.isDamageableItem()) {
++                if (itemstack.getItem().getMaxDamage() == 0) { // Paper - fix MC-264274, only shrink the stack if the item type actually has no durability
```
Lulu13022002: This is still not the right fix for this bug in the comment

kennytv (approved):

Issue #9438, opened 2023-07-02
Connections via Server Address and Velocity Server


No-Eul:
### Is your feature request related to a problem?

I wanted to support both connection through my domain as well as the Velocity lobby server using the plugin that interact with the player NPCs. So I have personally implemented this feature and I want it to be officially added.

### Describe the solution you'd like.

* Add `proxies.velocity.allow-direct-connection` config in `config/paper-global.yml` file
  * This is a boolean value, the default is `false`, which a server owner can enable if necessary.
  * If it is `true`, the clients can connect to the Paper server without connecting to the Velocity server.

### Total Configuration Specification
```yml
# config/paper-global.yml
proxies:
  velocity:
    allow-direct-connection: false
```

### Describe alternatives you've considered.
Before making this suggestion, I implemented and tried this feature myself by modifing server or by making [the mod](https://modrinth.com/mod/fabroxy). like this:

### Before Implementation
```java
// Nested Class GlobalConfiguration.Proxies.Velocity in io/papermc/paper/configuration/GlobalConfiguration.java
public class Velocity extends ConfigurationPart {
    // ...
}
```
```java
// void handleCustomQueryPacket(ServerboundCustomQueryPacket) in net/minecraft/server/network/ServerLoginPacketListenerImpl.java
// Paper start - Velocity support
if (io.papermc.paper.configuration.GlobalConfiguration.get().proxies.velocity.enabled && packet.getTransactionId() == this.velocityLoginMessageId) {
    net.minecraft.network.FriendlyByteBuf buf = packet.getData();
    if (buf == null) {
        this.disconnect("This server requires you to connect with Velocity.");
        return;
    }
    // ...
}
// ...
```

### After Implementation
```java
// Nested Class GlobalConfiguration.Proxies.Velocity in io/papermc/paper/configuration/GlobalConfiguration.java
public class Velocity extends ConfigurationPart {
    // ...
    public boolean allowDirectConnection;
}
```
```java
// void handleCustomQueryPacket(ServerboundCustomQueryPacket) in net/minecraft/server/network/ServerLoginPacketListenerImpl.java
// Paper start - Velocity support
if (io.papermc.paper.configuration.GlobalConfiguration.get().proxies.velocity.enabled && packet.getTransactionId() == this.velocityLoginMessageId) {
    net.minecraft.network.FriendlyByteBuf buf = packet.getData();
    if (buf == null) {
[...]

PR #9511, opened 2023-07-18
Add allowDirectConnection to Velocity proxy config

No-Eul:
Each Paper server can enable or disable this feature, which allows clients to connect to the server without having to connect to the Velocity proxy server.

The content of pull request is same to [this](https://github.com/PaperMC/Paper/issues/9438).

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -440,10 +440,10 @@ index 0000000000000000000000000000000000000000..9ef6712c70fcd8912a79f3f61e351aac
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..58cdb70c0b9de45ca28811416d871099eefec2ce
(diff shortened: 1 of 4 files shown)


Malfrador:
What's the use case of this? Why would anyone want to do this instead of just using velocitys forced-hosts feature?

You are not taking care of any authentication here. If velocity is used, the server will be running in offline mode. With your option anyone could connect to the server without being authenticated.

DaJokni:
why not just use direct connections in velocity?

No-Eul:
In my case, my friend's Velocity server and my Paper server are not on the same network, so forced-host doesn't work as well as I thought.

...or I may not have enough understanding about forced-host well.

emilyy-dev:
Hi, thank you for wanting to contribute to Paper!

Unfortunately, this is not something we want in Paper as it is, it does not take care of authentication whatsoever and it's not in our interest to have the server manage some sort of hybrid, broken, and exploitable mix between online and offline mode.

It is greatly appreciated to discuss and socialize changes before making/submitting them, that be via its own issue or in our discord. That being said, it was mentioned that the forced hosts setting in the proxy should solve your mentioned problem in #9438. If you have questions about feel free to reach out to us in our [discord server](https://discord.gg/papermc) or in the [Paper forums](https://forums.papermc.io/).

PR #9517, opened 2023-07-21
Fix endgateway teleportation offset

Lulu13022002:
The endgateway teleport the entity normally at the center of the found block in the horizontal axis (X+0.5 Y+0 Z+0.5) however on Spigot/Paper the offset is at the center of the block for both axis (X+0.5 Y+0.5 Z+0.5). On Paper it's even worst than Spigot since it affect all entities and so the teleportation is done in mid-air unlike vanilla.

### DIFF
--- a/patches/server/0201-Implement-EntityTeleportEndGatewayEvent.patch
+++ b/patches/server/0201-Implement-EntityTeleportEndGatewayEvent.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Implement EntityTeleportEndGatewayEvent
 
 
 diff --git a/src/main/java/net/minecraft/world/level/block/entity/TheEndGatewayBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/TheEndGatewayBlockEntity.java
-index 020314dded2c6305bd5e9013b108bc74be25ccc1..d3e2a78f4e562fac7f42687629a4807aec85037f 100644
+index 020314dded2c6305bd5e9013b108bc74be25ccc1..85914124014b4e6f0a561cf560918af68682b6f5 100644
 --- a/src/main/java/net/minecraft/world/level/block/entity/TheEndGatewayBlockEntity.java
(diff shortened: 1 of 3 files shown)


Machine-Maker (approved):

patches/server/0201-Implement-EntityTeleportEndGatewayEvent.patch
```diff
@@ -5,27 +5,27 @@ Subject: [PATCH] Implement EntityTeleportEndGatewayEvent
...
 +                com.destroystokyo.paper.event.entity.EntityTeleportEndGatewayEvent event = new com.destroystokyo.paper.event.entity.EntityTeleportEndGatewayEvent(bukkitEntity, bukkitEntity.getLocation(), location, new org.bukkit.craftbukkit.block.CraftEndGateway(world.getWorld(), blockEntity));
 +                if (!event.callEvent()) {
 +                    return;
 +                }
 +                // Paper end
  
                  entity1.setPortalCooldown();
--                entity1.teleportToWithTicket((double) blockposition1.getX() + 0.5D, (double) blockposition1.getY(), (double) blockposition1.getZ() + 0.5D);
+-                entity1.teleportToWithTicket((double) blockposition1.getX() + 0.5D, (double) blockposition1.getY(), (double) blockposition1.getZ() + 0.5D); // Paper - diff on change
 +                entity1.teleportToWithTicket(event.getTo().getX(), event.getTo().getY(), event.getTo().getZ()); // Paper
```
Machine-Maker, line 29: Just a comment on something I noticed later, I wonder if it'd be a nicer diff to reassign the `blockposition1` local var with the location from the event, that way this diff could be avoided. Not required for merge, but it could better to avoid the replacement diff (so we can keep the diff on change comment actually valid)

Lulu13022002: Yes that would be convenient but then plugin can't change the offset anymore. The possible way would be to compare if the destination location has changed and keep that line if it doesn't change

kennytv (approved):

Issue #6658, opened 2021-09-24
EntityTeleportEndGatewayEvent not work in passengers


Doc94:
### Expected behavior

When a entity mounted in another entity enter to a END_GATEWAY and is cancel using EntityTeleportEndGatewayEvent the entity  mounted cant be teleported.

### Observed/Actual behavior

When a player in a boat enter to the END_GATEWAY the boat not teleport because the EntityTeleportEndGatewayEvent is cancelled but the passengers (player) still can teleport.

### Steps/models to reproduce

`
@EventHandler
    public void portalEnter(EntityTeleportEndGatewayEvent event) {
       event.setCancelled(true);
    }
`

- Enter to the END
- Destroy the BEDROCK
- Put a boat
- Navigate to the portal 

### Plugin list

N/A

### Paper version

This server is running Paper version git-Paper-276 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 1707c10)
You are 2 version(s) behind
Previous version: git-Paper-266 (MC: 1.17.1)

### Agreements

- [X] I am running the latest version of Paper available from https://papermc.io/downloads.
- [X] I have searched for and ensured there isn't already an open issue regarding this.
- [X] My version of Minecraft is supported by Paper.

### Other

The event works like paralel to PlayerTeleportEvent not?

because this issue happen with enderpearls too, then the temporal fix is use the two events EntityTeleportEndGatewayEvent and PlayerTeleportEvent because PlayerPortalEvent is not valid for this cases.

lynxplay:
I can validate that this is the behavior on d0a2193d4a93ecc7dd69499cd1aa86c94d0f51f4. This is due to the fact that end gateways have an internal cooldown between each teleportation that is started even if the first entity that was to be teleported got its event cancelled.

Specifically `TheEndGatewayBlockEntity.java` line `112` breaks out of the loop over all entities after the first potential candidate, preventing any further logic from executing in regards to teleportation, whether the teleport was successful or not.

// edit[0]: The reason why I moved this to accepted was the fact that an entity in the same portal as another (at least in my opinion) should not block the other entity from teleporting. The passenger use-case I am not as sure about and should be subject to further discussion on here.

Machine-Maker:
@lynxplay where is the internal cooldown applied if the teleports fail? I see this 
```java
            if (blockEntity.age % 2400L == 0L) {
[...]

PR #9522, opened 2023-07-22
Allow passenger retaining teleport via end gateway

lynxplay:
Previously paper disabled the abilities for entities to teleport via end gateways if they were being used as a vehicle.

While the behaviour generally worked fine for entities riding other entities, players would quickly end up in an invalid state, not seeing their passenger anymore.

This commit removes the paper introduced limitation by now properly updating the passengers location of a player when they are teleporting through an end gateway.

### DIFF
--- a/patches/server/0403-Ensure-safe-gateway-teleport.patch
+++ b/patches/server/0403-Ensure-safe-gateway-teleport.patch
@@ -3,24 +3,18 @@ From: kickash32 <kickash32@gmail.com>
 Date: Fri, 15 May 2020 01:10:03 -0400
 Subject: [PATCH] Ensure safe gateway teleport
 
+== AT ==
+public net.minecraft.world.entity.Entity teleportPassengers()V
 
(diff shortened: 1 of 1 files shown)


kennytv (approved):

Doc94:
not sure but can be related to this?
https://github.com/PaperMC/Paper/issues/6658

lynxplay:
Partially, the other issue in my mind at least mainly relates to a cancelled EndgatewayTP event preventing any other entity from trying.

patches/server/0403-Ensure-safe-gateway-teleport.patch
```diff
@@ -3,24 +3,18 @@ From: kickash32 <kickash32@gmail.com>
 Date: Fri, 15 May 2020 01:10:03 -0400
 Subject: [PATCH] Ensure safe gateway teleport
```
Machine-Maker, line 4: This patch needs to be renamed to something more correct now. And just change the author too. Its really a complete patch replacement rather than an edit.

Machine-Maker (changes requested):

lynxplay:
Converted to draft for now, I do not have the time to look into the faulty behaviour of minecarts when steered through an endgateway.

PR #9525, opened 2023-07-23
[ci skip] Clarify viewers set mutability in chat events

### LINKED ISSUE #8461: AbstractChatEvent viewers should be mutable in all cases

When writing plugins that contain e.g. a team chat the obvious solution would proably be to modify the viewers in AbstractChatEvent with for example removeIf. But when looking at the viewers() method in the event the javadocs say the following: 

"Listeners should be aware that modifying the list may throw UnsupportedOperationException if the event caller provides an unmodifiable set."

But you could just cancel the event and send the message to the appropriete players with sendMessage right? Sadly this is not really a good solution since the message will be displayed as a system messsage to the players. And for other plugins monitoring the outcome of the event it will be unclear what players will recieve the message.

I fully understand that there isn't a good method to check if a set or list is mutable. But with how the javadoc is written its really unclear if paper itself calls it with an unmutable set or if its just third party plugins and it does not encurrage developers to provide a mutable set because it could break plugins when they dont.

I think the javadocs should state that a mutable set should be provided because it can cause incompatability if an immutable one is given. And the docs should also say wheather paper itself provides immutable sets in specific cases.

Catching the exception and using sendMessage in case of faliure. But in my opinion thats not really a good way to solve the issue

### Other

_No response_

Machine-Maker:
Closes https://github.com/PaperMC/Paper/issues/8461

Constructing and calling API events in a plugin isn't API. This is because the constructors for those events are not considered API. No where else in the API as far as I can tell adds a caveat in the getter for a collection that "if a plugin calls it, it *might* not be mutable". Every time the server calls it, it is mutable and that's enough to say its mutable in the javadoc.

### DIFF
--- a/patches/api/0005-Adventure.patch
+++ b/patches/api/0005-Adventure.patch
@@ -200,10 +200,10 @@ index 0000000000000000000000000000000000000000..2ad76b1751ba707f7ae0d283aa1cbaf6
 +}
 diff --git a/src/main/java/io/papermc/paper/event/player/AbstractChatEvent.java b/src/main/java/io/papermc/paper/event/player/AbstractChatEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..e7e13011c76285681ad420e6f356f6b83045d31a
+index 0000000000000000000000000000000000000000..ed8e885f226b02b9875b23ae2294a9056d2e8b29
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/player/AbstractChatEvent.java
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
I think that we should make it a habit to mark event constructors as internal TBH....

Machine-Maker:
> I think that we should make it a habit to mark event constructors as internal TBH....

Yeah, I've been doing that as I come across them while making a patch that is near or affects them.

kennytv (approved):

PR #9530, opened 2023-07-24
Add InventoryCloseEvent safety doc comment, similar to InventoryClickEvent

Gameoholic:
When invoking `HumanEntity.openInventory()` by an EventHandler for `InventoryCloseEvent`, clicking on inventory menus after invoking it does not fire the event `InventoryClickEvent`.
Scheduling a task for it fixed the issue.

`InventoryClickEvent` has the following doc:
```
... 
The following should never be invoked by an EventHandler for InventoryClickEvent using the HumanEntity or InventoryView associated with this event:
HumanEntity.openInventory(Inventory)
...
```
However, the comment for `InventoryCloseEvent` doesn't state that, even though it should.


kezz:
Should probably be merged into https://github.com/PaperMC/Paper/blob/master/patches/api/0055-Fix-upstream-javadocs.patch

PR #9533, opened 2023-07-25
Fix inventorycloseevent javadocs

Gameoholic:
Merge https://github.com/PaperMC/Paper/pull/9530


patches/api/0055-Fix-upstream-javadocs.patch
```diff
@@ -692,6 +691,37 @@ index 79797a2be7fb139d528116d34d13e51d39b96e56..0921484e921dbd200725b9298f655720
...
++ + * The following should never be invoked by an EventHandler for
++ + * InventoryCloseEvent using the HumanEntity or InventoryView associated with
++ + * this event:
++ + * <ul>
++ + * <li>{@link HumanEntity#closeInventory()}
++ + * <li>{@link HumanEntity#openInventory(Inventory)}
++ + * <li>{@link HumanEntity#openWorkbench(Location, boolean)}
++ + * <li>{@link HumanEntity#openEnchanting(Location, boolean)}
++ + * <li>{@link InventoryView#close()}
++ + * </ul>
```
Machine-Maker: I would change the wording on this to make it open ended. Currently it sounds like this list of methods is comprehensive and that any other method is fine to call when there are more methods that open inventory views. I would say something like "methods that change the view a player is looking at". And then you can list these as examples, not the full list.

Machine-Maker (changes requested):
Also, idk why the patch title changed, it should have spaces in it instead of hyphens. The hyphens replace invalid chars (like spaces) in the patch file name.

patches/api/0055-Fix-upstream-javadocs.patch
Gameoholic: Yeah to be fair it does say "Also be aware that this is not an exhaustive list, and other methods could potentially create issues as well." 
I just copied what it said for InventoryClickEvent.
What do you think of:
```
This event is called when a player closes an inventory.
Because InventoryCloseEvent occurs within a modification of the Inventory, not all Inventory related methods are safe to use.
Methods that change the view a player is looking at should never be invoked by an EventHandler for InventoryCloseEvent using the HumanEntity or InventoryView associated with this event.
Examples of these methods include:

- HumanEntity.closeInventory()
- HumanEntity.openInventory(Inventory)
- HumanEntity.openWorkbench(Location, boolean)
- HumanEntity.openEnchanting(Location, boolean)
- InventoryView.close()

To invoke one of these methods, schedule a task using BukkitScheduler.runTask(Plugin, Runnable), which will run the task on the next tick. Also be aware that this is not an exhaustive list, and other methods could potentially create issues as well.
```
I should also add this, right? (taken from InventoryClickEvent)
```
Assuming the EntityHuman associated with this event is an instance of a Player, manipulating the MaxStackSize or contents of an Inventory will require an Invocation of Player.updateInventory().
Modifications to slots that are modified by the results of this InventoryClickEvent can be overwritten. To change these slots, this event should be cancelled and all desired changes to the inventory applied. Alternatively, scheduling a task using BukkitScheduler.runTask(Plugin, Runnable), which would execute the task on the next tick, would work as well.
```

I've also just noticed that this javadoc is missing in other inventory related events so I'll test them tomorrow, and add them if those methods break for them as well.

I'll fix the patch title.

Machine-Maker: I think the first javadoc sounds good to me.

As for adding that second block from InventoryClickEvent, I actually don't know if that is true for the close event, that would require testing which you could do, or just ignore and just not add that now.

Gameoholic:
I've fixed the patch title, and updated the javadocs for both InventoryClickEvent and InventoryCloseEvent.

Gameoholic (commented):
I've fixed the patch title, and updated the javadocs for both InventoryClickEvent and InventoryCloseEvent.
(sorry for the double comment, happened on accident)

Machine-Maker (approved):

PR #9535, opened 2023-07-26
Add and implement PlayerFailMoveEvent

Moulberry:
Event runs when the client attempts to move, but is prevented from doing so by the server

The movement can be allowed by calling "setAllowed(true)"
Warning messages can be prevented by calling "setLogWarning(false)"

Event can be used to:
- Prevent lagback when interacting with certain game elements like launchers, double jump, etc.
- Disable movement checks for admins
- Prevent 'moved too quickly' and 'moved wrongly' messages from being logged

Event can also be implemented for vehicle movement, but I've skipped that in this PR.

### DIFF
--- a/patches/server/0994-Implement-PlayerFailMoveEvent.patch
+++ b/patches/server/0994-Implement-PlayerFailMoveEvent.patch
@@ -5,16 +5,27 @@ Subject: [PATCH] Implement PlayerFailMoveEvent
 
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index 316740b2ba4c85828f544249c8cdd6fa1b525d3f..fa6acce416d3c5fb821dc7da8e37c0adb6738218 100644
+index 316740b2ba4c85828f544249c8cdd6fa1b525d3f..f14f498203169797434a113c73a0dd2660d7cf21 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -1404,8 +1404,8 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                     double d0 = ServerGamePacketListenerImpl.clampHorizontal(packet.getX(this.player.getX())); final double toX = d0; // Paper - OBFHELPER
+                     double d1 = ServerGamePacketListenerImpl.clampVertical(packet.getY(this.player.getY())); final double toY = d1;
+                     double d2 = ServerGamePacketListenerImpl.clampHorizontal(packet.getZ(this.player.getZ())); final double toZ = d2; // Paper - OBFHELPER
+-                    float f = Mth.wrapDegrees(packet.getYRot(this.player.getYRot()));
+-                    float f1 = Mth.wrapDegrees(packet.getXRot(this.player.getXRot()));
++                    float f = Mth.wrapDegrees(packet.getYRot(this.player.getYRot())); final float toYaw = f; // Paper - OBFHELPER
++                    float f1 = Mth.wrapDegrees(packet.getXRot(this.player.getXRot())); final float toPitch = f1; // Paper - OBFHELPER
+ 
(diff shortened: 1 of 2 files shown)


Camotoy:
I could see this having use for Geyser wherever we can't figure out, or implement a reasonable fix, for incorrect movement.

It does seem like you could reduce diff by having a function that is called in the three spots you replace? Each codeblock seems to be doing, more or less, the same thing.

patches/server/0994-Implement-PlayerFailMoveEvent.patch
```diff
@@ -0,0 +1,102 @@
...
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -1471,8 +1471,14 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                             }
+                             // Paper start - Prevent moving into unloaded chunks
+                             if (this.player.level().paperConfig().chunks.preventMovingIntoUnloadedChunks && (this.player.getX() != toX || this.player.getZ() != toZ) && !worldserver.areChunksLoadedForMove(this.player.getBoundingBox().expandTowards(new Vec3(toX, toY, toZ).subtract(this.player.position())))) {
+-                                this.internalTeleport(this.player.getX(), this.player.getY(), this.player.getZ(), this.player.getYRot(), this.player.getXRot(), Collections.emptySet());
+-                                return;
++                                // Paper start - Add fail move event
++                                var event = fireFailMove(io.papermc.paper.event.player.PlayerFailMoveEvent.FailReason.MOVED_INTO_UNLOADED_CHUNK,
```
kennytv: var bad

patches/server/0994-Implement-PlayerFailMoveEvent.patch
```diff
@@ -0,0 +1,102 @@
...
++                                    toX, toY, toZ, f, f1, false);
++                                if (!event.isAllowed()) {
++                                    this.internalTeleport(this.player.getX(), this.player.getY(), this.player.getZ(), this.player.getYRot(), this.player.getXRot(), Collections.emptySet());
```
kennytv: If you dont change the line, as dumb as it'll look, keep the original indent of those particular lines so it's clear it is unchanged/no diff

kennytv (changes requested):

kennytv (approved):

patches/server/0994-Implement-PlayerFailMoveEvent.patch
```diff
@@ -0,0 +1,93 @@
...
+index 316740b2ba4c85828f544249c8cdd6fa1b525d3f..fa6acce416d3c5fb821dc7da8e37c0adb6738218 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -1471,8 +1471,14 @@ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, Tic
+                             }
+                             // Paper start - Prevent moving into unloaded chunks
+                             if (this.player.level().paperConfig().chunks.preventMovingIntoUnloadedChunks && (this.player.getX() != toX || this.player.getZ() != toZ) && !worldserver.areChunksLoadedForMove(this.player.getBoundingBox().expandTowards(new Vec3(toX, toY, toZ).subtract(this.player.position())))) {
++                                // Paper start - Add fail move event
++                                io.papermc.paper.event.player.PlayerFailMoveEvent event = fireFailMove(io.papermc.paper.event.player.PlayerFailMoveEvent.FailReason.MOVED_INTO_UNLOADED_CHUNK,
++                                    toX, toY, toZ, f, f1, false);
```
codebycam: Can you please add obf helpers for `f` & `f1`, same below

codebycam (changes requested):

patches/server/0994-Implement-PlayerFailMoveEvent.patch
codebycam: After this change it'll be ready for merge :-D

codebycam (approved):
Thanks!

CodeByCam:
@kennytv Please merge :-D

PR #9541, opened 2023-07-28
Use correct seed on api world load

Floppy012:
When loading an existing world using a `WorldCreator`, there is potential for two different seeds to be used. The seed used by the generator is the one loaded from the `level.dat` while the seed that gets provided to the `BiomeManager` (aka. `biomeZoomSeed`) is the one that the `WorldCreator` instance provides (which is random when not explicitly provided).

With this PR the seed from `WorldOptions` will be taken which will be the seed loaded from the `level.dat` in case the world already exists. In case a new world is being created the seed will be the one from `WorldCreator` as that is used to create `WorldOptions` in that case.

The new behavior would be the same as in `MinecraftServer#loadWorld0`

## Biome Changes

The impact of this PR on existing worlds is minor. Since the seed only seems responsible for biome borders. That is probably also the reason why this bug hasn't been noticed yet. For illustration, I've extracted a 1000x1000 area of the biomes as images:

| / | World Create | World load | Diff | 
|--|:------------------:|:----------------:|:------:|
|**Without fix** | worldCreate | before | create-before-diff
|**With fix** | worldCreate | after | create-after-diff

### DIFF
--- a/patches/server/1000-Use-correct-seed-on-api-world-load.patch
+++ b/patches/server/1000-Use-correct-seed-on-api-world-load.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Florian Schmidt <florian@f012.dev>
+Date: Fri, 28 Jul 2023 14:14:35 +0200
+Subject: [PATCH] Use correct seed on api world load
+
+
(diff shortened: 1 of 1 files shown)


kennytv (approved):

MiniDigger (approved):

Machine-Maker (approved):

Machine-Maker:
1000 patches

PR #9552, opened 2023-07-30
Fix rotation for spawning display entities

### LINKED ISSUE #9418: Pitch/Yaw ignored when spawning display entities
### Expected behavior

For pitch and yaw to be used when spawning, as with other entities.

### Observed/Actual behavior

When spawning a display entity, pitch and yaw seem to be ignored and always become `0`.

### Steps/models to reproduce

```java
Location location = player.getEyeLocation();
getLogger().info(location.toString());
Display display = player.getWorld().spawn(location, BlockDisplay.class);
getLogger().info(display.getLocation().toString());
getLogger().info(String.valueOf(display.getLocation().equals(location)));
```
Running the above code will log the first location with the player's yaw and pitch, then the display's location with `0` yaw and pitch, then show that they're not equal:
```
[23:44:07 INFO]: [PaperTestPlugin] Location{world=CraftWorld{name=world},x=-108.5082338140295,y=101.99942319657657,z=354.9048698050415,pitch=90.0,yaw=162.454}
[23:44:07 INFO]: [PaperTestPlugin] Location{world=CraftWorld{name=world},x=-108.5082338140295,y=101.99942319657657,z=354.9048698050415,pitch=0.0,yaw=0.0}
[23:44:07 INFO]: [PaperTestPlugin] false
```

### Plugin and Datapack List

```
> plugins
[23:54:02 INFO]: Server Plugins (1):
[23:54:02 INFO]: Bukkit Plugins:
[23:54:02 INFO]:  - PaperTestPlugin
```
```
> datapack list
[23:54:21 INFO]: There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
[23:54:21 INFO]: There are no more data packs available
````

### Paper version

```
> version
[00:00:49 INFO]: Checking version, please wait...
[00:00:50 INFO]: This server is running Paper version git-Paper-56 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 2c9c5e4)
Previous version: git-Paper-54 (MC: 1.20.1)
```

### Other

I understand that generally you'd use the built-in transformation and not the base yaw/pitch, but I believe there's still a benefit to fixing this if possible (e.g. custom entity systems with generic logic and whatnot).
Especially seeing as this works fine in vanilla commands: `/summon block_display ~ ~ ~ {Rotation:[50F,50F],block_state:{Name:"minecraft:grass_block"}}`

### PR DESCRIPTION
**Pull request by Gameoholic**
Fixes https://github.com/PaperMC/Paper/issues/9418

### CODE REVIEW

**Comment by Owen1212055** [comment]
So is this a blanket fix that makes all entities respect the yaw/pitch on spawn?

**Comment by Gameoholic** [comment]
> It may be better to have a blanket fix that makes all entities respect the yaw/pitch on spawn?

Right so I thought about that, but I'm not sure how to go about it.. [ You've said before](https://github.com/PaperMC/Paper/issues/9418#issuecomment-1612247903) that the spawning logic should be cleaned up, so besides just making all entities respect yaw/pitch (by making sure entity.moveTo() is executed regardless of the entity type) is there anything else I could fix / clean up?

**Comment by Gameoholic** [comment]
What I did for now is just move the entity before returning it:
```java
        if (entity != null) {
            entity.moveTo(x, y, z, yaw, pitch); // Paper - Set Entity Rotation
            return entity;
        }
```
However, looking at the code for spawning entities, sometimes it's `Entity#setPos(x, y, z)`,  `Entity#moveTo(x, y, z, yaw, pitch)` and `Entity#entity.absMoveTo(x, y, z, yaw, pitch)`

My code essentially overrides these statements, but I've tested spawning different entities and I see no issues? Why do these statements even exist when you can just do `Entity#moveTo`, which respects the yaw/pitch?

Machine-Maker (changes requested):
There is a lot of extra logic in moveTo that isn't in setPos. I think instead, you should just change the pitch and yaw individually with setYRot and setXRot.

An example of the logic in moveTo, is to record the old position to run any logic around something actually moving.

patches/server/0994-Fix-rotation-when-spawning-entities.patch
```diff
@@ -0,0 +1,40 @@
...
+             }
+-
+-            if (entity != null) {
+-                entity.setPos(x, y, z);
+-            }
+         }
+ 
+         if (entity != null) {
++            // Paper Start - Update position and rotation
++            entity.setPos(x, y, z);
```
Machine-Maker: Don't move the setPos here. Each entity already has it's own logic for settings its position, its at best redundant, and at worst breaking whatever logic behavior might happen above to set the position.

Machine-Maker (dismissed):
Are we sure that all entities should just have their yaw and pitch set at the end? No clue what that does for hanging entities for example.

patches/server/0994-Fix-rotation-when-spawning-entities.patch
Owen1212055: I am curious, is there any reason why we shouldn't just copy ofver the logic used by /summon?
They do this:
```java
                Entity entity1 = EntityType.loadEntityRecursive(nbttagcompound1, worldserver, (loadedEntity) -> { // Paper - remap fix
                loadedEntity.moveTo(pos.x, pos.y, pos.z, loadedEntity.getYRot(), loadedEntity.getXRot()); // Paper - remap fix
                return loadedEntity; // Paper - remap fix
                });
 ```

Machine-Maker: I mean I thought it was because there's a ton of extra logic in moveTo that isn't required for setting an initial position.

Gameoholic: I did that because Owen suggested to make a fix that makes all entities respect yaw/pitch, but you're right. In that case, can I revert back to my original commit, which just fixed the rotation for display entities?

Owen1212055: Probably. It's out of scope of this PR, can be a future thing to cleanup.

Warriorrrr (commented):
`setRot(yaw, pitch)` would fit best in my opinion, only a single line and does some sanity checks for NaN.

Warriorrrr (approved):

kennytv (approved):

PR #9554, opened 2023-07-30
Make the cursor limit option apply to banners

### LINKED ISSUE #9553: Clinet lag exploit using maps
### Expected behavior

  item-frame-cursor-limit works correctly

### Observed/Actual behavior

I have set item-frame-cursor-limit to 10, however maps can still be created with which the client hangs. Players do this with banners

### Steps/models to reproduce

Just put up a lot of banners and create a map
image
image

### Plugin and Datapack List

None

### Paper version

[23:45:00 INFO]: This server is running Paper version git-Paper-100 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: f402f89)

### Other

_No response_

Warriorrrr:
This PR makes the decoration limit for banners use the config's cursor limit instead of the vanilla limit of 256

Closes #9553


MiniDigger (approved):

MiniDigger:
I wonder if there could be a usecase for have the counts separated but good enough for now

lynxplay (approved):

PR #9558, opened 2023-07-31
Add Block#getFluid method

viciscat:
This PR adds the `getFluid()` method to the `Block` interface and implements it in `CraftBlock`. I also added a similar method to `getBukkitMaterial`, `getBukkitFluid`, in `BlockState`
It does pretty much what it says, returns the `org.bukkit.Fluid` present in the block. For example a waterlogged stair will return a `Fluid.WATER`, same for kelp and uh same for water. Did that make sense?


kennytv:
Thank you for your PR. Make sure you don't modify any imports; use fully qualified imports instead. I also don't think caching the fluid here is necessary

viciscat:
Thanks for the comment!
Will change the imports. And i cached the fluid to mimic the cachedMaterial value that was already present. Other than that I will probably close this PR because I found out that MachineMaker made a draft PR for a FluidState API.

PR #9569, opened 2023-08-04
Add Adventure components title and originalTitle in InventoryView

LeonidMem:
Honestly, I am not entirely sure if I changed the method CraftInventoryView.sendInventoryTitleChange(InventoryView, String) correctly for future work, because I decided to move from duplicated code, so, please, let me know how it is better to resolve such things.

### DIFF
--- a/patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
+++ b/patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
@@ -1,0 +1,157 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: LeonidM <wildhs.botm@gmail.com>
+Date: Fri, 4 Aug 2023 16:05:08 +0300
+Subject: [PATCH] Add Adventure components title and originalTitle in
+ InventoryView
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
+index 081e6c7ce4a0185ba75e48bbb581fd531cdb10a8..7d61c16ea889f83e84d80d37fcc5dc16ace6de15 100644
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftContainer.java
+@@ -50,6 +50,10 @@ public class CraftContainer extends AbstractContainerMenu {
+         this(new InventoryView() {
+ 
+             private final String originalTitle = inventory instanceof CraftInventoryCustom ? ((CraftInventoryCustom) inventory).getTitle() : inventory.getType().getDefaultTitle(); // Paper
++            // Paper start
++            private final net.kyori.adventure.text.Component originalTitleComponent = inventory instanceof CraftInventoryCustom ? ((CraftInventoryCustom) inventory).title() : inventory.getType().defaultTitle(); // Paper
++            private net.kyori.adventure.text.Component titleComponent = this.originalTitleComponent;
++            // Paper end
+             private String title = this.originalTitle;
+ 
+             @Override
+@@ -75,24 +79,39 @@ public class CraftContainer extends AbstractContainerMenu {
+             // Paper start
+             @Override
+             public net.kyori.adventure.text.Component title() {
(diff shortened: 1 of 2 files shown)


patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
```diff
@@ -0,0 +1,157 @@
...
++            }
++
++            @Override
++            public net.kyori.adventure.text.Component originalTitle() {
++                return this.originalTitleComponent;
+             }
+             // Paper end
+ 
+             @Override
++            @Deprecated // Paper
```
lynxplay, line 44: No need to annotate internal methods for deprecation.
(Repeats in a few spaces in the patch)

patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
```diff
@@ -0,0 +1,157 @@
...
+             public String getOriginalTitle() {
+                 return originalTitle;
+             }
+ 
+             @Override
++            @Deprecated // Paper
+             public void setTitle(String title) {
+                 CraftInventoryView.sendInventoryTitleChange(this, title);
```
lynxplay, line 58: Wouldn't it be smart to just invoke title() here with the result of the legacy deserialisation ?

patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
```diff
@@ -0,0 +1,157 @@
...
+     public String getOriginalTitle() {
+         return this.originalTitle;
+     }
+ 
+     @Override
++    @Deprecated // Paper
+     public void setTitle(String title) {
+         CraftInventoryView.sendInventoryTitleChange(this, title);
+         this.title = title;
++        this.titleComponent = net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(this.title); // Paper
```
lynxplay, line 127: Same here, probably best to just delegate this to the paper method to reduce duplicates.

patches/server/0994-Add-Adventure-components-title-and-originalTitle-in-.patch
```diff
@@ -0,0 +1,157 @@
...
+     }
+ 
+     public boolean isInTop(int rawSlot) {
+@@ -104,7 +127,16 @@ public class CraftInventoryView extends InventoryView {
+         return this.container;
+     }
+ 
++    @Deprecated // Paper
+     public static void sendInventoryTitleChange(InventoryView view, String title) {
++        // Paper start
```
lynxplay, line 137: This entire block can be made a bit pretty to look more like this:
```patch
     public static void sendInventoryTitleChange(InventoryView view, String title) {
+    // Paper start
+        Preconditions.checkArgument(title != null, "Title cannot be null");
+        CraftInventoryView.sendInventoryTitleChange(view, net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(title));
+    }
+
+    public static void sendInventoryTitleChange(InventoryView view, net.kyori.adventure.text.Component title) {
+    // Paper end
         Preconditions.checkArgument(view != null, "InventoryView cannot be null");
         Preconditions.checkArgument(title != null, "Title cannot be null");
         Preconditions.checkArgument(view.getPlayer() instanceof Player, "NPCs are not currently supported for this function");
@@ -113,7 +145,8 @@ public class CraftInventoryView extends InventoryView {
         final ServerPlayer entityPlayer = (ServerPlayer) ((CraftHumanEntity) view.getPlayer()).getHandle();
         final int containerId = entityPlayer.containerMenu.containerId;
         final MenuType<?> windowType = CraftContainer.getNotchInventoryType(view.getTopInventory());
-        entityPlayer.connection.send(new ClientboundOpenScreenPacket(containerId, windowType, CraftChatMessage.fromString(title)[0]));
+        entityPlayer.connection.send(new ClientboundOpenScreenPacket(containerId, windowType, io.papermc.paper.adventure.PaperAdventure.asVanilla(title))); // Paper 
         ((Player) view.getPlayer()).updateInventory();
     }

lynxplay (changes requested):
Welcome to paper :tada: grats on your first PR!

A few comments in the patch, beyond that generally, this change should live in the existing Adventure patch instead of being its own patch.
See the CONTRIBUTING.md on how to modify an existing patch.

Machine-Maker:
This is actually a duplicate PR (https://github.com/PaperMC/Paper/pull/9330). Since that one was made first, and they do the exact same thing, I'm going to close this one.

PR #9570, opened 2023-08-04
Only erase allay memory on non-item targets

### LINKED ISSUE #9567: Allay behavior differs significantly from Vanilla server.
### Expected behavior

Action to pick up selected items should be a priority.

### Observed/Actual behavior

Return to the player or note block before all items that Allay can find have been picked up.

### Steps/models to reproduce

There is the video.
https://cdn.discordapp.com/attachments/803652566571024404/1136724078330392676/Minecraft_003_AV1.mp4

### Plugin and Datapack List

> plugins
[03:16:48 INFO]: Server Plugins (0):

> datapack list
[03:17:00 INFO]: There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
[03:17:00 INFO]: There are no more data packs available

### Paper version

> version
[03:17:17 INFO]: Checking version, please wait...
[03:17:19 INFO]: This server is running Paper version git-Paper-100 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: f402f89)

### Other

_No response_

lynxplay:
Spigot incorrectly instanceOf checks the EntityTargetEvent#getTarget against the internal ItemEntity type and removes the nearest wanted item memory if said instanceOf check fails, (which is always the case) causing allays to behave differently as they constantly loose their target item.

This commit fixes the faulty behaviour by instance performing a check against the CraftItem type.

Resolves: #9567

Additionally also resolves a not updated distance variable that was wrongly ported to 1.19.

### DIFF
--- a/patches/server/0707-Remove-streams-for-villager-AI.patch
+++ b/patches/server/0707-Remove-streams-for-villager-AI.patch
@@ -112,7 +112,7 @@ index 731ef21dbbd25d6924717de42f4569a9b5935643..fe3ab3d388f0481fb0db06b7f730f868
      private final boolean isUnsafe; // Paper
  
 diff --git a/src/main/java/net/minecraft/world/entity/ai/sensing/NearestItemSensor.java b/src/main/java/net/minecraft/world/entity/ai/sensing/NearestItemSensor.java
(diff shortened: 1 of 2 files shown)


electronicboy (approved):
Can't test, but, based up on what I saw elsewhere, lgtm

Machine-Maker (approved):

Owen1212055 (approved):

PR #9571, opened 2023-08-04
[1.18.2] Update velocity-native to make it compile

powercasgamer:
"tested" it by compiling and joining which works.

discord conversation: https://discord.com/channels/289587909051416579/555462289851940864/1137005445098242161

Closes https://github.com/PaperMC/paperweight/issues/203


MiniDigger:
I restored the old proxied snapshot versions of the velocity natives. there is no guarantee that they will stick around forever. old versions are unsupported for a reason. people should migrate to supported versions.

PR #9578, opened 2023-08-05
Add-Mob-Spawner-support-to-AffectsSpawning

TreemanKing:
Introduce API for Mob Spawner support for AffectsSpawning to allow different type of spawning types. The reason for this as it would allow for the ability to only prevent natural spawns and keep mob spawns from spawners.

Originally I was just going to create another get/set, however, the code for checkDespawn is quite hard to change without breaking a previous patch hence I opted for editing original API by adding an extra boolean to the get (public void setAffectsSpawning(boolean affects, boolean affectsMobSpawners))

I believe everything should work by common sense but unsure if I've missed or possibly stuffed something up as it's my first time meddling around with this.

### DIFF
--- a/patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
+++ b/patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add Mob Spawner support to AffectsSpawning
 
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index 18aac3da3c88f33b1a71a5920a8daa27e9723913..9f621fe7bd33a0d9f2d68237eebc1130f8f09f9b 100644
+index 18aac3da3c88f33b1a71a5920a8daa27e9723913..fa1bb5f4939c8bd75ce297d34ee4e79717d041e4 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
 @@ -83,13 +83,7 @@ import net.minecraft.util.valueproviders.UniformInt;
@@ -23,15 +23,18 @@ index 18aac3da3c88f33b1a71a5920a8daa27e9723913..9f621fe7bd33a0d9f2d68237eebc1130
  import net.minecraft.world.entity.ai.navigation.PathNavigation;
  import net.minecraft.world.entity.ai.village.ReputationEventType;
  import net.minecraft.world.entity.ai.village.poi.PoiManager;
-@@ -565,6 +559,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -565,6 +559,10 @@ public class ServerLevel extends Level implements WorldGenLevel {
  
      // Paper start - optimise checkDespawn
      public final List<ServerPlayer> playersAffectingSpawning = new java.util.ArrayList<>();
 +    public final List<ServerPlayer> playersAffectingMobSpawnerSpawning = new java.util.ArrayList<>();
++    public final List<ServerPlayer> playersAffectingNaturalSpawning = new java.util.ArrayList<>();
++
++
      // Paper end - optimise checkDespawn
      // Paper start - optimise get nearest players for entity AI
      @Override
-@@ -773,6 +768,12 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -773,6 +771,19 @@ public class ServerLevel extends Level implements WorldGenLevel {
                  this.playersAffectingSpawning.add(player);
              }
          }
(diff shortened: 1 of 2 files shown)


patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
```diff
@@ -0,0 +1,111 @@
...
+index 3ff999734d14e2b6e7828e117f5ee32a60c26bc1..430d61bd36197879b13cf592eac9e9c3b3d3025f 100644
+--- a/src/main/java/net/minecraft/world/entity/EntitySelector.java
++++ b/src/main/java/net/minecraft/world/entity/EntitySelector.java
+@@ -45,8 +45,13 @@ public final class EntitySelector {
+     public static final Predicate<Entity> PLAYER_AFFECTS_SPAWNING = (entity) -> {
+         return !entity.isSpectator() && entity.isAlive() && entity instanceof Player player && player.affectsSpawning;
+     };
++
++    public static final Predicate<Entity> PLAYER_AFFECTS_MOB_SPAWNER_SPAWNING = (entity) -> {
++        return !entity.isSpectator() && entity.isAlive() && entity instanceof Player player && player.affectsSpawning;
```
Sytm, line 67: If I see this correctly you want to do this?
```suggestion
+        return !entity.isSpectator() && entity.isAlive() && entity instanceof Player player && player.affectsMobSpawnersSpawning;
```

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
```diff
@@ -0,0 +1,111 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -2682,8 +2682,9 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+     }
+ 
+     // Paper start
+-    public void setAffectsSpawning(boolean affects) {
++    public void setAffectsSpawning(boolean affects, boolean affectsMobSpawners) {
+         this.getHandle().affectsSpawning = affects;
++        this.getHandle().affectsMobSpawnersSpawning = affectsMobSpawners;
+     }
```
Sytm: At the moment this would be an API break.

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
```diff
@@ -0,0 +1,111 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
+index 18aac3da3c88f33b1a71a5920a8daa27e9723913..ceacb821d11290e6f0d8a624c910666e8a055908 100644
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -565,6 +565,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
+ 
+     // Paper start - optimise checkDespawn
+     public final List<ServerPlayer> playersAffectingSpawning = new java.util.ArrayList<>();
++    public final List<ServerPlayer> playersAffectingNaturalSpawning = new java.util.ArrayList<>();
```
Sytm, line 31: At the moment this is unused?

Sytm (commented):
It looks like you made changes to the API of Paper but the patch for that is missing

TreemanKing:
> It looks like you made changes to the API of Paper but the patch for that is missing

Oops I must've had it on the wrong directory, my bad.

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
TreemanKing: So I guess I would have to seperate this and make it seperately with a new set and get for this. I got stuck towards the section of src/main/java/net/minecraft/world/entity/Mob where you can't look at both. What I could do is seperate both so they are different but then again, this might also break API for those who use the method to get rid of all spawns which I want to prevent as well.

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
TreemanKing: Kind of? I want to make it only impact Natural Spawns and not spawns from Spawners themself.

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
TreemanKing: Thanks! Will resolve this.

TreemanKing: This should be resolved.

patches/server/0994-Add-Mob-Spawner-support-to-AffectsSpawning.patch
TreemanKing: I believe I have changed this, however, I am unable to figure out how to change the section listed (CheckDespawn) and SkeletonTrapGoals "canUse" section as it seems pre hardcoded in and I'm not too sure how to change this.

lynxplay:
I am going to close this as I don't think a secondary check here is really the way to go.
As you already laid out, this logic is rough to edit. Breaking API for this is also 100% not worth it, given how niche this API is going to be.

An alternative approach that I could see working in the future would be an event called before the spawner searches for nearby players. To counterbalance the performance impact, this event might also allow for the definition of a backoff cooldown from even searching for players (while still running the rest of the tick logic).

This would 
a) allow to remove specific players from specific spawners in regards to their activation
b) allow spawners to skip player searching for n ticks (as specified by the backoff cooldown)
c) still allows what this PR aimed to achieve, completely exclude players from affecting spawning on spawners.

I hope you can understand my points here, if you have further questions, we are always around on our discord :)

PR #9584, opened 2023-08-07
Cache map ids on item frames

### LINKED ISSUE #9531: Possible Optimizations to Item Frame Map Updates

Item Frame Map Cursor Updates are laggy, sometimes taking in 5-13% of our server tick. In the past, a PR was added to allow them to be configurable (item-frame-cursor-update-interval). However, when maps are used as decoration (as in our server which as 11 thousand) they cause immense lag regardless. 
These updates are also misnamed- while they are probably intended to update the cursors on maps, they are also used to send the map canvas as well, so turning these updates off would mean that no map canvas are sent to the client. This also means that our server rather wants to turn the interval down, as this allows these maps to load faster, without any visual errors. 

The lag is primarily caused by WorldMapItem#getMapId, which for some odd reason seems to carry a significant overhead despite just accessing the NBT of the item stack. 

There is one optimization I know that can be added to make this section of the code:

**Cache Map ID in the Item Frame Entity Class**
ItemFrame already has a nice method where the map id can be cached and placed into the entity itself (onItemUpdate), which runs any time the item stack is changed or loaded on/with the entity. This cached integer would range from [0, +infinity), with non-map-items being able to set the map id to -1 as that is not a valid map id. 
As you can see in the method, this primarily would save the overhead created from calling #getMapId every few ticks to only needing it to be cached. This also helps a small overhead created by accessing the item stack saved inside the synced entity data. 

Not sure what else can be done about this, but I talked to @Warriorrrr who wrote the former PR and he might have some other things to add. There's probably some nitty gritty stuff inside how the update packet is created. 

We aren't hitting 50 mspt yet, but since I estimate that we will have 4 times as many item frame maps as we currently do, I hope to avoid this limit as much as possible so we don't have to switch machine or possibly limit our player counts. 

### Other

image

Warriorrrr:
This patch reduces the cost of broadcasting framed map changes to players by caching the map ID on the item frame itself, instead of reading it from the framed map's tag each time. Also reduces a bit of the logic done in tickCarriedBy for framed maps, since framed maps aren't going to have player decorations on them.

In a lot of spark reports I've seen the call to MapItem#getMapId(ItemStack) take up a few percent of the tick, which seems pretty weird despite it being a simple method, so if that was fixed then this PR probably wouldn't even be necessary.

Closes #9531

---
Download the paperclip jar for this pull request: [paper-9584.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/858460673.zip)

### DIFF
--- a/patches/server/0997-Cache-map-ids-on-item-frames.patch
+++ b/patches/server/0997-Cache-map-ids-on-item-frames.patch
@@ -1,0 +1,39 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Warrior <50800980+Warriorrrr@users.noreply.github.com>
+Date: Mon, 7 Aug 2023 12:58:28 +0200
+Subject: [PATCH] Cache map ids on item frames
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerEntity.java b/src/main/java/net/minecraft/server/level/ServerEntity.java
+index d934d07ad761319f338d4386536f68fde211c041..196280f54e397c69d32bd4d1f6ae666efdd93773 100644
+--- a/src/main/java/net/minecraft/server/level/ServerEntity.java
++++ b/src/main/java/net/minecraft/server/level/ServerEntity.java
+@@ -115,7 +115,7 @@ public class ServerEntity {
+                 ItemStack itemstack = entityitemframe.getItem();
+ 
+                 if (this.level.paperConfig().maps.itemFrameCursorUpdateInterval > 0 && this.tickCount % this.level.paperConfig().maps.itemFrameCursorUpdateInterval == 0 && itemstack.getItem() instanceof MapItem) { // CraftBukkit - Moved this.tickCounter % 10 logic here so item frames do not enter the other blocks // Paper - Make item frame map cursor update interval configurable
+-                    Integer integer = MapItem.getMapId(itemstack);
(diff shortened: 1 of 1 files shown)


democracycraft:
In testing this patch, we observed a lot of performance gains as we have a lot of item frames. 

Here is a spark report for before we merged the patch - https://spark.lucko.me/xhfPrGtMz0

I'll include one after, once I can replicate close-ish the player count and entities. :)

patches/server/0997-Improve-framed-map-tracker-ticking.patch
```diff
@@ -0,0 +1,82 @@
...
+     public boolean fixed;
++    public Integer cachedMapId; // Paper
+ 
+     public ItemFrame(EntityType<? extends ItemFrame> type, Level world) {
+         super(type, world);
+@@ -296,6 +297,7 @@ public class ItemFrame extends HangingEntity {
+     // Paper end
+ 
+     private void removeFramedMap(ItemStack itemstack) {
++        this.cachedMapId = null; // Paper
```
lynxplay: Not really needed. removeFramedMap is more of a side effect of removal of an item.
onItemChanged is called whenever the item in the frame changes, which should be exactly what this cache is after. Given that getMapId returns null for non map items, that call would work as both the setter and reset.

kennytv (approved):

frqnny:
WOOO so happy this was merged, thank you!

PR #9592, opened 2023-08-09
Added Mushroom spread configuration to world config

GrayR0ot:


patches/server/0005-Paper-config-files.patch
```diff
@@ -1837,6 +1837,12 @@ index 0000000000000000000000000000000000000000..a33de97340f14219291c4175e9194914
 +        public boolean portalSearchVanillaDimensionScaling = true;
 +        public boolean disableTeleportationSuffocationCheck = false;
 +        public IntOr.Disabled netherCeilingVoidDamageHeight = IntOr.Disabled.DISABLED;
++
++        public Mushroom mushroom;
++
++        public class Mushroom extends ConfigurationPart {
++            public int spreadLimit = 5;
```
lynxplay: As stated on the other PR, this should probably be a IntOr.Default.
We generally want to move away from magic values (like -1 for disabled or the 5 being defined in here)

patches/server/1000-Increase-vanilla-mushroom-spread-limit.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -40,10 +40,17 @@ public class MushroomBlock extends BushBlock implements BonemealableBlock {
+     @Override
+     public void randomTick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
+         if (random.nextFloat() < (world.spigotConfig.mushroomModifier / (100.0f * 25))) { // Spigot - SPIGOT-7159: Better modifier resolution
+-            int i = 5;
+             boolean flag = true;
+             Iterator iterator = BlockPos.betweenClosed(pos.offset(-4, -1, -4), pos.offset(4, 1, 4)).iterator();
+ 
++            // Paper start - Increase vanilla mushroom spread
++            int mushroomSpreadLimit = world.paperConfig().environment.mushroom.spreadLimit;
```
lynxplay: I don't see the point in moving this down here ? Keep the logic where the initial i was defined so the removal and addition are in the same spot.

lynxplay (changes requested):

patches/server/0005-Paper-config-files.patch
GrayR0ot: I tryed, but I can't use any IDE with Paper, I have a gradle issue with my IDEA. I looked for examples, but I haven't found.

patches/server/1000-Increase-vanilla-mushroom-spread-limit.patch
GrayR0ot: Fixed

patches/server/0005-Paper-config-files.patch
GrayR0ot: -1 mean unlimited, not disabled. In fact if the player let -1, it will calculate the maximum theorical value. Which is 9 x 9 x 3 in this case. It's better than letting the player calculate himself.

Source: https://minecraft.fandom.com/wiki/Mushroom#Spread

"When receiving a [random tick](https://minecraft.fandom.com/wiki/Tick#Random_tick), a mushroom has a chance of spreading to a nearby block (including even above and below diagonally) if there are fewer than five mushrooms of the same type in a **9×9×3 volume** around the original mushroom."

GrayR0ot: [95c4e17](https://github.com/PaperMC/Paper/pull/9592/commits/95c4e17fe71bc8bdcdb391ae5cc4be363672deaf)

lynxplay: Wouldn't "disabling'" a spread limit imply it to be, well, not be a factor? If my spread limit is the "default" I expect it to be vanilla, if I set it to 3, I expect it to be 3. If I disable it I expect it to not be enforced, allowing it to spread as far as possible.

GrayR0ot: I can add a simple setting like "disable-limit" which would ignore the spread-limit option, and then set it to the maximum (9 x 9 x 3). What do you think?

lynxplay: @Machine-Maker do you think it is worth looking into some addition IntOr to allow for both "disabled" and "default" here ?

GrayR0ot: No, because if the player put -1, like a few other paper/spigot settings it's disabled by default. In my case this will works. If the player put -1 or 1000 or 1M it will disable the limit.

GrayR0ot: But it still possible if needed for more clarity

lynxplay: As said before, I think it is worth looking into the ability to move away from the magic constant that is -1 simply for clarity reasons.
We introduced this originally when world heights allowed for negative values, -1 would work here for sure, given that the limit cannot be negative.
This is purely a discussion in regards to clarity to the end user, specifically if we want to stick to -1 here or if we want to look into potentially expanding our configuration API to allow for the usage of both "default" and "disabled" in a single variable to communicate the implication of `-1` in a proper word.

GrayR0ot: Oh I understand, yes you can add something like that. But I think it's more simple for a user to understand a field to enable/disable, & a field with the value when enabled.

lynxplay: We'll wait for some more input on this, beyond that, the PR generally looks good tho, I'll have to think about the naming of the config option, given that spread limit could also be interpreted as "the radius" of the spread.

GrayR0ot: Yes sure. The limit is not really easy to understand. I made a lot of test, in fact the limit is almost useless.
Minecraft will check in a 9 x 9 x 3 radius if there are already 5+ mushrooms. But it will trigger for all of them (all mushrooms can spread). I was looking to add an option to change the radius, but it's useless with spreading recursivity.

Machine-Maker:
Yeah, I think in this case an IntOr.DefaultOrDisabled would be nice. I, by far, prefer using words instead of random constants. You never know what might change in the future. Like plenty of `-1` to disable some height thing broke when y went below 0. It's best to use `disabled` or `default` to convey precise meaning.

GrayR0ot:
I tryed to build the paper on my side i got the same error, without my commits BTW. Is the WorldConfiguration broken?

GrayR0ot (commented):
Added IntOrDefault as asked

patches/server/1000-Increase-vanilla-mushroom-spread-limit.patch
```diff
@@ -0,0 +1,29 @@
...
+index 5238b23cd3bca21e2b14c9be15699b95ad267e24..1afa460b388e1a772b278ac6710adc211fd5f2be 100644
+--- a/src/main/java/net/minecraft/world/level/block/MushroomBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/MushroomBlock.java
+@@ -40,10 +40,17 @@ public class MushroomBlock extends BushBlock implements BonemealableBlock {
+     @Override
+     public void randomTick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
+         if (random.nextFloat() < (world.spigotConfig.mushroomModifier / (100.0f * 25))) { // Spigot - SPIGOT-7159: Better modifier resolution
+-            int i = 5;
++            // Paper start - Increase vanilla mushroom spread
++            int mushroomSpreadLimit = world.paperConfig().environment.mushroom.spreadLimit.or(5);
```
Warriorrrr, line 17: Leaving the original i = 5 and using the i as the default in the .or might look a bit cleaner, but generally looks pretty good.

GrayR0ot:
I created a new PR with all new changes from Paper/master for better lisibility & workflow on my side.

https://github.com/PaperMC/Paper/pull/9599

GrayR0ot:
ggzgeg

PR #9595, opened 2023-08-12
Fix custom statistic criteria creation

### LINKED ISSUE #9507: PLAY_ONE_MINUTE statistic doesn't work as scoreboard objective
### Expected behavior

I set up an objective where the criteria is the PLAY_ONE_MINUTE statistic, it should track a players playtime (in ticks, bad name)

### Observed/Actual behavior

It stays at 0 (or whatever you initialize it to)

### Steps/models to reproduce

Set an objective with the statistic. 
Query the score.

### Plugin and Datapack List

LuckPerms and Votifier

### Paper version

This server is running Paper version git-Paper-83 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: c793bd9)
Previous version: git-Paper-9 (MC: 1.20)

### Other

I was able to use other statistics like crouch and fly distances as objective criteria, and they worked fine.

NoahvdAa:
When looking into fixing #9507 I noticed a lot more statistic criterion than just the play time were broken by the 'naive' enum -> key conversion. See [the wiki](https://minecraft.fandom.com/wiki/Statistics#List_of_custom_statistic_names) for the full list, everything that doesn't match with the enum is broken.

Rather than having to update the entire enum with the correct names in the way lynx showed [in this comment](https://github.com/PaperMC/Paper/issues/9507#issuecomment-1639815454), I opted for grabbing the correct statistic key from the server itself, using unsafevalues to bridge the gap. This should be easier to maintain and will prevent us missing the enum update should the keys change again.

A proper, long-term solution would be #5691, but until then I believe this is best fix.

Fixes #9507.

### DIFF
--- a/patches/api/0424-Fix-custom-statistic-criteria-creation.patch
+++ b/patches/api/0424-Fix-custom-statistic-criteria-creation.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Noah van der Aa <ndvdaa@gmail.com>
+Date: Sat, 12 Aug 2023 15:33:55 +0200
+Subject: [PATCH] Fix custom statistic criteria creation
+
+
+diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
+index c661eab343ae76488de701630424e2d589f44fc0..de36dab37d3af249cf51573d163731c12346066a 100644
+--- a/src/main/java/org/bukkit/UnsafeValues.java
++++ b/src/main/java/org/bukkit/UnsafeValues.java
+@@ -249,5 +249,7 @@ public interface UnsafeValues {
+      * @throws IllegalStateException if no biome by the given key is registered.
+      */
+     void setBiomeKey(RegionAccessor accessor, int x, int y, int z, NamespacedKey biomeKey);
++
++    String getStatisticCriteriaKey(@NotNull org.bukkit.Statistic statistic);
+     // Paper end
+ }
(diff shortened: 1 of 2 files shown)


patches/server/1002-Fix-custom-statistic-criteria-creation.patch
```diff
@@ -0,0 +1,23 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
++++ b/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
+@@ -622,6 +622,12 @@ public final class CraftMagicNumbers implements UnsafeValues {
+         net.minecraft.core.Holder<net.minecraft.world.level.biome.Biome> biomeBase = cra.getHandle().registryAccess().registryOrThrow(net.minecraft.core.registries.Registries.BIOME).getHolderOrThrow(net.minecraft.resources.ResourceKey.create(net.minecraft.core.registries.Registries.BIOME, org.bukkit.craftbukkit.util.CraftNamespacedKey.toMinecraft(biomeKey)));
+         cra.setBiome(x, y, z, biomeBase);
+     }
++
++    @Override
++    public String getStatisticCriteriaKey(org.bukkit.Statistic statistic) {
++        if (statistic.getType() != org.bukkit.Statistic.Type.UNTYPED) return "minecraft.custom:minecraft." + statistic.getKey().getKey();
```
NoahvdAa: For context; `CraftStatistic#getNMSStatistic` only allows looking up untyped statistics. Typed statistics don't/shouldn't work with the `Criteria#static` method anyway, as described in the javadocs, so this keeps the same behavior as before.

Machine-Maker (approved):

lynxplay (approved):

PR #7400, opened 2022-01-24
Add missing InventoryHolders to inventories

Machine-Maker:
All inventories that don't have block entity inventories associated with them, beacons, anvils, stonecutters, grindstone, smithing table, loom, cartography table had null inventory holders when that's not really correct. the `BlockInventoryHolder` seems to fit perfectly there.

PR #9596, opened 2023-08-12
Add getLocation to the composter output container

### LINKED ISSUE #9437: InventoryMoveItemEvent - calling event.getSource().getLocation() from the event on a composter causes NPE
### Expected behavior

You should be able to use event.getSource().getLocation() from the  InventoryMoveItemEvent to obtain the location of the source inventory.

### Observed/Actual behavior

when a composter is full and creates bonemeal, and a hopper is used to pull out the bonemeal, if you attempt to get the location of the source inventory from the InventoryMoveItemEvent a NPE will occur. 

`[08:19:40 INFO]: [IllegalStack] [STDOUT] Source is: COMPOSTER
[08:19:40 INFO]: [IllegalStack] [STDOUT] e.getSource().getLocation() was null.

[08:19:41 INFO]: [IllegalStack] [STDOUT] Source is: HOPPER
[08:19:41 INFO]: [IllegalStack] [STDOUT] Location from event: Location{world=CraftWorld{name=world},x=42.0,y=70.0,z=262.0,pitch=0.0,yaw=0.0}`

As you can see from the debug code above, when getLocation() is called from the event on a composter a NPE occurs, however when it is called on a HOPPER this same error does not occur.   It is also worth noting this only seems to happen in paper and not regular spigot.

### Steps/models to reproduce

`    @EventHandler
    public void onHopperTest(InventoryMoveItemEvent e) {
    	
    	
    	if(e.getSource().getHolder() instanceof BlockInventoryHolder) {
    		BlockInventoryHolder source = ((BlockInventoryHolder)e.getSource().getHolder());
    		System.out.println("Source is: " + source.getBlock().getType());
    		if(e.getSource().getLocation() != null)
    			System.out.println("Location from event: " + e.getSource().getLocation());
    		else
    			System.out.println("e.getSource().getLocation() was null.");
    	}
    	
    }`

### Plugin and Datapack List

[08:25:04 INFO]: Server Plugins (3):
[08:25:04 INFO]: Bukkit Plugins:
[08:25:04 INFO]:  - Essentials, IllegalStack, ProtocolLib

[08:25:41 INFO]: There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
[08:25:41 INFO]: There are no more data packs available

### Paper version

version
[08:26:12 INFO]: Checking version, please wait...
[08:26:13 INFO]: This server is running Paper version git-Paper-61 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 171ba7c)
Previous version: git-Paper-45 (MC: 1.20.1)

### Other

_No response_

Sytm:
As this is my first contribution to Paper I am open for suggestions for naming the patch or perhaps telling me if there is already an existing patch to merge this into.

Fixes #9437

### DIFF
--- a/patches/server/0936-Fix-certain-inventories-returning-null-Locations.patch
+++ b/patches/server/0936-Fix-certain-inventories-returning-null-Locations.patch
@@ -3,30 +3,32 @@ From: Jake Potrebic <jake.m.potrebic@gmail.com>
 Date: Wed, 15 Mar 2023 18:29:45 -0700
 Subject: [PATCH] Fix certain inventories returning null Locations
 
-Wandering Trader, AbstractHorse, and Beacon inventories returned null locations
+Wandering Trader, AbstractHorse, Beacon and Composter inventories returned null locations
 when a block or entity location is readily available
 
-diff --git a/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java b/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
-index 49d7109b6ca63c8073db777549a65b2fcb70967e..9b35a50eb99551f28d45fd5114e0401b4c54646e 100644
---- a/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
-+++ b/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
-@@ -345,7 +345,15 @@ public abstract class AbstractHorse extends Animal implements ContainerListener,
-     public void createInventory() {
-         SimpleContainer inventorysubcontainer = this.inventory;
+Co-authored-by: Lukas Planz <lukas.planz@web.de>
+
+diff --git a/src/main/java/net/minecraft/world/SimpleContainer.java b/src/main/java/net/minecraft/world/SimpleContainer.java
+index e938e46a142f3764f8224a1fd035e7d541b59c56..9d1ee40456a8d7001eee654a62e62cab2626305a 100644
+--- a/src/main/java/net/minecraft/world/SimpleContainer.java
++++ b/src/main/java/net/minecraft/world/SimpleContainer.java
+@@ -63,6 +63,16 @@ public class SimpleContainer implements Container, StackedContentsCompatible {
  
--        this.inventory = new SimpleContainer(this.getInventorySize(), (org.bukkit.entity.AbstractHorse) this.getBukkitEntity()); // CraftBukkit
+     @Override
+     public Location getLocation() {
 +        // Paper start
-+        this.inventory = new SimpleContainer(this.getInventorySize(), (org.bukkit.entity.AbstractHorse) this.getBukkitEntity()) // CraftBukkit
-+        {
-+            @Override
-+            public org.bukkit.Location getLocation() {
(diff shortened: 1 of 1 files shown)


kennytv:
Thank you for the PR! I haven't looked into this fully yet, but wouldn't the input container below it also need it overridden?

Sytm:
Yeah, you were right. I only tested this with the composter as the source inventory, but not as the destination inventory

Machine-Maker (changes requested):
I think there is a much better way to fix this issue.

Instead of overriding `getLocation`, just make `SimpleContainer#getLocation` check if the field `bukkitOwner` is set to an instance of `BlockInventoryHolder` and if it is, return the location of the block. That covers all 3 composter container types (any any future block inventory holders that might not override getLocation).

Also this should be merged into an existing patch "Fix certain inventories returning null locations"

----

You could go futher and check for other impls of InventoryHolder like `Entity` and return the location for that entity in `SimpleContainer#getLocation`. That would fix 2 (I think) more places where this happens, villager inventories and the horse inventories that the patch I mentioned fixes by doing the same thing you did.

Sytm:
When the entity implements `InventoryHolder` and `Container`, the getLocation call is handled by `CraftEntity.getLocation`, but the special case here is the horse where the `(Simple)Container` is not extended but just created.

patches/server/0938-Fix-certain-inventories-returning-null-Locations.patch
```diff
@@ -3,47 +3,48 @@ From: Jake Potrebic <jake.m.potrebic@gmail.com>
...
+@@ -40,6 +40,13 @@ public class BeaconMenu extends AbstractContainerMenu {
+         super(MenuType.BEACON, syncId);
+         this.player = (Inventory) inventory; // CraftBukkit - TODO: check this
+         this.beacon = new SimpleContainer(1) {
 +            // Paper start
-+            @Override
-+            public org.bukkit.Location getLocation() {
-+                return context.getLocation();
++            {
++                if (context.getLocation() != null) {
```
Sytm: I don't know if I should check the context like this or with `context != ContainerLevelAccess.NULL`, although this would permit for other implementations that would return null

Machine-Maker: Don't do beacon like this. The correct solution is to add a block inventory holder for beacons, which is done here https://github.com/PaperMC/Paper/pull/7400

Sytm: Reverted that last change again. I was thinking about doing it like that, but that was way more complicated than I thought and felt like the beacon was a bit more comparable to the composter.

But otherwise is this ok?

Machine-Maker (approved):
Feel free to add yourself as a co author on the patch. I'll probably merge that PR I mentioned into this patch after.

To add yourself, just put `Co-authored-by: <git info>` into the patch commit message.


EDIT:
> The Merchant container does not use SimpleContainer, so this must stay.

I meant the villager inventory itself. Like if you do getInventory on a villager entity, it returns a CraftInventory wrapping a SimpleContainer what wouldn't have its location set. But now it does.

PR #9599, opened 2023-08-13
Feat/mushroom spread config

GrayR0ot:

From: https://github.com/PaperMC/Paper/pull/9592

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1479,10 +1479,10 @@ index 0000000000000000000000000000000000000000..f0d4ec73bc8872a85e34f5c6b4d342e7
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..d47c57afafc01e25b965f1844938b2516a7bd031
+index 0000000000000000000000000000000000000000..8550600818842fa6c9a934dff8cb3282b8ec7d4f
 --- /dev/null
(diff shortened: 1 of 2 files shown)


Machine-Maker:
For future reference, you don't need to recreate the PR just to include the latest changes, just rebase your branch on the upstream master branch. Then you can force push to this branch.

patches/server/0005-Paper-config-files.patch
```diff
@@ -1841,6 +1841,12 @@ index 0000000000000000000000000000000000000000..d47c57afafc01e25b965f1844938b251
 +        public boolean portalSearchVanillaDimensionScaling = true;
 +        public boolean disableTeleportationSuffocationCheck = false;
 +        public IntOr.Disabled netherCeilingVoidDamageHeight = IntOr.Disabled.DISABLED;
++
++        public Mushroom mushroom;
++        public class Mushroom extends ConfigurationPart {
++            public BooleanOrDefault disableLimit = BooleanOrDefault.USE_DEFAULT;
++            public IntOr.Default spreadLimit = IntOr.Default.USE_DEFAULT;
```
Machine-Maker, line 1848: Like I mentioned in your previous PR, I think it would be better if this was combined into 1 setting. Then it wouldn't need a new section. You can create an IntOrDefaultOrDisabled which can be `default`, `disabled` or an integer. 

You can model that type based on the existing types, just add another constant string value that is valid.

Machine-Maker (changes requested):

patches/server/0005-Paper-config-files.patch
GrayR0ot: I can't do that for the moment. I'll ask someone on the discord.

PR #9601, opened 2023-08-13
Add ExplosionPower methods to LargeFireball

Doc94:
This PR add the get/set method to LargeFireball to the explosion power, if not wrong the only way to get that value is if the shooter is a Ghast but this not apply to custom LargeFireball generated.

~~This PR starts because wanna get the radius.. in another entites are more hardcoded..~~


Lulu13022002:
This is kinda redundant with getYield/setYield ?

Doc94:
> This is kinda redundant with getYield/setYield ?

Oh yeah... I confuse because see in another place the explosionPower, explosionPowerRadius and now like say the yield.....

XD

Warriorrrr:
Is this still needed if those methods do the same?

PR #9604, opened 2023-08-14
Add clicked position to PlayerUseUnknownEntityEvent

kennytv:
Fixes the event being called 4 times (2 per hand) when a player right clicks with no difference to the event data. Also adds some javadoc and removes an unnecessary access transformer

### DIFF
--- a/patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
+++ b/patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
@@ -10,7 +10,7 @@ Co-authored-by: Nassim Jahnke <nassim@njahnke.dev>
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerUseUnknownEntityEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerUseUnknownEntityEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..ffd2250dcf9041566381abc1d5e4fd3eebd3781b
+index 0000000000000000000000000000000000000000..c7d853121becd8f6980173aff6322e327f4037a1
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerUseUnknownEntityEvent.java
 @@ -0,0 +1,82 @@
@@ -48,7 +48,7 @@ index 0000000000000000000000000000000000000000..ffd2250dcf9041566381abc1d5e4fd3e
 +    }
 +
 +    /**
-+     * Returns the location of the unknown entity that was interacted with.
++     * Returns the entity id of the unknown entity that was interacted with.
 +     *
 +     * @return the entity id of the entity that was interacted with
 +     */
--- a/patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
+++ b/patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
@@ -3,56 +3,96 @@ From: Jedediah Smith <jedediah@silencegreys.com>
 Date: Sat, 2 Apr 2016 05:08:36 -0400
 Subject: [PATCH] Add PlayerUseUnknownEntityEvent
 
+Adds the PlayerUseUnknownEntityEvent to be used by plugins dealing with
+virtual entities/entities that are not actually known to the server.
+
+Co-authored-by: Nassim Jahnke <nassim@njahnke.dev>
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerUseUnknownEntityEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerUseUnknownEntityEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..09cfdf48ead8f03f3497646537292174241b0868
+index 0000000000000000000000000000000000000000..ffd2250dcf9041566381abc1d5e4fd3eebd3781b
 --- /dev/null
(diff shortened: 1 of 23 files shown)


patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
```diff
@@ -3,56 +3,96 @@ From: Jedediah Smith <jedediah@silencegreys.com>
...
 +import org.bukkit.entity.Player;
 +import org.bukkit.event.HandlerList;
 +import org.bukkit.event.player.PlayerEvent;
 +import org.bukkit.inventory.EquipmentSlot;
 +import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
 +
++/**
++ * Represents an event that is called when a player right-clicks an unknown entity.
++ * Useful for plugins dealing with virtual entities (entities that don't actually spawned on the server).
```
lynxplay, line 29:
```suggestion
+ * Useful for plugins dealing with virtual entities (entities that aren't actually spawned on the server).
```

patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
```diff
@@ -3,56 +3,96 @@ From: Jedediah Smith <jedediah@silencegreys.com>
...
++    public PlayerUseUnknownEntityEvent(@NotNull Player who, int entityId, boolean attack, @NotNull EquipmentSlot hand, @Nullable Position clickedPosition) {
 +        super(who);
 +        this.entityId = entityId;
 +        this.attack = attack;
 +        this.hand = hand;
++        this.clickedPosition = clickedPosition;
 +    }
 +
++    /**
++     * Returns the location of the unknown entity that was interacted with.
```
lynxplay:
```suggestion
+     * Returns the entity id of the unknown entity that was interacted with.
```

patches/server/0083-Add-PlayerUseUnknownEntityEvent.patch
```diff
@@ -3,63 +3,70 @@ From: Jedediah Smith <jedediah@silencegreys.com>
...
++    private void callPlayerUseUnknownEntityEvent(ServerboundInteractPacket packet, InteractionHand hand, @Nullable net.minecraft.world.phys.Vec3 pos) {
 +        this.cserver.getPluginManager().callEvent(new com.destroystokyo.paper.event.player.PlayerUseUnknownEntityEvent(
 +            this.getCraftPlayer(),
 +            packet.getEntityId(),
-+            packet.getActionType() == ServerboundInteractPacket.ActionType.ATTACK,
-+            hand == InteractionHand.MAIN_HAND ? EquipmentSlot.HAND : EquipmentSlot.OFF_HAND
-+        ));
++            packet.isAttack(),
++            hand == InteractionHand.MAIN_HAND ? EquipmentSlot.HAND : EquipmentSlot.OFF_HAND,
++            pos != null ? io.papermc.paper.math.Position.fine(pos.x, pos.y, pos.z) : null)
```
lynxplay: Should we be adding MCUtil helpers for position creation (from Vec3 and BlockPos respectively) if we want to expand the usage of position here ?

lynxplay (dismissed):

patches/server/0009-MC-Utils.patch
```diff
@@ -3960,6 +3964,14 @@ index 0000000000000000000000000000000000000000..6779a0cc401231c53545dd22827b404b
 +        return new BlockPos(vec.getBlockX(), vec.getBlockY(), vec.getBlockZ());
 +    }
 +
++    public FinePosition toPosition(Vec3 vector) {
++        return Position.fine(vector.x, vector.y, vector.z);
++    }
++
++    public BlockPosition toPosition(Vec3i vector) {
++        return Position.block(vector.getX(), vector.getY(), vector.getZ());
++    }
```
NoahvdAa: this doesn't compile

NoahvdAa (changes requested):
Looks good apart from the fact that it doesn't compile

kennytv:
Shhh

NoahvdAa (approved):

patches/api/0036-Add-PlayerUseUnknownEntityEvent.patch
```diff
@@ -3,56 +3,96 @@ From: Jedediah Smith <jedediah@silencegreys.com>
...
 +    }
 +
++    /**
++     * Returns the exact position that was clicked, or null if not available.
++     * See {@link org.bukkit.event.player.PlayerInteractAtEntityEvent} for more details.
++     *
++     * @return the exact position that was clicked, or null if not available
++     * @see org.bukkit.event.player.PlayerInteractAtEntityEvent
++     */
++    public @Nullable Position getClickedPosition() {
```
Machine-Maker: I think the vector in the packet is a relative vector to the armor stand, it's not a position.

Machine-Maker (changes requested):

Cryptite:
This isn't exactly the right place for this comment, probably, but is there some way in which this event could either be cancelled, or toggled as "consumed" in such a way where interacting with an unknown entity stops all subsequent future actions? And/or maybe there's another way to already handle this. Examples are:

- Right-clicking a packet NPC holding a consumable both interacts with it, but also you start eating.
- Same thing but when holding projectiles (pearls, potions, etc)

kennytv:
It'd be best if you created an issue for that

Machine-Maker (approved):

PR #9605, opened 2023-08-14
Rewrite framed map tracker ticking

### LINKED ISSUE #9597: Optimize MapItemSavedData#tickCarriedBy
### Description of issue

MapItemSavedData#tickCarriedBy is currently used by both a map itemstack ticked inside an ItemFrame as well as the items in the player inventory.

Given that maps can be duplicated, MapItemSavedData#tickCarriedBy is potentially called multiple times per tick, for each instance of the map around (either in player inventories or item frames). #tickCarriedBy however is terribly designed in regards to the duplicate existence of the same map and should be thoroughly investigated for potential performance improvements:

## Examples
`carriedByPlayers` is used to track which players currently carry the map. However, the itemframe calls this method for each player tracking the itemframe entity, meaning that in a single tick, a player may be added to said list just to be removed later  by the iteration through the carried list.

The tickCarriedBy additionally loads Decorations from the compound tag of the item stack over and over.

Warriorrrr:
Rewrites the tracking code for framed maps to remove the use of the tickCarriedBy method and the HoldingPlayer class. the tickCarriedBy method contained a lot of code that did not apply to framed maps at all, and by moving the parts that did elsewhere, we can essentially skip it.

Closes #9597

### DIFF
--- a/patches/server/1005-Rewrite-framed-map-tracker-ticking.patch
+++ b/patches/server/1005-Rewrite-framed-map-tracker-ticking.patch
@@ -1,0 +1,298 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Warrior <50800980+Warriorrrr@users.noreply.github.com>
+Date: Mon, 14 Aug 2023 17:34:53 +0200
+Subject: [PATCH] Rewrite framed map tracker ticking
+
+Rewrites the tracking code for framed maps to remove the use of the tickCarriedBy method and the HoldingPlayer class.
+The tickCarriedBy method contained a lot of code that did not apply to framed maps at all, and by moving the parts
+that did elsewhere, we can essentially skip it. The only logic that's ran inside the ServerEntity#sendChanges for maps
+now is just updating dirty map/decoration data.
+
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
Rebased for 1.20.6 https://github.com/SparklyPower/SparklyPaper/blob/ver/1.20.6/patches/server/0004-Rewrite-framed-map-tracker-ticking.patch

Outfit8TSB (approved):
this was inside experimental paper fork called sparklypaper, and now we get it in paper.
This is a Very Necessary Bugfix!
Super duper Very Very Double Backflip Beanboozled Approved Approved!

Warriorrrr:
I was looking over this PR again and it seems like my approach here was a bit flawed, since it did not account for one framed map being viewed in multiple locations and being updated, only players viewing the item frame that's first in the entity iteration order would've seen the changes.

There are still some other gains that could be made, such as reducing the amount of map update packets being sent to players while no changes actually occurred, and the mess that is the tickCarriedBy (seriously, maps in item frames shouldn't have to check if the player is wearing a pumpkin).

PR #9606, opened 2023-08-15
Fix startup and runtime crashes (experimental branch)

BlockyTheDev:
Like written on Discord with @Machine-Maker , here the pull request to fix a startup crash and the runtime crash occuring around three minutes after startup, found while wanting to experiment with the `experimental` Paper branch.

Changes following things:
- An erroneous change in the Custom-Potion-Mixes-Patch leading to a server crash at startup
- A server crash occuring around three minutes after startup, coming from the timings system because of an unexpected exception of the type `java.lang.ClassCastException`. While doing this I also migrated the code away from using the deprecated class `OldEnum`, introduced in the (Craft-)Bukkit pull request for enum to class migration for backwards compatibility, used in the `EntityType` related code of the timings system.

All the code changes were successfully tested through several server startups and several created timings reports (with different worlds and situations).

If i should post pictures of the exceptions, just write and I will post them here.

### DIFF
--- a/patches/api/0010-Timings-v2.patch
+++ b/patches/api/0010-Timings-v2.patch
@@ -717,10 +717,10 @@ index 0000000000000000000000000000000000000000..199789d56d22fcb1b77ebd56805cc28a
 +}
 diff --git a/src/main/java/co/aikar/timings/TimingHistory.java b/src/main/java/co/aikar/timings/TimingHistory.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..bf9f5d7a46c08c3352468cf738a9605bfe777b35
+index 0000000000000000000000000000000000000000..0a0e6308923f466ca96786b74811b9beeca83d73
 --- /dev/null
 +++ b/src/main/java/co/aikar/timings/TimingHistory.java
-@@ -0,0 +1,359 @@
+@@ -0,0 +1,358 @@
 +/*
 + * This file is licensed under the MIT License (MIT).
 + *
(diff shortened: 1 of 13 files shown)


patches/api/0010-Timings-v2.patch
```diff
@@ -749,24 +749,22 @@ index 0000000000000000000000000000000000000000..b9544c6f2adb0a8b32abd8ab3f1bdc5e
 +import co.aikar.timings.TimingHistory.RegionData.RegionId;
 +import com.google.common.base.Function;
 +import com.google.common.collect.Sets;
-+import java.util.HashMap;
++
++import java.util.*;
```
Warriorrrr: Star import should be undone

Warriorrrr (changes requested):

Leguan16:
Isn't the timing stuff subject to be replaced with spark? So any changes to that would be kind of a waste of time. Or not?

BlockyTheDev:
> Isn't the timing stuff subject to be replaced with spark? So any changes to that would be kind of a waste of time. Or not?

Yes it should be replaced in the future as I read, but  until it is replaced it should not lead to server crashes. The EntityType change was mainly for the reason that BlockType and EntityType are similiar.

patches/api/0010-Timings-v2.patch
BlockyTheDev: Moved to normal imports

patches/api/0010-Timings-v2.patch
```diff
@@ -749,24 +749,27 @@ index 0000000000000000000000000000000000000000..b9544c6f2adb0a8b32abd8ab3f1bdc5e
...
 +import co.aikar.util.LoadingMap;
 +import co.aikar.util.MRUMapCache;
 +
 +import java.lang.management.ManagementFactory;
-+import java.util.Collection;
-+import java.util.EnumMap;
-+import java.util.List;
-+import java.util.Map;
-+import java.util.Set;
++
```
Machine-Maker: Remove this diff, just looks like import re-ordering

Machine-Maker (changes requested):
Other than that, looks good.

I skipped the timings stuff cause it is all deprecated, but I didn't realize that it caused a crash so yeah, I guess it should be migrated.

patches/api/0010-Timings-v2.patch
BlockyTheDev: Should be fixed now.

Machine-Maker (approved):

PR #6680, opened 2021-09-29
Add Material#getPlacedMaterial

Machine-Maker:

PR #9607, opened 2023-08-15
Add `Material#getPlacedType` API

tal5:
## Additions

- `Material#getPlacedType` - returns the type an item would be placed as (usually the same, but is different in some cases, e.g. pumpkin seeds -> pumpkin stem).
- `UnsafeValues#getPlacedType(Material)` - bridge method for `Material#getPlacedType`.

## Notes

- Currently it returns a `Material`, as that's what's stored internally - might make sense to return `BlockData` to avoid potential issues with more complex data being added in the future/as it might make sense to represent it as `BlockData`?
- `BlockData#getPlacementMaterial` (other way around from this) seems to return `AIR` when a block doesn't have an item equivalent, while this returns `null` for items that can't be placed.
`null` makes more sense imo, but either is fine I guess - let me know if that should be changed to match.

### DIFF
--- a/patches/api/0426-Add-Material-getPlacedType-API.patch
+++ b/patches/api/0426-Add-Material-getPlacedType-API.patch
@@ -1,0 +1,61 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aya <31237389+tal5@users.noreply.github.com>
+Date: Tue, 15 Aug 2023 10:05:01 +0100
+Subject: [PATCH] Add Material#getPlacedType API
+
+
+diff --git a/src/main/java/org/bukkit/Material.java b/src/main/java/org/bukkit/Material.java
+index 03b47012447430a350e152920f754d993d4023db..ba087077f948bcdfa18708dc1804badb9cebc6b0 100644
+--- a/src/main/java/org/bukkit/Material.java
++++ b/src/main/java/org/bukkit/Material.java
+@@ -11047,4 +11047,23 @@ public enum Material implements Keyed, Translatable, net.kyori.adventure.transla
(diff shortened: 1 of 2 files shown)


patches/api/0426-Add-Material-getPlacedType-API.patch
```diff
@@ -0,0 +1,61 @@
...
++     * <p>
++     * Some examples include:
++     * <pre>
++     *     {@link Material#REDSTONE} -> {@link Material#REDSTONE_WIRE}
++     *     {@link Material#PUMPKIN_SEEDS} -> {@link Material#PUMPKIN_STEM}
++     * </pre>
++     * @param item An item material.
++     * @return The material that would be placed down by {@code item}, or {@code null} if it can't be placed.
++     * @throws IllegalArgumentException if {@code item} isn't an item ({@link Material#isItem()} is {@code false})
++     */
```
Machine-Maker, line 57: Can drop the javadoc here, just increases workload to maintain it since its a copy

Machine-Maker (changes requested):
I think returning BlockData would be more appropriate if we had a method that called `BlockItem#getPlacementState` which returns the equivalent BlockData in nms (BlockState). For just this, Material is fine.

----

I guess since Material is just going to be straight up deprecated by upstream hopefully next major version, I don't really have an issue adding more methods to Material that only apply to items or blocks but not both

Warriorrrr:
This seems to be a duplicate of https://github.com/PaperMC/Paper/pull/6680 if I'm not mistaken

Owen1212055:
Closing, as in general we are waiting for upstream's blocktype/itemtype API to better expose this types of API.

PR #9608, opened 2023-08-15
[ci skip] update editorconfig and gitattributes

Machine-Maker:


.gitattributes
```diff
@@ -1,4 +1,4 @@
-* text=auto
+* text=auto eol=lf
```
Machine-Maker, line 1: Is this correct? or should it just be `text eol=lf`? I think without the `=auto` it won't ignore binary files?

MiniDigger: since git 2.10, ye, so this is correct
https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248

MiniDigger (approved):

Owen1212055:
``
ij_java_generate_final_locals = true
ij_java_generate_final_parameters = true
``

Is this something we are pushing? This is a bit of a change as I am not sure everyone does this, unless we want to push this within our codebase.

kennytv (approved):

PR #5242, opened 2021-02-24
Add EntityExtinguishEvent

aurorasmiles:
Closes #5143
Replaces #5144 
I tried to see if I could manage to fix that issue and add that event. After finishing, I noticed that there already way a PR for this, but as it was made by Ivan and he is banned from the Org, I figured to open a PR anyway.
I haven't implemented cancellable, because I figured this would cause a lot of issues in the long run, and I wasn't sure how to properly address that.
Other than that there is another extinguish called in EntityHuman line 1211, but I decided to not call the event there, since the entity was never technically on fire

PR #9610, opened 2023-08-15
Add EntityExtinguishEvent

### LINKED ISSUE #5143: EntityExtinguishEvent
Currently there is no way to prevent an entity's fire ticks being reset to zero when the entity enters water or is in the rain.  It would be helpful if there was an EntityExtinguishEvent that is fired when an entity enters water or rain while on fire.  It would be possible to cancel, and you would be able to get the entity that is being extinguished and also the cause of the entity being extinguished.  As far as I know, there is currently no way to accomplish this.  I have tried adding to the fire ticks when an entity (in this case a player) is in water, but it is just being reset to zero instantly.  For 1.17, this event would also need to be fired when an entity enters powder snow.

Warriorrrr:
Closes https://github.com/PaperMC/Paper/issues/5143, continuation of https://github.com/PaperMC/Paper/pull/5242

Also fixes two vanilla bugs related to extinguishing

### DIFF
--- a/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
@@ -54,6 +54,18 @@ https://bugs.mojang.com/browse/MC-264285
 https://bugs.mojang.com/browse/MC-84789
   Fix wild wolves not considering bones interesting
 
+https://bugs.mojang.com/browse/MC-245603
+  Fix powdered snow not extinguishing fire arrows
+
+https://bugs.mojang.com/browse/MC-214647
+  Fix arrows not making a sound when extinguished
+
+https://bugs.mojang.com/browse/MC-245680
+  Fix placing powdered snow on a flaming arrow not consuming it
+
+== AT ==
+public net/minecraft/world/entity/Entity playEntityOnFireExtinguishedSound()V
+
 Co-authored-by: William Blake Galbreath <blake.galbreath@gmail.com>
 
 diff --git a/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java b/src/main/java/net/minecraft/core/dispenser/DispenseItemBehavior.java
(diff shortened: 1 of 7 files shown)


NoahvdAa:
Bug fixes should definitely go in their own patch, not in the implementation of an event.

patches/server/1005-Add-EntityExtinguishEvent.patch
```diff
@@ -0,0 +1,243 @@
...
+index dc5fda83a66afbfeb7897fc20b4742899d8aca08..ea1eb78bbb4dd1a78df340d14ba688109f71966b 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1034,7 +1034,8 @@ public class ServerPlayer extends Player {
+         this.awardStat(Stats.DEATHS);
+         this.resetStat(Stats.CUSTOM.get(Stats.TIME_SINCE_DEATH));
+         this.resetStat(Stats.CUSTOM.get(Stats.TIME_SINCE_REST));
+-        this.clearFire();
++        if (this.getRemainingFireTicks() > 0) new io.papermc.paper.event.entity.EntityExtinguishEvent(this.getBukkitEntity(), io.papermc.paper.event.entity.EntityExtinguishEvent.Cause.DEATH).callEvent(); // Paper - add EntityExtinguishEvent
++        this.clearFire(io.papermc.paper.event.entity.EntityExtinguishEvent.Cause.DEATH); // Paper
```
Lulu13022002: This event is called twice here.

patches/server/1005-Add-EntityExtinguishEvent.patch
```diff
@@ -0,0 +1,243 @@
...
+diff --git a/src/main/java/net/minecraft/world/level/block/PowderSnowBlock.java b/src/main/java/net/minecraft/world/level/block/PowderSnowBlock.java
+index fb0cdcf1d7c26baaaf03a730af7f2d21973d4ea0..d7f81417443db972a594d0354f9deec8dd8af872 100644
+--- a/src/main/java/net/minecraft/world/level/block/PowderSnowBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/PowderSnowBlock.java
+@@ -78,7 +78,7 @@ public class PowderSnowBlock extends Block implements BucketPickup {
+                 // CraftBukkit end
+                 // Paper start - fix MC-245603
+                 if (entity instanceof net.minecraft.world.entity.projectile.AbstractArrow) {
+-                    entity.clearFire();
++                    if (!entity.clearFire(io.papermc.paper.event.entity.EntityExtinguishEvent.Cause.POWDER_SNOW)) return; // Paper - add EntityExtinguishEvent
```
Lulu13022002, line 248: I think the snow block should still be destroyed for flame arrows like the burning player. Because here you cancel the extinguish of the entity and so if the entity still burn the block should melt. You can move your bugfix block one line down

patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
```diff
@@ -294,6 +318,25 @@ index f174094febfdfdc309f1b50877be60bae8a98156..5f407535298a31a34cfe114dd863fd6a
...
+@@ -188,6 +188,14 @@ public abstract class AbstractArrow extends Projectile {
+         }
+ 
+         if (this.isInWaterOrRain() || iblockdata.is(Blocks.POWDER_SNOW)) {
++            // Paper start - fix MC-214647 & MC-245680
++            if (this.getRemainingFireTicks() > 0) {
++                this.playEntityOnFireExtinguishedSound();
++
++                if (iblockdata.is(Blocks.POWDER_SNOW))
++                    this.level().destroyBlock(blockposition, false);
```
kennytv, line 334: I think this should also fire the EntityChangeBlockEvent

PR #9611, opened 2023-08-15
Do crystal-portal proximity check before entity lookup

MartijnMuijsers:
This adds a very cheap distance check when an end crystal is placed.

Attempting to respawn the dragon, which involves looking up the end crystal entities near the portal, every time an end crystal is placed, can be slow on some servers that have players placing end crystals as a style of combat.

The very cheap distance check prevents running the entity lookup every time.

### DIFF
--- a/patches/server/1029-Do-crystal-portal-proximity-check-before-entity-look.patch
+++ b/patches/server/1029-Do-crystal-portal-proximity-check-before-entity-look.patch
@@ -1,0 +1,75 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Martijn Muijsers <martijnmuijsers@live.nl>
+Date: Tue, 15 Aug 2023 21:04:55 +0200
+Subject: [PATCH] Do crystal-portal proximity check before entity lookup
+
+This adds a very cheap distance check when an end crystal is placed.
+
+Attempting to respawn the dragon, which involves looking up the end crystal
+entities near the portal, every time an end crystal is placed, can be slow on
+some servers that have players placing end crystals as a style of combat.
+
+The very cheap distance check prevents running the entity lookup every time.
+
+diff --git a/src/main/java/net/minecraft/world/item/EndCrystalItem.java b/src/main/java/net/minecraft/world/item/EndCrystalItem.java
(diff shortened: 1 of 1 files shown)


Machine-Maker (approved):
Didn't realize it was so generous in detecting where you placed crystals, always thought it was just those exact blocks.

MartijnMuijsers:
> Didn't realize it was so generous in detecting where you placed crystals, always thought it was just those exact blocks.

Yeah I was surprised too. Normally you can't place crystals in the other positions (one closer to the portal, which is above a portal block; or one below, which would be inside bedrock or a portal block), but if you do place them there (by modifying the portal bedrock/portal blocks in advance) it does work. The vanilla code checks for an entity in a bounding box that just barely touches those normally-impossible locations.

patches/server/1005-Do-crystal-portal-proximity-check-before-entity-look.patch
```diff
@@ -0,0 +1,66 @@
...
+diff --git a/src/main/java/net/minecraft/world/item/EndCrystalItem.java b/src/main/java/net/minecraft/world/item/EndCrystalItem.java
+index ca1edc083847b47bb450b291723aca778a5912dc..fbe6cf2bd05a9bb4953d1110fbead842c5371fab 100644
+--- a/src/main/java/net/minecraft/world/item/EndCrystalItem.java
++++ b/src/main/java/net/minecraft/world/item/EndCrystalItem.java
+@@ -56,7 +56,7 @@ public class EndCrystalItem extends Item {
+                         EndDragonFight enderdragonbattle = ((ServerLevel) world).getDragonFight();
+ 
+                         if (enderdragonbattle != null) {
+-                            enderdragonbattle.tryRespawn();
++                            enderdragonbattle.tryRespawn(blockposition1); // Paper - pass placed end crystal position to pre-check proximity to portal
```
kennytv: blockposition1 should get an obfuscation helper

kennytv (approved):

patches/server/1005-Do-crystal-portal-proximity-check-before-entity-look.patch
MartijnMuijsers: Rebased and added an obfuscation helper variable. Is it added the right way?

PR #9614, opened 2023-08-16
Add SpawnerPlayerSearchEvent

TreemanKing:
Hopefully second try is the charm. With the feedback of lynxplay from my previous pull, I created a new Event which would act before the Spawner searches for nearby spawners.

This would:
a) allow to remove specific players from specific spawners in regards to their activation
b) allow spawners to skip player searching for n ticks (as specified by the backoff cooldown)
c) completely exclude players from affecting spawning on spawners.

I was also going to add NotNulls to the event but I wasn't sure so your feedback would be highly appreciated!

### DIFF
--- a/patches/api/0484-Add-SpawnerPlayerSearchEvent.patch
+++ b/patches/api/0484-Add-SpawnerPlayerSearchEvent.patch
@@ -1,0 +1,135 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TreemanKing <67459602+TreemanKing@users.noreply.github.com>
+Date: Wed, 16 Aug 2023 11:59:08 +1000
+Subject: [PATCH] Add SpawnerPlayerSearchEvent
+
+This patch introduces a new event, SpawnerPlayerSearchEvent. This event is designed to be fired when a mob spawner searches for nearby players to activate. It provides the ability to control and customize the activation logic of mob spawners, preventing them from searching and spawning mobs if desired
+
+diff --git a/src/main/java/io/papermc/paper/event/block/SpawnerPlayerSearchEvent.java b/src/main/java/io/papermc/paper/event/block/SpawnerPlayerSearchEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..3ab3ebdfc560aafd043bef23006ee19d8cd67ed3
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/block/SpawnerPlayerSearchEvent.java
+@@ -0,0 +1,122 @@
++package io.papermc.paper.event.block;
++
++import com.google.common.base.Preconditions;
++import org.bukkit.Location;
++import org.bukkit.entity.Player;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Range;
++import org.jetbrains.annotations.Unmodifiable;
++import java.util.Collection;
++import java.util.Collections;
++
++/**
++ * Represents an event that is fired when a mob spawner searches for nearby players to activate.
++ * This event occurs before any of the spawner logic, allowing cancellation to prevent the spawner 
++ * from searching and spawning any mobs. It can also bypass the nearby player check if required.
++ */
++public class SpawnerPlayerSearchEvent extends Event {
++    @NotNull private final Location spawnerLocation;
++    @NotNull private final Collection<Player> activatingPlayers;
++    private int backoffTicks;
++    @NotNull private Result searchResult = Result.DEFAULT;
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    @ApiStatus.Internal
++    public SpawnerPlayerSearchEvent(@NotNull Location spawnerLocation, @NotNull Collection<Player> activatingPlayers, int backoffTicks) {
++        this.spawnerLocation = Preconditions.checkNotNull(spawnerLocation, "Spawner Location can not be null");
++        this.activatingPlayers = Preconditions.checkNotNull(activatingPlayers, "Activating players list cannot be null");
++        this.backoffTicks = backoffTicks;
++    }
++
++    /**
++     * Retrieves the location of the mob spawner associated with the event
++     *
++     * @return The location of the spawner
++     */
++    @NotNull
++    public Location getSpawnerLocation() {
++        return spawnerLocation.clone();
++    }
++
++    /**
++     * Retrieves an immutable collection of players who are close enough to activate the spawner
++     *
++     * @return A collection of activating players
++     */
++    @NotNull @Unmodifiable
++    public Collection<Player> getActivatingPlayers() {
++        return Collections.unmodifiableCollection(activatingPlayers);
++    }
++
++    /**
++     * Retrieves the number of backoff ticks before the mob spawner proceeds with spawner logic
++     *
++     * <p>Backoff ticks determine the duration the spawner pauses for before it proceeds with the spawner logic
++     * and searches for players. A higher value means a longer delay before the spawner attempts to search for
++     * activating players again. It's important to note that negative backoff ticks do not make sense and will be
++     * corrected to 0 automatically.</p>
++     *
++     * <p>While backoff ticks are active, the spawner suspends the spawner logic and hence the search for players.
++     * This means it will not trigger the spawning of monsters. This effectively postpones the activation of the spawner until
++     * the backoff ticks expire.</p>
++     *
++     * @return The amount of backoff ticks before the spawner proceeds with spawner logic
++     */
++    @Range(from = 0, to = Integer.MAX_VALUE)
++    public int getBackoffTicks() {
++        return backoffTicks;
++    }
++
++    /**
++     * Sets the number of backoff ticks before the mob spawner proceeds with the spawner's logic again.
++     * <p>{@link SpawnerPlayerSearchEvent#getBackoffTicks} on an explanation of how backoff ticks function<p>
++     *
++     * @param ticks sets the amount of ticks before the spawner proceeds with the spawner's logic
++     * @throws IllegalArgumentException If the provided backoff ticks value is negative.
++     * @see SpawnerPlayerSearchEvent#getBackoffTicks
++     */
++    public void setBackoffTicks(@Range(from = 0, to = Integer.MAX_VALUE) int ticks) {
++        Preconditions.checkArgument(ticks >= 0, "Backoff Ticks cannot be less than 0");
++        this.backoffTicks = ticks;
++    }
++
++    /**
++     * Gets the result of the spawner event.
++     * <p>See {@link SpawnerPlayerSearchEvent#setSearchResult(Result)} for a description of each result</p>
++     *
++     * @return The result of the event
++     * @see SpawnerPlayerSearchEvent#setSearchResult(Result)
++     */
++    public @NotNull Result getSearchResult() {
++        return this.searchResult;
++    }
++
++    /**
++     * Sets the result of the spawner search event
++     * <ul>
++     *          <li>{@link org.bukkit.event.Event.Result#ALLOW}: Proceed with spawner logic, regardless of activating players </li>
++     *          <li>{@link org.bukkit.event.Event.Result#DENY}: Skip spawner logic, even with activating players</li>
++     *          <li>{@link org.bukkit.event.Event.Result#DEFAULT}: Use default behaviour based on activating players </li>
++     * </ul>
++     *
++     * @param result The result to set for the event
++     */
++    public void setSearchResult(@NotNull org.bukkit.event.Event.Result result) {
++        Preconditions.checkArgument(result != null, "Result cannot be null");
++        this.searchResult = result;
++    }
++
++    @NotNull
++    public HandlerList getHandlers() {
++        return HANDLER_LIST;
++    }
++
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return HANDLER_LIST;
++    }
++}
++
--- a/patches/server/1020-Optimise-nearby-player-retrieval.patch
+++ b/patches/server/1020-Optimise-nearby-player-retrieval.patch
@@ -8,10 +8,10 @@ we can instead use the nearby player tracking system to reduce
 the number of tests per search.
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index d4376ed215d97066a21e462fae2a0e25ad8a16a1..aab652174a8175765cad548f7c61ce353ca74803 100644
+index d4376ed215d97066a21e462fae2a0e25ad8a16a1..9755ad4b7ab94e94b7765fcb88bb21273ae64eb4 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
-@@ -581,6 +581,115 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -581,6 +581,157 @@ public class ServerLevel extends Level implements WorldGenLevel {
          this.lagCompensationTick = (System.nanoTime() - net.minecraft.server.MinecraftServer.SERVER_INIT) / (java.util.concurrent.TimeUnit.MILLISECONDS.toNanos(50L));
      }
      // Paper end - lag compensation
@@ -21,6 +21,48 @@ index d4376ed215d97066a21e462fae2a0e25ad8a16a1..aab652174a8175765cad548f7c61ce35
 +                                                                                     net.minecraft.world.entity.LivingEntity entity,
 +                                                                                     net.minecraft.world.phys.AABB box) {
 +        return this.getNearbyEntities(Player.class, targetPredicate, entity, box);
++    }
++
++    public List<ServerPlayer> getNearbyPlayers(double sourceX, double sourceY, double sourceZ, double maxDistance, @Nullable Predicate<Entity> targetPredicate) {
++
++        List<ServerPlayer> nearbyPlayers = new ArrayList<>();
++
++        if (maxDistance > 0) {
++
++            io.papermc.paper.util.player.NearbyPlayers players = this.chunkSource.chunkMap.getNearbyPlayers();
++
++            com.destroystokyo.paper.util.maplist.ReferenceList<ServerPlayer> nearby = players.getPlayersByBlock(
++                io.papermc.paper.util.CoordinateUtils.getBlockCoordinate(sourceX),
++                io.papermc.paper.util.CoordinateUtils.getBlockCoordinate(sourceZ),
++                io.papermc.paper.util.player.NearbyPlayers.NearbyMapType.GENERAL
++            );
++
++            if (nearby == null) return nearbyPlayers;
++
++            double maxDistanceSqr = maxDistance * maxDistance;
++            Object[] rawData = nearby.getRawData();
++            for (int i = 0, len = nearby.size(); i < len; ++i) {
++                ServerPlayer player = (ServerPlayer) rawData[i];
++                double distanceSquared = player.distanceToSqr(sourceX, sourceY, sourceZ);
++                if (distanceSquared > maxDistanceSqr) {
++                    continue;
++                }
++
++                if (targetPredicate == null || targetPredicate.test(player)) {
++                    nearbyPlayers.add(player);
++                }
++            }
++
++            return nearbyPlayers;
++        } else {
++            for (ServerPlayer player : this.players()) {
++                if (targetPredicate == null || targetPredicate.test(player)) {
++                    nearbyPlayers.add(player);
++                }
++            }
++
++            return nearbyPlayers;
++        }
 +    }
 +
 +    @Override
--- a/patches/server/1029-Write-SavedData-IO-async.patch
+++ b/patches/server/1029-Write-SavedData-IO-async.patch
@@ -24,10 +24,10 @@ index 36caf354634d6675a3f1ec6829f4778e1d0623bc..b99f50604bafecbc68835974c9ed0caa
  
      // CraftBukkit start - modelled on below
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index aab652174a8175765cad548f7c61ce353ca74803..ca56a0b596976448da6bb2a0e82b3d5cd4133e12 100644
+index 9755ad4b7ab94e94b7765fcb88bb21273ae64eb4..b4c3fb8d8ec7cdb2548da0b4dc50eeab27e4b760 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
-@@ -1492,7 +1492,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -1534,7 +1534,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
  
          try (co.aikar.timings.Timing ignored = this.timings.worldSave.startTiming()) {
              if (doFull) {
@@ -36,7 +36,7 @@ index aab652174a8175765cad548f7c61ce353ca74803..ca56a0b596976448da6bb2a0e82b3d5c
              }
  
              this.timings.worldSaveChunks.startTiming(); // Paper
-@@ -1528,7 +1528,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -1570,7 +1570,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
                  progressListener.progressStartNoAbort(Component.translatable("menu.savingLevel"));
              }
  
@@ -45,7 +45,7 @@ index aab652174a8175765cad548f7c61ce353ca74803..ca56a0b596976448da6bb2a0e82b3d5c
              if (progressListener != null) {
                  progressListener.progressStage(Component.translatable("menu.savingChunks"));
              }
-@@ -1551,12 +1551,12 @@ public class ServerLevel extends Level implements WorldGenLevel {
+@@ -1593,12 +1593,12 @@ public class ServerLevel extends Level implements WorldGenLevel {
          // CraftBukkit end
      }
  
--- a/patches/server/1053-feat-Add-SpawnerPlayerSearchEvent.patch
+++ b/patches/server/1053-feat-Add-SpawnerPlayerSearchEvent.patch
@@ -1,0 +1,62 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TreemanK <tristankrst@gmail.com>
+Date: Tue, 4 Jun 2024 14:17:55 +1000
+Subject: [PATCH] feat: Add SpawnerPlayerSearchEvent
+
+
+diff --git a/src/main/java/net/minecraft/world/level/BaseSpawner.java b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+index f57e1b78204dff661ad5d3ee93a88a00330af2dc..cdf261955d69220ec9f1b12c297030b0f7b0940f 100644
+--- a/src/main/java/net/minecraft/world/level/BaseSpawner.java
++++ b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+@@ -59,7 +59,7 @@ public abstract class BaseSpawner {
+     }
+ 
+     public boolean isNearPlayer(Level world, BlockPos pos) {
+-        return world.hasNearbyAlivePlayerThatAffectsSpawning((double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange); // Paper - Affects Spawning API
++        return world.hasNearbyAlivePlayerThatAffectsSpawning((double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange); // Paper - Affects Spawning API // diff on change (activatingPlayers)
+     }
+ 
+     public void clientTick(Level world, BlockPos pos) {
+@@ -87,10 +87,41 @@ public abstract class BaseSpawner {
+         if (spawnCount <= 0 || maxNearbyEntities <= 0) return; // Paper - Ignore impossible spawn tick
+         // Paper start - Configurable mob spawner tick rate
+         if (spawnDelay > 0 && --tickDelay > 0) return;
++
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end - Configurable mob spawner tick rate
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - SpawnerSearchPlayerEvent
++        java.util.function.Supplier<java.util.Collection<org.bukkit.entity.Player>> lazyActivatingPlayersSup = com.google.common.base.Suppliers.memoize(() ->
++            com.google.common.collect.Collections2.transform(
++                world.getNearbyPlayers(
++                    (double) pos.getX() + 0.5D,
++                    (double) pos.getY() + 0.5D,
++                    (double) pos.getZ() + 0.5D,
++                    (double) this.requiredPlayerRange,
++                    net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING),
++                net.minecraft.server.level.ServerPlayer::getBukkitEntity)
++        );
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = new com.google.common.collect.ForwardingCollection<>() {
++            @Override
++            protected java.util.Collection<org.bukkit.entity.Player> delegate() {
++                return lazyActivatingPlayersSup.get();
++            }
++        };
++
++        io.papermc.paper.event.block.SpawnerPlayerSearchEvent spawnerPlayerSearchEvent =
++            new io.papermc.paper.event.block.SpawnerPlayerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, tickDelay));
++
++        spawnerPlayerSearchEvent.callEvent();
++        org.bukkit.event.Event.Result result = spawnerPlayerSearchEvent.getSearchResult();
++        tickDelay = Math.max(tickDelay, spawnerPlayerSearchEvent.getBackoffTicks());
++
++        if (result == org.bukkit.event.Event.Result.DENY) return;
++
++        if (result == org.bukkit.event.Event.Result.ALLOW || !activatingPlayers.isEmpty()) {
++        // Paper end - SpawnerSearchPlayerEvent
+             if (this.spawnDelay < -tickDelay) { // Paper - Configurable mob spawner tick rate
+                 this.delay(world, pos);
+             }


patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,79 @@
...
+Subject: [PATCH] PreSpawnerSearchEvent
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/entity/PreSpawnerSearchEvent.java b/src/main/java/com/destroystokyo/paper/event/entity/PreSpawnerSearchEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..24c7c05800386f5ad37d626c289696e4853760d1
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/entity/PreSpawnerSearchEvent.java
+@@ -0,0 +1,67 @@
++package com.destroystokyo.paper.event.entity;
```
Warriorrrr: New events should go into the io.papermc.paper package

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++
++import java.util.List;
++
++    public class PreSpawnerSearchEvent extends Event implements Cancellable {
++
++        private final Location spawnerLocation;
++        private final List<Player> activatingPlayers;
++        private int backoffTicks;
++        private boolean cancelled = false;
++        private static final HandlerList handlers = new HandlerList();
```
Warriorrrr: static final fields should be upper snake case, HANDLER_LIST

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++
++        public PreSpawnerSearchEvent(Location spawnerLocation, List<Player> activatingPlayers, int backoffTicks) {
```
Warriorrrr: I think it might be nicer to have a Collection<Player> instead of a List, you can also annotate this constructor with a `@APIStatus.Internal`

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++            this.spawnerLocation = spawnerLocation;
++            this.activatingPlayers = activatingPlayers;
++            this.backoffTicks = backoffTicks;
++        }
++
++        /**
++         * @return The spawner's location
++         */
++        public Location getSpawnerLocation() { return spawnerLocation; }
```
Warriorrrr: These getters/setters should be on multiple lines instead of one to be more in line with everywhere else.

patches/server/1006-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,66 @@
...
+ import net.minecraft.util.random.WeightedEntry;
+ import net.minecraft.world.Difficulty;
+-import net.minecraft.world.entity.Entity;
+-import net.minecraft.world.entity.EntityType;
+-import net.minecraft.world.entity.Mob;
+-import net.minecraft.world.entity.MobSpawnType;
+-import net.minecraft.world.entity.SpawnGroupData;
+-import net.minecraft.world.entity.SpawnPlacements;
++import net.minecraft.world.entity.*;
++import net.minecraft.world.entity.player.Player;
```
Warriorrrr: This star import should be undone. Also use fully qualified names to avoid imports & reduce diff

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++
++        /**
++         * @return The player who is activating the player
```
Warriorrrr:
```suggestion
+         * @return A collection of players who are close enough to activate the spawner.
```

patches/server/1006-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,66 @@
...
++        com.destroystokyo.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent;
++        preSpawnerSearchEvent = new com.destroystokyo.paper.event.entity.PreSpawnerSearchEvent(
++            io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, searchCooldownTicks);
++
++        searchCooldownTicks = preSpawnerSearchEvent.getBackoffTicks();
++
++        if (preSpawnerSearchEvent.isCancelled()) return;
++        // Paper end
++
+         if (this.isNearPlayer(world, pos)) {
```
Warriorrrr: Is it intended that this still checks for nearbyPlayers instead of using activatingPlayers?

Warriorrrr (changes requested):
PR looks good aside from a few small comments.

patches/server/1006-PreSpawnerSearchEvent.patch
TreemanKing: Not sure. It wasn't what my PR was attempting to achieve but I guess this could be implemented instead of the previous? Thoughts?
```suggestion
         if (!activatingPlayers.isEmpty()) {
```

patches/server/1006-PreSpawnerSearchEvent.patch
TreemanKing: I feel like when I put everything back something cooked up and it isn't showing all the imports in the patch but it is in my IDE. Could someone check this?

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,96 @@
...
++import org.bukkit.entity.Player;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++import java.util.Collection;
++import java.util.List;
++
++    public class PreSpawnerSearchEvent extends Event implements Cancellable {
```
Lulu13022002: No need for this indentation

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,61 @@
...
++
++import java.util.Collection;
+ import java.util.Optional;
+ import java.util.function.Function;
+ import javax.annotation.Nullable;
+@@ -22,6 +24,7 @@ import net.minecraft.world.entity.Mob;
+ import net.minecraft.world.entity.MobSpawnType;
+ import net.minecraft.world.entity.SpawnGroupData;
+ import net.minecraft.world.entity.SpawnPlacements;
++import net.minecraft.world.entity.EntitySelector;
```
Lulu13022002: You need to use the fully qualified name to get away from these. So instead of using EntitySelector you put net.minecraft.world.entity.EntitySelector for example

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,61 @@
...
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
++
++        Collection<org.bukkit.entity.Player> activatingPlayers = (Collection)world.getNearbyPlayers(null,
++            (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double)
++                this.requiredPlayerRange, EntitySelector.PLAYER_AFFECTS_SPAWNING);
++
++        io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent;
++        preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(
++            io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, searchCooldownTicks);
```
Lulu13022002: This can be on the same line but the event doesn't seems to be called anywhere in the first place?

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,61 @@
...
+@@ -86,7 +90,24 @@ public abstract class BaseSpawner {
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
++
++        Collection<org.bukkit.entity.Player> activatingPlayers = (Collection)world.getNearbyPlayers(null,
```
Lulu13022002: You absolutely can't cast a List<net.minecraft.server.level.ServerPlayer> into a Collection<org.bukkit.entity.Player>. And this logic should mimic the isNearPlayer boolean but changing into a list

patches/server/1006-PreSpawnerSearchEvent.patch
Lulu13022002: This is fine, i don't except a getter to be mutable for a list (by default), and that wouldn't be ~useful~ here but for safety you can change the list to an unmodifiable one for the getter only.
The only missing thing is a small comment saying you copied the logic from hasNearbyAlivePlayerThatAffectsSpawning or you can keep your change to avoid two loops but in both case the list shouldn't be mutable by the plugin.

Edit: it might be interesting to have a mutable list here, you might change it to a getter/setter or just expose the final boolean in that case but that's up to you. If you keep a mutable list things should be clear in the javadoc.

Lulu13022002:
I forgot to say that in the review, but you need the notnull/nullable annotations in the API otherwise the test will fail

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: What would you recommend alternatively? Isn't it required so that "setCancelled" can be checked?

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: > I think it might be nicer to have a Collection instead of a List, you can also annotate this constructor with a `@APIStatus.Internal`

Was originally a list but followed the feedback of above. Probably either did it wrong or don't have the right methods in place for a Collection.

Lulu13022002: The problem is not really the collection but more the type ServerPlayer is internal so you can't cast directly that and if you do the right cast (Collection<org.bukkit.entity.Player> with the generic type) things won't compile

patches/server/1005-PreSpawnerSearchEvent.patch
Lulu13022002: The event doesn't seems to be called anywhere normally you do a #callEvent or Bukkit...callEvent

TreemanKing: So something like this?
`if(!preSpawnerSearchEvent.callEvent()){
if(event.setCancelled()){ return } }`

Leguan16: you dont need the second line iirc.

patches/server/1005-PreSpawnerSearchEvent.patch
Lulu13022002: This is not resolved, please don't close a conversation that is not yet done, this is already the second time, i will not unresolve stuff each day. I'm fine if you need some help however

TreemanKing: EDIT: Sorry, thought I had fixed it! Should be fine now...

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: Should only be using FQN now.

Lulu13022002: you missed one import (java.util.List)

patches/server/1005-PreSpawnerSearchEvent.patch
Lulu13022002: This is not fixed the way you convert a ServerPlayer to a Bukkit player is using #getBukkitEntity on the internal object. No cast should be needed

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,65 @@
...
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = convertServerPlayersToBukkitPlayers(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING)); // Uses logic from hasNearbyAlivePlayerThatAffectsSpawning
++
++        io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, searchCooldownTicks);
++
++        if (!preSpawnerSearchEvent.callEvent()) return;
++
++        searchCooldownTicks = preSpawnerSearchEvent.getBackoffTicks();
++        // Paper end
++
++        if (!activatingPlayers.isEmpty()) { // Paper - Use PreSpawnerSearchEvent API
```
TreemanKing:
```suggestion
+        if (!activatingPlayers.isEmpty()) { // Paper - Use PreSpawnerSearchEvent API

+            io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, searchCooldownTicks);
+
+            if (!preSpawnerSearchEvent.callEvent()) return;
+
+            searchCooldownTicks = preSpawnerSearchEvent.getBackoffTicks();
+        // Paper end
+
```

Perhaps this is better which would make activatingPlayers `@NotNull` within the event/api. Thoughts?

TreemanKing:
I believe all of the above should be resolved.

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,95 @@
...
++
++    /**
++     * @return The amount of backoff ticks before the spawner searches for players again
++     */
++    public int getBackoffTicks() {
++        return backoffTicks;
++    }
++
++    /**
++     * @param ticks sets the amount of ticks before the spawner searches again for the player
```
lynxplay: Generally a rather weird javadoc structure on the event, you are skipping the main javadoc comments on every method and just describe the method in either `@return` or `@param` flags.

Can you replace this with proper javadocs as found in the rest of the codebase, describing the method in a normal javadoc body and the specific return/param value in the tags.

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,95 @@
...
++    /**
++     * @return The spawner's location
++     */
++    @NotNull
++    public Location getSpawnerLocation() {
++        return spawnerLocation;
++    }
++
++    /**
++     * @return A collection of players who are close enough to activate the spawner
```
lynxplay: Should denote a) via annotation and b) in the javadocs that the returned collection is immutable.

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++    private int searchCooldownTicks = 0; // Paper
+ 
+     public BaseSpawner() {}
+ 
+@@ -57,6 +58,16 @@ public abstract class BaseSpawner {
+     public boolean isNearPlayer(Level world, BlockPos pos) {
+         return world.hasNearbyAlivePlayerThatAffectsSpawning((double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange); // Paper - Affects Spawning API
+     }
++    // Paper Start - PreSpawnerSearchEvent (Convert ServerPlayers into BukkitPlayer)
++    public java.util.Collection<org.bukkit.entity.Player> convertServerPlayersToBukkitPlayers(java.util.List<net.minecraft.server.level.ServerPlayer> serverPlayers) {
```
lynxplay: Can just be replaced by a `com.google.common.collect.Collections2.transform(serverPlayerList, ServerPlayer::getBukkitEntity)` in the code itself.

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
+@@ -86,7 +97,20 @@ public abstract class BaseSpawner {
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = convertServerPlayersToBukkitPlayers(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING)); // Uses logic from hasNearbyAlivePlayerThatAffectsSpawning
```
lynxplay: This needs a diff on change in MobSpawnerAbstract#isNearPlayer so that changes in there cause the patch to fail, otherwise we might miss changes to said method in an update, leaving us with then faulty copied logic.

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,95 @@
...
++     */
++    @NotNull
++    public Collection<Player> getActivatingPlayers() {
++        return Collections.unmodifiableCollection(activatingPlayers);
++    }
++
++    /**
++     * @return The amount of backoff ticks before the spawner searches for players again
```
lynxplay: This method should rigorously describe how backoff ticks work. setBackoffTicks should also `@see getBackoffTicks` to link to said explination.

Right now backoff ticks literally mean the spawners does nothing for the next n ticks, which needs to be documented as search backoff could also just mean the spawners simply does not search for the next n ticks but still keeps spawning if it found players for this event.

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,95 @@
...
++import org.bukkit.Location;
++import org.bukkit.entity.Player;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++import java.util.Collection;
++import java.util.Collections;
++
++public class PreSpawnerSearchEvent extends Event implements Cancellable {
```
lynxplay: Missing javadocs on when and why the event is fired.

patches/server/1005-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
+     public void clientTick(Level world, BlockPos pos) {
+         if (!this.isNearPlayer(world, pos)) {
+@@ -86,7 +97,20 @@ public abstract class BaseSpawner {
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
```
lynxplay: We might want some input on whether this is the behaviour we want backoff ticks to have or if they should simply *just* skip the player search and use the "boolean" if players were nearby from the last attempt before the backoff.

patches/api/0426-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,95 @@
...
++     */
++    public int getBackoffTicks() {
++        return backoffTicks;
++    }
++
++    /**
++     * @param ticks sets the amount of ticks before the spawner searches again for the player
++     */
++    public void setBackoffTicks(int ticks) {
++        this.backoffTicks = ticks;
```
lynxplay: Negative backoff ticks make no sense. This could use a `@Range` annotation as well as obvious javadocs and a runtime validation via Preconditions.

patches/server/1005-PreSpawnerSearchEvent.patch
lynxplay: Additionally, letting that field dive into the negative numbers might not be the smartest idea ? Obviously underflow is far away given this is an int, but it is kind of bleh state nonetheless.

lynxplay (changes requested):

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: Do you mean getNearbyPlayers? Or are you referring to  putting `if (!activatingPlayers.isEmpty())`?

lynxplay: From where you copied this from (the inside of the `isNearPlayer` method in `MobSpawnerAbstract`).

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: > We might want some input on whether this is the behaviour we want backoff ticks to have or if they should simply _just_ skip the player search and use the "boolean" if players were nearby from the last attempt before the backoff.

Hmmm, that's a great point. I prefer the current functionality of it but it'd be interesting what others would have to say.

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: Not really sure how to do this. Is it by annotation or something similiar?

Also, I'm not sure why it reordered the patches the way it did.

TreemanKing: Not really sure how to do this. Is it by annotation or something similiar?

~~Also, I'm not sure why it reordered the patches the way it did.~~ Just rebased, not sure why it added the previous commits twice...

Lulu13022002: You can put a comment "// diff on change" in the isNearPlayer method to be sure to prevent desync even if this method is completely rewritten by paper. That way the diff will conflict when things change and the maintainer could update your event. Also currently the pr doesn't allow point a and c to happens so see my first comment about that (the edit) but i think the event should be called before the player check to allow a and c.

patches/server/1005-PreSpawnerSearchEvent.patch
Lulu13022002: i don't think this list was null even before your change? an empty list is not the same as a null object

patches/api/0439-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,121 @@
...
++    public Location getSpawnerLocation() {
++        return spawnerLocation;
++    }
++
++    /**
++     * Retrieves an immutable collection of players who are close enough to activate the spawner
++     * @return A collection of activating players
++     */
++    @NotNull
++    public Collection<Player> getActivatingPlayers() {
```
lynxplay: Could use a `@Unmodifiable`

patches/api/0439-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,121 @@
...
++     * means a longer delay before the spawner attempts to search for activating players again. It's important
++     * to note that negative backoff ticks do not make sense and will be corrected to 0 automatically.</p>
++     *
++     * <p>While backoff ticks are active, the spawner suspends its search for players and will not trigger
++     *  the spawning of monsters. This effectively postpones the activation of the spawner until the backoff
++     *  ticks expire.</p>
++     *
++     * @return The amount of backoff ticks before the spawner searches for players again
++     */
++    public int getBackoffTicks() {
```
lynxplay: `@Range(from = 0, to = Integer.MAX_VALUE)`

patches/api/0439-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,121 @@
...
++
++    /**
++     * Sets the number of backoff ticks before the mob spawner searches for players again.
++     *
++     * @see PreSpawnerSearchEvent#getBackoffTicks() on an explanation of how backoff ticks function
++     *
++     * @param ticks sets the amount of ticks before the spawner searches again for the player
++     * @throws IllegalArgumentException If the provided backoff ticks value is negative.
++     */
++    public void setBackoffTicks(int ticks) {
```
lynxplay: `@Range(from = 0, to = Integer.MAX_VALUE)` on the parameter.

patches/server/1027-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,40 @@
...
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING),net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
++
++        if (!activatingPlayers.isEmpty()) { // Paper - Use PreSpawnerSearchEvent API
++
++            io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, searchCooldownTicks);
```
lynxplay:
```suggestion
+            io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, searchCooldownTicks));
```

patches/server/1027-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,40 @@
...
+@@ -86,7 +87,20 @@ public abstract class BaseSpawner {
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (--searchCooldownTicks > 0) return;
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING),net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
```
lynxplay:
```suggestion
+        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING), net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
```

lynxplay (changes requested):
Generally really nice work 👍 a few annotations left but beyond that good and ready for testing/review.

patches/api/0439-PreSpawnerSearchEvent.patch
TreemanKing: Annotations aren't allowed for Void functions.

patches/server/1005-PreSpawnerSearchEvent.patch
TreemanKing: The list is never null though? I'm somewhat confused regarding this comment.

patches/server/1006-PreSpawnerSearchEvent.patch
TreemanKing: Basically what you mean is to allow for the list to be able to add/remove players (basically allow spawners to be activated at any range?). 

In my opinion I don't think this is a desirable outcome as basically it will allow spawner's to functionally be always active which in my case would be opposite of the PR (if it checks the activatingPlayers).

Although, I could add something like the following (not sure if this is what you meant!), although I don't think it would be great.
```java
private List<Player> activatingPlayers = new ArrayList<>();
private boolean activatingPlayersModified = false;

public Collection<Player> getActivatingPlayers() {
    return Collections.unmodifiableList(activatingPlayers);
}

public void addPlayer(Player player) {
    activatingPlayers.add(player);
    activatingPlayersModified = true;
}

public void setActivatingPlayers(Collection<Player> activatingPlayers){
  //whatever logic to be implemented
}

public boolean hasActivatingPlayersBeenModified() {
    return activatingPlayersModified;
}

File: patches/api/0439-PreSpawnerSearchEvent.patch (hunk 23 above)
Comment (Lulu13022002) [reply patches/api/0439-PreSpawnerSearchEvent.patch]: The annotation should go on the int parameter but i wonder if zero really makes sense?

File: patches/server/1006-PreSpawnerSearchEvent.patch (hunk 7 above)
Comment (Lulu13022002) [reply patches/server/1006-PreSpawnerSearchEvent.patch]: You don't really need addPlayer here and the boolean but the goal is to allow a and c otherwise how you would do?
For your point c you could have a #setResult([Event.Result](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/Event.Result.html)) / e.getResult() that would bypass the player check when not set to default for example.

File: patches/server/1005-PreSpawnerSearchEvent.patch (hunk 12 above)
Comment (Lulu13022002) [reply patches/server/1005-PreSpawnerSearchEvent.patch]: what i mean is that even before this commit (https://github.com/PaperMC/Paper/pull/9614/commits/dd7964fcbe11e05c7e3099e2f8443aaa3e8a649f), the list was empty at best and that your change in this commit goes against your point a and c (can't exclude players in the check if the check is done before the event).

File: patches/server/1006-PreSpawnerSearchEvent.patch (hunk 7 above)
Comment (TreemanKing) [reply patches/server/1006-PreSpawnerSearchEvent.patch]: So basically I won't need the cancellable in this case if I have set/getResult is what I'm understanding?
EDIT: I understand now and I've changed it to the tristate intstead and removing the Cancellable . I believe this fixes a and c.

~~Additionally I don't know why the sync added all the other files. I tried reverting but it just added more files. Is there anyway to just remove them from this PR?~~

Edit: I think I fixed it? Still wrapping my head around this (rebasing)

File: patches/api/0426-PreSpawnerSearchEvent.patch (hunk 17 above)
Comment (TreemanKing) [reply patches/api/0426-PreSpawnerSearchEvent.patch]: Should be fixed.

File: patches/api/0440-PreSpawnerSearchEvent.patch [hunk 26]
```diff
@@ -0,0 +1,138 @@
...
++/**
++ * Represents an event that is fired before a mob spawner attempts to search
++ * for nearby players to activate. This event occurs before any of the spawner logic,
++ * allowing cancellation to prevent the spawner from searching and spawning any mobs.
++ */
++public class PreSpawnerSearchEvent extends Event {
++    @NotNull private final Location spawnerLocation;
++    @NotNull private final Collection<Player> activatingPlayers;
++    @Range(from = 0, to = Integer.MAX_VALUE) private int backoffTicks;
++    @NotNull private static PreSpawnerSearchResult SEARCH_RESULT = PreSpawnerSearchResult.DEFAULT;
```
lynxplay: Why is this static ?

TreemanKing: Fixed. Referenced wrong thing.

patches/api/0440-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,139 @@
...
++/**
++ * Represents an event that is fired before a mob spawner attempts to search
++ * for nearby players to activate. This event occurs before any of the spawner logic,
++ * allowing cancellation to prevent the spawner from searching and spawning any mobs.
++ */
++public class PreSpawnerSearchEvent extends Event {
++    @NotNull private final Location spawnerLocation;
++    @NotNull private final Collection<Player> activatingPlayers;
++    @Range(from = 0, to = Integer.MAX_VALUE) private int backoffTicks;
++    @NotNull private Result SEARCH_RESULT = Result.DEFAULT;
```
lynxplay: `searchResult` not SEARCH_RESULT

patches/api/0440-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,139 @@
...
++        this.backoffTicks = backoffTicks;
++    }
++
++    /**
++     * Retrieves the location of the mob spawner associated with the event
++     * @return The location of the spawner
++     */
++    @NotNull
++    public Location getSpawnerLocation() {
++        return spawnerLocation;
```
lynxplay: we should probably clone here. Mutation of this variable via plugins is rather undesirable.

patches/api/0440-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,139 @@
...
++    public HandlerList getHandlers() {
++        return HANDLER_LIST;
++    }
++
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return HANDLER_LIST;
++    }
++
++
```
lynxplay: Useless new line

lynxplay (commented):
Looks good to me, I'll test the PR later today when I make it back home.
The comments can be resolved by the merger or you in time, but should not prevent the PR from being merged 👍

patches/server/1006-PreSpawnerSearchEvent.patch
Lulu13022002: Yes, this looks good

patches/api/0440-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,131 @@
...
++/**
++ * Represents an event that is fired before a mob spawner attempts to search
++ * for nearby players to activate. This event occurs before any of the spawner logic,
++ * allowing cancellation to prevent the spawner from searching and spawning any mobs.
++ */
++public class PreSpawnerSearchEvent extends Event {
++    @NotNull private final Location spawnerLocation;
++    @NotNull private final Collection<Player> activatingPlayers;
++    @Range(from = 0, to = Integer.MAX_VALUE) private int backoffTicks;
```
Lulu13022002: The range annotation is only needed on the getter/setter

patches/api/0440-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,131 @@
...
++    @NotNull private Result searchResult = Result.DEFAULT;
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    @ApiStatus.Internal
++    public PreSpawnerSearchEvent(@NotNull Location spawnerLocation, @NotNull Collection<Player> activatingPlayers, @Range(from = 0, to = Integer.MAX_VALUE) int backoffTicks) {
++        this.spawnerLocation = Preconditions.checkNotNull(spawnerLocation, "Spawner Location can not be null");
++        this.activatingPlayers = Preconditions.checkNotNull(activatingPlayers, "There must be a activating player");
```
Lulu13022002: The message is not relevant anymore since it's totally fine if there's no activating player. probably you can just copy the above line pattern.

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++
++        io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, searchCooldownTicks));
++
++        if (!preSpawnerSearchEvent.callEvent()) return;
++
++        org.bukkit.event.Event.Result result = preSpawnerSearchEvent.getSearchResult();
++        activatingPlayers = preSpawnerSearchEvent.getActivatingPlayers();
++        searchCooldownTicks = preSpawnerSearchEvent.getBackoffTicks();
++
++        if(result == org.bukkit.event.Event.Result.DENY) return;
```
Lulu13022002: Missing a space between the `if` and `(`

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++        }
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING), net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
++
++        io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, searchCooldownTicks));
++
++        if (!preSpawnerSearchEvent.callEvent()) return;
++
++        org.bukkit.event.Event.Result result = preSpawnerSearchEvent.getSearchResult();
++        activatingPlayers = preSpawnerSearchEvent.getActivatingPlayers();
```
Lulu13022002: This line is not needed since the list is immutable

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++        searchCooldownTicks = preSpawnerSearchEvent.getBackoffTicks();
++
++        if(result == org.bukkit.event.Event.Result.DENY) return;
++
++        if (result == org.bukkit.event.Event.Result.ALLOW || !activatingPlayers.isEmpty()) { // Paper - Use PreSpawnerSearchEvent API
```
Lulu13022002: This comment is not needed, you're already in a start/end block

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++        if (searchCooldownTicks > 0) {
++            --searchCooldownTicks;
++            return;
++        }
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING), net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
++
++        io.papermc.paper.event.entity.PreSpawnerSearchEvent preSpawnerSearchEvent = new io.papermc.paper.event.entity.PreSpawnerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, searchCooldownTicks));
++
++        if (!preSpawnerSearchEvent.callEvent()) return;
```
Lulu13022002: The event is not anymore cancellable with this way so this can just be a simple #callEvent

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - PreSpawnerSearchEvent
++        if (searchCooldownTicks > 0) {
++            --searchCooldownTicks;
++            return;
++        }
++
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers(null, (double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING), net.minecraft.server.level.ServerPlayer::getBukkitEntity); // Uses logic from getNearbyPlayer
```
Lulu13022002: The comment is not needed here since you have already a diff on change or you can just point toward the right method if you feel it's more clear with both

patches/server/1030-PreSpawnerSearchEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++    private int searchCooldownTicks = 0; // Paper
+ 
+     public BaseSpawner() {}
+ 
+@@ -54,7 +55,7 @@ public abstract class BaseSpawner {
+         this.spawnPotentials = SimpleWeightedRandomList.empty(); // CraftBukkit - SPIGOT-3496, MC-92282
+     }
+ 
+-    public boolean isNearPlayer(Level world, BlockPos pos) {
++    public boolean isNearPlayer(Level world, BlockPos pos) { // diff on change - PreSpawnerSearchEvent
```
Lulu13022002: The diff on change should be on the hasNearbyAlivePlayerThatAffectsSpawning line, you can merge both comments in a single one to avoid duplicate // Paper

patches/api/0440-PreSpawnerSearchEvent.patch
TreemanKing: Is it necessary to have a `@NotNull` in there or should I change the annotation to `@Nullable` or does it not matter?

Lulu13022002: The annotation is fine (and it matter otherwise that means the list can be null and #unmodifiableCollection on it will probably throw a NPE if that's really the case), but the message was outdated you can check with Preconditions.checkNotNull(activatingPlayers, "Activating players list can not be null"); instead

lynxplay (commented):
After testing, seems to work well.
However, I am a bit bleh about the naming of the event after thinking about it.
It isn't really pre-search, the "heavy" searching logic is still executed.
The backoff ticks can prevent that, but the event is not really pre- anything.

TreemanKing:
> After testing, seems to work well.
> However, I am a bit bleh about the naming of the event after thinking about it.
> It isn't really pre-search, the "heavy" searching logic is still executed.
> The backoff ticks can prevent that, but the event is not really pre- anything.

Yep, more I coded, the more I thought maybe it should just be called "SpawnerSearchEvent".

patches/api/0440-SpawnerSearchEvent.patch
```diff
@@ -0,0 +1,131 @@
...
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Range;
++import org.jetbrains.annotations.Unmodifiable;
++import java.util.Collection;
++import java.util.Collections;
++
++/**
++ * Represents an event that is fired before a mob spawner attempts to search
++ * for nearby players to activate. This event occurs before any of the spawner logic,
++ * allowing cancellation to prevent the spawner from searching and spawning any mobs.
```
Leguan16: Am I bind or does it actually not allow cancellation?

lynxplay: It does via setResult(DENY)

Leguan16: my bad. Thank you.

patches/api/0440-SpawnerSearchEvent.patch
```diff
@@ -0,0 +1,131 @@
...
++     * @return The location of the spawner
++     */
++    @NotNull
++    public Location getSpawnerLocation() {
++        return spawnerLocation.clone();
++    }
++
++    /**
++     * Retrieves an immutable collection of players who are close enough to activate the spawner
++     * @return A collection of activating players
```
Lulu13022002: Missing a new line between the description and the return/parameters, this also applies for getSpawnerLocation and setSearchResult

patches/api/0440-SpawnerSearchEvent.patch
```diff
@@ -0,0 +1,131 @@
...
++    /**
++     * Sets the result of the spawner search event
++     * <ul>
++     *          <li>{@link org.bukkit.event.Event.Result#ALLOW}: Proceed with spawner logic, regardless of activating players </li>
++     *          <li>{@link org.bukkit.event.Event.Result#DENY}: Skip spawner logic, even with activating players</li>
++     *          <li>{@link org.bukkit.event.Event.Result#DEFAULT}: Use default behaviour based on activating players </li>
++     * </ul>
++     * @param result The result to set for the event
++     */
++    public void setSearchResult(@NotNull org.bukkit.event.Event.Result result){
```
Lulu13022002: Missing a space at the end between `)` and `{`

patches/api/0440-PreSpawnerSearchEvent.patch
Lulu13022002: You missed the constructor

TreemanKing:
Should I move the event to Block instead of Entity as well?

Lulu13022002:
It's a little bit tricky since the base spawner is used by the spawner block but also by the minecart spawner, i don't mind that whatever you chose but i think people are more used to see a spawner block

Leguan16:
What is the general opinion on this one:
https://discord.com/channels/289587909051416579/555469074080202765/1148993218885718076
(adding `Cancelable` which defaults to `DENY`)

lynxplay:
Again, i don't know if we have a precedent for this.
Issue isn't really setCancelled(true), issue is setCancelled(false).

Does it use DEFAULT or ALLOW ?
Its a bit meh

TreemanKing:
Anything else I need to do or is this just awaiting reviews?

lynxplay:
At least from my end, I just need to find time to give this a final review and testing :+1:

TreemanKing:
Having a second thought, maybe calling it SpawnerPlayerSearchEvent is more accurate with the event. Thoughts?

patches/api/0441-SpawnerSearchEvent.patch
```diff
@@ -0,0 +1,134 @@
...
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Range;
++import org.jetbrains.annotations.Unmodifiable;
++import java.util.Collection;
++import java.util.Collections;
++
++/**
++ * Represents an event that is fired before a mob spawner attempts to search
```
lynxplay: The javadocs need updating, the event is called after players have been searched.

patches/server/1032-Optimise-nearby-player-retrieval.patch
```diff
@@ -23,6 +23,39 @@ index 47e8a6b28960883da868061ed649152ea792c184..f502b01b564bd33c449cbe621966ef40
...
++
++        if (nearby == null) return nearbyPlayers;
++
++        double rangeSquared = range * range;
++
++        Object[] rawData = nearby.getRawData();
++        for (int i = 0, len = nearby.size(); i < len; ++i) {
++            ServerPlayer player = (ServerPlayer)rawData[i];
++
++            if (range >= 0.0) {
```
lynxplay: This could just be moved to the top ye ?
if the range is negative, just yield the entire list of players this iterates over/in the world/in the server.

lynxplay (changes requested):
After some thought with MM about this, I think we have to rethink a bit how we want to handle the backoff ticks.

Mostly for the following scenario:
Plugin developer A sets the backoff ticks to 20.
Plugin developer A expects the spawner to start searching again in 20 ticks.
Server owner B set a tick delay for mob spawners in the paper world config to 20 ticks.
This means that every 20 ticks the spawner will attempt to search, only decrementing the backoff ticks every 20 ticks, leading to a total of 400 ticks before backoff ticks reaches 0.

Plugin developer A wants to avoid this, but is undable to do so because the tick delay is not exposed right now in the API so computing how long the backoff should be is not possible.

An alternative here would be to count down tickDelay and backoff ticks at the same time.
That way plugin developer A can set the backoff ticks to, lets say 20.
If the tick delay is e.g. 30, after 30 ticks the spawner would search again directly.

This way the plugin developer does not have to deal in spawner ticks but in world ticks, which should be a much more reliable/easy unit as it is not affected by the unaccessible tick delay.

patches/server/1032-Optimise-nearby-player-retrieval.patch
TreemanKing: Yup, I've changed it to replicate the original getNearestPlayer function.

Leguan16:
does the Spawner search Players? Might rename it to `SpawnerSearchPlayerEvent`?

patches/api/0446-Add-SpawnerPlayerSearchEvent.patch
```diff
@@ -0,0 +1,135 @@
...
++    /**
++     * Gets the result of the spawner event.
++     * <p>See {@link SpawnerPlayerSearchEvent#setSearchResult(Result)} for a description of each result</p>
++     *
++     * @return The result of the event
++     * @see SpawnerPlayerSearchEvent#setSearchResult(Result)
++     */
++    @NotNull
++    public Result getSearchResult() {
++        return Preconditions.checkNotNull(this.searchResult, "Result cannot be null");
```
lynxplay: This is not really needed, we are already checking the value to not be null in the setter.

patches/api/0446-Add-SpawnerPlayerSearchEvent.patch
```diff
@@ -0,0 +1,135 @@
...
++     * <ul>
++     *          <li>{@link org.bukkit.event.Event.Result#ALLOW}: Proceed with spawner logic, regardless of activating players </li>
++     *          <li>{@link org.bukkit.event.Event.Result#DENY}: Skip spawner logic, even with activating players</li>
++     *          <li>{@link org.bukkit.event.Event.Result#DEFAULT}: Use default behaviour based on activating players </li>
++     * </ul>
++     *
++     * @param result The result to set for the event
++     */
++    public void setSearchResult(@NotNull org.bukkit.event.Event.Result result) {
++        this.searchResult = Preconditions.checkNotNull(result, "Result cannot be null");
```
lynxplay: This should use the checkArgument one. checkNotNull will throw a NPE if it fails which is a rather ugly exception compared to the illegal args one.

patches/server/1047-Add-SpawnerPlayerSearchEvent.patch
```diff
@@ -0,0 +1,55 @@
...
++        return world.hasNearbyAlivePlayerThatAffectsSpawning((double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange); // Paper - Affects Spawning API // diff on change (activatingPlayers)
+     }
+ 
+     public void clientTick(Level world, BlockPos pos) {
+@@ -82,11 +83,24 @@ public abstract class BaseSpawner {
+     public void serverTick(ServerLevel world, BlockPos pos) {
+         if (spawnCount <= 0 || maxNearbyEntities <= 0) return; // Paper - Ignore impossible spawn tick
+         // Paper start - Configurable mob spawner tick rate
+-        if (spawnDelay > 0 && --tickDelay > 0) return;
++        if (spawnDelay > 0 && --tickDelay > 0 && --backoffTicks > 0) return; // Paper
```
lynxplay: Maybe my brain is fired here, but this would fail to reduce the backoff ticks if the tick delay is <= 0 after its decrementation. 

If my server has a tick delay configure of `0`, `--backoffTicks > 0` is never evaluated.

patches/server/1047-Add-SpawnerPlayerSearchEvent.patch
```diff
@@ -0,0 +1,55 @@
...
+         tickDelay = world.paperConfig().tickRates.mobSpawner;
+         if (tickDelay == -1) { return; } // If disabled
+         // Paper end
+-        if (this.isNearPlayer(world, pos)) {
++
++        // Paper start - SpawnerPlayerSearchEvent
++        java.util.Collection<org.bukkit.entity.Player> activatingPlayers = com.google.common.collect.Collections2.transform(world.getNearbyPlayers((double) pos.getX() + 0.5D, (double) pos.getY() + 0.5D, (double) pos.getZ() + 0.5D, (double) this.requiredPlayerRange, net.minecraft.world.entity.EntitySelector.PLAYER_AFFECTS_SPAWNING), net.minecraft.server.level.ServerPlayer::getBukkitEntity);
++
++        io.papermc.paper.event.block.SpawnerPlayerSearchEvent spawnerPlayerSearchEvent = new io.papermc.paper.event.block.SpawnerPlayerSearchEvent(io.papermc.paper.util.MCUtil.toLocation(world, pos), activatingPlayers, Math.max(0, backoffTicks));
```
lynxplay: We might be able to use a `com.google.common.collect.ForwardingCollection` collecting in combination with a 
`com.google.common.base.Suppliers.memoize()` to only compute the collection of players when a plugin actually needs it during the event.

So e.g. (pseudo code)
```
Supplier<Collection<Player>> lazyActivatingPlayersSup = Suppliers.memo(() -> transfor(worlds.getNearby.....));
Collection<Player> activatingPlayers = new FowardingCollection() {
  delegate() {  return lazySup.get()  }
}
```

lynxplay (changes requested):

patches/server/1047-Add-SpawnerPlayerSearchEvent.patch
TreemanKing: Well if the `tickDelay` is `-1`, the spawner is disabled no?
```java
if (tickDelay == -1) { return; } // If disabled
```

patches/server/1047-Add-SpawnerPlayerSearchEvent.patch
```diff
@@ -0,0 +1,73 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/level/BaseSpawner.java b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+index 633500aefd515df5dadda3802b94079f75a03fa0..10e9596465523495637a442b2988a36cdaacd6af 100644
+--- a/src/main/java/net/minecraft/world/level/BaseSpawner.java
++++ b/src/main/java/net/minecraft/world/level/BaseSpawner.java
+@@ -46,6 +46,7 @@ public abstract class BaseSpawner {
+     public int requiredPlayerRange = 16;
+     public int spawnRange = 4;
+     private int tickDelay = 0; // Paper
++    private int backoffTicks = 0; // Paper
```
lynxplay: Machine and I discussed this yesterday but wouldn't it make sense to "merge" backoffTicks with tickDelay ?

The tickDelay variable does pretty much the exact same.
We'd save ourself a field here by `this.tickDelay = Math.max(this.tickDelay, event.getBackoffTicks())` instead.

Passing the current backoff ticks to the event seems, in retrospective, kind of useless ? given the event is only called when those reach 0, I don't think there is much value here.

PR #9615, opened 2023-08-16
Don't let player loot if reloot time is disabled

### LINKED ISSUE #9613: Players can reloot chests even when restrict-player-reloot is true
### Expected behavior

It is expected that players are unable to loot a chest more than once with the restrict-player-reloot configuration option set to true.

### Observed/Actual behavior

Players can loot a chest more than once.

### Steps/models to reproduce

Start a fresh 1.20.1 Paper server with build 119.

Set the following in paper-world-defaults.yml:
lootables:
auto-replenish: true
refresh-max: 15s
refresh-min: 15s

Log in, find a lootable chest and loot it.

Observe that after 15 seconds, you are unable to loot the chest again (it remains empty).

Upgrade to Paper build 120.

Log in and reloot the same chest repeatedly every 15 seconds.

### Plugin and Datapack List

None, I used a fresh installation to test.

### Paper version

This server is running Paper version git-Paper-120 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 20a71c5)

### Other

This happens even on build 126.  It was noticed after I updated my production Paper server from build 111 to 126.

lynxplay:
If the reloot time is disabled while reloot is restricted, the player should not be able to reloot. The previous logic was incorrect and allowed players to reloot explicitly when the reloot time was disabled.

Resolves: #9613

### DIFF
--- a/patches/server/0095-LootTable-API-Replenishable-Lootables-Feature.patch
+++ b/patches/server/0095-LootTable-API-Replenishable-Lootables-Feature.patch
@@ -240,7 +240,7 @@ index 0000000000000000000000000000000000000000..8e6dac2cef7af26ad74928eff631c182
 +}
 diff --git a/src/main/java/com/destroystokyo/paper/loottable/PaperLootableInventoryData.java b/src/main/java/com/destroystokyo/paper/loottable/PaperLootableInventoryData.java
 new file mode 100644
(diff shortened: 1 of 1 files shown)


electronicboy (approved):
No eyes or thinking rock, but, based on convos elsewhere, lgtm

Machine-Maker (approved):

PR #9625, opened 2023-08-18
Add MiniMessage motd option

### LINKED ISSUE #8230: MiniMessage support for MOTD

Warriorrrr:
Closes #8230, supersedes #8242

By default, the option will have a value of "default" which will fallback to the motd set in the server.properties. Adds a new ComponentOrDefault class which uses the already existing ComponentSerializer to (de)serialize the component.


Warriorrrr:
It might also be nice to set the motd inside the /paper reload command, currently it's only after every setMotd call. Can't update this PR until the end of the week unfortunately.

PR #9629, opened 2023-08-20
Lifecycle Event System

Machine-Maker:
A separate and minimal event system is required for lifecycle-related registration, such as commands, the future Registry Modification API and other similar resources which are either registered once on startup or registered on startup and re-registered during a `/minecraft:reload`-type reload.

~~This PR includes dummy events as none are actually added yet. The Brigadier Commands API would use this. See the Test Plugin for how these dummy events can be used.~~removed, see the examples below.

----

Lifecycle events are tied to one of 2 types, the `Plugin` instance (must be registered during a plugin's onEnable) and the `BootstrapContext` (must be registered during the `PluginBootstrap#bootstrap` call). These 2 are separate due to the nature of when certain things are registered or can be regsitered. For example, Commands are fine to be registered during `Plugin#onEnable`, but this like custom advancement triggers would have to be registered *much* earlier, so that would be in the `BootstrapContext` phase.

### Brigadier Command API
*in a class that extends `JavaPlugin`*
```java
@Override
public void onEnable() {
    // the LifecycleEventManager is where you register handlers for event types
    final LifecycleEventManager<Plugin> lifecycleManager = this.getLifecycleManager();
    // to register a handler, you need an event type, usually found in LifecycleEvents
    // here the event type is COMMANDS
    lifecycleManager.registerEventHandler(LifecycleEvents.COMMANDS, event -> {
        final Commands commands = event.registrar();
        // ensure plugin commands override
        commands.register(Commands.literal("tag")
                .executes(ctx -> {
                    ctx.getSource().getSender().sendPlainMessage("overriden command");
                    return Command.SINGLE_SUCCESS;
                })
                .build(),
            null,
            Collections.emptyList()
        );
    });
}
```
*in a class that implements `PluginBootstrap`*
```java
@Override
    final LifecycleEventManager<BootstrapContext> lifecycleManager = context.getLifecycleManager();
    lifecycleManager.registerEventHandler(LifecycleEvents.COMMANDS.newHandler(event -> {
        final Commands commands = event.registrar();
        commands.register(Commands.literal("heya")
                .then(Commands.argument("range", VanillaArgumentTypes.doubleRange())
                    .executes((ct) -> {
                        ct.getSource().getSender().sendPlainMessage(VanillaArgumentTypes.getDoubleRange(ct, "range").toString());
                        return 1;
                    })
                ).build(),
            null,
            Collections.emptyList()
        );
    // you can add a priority or set your handler to be a "monitor"
    }).priority(10));
}
```
### Registry Modification API
*in a class that implements `PluginBootstrap`*
```java
@Override
    // the LifecycleEventManager is where you register handlers for event types
    final LifecycleEventManager<BootstrapContext> lifecycleManager = context.getLifecycleManager();
    lifecycleManager.registerEventHandler(RegistryEvents.GAME_EVENT.newAdditionHandler(event -> {
        if (event.key().equals(GameEventKeys.BLOCK_OPEN)) {
            event.builder().range(event.builder().range() * 2);
        }
    }).priority(10));
    lifecycleManager.registerEventHandler(RegistryEvents.GAME_EVENT.preFreeze(), event -> {
        event.registry().register(NEW_EVENT, builder -> {
            builder.range(2);
        });
    });
}
```
You can see examples of this being used in both the [plugin bootstrap](https://github.com/Owen1212055/Paper/blob/command-recode-thing/test-plugin/src/main/java/io/papermc/testplugin/TestPluginBootstrap.java) and [plugin JavaPlugin](https://github.com/Owen1212055/Paper/blob/command-recode-thing/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java#L35).

```[tasklist]
- [x] Implement locking of registration for `BootstrapContext` events (requires https://github.com/PaperMC/Paper/pull/9589)
- [x] ~~Should all events allow priorities?~~ No
- [x] ~~Should there be Read-Only interfaces for the "monitor" event handlers?~~ No, too complex, annoying to create/use
- [x] Add javadocs
```


patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++import io.papermc.paper.plugin.configuration.PluginMeta;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Implemented by types which are considered owners
++ * of registered handlers for lifecycle events. Generally
++ * the types which implement this interface also provide
++ * a {@link LifecycleEventManager} where you can register
++ * event handlers.
```
kennytv:
```suggestion
+ * Implemented by types that are considered owners
+ * of registered handlers for lifecycle events. Generally,
+ * the types that implement this interface also provide
+ * a {@link LifecycleEventManager} where you can register
+ * event handlers.
```

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++ * @param <O> the required owner type
++ */
++@ApiStatus.Experimental
++@ApiStatus.NonExtendable
++public interface MonitorLifecycleEventHandlerConfiguration<O extends LifecycleEventOwner> extends LifecycleEventHandlerConfiguration<O> {
++
++    /**
++     * Sets this handler configuration to be considered a "monitor".
++     * These handlers will run last and should only be used by plugins
++     * to observe any changes from previously ran handlers.
```
kennytv:
```suggestion
+     * to observe changes from previously run handlers.
```

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++    /**
++     * A lifecycle event that exposes a {@link Registrar} that is
++     * reloadable.
++     *
++     * @param <R>
++     * @see RegistrarEvent
++     */
++    @ApiStatus.Experimental
++    @ApiStatus.NonExtendable
++    interface Reloadable<R extends Registrar> extends RegistrarEvent<R> {
```
kennytv: Moving this out as `ReloadableRegistrarEvent` would make its use more intuitive I think

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++package io.papermc.paper.plugin.lifecycle.event.types;
++
++import org.jetbrains.annotations.ApiStatus;
++
++/**
++ * Base type for all Lifecycle Events.
++ * <p>
++ * A Lifecycle Event generally is an event which
++ * fires during times when the older event system is not
++ * available, like during early server initialization.
```
kennytv: A bit too wordy; `Lifecycle events are generally fired where the older event system is not available, like during early server initialization.`

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++import io.papermc.paper.plugin.lifecycle.event.handler.configuration.LifecycleEventHandlerConfiguration;
++import io.papermc.paper.plugin.lifecycle.event.handler.configuration.MonitorLifecycleEventHandlerConfiguration;
++import io.papermc.paper.plugin.lifecycle.event.handler.configuration.PrioritizedLifecycleEventHandlerConfiguration;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Base type for all types of lifecycle events. Differs from
++ * {@link LifecycleEvent} which is the actual event object whereas
```
kennytv:
```suggestion
+ * {@link LifecycleEvent} which is the actual event object, whereas
```

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++     * of handlers as "monitors" which are run last. Useful
++     * if a plugin wants to only observe the changes other handlers
++     * made.
++     *
++     * @param <O> the required owner type
++     * @param <E> the event object type
++     */
++    @ApiStatus.Experimental
++    @ApiStatus.NonExtendable
++    interface Monitor<O extends LifecycleEventOwner, E extends LifecycleEvent> extends LifecycleEventType<O, E, MonitorLifecycleEventHandlerConfiguration<O>> {
```
kennytv: `Monitor` as a name feels wrong here, it's more of a Monitorable event, though that's a mouthful. Same for `Prioritized` below

patches/api/0449-Add-Lifecycle-Event-system.patch
```diff
@@ -0,0 +1,577 @@
...
++ */
++@ApiStatus.Experimental
++@ApiStatus.NonExtendable
++public interface LifecycleEventManager<O extends LifecycleEventOwner> {
++
++    default <E extends LifecycleEvent> void registerEventHandler(final @NotNull LifecycleEventType<? super O, ? extends E, ?> eventType, final @NotNull LifecycleEventHandler<? super E> eventHandler) {
++        this.registerEventHandler(eventType.newHandler(eventHandler));
++    }
++
++    void registerEventHandler(@NotNull LifecycleEventHandlerConfiguration<? super O> handlerConfiguration);
```
kennytv: This needs docs, ideally with examples

kennytv (changes requested):
In general, Ctrl+F "which" and replace it with "that" in all doc comments. "Which" is for nonessential subordinate clauses, which I don't think were used anywhere

Machine-Maker:
Ok, addressed most of your comments in [`1e29f73` (#9629)](https://github.com/PaperMC/Paper/pull/9629/commits/1e29f73edd55f00ec39b20968e788635b385d9af).

patches/api/0449-Add-Lifecycle-Event-system.patch
Machine-Maker: So you think Monitorable and Prioritizable?

kennytv: Yeah, not sure if there's better alternatives

Machine-Maker: done

kennytv (approved):

Machine-Maker:
You can see examples of this being used in both the [plugin bootstrap](https://github.com/Owen1212055/Paper/blob/command-recode-thing/test-plugin/src/main/java/io/papermc/testplugin/TestPluginBootstrap.java) and [plugin JavaPlugin](https://github.com/Owen1212055/Paper/blob/command-recode-thing/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java#L35).

electronicboy (approved):

lynxplay (approved):

PR #9636, opened 2023-08-21
Expose Hand during BlockCanBuildEvent

456dev:
useful for checking metadata / pdc of block to be placed


MiniDigger:
cant you just call the method on player?

the456gamer:
as general usage, or in this prs implementation?
because currently the hand the block is place from isn't exposed.

MiniDigger:
then it makes more sense to expose the hand, as thats the only thing you cant right now, no?

patches/server/1015-Expose-hand-during-BlockCanBuildEvent.patch
```diff
@@ -0,0 +1,32 @@
...
+index 39b8b3675ac58409e05fac07e07c8016c5280d81..b6f8b6f41bf7a4cb1a66d54775b1cc6fb3e0e5fc 100644
+--- a/src/main/java/net/minecraft/world/item/StandingAndWallBlockItem.java
++++ b/src/main/java/net/minecraft/world/item/StandingAndWallBlockItem.java
+@@ -59,7 +59,7 @@ public class StandingAndWallBlockItem extends BlockItem {
+             boolean defaultReturn = world.isUnobstructed(iblockdata1, blockposition, CollisionContext.empty());
+             org.bukkit.entity.Player player = (context.getPlayer() instanceof ServerPlayer) ? (org.bukkit.entity.Player) context.getPlayer().getBukkitEntity() : null;
+ 
+-            BlockCanBuildEvent event = new BlockCanBuildEvent(CraftBlock.at(world, blockposition), player, CraftBlockData.fromData(iblockdata1), defaultReturn);
++            BlockCanBuildEvent event = new BlockCanBuildEvent(CraftBlock.at(world, blockposition), player, CraftBlockData.fromData(iblockdata1), defaultReturn, context.getHand() == net.minecraft.world.InteractionHand.MAIN_HAND ? org.bukkit.inventory.EquipmentSlot.HAND : org.bukkit.inventory.EquipmentSlot.OFF_HAND); // Paper - expose hand
```
Machine-Maker: There is a method in CraftEquipmentSlot to convert the hand.

patches/api/0429-Expose-hand-during-BlockCanBuildEvent.patch
```diff
@@ -0,0 +1,48 @@
...
++        this(block, null, type, canBuild, org.bukkit.inventory.EquipmentSlot.HAND); // Paper - expose hand
+     }
+ 
+     /**
+@@ -37,12 +38,23 @@ public class BlockCanBuildEvent extends BlockEvent {
+      * @param type the id of the block to place
+      * @param canBuild whether we can build
+      */
+-    public BlockCanBuildEvent(@NotNull final Block block, @Nullable final Player player, @NotNull final BlockData type, final boolean canBuild) {
++    public BlockCanBuildEvent(@NotNull final Block block, @Nullable final Player player, @NotNull final BlockData type, final boolean canBuild, @NotNull final org.bukkit.inventory.EquipmentSlot hand) { // Paper - expose hand
```
Machine-Maker: Add `@ApiStatus.Internal` to the constructor. It's possible we want this to be a new constructor. Usually for most events I don't care, but its possible this event is created/fired by plugins more than others.

patches/api/0429-Expose-hand-during-BlockCanBuildEvent.patch
```diff
@@ -0,0 +1,48 @@
...
+         this.blockData = type;
++        this.hand = hand; // Paper - expose hand
+     }
++    // Paper start - expose hand
++    /**
++     * Gets the hand the player will use to place the block
++     *
++     * @return the EquipmentSlot representing the players hand.
++     */
++    public org.bukkit.inventory.EquipmentSlot getHand() {
```
Machine-Maker: Missing `@NotNull`

Machine-Maker (changes requested):

patches/api/0429-Expose-hand-during-BlockCanBuildEvent.patch
```diff
@@ -0,0 +1,48 @@
...
+     private final Player player;
++    private final org.bukkit.inventory.EquipmentSlot hand; // Paper - expose hand
+ 
+     @Deprecated
+     public BlockCanBuildEvent(@NotNull final Block block, @NotNull final BlockData type, final boolean canBuild) {
+@@ -37,12 +38,29 @@ public class BlockCanBuildEvent extends BlockEvent {
+      * @param type the id of the block to place
+      * @param canBuild whether we can build
+      */
++    @org.jetbrains.annotations.ApiStatus.Internal // Paper
```
Machine-Maker: Can just mark this as deprecated instead. Actually, also mark this as `DoNotUse` to make sure we catch any that upstream might add. 

So replace `Internal` with `Deprecated` and `DoNotUse`. Just for this method tho, the other one is good.

456dev: what about the existing deprecated one, or leave that as-is. should i also add a deprecation reason in the javadoc, or leave it

Machine-Maker: Just leave that one as is. No need to mention anything in the javadocs either

kennytv (approved):

Machine-Maker (approved):

PR #9641, opened 2023-08-23
Add configurable crop growth light levels

### LINKED ISSUE #9631: Growth Required Light Level

I would like this feature added to allow players to grow crops without having torches or another light source of 9 near.

A light level value in the configuration file where users can specifiy the minimum light level to grow crops. Setting to 0 would practically disable the requirement of a light to grow crops.

The light level of 9 is hard coded into Spigot and/or paper and is not configurable. Creating a plugin to combat this is not possible as Paper is forcing the light level of 9 to be the minimum of crop growth.

### Other

_No response_

JamieIsGeek:
Closes #9631


Machine-Maker:
So the problem with this, is that crops will break if they can't see the sky and have a light level lower than 8 (or smth close). 

There are multiple light level and conditions for crops to grow, so idk how well a single configuration value will handle those.

JamieIsGeek:
Yeah I saw something lower down in the class about crop survival which I guess could also be turned into a config option?

JamieIsGeek:
There is this which is the only othert light dependant part of the CropBlock class so I'd assuming making this configurable would fix your issue?
```java
@Override
public boolean canSurvive(BlockState state, LevelReader world, BlockPos pos) {
    return (world.getRawBrightness(pos, 0) >= 8 || world.canSeeSky(pos)) && super.canSurvive(state, world, pos);
}
```

Machine-Maker:
Yeah, that would be one potential fix. I just wonder if an event allowing plugins to customize it more finely would be better. Idk how such an event would work tho.

PR #9643, opened 2023-08-24
Provide registered listeners in PluginManager

Aeltumn:
This introduces new methods on the PluginManager that supersede the old registerEvent/registerEvents methods. These new methods return a RegisteredListener object which can be used by plugins to better handle their registered listeners. The RegisteredListener now also has an unregister() method which can be used to directly unregister that listener from all relevant handler lists instead of having to iterate over them all.

### DIFF
--- a/patches/api/0455-Provide-registered-listeners-in-the-PluginManager-an.patch
+++ b/patches/api/0455-Provide-registered-listeners-in-the-PluginManager-an.patch
@@ -1,0 +1,238 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aeltumn <daniel@goossens.ch>
+Date: Sun, 20 Aug 2023 19:39:43 +0200
+Subject: [PATCH] Provide registered listeners in the PluginManager and add
+ faster unregistration method
+
+Adds new methods that supersede the old registerEvent methods but return a RegisteredListener object. And adds an unregister() method to RegisteredListener which allows for targetted unregistration instead of unregistering from all HandlerLists which is inefficient when unregistering listeners often.
+
+diff --git a/src/main/java/org/bukkit/event/HandlerList.java b/src/main/java/org/bukkit/event/HandlerList.java
+index 2292bd460ce2be113beb4ba6b4eb19350060f01c..6d66a6293986ff4e6b195c3de8c58730f6441d51 100644
+--- a/src/main/java/org/bukkit/event/HandlerList.java
++++ b/src/main/java/org/bukkit/event/HandlerList.java
+@@ -61,6 +61,11 @@ public class HandlerList {
+             for (HandlerList h : allLists) {
+                 synchronized (h) {
+                     for (List<RegisteredListener> list : h.handlerslots.values()) {
++                        // Paper start - store handler list on registered listener
++                        for (RegisteredListener listener : list) {
++                            listener.unregisterHandlerList(h);
++                        }
++                        // Paper end - store handler list on registered listener
+                         list.clear();
+                     }
+                     h.handlers = null;
+@@ -126,6 +131,7 @@ public class HandlerList {
+             throw new IllegalStateException("This listener is already registered to priority " + listener.getPriority().toString());
+         handlers = null;
+         handlerslots.get(listener.getPriority()).add(listener);
++        listener.registerHandlerList(this); // Paper - store handler list on registered listener
+     }
+ 
+     /**
+@@ -146,6 +152,7 @@ public class HandlerList {
+      */
+     public synchronized void unregister(@NotNull RegisteredListener listener) {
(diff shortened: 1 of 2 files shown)


patches/api/0439-Provide-registered-listeners-in-the-PluginManager.patch
```diff
@@ -0,0 +1,215 @@
...
+      */
+     @org.jetbrains.annotations.ApiStatus.Experimental
+     void overridePermissionManager(@NotNull Plugin plugin, @Nullable io.papermc.paper.plugin.PermissionManager permissionManager);
++
++    /**
++     * Registers all the events in the given listener class
++     *
++     * @param listener Listener to register
++     * @param plugin Plugin to register
++     * @return All registered listeners
```
Machine-Maker: Clarify that this returns only registered listeners created from the supplier Listener, not *all* of them ever registered.

patches/api/0439-Provide-registered-listeners-in-the-PluginManager.patch
```diff
@@ -0,0 +1,215 @@
...
+      */
++    @Deprecated(forRemoval = true) // Paper - superseded by registerListener
+     public void registerEvent(@NotNull Class<? extends Event> event, @NotNull Listener listener, @NotNull EventPriority priority, @NotNull EventExecutor executor, @NotNull Plugin plugin);
+ 
+     /**
+@@ -141,6 +143,7 @@ public interface PluginManager extends io.papermc.paper.plugin.PermissionManager
+      * @param plugin Plugin to register
+      * @param ignoreCancelled Whether to pass cancelled events or not
+      */
++    @Deprecated(forRemoval = true) // Paper - superseded by registerListener
```
Machine-Maker: I think these deprecations are going too far. I don't ever see us removing these since they work fine. I think they are a better candidate for the "Obsolete" annotation.

Machine-Maker (changes requested):
Also, we generally steer clear of using "var" anywhere.

patches/api/0439-Provide-registered-listeners-in-the-PluginManager.patch
Aeltumn: Is there an annotation for Obsolete or should I just only note its obsoleteness in the documentation?

Machine-Maker: There is, it’s “ApiStatus.Obsolete” from jetbrains’ annotation library

Warriorrrr: Marking the current methods as obsolete just because of the return value seems a bit "strict" to me, imo the new methods could just be referenced in a seeAlso javadoc comment

Aeltumn:
Rebased and ready for re-review.

patches/api/0447-Provide-registered-listeners-in-the-PluginManager.patch
```diff
@@ -0,0 +1,244 @@
...
++        handlersLists.remove(list);
++    }
++
++    /**
++     * Unregisters this listener from all {@link org.bukkit.event.HandlerList} instances.
++     * This is faster than calling HandlerList.unregisterAll as this directly goes to the
++     * correct handler lists to unregister this listener.
++     */
++    public void unregister() {
++        for (var list : new java.util.HashSet<>(handlersLists)) {
```
powercasgamer:
```suggestion
+        for (org.bukkit.event.HandlerList list : new java.util.HashSet<>(handlersLists)) {
```

powercasgamer (changes requested):

powercasgamer (approved):

patches/api/0447-Provide-registered-listeners-in-the-PluginManager.patch
```diff
@@ -0,0 +1,244 @@
...
++
++    /**
++     * Registers all the events in the given listener class
++     *
++     * @param listener Listener to register
++     * @param plugin Plugin to register
++     * @return The new registered listeners, may be empty if no event handlers were found
++     */
++    @NotNull
++    public java.util.Collection<RegisteredListener> registerListeners(@NotNull Listener listener, @NotNull Plugin plugin);
```
DerEchtePilz: Naming this `registerListener` too might also be good? Provided that it only accepts one Listener and not multiple, I think naming this `registerListener` might be a good idea.
Alternatively, could this maybe accept a vararg Listener array?

Aeltumn: That's fair, I had it as `registerListeners` as it replaces `registerEvents`, but the new one still only registers one listener, even if it registers multiple events.

Aeltumn:
Rebased and cleaned up to match new guidelines.

PR #9649, opened 2023-08-25
Fix MC-255756

roro1506HD:
This PR fixes shields not playing sounds when being broken by an axe or when blocking a damage. This bug is client-sided but can easily be fixed by playing the sounds instead of the entity events, this is exactly what this PR does.


imDaniX:
Maybe it would be better to both broadcast the entity event and play the sound?

patches/server/1027-Fix-MC-255756-shield-break-block-sounds-not-playing.patch
```diff
@@ -0,0 +1,66 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Mob.java b/src/main/java/net/minecraft/world/entity/Mob.java
+index e2a25c29ec74147b3e66aa0b3deb85a8f6ee53a5..f9a274932049e10e0a73b94e7e681cd37c0aaaa2 100644
+--- a/src/main/java/net/minecraft/world/entity/Mob.java
++++ b/src/main/java/net/minecraft/world/entity/Mob.java
+@@ -26,6 +26,8 @@ import net.minecraft.resources.ResourceLocation;
+ import net.minecraft.server.level.ServerLevel;
+ import net.minecraft.server.level.ServerPlayer;
+ import net.minecraft.sounds.SoundEvent;
++import net.minecraft.sounds.SoundEvents;
++import net.minecraft.sounds.SoundSource;
```
imDaniX: Don't import - use fully qualified names instead.

roro1506HD: Oh yeah my bad, forgot to check the imports for this file

roro1506HD:
> Maybe it would be better to both broadcast the entity event and play the sound?

I chose not to send both because when mojang is gonna fix the issue on their end, we're gonna have a time where the sound is gonna be played twice

patches/server/1027-Fix-MC-255756-shield-break-block-sounds-not-playing.patch
```diff
@@ -0,0 +1,57 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+index e11d7283662834047b2ff81a2fd25a4263792deb..7e02de5ae22abeaad6607c5856e8326538716acb 100644
+--- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
++++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+@@ -1482,7 +1482,7 @@ public abstract class LivingEntity extends Entity implements Attackable {
+             boolean knockbackCancelled = this.level().paperConfig().environment.disableExplosionKnockback && source.is(DamageTypeTags.IS_EXPLOSION) && this instanceof net.minecraft.world.entity.player.Player; // Paper - Disable explosion knockback
+             if (flag1) {
+                 if (flag) {
+-                    this.level().broadcastEntityEvent(this, (byte) 29);
++                    this.level().playSound(null, this, SoundEvents.SHIELD_BLOCK, SoundSource.PLAYERS, 0.8F, 0.8F + this.level().random.nextFloat() * 0.4F); // Paper - Fixes MC-255756: Play the sound instead of the entity event
```
Machine-Maker: Part of the comment should be where exactly you got this line (which I assume is from the handleEntityEvent in LivingEntity). Also, I think that first float param should be `1.0F` for event 29.

patches/server/1027-Fix-MC-255756-shield-break-block-sounds-not-playing.patch
```diff
@@ -0,0 +1,57 @@
...
+                     this.level().broadcastDamageEvent(this, source);
+@@ -2225,7 +2225,10 @@ public abstract class LivingEntity extends Entity implements Attackable {
+ 
+             // Apply blocking code // PAIL: steal from above
+             if (event.getDamage(DamageModifier.BLOCKING) < 0) {
+-                this.level().broadcastEntityEvent(this, (byte) 29); // SPIGOT-4635 - shield damage sound
++                // Paper start - Fixes MC-255756: Play the sound instead of the entity event
++                net.minecraft.world.entity.player.Player exemptedPlayer = this instanceof net.minecraft.world.entity.player.Player player ? player : null;
++                this.level().playSound(exemptedPlayer, this, SoundEvents.SHIELD_BLOCK, SoundSource.PLAYERS, 0.8F, 0.8F + this.level().random.nextFloat() * 0.4F);
++                // Paper end
```
Machine-Maker: Why is the player blocking exempted here? broadcastEntityEvent sends it to the current player as well as nearby players.

Machine-Maker (changes requested):
This whole patch can be merged into the patch that fixes a ton of other random vanilla bugs.

patches/server/1027-Fix-MC-255756-shield-break-block-sounds-not-playing.patch
roro1506HD: Uhm, I missed the part where it sends it to the player, going to fix that on other places as well

patches/server/1027-Fix-MC-255756-shield-break-block-sounds-not-playing.patch
roro1506HD: Nice catch, sorry for not paying too much attention on this. And got you for the comment, going to add that

Owen1212055 (approved):

patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
Machine-Maker, line 1: Lets put a `// Paper - diff on change` on the specific line in LivingEntity#handleEntityEvent that you took the playing sound call from

Machine-Maker (changes requested):
I think this bypasses the `isSilent` check that exists in the regular logic.

The more I look into this specific issue, the less satisfied I am with fixing it on the server. The correct fix is a client-sided one (as illustrated in the mojira issue). There are other inconsistencies we are creating. The sound now has a set seed and has the exact same pitch for all listeners which isn't the case in vanilla. 

I think we should just not fix this as it would be too complicated to fix from the server.

PR #9658, opened 2023-08-27
Expand open container api on HumanEntity to support custom titles

Sytm:
The methods now take an optional component that overrides the title of the container.

I tested these changes with every method and every container type accepts the new overridden title just fine with the villager / merchant being the exception because the level of the villager is appended to the name

### DIFF
--- a/patches/api/0209-Add-additional-open-container-api-to-HumanEntity.patch
+++ b/patches/api/0209-Add-additional-open-container-api-to-HumanEntity.patch
@@ -3,16 +3,137 @@ From: JRoy <joshroy126@gmail.com>
 Date: Wed, 26 Aug 2020 02:11:58 -0400
 Subject: [PATCH] Add additional open container api to HumanEntity
 
+Expand existing methods to also allow for title overrides
+
+Co-authored-by: Lukas Planz <me@md5lukas.de>
 
 diff --git a/src/main/java/org/bukkit/entity/HumanEntity.java b/src/main/java/org/bukkit/entity/HumanEntity.java
-index 2557ddcc0528d4f9d811883b3ddc61148ebc3998..5ecfb98540c00da05b13bc5370debb89c52cc76f 100644
+index 2557ddcc0528d4f9d811883b3ddc61148ebc3998..48f5934f1e4e570606c7cd57e9fc349cc8763f68 100644
 --- a/src/main/java/org/bukkit/entity/HumanEntity.java
 +++ b/src/main/java/org/bukkit/entity/HumanEntity.java
-@@ -174,6 +174,92 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
+@@ -123,7 +123,26 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
+      *     opened.
+      */
      @Nullable
-     public InventoryView openMerchant(@NotNull Merchant merchant, boolean force);
+-    public InventoryView openWorkbench(@Nullable Location location, boolean force);
++    // Paper start - Add additional open container api
++    default InventoryView openWorkbench(@Nullable Location location, boolean force) {
++        return openWorkbench(location, force, null);
++    }
++
++    /**
++     * Opens an empty workbench inventory window with the player's inventory
++     * on the bottom.
++     *
++     * @param location The location to attach it to. If null, the player's
++     *     location is used.
++     * @param force If false, and there is no workbench block at the location,
++     *     no inventory will be opened and null will be returned.
++     * @param title The title to use instead of the default one
++     * @return The newly opened inventory view, or null if it could not be
++     *     opened.
++     */
++    @Nullable
++    InventoryView openWorkbench(@Nullable Location location, boolean force, @Nullable net.kyori.adventure.text.Component title);
++    // Paper end - Add additional open container api
  
-+    // Paper start - Add additional containers
+     /**
+      * Opens an empty enchanting inventory window with the player's inventory
+@@ -137,7 +156,26 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
+      *     opened.
+      */
+     @Nullable
+-    public InventoryView openEnchanting(@Nullable Location location, boolean force);
++    // Paper start - Add additional open container api
++    default InventoryView openEnchanting(@Nullable Location location, boolean force) {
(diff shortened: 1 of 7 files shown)


patches/api/0213-Add-additional-open-container-api-to-HumanEntity.patch
```diff
@@ -3,16 +3,134 @@ From: JRoy <joshroy126@gmail.com>
...
+      *
+      * @param merchant The merchant to trade with. Cannot be null.
+      * @param force whether to force the trade even if another player is trading
++     * @param name The name to use instead of the default one
+      * @return The newly opened inventory view, or null if it could not be
+      * opened.
+      */
+     @Nullable
+-    public InventoryView openMerchant(@NotNull Merchant merchant, boolean force);
++    public InventoryView openMerchant(@NotNull Merchant merchant, boolean force, @Nullable net.kyori.adventure.text.Component name);
```
Machine-Maker: Use `title` to maintain consistency.

patches/api/0213-Add-additional-open-container-api-to-HumanEntity.patch
```diff
@@ -3,16 +3,134 @@ From: JRoy <joshroy126@gmail.com>
...
++     * the force parameter for this.
++     *
++     * @param trader The merchant to trade with. Cannot be null.
++     * @param force whether to force the trade even if another player is trading
++     * @param name The name to use instead of the default one
++     * @return The newly opened inventory view, or null if it could not be
++     * opened.
++     */
++    @Nullable
++    public InventoryView openMerchant(@NotNull Villager trader, boolean force, @Nullable net.kyori.adventure.text.Component name);
```
Machine-Maker: Use `title` to maintain consistency.

patches/server/0464-Add-additional-open-container-api-to-HumanEntity.patch
```diff
@@ -3,47 +3,148 @@ From: JRoy <joshroy126@gmail.com>
...
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index e087b62f88157d57eb037feb0d39c076b286f755..8325d03b8f1f08c7ef68aaf10f2e5b8d7d35675b 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1502,7 +1502,7 @@ public class ServerPlayer extends Player {
+     }
+ 
+     @Override
+-    public OptionalInt openMenu(@Nullable MenuProvider factory) {
++    public OptionalInt openMenu(@Nullable MenuProvider factory, @Nullable net.kyori.adventure.text.Component titleOverride) { // Paper - overridable title
```
Machine-Maker: I don't like calling this `titleOverride`. Even tho its just a param name, it should be something like `defaultTitleOverride` or something to distinguish it from the actual title override just added to InventoryOpenEvent

Machine-Maker (changes requested):

patches/api/0213-Add-additional-open-container-api-to-HumanEntity.patch
Sytm: My rationale for calling this parameter name instead of title was that when you set the title override to "Test" you would see "Test - Novice" instead of just "Test".

I think it is better to add some additional JavaDoc explaining that.

Should I still rename the param?

---
Yeah nvm, setting the villager level to 0 removes the suffix

patches/server/0465-Add-additional-open-container-api-to-HumanEntity.patch
```diff
@@ -3,47 +3,155 @@ From: JRoy <joshroy126@gmail.com>
...
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index e087b62f88157d57eb037feb0d39c076b286f755..4b9319efaca5de6e60f82e35d643bc72b217c9e2 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -1502,7 +1502,7 @@ public class ServerPlayer extends Player {
+     }
+ 
+     @Override
+-    public OptionalInt openMenu(@Nullable MenuProvider factory) {
++    public OptionalInt openMenu(@Nullable MenuProvider factory, @Nullable net.kyori.adventure.text.Component defaultTitleOverride) { // Paper - overridable title
```
Machine-Maker: I think instead of this diff, you can just create new instances of SimpleMenuProvider and pass in the new Component title. That way it avoids all the diff in this methods and doesn't add yet another local var that is related to a title.

patches/server/0465-Add-additional-open-container-api-to-HumanEntity.patch
```diff
@@ -3,47 +3,155 @@ From: JRoy <joshroy126@gmail.com>
...
+@@ -455,12 +455,76 @@ public class CraftHumanEntity extends CraftLivingEntity implements HumanEntity {
+             throw new IllegalArgumentException("Can't open merchant " + merchant.toString());
+         }
+ 
++        // Paper start - overridable title
++        if (title != null) {
++            name = io.papermc.paper.adventure.PaperAdventure.asVanilla(title);
++            level = 0; // Set level to 0 to remove title suffix
++        }
++        // Paper end
```
Machine-Maker: Not a fan of this because it just overwrites what they entity might have the level set to. I didn't know the title came with a suffix. Calling it `name` then is fine since its not the full title. People can manipulate the `Merchant` if they want to change the level. (I see there's no API for that for the custom merchants, just the entities but that can be added)

Machine-Maker (changes requested):
We don't have to stick with upstream's ridiculousness in adding redundant `public` keywords to interface methods. For new methods, just don't include that.

patches/server/0465-Add-additional-open-container-api-to-HumanEntity.patch
Sytm: For the CustomMerchant that is irrelevant because they never append a suffix at all

Sytm:
Due to the maybe™ upcoming enum eradication I went ahead and removed the internal usage of the Material enum and use the blocks directly

PR #9664, opened 2023-08-28
Add ItemStack#isEmpty and related methods

Aeltumn:
Introduces a proper way to check if an ItemStack is empty just like in vanilla's code. The added isEmpty() method is even redirected into Mojang's method but an identical implementation is provided on ItemStack itself.

Also fixes an incorrect nullability annotation on getCursor.


patches/api/0440-Allow-proper-checking-of-empty-item-stacks.patch
```diff
@@ -0,0 +1,90 @@
...
+ public class ItemStack implements Cloneable, ConfigurationSerializable, Translatable, net.kyori.adventure.text.event.HoverEventSource<net.kyori.adventure.text.event.HoverEvent.ShowItem>, net.kyori.adventure.translation.Translatable { // Paper
++
++    // Paper start
++    /**
++     * An empty item stack, consists of an air material and a stack size of 0.
++     *
++     * Any item stack with a material of air or a stack size of 0 is seen
++     * as being empty by {@link ItemStack#isEmpty}.
++     */
++    public static final ItemStack EMPTY = new ItemStack();
```
kennytv: Can't do that since everything inside ItemStack is mutable

kennytv (changes requested):

patches/api/0440-Allow-proper-checking-of-empty-item-stacks.patch
Aeltumn: Good point, I guess an EMPTY static value is not necessary, it was more for vanilla parity. Now that I think about it vanilla's empty is also mutable? Well weird. 

I'll change it to an empty() method instead actually, since that should be fine even if it's mutable.

kennytv (approved):

patches/api/0440-Allow-proper-checking-of-empty-item-stacks.patch
```diff
@@ -0,0 +1,81 @@
...
+diff --git a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+index fe58058f9b5d29388d48115cc81dc48ab08c58c1..ac67a4e321f43d1ede09dafe2daa1f07de8a923f 100644
+--- a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+@@ -85,7 +85,7 @@ public class InventoryClickEvent extends InventoryInteractEvent {
+      *
+      * @return the cursor ItemStack
+      */
+-    @Nullable
++    @NotNull // Paper - fix nullability
```
Sytm: I think these two nullability fixes are a better fit for this patch: https://github.com/PaperMC/Paper/blob/b4e3b3d1dd447bac4cbf478595c1ec320bc6dd4b/patches/api/0172-Fix-Spigot-annotation-mistakes.patch

Aeltumn:
> nullIfEmpty seems weird to have as a method… why is that being added?

It's something that's more common in Kotlin code, it allows you to pick if you want to treat something as nullable or as a possibly empty item stack. But since in Java it doesn't do much I'll remove it.

patches/api/0172-Fix-Spigot-annotation-mistakes.patch
```diff
@@ -510,6 +510,19 @@ index e1123295b9511a2c610a1baf7195638f7f3e64c4..273ae8e5da0a858d3b82d1b0f5992318
...
+diff --git a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+index fe58058f9b5d29388d48115cc81dc48ab08c58c1..ac67a4e321f43d1ede09dafe2daa1f07de8a923f 100644
+--- a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+@@ -85,7 +85,7 @@ public class InventoryClickEvent extends InventoryInteractEvent {
+      *
+      * @return the cursor ItemStack
+      */
+-    @Nullable
++    @NotNull // Paper - fix nullability
```
imDaniX, line 604: I think this belongs [there](https://github.com/PaperMC/Paper/blob/b4e3b3d1dd447bac4cbf478595c1ec320bc6dd4b/patches/api/0172-Fix-Spigot-annotation-mistakes.patch).

patches/api/0172-Fix-Spigot-annotation-mistakes.patch
```diff
@@ -691,6 +704,24 @@ index f1a48eab1a357ae64545e1f1dc941c383cff8707..466d1bd7089b76f48f953e1a51c611ec
...
+     /**
+      * Get the item on the cursor of one of the viewing players.
+      *
+-     * @return The item on the player's cursor, or null if they aren't holding
+-     *     one.
++     * @return The item on the player's cursor, or an empty stack
++     * if they aren't holding one.
+      */
+-    @Nullable
++    @NotNull // Paper - fix nullability
```
imDaniX, line 825: Same here.

Machine-Maker (approved):

patches/api/0172-Fix-Spigot-annotation-mistakes.patch
Aeltumn: These changes should already be in that patch?

imDaniX: What I'm implying is that you may consider [modifying](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#modifying-patches) said patch with those annotation changes instead of having them in this one.
Not sure if this is required since the team already approved the changes tho.

Leguan16: tell me the name of the patch the changes are right now.
and tell me the name of the patch you want them to be.

Aeltumn: @Sytm already beat you to the punch, I already moved them to spigot annotation fixes patch. 😛

PR #9665, opened 2023-08-28
Add additional tags to ItemMeta obfuscation filtering

### LINKED ISSUE #9451: [Feature] Better Offuscation

Currently obfuscation hides alot of stuff (attributes, lore, item-name, item-damage, enchantments) but I want it to hide more if possible.

If possible, I want only CustomModelData to show and nothing else.

**People are able to use nbt inspectors** to see PublicBukkitValues and see if an item is unbreakable or not still.

### Other

current paper-world-defaults.yml

  obfuscation:
    items:
      hide-durability: true
      hide-itemmeta: true
      hide-itemmeta-with-visual-effects: false

new paper-world-defaults.yml?

  obfuscation:
    items:
      hide-durability: true
      hide-itemmeta: true
      hide-textures: false
      hide-itemmeta-with-visual-effects: false
      hide-other: true

Sytm:
Fixes #9451

(Is blacklist filtering even feasible like Machine said in the linked issue or would it be too much work to cover all edge cases?)

### DIFF
--- a/patches/server/0721-Hide-unnecessary-itemmeta-from-clients.patch
+++ b/patches/server/0721-Hide-unnecessary-itemmeta-from-clients.patch
@@ -33,7 +33,7 @@ index e8ca93305e8baaee3e72932b630a2d6907444593..65f36454b08d7527aca0d3e667979fd5
                                  }
                              }
 diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
-index f8738a28fb30c8fdb52ed3e6f8000144129c1f00..58186adba67a43188784efffd4bc9c8d98ac2848 100644
+index f8738a28fb30c8fdb52ed3e6f8000144129c1f00..4dc9647b9c7206b1e98554d2db0f193b0dce0a94 100644
 --- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
 +++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
 @@ -3196,7 +3196,7 @@ public abstract class LivingEntity extends Entity implements Attackable {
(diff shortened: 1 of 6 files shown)


electronicboy:
Filtering custom model data is a no go

Owen1212055 (dismissed):
Removing CustomModelData will change the look of the item for other players. That should not be removed like that.

Sytm:
That's why I deliberately put it inside the `hideItemmetaWithVisualEffects` block

electronicboy:
visual effects are generally more trivial things, idk how a lode compass appears differently to others if it has a pos or not, but, CMD is purely intended to allow people to change how items look in general using datapacks, etc; it being removed ever sounds like a stupidly niche edge-case at best

patches/server/0732-Hide-unnecessary-itemmeta-from-clients.patch
```diff
@@ -107,6 +111,8 @@ index a5cf00a41c61cc9f851781cc4f55c9bb47c2f5b6..f306c9e8f46d80d4c4fdc671e3754b3f
 +                // the position and fake the dimension
 +                tag.putString("LodestoneDimension", "paper:paper");
 +            }
++            // The id of a filled map
++            tag.remove("map");
```
NoahvdAa: What changes visually when this tag is missing?

NoahvdAa (commented):
> idk how a lode compass appears differently to others if it has a pos or not

[it just shows as spinning](https://i.noah.pm/xWg6hF.gif)

patches/server/0732-Hide-unnecessary-itemmeta-from-clients.patch
Sytm: Vanilla does not have any visual changes, but [some mods](https://modrinth.com/mod/not-enough-animations) which aren't necessarily cheats could behave differently.

Machine-Maker (approved):

lynxplay (approved):

PR #9666, opened 2023-08-29
Added LootTableRollEvent

aerulion:
Adds an event that gets called whenever a LootTable rolls any loot.
There are two aspects I'm not sure about:
- Whether this should be a ServerEvent, or a WorldEvent like the LootGenerateEvent (but just forwarding the world from the LootContext location seems a bit pointless)?
- If this is the best way to implement / call such an event. The goal was to have a general event for all loot table rolls, but this implementation only covers loot generated via LootTable#getRandomItems method returning an ObjectArrayList of the generated items. That for example excludes the loot generated by a LivingEntity, since it calls the void method with a consumer consuming the generated items, and I couldn't think of any good way to implement the event call in the consumer method.

Nonetheless this implementation allows for modifying the loot of Brushable Blocks, which wouldn't be possible otherwise (apart from replacing the item during the BlockDropItemEvent, but that leads to the "old item" sticking out the brushable block during brushing).

### DIFF
--- a/patches/server/1029-Added-LootTableRollEvent.patch
+++ b/patches/server/1029-Added-LootTableRollEvent.patch
@@ -1,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: aerulion <aerulion@gmail.com>
+Date: Tue, 29 Aug 2023 04:52:12 +0200
+Subject: [PATCH] Added LootTableRollEvent
+
+
+diff --git a/src/main/java/net/minecraft/world/level/storage/loot/LootTable.java b/src/main/java/net/minecraft/world/level/storage/loot/LootTable.java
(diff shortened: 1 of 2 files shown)


patches/server/1029-Added-LootTableRollEvent.patch
```diff
@@ -0,0 +1,43 @@
...
++++ b/src/main/java/net/minecraft/world/level/storage/loot/LootTable.java
+@@ -132,7 +132,10 @@ public class LootTable {
+ 
+         Objects.requireNonNull(objectarraylist);
+         this.getRandomItems(context, objectarraylist::add);
+-        return objectarraylist;
++        // Paper start - LootTableRollEvent
++        final io.papermc.paper.event.server.LootTableRollEvent lootTableRollEvent = CraftEventFactory.callLootTableRollEvent(this, context, objectarraylist);
++        return lootTableRollEvent.getLoot().stream().map(CraftItemStack::asNMSCopy).collect(ObjectArrayList.toList());
++        // Paper start - LootTableRollEvent
```
SO-32, line 19: Change to paper end

PR #9676, opened 2023-08-31
Take in account waterlogged blocks for scaffolding

Lulu13022002:
The scaffolding is waterloggable and when the limit of suspended blocks is reached in water, the block fade into air (for the BlockFadeEvent) and into water (for the EntityChangeBlockEvent/BlockDestroyEvent)


lynxplay (approved):
Might be worth kinda just merging with the existing `0851-Add-missing-BlockFadeEvents.patch` patch, beyond that, lgtm

Machine-Maker (approved):
I agree with lynx, can be merged into that other patch and the patch renamed to smth about adding and fixing fade events.

Issue #6911, opened 2021-11-20
Add option to disconnect all players on stopping before disabling plugins


molor:
### Is your feature request related to a problem?

Partially. Every time after I stop the server, all connected players was just freezes for 3-10 seconds and can't open chests, place blocks (they can place blocks on client side, but server just ignores this during shutdown), use bow etc.

### Describe the solution you'd like.

I think that this problem can solved if we have setting like `disconnect-everyone-before-shutdown`, which defaults to `false` to keep initial behavior, so any plugin can handle PlayerQuitEvent before it was disabled, and players don't need to wait and asking us a question like "it's my network lag or server crashed?!".

Also it's would be nice if the `stop` command had the ability to specify the reason why the server was stopped instead of using default bukkit's `shutdown-message`

### Describe alternatives you've considered.

Maybe this can be solved by plugin that overrides `stop` command and kicks everyone before calling Server#shutdown(), but I don't know is this possible and will Paper CTLR+C handler support it..

olijeffers0n:
Hmm, I'm not sure im a fan of this as i feel some plugins may (ab)use the fact that when their onDisable method is called they can then get all online players and deal with any data accordingly. However i see what you mean about just using the PlayerQuitEvent.

molor:
I think that if plugin works with player data, they save this data anyway: 1) after player quit, or 2) on disable if it uses cache, so it is unlikely that any data loss can occur here. Also we sometimes use the custom `kickall` command before shutdown, and this works without problems

electronicboy:
kick all command is co-ordinated and not ran in the same tick as the server falling apart generally, an option can be added given that there is a good reason for this, but, dragons beware... :/

PR #9679, opened 2023-09-02
Save players before plugin disabling on server shutdown

DerEchtePilz:
For reference: https://discord.com/channels/289587909051416579/555462289851940864/1147468559887384586

If this behaviour is intended, feel free to close this PR, but as written in that message, I think `/restart` and `/stop` should follow the same order of operations when disabling plugins and saving players.

### DIFF
--- a/patches/server/1043-Save-players-before-disabling-plugins-on-server-shut.patch
+++ b/patches/server/1043-Save-players-before-disabling-plugins-on-server-shut.patch
@@ -1,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DerEchtePilz <81232921+DerEchtePilz@users.noreply.github.com>
+Date: Sat, 2 Sep 2023 12:48:58 +0200
+Subject: [PATCH] Save players before disabling plugins on server shutdown
+
+
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index aeae4f8d4ead24db315631c3d2c0b930d0d51e02..93145f9c35e249e4a223109a1b233ffc68e90aac 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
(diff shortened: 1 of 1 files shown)


DerEchtePilz:
I also wanted to say that this allows the `PlayerKickEvent` to work with the `/stop` command because players are kicked before plugins get disabled and therefore the `PlayerKickEvent` is called.

molor:
Fixes [#6911](https://github.com/PaperMC/Paper/issues/6911)
Also see the comments in it

electronicboy:
This is 100% not something that can be done mid patch release, this is more a major change in behavior in which needs to be done on a major release with proper testing to ensure that we don't break anything;

This is how this logic has worked for a decade, and while you can maybe argue that it's weird, it is a side-effect of various dozen other choices in the codebase over a decade which has left us here

DerEchtePilz:
So you say this could be added but this is more likely something for a new major MC version?
Or did you mean something else with "major release"?

lynxplay:
Yea, this logic would be desirable in the long run imo, it just
a) would best fit something like, 1.21 release
b) might be best to wait for papers hardfork or be proposed to spigot.

Given that the quit/onDisable stuff has worked like this for years and most logic that relies on this logic deals with rather crucial parts of plugins (e.g. data saving) changing this mid version is rather terrible for plugin developers to update.
Additionally, given that this breaks spigot compatibility, imo it is a rough PR even during a major release, as spigot plugins would just silently fail if they rely on spigots behaviour.

DerEchtePilz:
> b) might be best to wait for papers hardfork or be proposed to spigot.
#6911 suggests to add a config option. Would that, assuming this is getting merged at some point, be a feasable option until a hardfork? Or is this something that better waits until the hardfork?

electronicboy:
a config option was proposed for /stop to boot everybody, but there is no expectation that it would change event behavior, the server has stopped ticking and many things relied upon that in order to fire events, etc; 

If we're going to change the behavior around this, it should just be changed without a config option, rather than having such a dangerous setting sitting around with massive side-effects

Leguan16:
I guess that would be an option. But probably only if it's disabled per default, logically. 

On the other side if you enable it and have spigot plugins that rely on the old system it breaks again. So maybe an unsupported option until Hardfork?

PR #9683, opened 2023-09-03
Force sync inventory after PrepareResultEvent

Sytm:
This is required because when changing the result for an anvil (when the event is called because of AnvilMenu.setRenameText method) to an invalid result the item is not synced back to the client


Malfrador (approved):
Looks good to me and works as expected. Considering the client will refuse to pick-up results that don't match the input, forcing a re-sync of the result is necessary to make the event work properly in other cases too.

Owen1212055:
what is an "invalid result"?

Sytm:
> what is an "invalid result"?

For example if the inputs are a grass block and a stone block, the client predicts that this is not possible and therefore displays that: a crossed out result arrow.

Meanwhile the server could allow this and set the result to some other arbitrary item, but this is not synced to the client

Issue #7955, opened 2022-06-12
Usage of a paper-plugin.yml file for Paper-specific plugins


Andre601:
### Is your feature request related to a problem?

PaperMC is currently using the same `plugin.yml` like Spigot does.
This comes with all its pros and cons, especially when you make a plugin that should cater towards both and don't want to make separate jars for each.

A prime example would be the `libraries` option available.
While loading specific libraries - such as Kyori's Adventure library - makes sense on Spigot, is it sometimes wasted effort on a Paper Server due to it already offering it natively.

Other issues are that you may need to create stuff like interfaces to have convenience methods to f.e. send messages using Components on both Spigot and Paper which require extra stuff such as `Class.forName` checks to find out which server is actually used and if components are available.

### Describe the solution you'd like.

Paper should offer a way to define separate options using a file like `paper-plugin.yml`
This basically would inherit what BungeeCord is already doing: Prioritize and use that file if present and otherwise fall back to the default `plugin.yml` file.

Such a file could yield several benefits:

- Define a separate main class to load, allowing you to split up paper-only stuff into a different section while keeping a single, universal jar.
- Have separate libraries defined to load on a Paper Server. Avoids loading of already provided libraries such as Kyori adventure.
- More enhanced/improved formats or options in general without risk of breaking existing `plugin.yml` formatting.

To add to the first point: Having a separate main class could then allow to use Paper-specific methods without the risk of having them used on a Spigot Server.
This could then allow Paper to provide convenience methods to f.e. check if a particular feature would be present, in case that is wanted.

### Describe alternatives you've considered.

There are 2 alternatives for the libraries part I can think of: Shade it in or load it either way.

Both have their downsides (Increased jar size or loading of an already provided library)

For the other stuff did I already mention currently used methods such as `Class.forName` checks.

Owen1212055:
Added in https://github.com/PaperMC/Paper/pull/8108

DereC4:
So this is where it began..

PR #9685, opened 2023-09-04
Print warning if people try to call JavaPlugin#getCommand with paper plugins

### LINKED ISSUE #9669: Print helpful warning if 'commands' section is found in paper-plugin.yml

Some developers expect to use the `commands` section in the paper-plugin.yml as they do in the plugin.yml. This is unsupported, which causes calls such as `getCommand("myCommand").setExecutor(someExecutor)` to fail with a NPE.

A NPE does not immediately explain the cause of the bug, namely that the developer expects to use the commands section but the commands section is no longer a feature. So the developer revisits the paper-plugin.yml, checks for typos, checks the code again, recompiles and reruns, and it still doesn't work. Then they maybe read the documentation on Paper plugins one more time, or join the PaperMC discord (as has happened recently) to debug.

If the commands section is found, print a warning message informing developers it is not available using paper-plugin.yml and programmatic registration is preferred.

It may be worth even throwing an exception and refusing to load the plugin. I can see no other reason to use the commands section if paper-plugin.yml is not in use.

Do nothing, and continue to sustain inquiries and support requests on this.

The documentation is already sufficiently prominent, so I see nothing to change there.

### Other

_No response_

Leguan16:
This PR adds a simple warning if a `commands` section has been found in the `paper-plugin.yml` file.

I am taking suggestions for the warning message since I don't know if I should include how to register commands with Paper Plugins. 
I also did not know which logger I should use since I don't think a plugin logger has already been created at the time where the warning is printed. 

Which brings me to the next point:
Currently the warning is printed before the server is started: 

```.sv
Loading libraries, please wait...
2023-09-04 17:30:38,271 main WARN Advanced terminal features are not available in this environment
[17:30:38 WARN]: [Paper-Test-Plugin] Defining commands in paper-plugin.yml is not supported! <-------
[17:30:41 INFO]: Environment: authHost='https://authserver.mojang.com', accountsHost='https://api.mojang.com', sessionHost='https://sessionserver.mojang.com', servicesHost='https://api.minecraftservices.com', name='PROD'
[17:30:42 INFO]: Loaded 7 recipes
[17:30:42 INFO]: Starting minecraft server version 1.20.1
```
I don't think there is a way of printing the warning during the plugin load period (unless we add a boolean or something) since I don't think it is not possible to get the node after the `PaperPluginMeta#create()` method. 

Closes #9669

### DIFF
--- a/patches/server/0013-Paper-Plugins.patch
+++ b/patches/server/0013-Paper-Plugins.patch
@@ -4748,10 +4748,10 @@ index 0000000000000000000000000000000000000000..e3430f535e8e9c3b8b44bf2daece8c47
 +}
 diff --git a/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..489a1737516eead7e11eaa226d60dd647228cbdb
+index 0000000000000000000000000000000000000000..45bd29b70782e29eb11c36eaca0f940aee49799b
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
-@@ -0,0 +1,252 @@
+@@ -0,0 +1,248 @@
 +package io.papermc.paper.plugin.provider.configuration;
 +
 +import com.google.common.base.Preconditions;
@@ -4857,10 +4857,6 @@ index 0000000000000000000000000000000000000000..489a1737516eead7e11eaa226d60dd64
 +                .build();
 +        }
 +
-+        if (!node.node("commands").virtual()) {
-+            org.slf4j.LoggerFactory.getLogger(pluginConfiguration.getName()).warn("Defining commands in paper-plugin.yml is not supported! - Commands will not be available!");
-+        }
-+        
 +        return pluginConfiguration;
 +    }
 +
--- a/patches/server/0013-Paper-Plugins.patch
+++ b/patches/server/0013-Paper-Plugins.patch
@@ -4748,10 +4748,10 @@ index 0000000000000000000000000000000000000000..e3430f535e8e9c3b8b44bf2daece8c47
 +}
 diff --git a/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..45bd29b70782e29eb11c36eaca0f940aee49799b
+index 0000000000000000000000000000000000000000..489a1737516eead7e11eaa226d60dd647228cbdb
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
-@@ -0,0 +1,248 @@
+@@ -0,0 +1,252 @@
 +package io.papermc.paper.plugin.provider.configuration;
 +
 +import com.google.common.base.Preconditions;
@@ -4857,6 +4857,10 @@ index 0000000000000000000000000000000000000000..45bd29b70782e29eb11c36eaca0f940a
 +                .build();
 +        }
 +
++        if (!node.node("commands").virtual()) {
++            org.slf4j.LoggerFactory.getLogger(pluginConfiguration.getName()).warn("Defining commands in paper-plugin.yml is not supported! - Commands will not be available!");
++        }
++        
 +        return pluginConfiguration;
 +    }
 +
--- a/patches/server/0013-Paper-Plugins.patch
+++ b/patches/server/0013-Paper-Plugins.patch
@@ -4748,10 +4748,10 @@ index 0000000000000000000000000000000000000000..e3430f535e8e9c3b8b44bf2daece8c47
 +}
 diff --git a/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..45bd29b70782e29eb11c36eaca0f940aee49799b
+index 0000000000000000000000000000000000000000..705ea6f2ddee0b38aa33ca219f0bb695cc3fafad
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/plugin/provider/configuration/PaperPluginMeta.java
-@@ -0,0 +1,248 @@
+@@ -0,0 +1,253 @@
 +package io.papermc.paper.plugin.provider.configuration;
 +
 +import com.google.common.base.Preconditions;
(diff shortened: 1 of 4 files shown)


patches/server/0013-Paper-Plugins.patch
```diff
@@ -4857,6 +4857,10 @@ index 0000000000000000000000000000000000000000..45bd29b70782e29eb11c36eaca0f940a
 +                .build();
 +        }
 +
++        if (!node.node("commands").virtual()) {
++            org.slf4j.LoggerFactory.getLogger(pluginConfiguration.getName()).warn("Defining commands in paper-plugin.yml is not supported! - Commands will not be available!");
```
Warriorrrr: You could probably link to the docs page (https://docs.papermc.io/paper/dev/getting-started/paper-plugins#commands) that explains what to do instead.

Owen1212055:
Tbh, It's a bit worrying that people are using these plugins like this, to the point where we need to warn if they aren't even reading the documentation.
image

That format is still very much subject to change, and in all honesty, I would rather have some kind of warning if a paper plugin is being loaded... because those are very much still experimental and are waiting for future API.

Leguan16:
That is a good point. But I think the one thing has little to do with the other. 
I think many people are creating a Paper plugin with the MinecraftDev IJ plugin and don't even realize that its different to Bukkit Plugins. So what you suggested should probably be handled over there by adding a little warning next to the paper manifest checkbox. 

You can see it in the Discord that people are even too lazy to read JavaDocs so we cannot assume that they read the docs. The best thing we can do is give them a hint that they did something wrong on our side.

A248:
> And if a developer is actively handing out Paper plugins, /shrug what fun.

There are already plugins published on Modrinth using the paper-plugin.yml. I recall viewing this myself.

With all due respect, one should not publish an API to a popular project without expecting its use. The adoption of Paper plugins began as soon as a newfangled API for declaring plugin metadata was released. These shifts are not necessarily misguided, as the original impetus for Paper plugins outlined several use-cases which remain relevant: https://github.com/PaperMC/Paper/issues/7955

lynxplay:
> With all due respect, one should not publish an API to a popular project without expecting its use.

We do expect *usage* of the API, hence we published it. To collect feedback on it as we further improve it.
I don't think anyone minds people publishing paper plugins, ***IF*** they are aware of the contracts that come along with the usage of paper plugins, specifically changes to format, abi, api without backwards compatibility promises, as explicitly defined via the docs page and the experimental annotation.

Developers that publish such plugins *need* to be aware of those restrictions and risks and what those might mean for their users, hence we mostly expected this API to be used in private plugin development as updating to changes to paper plugins becomes a lot easier.

Tho given how common the question became regarding command definition in paper-plugin.yml, this is clearly not the case as it seems that a lot of people have started using paper plugins without the required insights into the *unstable* API and its limitations. Which is why I 100% agree with your Issue about actively warning to further raise the fact, that they are trying to use unstable API incorrectly.

If we have to go as far as printing a warning for paper plugins in general, I am unsure of.
Again, either developers are already aware of the risks they accept when using paper plugins and use it correctly (aware of breaking changes), not much need for a warning there, or people have ignored *every* warning thrown at them by us, at which point, another warning in their server logs isn't going to help much.

4drian3d:
> There are already plugins published on Modrinth using the paper-plugin.yml. I recall viewing this myself.

Personally, I have more than 10 public plugins made based on a `paper-plugin.yml`, I am fully aware that there will be changes in the structure of this type of plugins in the future and I have no problem to update my projects based on the new changes.
What I don't agree is to add a warning in console about a paper plugin being loaded, because there would be users asking about the warning and what to do to remove it and the only solution would be to lose exclusivity of all my plugins to be loaded only in Paper, which I want to avoid doing at all cost.

lynxplay:
With my comment about a warning I was only referring to a general warning comment for loading paper plugins, sorry if that wasn't clear.

I am 100% on board with a warning/exception for an invalid yaml format in the paper-plugin.yml.

MiniDigger:
cant we have a general yaml format and throw if the format doesnt match? special casing this feels strange

Leguan16:
> special casing this feels strange

usually it isn't really a problem to have random tags in the config.
This is rather meant as a hint that defining commands has changed.

Leguan16:
> cant we have a general yaml format and throw if the format doesnt match?

A generic exception like "invalid section <name> found in paper-plugin.yml" would just again bring the people to the discord asking what they should use instead. 
The warning is supposed to minimize this.

Leguan16:
What if, instead of checking the `paper-plugin.yml` for a `commands` section, we print a warning if `JavaPlugin#getCommand()` is called by a paper plugin?

```java
@Nullable
public PluginCommand getCommand(@NotNull String name) {
    // Paper start - print warning if paper plugin
    if (getResource("paper-plugin.yml") != null) {
        this.logger.warning("---------------------------------WARNING--------------------------------");
        this.logger.warning("         Defining commands in paper-plugin.yml is unsupported!          ");
        this.logger.warning("      Please refer to the docs for proper command initialization!       ");
        this.logger.warning("https://docs.papermc.io/paper/dev/getting-started/paper-plugins#commands");
        this.logger.warning("                    Command will not be initialized!                    ");
        this.logger.warning("---------------------------------WARNING--------------------------------");
        return null;
    }
    // Paper end
    ...
}
```

MiniDigger:
that seems the nicer way for that particular usecase
not sure tho if checking for the file is the best way, idk much about the infra behind paper plugins, but there gotta be a nice instanceof check we could do

Leguan16:
> but there gotta be a nice instanceof check we could do

i changed it to 
```java
if (!(pluginMeta instanceof PluginDescriptionFile)) {
    this.logger.warning("---------------------------------WARNING--------------------------------");
    this.logger.warning("         Defining commands in paper-plugin.yml is unsupported!          ");
    this.logger.warning("      Please refer to the docs for proper command initialization!       ");
    this.logger.warning("https://docs.papermc.io/paper/dev/getting-started/paper-plugins#commands");
    this.logger.warning("                    Command will not be initialized!                    ");
    this.logger.warning("---------------------------------WARNING--------------------------------");
    return null;
}
```
I am not sure if I should throw an exception instead of returning null.

Leguan16:
Updated it to reflect discussed changes. 
I also decided to throw a RuntimeException instead of returning null. 

The only downside is that it doesn't respond to 
`getServer().getPluginCommand()` yet.

electronicboy:
I really dislike the notion of this; There are still some Qs over what we want to do long term here, ideally we wanna move away from the broken bukkit command system, however, the expectation of there still being an easy to use command system still exists, etc, etc; it's not 100% decided that such a commands block will never exist, I think that this issue would generally be better off solved with documentation and such

Leguan16:
Generally I agree with you but good luck getting new people to read the docs. As I have mentioned before this is solely for the purpose of hinting new devs that they are doing something wrong. This can always be removed in the future but to be honest, there are too much plugin tutorials and people will use those tutorials to develope Paper Plugins in the future. So getting people away from the old system will never really work until we remove the methods entirely. And until then I think its a good idea to print such a warning. 

This is also not final but in my opinion its important and more user-friendly to add this (or in another form).

electronicboy:
Except this ends up leaking warnings to other plugin invocations of this stuff, which imho makes this a no go on this solution

Leguan16:
I think I am closing this in favor of https://github.com/minecraft-dev/MinecraftDev/pull/2122 since it got merged recently. If you ignore this warning there is really nothing else we can do. 

If someone still thinks this is something worth adding (in another form than how it works at the moment) just let me know :)

PR #9686, opened 2023-09-05
Fix leashed pets teleporting to owner when loaded

### LINKED ISSUE #8462: Tameable entity get teleportet to player after re-connect
### Expected behavior

Tameable entity have to remain leashed at their positions

### Observed/Actual behavior

Tameable entity teleports to player and unleashes

### Steps/models to reproduce

https://youtu.be/8fpqHAmN1uU

### Plugin and Datapack List

no plugins
standard datapack: [vanila (build-in)], [file/bukkit (world)]

### Paper version

This server is running Paper version git-Paper-201 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: 57555b8)
You are 5 version(s) behind

### Other

maybe https://bugs.mojang.com/browse/MC-173303

Warriorrrr:
Closes #8462

The FollowOwnerGoal is ticked before the re-leashing logic inside Mob#restoreLeashFromSave is, which allows tamables to teleport to their owners in their first tick. This PR fixes that by also checking whether the entity has a "pending" leash before running the goal.

### DIFF
--- a/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
@@ -57,6 +57,12 @@ https://bugs.mojang.com/browse/MC-84789
 https://bugs.mojang.com/browse/MC-225381
   Fix overfilled bundles duplicating items / being filled with air
 
+https://bugs.mojang.com/browse/MC-173303
+  Fix leashed pets teleporting to owner when loaded
+
+== AT ==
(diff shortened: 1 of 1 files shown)


lynxplay (approved):
imo better than messing around with isLeashed, given that we would not want isLeashed to yield true until the entity is assigned.

Machine-Maker (approved):

electronicboy (approved):

PR #9688, opened 2023-09-07
Add experience points api

Sytm:
Adds API to handle the individual experience points of a player. This is what the existing get/setTotalExperience should have been, but we never got.

I know that the implementation is kind of top-tier cursed, but thanks to the three different equations Mojang has choosen this is what it is. If something like _this_ is not desired in Paper just close the PR, although it would be nice to have.

I have tested this method up to I think it was around 39,000,000 experience points/level 2961 and the math still checked out

### DIFF
--- a/patches/api/0449-Add-experience-points-API.patch
+++ b/patches/api/0449-Add-experience-points-API.patch
@@ -1,0 +1,56 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lukas Planz <lukas.planz@web.de>
+Date: Tue, 5 Sep 2023 20:33:52 +0200
+Subject: [PATCH] Add experience points API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
+index 9130a57cf6ef5d543703a03aeed07aa17b1ab7e8..09629652fc13964f2ff2ef549d36292d0b8a89be 100644
+--- a/src/main/java/org/bukkit/entity/Player.java
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -1764,6 +1764,45 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+      * @param exp New total experience points
+      */
+     public void setTotalExperience(int exp);
++    // Paper start
++    /**
++     * Gets the players total amount of experience points he collected to reach the current level and level progress.
++     *
++     * <p>This method differs from {@link #getTotalExperience()} in that this method always returns an
++     * up-to-date value that reflects the players{@link #getLevel() level} and {@link #getExp() level progress}</p>
++     *
++     * @return Current total experience points
++     * @see #getLevel()
++     * @see #getExp()
++     * @see #setExperienceLevelAndProgress(int)
++     */
++    @org.jetbrains.annotations.Range(from = 0, to = Integer.MAX_VALUE) int calculateTotalExperiencePoints();
++
++    /**
++     * Updates the players level and level progress to that what would be reached when the total amount of experience
++     * had been collected.
++     *
++     * <p>This method differs from {@link #setTotalExperience(int)} in that this method actually updates the
++     * {@link #getLevel() level} and {@link #getExp() level progress} so that a subsequent call of
++     * {@link #calculateTotalExperiencePoints()} yields the same amount of points that have been set</p>
++     *
++     * @param totalExperience New total experience points
++     * @see #setLevel(int)
++     * @see #setExp(float)
++     * @see #calculateTotalExperiencePoints()
++     */
++    void setExperienceLevelAndProgress(@org.jetbrains.annotations.Range(from = 0, to = Integer.MAX_VALUE) int totalExperience);
++
++    /**
++     * Gets the total amount of experience points that are needed to reach the next level from zero progress towards it.
++     *
++     * <p>Can be used with {@link #getExp()} to calculate the current points for the current level and alike</p>
++     *
++     * @return The required experience points
++     * @see #getExp()
++     */
++    int getExperiencePointsNeededForNextLevel();
++    // Paper end
+ 
+     /**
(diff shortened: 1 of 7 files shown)


patches/server/1029-Add-experience-points-API.patch
```diff
@@ -0,0 +1,58 @@
...
++        this.getHandle().experienceProgress = (float) remainingPoints / this.getExperiencePointsNeededForNextLevel();
++        this.getHandle().lastSentExp = -1;
++    }
++
++    @Override
++    public int getExperiencePointsNeededForNextLevel() {
++        return this.getHandle().getXpNeededForNextLevel();
++    }
++
++    private int getTotalExperiencePoints(int level) {
```
MiniDigger: if we pull this, we def need a link to some documentation here
is there no way we could call a vanilla method that does something like this?

MiniDigger:
what is the difference here to the other method?
that should definitely be documented on both methods, old and new, so that we dont leave people guessing or having to look at the code (which I cant right now, hence the question)

patches/server/1029-Add-experience-points-API.patch
Sytm: Well, I pulled the equations from the [Minecraft Wiki](https://minecraft.fandom.com/wiki/Experience?so=search#Leveling_up) tbh. Sadly, the player experience is only stored as the level and the progression towards the next level as a float from 0.0 -> 1.0.

If you were to collect experience ingame the points are first converted into progression and then added to the current progression.

MiniDigger: just drop a link to that page in the code then, then people know where the magic numbers come from when reading the code

patches/api/0440-Add-experience-points-API.patch
```diff
@@ -0,0 +1,54 @@
...
++     * Gets the players total amount of experience points he collected to reach the current level and level progress.
++     *
++     * <p>This method differs from {@link #getTotalExperience()} in that this method always returns an
++     * up-to-date value that reflects the players{@link #getLevel() level} and {@link #getExp() level progress}</p>
++     *
++     * @return Current total experience points
++     * @see #getLevel()
++     * @see #getExp()
++     */
++    public @org.jetbrains.annotations.Range(from = 0, to = Integer.MAX_VALUE) int getTotalExperiencePoints();
```
Machine-Maker: I think to further differentiate this method and `getTotalExperience`, this method should be named something like `calculateTotalExperience` since it's not *getting* a value, but rather calculating it based on the xp level and progress towards next level.

patches/api/0440-Add-experience-points-API.patch
```diff
@@ -0,0 +1,54 @@
...
++     *
++     * <p>This method differs from {@link #setTotalExperience(int)} in that this method actually updates the
++     * {@link #getLevel() level} and {@link #getExp() level progress} so that a subsequent call of
++     * {@link #getTotalExperiencePoints()} yields the same amount of points that have been set</p>
++     *
++     * @param exp New total experience points
++     * @see #setLevel(int)
++     * @see #setExp(float)
++     */
++    public void setTotalExperiencePoints(@org.jetbrains.annotations.Range(from = 0, to = Integer.MAX_VALUE) int exp);
```
Machine-Maker: I would then also rename this method to something like `setExperienceLevelAndProgress` which takes the total xp amount like it does now.

patches/server/1031-Add-experience-points-API.patch
```diff
@@ -0,0 +1,60 @@
...
++    private int getLevelsForExperiencePoints(int points) {
++        if (points <= 352) { // Level 0-16
++            return (int) Math.floor(Math.sqrt(points + 9) - 3);
++        } else if (points <= 1507) { // Level 17-31
++            return (int) Math.floor(8.1 + Math.sqrt(0.4 * (points - (7839.0 / 40.0))));
++        } else { // 32+
++            return (int) Math.floor((325.0 / 18.0) + Math.sqrt((2.0 / 9.0) * (points - (54215.0 / 72.0))));
++        }
++    }
++    // Paper end
```
Machine-Maker: I'm not a super fan of copying the logic for converting level into total points. Although I'm not super clear on what a good way to refactor the vanilla logic would be to allow us to use it directly. I think at the minimum, some `// Paper - diff on change` comments should be added to the key lines in `Player#getXpNeededForNextLevel` as that seems to be where the logic comes from.

Machine-Maker (changes requested):

patches/server/1031-Add-experience-points-API.patch
Sytm: Yeah, I agree that this is a somewhat unfortunate way of handling this, but due to how the experience stuff is implemented it is necessary.

If you look at the initial commit of this PR, you can see that at least for setting I was able to use parts of vanilla code to set the level and progress. Getting still had to be done with the formula.

Machine-Maker: Yeah, then at least add some `// Paper - diff on change` comments in a few places to try and catch if mojang ever changes this.

Machine-Maker (approved):

Owen1212055 (approved):
Please rebase

lynxplay:
I think it is worth to add a test here. You can create a mock player relatively quickly (see how GameTestHelper does it).
That way, in future releases and changes, this API does fail a test when it goes out of sync with vanilla given we cannot really use internal logic for the API but have to re-create it.

Sytm:
Mocking a Bukkit Player isn't as easy:
- Bukkit Player
  -  nms Player
   - MinecraftServer <- Requires a lot more stuff
   - ServerLevel
     - Level <- Mocking errors because some registries are referenced in static fields that are obviously not initialized

lynxplay:
Yea, you are right. The fact that some of the paper code takes a level for granted makes this annoying.
I guess this is fine then, we just need to invest time into integration testing frameworks down the line :upside_down_face:

lynxplay (approved):

PR #9689, opened 2023-09-07
Enforce sign line nullability when setting line with Adventure Method

thelooter:
Enforces the nullability when setting a Sign Line with the Adventure Method

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -3279,7 +3279,7 @@ index bd43440fcbcf461fb8ca94b96cd256f392efeb84..5af30bff1bfdf09daa585071f2ec1fe0
          Component[] components = new Component[4];
  
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java b/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java
-index 43f6ac44525048faf0b0cd2c11f9bc36fa0cab5d..ef624d9e71dde2a9f3d32b4436a4dd7acbb02d53 100644
+index 43f6ac44525048faf0b0cd2c11f9bc36fa0cab5d..4747d77fd5fd12116ef40d897a08c7baca60a399 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java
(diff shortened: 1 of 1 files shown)


patches/server/0010-Adventure.patch
```diff
@@ -3313,10 +3313,17 @@ index bd43440fcbcf461fb8ca94b96cd256f392efeb84..5af30bff1bfdf09daa585071f2ec1fe0
...
 diff --git a/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java b/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java
-index 43f6ac44525048faf0b0cd2c11f9bc36fa0cab5d..1317cdd48c1c27c9e5152da5403237b1bb15ac52 100644
+index 43f6ac44525048faf0b0cd2c11f9bc36fa0cab5d..22af5b9e52ddbee333f14047e9b19693059485be 100644
 --- a/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java
 +++ b/src/main/java/org/bukkit/craftbukkit/block/sign/CraftSignSide.java
-@@ -12,37 +12,69 @@ import org.jetbrains.annotations.Nullable;
+@@ -1,5 +1,6 @@
+ package org.bukkit.craftbukkit.block.sign;
+ 
++import com.google.common.base.Preconditions;
```
Leguan16: please don't use imports. Use Full Qualified Names instead.

patches/server/0010-Adventure.patch
```diff
@@ -3347,6 +3354,7 @@ index 43f6ac44525048faf0b0cd2c11f9bc36fa0cab5d..1317cdd48c1c27c9e5152da5403237b1
 +
 +    @Override
 +    public void line(final int index, final net.kyori.adventure.text.@NotNull Component line) throws IndexOutOfBoundsException {
++        Preconditions.checkNotNull(line, "Line cannot be null");
```
Leguan16: use `com.google.common.base.Preconditions.checkNotNull(line, "Line cannot be null");`

Warriorrrr (approved):

Machine-Maker (approved):
The rest of the API uses checkArgument so the exception is of the IllegalArgumentException type instead of NPE, but idk how big a deal that is.

electronicboy (approved):

PR #9691, opened 2023-09-08
Add predicate for block when raytracing

TonytheMacaroni:
Adds an overload to `World#rayTrace` and `World#rayTraceBlocks` that accepts a `Predicate<Block>`. Blocks that match the predicate are ignored for collision calculations.


patches/api/0440-Add-predicate-for-block-data-when-raytracing-blocks.patch
```diff
@@ -0,0 +1,90 @@
...
++     * @param maxDistance the maximum distance
++     * @param fluidCollisionMode the fluid collision mode
++     * @param ignorePassableBlocks whether to ignore passable but collidable
++     *     blocks (ex. tall grass, signs, fluids, ..)
++     * @param ignoreFilter predicate for block data to ignore when passing
++     *     through, or <code>null</code> to consider all blocks
++     * @return the ray trace hit result, or <code>null</code> if there is no hit
++     */
++    @Nullable
++    public RayTraceResult rayTraceBlocks(@NotNull Location start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, @Nullable Predicate<BlockData> ignoreFilter);
```
Machine-Maker: Both of these methods on World should use Position instead of Location as a world isn't required.

Machine-Maker (changes requested):

patches/api/0440-Add-predicate-for-block-data-when-raytracing-blocks.patch
TonytheMacaroni: What should be done with the world and finite preconditions? Adding a method method like `Location#checkFinite` to `Position` wouldn't be difficult, but the ray trace methods currently validate that the passed location has the same world. Should the newer ones just omit the check, and only check on the older ray trace methods?

Machine-Maker: Just check if the position is an instance of Location and then run those checks. A `checkFinite` method for Position might be good, but shouldn't be done in this PR.

Machine-Maker:
> Should the predicate be for BlockState instead of Block?

No, `Block` is perfect for this. It is just a wrapper for a world and position so it is more efficient than creating BlockState for every position checked.

> Is the cast to LevelAccessor fine? The passThrough predicate is only non-null when the BlockGetter is an instance of ServerLevel.

To be safe, I would also, in addition to the null check, have an instanceof check.

patches/api/0440-Add-predicate-for-block-data-when-raytracing-blocks.patch
TonytheMacaroni: Should more overloads be added that take a `Position` instead of a `Location` for the pre-existing ray trace methods? Would require an additional one to call `World#rayTraceEntities` in `World#rayTrace` with a `Position`, and was wondering if it's necessary to add overloads for the shorter variants.

Machine-Maker: At some point we will probably add some bytecode rewriting to just allow us to change those Location params to Position since Location implements Position. But that is for another time. For now, just use Position in new stuff where it is applicable.

TonytheMacaroni: Alright, added an overload for `World#rayTraceEntities`, and changed the two added methods to take a `Position` .

patches/api/0441-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,119 @@
...
++     * @param raySize entity bounding boxes will be uniformly expanded (or
++     *     shrinked) by this value before doing collision checks
++     * @param filter only entities that fulfill this predicate are considered,
++     *     or <code>null</code> to consider all entities
++     * @return the closest ray trace hit result, or <code>null</code> if there
++     *     is no hit
++     */
++    @Nullable
++    public RayTraceResult rayTraceEntities(@NotNull io.papermc.paper.math.Position start, @NotNull Vector direction, double maxDistance, double raySize, @Nullable Predicate<Entity> filter);
++    // Paper End
```
Machine-Maker: Don't need to add this. We will change Location to Position at some point in the future.

patches/api/0441-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,119 @@
...
++     * @param maxDistance the maximum distance
++     * @param fluidCollisionMode the fluid collision mode
++     * @param ignorePassableBlocks whether to ignore passable but collidable
++     *     blocks (ex. tall grass, signs, fluids, ..)
++     * @param passThrough predicate for blocks to ignore when passing
++     *     through, or <code>null</code> to consider all blocks
++     * @return the ray trace hit result, or <code>null</code> if there is no hit
++     */
++    @Nullable
++    public RayTraceResult rayTraceBlocks(@NotNull io.papermc.paper.math.Position start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, @Nullable Predicate<org.bukkit.block.Block> passThrough);
```
Machine-Maker: The Predicate here should be inverted so as to match the Predicate the rayTraceEntities method uses. Returning true includes the block in the search.

patches/api/0441-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,119 @@
...
++     *     shrinked) by this value before doing collision checks
++     * @param filter only entities that fulfill this predicate are considered,
++     *     or <code>null</code> to consider all entities
++     * @param passThrough predicate for blocks to ignore when passing through,
++     *      or <code>null</code> to consider all blocks
++     * @return the closest ray trace hit result with either a block or an
++     *     entity, or <code>null</code> if there is no hit
++     */
++    @Nullable
++    public RayTraceResult rayTrace(@NotNull io.papermc.paper.math.Position start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, double raySize, @Nullable Predicate<Entity> filter, @Nullable Predicate<Block> passThrough);
```
Machine-Maker: Same here regarding the inversion of the block predicate.

patches/server/1036-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,120 @@
...
++    @Override
++    public RayTraceResult rayTraceBlocks(io.papermc.paper.math.Position start, Vector direction, double maxDistance, FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, Predicate<Block> passThrough) {
+         Preconditions.checkArgument(start != null, "Location start cannot be null");
+-        Preconditions.checkArgument(this.equals(start.getWorld()), "Location start cannot be in a different world");
+-        start.checkFinite();
++        Preconditions.checkArgument(!(start instanceof Location location) || this.equals(location.getWorld()), "Location start cannot be in a different world");
++        org.bukkit.util.NumberConversions.checkFinite(start.x(), "x not finite");
++        org.bukkit.util.NumberConversions.checkFinite(start.y(), "y not finite");
++        org.bukkit.util.NumberConversions.checkFinite(start.z(), "z not finite");
++        // Paper end
```
Machine-Maker: Go ahead and at a `boolean isFinite` method to `Position` which the default implementation just checks if each component x, y, z are finite and returns true if it is. Then override that method in Location to also check the pitch and yaw. Then if its not, you can throw illegal argument exception.

Machine-Maker (changes requested):

patches/server/1036-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,120 @@
...
+ 
+         Preconditions.checkArgument(direction != null, "Vector direction cannot be null");
+         direction.checkFinite();
+@@ -1184,16 +1201,23 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+         }
+ 
+         Vector dir = direction.clone().normalize().multiply(maxDistance);
+-        Vec3 startPos = CraftLocation.toVec3D(start);
++        Vec3 startPos = new Vec3(start.x(), start.y(), start.z()); // Paper
```
lynxplay: This should probably live in the MCUtil patch as a utility.

patches/api/0441-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Required because `CraftWorld#rayTrace` calls `CraftWorld#rayTraceEntities` with its start position. Could construct a location from the position, but that seems a bit jank.

patches/api/0441-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Changed. Personally, a predicate for which blocks to pass through made more sense to me, but matching the other predicate seems fine as well.

patches/server/1036-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Added.

patches/server/1036-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Added.

patches/api/0442-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,119 @@
...
+
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index da524a71af74b02515b037f7fe09ba6988e2c8bf..06c0c759cf9abd11008795d31b0da97ab3b2163e 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -1713,6 +1713,28 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+     @Nullable
+     public RayTraceResult rayTraceEntities(@NotNull Location start, @NotNull Vector direction, double maxDistance, double raySize, @Nullable Predicate<Entity> filter);
+ 
++    // Paper Start
```
Lulu13022002: The start/end part of the paper comments should be in lowercase

patches/server/1039-Add-predicate-for-blocks-when-raytracing.patch
```diff
@@ -0,0 +1,118 @@
...
+ 
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+index 0e670de77a7f9926e295e1dd63d909bed1a959ca..b49f78d06bcb06676b77fda444f4db60a993f99f 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -1117,9 +1117,17 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+ 
+     @Override
+     public RayTraceResult rayTraceEntities(Location start, Vector direction, double maxDistance, double raySize, Predicate<Entity> filter) {
++        // Paper Start
```
Lulu13022002: Same here

patches/api/0009-Add-Position.patch
```diff
@@ -273,6 +274,13 @@ index 0000000000000000000000000000000000000000..300da713dcc303b340efad70efe57fac
 +    boolean isFine();
 +
 +    /**
++     * Checks if each component of this position is finite.
++     */
++    default boolean isFinite() {
++        return NumberConversions.isFinite(this.x()) && NumberConversions.isFinite(this.y()) && NumberConversions.isFinite(this.z());
```
Lulu13022002: You can use the java equivalent with Double.isFinite

patches/api/0009-Add-Position.patch
```diff
@@ -415,6 +423,11 @@ index 7c4db051472fb6a6c6d24092dc6f75487356690a..3b99f359f556e6f2c341d55fa69b7462
 +    }
 +
 +    @Override
++    public boolean isFinite() {
++        return io.papermc.paper.math.FinePosition.super.isFinite() && NumberConversions.isFinite(this.getYaw()) && NumberConversions.isFinite(this.getPitch());
```
Lulu13022002: You can use the java equivalent with Float.isFinite

TonytheMacaroni: Fixed.

patches/api/0009-Add-Position.patch
TonytheMacaroni: Fixed.

patches/server/1039-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Fixed.

patches/api/0442-Add-predicate-for-blocks-when-raytracing.patch
TonytheMacaroni: Fixed.

patches/api/0009-Add-Position.patch
TonytheMacaroni: Fixed.

Machine-Maker (approved):

Issue #5274, opened 2021-02-27
API to make ItemStacks glowing (without adding real enchants)


Trigary:
**Is your feature request related to a problem? Please describe.**
I want to make items (`ItemStack` / `ItemMeta`) glowing without changing anything else.

**Describe the solution you'd like**
Addition of two methods: `ItemMeta#setShiny(boolean)` and `ItemMeta#isShiny()`.
The setter shouldn't apply any real enchantments and definitely shouldn't add `ItemFlag.HIDE_ENCHANTS`: the item might already have enchantments (or if not, it might have enchantments in the future), so adding that flag would be bad. Adding a real enchantment is also bad since I don't want enchantments to show up under `#getEnchantments()` just because I called `.setShiny(true)`.
The return value of `isShiny()` should stay the same prior to and after serialization/deserialization, whatever that value might be. So in other words, this feature should support serialization.

**Describe alternatives you've considered**
Solutions that rely on `ItemFlag#HIDE_ENCHANTS` are possible but are complicated if we want to do it right: we have to keep in mind that items can get enchanted via commands, via enchanting tables, by other plugins. So we have to remove the `ItemFlag` when necessary, etc.
@HexedHero claims to achieve this by registering a custom `Enchantment`, but that's not supported by the Bukkit API, therefore is a non-solution.

**Additional context**
kashike believes implementing this is possible.

ghost:
.addenchant lure 1
.addflag hideenchants

Machine-Maker:
This looks like it *used* to be possible by setting the `Enchantments` key on an items tag to an empty NBTTagList, but it doesn't look like that is the case anymore. 

I don't know exactly what the client uses to determine if it should display it as glowing.

Draycia:
Looks like the effect is possible via setting an invalid enchantment.
`/give @s minecraft:diamond_sword{Enchantments:[{id:"",lvl:0}]} 1`
The lore is rendered exactly the same as a vanilla sword, but with the glowing effect and name change.

Something thing to note though is that setting Enchantments to an empty list or with an invalid enchantment causes the item to be unenchantable via enchanting tables. Anvils however work perfectly fine and as expected from my minimal testing. Though this behaviour has remained consistent with the old method prior to 1.16

Machine-Maker:
I wonder if the “unenchantable-ness” can be fixed on the server side or not? 

[...]

PR #9693, opened 2023-09-08
Add Foiled/Glow NBT + API

TreemanKing:
Adds Foiled/Glow API using NBT. 

Possibly can add the API to the Overrides for the default items that are foiled. Follows most of the logic of the unbreakable tag.

Relevant:
#5274 

Also not sure if this is going to be relevant due to #8711.

P.S. For some reason Patch 1018 was edited and it says it removed something? I also can't rebuild or apply new patches after rebuilding the first time which is confusing to me.

### DIFF
--- a/patches/api/0148-Add-an-API-for-CanPlaceOn-and-CanDestroy-NBT-Values.patch
+++ b/patches/api/0148-Add-an-API-for-CanPlaceOn-and-CanDestroy-NBT-Values.patch
@@ -226,10 +226,10 @@ index a42f1d53340e4073038d46b7fabf5d44248d5b32..dbc22807a33606f8fe326cc2f5f755fe
          return key;
      }
 diff --git a/src/main/java/org/bukkit/inventory/meta/ItemMeta.java b/src/main/java/org/bukkit/inventory/meta/ItemMeta.java
-index f78714c2a6d9da162c9802552984cbfad76ff728..55e9dc5d1d87fbc9d0dabbb7bd59584fe2c5f379 100644
+index f78714c2a6d9da162c9802552984cbfad76ff728..32c05427858d19a70df40381ee2238f1d34d7190 100644
 --- a/src/main/java/org/bukkit/inventory/meta/ItemMeta.java
 +++ b/src/main/java/org/bukkit/inventory/meta/ItemMeta.java
-@@ -444,4 +444,87 @@ public interface ItemMeta extends Cloneable, ConfigurationSerializable, Persiste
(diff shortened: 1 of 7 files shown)


lynxplay:
I honestly think this might be best to keep for property API.

There is certainly a need for "glowing" but given that mojang does not support this natively without the empty enchant array, exposing this to the API seems a tad bleh, we are kinda just hacking for the plugin user abusing the "quirk" that the client still displays enchantment glint on an empty but present enchantment array.

imo it is a mistake to promote this to a stable API interface like ItemMeta. The property API would make a lot more sense as the user is responsible for the data they input and can hack around in there, making it a lot clearer that what they are doing might not be 100% valid state.

Given we also closed MM's PR for this (#5725), imo this is probably not worth continuing your effort here and just letting property API handle it once out. Obviously just my opinion, I'd love to hear your and other people's thought on this.

Owen1212055:
Thank you for your contribution, however yes I am more inclined to side with lynx's comment... especially with the sensitive nature of itemstacks currently.

TreemanKing:
> There is certainly a need for "glowing" but given that mojang does not support this natively without the empty enchant array, exposing this to the API seems a tad bleh, we are kinda just hacking for the plugin user abusing the "quirk" that the client still displays enchantment glint on an empty but present enchantment array.

So is the client not detecting the following server code?
```java
public boolean isEnchanted() {
        return this.tag != null && this.tag.contains("Enchantments", 9) ? !this.tag.getList("Enchantments", 10).isEmpty() : false; 
}
```
> Given we also closed MM's PR for this (https://github.com/PaperMC/Paper/pull/5725), imo this is probably not worth continuing your effort here and just letting property API handle it once out. Obviously just my opinion, I'd love to hear your and other people's thought on this.

I believe MM made a fake enchantment, whereas this focuses on having either an enchantment or NBT associated with the item.
```java
// This what all the Override items reference, I've just added an additional check to the is Foiled method, similar to that of the isEnchanted
public boolean isFoil(ItemStack stack) {
        return stack.isEnchanted() || stack.isFoiled();
}

public boolean isFoiled() {
        return this.tag != null && this.tag.contains("Foiled") && this.tag.getBoolean("Foiled");
}
```
I believe with a touchup, it would be more than stable as it is essentially the same as the "Unbreakable" tag. Also left it as draft as I think a discussion would be better than just going straight into review.

TreemanKing:
From what it seems from testing, the glowing is handled by the client and the conditions as well so any server side changes won't impact if an item glows.

PR #9695, opened 2023-09-09
Fix sign update desync

ScriptLineStudios:
Stops desync (most likely caused by latency) in the event of a player updating a block which the server hasn't recognized as a sign.


Warriorrrr:
Do you have any link to a place this issue you're fixing has been reported at and/or reproduction instructions? As it stands, this PR has a few issues which makes it not mergable:
- Sending a block entity update packet doesn't do anything in terms of desync
- This will cause an NPE if there is no block entity where the client thinks there is one
- There is no range check done

Owen1212055:
Closing this as in general it's not really certain what you are trying to fix here. Feel free to respond with a more detailed description of your changes, but it's not really understood what kind of "desync" this is resolving here.

PR #9703, opened 2023-09-10
Compression level change, speeds up compression speeds.

PedroMPagani:
You are free to benchmark this on your own, this benchmark from my part was done on DonutSMP events, that get around 700-800 players in a close 400x400 area where there's a lot of Crystal PVP everywhere, lots of chunk interactions etc, compression is highly necessary on environments like this, and a few servers also use this, it's a quick change and most users shouldn't ever feel any need to have this changed but it's a good software improvement if you confirm this should be applied.

This made a buffer compression of around 40kb to 25kb just with little difference from -1, and with MUCH faster speeds.

### DIFF
--- a/patches/server/1031-Compression-Level-Change.patch
+++ b/patches/server/1031-Compression-Level-Change.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: PedroMPagani <pedro.paulo.pagani@gmail.com>
+Date: Sun, 10 Sep 2023 08:39:10 -0300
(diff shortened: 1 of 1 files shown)


kennytv (changes requested):
This should be merged into the original patch `0706-Use-Velocity-compression-and-cipher-natives.patch`, ideally adding a config option for the compression level with the default staying -1 as the middle ground between performance and compression level

PedroMPagani:
I don't have much time to do this and I also haven't touched much of the patch system, if anyone else who wants to become a contributor of the project wants to take this on, feel free to do the required changes ;)

Owen1212055:
Superseded

PR #9709, opened 2023-09-11
Fix MC-868

### LINKED ISSUE #9698: Detector rail is sometimes triggered when a minecart drives next to it.
### Expected behavior

A detector rail is only triggered when a minecart drives over it.

### Observed/Actual behavior

Detector rail is sometimes triggered when a minecart drives next to it.

### Steps/models to reproduce

YT-Video: https://www.youtube.com/watch?v=frW0af0Dhbs
Litematica schematic: [detector rail bug.zip](https://github.com/PaperMC/Paper/files/12566208/detector.rail.bug.zip)

### Plugin and Datapack List

Bukkit Plugins:
GriefPrevention, Multiverse-Core, Multiverse-Inventories, Multiverse-NetherPortals, Multiverse-Portals, Vault

### Paper version

This server is running Paper version git-Paper-169 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: b4e3b3d)\nYou are 2 version(s) behind\nDownload the new version at: https://papermc.io/downloads/paper\nPrevious version: git-Paper-516 (MC: 1.19.4)

### Other

_No response_

TreemanKing:
Fixes MC-868
Adds a check if distance between minecart and detector rail is less than maximum distance of the centre of a block to the edge of a block (0.5)

Closes
#9698

### DIFF
--- a/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
@@ -382,7 +382,7 @@ index 05dfb1790a292f9f85b641377c2ca3675726c127..fe4740cd2a5092fb5c9a597c2ea3774d
              boolean flag1 = false;
              List<AbstractMinecart> list = this.getInteractingMinecartOfType(world, pos, AbstractMinecart.class, (entity) -> {
 -                return true;
-+                return java.awt.geom.Point2D.distance(pos.getX() + 0.5, pos.getZ() + 0.5, entity.getX(), entity.getZ()) < 0.5D; // Paper - MC-868 - Check if distance between minecart and detector rail is less than maximum distance to edge of a block
++                return java.awt.geom.Point2D.distanceSq(pos.getX() + 0.5, pos.getZ() + 0.5, entity.getX(), entity.getZ()) < 0.25D; // Paper - MC-868 - Check if distance between minecart and detector rail is less than maximum distance to edge of a block
              });
  
              if (!list.isEmpty()) {
--- a/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch
@@ -57,6 +57,9 @@ https://bugs.mojang.com/browse/MC-84789
 https://bugs.mojang.com/browse/MC-225381
   Fix overfilled bundles duplicating items / being filled with air
 
+https://bugs.mojang.com/browse/MC-868
+  Fix Detector Rail switches junction before Minecart passes detector (happens only with minecarts of certain speed)
+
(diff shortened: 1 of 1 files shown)


vadcx:
Throughout the MC codebase to avoid the math.sqrt function (expensive) the distance checks are made on squared values.

If the classic formula is math.sqrt((x2-x1)²+(y2-y1)²) < limit

then removing sqrt on both sides of the equation leaves (x2-x1)²+(y2-y1)² < limit²

The ...distanceSquared function should be somewhere in vanilla. Please double-check me on the 0.5d² value.

Machine-Maker (changes requested):
This breaks (possibly) expected behavior for the detector rail as it currently operates. I created the setup below and summoned a minecart at varying points on the track horizontally. In vanilla, I could spawn it on 3/10ths of the adjacent block and it would still trigger the detector rail. With this PR, the position has to be within the detector block x/z coord.

TreemanKing:
> This breaks (possibly) expected behavior for the detector rail as it currently operates. I created the setup below and summoned a minecart at varying points on the track horizontally. In vanilla, I could spawn it on 3/10ths of the adjacent block and it would still trigger the detector rail. With this PR, the position has to be within the detector block x/z coord.

From testing that I did prior to posting the PR, I used the sqrt(2) as the maximum distance (centre to corner). When I had used this, the bug would still persist (considering a velocity of 0.7 or greater from memory), which I believe what current basic vanilla function uses.

As the block should only take the entities from the basic orthogonal directions as it can only face 4 directions, and to this that it's attempting to detect then interacting entity (entity on block). 

I do believe some redstone contraptions that use detector rails should be used as a test to see if any current redstone contraptions will break. Only thing I can think of is if the redstone is required to be on for that fraction of a second longer.

vadcx:
Treeman, did you consider the fix posted in the originial MC-868? [Link to comment by FX - PR0CESS](https://bugs.mojang.com/browse/MC-868?focusedId=1163054&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1163054)

> Now I will mention why this should not be fixed xD
> 
> When changing this code you run into an issue with client rendering. Due to the client using interpolation, the client will render the minecart as if it's turning and then put it in the right place. So that issue should be solved first... which is not an easy issue to fix.

TreemanKing:
> Treeman, did you consider the fix posted in the originial MC-868? [Link to comment by FX - PR0CESS](https://bugs.mojang.com/browse/MC-868?focusedId=1163054&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1163054)
> 
> > Now I will mention why this should not be fixed xD
> > 
> > When changing this code you run into an issue with client rendering. Due to the client using interpolation, the client will render the minecart as if it's turning and then put it in the right place. So that issue should be solved first... which is not an easy issue to fix.

His solution checks a single co-ordinate.

PR #9711, opened 2023-09-11
compression level change with config option

Leguan16:
Supersedes #9703 

As from the author requested: Here a new version with the correct patches and a config option.

The config option is located under the `misc` section. I did not know if I should change the version of the config but that should not stop this PR from being merged. 

Is it possible to add people as co-authors so the original author also gets credits?

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -440,15 +440,16 @@ index 0000000000000000000000000000000000000000..9ef6712c70fcd8912a79f3f61e351aac
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..2f6b0bc82d592acff1cede90956c852c1fcd4ef2
+index 0000000000000000000000000000000000000000..4b3bb5fad1d80c9baea20ca224663b1ebc2b83e9
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,293 @@
+@@ -0,0 +1,294 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
 +import io.papermc.paper.configuration.constraint.Constraint;
 +import io.papermc.paper.configuration.constraint.Constraints;
++import io.papermc.paper.configuration.type.IntOr;
 +import net.kyori.adventure.text.Component;
 +import net.kyori.adventure.text.format.NamedTextColor;
 +import net.minecraft.network.protocol.Packet;
@@ -465,7 +466,7 @@ index 0000000000000000000000000000000000000000..2f6b0bc82d592acff1cede90956c852c
 +
 +@SuppressWarnings({"CanBeFinal", "FieldCanBeLocal", "FieldMayBeFinal", "NotNullFieldNotInitialized", "InnerClassMayBeStatic"})
 +public class GlobalConfiguration extends ConfigurationPart {
-+    static final int CURRENT_VERSION = 29;
++    static final int CURRENT_VERSION = 28;
 +    private static GlobalConfiguration instance;
 +    public static GlobalConfiguration get() {
 +        return instance;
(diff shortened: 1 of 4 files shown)


patches/server/0005-Paper-config-files.patch
```diff
@@ -465,7 +465,7 @@ index 0000000000000000000000000000000000000000..45f1436cdd4b81b621ab71e4336c2aa6
 +
 +@SuppressWarnings({"CanBeFinal", "FieldCanBeLocal", "FieldMayBeFinal", "NotNullFieldNotInitialized", "InnerClassMayBeStatic"})
 +public class GlobalConfiguration extends ConfigurationPart {
-+    static final int CURRENT_VERSION = 28;
++    static final int CURRENT_VERSION = 29;
```
Owen1212055: Don't think we need to update the version here

patches/server/0005-Paper-config-files.patch
```diff
@@ -725,6 +725,7 @@ index 0000000000000000000000000000000000000000..45f1436cdd4b81b621ab71e4336c2aa6
 +        public boolean useAlternativeLuckFormula = false;
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
++        public int compressionLevel = -1;
```
lynxplay: Would probably be better to use IntOrDefault here.
-1 yields default compression for the zip deflater as well as for the native ones velocity implements (6 iirc)

lynxplay (changes requested):

patches/server/0005-Paper-config-files.patch
lynxplay: We don't yea. Version bumps are  only needed if we need a migration to be run on existing data.

lynxplay:
You can generally add a coauthor by simply adding their name into the patch, see e.g. https://github.com/PaperMC/Paper/blob/master/patches/server/0872-Fix-a-bunch-of-vanilla-bugs.patch#L60

Leguan16:
the placement of the option is ok or should I move that as well?

patches/server/0005-Paper-config-files.patch
Leguan16: alright thanks!

lynxplay:
> the placement of the option is ok or should I move that as well?

yea :+1: imo it is.

patches/server/0005-Paper-config-files.patch
Leguan16: Ok so I have tried this and I don't seem to be able to do that.
It tells me that `No TypeSerializer found for field compressionLevel of type class io.papermc.paper.configuration.type.IntOr$Default`, before it just did not want to create the `misc` section at all.
To my eyes I have made everything like where it is already in use. 
The updated `GlobalConfiguration` variable:
```java
public IntOr.Default compressionLevel = IntOr.Default.USE_DEFAULT
```
The updated line in `Connection#setupCompression()`:
```java
[...].create(io.papermc.paper.configuration.GlobalConfiguration.get().misc.compressionLevel.or(-1));
```
Do I have to change anything else besides that?

Leguan16: Ah so i think this is why:
https://github.com/PaperMC/Paper/blob/master/patches/server/0005-Paper-config-files.patch#L1166
(only the world-config has the IntOr Serializers registered)
What are we doing now? Should we create a separate PR adding the stuff to the global config or should I include that here?

lynxplay: include the registration here :+1:

Leguan16: should be fixed now.

lynxplay (approved):

kennytv (approved):

PedroMPagani:
@Leguan16 good job ;), it'd be nice if paper added on the docs examples of the available compression options with example of rates, as a benchmark of course, this should benefit a few servers a lot with bandwidth decreasing as well as more performance and cost decreasing :)

lynxplay:
This does require a PR to the docs before it may be merged yea. @Leguan16 if you have the time, it would be great if you can propose the change there as well :+1:

Leguan16:
I can add a description to the `global-configuration.mdx` what it does but I have no time to benchmark this and create a proper docs page for it. Besides, I don't even know how I should benchmark this. 
@PedroMPagani you seem to know what it does exactly so what should I put in the description (So I have it correct)?

lynxplay:
I don't see the usage for benchmarking on the docs. A quick explanation that a higher compression level means more CPU time spend for less data transmitted and a lower value meaning less CPU time spend but more data transmitted is fine.

Leguan16:
Alright! I will add that later.

Leguan16:
I opened a PR in the docs repo
https://github.com/PaperMC/docs/pull/242

olijeffers0n:
Docs is ready to go 😄

PR #9712, opened 2023-09-12
Dynamic brain activation

kennytv:
Based on https://github.com/TECHNOVE/Airplane/blob/ver/1.17/patches/server/0010-Dynamic-Activation-of-Brain.patch. Modified to not have to manually patch every new entity using Brain (other than adding new config defaults), also skipping unnecessary activity checks

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1480,10 +1480,10 @@ index 0000000000000000000000000000000000000000..f0d4ec73bc8872a85e34f5c6b4d342e7
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..da7c899fbab162ee197a0593f455ebd9c5286d3c
+index 0000000000000000000000000000000000000000..f5dd8ad4c97b8ce4db0d530c34926b465985ef26
 --- /dev/null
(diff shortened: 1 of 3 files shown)


MartijnMuijsers:
I spent a lot of time fielding complaints from players about this: entities in the distance behaving dumbly (piglin brutes being basically pacifist sometimes), mobs dying for all sorts of reasons (see above) like drowning or suffocation, mobs not reacting properly to plugin actions (like being statue-ified for a bit after teleporting), lots of normally Paper-compatible and fairly innocent farms sporadically and unpredictably breaking, or becoming so slow or stuck at typical distances that it is indistinguishable from broken mob AI.

I would suggest not enabling this by default.

Additionally, the `distanceModifier` is really hard to tweak because the effects between its subsequent values are huge. I would suggest, before merging this, to determine a different, better parameter (potentially less performant, but I really don't think a few floating-point computations are going to be a killer here).

And lastly, even when server owners wish to optimize this, the desirable parameter values can be quite different for different mob types and server observations. I would support adding a way to provide more specific values for particular mob type (groups).

electronicboy:
eyes too hurty to say much, but, stuff like that video is basically in part why EAR 2.0 exists, if we want to do this, we'd probably want to look at adopting a similar set of concepts

kennytv:
Honestly yeah I agree after looking into it some more, even defaulting to false seems weird with an option that just breaks entities this hard. Here's non-compromising performance improvements I added put into its own patch (keeping this pr as is for history purposes): https://github.com/PaperMC/Paper/pull/9713

PR #9715, opened 2023-09-12
Fix Projectile#setOwner(null) not clearing owner

### LINKED ISSUE #9710: Projectile#setShooter ignores null values
### Expected behavior

Calling [Projectile#setShooter](https://jd.papermc.io/paper/1.20/org/bukkit/entity/Projectile.html#getShooter()) with a null value as the shooter should remove the shooter from the projectile

### Observed/Actual behavior

Shooter is not removed from the projectile

### Steps/models to reproduce

1. Add this listener to a plugin:
```java
@EventHandler
public void onLaunch(PlayerLaunchProjectileEvent event) {
  event.getProjectile().setShooter(null);
  event.getPlayer().sendPlainMessage("Shooter set to null");
}
```
2. Throw an ender pearl
3. The ender pearl teleports you despite the shooter being set to null in the listener

### Plugin and Datapack List

Irrelevant

### Paper version

[19:38:15 INFO]: Checking version, please wait...
[19:38:15 INFO]: This server is running Paper version git-Paper-174 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: c9cd94f)
Previous version: git-Paper-134 (MC: 1.20.1)

### Other

The implementation uses the `setOwner` method in `net.minecraft.world.entity.projectile.Projectile` which ignores null values.
```java
public void setOwner(@Nullable Entity entity) {
  if (entity != null) {
    this.ownerUUID = entity.getUUID();
    this.cachedOwner = entity;
  }
  this.refreshProjectileSource(false); // Paper
}
```

Warriorrrr:
When a plugin uses Projectile#setShooter(null), the projectileSource will be nulled out, but the cached owner & owner uuid won't be. Additionally, the projectileSource will also be filled back in the next time Projectile#getOwner is called.

Closes #9710


kennytv (approved):

lynxplay (approved):
Potentially mergable with the other fix we have, https://github.com/PaperMC/Paper/blob/master/patches/server/0960-Refresh-ProjectileSource-for-projectiles.patch.

Otherwise, LGTM

PR #9717, opened 2023-09-12
Ensure destination is unloaded when preventing movement into unloaded chunks.

pseudospace0:
When preventing players from moving into unloaded chunks; I would argue that if a scenario arises where the player is attempting to move to a chunk which is indeed loaded, but is sandwiched between a one or more unloaded chunks it would be more beneficial for the method to return in a place where a warning message is explicitly logged rather than just silently not allowing the movement. 

This can be done by ensuring that the location of the movement resides in an unloaded chunk.

### DIFF
--- a/patches/server/1032-Ensure-final-chunk-is-unloaded-when-preventing-movem.patch
+++ b/patches/server/1032-Ensure-final-chunk-is-unloaded-when-preventing-movem.patch
@@ -1,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: pseudospace0 <pseudospace0@proton.me>
+Date: Mon, 11 Sep 2023 21:32:47 +0200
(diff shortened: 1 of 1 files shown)


lynxplay:
No comment from me on this PR, There is no reports of this actually happening, I don't see a point of "fixing" such a hypothetical.

pseudospace0:
> No comment from me on this PR, There is no reports of this actually happening, I don't see a point of "fixing" such a hypothetical.

I wouldn't really refer to it as a "fix" per say, more just a slightly more elegant way of handling things. I would imagine a scenario like this would most likely arise out of a client using some form of movement hacking in which case it would probably be better to log as either "Player moved too fast" or "Player moved wrongly" rather than just returning silently.

lynxplay:
I disagree that allowing clients to seemingly teleport through unloaded chunks is a more elegant way of handling this than the current implementation.

I don't think we are interested in this fix, I hope you can understand where we are coming from here as this is a non issue you are trying to fix.

Thanks non the less!

PR #9724, opened 2023-09-14
Fix BlockRedstoneEvent for tripwire hooks

### LINKED ISSUE #9714: Tripwire hook not registering properly on BlockRedstoneEvents
### Expected behavior

Whenever an entity goes onto a wire connecting two tripwire hooks, there should fire two induvidual BlockRedstoneEvents for the two related tripwire hooks.

For example, when an entity goes onto a tripwire with one end having a tripwire hook facing east, and the other end a tripwire hook facing west, then there should trigger a BlockRedstoneEvent once for both of the tripwire hooks.

### Observed/Actual behavior

Only east facing and north facing tripwire hooks show up in BlockRedstoneEvent. The Event also triggers at least 2 times to the maximum tested value of 3 times (depending on the length of the wire) for the same tripwire hook.

### Steps/models to reproduce

Make a plugin with the following listener
```java
@EventHandler
void onBlockRedstoneEvent(BlockRedstoneEvent event){
    plugin.getLogger().log(Level.INFO, event.getBlock().toString())
}

```

Then make a tripwire contraption (Place tripwire hooks, and strings connecting the two hooks, the connection has to be a straight line)

After that, go into the tripwire and see the logged messages

### Plugin and Datapack List

Only a testplugin with the above mentioned listener

### Paper version

This server is running Paper version git-Paper-176 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 517f8fb)

### Other

_No response_

Lulu13022002:
Closes #9714

Previously the BlockRedstoneEvent was called for every update even non redstone related one and was always unpowered (15 -> 0). Now only two events at least are called (four when cancelled, i'm not 100% sure about) and the redstone current changes are properly respected.


patches/server/1032-Fix-BlockRedstoneEvent-for-tripwire-hooks.patch
```diff
@@ -0,0 +1,72 @@
...
+             Direction enumdirection1 = enumdirection.getOpposite();
+ 
+             world.setBlock(blockposition1, (BlockState) iblockdata3.setValue(TripWireHookBlock.FACING, enumdirection1), 3);
+             this.notifyNeighbors(world, blockposition1, enumdirection1);
+             this.emitState(world, blockposition1, flag4, flag5, flag2, flag3);
++            } // Paper
+         }
+ 
+         // CraftBukkit start
++        if (flag3 != flag5) { // Paper
```
Machine-Maker: Probably some obfhelpers for `flag3` and `flag5` would be nice to have here, and in the if statement added above as well.

lynxplay:
Can you merge this into https://github.com/PaperMC/Paper/blob/master/patches/server/1032-Call-BlockRedstoneEvents-for-lecterns.patch, rename the title to something like "Call BlockRedstoneEvent properly" or something the likes ?

Don't forget smack yourself as a co-author :pray:

SO-32:
Tested on base and pr. Issues reproduced on base for both tripwire hooks and lecterns. Fix confirmed for both in pr

patches/server/1032-Call-BlockRedstoneEvent-properly.patch
```diff
@@ -0,0 +1,99 @@
...
++        BlockRedstoneEvent eventRedstone = new BlockRedstoneEvent(block, wasPowered ? 15 : 0, newCurrent); // Paper
+         world.getCraftServer().getPluginManager().callEvent(eventRedstone);
+ 
+-        if (eventRedstone.getNewCurrent() > 0) {
+-            return;
++            // Paper start
++            if (eventRedstone.getNewCurrent() != newCurrent) {
++                cancelledEmitterHook = true;
++            }
++            // Paper end
```
Warriorrrr: Indent here is a bit weird

Warriorrrr (approved):
This is missing identifying patch comments but rest looks good

PR #9727, opened 2023-09-15
Disallow using items in unloaded chunks 

pseudospace0:
Pretty much the same thing as this: https://github.com/PaperMC/Paper/blob/29d1c7b60244bb002d29a5dcfc9c995019f550ab/patches/server/0281-Don-t-allow-digging-into-unloaded-chunks.patch#L4 just for ServerboundUseItemOnPacket. There isn't really a need to use up any more time handling the packet if its in an unloaded chunk.


lynxplay:
Thank you for your third attempt at ensuring unloaded chunks stay untouched!
As with the two previous ones, this one does not look particularly bullet proof.
If you want to contribute to paper, maybe check for some easier PRs first before touching such delicate logic :+1: 

To potentially save you and us some more time, none of these changes are going to be backported to 1.19 or Folia.
The version is end of life and will at best receive critical bug fixes.

Thank you for your contribution and the good attempts :+1:

PR #9748, opened 2023-09-23
Migrate paper log ips option to new server.properties option

NoahvdAa:


patches/server/0005-Paper-config-files.patch
```diff
@@ -1212,8 +1212,11 @@ index 0000000000000000000000000000000000000000..2d3068b7748032342edd81e6ea4a7c08
 +        for (NodePath path : RemovedConfigurations.REMOVED_GLOBAL_PATHS) {
 +            builder.addAction(path, TransformAction.remove());
 +        }
-+        // ADD FUTURE TRANSFORMS TO builder HERE
```
Machine-Maker, line 1215: Keep the comment like the world config transformations code does

patches/server/0005-Paper-config-files.patch
```diff
@@ -3373,6 +3376,54 @@ index 0000000000000000000000000000000000000000..ef0e834c164b0ccc1a61b349348e6799
...
++
++    @Nullable
++    @Override
++    public Object[] visitPath(final NodePath path, final ConfigurationNode value) throws ConfigurateException {
++        DedicatedServer server = ((DedicatedServer) MinecraftServer.getServer());
++
++        boolean val = value.getBoolean(server.settings.getProperties().logIPs);
++        server.settings.getProperties().logIPs = val;
++        server.settings.getProperties().properties.setProperty("log-ips", String.valueOf(val));
++        server.settings.forceSave();
```
Machine-Maker: Is there some way we can use DedicatedServerSettings#update to avoid needing to remove the final field? Like just change the property and then reconstruct the properties? Maybe its not actually better, but if it is, it removes the need to remove the final

Machine-Maker (changes requested):

patches/server/0005-Paper-config-files.patch
NoahvdAa: Nope, those are specifically mutable values, changing the logIPs value to also be a mutable value would involve a lot more diff and work than just stripping final;
image

Machine-Maker: Yeah, I saw that, but you shouldn't need to change it to be a MutableValue, just do something like 
```java
this.settings.update(properties -> {
    var newProps = properties.cloneProperties();
    newProps.setProperty("logIps", newValue);
    return properties.reload(registryAccess(), newProps, options);
});
```

Machine-Maker (approved):

PR #9754, opened 2023-09-24
Fix brand name always returns null

LemonCaramel:
Honestly, I don't know if it is a good way

### DIFF
--- a/patches/server/0444-Brand-support.patch
+++ b/patches/server/0444-Brand-support.patch
@@ -4,6 +4,19 @@ Date: Sat, 11 Apr 2020 13:10:58 +0200
 Subject: [PATCH] Brand support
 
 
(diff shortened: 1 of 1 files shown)


Warriorrrr (commented):
I was thinking it'd also be possible to set the brand before the instanceof & return in handleCustomPayload, but that would result in the message received event not being called for plugins, so this seems like it'd be the simplest way to solve it at least

electronicboy:
The proper solution would be to actually handle the new type properly instead of just trying to bodge it out

Lulu13022002:
Closing this pr, since this is a hack not a fix and your change doesn't work for the same reason electronicboy's commit doesn't either. However thank you for the report, the issue is still valid.

PR #9756, opened 2023-09-24
Fix missing map initialize event call & missing map id assignment

Warriorrrr:
Fixes maps not having their id field assigned which causes an NPE when a plugin tries to get the id, also fixes the initialize event not being called in this spot.

If it's possible to replace the private -> public with an AT then let me know, wasn't sure how to handle the generic return type


electronicboy (approved):

electronicboy:
AT would just use the erased type, not sure I care though atm

PR #9763, opened 2023-09-25
Fix PlayerSwapHandItemsEvent NPE when a hand set to null

### LINKED ISSUE #9744: PlayerSwapHandItemsEvent throws exception when mainhand or offhand set to null
### Expected behavior

The main or offhand should be set to AIR 

### Observed/Actual behavior

A packet exception is thrown causing dupes in some cases

`Failed to handle packet net.minecraft.network.protocol.game.PacketPlayInBlockDig@4c36c83e, suppressing error
java.lang.NullPointerException: Cannot invoke "org.bukkit.inventory.ItemStack.equals(Object)" because the return value of "org.bukkit.event.player.PlayerSwapHandItemsEvent.getMainHandItem()" is null
        at net.minecraft.server.network.ServerGamePacketListenerImpl.handlePlayerAction(ServerGamePacketListenerImpl.java:1874) ~[?:?]`

### Steps/models to reproduce

Have a sword/any item in your mainhand and any item in your offhand at the same time, hit f while holding the main hand item and have this handler registered
      
      @EventHandler public void onSwap(PlayerSwapHandItemsEvent event) {
        event.setMainHandItem(null);
      }

### Plugin and Datapack List

TestPlugin

### Paper version

This server is running Paper version git-Paper-196 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 773dd72)
Previous version: git-Paper-"189211f" (MC: 1.20.1)

### Other

_No response_

notTamion:
resolves #9744

### DIFF
--- a/patches/api/0441-Fix-PlayerSwapHandItemsEvent-throwing-exception-when.patch
+++ b/patches/api/0441-Fix-PlayerSwapHandItemsEvent-throwing-exception-when.patch
@@ -6,9 +6,18 @@ Subject: [PATCH] Fix PlayerSwapHandItemsEvent throwing exception when mainhand
 
 
 diff --git a/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java b/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
-index 9f592317c920589c22a5fb8e916c6ca58ebe5c59..066e76938ef74e9ca5f3909ffb2864b8581f06f5 100644
+index 9f592317c920589c22a5fb8e916c6ca58ebe5c59..edfa0e1404a38bb0113ecc9ccd9002d6b7dc67b3 100644
 --- a/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
 +++ b/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
+@@ -31,7 +31,7 @@ public class PlayerSwapHandItemsEvent extends PlayerEvent implements Cancellable
+      *
+      * @return item in the main hand
+      */
+-    @Nullable
++    @NotNull
+     public ItemStack getMainHandItem() {
+         return mainHandItem;
+     }
 @@ -42,7 +42,7 @@ public class PlayerSwapHandItemsEvent extends PlayerEvent implements Cancellable
       * @param mainHandItem new item in the main hand
       */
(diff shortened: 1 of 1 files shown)


patches/api/0441-Fix-PlayerSwapHandItemsEvent-throwing-exception-when.patch
```diff
@@ -0,0 +1,29 @@
...
+diff --git a/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java b/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
+index 9f592317c920589c22a5fb8e916c6ca58ebe5c59..8ec2d1689399f9f89684ca4fe145d7553a72dff4 100644
+--- a/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
++++ b/src/main/java/org/bukkit/event/player/PlayerSwapHandItemsEvent.java
+@@ -42,7 +42,7 @@ public class PlayerSwapHandItemsEvent extends PlayerEvent implements Cancellable
+      * @param mainHandItem new item in the main hand
+      */
+     public void setMainHandItem(@Nullable ItemStack mainHandItem) {
+-        this.mainHandItem = mainHandItem;
++        this.mainHandItem = mainHandItem == null ? new ItemStack(org.bukkit.Material.AIR) : mainHandItem; // Paper
```
Machine-Maker: Use `ItemStack#empty()` instead of a new stack of Air so the amount is set to 0. Same for below.

Machine-Maker (changes requested):

patches/api/0441-Fix-PlayerSwapHandItemsEvent-throwing-exception-when.patch
notTamion: Thanks for the improvements

Warriorrrr (approved):
I was thinking about moving the .equals around to prevent the npe but this also works, so lgtm

Machine-Maker:
I think this fix is better because it’s less nullable itemstacks. Now we can also change the annotation on the getter to NotNull. The ItemStack passed into the event already was NotNull, so that doesn’t break behavior expectations.

Machine-Maker (approved):

Sytm:
Are Paper comments required behind the nullability annotations?

lynxplay (approved):

PR #9770, opened 2023-09-28
Fixup player profile getters and constructor to expected nullability

kennytv:
Changing nullability here would go beyond just getters with its creation and completion, so that feels more like a future thing

### DIFF
--- a/patches/server/0141-Basic-PlayerProfile-API.patch
+++ b/patches/server/0141-Basic-PlayerProfile-API.patch
@@ -615,7 +615,7 @@ index c70cd016e1978931d115cfca94664897f0158196..eac9658fa4cab7a651e10e4e18c679e0
          String s1 = name.toLowerCase(Locale.ROOT);
          GameProfileCache.GameProfileInfo usercache_usercacheentry = (GameProfileCache.GameProfileInfo) this.profilesByName.get(s1);
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
-index 0ef8750c7c862a44dfb0e15602ef819790c9f1a4..cfb0846381f5ebfe33fc4074f21fb0993812f75c 100644
+index 0ef8750c7c862a44dfb0e15602ef819790c9f1a4..a230e2ac2de48d4d5d963e1de2bd87999b4ad2fc 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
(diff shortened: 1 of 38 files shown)


patches/server/0141-Basic-PlayerProfile-API.patch
```diff
@@ -103,14 +103,14 @@ index 0000000000000000000000000000000000000000..e513ce5bcb5070b8a57737228904381f
...
 +    }
 +
 +    @Override
 +    @Deprecated(forRemoval = true)
 +    public UUID setId(@Nullable UUID uuid) {
 +        GameProfile prev = this.profile;
-+        this.profile = new GameProfile(uuid, prev.getName());
++        this.profile = new GameProfile(uuid != null ? uuid : Util.NIL_UUID, prev.getName());
 +        copyProfileProperties(prev, this.profile);
 +        return prev.getId();
```
Lulu13022002: This maybe needs to be updated to avoid a contract change from nullable to notnull?

rudde0:
Is this PR fixes the [addressed](https://hub.spigotmc.org/jira/browse/SPIGOT-7492) issue?

Lulu13022002:
I think the method Server#createProfileExact should also be updated, the method specify than having one of its parameter null is valid. However when the method will find an online player and one of its parameter is null, the constructor will throw. For example: `createProfileExact(null, p.getName())`, `createProfileExact(p.getUniqueId(), null)`

WindSpiritSR:
@Lulu13022002

patches/server/0141-Basic-PlayerProfile-API.patch
```diff
@@ -103,16 +103,17 @@ index 0000000000000000000000000000000000000000..e513ce5bcb5070b8a57737228904381f
 +    @Nullable
 +    @Override
 +    public UUID getId() {
-+        return profile.getId();
++        return profile.getId() != Util.NIL_UUID ? profile.getId() : null;
```
Lulu13022002: I think doing an equals instead of a ref equality for the uuid check would be better to have a consistent state across the server since vanilla and spigot does that as well.

Lulu13022002 (commented):
The two linked issues above seems relevant and could be fixed here (even if it's a spigot one). The method SkullMeta#setOwningPlayer didn't throw before with an offline player without a name:
```Java
stack.editMeta(SkullMeta.class, meta -> {
    meta.setOwningPlayer(Bukkit.getOfflinePlayer(UUID.fromString("a8179ff3-c201-4a75-bdaa-9d14aca6f83f")));
});
```
Also this patch need to be reevaluate (https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/browse/CraftBukkit-Patches/0055-Fix-Player-Banning.patch): cause it can throw in the case it's still relevant

kennytv:
The other issues are better fixed in another pr/of lesser priority

PR #9776, opened 2023-09-29
Add support for custom titles to HumanEntity inventory open methods

TrollyLoki:
Adds additional overloads to the methods in the HumanEntity interface that open inventories (such as openAnvil, openWorkbench, ect.) allowing for customization of the displayed titles.


Sytm:
I already opened a PR that results in the same behaviour. See #9658

Warriorrrr:
Thank you for your contribution, unfortunately this PR seems to do roughly the same thing as the existing #9658, so this is being closed in favour of that one.

PR #9780, opened 2023-09-30
Fix strikeLightningEffect powers lightning rods & clears copper

### LINKED ISSUE #9736: World#strikeLightningEffect powers lightning rods and clears copper
### Expected behavior

A purely visual lightning bolt should neither power lightning rods nor clear weathering copper blocks in the world.

### Observed/Actual behavior

Lightning rods and copper blocks are mutated by a visual lightning strike.

### Steps/models to reproduce

Summon a lightning strike using [World#strikeLightningEffect](https://jd.papermc.io/paper/1.20/org/bukkit/World.html#strikeLightningEffect(org.bukkit.Location)) onto weathered copper blocks or lightning rods.

Observe mutation.

### Plugin and Datapack List

Nan

### Paper version

1b1c23010ac55e610e3a36f4e5b93a8198cff69c

### Other

A follow up to #8886, as the silent boolean, while poorly named, still has the function that visual lightning *should* have.
Specifically, net/minecraft/world/entity/LightningBolt.java:89 only respects the isSilent boolean, not the visual flag.

notTamion:
resolves #9736 
Not sure if the check for isSilent should be removed i am just going to leave it in for now incase any plugins are using the deprecated methods as a workaround

### DIFF
--- a/patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
+++ b/patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Fix strikeLightningEffect powers lightning rods and clears
 
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
-index 17610196db7a1c6feb2cf74a02479a8691aa323f..ca811a3dcc751fcf663b5ef49c892097b20bfb83 100644
+index 17610196db7a1c6feb2cf74a02479a8691aa323f..ef7a721df01723cc64c2ddffa536cd842997282b 100644
 --- a/src/main/java/net/minecraft/server/level/ServerLevel.java
 +++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
 @@ -997,7 +997,7 @@ public class ServerLevel extends Level implements WorldGenLevel {
@@ -14,12 +14,12 @@ index 17610196db7a1c6feb2cf74a02479a8691aa323f..ca811a3dcc751fcf663b5ef49c892097
                  if (entitylightning != null) {
                      entitylightning.moveTo(Vec3.atBottomCenterOf(blockposition));
 -                    entitylightning.setVisualOnly(flag1);
-+                    entitylightning.setSkeletonTrap(flag1); // Paper
++                    entitylightning.skeletonTrap = flag1; // Paper
                      this.strikeLightning(entitylightning, org.bukkit.event.weather.LightningStrikeEvent.Cause.WEATHER); // CraftBukkit
                  }
              }
 diff --git a/src/main/java/net/minecraft/world/entity/LightningBolt.java b/src/main/java/net/minecraft/world/entity/LightningBolt.java
-index 471275c5362b61ce8b5b9dd5c85b3e93cabd3f76..14244fb611ce0c68429831b42d36e412df73a49a 100644
+index 471275c5362b61ce8b5b9dd5c85b3e93cabd3f76..829778169b74e1543d034fd3294ce39151f48800 100644
 --- a/src/main/java/net/minecraft/world/entity/LightningBolt.java
 +++ b/src/main/java/net/minecraft/world/entity/LightningBolt.java
 @@ -46,6 +46,7 @@ public class LightningBolt extends Entity {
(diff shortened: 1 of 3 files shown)


kennytv:
Thanks - I think another if needs the visual check where it is able to set fire to the world

notTamion:
Is it also in the LightningBolt class? Incase you are talking about the spawnFire method, that one already got a check in it.
When testing the changes it doesn't set anything on fire. Or is there something i am missing?

kennytv (approved):
Oh yeah right, thanks!

Machine-Maker (changes requested):
This is a vanilla behavior break. The vanilla `visualOnly` boolean is exclusively used for the skeleton trap lightning, which in vanilla, does affect lightning rods and copper blocks. To add the check for it here would change that behavior.

notTamion:
Would creating an extra boolean and using that for the skeleton trap be an good way of fixing this? This way we can use the visualOnly boolean for the strikeLightningEffect method. This also fixes misleading variable names. Is this a good way to do it or are there better ways?

notTamion:
Fixed the behavior break by creating an extra skeletonTrap boolean that is used for skeleton traps instead of the visualOnly boolean. This way we also avoid confusion with variable names. I hope this way of fixing it is acceptable

patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
```diff
@@ -0,0 +1,105 @@
...
+@@ -133,10 +140,10 @@ public class LightningBolt extends Entity {
+             }
+         }
+ 
+-        if (this.life >= 0 && !this.visualOnly) { // CraftBukkit - add !this.visualOnly
++        if (this.life >= 0 && !this.visualOnly && !this.skeletonTrap) { // CraftBukkit - add !this.visualOnly // Paper
+             if (!(this.level() instanceof ServerLevel)) {
+                 this.level().setSkyFlashTime(2);
+-            } else if (!this.visualOnly) {
++            } else { // Paper
```
Lulu13022002: This diff is not needed it's better indeed but this will be part of the spigot cleanup it shouldn't be there

patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
```diff
@@ -0,0 +1,105 @@
...
+     private void spawnFire(int spreadAttempts) {
+-        if (!this.visualOnly && !this.level().isClientSide && this.level().getGameRules().getBoolean(GameRules.RULE_DOFIRETICK)) {
++        if (!this.visualOnly && !this.skeletonTrap && !this.level().isClientSide && this.level().getGameRules().getBoolean(GameRules.RULE_DOFIRETICK)) { // Paper
+             BlockPos blockposition = this.blockPosition();
+             BlockState iblockdata = BaseFireBlock.getState(this.level(), blockposition);
+ 
+             if (this.level().getBlockState(blockposition).isAir() && iblockdata.canSurvive(this.level(), blockposition)) {
+                 // CraftBukkit start - add "!visualOnly"
+-                if (!this.visualOnly && !CraftEventFactory.callBlockIgniteEvent(this.level(), blockposition, this).isCancelled()) {
++                if (!CraftEventFactory.callBlockIgniteEvent(this.level(), blockposition, this).isCancelled()) { // Paper
```
Lulu13022002: Same here

patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
```diff
@@ -0,0 +1,105 @@
...
+                     this.level().setBlockAndUpdate(blockposition, iblockdata);
+                     ++this.blocksSetOnFire;
+                 }
+@@ -181,7 +188,7 @@ public class LightningBolt extends Entity {
+                 iblockdata = BaseFireBlock.getState(this.level(), blockposition1);
+                 if (this.level().getBlockState(blockposition1).isAir() && iblockdata.canSurvive(this.level(), blockposition1)) {
+                     // CraftBukkit start - add "!visualOnly"
+-                    if (!this.visualOnly && !CraftEventFactory.callBlockIgniteEvent(this.level(), blockposition1, this).isCancelled()) {
++                    if (!CraftEventFactory.callBlockIgniteEvent(this.level(), blockposition1, this).isCancelled()) { // Paper
```
Lulu13022002: Same here

patches/server/1039-Fix-strikeLightningEffect-powers-lightning-rods-and-.patch
```diff
@@ -0,0 +1,105 @@
...
+         super(type, world);
+@@ -59,6 +60,12 @@ public class LightningBolt extends Entity {
+         this.visualOnly = cosmetic;
+     }
+ 
++    // Paper start
++    public void setSkeletonTrap(boolean skeletonTrap) {
++        this.skeletonTrap = skeletonTrap;
++    }
++    //Paper end
```
Lulu13022002: Missing a space between // and Paper end, also the setter is not really needed if the field is already public so you can either make the field private and use the setter or directly use the field access

Machine-Maker:
Vanilla's `visualOnly` field **is** the skeleton trap boolean. All other uses of it for anything else should be reverted. Instead of using the vanilla field for what it originally ways, we are just changing its meaning, and adding a field to replace it. It should be the other way around. Add an "isEffect" boolean that is checked in the correct places, and remove `visualOnly` from anywhere it's not used in vanilla. Then make the API isEffect method return that new boolean.

Machine-Maker (approved):

PR #9786, opened 2023-10-02
Add API around client-side only Signs

Sytm:
Signs are commonly used as multi-line text input, so this should be part of the API.

Anvils are also frequently used for that, but they only offer single-line text (although they are already part of the Paper API)

### DIFF
--- a/patches/api/0484-Add-API-for-client-side-signs.patch
+++ b/patches/api/0484-Add-API-for-client-side-signs.patch
@@ -1,0 +1,121 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lukas Planz <me@md5lukas.de>
+Date: Mon, 2 Oct 2023 15:47:27 +0200
+Subject: [PATCH] Add API for client-side signs
+
+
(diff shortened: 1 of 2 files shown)


lynxplay (changes requested):
Given that there is a `PlayerUseUnknownEntityEvent`, I think this is better left like it initially was, only called for unknown signs to not interfere with "normal" API usage.

Sytm:
My reasoning for doing it this way has been that while the unknown entities are known to be actually unknown because of the continually increasing entity id, there could be already a sign placed at the location where the local sign should be opened, so when he closes the sign editor the sign change handling would be passed to the real api which shoots it doen before the event is called because he wasn't allowed to do so.

This wouldn't require searching for a free space around the player to send him a sign and unify the possible handling of them with one event instead of maybe two (if it were to actually work)

Sytm:
So what's the current stance on this PR? I'd still like to see it added in this form, but I haven't gotten an answer on my previous comment

Owen1212055:
Hmm, I agree something like this is certainly useful. But the current event name is defo a bit misleading as we definitely want to make sure there is a distinction as people might blindly listen to this event.

I agree that it might be useful to have an "unknown" one and then a normal one.

Sytm:
I have changed the name of the event to differentiate it more from the normal event.

I am still not sold on the concept to only call this event when there isn't an actual sign to edit, because of the points I already mentioned above.

PR #9787, opened 2023-10-02
Reworked adding native map methods to PDC

Xander-de-Keijzer:
Linked issue: #8925
As the previous pull request #9064 seems to have gone inactive I have reworked the changes, expanded the documentation and applied the requested changes from that pull request.

PDC now inherits Iterable<NamespacedKey>, and adds these methods

    #setIfAbsent(Key, PersistentDataType, Object)
    #computeIfAbsent(Key, PersistentDataType, Function)
    #computeIfPresent(Key, PersistentDataType, BiFunction)
    #compute(Key, PersistentDataType, BiFunction)
    #getOrThrow(Key, PersistentDataType, Supplier)
    #getOrThrow(Key, PersistentDataType)
    #replace(Key, PersistentDataType, Object, Object)
    #replace(Key, PersistentDataType, Object)
    #merge(Key, PersistentDataType, Object, BiFunction)


olijeffers0n:
The actual patch names are weird btw. I'm a fan of these methods though - you should add attribution for the previous codeowner though as a co-author in the patch description please.

imDaniX:
Can you also look into implementing [`merge`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#merge(K,V,java.util.function.BiFunction)) pretty please?

Xander-de-Keijzer:
> The actual patch names are weird btw. I'm a fan of these methods though - you should add attribution for the previous codeowner though as a co-author in the patch description please.

> Can you also look into implementing [`merge`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#merge(K,V,java.util.function.BiFunction)) pretty please?

Added previous codeowner as Co-author, renamed the patches and added the merge method

patches/api/0441-Expose-native-map-methods-in-PDC.patch
```diff
@@ -0,0 +1,288 @@
...
+diff --git a/src/main/java/org/bukkit/persistence/PersistentDataContainer.java b/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
+index 57609b7793122e135fa0c3b926500849379637b2..b221b97fd540b1f16aea3d6fce72d8a09c7a5885 100644
+--- a/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
++++ b/src/main/java/org/bukkit/persistence/PersistentDataContainer.java
+@@ -9,7 +9,7 @@ import org.jetbrains.annotations.Nullable;
+  * This interface represents a map like object, capable of storing custom tags
+  * in it.
+  */
+-public interface PersistentDataContainer {
++public interface PersistentDataContainer extends Iterable<net.kyori.adventure.key.Key> {
```
Machine-Maker: I don't think its appropriate to implement Iterable, especially when the iterator's remove doesn't actually remove the key from the PDC because the `keySet` set isn't backed by the pdc map.

patches/api/0441-Expose-native-map-methods-in-PDC.patch
```diff
@@ -0,0 +1,288 @@
...
++     * Get a set of {@link net.kyori.adventure.key.Key}s present on this {@link PersistentDataContainer}
++     * instance.
++     * <p>
++     * Any changes made to the returned set will not be reflected on the
++     * instance.
++     *
++     * @return the key set
++     */
++    @NotNull
++    Set<net.kyori.adventure.key.Key> keySet();
```
Machine-Maker: I don't think we have to add this either, especially since it doesn't follow `Map#keySet`'s contract. Plus, I don't really see a ton of use for doing this which you can't already do and we don't want to encourage a ton of messing with other plugin's PDC.

patches/api/0441-Expose-native-map-methods-in-PDC.patch
```diff
@@ -0,0 +1,288 @@
...
++     * @param <T>   the generic java type of the tag value
++     * @param <Z>   the generic type of the object to store
++     *
++     * @throws NullPointerException     if the key is null
++     * @throws NullPointerException     if the type is null
++     * @throws NullPointerException     if the value is null
++     * @throws IllegalArgumentException if no suitable adapter will be found for
++     *                                  the {@link PersistentDataType#getPrimitiveType()}
++     */
++    <T, Z> void setIfAbsent(@NotNull net.kyori.adventure.key.Key key, @NotNull PersistentDataType<T, Z> type, @NotNull Z value);
```
Machine-Maker: ~~This isn't a Map method and I think everything it does is covered by `computeIfAbsent` and `Functions.constant()`.~~

Apparently putIfAbsent is a thing, so it should be called that.

patches/server/1036-Expose-native-map-methods-in-PDC.patch
```diff
@@ -0,0 +1,250 @@
...
++import net.kyori.adventure.key.Key;
+ import net.minecraft.nbt.CompoundTag;
+ import net.minecraft.nbt.Tag;
+ import org.bukkit.NamespacedKey;
+@@ -14,6 +18,8 @@ import org.bukkit.craftbukkit.util.CraftNBTTagConfigSerializer;
+ import org.bukkit.persistence.PersistentDataAdapterContext;
+ import org.bukkit.persistence.PersistentDataContainer;
+ import org.bukkit.persistence.PersistentDataType;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
```
Machine-Maker: Inline the imports

patches/server/1036-Expose-native-map-methods-in-PDC.patch
```diff
@@ -0,0 +1,250 @@
...
+         Preconditions.checkArgument(key != null, "The NamespacedKey key cannot be null");
+         Preconditions.checkArgument(type != null, "The provided type cannot be null");
+         Preconditions.checkArgument(value != null, "The provided value cannot be null");
++        // Paper start
+ 
+-        this.customDataTags.put(key.toString(), this.registry.wrap(type.getPrimitiveType(), type.toPrimitive(value, adapterContext)));
++        set(key.toString(), type, value);
++    }
++
++    private <T, Z> void set(String key, PersistentDataType<T, Z> type, Z value) {
```
Machine-Maker: Instead of taking a String, these methods can actually just take an adventure `Key`. That way the entire method body of `get`, `has`, `remove`, etc. can be moved into the private methods, and the public implementations can delegate to those private methods. It simplifies the diff dramatically. Make sure to change the toString() methods to asString() for the adventure Key tho.

Machine-Maker (changes requested):
Make sure Paper start and end comments surround areas of diff, there are quite a few missing.


This is also something which is very easily unit-testable. It could be nice to have unit tests covering the various methods to ensure they function correctly.

patches/api/0441-Expose-native-map-methods-in-PDC.patch
Xander-de-Keijzer: I think the previous author chose this name because PDC uses set() instead of put() but I'm not sure

lynxplay: I think if we are moving most of this to me as map like as possible, I think it would be best to use put here.

Xander-de-Keijzer: So should we change the set method to put then to, or add put as the adventure key method to not break api?

Xander-de-Keijzer: So:
`set(NamespacedKey, ...)`
`put(Key, ...)`
`put(String, ...)`?

lynxplay: We certainly cannot rename the method and break API.
I think it is honestly fine to leave it as put.
A "mistake" made when naming it, but certainly not remotely bad enough to warrent a rename.

But if we add new stuff, I think it is a good idea to stick to the map like naming if that is what the new methods aim to achieve.

PR #9789, opened 2023-10-03
Downgrade bundled SQLite as per https://hub.spigotmc.org/jira/browse/SPIGOT-7491

Benjamin-Dobell:
`org.xerial.sqlite-jdbc` released a breaking change for a fairly commonly used API — used extensively in [plotsquared for example](https://github.com/IntellectualSites/PlotSquared/blob/26c55a318fac673f629faef3f0e7379893ced0d5/Core/src/main/java/com/plotsquared/core/database/SQLManager.java#L1078). CraftBukkit initially picked up this update but it has since been rolled back (https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/40d41acc1aba91ead44adc469d715debe6f041a8#pom.xml).

Unfortunately, Paper [updated CraftBukkit prior to rollback](https://github.com/PaperMC/Paper/commit/4cdbb0c86c40f5ff93a9a7cdca3102636b35b6d4).

This PR rolls back to `org.xerial:sqlite-jdbc:3.42.0.1` everywhere and upgrades Bukkit, CraftBukkit and Spigot again, so that CraftBukkit plugins that make use of SQLIte don't break.

In the process a few patches needed to be updated because the changes they introduced have since been upstreamed.


olijeffers0n:
If i am understanding this PR correctly, you should just wait until we pull upstream again

electronicboy:
This is an upstream merge PR, they just generally failed to entitle it as such

kennytv:
We generally like to do upstream updates ourselves, especially when Spigot adds new API we already have, sorry

Benjamin-Dobell:
@kennytv No problem at all. 👍 This PR is at least is here for reference, both for upgrading to the latest upstream and on the off chance someone happens to try HEAD wonders what's up with SQLite 😅

PR #9792, opened 2023-10-04
Add consumer parameter for spawning falling blocks

booky10:
Bukkit already has this consumer for spawning normal entities and dropping items, but the method for spawning falling blocks is missing it

### DIFF
--- a/patches/api/0441-Add-consumer-parameter-for-spawning-falling-blocks.patch
+++ b/patches/api/0441-Add-consumer-parameter-for-spawning-falling-blocks.patch
@@ -1,0 +1,34 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: booky10 <boooky10@gmail.com>
+Date: Sun, 29 Jan 2023 01:38:41 +0100
+Subject: [PATCH] Add consumer parameter for spawning falling blocks
+
+
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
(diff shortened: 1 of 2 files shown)


Malfrador:
I would personally prefer having a `setBlockData` method in `FallingBlock`, so that its possible to just use the existing consumer-based methods with FallingBlocks. 

FallingBlocks being the only entity that require a special spawning method in `World` is kinda weird, and the only reason for this is that they require BlockData to be spawned, which was impossible before the consumer methods existed.

booky10:
Yeah, would make sense to add api for changing the blockstate.

But still, the falling block spawning method is the only one which doesn't have a consumer parameter. This seems like an inconsistency to me.

Machine-Maker:
I agree with @Malfrador that a setBlockData method on FallingBlock is a better solution. I don't really like all the unique spawn methods for various things, Paper already has changes to make other entities work with the general spawn methods. Not sure we should be adding more.

booky10:
So should I scrap the current PR and implement the proposed alternative solution?
Also, should the specific spawning-methods (for spawning falling blocks/dropping items) be deprecated/marked as obsolete?

lynxplay:
If we manage to get a FallingBlock#setBlockData (or the likes) to properly work, then yea, we could mark the special spawning method as `@Obsolete` and link the proper methods.

I'd keep this branch around tho, if implementing the setBlockData method seems too troublesome/runs into client restrictions down the line, we might just have to fall back onto this proposal.

PR #9806, opened 2023-10-07
Fix bugs in collision detection when itemframe is outside the worldborder

### LINKED ISSUE #9805: The Item frame outside the worldborder drops abnormally
### Expected behavior

same as vanilla the item frame does not drop

### Observed/Actual behavior

spigot same as vanilla 

all of the paper fork has the problem.

when you let worldboder behand the item frame, it will drop without reason.
93C50C369CF458D846436F10A04F0988

### Steps/models to reproduce

1.put a item frame
2.move world border behand item frame
3.wait

### Plugin and Datapack List

> datapack list
[21:23:29 INFO]: There are 6 data pack(s) enabled: [vanilla (built-in)], [trade_rebalance (feature)], [file/Structory_1.20_v1.3.3.zip (world)], [file/Terralith_1.20_v2.4.6.zip (world)], [file/XKendermanFix.zip (world)], [file/bukkit (world)]

> plugins
[21:24:01 INFO]: Server Plugins (2):
[21:24:01 INFO]: Bukkit Plugins:
[21:24:01 INFO]:  - Chunky, FreedomChat

### Paper version

> version
[21:24:18 INFO]: Checking version, please wait...
[21:24:21 INFO]: This server is running Paper version git-Paper-217 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: cfe311d)

### Other

_No response_

Hydrostic:
repair for https://github.com/PaperMC/Paper/issues/9805
in `net.minecraft.world.entity.decoration.ItemFrame#survives`
```
 @Override
    public boolean survives() {
        if (this.fixed) {
            return true;
        } else if (!this.level().noCollision((Entity) this)) {
            return false;
        } else {
            BlockState iblockdata = this.level().getBlockState(this.pos.relative(this.direction.getOpposite()));

            return !iblockdata.isSolid() && (!this.direction.getAxis().isHorizontal() || !DiodeBlock.isDiode(iblockdata)) ? false : this.level().getEntities((Entity) this, this.getBoundingBox(), ItemFrame.HANGING_ENTITY).isEmpty();
        }
    }
```
the itemframe will detect whether itself collides with other items by calling `noCollision`, but in `net.minecraft.world.level.Level#noCollision(Entity, AABB)`, which has been modified by paper, hasn't consider the situation that itemframe is outside the world border and directly call the func `getCollisionsForBlocksOrWorldBorder`, which return false when entity is outside the world border
```
// Paper start - Prevent armor stands from doing entity lookups
    @Override
    public boolean noCollision(@Nullable Entity entity, AABB box) {
        if (entity instanceof net.minecraft.world.entity.decoration.ArmorStand && !entity.level().paperConfig().entities.armorStands.doCollisionEntityLookups) return false;
        // Paper start - optimise collisions
        int flags = io.papermc.paper.util.CollisionUtil.COLLISION_FLAG_CHECK_ONLY;
        if (entity != null) {
            flags |= io.papermc.paper.util.CollisionUtil.COLLISION_FLAG_CHECK_BORDER;
        }
        if (io.papermc.paper.util.CollisionUtil.getCollisionsForBlocksOrWorldBorder(this, entity, box, null, null, flags, null)) {
            return false;
        }

        return !io.papermc.paper.util.CollisionUtil.getEntityHardCollisions(this, entity, box, null, flags, null);
        // Paper end - optimise collisions
    }
    // Paper end
```
so I change the patch to
```
if (!(entity instanceof net.minecraft.world.entity.decoration.ItemFrame) && io.papermc.paper.util.CollisionUtil.getCollisionsForBlocksOrWorldBorder(this, entity, box, null, null, flags, null)) {
+            return false;
+        }
```
to prevent item frame from killed

### DIFF
--- a/patches/server/0727-Collision-optimisations.patch
+++ b/patches/server/0727-Collision-optimisations.patch
@@ -2629,7 +2629,7 @@ index a25497eec004add7408a63b1a0f09e3fa443b324..9f892de55ab03367daed4c30cc44c9dd
  
      // Paper start
 diff --git a/src/main/java/net/minecraft/world/level/Level.java b/src/main/java/net/minecraft/world/level/Level.java
-index 270ce3995229aa79074e981bb45e5480a5e924d4..627758e4f55283c5dcf23d494efb92bfa9f6b005 100644
+index 270ce3995229aa79074e981bb45e5480a5e924d4..3163058a28092b1708d58cdc54f9ca8614a4d4ac 100644
 --- a/src/main/java/net/minecraft/world/level/Level.java
 +++ b/src/main/java/net/minecraft/world/level/Level.java
 @@ -297,6 +297,10 @@ public abstract class Level implements LevelAccessor, AutoCloseable {
@@ -3024,7 +3024,7 @@ index 270ce3995229aa79074e981bb45e5480a5e924d4..627758e4f55283c5dcf23d494efb92bf
 +        if (entity != null) {
 +            flags |= io.papermc.paper.util.CollisionUtil.COLLISION_FLAG_CHECK_BORDER;
 +        }
-+        if (io.papermc.paper.util.CollisionUtil.getCollisionsForBlocksOrWorldBorder(this, entity, box, null, null, flags, null)) {
++        if (!(entity instanceof net.minecraft.world.entity.decoration.ItemFrame) && io.papermc.paper.util.CollisionUtil.getCollisionsForBlocksOrWorldBorder(this, entity, box, null, null, flags, null)) {
 +            return false;
 +        }
 +


lynxplay:
I wonder if we can somehow pass if the entity noCollision should use the `COLLISION_FLAG_CHECK_BORDER` flag or not.
The instanceOf, while presumably fine, seems kind of meh there given we know the exact spot this check would have to trigger for so we could attempt to shortcut here a bit.

My brain is absolutely friend right now tho, so I sadly cannot look deeper into that.

Lulu13022002:
Closing this pr since this is hack not a fix, the proper way is to do what vanilla do instead of what the new system do. There's more issue with the world border collision. The item frame do indeed drop when near the border (<2*|boxSize|) but not when far outside of the border and that's the issue. Essentially in vanilla an entity collide with a border without collide epsilon and when the entity is near the border and inside (with a margin) but not far outside then the two voxel shape intersection are checked.

Hydrostic:
> Closing this pr since this is a hack not a fix, the proper way is to do what vanilla do instead of what the new system do. There's more issue with the world border collision. The item frame do indeed drop when near the border (<2*|boxSize|) but not when far outside of the border and that's the issue. Essentially in vanilla an entity collide with a border without collide epsilon and when the entity is near the border and inside (with a margin) but not far outside then the two voxel shape intersection are checked.

If the border is set to expand automatically e.g. /worldborder add 10000 10，then will the moving border cause item frame to disappear？

Lulu13022002:
Yes, this might happen if the timing is right. For example if the border is slow enough that the item frame check its state (that happens each 100 ticks by default) and the condition are met.

PR #9815, opened 2023-10-10
Make setVelocity method of Fireballs change the travel direction to an arbitrary vector

TrollyLoki:
Currently calling the setVelocity() has undefined behavior at best, and in addition there is no way to set a fireballs direction to anything other than a normalized vector. This PR fixes both of these problems at once by overriding the setVelocity method of fireballs to set the travel direction while bypassing the vanilla normalization, allowing for things such as freezing a fireball in-place or changing its speed.

### DIFF
--- a/patches/server/1037-Make-setVelocity-method-of-Fireballs-change-the-trav.patch
+++ b/patches/server/1037-Make-setVelocity-method-of-Fireballs-change-the-trav.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Make setVelocity method of Fireballs change the travel
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java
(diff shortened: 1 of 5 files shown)


Machine-Maker (approved):

lynxplay (approved):

patches/server/1037-Make-setVelocity-method-of-Fireballs-change-the-trav.patch
```diff
@@ -0,0 +1,30 @@
...
+     }
+ 
++    // Paper start - set direction without normalizing
++    @Override
++    public void setVelocity(Vector velocity) {
++        Preconditions.checkArgument(velocity != null, "Vector velocity cannot be null");
++        velocity.checkFinite();
++        this.getHandle().xPower = velocity.getX();
++        this.getHandle().yPower = velocity.getY();
++        this.getHandle().zPower = velocity.getZ();
```
mcmonkey4eva, line 23: X/Y/Z Power is Acceleration, not Velocity - this PR causes fireballs to accelerate continually over time if you set the velocity

PR #9821, opened 2023-10-11
Add API to get the collision shape of a block before it's placed

TrollyLoki:


patches/api/0442-Add-API-to-get-the-collision-shape-of-a-block-before.patch
```diff
@@ -0,0 +1,29 @@
...
++++ b/src/main/java/org/bukkit/block/data/BlockData.java
+@@ -204,6 +204,17 @@ public interface BlockData extends Cloneable {
+      */
+     boolean isFaceSturdy(@NotNull BlockFace face, @NotNull BlockSupport support);
+ 
++    // Paper start
++    /**
++     * Gets the block collision shape that would result if this block data were placed at a particular location.
++     *
++     * @param location the location to simulate placing the block at
```
Machine-Maker: I don't think `simulate` is the correct word. It feels like it means its doing more than just checking what the collision would be. I think a better description would be `the location used to calculate the collision shape`, or something like that.

Machine-Maker (changes requested):

Machine-Maker (approved):

patches/api/0442-Add-API-to-get-the-collision-shape-of-a-block-before.patch
```diff
@@ -0,0 +1,29 @@
...
+index 31111cea5ffd018c3c011c1f3b8befbbd33db5e5..a46b494f66bf8128303a65feec50ef92d88014b1 100644
+--- a/src/main/java/org/bukkit/block/data/BlockData.java
++++ b/src/main/java/org/bukkit/block/data/BlockData.java
+@@ -204,6 +204,17 @@ public interface BlockData extends Cloneable {
+      */
+     boolean isFaceSturdy(@NotNull BlockFace face, @NotNull BlockSupport support);
+ 
++    // Paper start
++    /**
++     * Gets the block collision shape that would result if this block data were placed at a particular location.
```
lynxplay: I think the javadocs here are somewhat misleading.
The method only computes the collision shape of the block data in the passed location without any placement like logic running.

That is fine, I can 100% see the usage of this method, but I think we need to be very detailed about *what* this method does.
From a) the name and b) the documentation it is not instantly obvious that this, e.g., would not have a Material.OAK_FENCE.createBlockData() return the shape a fence would be in at location A if a connecting fence is next to it.

The docs are in that regard "technically" correct, the fence block data would have no direction marked and would have a single bounding box of a fence that is not connected, but most people would not expect this behaviour.

I think we'd save ourself and plugin developers down the line a lot of time if we further explain the logic this method provides :)

TrollyLoki:
I removed the word placed and replaced it with calculated as I did in the parameter description. Also added another line clarifying that the result may differ from what would actually result if the block was actually placed.

electronicboy (approved):

Owen1212055 (approved):

PR #9822, opened 2023-10-11
Fix CraftPlayerProfile#setId regression

### LINKED ISSUE #9819: PlayerProfile#setId breaks after update to 1.20.2
### Expected behavior

I would expect the PlayerProfile#setId method to set the UUID of the PlayerProfile.

### Observed/Actual behavior

The UUID does not change.

### Steps/models to reproduce

1) Create a plugin and hook into the PlayerPreLoginEvent
2) Attempt to modify the UUID to some arbitrary value
3) Observe that the UUID does not change

Example snippet:
```java
@EventHandler(priority = EventPriority.LOWEST)
public void onPreLogin(AsyncPlayerPreLoginEvent event) {
    var existing = event.getPlayerProfile();
    var uuid = UUID.randomUUID();

    existing.setId(uuid);

    System.out.println(uuid);
    System.out.println(event.getUniqueId());
}
```

Logically, both lines should print the same thing. However, they do not.

### Plugin and Datapack List

None, only my test plugin.

### Paper version

```
[19:30:06 INFO]: Checking version, please wait...
[19:30:06 INFO]: This server is running Paper version git-Paper-226 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: a2ba452)
Previous version: git-Paper-100 (MC: 1.20.1)
```

### Other

*I do realize that the method is marked for removal, however unless it gets removed I expect it to work and not silently fail*

Perhaps this was a typo during the commit: https://github.com/PaperMC/Paper/commit/431e6418ab33ff0decd16f82238e9d50daf328f5

From CraftPlayerProfile.java after the commit:
```java
    @Override
    @Deprecated(forRemoval = true)
    public UUID setId(@Nullable UUID uuid) {
        final GameProfile previousProfile = this.profile;
        final UUID previousId = this.getId();
        this.profile = new GameProfile(previousProfile.getId(), previousProfile.getName());
        copyProfileProperties(previousProfile, this.profile);
        return previousId;
    }
```

From CraftPlayerProfile.java on 1.20.1:
```java
    @Override
    @Deprecated(forRemoval = true)
    public UUID setId(@Nullable UUID uuid) {
        GameProfile prev = this.profile;
        this.profile = new GameProfile(uuid, prev.getName());
        copyProfileProperties(prev, this.profile);
        return prev.getId();
    }
```

If this was intentional, please remove the method from the API.

SplotyCode:
Fixes https://github.com/PaperMC/Paper/issues/9819

### DIFF
--- a/patches/server/0141-Basic-PlayerProfile-API.patch
+++ b/patches/server/0141-Basic-PlayerProfile-API.patch
@@ -17,10 +17,10 @@ public org.bukkit.craftbukkit.profile.CraftPlayerProfile setProperty(Ljava/lang/
 
 diff --git a/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java b/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..0368571e19f37bd259bdb75f918f26d1dd90490a
+index 0000000000000000000000000000000000000000..daa157eaa021d039f9a092bea0b78f7c1f746e3b
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java
-@@ -0,0 +1,402 @@
+@@ -0,0 +1,409 @@
 +package com.destroystokyo.paper.profile;
 +
 +import com.mojang.authlib.yggdrasil.ProfileResult;
@@ -59,7 +59,7 @@ index 0000000000000000000000000000000000000000..0368571e19f37bd259bdb75f918f26d1
 +    }
 +
 +    public CraftPlayerProfile(UUID id, String name) {
-+        this.profile = new GameProfile(id != null ? id : Util.NIL_UUID, name != null ? name : "");
++        this.profile = createAuthLibProfile(id, name);
 +    }
 +
 +    public CraftPlayerProfile(GameProfile profile) {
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

kennytv:
You can't create a game profile with a null uuid

patches/server/0141-Basic-PlayerProfile-API.patch
```diff
@@ -111,7 +111,7 @@ index 0000000000000000000000000000000000000000..b7da8606e24b216b39020130fd2c42c7
 +    public UUID setId(@Nullable UUID uuid) {
 +        final GameProfile previousProfile = this.profile;
 +        final UUID previousId = this.getId();
-+        this.profile = new GameProfile(previousProfile.getId(), previousProfile.getName());
++        this.profile = new GameProfile(uuid, previousProfile.getName());
```
Machine-Maker: If the `uuid` argument is null, you should use `Util.NIL_UUID`. The GameProfile constructor does not accept null for either argument.

Machine-Maker (dismissed):

patches/server/0141-Basic-PlayerProfile-API.patch
SplotyCode: yeah sorry about that

kennytv (approved):

PR #9826, opened 2023-10-12
Support applicable API during config stage

Machine-Maker:
There are at least 2 events I know of that fire during the configuration stage before the `ServerGamePacketListenerImpl` has been created: `PlayerClientOptionsChangeEvent` and `PlayerSpawnLocationEvent`. In addition, there are the listeners for custom payloads that also run. This PR adds support for the functionality that is supported during the configuration phase, namely kicking players, sending custom resource packs, and sending custom payloads.

```[tasklist]
- [x] kick API
- [x] sending resource pack API
- [x] ~~custom channel stuff~~will handle at a later time
```

### DIFF
--- a/patches/server/1052-Support-applicable-API-during-config-stage.patch
+++ b/patches/server/1052-Support-applicable-API-during-config-stage.patch
@@ -1,0 +1,189 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 11 Oct 2023 18:57:45 -0700
+Subject: [PATCH] Support applicable API during config stage
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+index 6a4637eef14cbd84bbe26ef16f004b8f93367a3d..f7923c317672643e16ec605672658b507da34aaa 100644
+--- a/src/main/java/net/minecraft/server/level/ServerPlayer.java
++++ b/src/main/java/net/minecraft/server/level/ServerPlayer.java
+@@ -207,6 +207,7 @@ public class ServerPlayer extends Player {
+     private static final AttributeModifier CREATIVE_BLOCK_INTERACTION_RANGE_MODIFIER = new AttributeModifier(UUID.fromString("736565d2-e1a7-403d-a3f8-1aeb3e302542"), "Creative block interaction range modifier", 0.5D, AttributeModifier.Operation.ADD_VALUE);
+     private static final AttributeModifier CREATIVE_ENTITY_INTERACTION_RANGE_MODIFIER = new AttributeModifier(UUID.fromString("98491ef6-97b1-4584-ae82-71a8cc85cf73"), "Creative entity interaction range modifier", 2.0D, AttributeModifier.Operation.ADD_VALUE);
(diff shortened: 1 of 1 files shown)


MenschenToaster:
I think we should have an API to keep the player in config for a while as well while e.g. the resource pack is loading. Especially because of this issue right here https://github.com/PaperMC/Velocity/issues/1164.

Machine-Maker:
That is outside the scope of this PR. This is just intending to support already-existing API that is valid for use during the configuration phase. Don't interpret this as saying such API shouldn't exist however.

cometcake575:
Anyone have any idea how long until this PR is merged? Isn't it as simple as just clicking the button now?

MiniDigger:
@cometcake575 this PR is still in draft, it even has a todo list in the description with at least one item that needs looking at.
then the code needs to be reviewed by the team, tested and then it can be integrated. integrating is more complicated than just pressing the button because of how our patching works.
tldr: be patient

patches/server/1051-Support-applicable-API-during-config-stage.patch
```diff
@@ -0,0 +1,187 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+index 7e6116963d835d4606ef3d93b69d3e44b61288e1..d5a87da4d64651b32e5b1197e932d7722c606978 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -217,6 +217,27 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+         this.firstPlayed = System.currentTimeMillis();
+     }
+ 
++    // Paper start - send packets during config stage
++    public void sendPacketDuringMultipleStages(final Packet<? extends net.minecraft.network.protocol.common.ClientCommonPacketListener> packet) {
```
Lulu13022002: I don't think this method is needed. Can't you just use commonListener with a null check where needed (i.e the setResourcePack call). Or at least delegate to that method here.

Machine-Maker: done

Twarug:
Is there any progress?

kennytv (approved):

Owen1212055 (changes requested):
In general just want to put this here for more conversation but I feel like we shouldn't "half" support this stage like this. We need to completely deviate from the player inheritance stage in order to properly support this stage... (a disconnect event that occurs here, etc)

Having a player during this stage at all is flawed in itself, and is just due to a bad spigot representation. So we should move away from that in the future anyways.

electronicboy:
I did say elsewhere about making Player more attached to the Network, but, we could maybe go for a more overarching thing of just offering a wrapper for the connection itself which lets you stem off into a configuring client, which lets you see methods only available in that phase, as well as a more general one that lets you get the player? not really sure how that would look, but, we also need to approach this for earlier stage stuff too, anyways, i.e. login plugin messaging API

Machine-Maker:
The Player should be supported in PlayerSpawnLocationEvent which is in a location *after* vanilla creates ServerPlayer but before the game connection is created. So either way, the `Player` interface has to have some support for stuff before the game packet listener is created.

Machine-Maker:
Replaced by configuration phase API.

PR #9828, opened 2023-10-12
Use PluginMessage to allow Velocity to cancel signed chat and commands

Joo200:
This patch adds a custom payload handler in ServerGamePacketListenerImpl to handle cancelled commands and cancelled chat messages from the proxy.

My current implementation wraps the cancelled packet into a plugin message, protocol changes will be handled by the updated  implementation in Minecraft code and Velocity.

This is a draft and can/should be used for https://github.com/PaperMC/Velocity/pull/1100 and possible for Waterfall/BungeeCord.

### DIFF
--- a/patches/server/1039-cancellable-proxy-commands-and-chats.patch
+++ b/patches/server/1039-cancellable-proxy-commands-and-chats.patch
@@ -1,0 +1,45 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Joo200 <github@joo200.de>
+Date: Wed, 4 Oct 2023 21:01:01 +0200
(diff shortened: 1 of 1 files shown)


electronicboy:
for a good while, I've wanted the means to "handshake" between the proxy and the server some form of capability -esque sysem, that way the proxy and the server can see if they support this stuff, i.e. we can detect if we're able to tell the server that chat cycled rather than just having to assume it based on proto version or whatever, but, I did kinda get veto'd on this, so, idk

Joo200:
Yes, that's some open todo for velocity. For Paper it *should* be fine to listen to those plugin messages because there is no active logic that changes behavior.

Maybe we should just add it as config option to the velocity config and send that package then.

PR #9829, opened 2023-10-12
Add BlockData setter for FallingBlock

Sytm:
Due to the fact that the block data of the FallingBlock is only sent to the client with the spawn packet, changing it at a later point would be tricky.
The entity would have to be removed for every player that is tracking it and then resent. If this is desired please let me know and I'll try to (hopefully) make it work

Closes #9792


Owen1212055:
You should be able to call the .update() method after the block data is set, which should correctly resend the entity.

Machine-Maker:
We need to verify that it *looks* correct on vanilla clients if you do that since it technically removes the old entity and adds a new one. It works for other entities like fireballs, paintings and item frames, but it should be verified that it looks correct.

Sytm:
That doesn't work. It isn't entity meta data, it is kind of an entity variant. See [Spawn Entity Packet](https://wiki.vg/Protocol#Spawn_Entity) and [Object Data](https://wiki.vg/Object_Data#Falling_Block)

patches/server/0808-Expand-FallingBlock-API.patch
```diff
@@ -46,10 +51,28 @@ index c64dcceabc41c11542b535d104b7f43172032842..9105418b29c89f092378da11b14e3d32
...
+     public BlockData getBlockData() {
+         return CraftBlockData.fromData(this.getHandle().getBlockState());
+     }
++    // Paper start
++    @Override
++    public void setBlockData(@org.jetbrains.annotations.NotNull final BlockData blockData) {
++        Preconditions.checkNotNull(blockData, "blockData");
++        if (!(blockData instanceof CraftBlockData craftBlockData))  {
++            throw new IllegalArgumentException("Unknown implementation of BlockData: " + blockData);
++        }
```
Machine-Maker: I don't think we need this. We don't confirm that its the correct instance in other places afaik so just a straight cast should be fine.

patches/api/0363-Expand-FallingBlock-API.patch
```diff
@@ -0,0 +1,66 @@
...
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -2257,8 +2257,10 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+      * @return The spawned {@link FallingBlock} instance
+      * @throws IllegalArgumentException if {@link Location} or {@link
+      *     BlockData} are null
++     * @deprecated use {@link #spawn(Location, Class, Consumer)} (or a variation thereof) in combination with {@link FallingBlock#setBlockData(BlockData)}
+      */
+     @NotNull
++    @org.jetbrains.annotations.ApiStatus.Obsolete // Paper
```
Machine-Maker: There are 2 other `spawnFallingBlock` methods and although they are already deprecated, the deprecation notice should be updated to point to the new method. We like to avoid double jumps in deprecation notices, like one deprecated method that points to another that points to the actual one.

Machine-Maker (changes requested):
Feel free to add yourself as a co-author to the patches.

Sytm:
Falling block entities have a field that can contain tile entity data. I guess that should get cleared when changing the BlockData of the entity?

Should I also add an getter / setter for that?

Sytm:
Added getters and setters for the internal CompoundTag that represents the FallingBlocks placed BlockState

patches/api/0363-Expand-FallingBlock-API.patch
```diff
@@ -0,0 +1,108 @@
...
++    @NotNull
++    org.bukkit.block.BlockState getPlacedBlockState();
++
++
++    /**
++     * Sets the {@link BlockData} and possibly present tile entity data for the falling block.
++     *
++     * @param blockState the BlockState to use
++     */
++    void setPlacedBlockState(@NotNull org.bukkit.block.BlockState blockState);
```
lynxplay: Any reason for the "placed" prefix here ?

Sytm: I added that because the block state is not applied until the block is placed on landing.

But most blocks where that really matters aren't rendered properly, so it could be removed anyways

lynxplay: Yea, I mean given the fact that this method immediately changes the "block data"/looks of the falling block, I think it would be best to drop the placed prefix.

lynxplay (approved):
LGTM beyond the renaming issue, renaming can be resolved by the merger or author, whatever happens first.

Machine-Maker (approved):

PR #9830, opened 2023-10-12
Add raw byte block state serialization

TrollyLoki:
Adds two methods for (de)serialization of block states similar to the existing methods for (de)serialization of entities and item stacks.

I decided on serializing block states rather than blocks themselves because it allows plugins to examine the properties of deserialized data via the existing APIs before actually placing them in the world by using the update() method.

### Example Usage:
```java
private byte[] blockBytes;

@EventHandler
public void onBlockBreak(@NotNull BlockBreakEvent event) {
    // serialize block when it's broken
    BlockState state = event.getBlock().getState();
    blockBytes = Bukkit.getUnsafe().serializeBlock(state);
}

@EventHandler
public void onInteract(@NotNull PlayerInteractEvent event) {
    if (blockBytes == null)
        return;
    if (event.getAction() != Action.RIGHT_CLICK_BLOCK)
        return;
    if (event.getItem() == null || event.getItem().getType() != Material.BLAZE_ROD)
        return;

    // deserialize block at another location using a blaze rod
    Block block = event.getClickedBlock().getRelative(event.getBlockFace());
    BlockState state = Bukkit.getUnsafe().deserializeBlock(blockBytes, block);
    state.update(true);
}
```


patches/server/1040-Add-raw-byte-block-state-serialization.patch
```diff
@@ -0,0 +1,77 @@
...
++    public static BlockState getBlockState(Block block, boolean useSnapshot, net.minecraft.world.level.block.state.BlockState customBlockData, BlockEntity customTileEntity) {
+         // Paper end
+         Preconditions.checkNotNull(block, "block is null");
+         CraftBlock craftBlock = (CraftBlock) block;
+         CraftWorld world = (CraftWorld) block.getWorld();
+         BlockPos blockPosition = craftBlock.getPosition();
+-        net.minecraft.world.level.block.state.BlockState blockData = craftBlock.getNMS();
+-        BlockEntity tileEntity = craftBlock.getHandle().getBlockEntity(blockPosition);
++        net.minecraft.world.level.block.state.BlockState blockData = customBlockData != null ? customBlockData : craftBlock.getNMS(); // Paper - use customBlockData if provided
++        BlockEntity tileEntity = customBlockData != null ? customTileEntity : craftBlock.getHandle().getBlockEntity(blockPosition); // Paper - use customTileEntity if customBlockData provided
```
lynxplay: Should we be checking both customBlockData and customTileEntity to not be null here?
Right now it is a pretty implicit contract to pass either both null or both not null.

TrollyLoki: If a deserialized block does not have an associated tile entity, the passed customTileEntity will be null but customBlockData will not be. I am essentially using the non-null status of customBlockData as a flag to indicate that it should use the customTileEntity be it null or non-null instead of getting the tile entity from the world.
Would it be better to have a separate boolean argument specifying which block data and tile entity should be used?

lynxplay: Hmmm, no I think this is fine then. Would probably update the comment above the method talking about this to be a bit more specific then.

I'll go through a deeper review this weekend (tomorrow hopefully), but yea, should be fine then.

PR #9831, opened 2023-10-12
Add LivingEntity#getAllPotionEffects

### LINKED ISSUE #9811: Get Hidden Weaker Potion Effects

Yes. When a stronger potion effect overrides the weaker potion effect, the weaker potion effect remains, but is hidden. It returns after the stronger potion effect expires, if the weaker potion effect has a higher duration than the stronger potion effect.

However, from the API, using `LivingEntity#getActivePotionEffects()` doesn't return the hidden potion effects. There doesn't seem to be any way to get these hidden effects.

Create a new overloading method for `LivingEntity#getActivePotionEffects()` to retrieve hidden potion effects as well. Maybe a boolean for the additional parameter to retrieve said hidden potion effects.

Or, a whole new method is also fine, whichever is better.

I've tried to add a Potion Effects tracker into my plugin, however, that seems like a whole load of extra codes when the feature is already in vanilla Minecraft.

### Other

_No response_

notTamion:
closes #9811


patches/server/1039-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,53 @@
...
++++ b/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
+@@ -327,6 +327,17 @@ public class MobEffectInstance implements Comparable<MobEffectInstance> {
+         return (this.getDuration() <= 32147 || mobEffectInstance.getDuration() <= 32147) && (!this.isAmbient() || !mobEffectInstance.isAmbient()) ? ComparisonChain.start().compareFalseFirst(this.isAmbient(), mobEffectInstance.isAmbient()).compareFalseFirst(this.isInfiniteDuration(), mobEffectInstance.isInfiniteDuration()).compare(this.getDuration(), mobEffectInstance.getDuration()).compare(this.getEffect().getColor(), mobEffectInstance.getEffect().getColor()).result() : ComparisonChain.start().compare(this.isAmbient(), mobEffectInstance.isAmbient()).compare(this.getEffect().getColor(), mobEffectInstance.getEffect().getColor()).result();
+     }
+ 
++    // Paper start
++    @org.jetbrains.annotations.NotNull
++    public java.util.List<MobEffectInstance> getHiddenEffects() {
++        java.util.List<MobEffectInstance> effects = new java.util.ArrayList<>();
++        for(MobEffectInstance effect = this.hiddenEffect; effect != null; effect = effect.hiddenEffect) {
```
Doc94:
```suggestion
+        for (MobEffectInstance effect = this.hiddenEffect; effect != null; effect = effect.hiddenEffect) {
```

patches/server/1039-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,53 @@
...
+         return effects;
+     }
+ 
++    // Paper start
++    @Override
++    public Collection<PotionEffect> getPotionEffects() {
++        List<PotionEffect> effects = new ArrayList<PotionEffect>();
++        for (MobEffectInstance handle : this.getHandle().activeEffects.values()) {
++            effects.add(new PotionEffect(CraftPotionEffectType.minecraftToBukkit(handle.getEffect()), handle.getDuration(), handle.getAmplifier(), handle.isAmbient(), handle.isVisible()));
++            for(MobEffectInstance effect : handle.getHiddenEffects()) {
```
Doc94:
```suggestion
+            for (MobEffectInstance effect : handle.getHiddenEffects()) {
```

patches/api/0442-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,28 @@
...
+ 
++    // Paper start
++    /**
++     * Returns all {@link PotionEffect}s on the living
++     * entity.
++     *
++     * @return a collection of {@link PotionEffect}s
++     */
++    @NotNull
++    public Collection<PotionEffect> getPotionEffects();
```
lynxplay: Idk if I am happy with the naming.
While technically correct, I think we should include the "hidden" part somehow in the name.
Given it is so close to the existing method, imo we want to make sure it is not confused easily.

notTamion: would getAllPotionEffects be good? something like getHiddenPotionEffects is only an option if we make it not return the active ones. do you have any recommendations?

patches/server/1039-Add-LivingEntity-getPotionEffects.patch
notTamion: will fix in next commit

patches/server/1039-Add-LivingEntity-getPotionEffects.patch
notTamion: will fix in next commit

patches/api/0442-Add-LivingEntity-getPotionEffects.patch
riazufila: If you don't mind, I'd like to chime in. I agree with the suggestion `getAllPotionEffects`. To me, it indicates that the method will return all potion effects, hidden included.

notTamion: I changed the method name to getAllPotionEffects since that seemed like the best option to me. i am still going to leave this open incase anyone objects

patches/api/0442-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,28 @@
...
+ 
++    // Paper start
++    /**
++     * Returns all currently active and hidden {@link PotionEffect}s on the living
++     * entity.
++     *
++     * @return a collection of {@link PotionEffect}s
++     */
++    @NotNull
++    public Collection<PotionEffect> getAllPotionEffects();
```
Lulu13022002: Public access is not needed in interface

patches/api/0442-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,28 @@
...
++    // Paper stop
```
Lulu13022002: Should be // Paper end

patches/server/1039-Add-LivingEntity-getPotionEffects.patch
```diff
@@ -0,0 +1,53 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
+@@ -503,6 +503,20 @@ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
+         return effects;
+     }
+ 
++    // Paper start
++    @Override
++    public Collection<PotionEffect> getAllPotionEffects() {
++        List<PotionEffect> effects = new ArrayList<PotionEffect>();
```
Lulu13022002: The generic type <PotionEffect> is not needed in the right assignment, also in this part you skip completely the showIcon part that could be relevant on player with custom made potion.

notTamion: what exactly do you mean by the showIcon part? i assume that variable does exactly what its name suggests and i don't see how exactly that would be relevant for this API method

Lulu13022002: It's a nbt key/field, when used to create a potion using a command or the api, the icon in the top right corner can be hidden in game (not in the inventory). By default it's set to follow the show_particles parameter but it could be different for custom effect. However in your method you ignore that, and so for those type of potions the getter (PotionEffect#hasIcon) will be wrong. It's really niche api but there's no reason to ignore it i think.

notTamion: ah thanks. i used the getActivePotionEffects method as a starting point so i used the wrong overload

notTamion:
Please wait until #9864 is merged since i can then rebase and use CraftPotionUtil

notTamion:
superseded by #9910

PR #6299, opened 2021-07-31
Expose Player Idle Time

tahmid-23:

PR #9833, opened 2023-10-14
Add player idle duration API

booky10:
Implements API for getting, setting and resetting a player's idle duration.

---

Related to #6299, which exposed a getter for the idle time, but got closed because it went stale.

### DIFF
--- a/patches/api/0442-Add-player-idle-duration-API.patch
+++ b/patches/api/0442-Add-player-idle-duration-API.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Add player idle duration API
 Implements API for getting, setting and resetting a player's idle duration.
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 20fa1024f9ad8f478a347be5c554b5e45b398a1c..2fc042b8ffe610a7ec431aad3c1d3ba407f05d40 100644
+index 20fa1024f9ad8f478a347be5c554b5e45b398a1c..2996db88343060624d6aab7889290ae74a973fe8 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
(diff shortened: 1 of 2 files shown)


patches/api/0442-Add-player-idle-duration-API.patch
```diff
@@ -0,0 +1,49 @@
...
++     */
++    @NotNull Duration getIdleDuration();
++
++    /**
++     * After the idle duration exceeds {@link org.bukkit.Bukkit#getIdleTimeout()}, the
++     * player will be kicked for {@link org.bukkit.event.player.PlayerKickEvent.Cause#IDLING}.
++     *
++     * @param duration the new idle duration of this player
++     */
++    void setIdleDuration(@NotNull Duration duration);
```
lynxplay: What is a potential usecase for setting the idle duration here?
Resetting I certainly get, but what would a manual setting be useful for ?

booky10: I've mainly used the get and reset method before, but when creating this PR it seemed odd to have a getter but no setter. I see no reason not to add it.

One possible use case for this would be to persist the idle duration between reconnects, but that seems like a very niche use case to me.

lynxplay: Well I am more concerned about the fact that, at least rn, Duration is a rather wide spectrum of numbers possible to be passed.

Passing some terribly high duration here would quickly end up with the player having an idle duration that makes 0 sense, e.g. longer than they have been online or longer than the server has been up.

booky10: Yeah, makes sense, I'm gonna remove that then

electronicboy (approved):
Header might want a quick change

Machine-Maker (approved):

PR #9834, opened 2023-10-14
Don't call options events on login

froobynooby:
Starting in 1.20.2, when a new `ServerPlayer` instance is created its client options are populated with defaults using the `ServerPlayer.updateOptions` method.

This results in `PlayerClientOptionsChangeEvent` and `PlayerLocaleChangeEvent` being called before the server has actually received that information from the client.

This PR just makes it so the events are not called in the initialisation of `ServerPlayer` (i.e. on player login), consistent with behaviour before 1.20.2.


Machine-Maker (changes requested):
This can be merged into the patch which adds the option event.

kennytv (approved):

Machine-Maker (approved):

PR #9835, opened 2023-10-14
Add Grindstone experience reward API

Sytm:
The only "difference" to Vanilla behavior is when then experience reward will be determined, but otherwise it is left as is.

This supersedes #9762

### DIFF
--- a/patches/api/0442-Add-Grindstone-experience-reward-API.patch
+++ b/patches/api/0442-Add-Grindstone-experience-reward-API.patch
@@ -1,0 +1,34 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lukas Planz <lukas.planz@web.de>
+Date: Sat, 14 Oct 2023 14:28:18 +0200
+Subject: [PATCH] Add Grindstone experience reward API
+
+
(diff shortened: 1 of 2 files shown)


lynxplay (commented):
After a bit of thought, I think moving the time of computation and "ownership" of the experience value from the slot to the inventory is a tad of a dangerous game.

The experience does not exist, we move the time of computation here to a place that may potentially now be out of sync with the actual computation down the line.

Having the inventory "manage" the exp also allows plugins to magically change this value outside of any events being called if they simply keep a reference to it,  leading to potentially more disjoint between expected exp and rewarded one.

This is by far not a hard no from me, I just think the API suggestions needs a lot more evaluation.

Sytm:
I was also thinking about the possibility of moving the accessors into the PrepareGrindstoreEvent.

The code in the CraftEventFactory would look a bit janky due to mutliple instanceof checks and casts but it would work.

Your opinion on that?

Machine-Maker:
Going to close this since https://github.com/PaperMC/Paper/pull/11346 handles this better. The only thing missing is the lack of player context in the BlockExpEvent, but instead, a separate PR to add a new type that extends BlockExpEvent (but doesn't have its own handler list) can be added which adds the player context (like a GrindstoneExpEvent or something).

PR #9838, opened 2023-10-14
Add aggressive mob API

booky10:
Moves the methods from Zombie to Mob, as more Mobs can raise their arm(s).

---

I've marked the existing arms-raised API for Zombies as deprecated. Should this be deprecated, marked as obsolete or just left unchanged?


patches/server/1039-Add-aggressive-mob-API.patch
```diff
@@ -0,0 +1,50 @@
...
+ }
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPanda.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPanda.java
+index 56f9630dbe5d18d5ec33dc85f6531723022d6a3b..646d915552733603f9884012646d46ed82df5758 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPanda.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPanda.java
+@@ -127,6 +127,14 @@ public class CraftPanda extends CraftAnimals implements Panda {
+         return this.getHandle().getUnhappyCounter();
+     }
+ 
++    // Paper start
```
Lulu13022002: This override shouldn't be needed since internal already override this one?

Lulu13022002 (commented):
The setter already exists in RangedEntity#setChargingAttack so i wonder if the setter is still needed?

patches/server/1039-Add-aggressive-mob-API.patch
booky10: What exactly do you mean?
I've added an override for pandas because vanilla always returns true for the internal `isAggressive()` if its variant is `AGGRESSIVE`. If I would remove this override, calling `setAggressive(false)` on a panda would have no effect on the return value of `isAggressive()`.

booky10:
The `setChargingAttack` and `isChargingAttack` stuff seems weird to me. `isChargingAttack` just checks if the entity is currently using an item (by redirecting to `LivingEntity#isHandRaised`), but `setChargingAttack` makes the entity aggressive which doesn't make the entity start/stop using an item immediately.

I'll deprecate them later today.

patches/server/1039-Add-aggressive-mob-API.patch
Lulu13022002: Oh yes sorry i got confused.

patches/api/0162-Add-more-Zombie-API.patch
```diff
@@ -40,14 +40,18 @@ index 336b3efaf0a8ed5a238e8b941193d690e8b72d71..227560e04854088d162b4b6ed4cd1503
 +     * Set if zombie has its arms raised
 +     *
 +     * @param raised True to raise arms
++     * @deprecated use {{@link #setAggressive(boolean)}}
 +     */
++    @Deprecated // Paper
```
Lulu13022002: Paper comments are not needed here you're already in a Paper comment block, same below

patches/api/0442-Add-aggressive-mob-API.patch
```diff
@@ -0,0 +1,42 @@
...
++     *     <li>{@link Pillager}</li>
++     *     <li>{@link PiglinBrute}</li>
++     * </ul>
++     *
++     * @return wether the mob is aggressive or not
++     */
++    boolean isAggressive();
++
++    /**
++     * @see #isAggressive()
```
Lulu13022002: Feel free to add a bit of documentation here redirection only is annoying in the ide. Can say what will happen for some mob then redirect to the complete list of mobs

Owen1212055:
Please squash this into the add missing entity API patch

booky10:
> Please squash this into the add missing entity API patch

Rebased on master and merged patches

patches/server/0626-Missing-Entity-API.patch
```diff
@@ -882,6 +904,21 @@ index 5467e4a74b70ff57b49d9e6bc686c493178f8511..56f9630dbe5d18d5ec33dc85f6531723
...
      public boolean isRolling() {
+@@ -101,6 +127,14 @@ public class CraftPanda extends CraftAnimals implements Panda {
+         return this.getHandle().getUnhappyCounter();
+     }
+ 
++    // Paper start
++    @Override
++    public boolean isAggressive() {
++        // Pandas are always aggressive if their variant is AGGRESSIVE
++        return (this.getHandle().getEntityData().get(net.minecraft.world.entity.Mob.DATA_MOB_FLAGS_ID) & 4) != 0;
```
Owen1212055: Why are you overriding this? The nms call overrides the method anyways.

Owen1212055: Actually, i have read the above comment. I am not sure why tho we still need this. Since this is a deviation of vanilla behavior, as the panda that has this gene will ALWAYS be considered aggressive. I am not sure if we should make this distinction here.

booky10: It didn't make sense to me as setting aggressive to false and then checking it would still return true. Still, like you said, it also doesn't align with vanilla.
Either way, I'll add a note to the javadoc for this.

Owen1212055: This behavior is mostly expected id see. Cause even so, it’s just manually editing the flag, and so as long as docs properly say that this may sometimes have undefined behavior in relation to get/set aggressive, I think that’s fine.

patches/server/0626-Missing-Entity-API.patch
```diff
@@ -28,6 +28,7 @@ public net.minecraft.world.entity.animal.AbstractSchoolingFish leader
 public net.minecraft.world.entity.animal.AbstractSchoolingFish schoolSize
 public net.minecraft.world.entity.animal.Rabbit moreCarrotTicks
 public net.minecraft.world.entity.AreaEffectCloud ownerUUID
+public net.minecraft.world.entity.Mob DATA_MOB_FLAGS_ID
```
Lulu13022002: This AT is not needed anymore

Owen1212055 (approved):

lynxplay (approved):

electronicboy (approved):

PR #9856, opened 2023-10-20
Add Enchantment cost API

luisc99:
Cost is a property of individual enchantments, and is used by vanilla in combination with environmental aspects like tool enchantability and bookshelf count to determine the final cost of an enchantment as shown in an enchanting table.

Having access to the base cost of an enchantment using these vanilla methods can allow plugin developers to determine the "value" of an enchantment, and use it in custom logic where needed. I came across this recently when trying to assign an economic value to enchantments during tool repairing, and noticed these values don't seem to be obtainable under the current API.


Machine-Maker (changes requested):
I think the naming and javadocs of these methods need to make it more clear that its not the `number of experience levels needed`. As you briefly mention, its not the direct cost. I don't think calling it a "cost" is really correct. And the `@return` tag in the javadoc doesn't do anything to dispel that notion.

Maybe something like `getMin/MaxModifiedCost`? Since it is used when comparing it to the modified level amount on the enchant clicked.

luisc99:
I agree this concept of "cost" is a bit confusing. The original method names were mainly to match the Mojang ones, as the existing similar surrounding methods all did, but I've changed it them to be clearer. I've also updated the javadocs a bit, to hopefully try make it more obvious the cost isn't directly related to experience levels and to add a bit more context.

Machine-Maker (changes requested):
These patches can be nicely squashed into the `More Enchantment API` patch that already exist for the server and the API. You can then add yourself as a co author of the patch with
`Co-authored-by: Luis <luisc99@icloud.com>`

Machine-Maker (approved):

NoahvdAa (approved):

Owen1212055 (approved):

PR #9858, opened 2023-10-20
Cleanup disable explosion knockback patch

LemonCaramel:
EntityEvent 2 was previously used to show damage effects. This is no longer used in modern clients.
+Fixes the issue of damage sounds not playing when the option is enabled.

### DIFF
--- a/patches/server/0041-Disable-explosion-knockback.patch
+++ b/patches/server/0041-Disable-explosion-knockback.patch
@@ -4,31 +4,6 @@ Date: Wed, 2 Mar 2016 14:48:03 -0600
 Subject: [PATCH] Disable explosion knockback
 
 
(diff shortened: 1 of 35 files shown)


Machine-Maker:
I'm curious what this bit of diff was supposed to do originally? On the surface, it doesn't look like it has anything to do with knockback. 

I did confirm that this change does change behavior back to what it was during 1.18.2.

LemonCaramel:
This diff was created in Paper 1.9. It's unrelated to knockback.

However, in Paper 1.8, it was actually used to [disable knockback](https://github.com/PaperMC/Paper/blob/ver/1.8.8/Spigot-Server-Patches/0055-Disable-explosion-knockback.patch#L18-L22).


spigot 1.8

Machine-Maker (approved):

lynxplay (approved):

PR #9860, opened 2023-10-22
Fix log level of advancement debug message

jpenilla:
This message when advancements are removed was added in 1.20.2. It doesn't provide anything useful and spams the console when removing advancements.


lynxplay:
Would probably fit nicely into 0901-Improve-logging-and-errors.patch. That patch already has removal of unneeded logs.

electronicboy (approved):

Machine-Maker (dismissed):
What lynx said, this can go into the logging fixes patch.

lynxplay (approved):

PR #9861, opened 2023-10-22
Updated Upstream (Bukkit/CraftBukkit/Spigot)

Machine-Maker:
Upstream has released updates that appear to apply and compile correctly. This update has not been tested by PaperMC and as with ANY update, please do your own testing

Bukkit Changes:
5010ed00 PR-923: Back Particle by a minecraft registry
d47dd212 Increase clarity of errors when loading malformed plugin main classes
2c3c3337 PR-914: Make use of ApiStatus.Internal reather than deprecation
2b4c3b46 PR-922: Add keepSpawnInMemory setting to WorldCreator

CraftBukkit Changes:
75502b6dd SPIGOT-7507: Fall back to world UUIDs if Dimension tag missing or invalid
67a52a648 PR-1279: Back Particle by a minecraft registry
f4d977e79 Simplify texture updating in CraftMetaSkull
e42510b06 Increase outdated build delay
f3ad63aad SPIGOT-7506: SculkBloomEvent.getBlock() world is null
b59004114 PR-1278: Return correct default light in chunk snapshot
2a381856b SPIGOT-7503: Remove special handling for minecraft:brand custom channel
2fa7644d0 PR-1277: Add keepSpawnInMemory setting to WorldCreator

Spigot Changes:
dba3cdc3 Rebuild patches

PR #9863, opened 2023-10-23
Rewrite custom payload handler

LemonCaramel:
Rewrite the CustomPayload handler using Mojang mechanism
https://github.com/PaperMC/Paper/pull/9861#issuecomment-1774357124

```java
public final class TestPlugin extends JavaPlugin implements Listener, PluginMessageListener {

    @Override
    public void onEnable() {
        this.getServer().getMessenger().registerIncomingPluginChannel(this, "minecraft:brand", this);
    }

    @Override
    public void onPluginMessageReceived(@NotNull String channel, @NotNull Player player, byte[] msg) {
        // Brand Test
        final String spigotMethod = new String(msg, StandardCharsets.UTF_8).substring(1);
        this.getSLF4JLogger().info("Spigot Method: [{}]", spigotMethod);
        this.getSLF4JLogger().info("Paper API: [{}]", player.getClientBrandName());
    }

    @EventHandler
    void onPlayerRegisterChannel(final PlayerRegisterChannelEvent event) {
        this.getSLF4JLogger().info("Register Channel: [{}]", event.getChannel());
    }

    @EventHandler
    void onPlayerUnregisterChannel(final PlayerUnregisterChannelEvent event) {
        this.getSLF4JLogger().info("Unregister Channel: [{}]", event.getChannel());
    }
}
```


electronicboy:
I don't think that we need to move the handling stuff over is needed, just moreso the fact that it retains the OG buffer and then we can pass that through, but, I generally have nothing against this, it just doesn't retain the buffer as was desired, as then that can be passed to stuff

post that, we can generally look into the state of the plugin messaging API

patches/server/0445-Rewrite-custom-payload-handler.patch
```diff
@@ -0,0 +1,180 @@
...
+@@ -47,7 +53,15 @@ public record ServerboundCustomPayloadPacket(CustomPacketPayload payload) implem
+     }
+ 
+     // CraftBukkit start
+-    public record UnknownPayload(ResourceLocation id, io.netty.buffer.ByteBuf data) implements CustomPacketPayload {
++    // Paper start - Rewrite custom payload handler
++    public record UnknownPayload(ResourceLocation id, io.netty.buffer.ByteBuf data) implements io.papermc.paper.network.payload.ReadableMessage {
++        @Override
++        public byte[] message() {
++            final byte[] message = new byte[data.readableBytes()];
```
electronicboy: Subsequent calls to this will blow up as readBytes modifies the reader index

patches/server/0445-Rewrite-custom-payload-handler.patch
```diff
@@ -0,0 +1,180 @@
...
+diff --git a/src/main/java/net/minecraft/network/protocol/common/ServerboundCustomPayloadPacket.java b/src/main/java/net/minecraft/network/protocol/common/ServerboundCustomPayloadPacket.java
+index af86f752c33a2990405fea058b7c41c437ba9d46..8f5ab7d6cceb3558960eeea1b90a75d227ba367a 100644
+--- a/src/main/java/net/minecraft/network/protocol/common/ServerboundCustomPayloadPacket.java
++++ b/src/main/java/net/minecraft/network/protocol/common/ServerboundCustomPayloadPacket.java
+@@ -12,7 +12,13 @@ import net.minecraft.resources.ResourceLocation;
+ public record ServerboundCustomPayloadPacket(CustomPacketPayload payload) implements Packet<ServerCommonPacketListener> {
+ 
+     private static final int MAX_PAYLOAD_SIZE = 32767;
+-    private static final Map<ResourceLocation, FriendlyByteBuf.Reader<? extends CustomPacketPayload>> KNOWN_TYPES = ImmutableMap.<ResourceLocation, FriendlyByteBuf.Reader<? extends CustomPacketPayload>>builder().build(); // CraftBukkit - no special handling
++    // Paper start - Rewrite custom payload handler
```
electronicboy: I generally want some comments on what best to do here, 32kb of data is pretty small by default and so I did maybe envision that we'd add way to allow plugins to register their own channels AND override this limit, having this this thing before, it's a PITA as the network I was doing stuff for really didn't wanna pick up additional infra, for ex

PR #9865, opened 2023-10-23
add getAdvancementProgress() to PlayerAdvancementCriterionGrantEvent

456dev:
this allows for simpler checks for if an advancement would be given to a player, 
and denying it (as the advancement grant event is non-cancelable)

# feedback wanted:

- [x] is it ok to rename the patch from just `PlayerAdvancementCriterionGrantEvent`?
- [ ] is it worth adding this to api at all?

# alternatives:
- `event.getPlayer().getAdvancementProgress(event.getAdvancement()).isDone();`
- providing just a subset of advancement progress (isDone, like it was originally)

### DIFF
--- a/patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
+++ b/patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
@@ -7,10 +7,10 @@ Co-authored-by: The456gamer <the456gamer@the456gamer.dev>
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..b8cdeb48252550a8cc36b0673aeab2a9863bccfb
+index 0000000000000000000000000000000000000000..6e0395f4dc48bd95e29250408aa38176b9c97c70
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java
-@@ -0,0 +1,76 @@
+@@ -0,0 +1,80 @@
 +package com.destroystokyo.paper.event.player;
 +
 +import org.bukkit.advancement.Advancement;
@@ -23,6 +23,8 @@ index 0000000000000000000000000000000000000000..b8cdeb48252550a8cc36b0673aeab2a9
 +
 +/**
 + * Called when a player is granted a criteria in an advancement.
++ *
++ * Unlike most other events, this happens after the criterion is granted, and is rolled back if canceled
 + */
 +public class PlayerAdvancementCriterionGrantEvent extends PlayerEvent implements Cancellable {
 +    private static final HandlerList handlers = new HandlerList();
@@ -64,6 +66,8 @@ index 0000000000000000000000000000000000000000..b8cdeb48252550a8cc36b0673aeab2a9
 +
 +    /**
 +     * Gets the AdvancementProgress.
++     * Unlike other methods, it returns the state after the event has completed.
++     * If the event is canceled, the change is rolled back after the event finishes.
 +     *
 +     * @return advancement progress
 +     */
--- a/patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
+++ b/patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
@@ -7,10 +7,10 @@ Co-authored-by: The456gamer <the456gamer@the456gamer.dev>
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..6e0395f4dc48bd95e29250408aa38176b9c97c70
+index 0000000000000000000000000000000000000000..9e43739404814781ede832ec5931cd2d963fff19
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerAdvancementCriterionGrantEvent.java
-@@ -0,0 +1,80 @@
+@@ -0,0 +1,78 @@
 +package com.destroystokyo.paper.event.player;
 +
 +import org.bukkit.advancement.Advancement;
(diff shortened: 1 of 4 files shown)


notTamion (commented):
not sure if this is really needed since you can already get the bukkit AdvancementProgress using CraftPlayer#getAdvancementProgess(Advancement). Though i think adding a getAdvancementProgress method to the event could still be a nice quality of life feature. But only returning a boolean seems not very useful to me when you could also return way more information

patches/api/0085-add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -1,15 +1,15 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: BillyGalbreath <Blake.Galbreath@GMail.com>
 Date: Fri, 19 Jan 2018 08:15:14 -0600
-Subject: [PATCH] PlayerAdvancementCriterionGrantEvent
+Subject: [PATCH] add PlayerAdvancementCriterionGrantEvent
```
notTamion:
```suggestion
Subject: [PATCH] Add PlayerAdvancementCriterionGrantEvent
```
Renaming this isn't really required but if you would like to go through with it then i think you should make it start with an uppercase letter since 99 % of patches start with one

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -58,6 +62,16 @@ index 0000000000000000000000000000000000000000..bb8d7c959cdea4b66455a49e74804ea4
 +        return cancel;
 +    }
 +
++    /**
++     * Gets the AdvancementProgress.
```
Machine-Maker: I would clarify that this is the advancement progress **with** the criteron granted. This event kinda breaks the norm by having the change happen, then fire the event and undo the change if it was cancelled. Maybe that should be changed, but that can be for another time.

Machine-Maker (changes requested):

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -21,17 +23,21 @@ index 0000000000000000000000000000000000000000..bb8d7c959cdea4b66455a49e74804ea4
 +
 +/**
 + * Called when a player is granted a criteria in an advancement.
++ *
++ * Unlike most other events, this happens after the criterion is granted, and is rolled back if canceled
```
notTamion: I think better wording would be something like: Called after a player is granted a criterion in an advancement. If cancelled the criteria will be revoked.

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -58,6 +64,18 @@ index 0000000000000000000000000000000000000000..bb8d7c959cdea4b66455a49e74804ea4
 +        return cancel;
 +    }
 +
++    /**
++     * Gets the AdvancementProgress.
++     * Unlike other methods, it returns the state after the event has completed.
++     * If the event is canceled, the change is rolled back after the event finishes.
```
notTamion: I think a better way would be something like this: Gets the AdvancementProgress with the involved criterion marked as granted. Would love to hear feedback on these suggestions though

notTamion (changes requested):

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
Machine-Maker: I agree that this is better wording.

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
456dev: I think I might be confused that it implies only this advancement progress has it granted: other sources would still have it not marked, which is incorrect

maybe
gets the current AdvancementProgress after the criterion has been granted.

the456gamer:
oh oops

notTamion:
woops lol. not exaclty sure what happened here but seems to me that you didn't rebase correctly: https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#rebasing-prs

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
notTamion: i am unsure about this one. "after the criterion has been granted" kinda makes it seem like the criterion will get granted regardless of if the event is cancelled etc. perhaps we could add something like "See {@link PlayerAdvancementCriterionGrantEvent}" to my suggestion. since we already clarified it nicely in the class jdocs

456dev: i think including the word current helps a lot, but i agree that linking back to original event is useful. is that compromise ok?

notTamion: yeah that seems nice

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -1,37 +1,42 @@
...
 +import org.bukkit.entity.Player;
 +import org.bukkit.event.Cancellable;
 +import org.bukkit.event.HandlerList;
 +import org.bukkit.event.player.PlayerEvent;
 +import org.jetbrains.annotations.NotNull;
 +
 +/**
-+ * Called when a player is granted a criteria in an advancement.
++ * Called after a player is granted a criteria in an advancement.
++ *I f cancelled the criteria will be revoked.
```
notTamion:
```suggestion
+ * If cancelled the criteria will be revoked.
```

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
```diff
@@ -58,6 +63,17 @@ index 0000000000000000000000000000000000000000..bb8d7c959cdea4b66455a49e74804ea4
 +        return cancel;
 +    }
 +
++    /**
++     * Gets the current AdvancementProgress.
```
notTamion, line 67: thing is if we do say it returns the current one. it actually has to return the current one which it doesn't since the advancement progress is being retrieved during event creation and not during the method call. i am unsure if the jdocs or the way its stored is the issue. somebody from paper team should probably look over that one and say how its usually done with events

notTamion (changes requested):

patches/api/0085-Add-PlayerAdvancementCriterionGrantEvent.patch
notTamion: might be fine though. probably just me being dyslexic

456dev: From what I can tell, anything that would invalidate the advancement progress would also invalidate the advancement itself that is provided in the same way, so I thought it was unnecessary

notTamion: yeah good point. LGTM then

Machine-Maker (approved):

Owen1212055 (approved):

PR #9869, opened 2023-10-25
Fix null item in sendEquipmentChange

SplotyCode:


SplotyCode:
Should i create a new paper patch instead of a fixup?

Leguan16:
This looks like its a new patch. 
For now, don't use imports, but FQN instead.

SplotyCode:
> This looks like its a new patch. For now, don't use imports, but FQN instead.

Ah sorry now that you mention it males total since as it would be annoying when resolving upstream conflicts

Machine-Maker (approved):

patches/server/1040-Fix-null-item-for-Player-sendEquipmentChange.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+index 604ed1f6742a6b028b9db2809f7bd4b9a3b38f4d..96ab6b5787839bfb6be682bdbc29a011eab8384c 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -1072,7 +1072,7 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+ 
+     @Override
+     public void sendEquipmentChange(LivingEntity entity, EquipmentSlot slot, ItemStack item) {
+-        this.sendEquipmentChange(entity, Map.of(slot, item));
++        this.sendEquipmentChange(entity, java.util.Collections.singletonMap(slot, item)); // Paper - replace Map.of
```
MiniDigger:
```suggestion
+        this.sendEquipmentChange(entity, java.util.Collections.singletonMap(slot, item)); // Paper - replace Map.of to allow null values
```
maybe a bit more explicit?

MiniDigger: maybe change the patch title to into allow null items for ...

lynxplay (approved):

PR #9880, opened 2023-10-28
Fix enchantable brush

LaylaSilbernberg:
I think this should do the trick. I also want to add some tests but I need some more familiarity with the structure of paper so might need some help there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+A fork of paper focused on bugfixing.
+
+===========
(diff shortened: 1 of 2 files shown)


Doc94:
For starting the PR is good handle the main issue, also i recommed you try to edit the original patch where this Tags was added. you can check here https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#modifying-patches for know how to edit a patch.

electronicboy:
My original comment generally showcased that there was a bunch of stuff missing from the enchantables set which should also be added

LaylaSilbernberg:
Oh I feel very silly, I felt like I was missing something obvious. I will have a look at these things tomorrow.

PR #9883, opened 2023-10-29
Fix entity camera not being reset when cancelling spectating start/stop events

booky10:
Issue can be reproduced by cancelling all `PlayerStartSpectatingEntity` events and sneaking after trying to start spectating an entity.

(btw 415d7081d2cdddba83b91a68cf2397ce95ac115b broke patching, I think the submodule-update wasn't intentional)


electronicboy:
I apparently had some random branch checked out before I pushed that commit, should be fixed now, derp

electronicboy (approved):

PR #9885, opened 2023-10-29
Fix NPE when no valid world is found on legacy Players

Machine-Maker:
Upstream's commit [here](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/75502b6ddcae5aa84b7e97fd02d2e1551afddebc#nms-patches/net/minecraft/server/players/PlayerList.patch) failed to account for the level not being set on the ServerPlayer if it doesn't point to a valid world. This adds a null check on the ServerLevel before getting the level key.

---
Download the paperclip jar for this pull request: [paper-9885.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1014594901.zip)


electronicboy (approved):

electronicboy:
So, part of the thing here is that spigot sets the default world to the overworld earlier if a world isn't loaded/found, which is likely what they're relying on here

Given that the OG issue that fixes is probably still a thing, this is probably the best fix as I don't think that we've had reports of the issue spigot is trying to fix?

PR #9887, opened 2023-10-29
Fix a bunch of stuff with player spawn locations

Machine-Maker:
If a playerdata doesn't contain a valid, loaded world, reset to the main world spawn point.

This also squashes 2 very related patches

Supersedes https://github.com/PaperMC/Paper/pull/9116 but doesn't close the linked issue to that PR because I do think adding an event to control the target world is useful, but that will be in a PR after this one.


lynxplay (approved):

electronicboy (approved):
If this blows up, I blame Lynx

PR #9888, opened 2023-10-29
Added missing enchantables to material tags

### LINKED ISSUE #9862: MaterialTags.ENCHANTABLE.isTagged(Material.BRUSH) return false
### Expected behavior

The brush is enchantable, so `MaterialTags.ENCHANTABLE.isTagged(Material.BRUSH)` should return true

### Observed/Actual behavior

`MaterialTags.ENCHANTABLE.isTagged(Material.BRUSH)` return false

### Steps/models to reproduce

.

### Plugin and Datapack List

.

### Paper version

I am not using Paper, but I confirm that this occurred on the latest Paper

### Other

_No response_

LaylaSilbernberg:
Sorry for opening a new pull request instead of using the previous one. I added the material tags I overlooked and instead of creating a new patch, edited a previous one. What do you think? I would still like to add some tests but might have to think about the implementation.


patches/api/0153-Add-Material-Tags.patch
```diff
@@ -8,6 +8,7 @@ are related to each other by a trait.
 
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 Co-authored by: Lena Kolb <lenakolb2204@gmail.com>
```
Leguan16: you missed the `-` between the `authored` and `by` :) maybe you can add it to the line above too? But another person should confirm this.

LaylaSilbernberg: doh! but should not be too big a deal.

LaylaSilbernberg: I'll wait for a second confirmation first.

patches/api/0153-Add-Material-Tags.patch
```diff
@@ -682,7 +683,11 @@ index 0000000000000000000000000000000000000000..0b92f3279fe5dcbea77b22803447b7ad
 +     */
 +    public static final MaterialSetTag ENCHANTABLE = new MaterialSetTag(keyFor("enchantable"))
 +        .add(PICKAXES, SWORDS, SHOVELS, AXES, HOES, HELMETS, CHEST_EQUIPPABLE, LEGGINGS, BOOTS, BOWS)
-+        .add(Material.TRIDENT, Material.SHIELD, Material.FISHING_ROD, Material.SHEARS, Material.FLINT_AND_STEEL, Material.CARROT_ON_A_STICK, Material.WARPED_FUNGUS_ON_A_STICK)
++        .add(Material.TRIDENT, Material.SHIELD, Material.FISHING_ROD, Material.SHEARS,
++            Material.FLINT_AND_STEEL, Material.CARROT_ON_A_STICK, Material.WARPED_FUNGUS_ON_A_STICK,
++            Material.BRUSH, Material.CARVED_PUMPKIN, Material.COMPASS, Material.SKELETON_SKULL,
++            Material.WITHER_SKELETON_SKULL, Material.PLAYER_HEAD, Material.ZOMBIE_HEAD,
++            Material.CREEPER_HEAD, Material.DRAGON_HEAD, Material.PIGLIN_HEAD)
 +        .ensureSize("ENCHANTABLE", 65).lock();
```
Lulu13022002: The current test will fail (and it will throw) if the size is not increased by the proper number (75 here it seems)

LaylaSilbernberg: Gotcha! I will make the appropriate changes this evening then. Thanks for the feedback.

LaylaSilbernberg:
Added the changes.

notTamion:
#9862 just so people looking at the issue know there is a pr for it

Machine-Maker (approved):
Owen is going to add the test cat wrote in the issue to this, and then its good.

PR #9890, opened 2023-10-30
Add Llama caravan events

SplotyCode:


patches/api/0443-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,172 @@
...
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.entity.EntityEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called just before a {@link Llama} joins a caravan.
++ */
++public class LlamaJoinCaravanEvent extends EntityEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker:
```suggestion
+    private static final HandlerList HANDLER_LIST = new HandlerList();
```

patches/api/0443-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,172 @@
...
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.entity.EntityEvent;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Called just before a {@link Llama} leaves a caravan.
++ */
++public class LlamaLeaveCaravanEvent extends EntityEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker:
```suggestion
+    private static final HandlerList HANDLER_LIST = new HandlerList();
```

patches/server/1044-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,98 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java b/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java
+index 5f61c97478f005aaaaad1b027118079db7275cf7..83ffa6b121460d769245c82337d06332e89a7289 100644
+--- a/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java
++++ b/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java
+@@ -437,8 +437,19 @@ public class Llama extends AbstractChestedHorse implements VariantHolder<Llama.V
+         }
+     }
+ 
++    // Paper start - Add Llama caravan events
+     public void leaveCaravan() {
```
Machine-Maker, line 43: I think method should marked with `@Deprecated` and `@DoNotUse` to catch any future uses of it since they should all use the method with a reason.

patches/server/1044-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,98 @@
...
++                return;
++            }
++            // Paper end - Add Llama caravan events
+             this.caravanHead.caravanTail = null;
+         }
+ 
+@@ -446,10 +457,26 @@ public class Llama extends AbstractChestedHorse implements VariantHolder<Llama.V
+     }
+ 
+     public void joinCaravan(Llama llama) {
```
Machine-Maker, line 67: Same situation here with annotated this method. Also, missing paper comments for diff directly below

patches/server/1044-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,98 @@
...
+     @Override
+     public void joinCaravan(@org.jetbrains.annotations.NotNull Llama llama) {
+-        this.getHandle().joinCaravan(((CraftLlama) llama).getHandle());
++        this.getHandle().joinCaravan(((CraftLlama) llama).getHandle(), io.papermc.paper.event.entity.LlamaJoinCaravanEvent.Reason.PLUGIN);
+     }
+ 
+     @Override
+     public void leaveCaravan() {
+-        this.getHandle().leaveCaravan();
++        this.getHandle().leaveCaravan(io.papermc.paper.event.entity.LlamaLeaveCaravanEvent.Reason.PLUGIN);
```
Machine-Maker: Despite this being inside a paper block, best to add an inline paper comment with a comment of what is changing.

patches/server/1044-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,98 @@
...
+         this.caravanHead.caravanTail = this;
+     }
+ 
++    // Paper start
++    @Override
++    public org.bukkit.craftbukkit.entity.CraftLlama getBukkitEntity() {
++        return (org.bukkit.craftbukkit.entity.CraftLlama) super.getBukkitEntity();
++    }
++    // Paper end
++
```
Machine-Maker, line 88: Not sure this is worth not just having 2 casts when the llama is passed to the event constructor. I don't think we do this for other similar situations.

patches/api/0443-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,172 @@
...
++ */
++public class LlamaJoinCaravanEvent extends EntityEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
++    @NotNull
++    private final Llama appendTo;
++    @NotNull
++    private final Reason reason;
++    private boolean cancelled;
++
++    public LlamaJoinCaravanEvent(final @NotNull Llama what, final @NotNull Llama appendTo, final @NotNull Reason reason) {
```
Machine-Maker, line 34: Annotated the constructor with `@ApiStatus.Internal`.

patches/api/0443-Add-Llama-caravan-events.patch
```diff
@@ -0,0 +1,172 @@
...
++/**
++ * Called just before a {@link Llama} leaves a caravan.
++ */
++public class LlamaLeaveCaravanEvent extends EntityEvent implements Cancellable {
++    private static final HandlerList handlers = new HandlerList();
++    @NotNull
++    private final Reason reason;
++    private boolean cancelled;
++
++    public LlamaLeaveCaravanEvent(final @NotNull Llama what, final @NotNull Reason reason) {
```
Machine-Maker, line 124: Annotated the constructor with `@ApiStatus.Internal`.

Machine-Maker (changes requested):

patches/server/1044-Add-Llama-caravan-events.patch
SplotyCode: I dont really have a strong opinion on this. In our CraftLlama we could then also remove some casts and for plugin devs it is always tedious to use  `((X) ((CraftX) x).getHandle()).methodCall()`
image
We can also use casts but even with var it looks quite ugly using the fqn. Do you have fears of merge conflicts with upstream?

PR #9893, opened 2023-10-31
Fix vanilla components not being translated

booky10:
Vanilla components were just converted to normal json strings without being translated by the `GlobalTranslator`. This caused custom name's of entities to no longer be translated properly after a server restart or an entity unload.

This does have a small performance cost, as all vanilla components (not backed by an adventure component) are now converted to an adventure component by serializing to json and deserializing from json again.

<details>
<summary>Test plugin diff</summary>

```diff
Index: test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java b/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java
--- a/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java	(revision c71e79c2d56d2cd374995cbd5a5ac7c3d1fcf5b2)
+++ b/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java	(date 1698715701243)
@@ -1,5 +1,16 @@
 package io.papermc.testplugin;
 
+import io.papermc.paper.event.player.AsyncChatEvent;
+import java.text.MessageFormat;
+import java.util.Locale;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.translation.GlobalTranslator;
+import net.kyori.adventure.translation.TranslationRegistry;
+import org.bukkit.Bukkit;
+import org.bukkit.Location;
+import org.bukkit.NamespacedKey;
+import org.bukkit.entity.ArmorStand;
+import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.plugin.java.JavaPlugin;
 
@@ -8,5 +19,20 @@
     @Override
     public void onEnable() {
         this.getServer().getPluginManager().registerEvents(this, this);
+
+        TranslationRegistry registry = TranslationRegistry.create(new NamespacedKey(this, "i18n"));
+        registry.register("testplugin.armorstand", Locale.ENGLISH, new MessageFormat("Success!"));
+        GlobalTranslator.translator().addSource(registry);
+    }
+
+    @EventHandler
+    public void onChat(AsyncChatEvent event) {
+        Bukkit.getScheduler().runTask(this, () -> {
+            Location spawnLoc = event.getPlayer().getLocation();
+            spawnLoc.getWorld().spawn(spawnLoc, ArmorStand.class, armorStand -> {
+                armorStand.setCustomNameVisible(true);
+                armorStand.customName(Component.translatable("testplugin.armorstand"));
+            });
+        });
     }
 }

```
</details>

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -1221,7 +1221,7 @@ index 0000000000000000000000000000000000000000..2fd6c3e65354071af71c7d8ebb97b559
 +}
 diff --git a/src/main/java/io/papermc/paper/adventure/PaperAdventure.java b/src/main/java/io/papermc/paper/adventure/PaperAdventure.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..1fb9b0ea9f2dbff12d8d2fbc2ad2f4d88ca9f7ad
+index 0000000000000000000000000000000000000000..f5ba56f9c8406aef2fe91d6b4174762cfed11fce
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/adventure/PaperAdventure.java
 @@ -0,0 +1,394 @@
@@ -1430,7 +1430,7 @@ index 0000000000000000000000000000000000000000..1fb9b0ea9f2dbff12d8d2fbc2ad2f4d8
 +        return PlainTextComponentSerializer.plainText().serialize(translated(component, locale));
 +    }
 +
-+    private static Component translated(final Component component, final Locale locale) {
++    public static Component translated(final Component component, final Locale locale) {
 +        return GlobalTranslator.render(
 +            component,
 +            // play it safe
@@ -2120,7 +2120,7 @@ index d120fff432d9c4fc7a35ddffdc4186459e45e950..73c15a0c56a103ba4e62f0a51af8d425
          }
      }
 diff --git a/src/main/java/net/minecraft/network/FriendlyByteBuf.java b/src/main/java/net/minecraft/network/FriendlyByteBuf.java
-index 9373502ede6c8a881af67db005cf12fd9313f37f..50c7e0a449ea9d71e21e86b4f03dac0298f96130 100644
+index 9373502ede6c8a881af67db005cf12fd9313f37f..d9be3fada2e603684275a2094954e29039fb07c7 100644
 --- a/src/main/java/net/minecraft/network/FriendlyByteBuf.java
 +++ b/src/main/java/net/minecraft/network/FriendlyByteBuf.java
 @@ -87,6 +87,7 @@ public class FriendlyByteBuf extends ByteBuf {
@@ -2131,7 +2131,7 @@ index 9373502ede6c8a881af67db005cf12fd9313f37f..50c7e0a449ea9d71e21e86b4f03dac02
      public static final short MAX_STRING_LENGTH = 32767;
      public static final int MAX_COMPONENT_STRING_LENGTH = 262144;
      private static final int PUBLIC_KEY_SIZE = 256;
-@@ -526,8 +527,19 @@ public class FriendlyByteBuf extends ByteBuf {
+@@ -526,8 +527,18 @@ public class FriendlyByteBuf extends ByteBuf {
          return (Component) this.readWithCodecTrusted(NbtOps.INSTANCE, ComponentSerialization.CODEC);
      }
  
@@ -2141,12 +2141,12 @@ index 9373502ede6c8a881af67db005cf12fd9313f37f..50c7e0a449ea9d71e21e86b4f03dac02
 +    }
 +
      public FriendlyByteBuf writeComponent(Component text) {
+-        return this.writeWithCodec(NbtOps.INSTANCE, ComponentSerialization.CODEC, text);
 +        if (text instanceof io.papermc.paper.adventure.AdventureComponent adv) {
 +            return this.writeComponent(adv.adventure$component());
 +        }
 +
-+        // TODO this.adventure$locale
-         return this.writeWithCodec(NbtOps.INSTANCE, ComponentSerialization.CODEC, text);
++        return this.writeWithCodec(NbtOps.INSTANCE, ComponentSerialization.localizedCodec(this.adventure$locale), text);
 +        // Paper end - adventure
      }
  
@@ -2270,14 +2270,35 @@ index d3a80d0a23be762c05931ae8001d98e43cab2b4a..e0037b99838519eee5fb6f7e95ffaad2
          }
      }
 diff --git a/src/main/java/net/minecraft/network/chat/ComponentSerialization.java b/src/main/java/net/minecraft/network/chat/ComponentSerialization.java
-index 49138ccda0f378b13c7f425be765876eb4026b06..ddc91d376fd59da982e594188fcef7202a2a5972 100644
+index 49138ccda0f378b13c7f425be765876eb4026b06..d6008d7d800ecfa7c97b9c89a6058a17342d8552 100644
 --- a/src/main/java/net/minecraft/network/chat/ComponentSerialization.java
 +++ b/src/main/java/net/minecraft/network/chat/ComponentSerialization.java
-@@ -61,6 +61,33 @@ public class ComponentSerialization {
+@@ -55,12 +55,61 @@ public class ComponentSerialization {
+         return ExtraCodecs.orCompressed(mapCodec3, mapCodec2);
+     }
+ 
++    // Paper start - adventure
++    private static final java.util.Map<java.util.Locale, Codec<Component>> LOCALIZED_CODECS = new java.util.concurrent.ConcurrentHashMap<>();
++
++    public static Codec<Component> localizedCodec(java.util.Locale locale) {
++        return LOCALIZED_CODECS.computeIfAbsent(locale,
++            loc -> ExtraCodecs.recursive("Component", selfCodec -> createCodec(selfCodec, loc)));
++    }
++    // Paper end
++
+     private static Codec<Component> createCodec(Codec<Component> selfCodec) {
++        // Paper start - adventure
++        return createCodec(selfCodec, null);
++    }
++
++    private static Codec<Component> createCodec(Codec<Component> selfCodec, @javax.annotation.Nullable java.util.Locale locale) {
++        // Paper end
+         ComponentContents.Type<?>[] types = new ComponentContents.Type[]{PlainTextContents.TYPE, TranslatableContents.TYPE, KeybindContents.TYPE, ScoreContents.TYPE, SelectorContents.TYPE, NbtContents.TYPE};
+         MapCodec<ComponentContents> mapCodec = createLegacyComponentMatcher(types, ComponentContents.Type::codec, ComponentContents::type, "type");
          Codec<Component> codec = RecordCodecBuilder.create((instance) -> {
              return instance.group(mapCodec.forGetter(Component::getContents), ExtraCodecs.strictOptionalField(ExtraCodecs.nonEmptyList(selfCodec.listOf()), "extra", List.of()).forGetter(Component::getSiblings), Style.Serializer.MAP_CODEC.forGetter(Component::getStyle)).apply(instance, MutableComponent::new);
          });
-+        // Paper start
++        // Paper start - adventure
 +        final Codec<Component> origCodec = codec;
 +        codec = new Codec<>() {
 +            @Override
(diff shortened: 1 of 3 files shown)


electronicboy (approved):

Machine-Maker:
I wonder if it's worth hooking into vanilla's serialization logic (gson in this version, codecs in the next version) and render the translatable components when serializing to json. That would avoid the double conversion.

booky10:
How would that work with adventure's translator? The adventure translator only accepts adventure components, not vanilla components or json.

lynxplay:
Well technically there is Translator's `MessageFormat translate(final @NotNull String key, final @NotNull Locale locale);`
that could be used, however that was recently expanded to also have a method that takes a TranslatableComponent.

Machine-Maker:
I was thinking along the lines of... if, while serializing a vanilla component, it encounters a translatable, it can only convert that component to adventure, render it, and then serialize it to json. That way, *most* components, or chunks of components avoid the double serialization

booky10:
I think I've now implemented your suggested solution to this problem. I've used a `ThreadLocal` to store the adventure locale, as the static gson serializer doesn't contain (or support) any context as far as I know. I'm open to suggestions on alternatives to store the locale.

Machine-Maker:
I think it can be even better. Can't we first check if the translation key exists in adventure's global translator? If it doesn't, do we need to bother running it through adventure's renderer? Maybe @kashike can weigh in on this if I'm missing some functionality of adventure's translation system.

booky10:
> I think it can be even better. Can't we first check if the translation key exists in adventure's global translator? If it doesn't, do we need to bother running it through adventure's renderer? Maybe @kashike can weigh in on this if I'm missing some functionality of adventure's translation system.

I'm not sure this is possible. The adventure `Translator` interface accepts `TranslatableComponent`s or translation keys, but doesn't provide API for listing registered translation keys. A `TranslationRegistry` does have this information.

Machine-Maker:
Hmm yeah... I guess there isn't a way to know if it can be converted. 

This might be as good as it gets then for now.

patches/server/0010-Adventure.patch
```diff
@@ -977,6 +977,10 @@ index 0000000000000000000000000000000000000000..3dc613116c086444ece88bcb0a569eee
 +    // Component
 +
 +    public static Component asAdventure(final net.minecraft.network.chat.Component component) {
++        if (component instanceof AdventureComponent) {
```
dawon: You can use instanceof cast, that would simplify the code and make it more readable:

```java
if (component instanceof AdventureComponent aComp) {
           // don't convert to json and back if already backed by an adventure Component
           return aComp.adventure;
}
```

booky10: This style is not used anywhere else in the class (see line 1030), Idk if this should be changed.

Machine-Maker: Well... some of the class was written before that was a thing. I think we'd prefer to use the newer instanceof where we can.

booky10: Ok, changed :+1:

booky10:
I've rewritten this PR for 1.20.3/1.20.4. This now translates translatable components when encoding using the serialization codec. I don't have much experience in working with Codecs, so I don't know if there is a more optiomal spot for converting and translating the components.

Machine-Maker (dismissed):
So while I think the changes work... I would rather not use a ThreadLocal for this I think. Since this CODEC is global, it runs the risk of causing a problem. Lets say in someone's custom Translator, they do something that causes that CODEC to be used, since its on the same thread it will cause the rendering to take effect. I would rather there be a separate "rendering" codec that can be created with a locale. Could be a global map of them with the locale as the key, created as needed. Then that codec would be used to convert the component with the rendering/translating.

This would be done by extracting out that last return statement in createCodec into a separate method and returning the direct Codec<Component> into a field instead. Then, a separate method could be created that took a Locale and wrapped that component in the eithers with the rendering logic.

booky10:
I've now removed the ThreadLocal now and added a map with the translating codecs. I'm not quite sure how storing the direct codec would work, as I think children of components wouldn't be translated because its a recursive codec and we would only translate the component if the parent is also translatable.

Machine-Maker:
What you ended up doing is fine, it is probably just possible to not have to recreate the whole codec each time, and still retain the recursive nature of it. But idk how big a deal that is, not like its gonna be creating a zillion codecs anyways.

booky10:
<details>
<summary>ComponentSerialization Diff</summary>

``` patch
Index: src/main/java/net/minecraft/network/chat/ComponentSerialization.java
IDEA additional info:
<+>UTF-8
diff --git a/src/main/java/net/minecraft/network/chat/ComponentSerialization.java b/src/main/java/net/minecraft/network/chat/ComponentSerialization.java
--- a/src/main/java/net/minecraft/network/chat/ComponentSerialization.java	(revision 553d84e32f0649d97ac888f9a96c3f5d1d954c2c)
+++ b/src/main/java/net/minecraft/network/chat/ComponentSerialization.java	(date 1702305732428)
@@ -32,6 +32,7 @@
     });
 
     // Paper start - adventure localization
+    private static final Codec<Component> DIRECT_CODEC = ExtraCodecs.recursive("Component", ComponentSerialization::createDirectCodec);
     private static final java.util.Map<java.util.Locale, Codec<Component>> TRANSLATING_CODECS = new java.util.HashMap<>();
 
     public static Codec<Component> getTranslatingCodec(@javax.annotation.Nullable java.util.Locale locale) {
@@ -69,18 +70,23 @@
         return ExtraCodecs.orCompressed(mapCodec3, mapCodec2);
     }
 
-    private static Codec<Component> createCodec(Codec<Component> selfCodec) {
-        // Paper start - adventure localization
-        return createCodec(selfCodec, null);
-    }
-
-    private static Codec<Component> createCodec(Codec<Component> selfCodec, @javax.annotation.Nullable java.util.Locale locale) {
-        // Paper end
+    private static Codec<Component> createDirectCodec(Codec<Component> selfCodec) { // Paper - adventure localization
         ComponentContents.Type<?>[] types = new ComponentContents.Type[]{PlainTextContents.TYPE, TranslatableContents.TYPE, KeybindContents.TYPE, ScoreContents.TYPE, SelectorContents.TYPE, NbtContents.TYPE};
         MapCodec<ComponentContents> mapCodec = createLegacyComponentMatcher(types, ComponentContents.Type::codec, ComponentContents::type, "type");
         Codec<Component> codec = RecordCodecBuilder.create((instance) -> {
             return instance.group(mapCodec.forGetter(Component::getContents), ExtraCodecs.strictOptionalField(ExtraCodecs.nonEmptyList(selfCodec.listOf()), "extra", List.of()).forGetter(Component::getSiblings), Style.Serializer.MAP_CODEC.forGetter(Component::getStyle)).apply(instance, MutableComponent::new);
         });
+        // Paper start - adventure localization
+        return codec;
+    }
+
+    private static Codec<Component> createCodec(Codec<Component> selfCodec) {
+        return createCodec(selfCodec, null);
+    }
+
+    private static Codec<Component> createCodec(Codec<Component> selfCodec, final @javax.annotation.Nullable java.util.Locale locale) {
+        final Codec<Component> codec = DIRECT_CODEC;
+        // Paper end
         return Codec.either(Codec.either(Codec.STRING, ExtraCodecs.nonEmptyList(selfCodec.listOf())), codec).xmap((either) -> {
             return either.map((either2) -> {
                 return either2.map(Component::literal, ComponentSerialization::createFromList);
```

</details>

<details>
<summary>TestPlugin Diff</summary>

```diff
Index: test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java b/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java
--- a/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java	(revision 75d74272d91cc1b00a2e552d61b20a1d57cae799)
+++ b/test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java	(date 1702305590501)
@@ -1,5 +1,16 @@
 package io.papermc.testplugin;
 
+import io.papermc.paper.event.player.AsyncChatEvent;
+import java.text.MessageFormat;
+import java.util.Locale;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.translation.GlobalTranslator;
+import net.kyori.adventure.translation.TranslationRegistry;
+import org.bukkit.Bukkit;
+import org.bukkit.Location;
+import org.bukkit.NamespacedKey;
+import org.bukkit.entity.ArmorStand;
+import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.plugin.java.JavaPlugin;
 
@@ -8,5 +19,28 @@
     @Override
     public void onEnable() {
         this.getServer().getPluginManager().registerEvents(this, this);
+
+        TranslationRegistry registry = TranslationRegistry.create(new NamespacedKey(this, "i18n"));
+        registry.register("testplugin.armorstand", Locale.ENGLISH, new MessageFormat("Success! ({0})"));
+        registry.register("testplugin.armorstand.arg", Locale.ENGLISH, new MessageFormat("Argument"));
+        GlobalTranslator.translator().addSource(registry);
+    }
+
+    @EventHandler
+    public void onChat(AsyncChatEvent event) {
+        Bukkit.getScheduler().runTask(this, () -> {
+            Location spawnLoc = event.getPlayer().getLocation();
+            spawnLoc.getWorld().spawn(spawnLoc, ArmorStand.class, armorStand -> {
+                armorStand.setCustomNameVisible(true);
+                armorStand.customName(Component.translatable("testplugin.armorstand",
+                    Component.translatable("testplugin.armorstand.arg")));
+            });
+            spawnLoc.add(5d, 0d, 0d);
+            spawnLoc.getWorld().spawn(spawnLoc, ArmorStand.class, armorStand -> {
+                armorStand.setCustomNameVisible(true);
+                armorStand.customName(Component.text("Testing: ")
+                    .append(Component.translatable("testplugin.armorstand.arg")));
+            });
+        });
     }
 }
```

</details>

With these changes, Components are now no longer translated recursively (as expected):
image

I'm not sure if recursive translation will be possible when saving the direct codec because of how recursive codecs work - or I've just messed something else up.

patches/server/0010-Adventure.patch
```diff
@@ -1848,6 +1861,82 @@ index d3a80d0a23be762c05931ae8001d98e43cab2b4a..e0037b99838519eee5fb6f7e95ffaad2
...
+     });
+ 
++    // Paper start - adventure localization
++    private static final java.util.Map<java.util.Locale, Codec<Component>> TRANSLATING_CODECS = new java.util.HashMap<>();
++
++    public static Codec<Component> getTranslatingCodec(@javax.annotation.Nullable java.util.Locale locale) {
++        if (locale == null) {
++            return CODEC;
++        }
++        synchronized (TRANSLATING_CODECS) {
```
jpenilla: just use a ConcurrentHashMap instead

jpenilla:
Just a heads up that you probably want to wait until #10028 and #10014 are merged to make more changes, as they touch the same areas.

patches/server/0010-Adventure.patch
```diff
@@ -2288,11 +2309,18 @@ index 49138ccda0f378b13c7f425be765876eb4026b06..ddc91d376fd59da982e594188fcef720
 +            @Override
 +            public <T> DataResult<T> encode(final Component input, final DynamicOps<T> ops, final T prefix) {
 +                if (input instanceof io.papermc.paper.adventure.AdventureComponent adv) {
-+                    if (adv.deepConvertedIfPresent() != null) {
-+                        return origCodec.encode(java.util.Objects.requireNonNull(adv.deepConvertedIfPresent()), ops, prefix);
-+                    } else {
-+                        // return io.papermc.paper.adventure.PaperAdventure.localizedCodec(locale).encode(adv.adventure$component(), ops, prefix); // TODO
-+                        return io.papermc.paper.adventure.AdventureCodecs.COMPONENT_CODEC.encode(adv.adventure$component(), ops, prefix);
++                    return io.papermc.paper.adventure.PaperAdventure.localizedCodec(locale)
```
jpenilla: I think this needs a null check for the Locale, falling back to the AdventureCodecs.COMPONENT_CODEC

patches/server/0010-Adventure.patch
```diff
@@ -2288,11 +2309,18 @@ index 49138ccda0f378b13c7f425be765876eb4026b06..ddc91d376fd59da982e594188fcef720
...
++                        .encode(adv.adventure$component(), ops, prefix);
++                }
++                if (locale != null && input.getContents() instanceof TranslatableContents) {
++                    final net.kyori.adventure.text.Component adventureComponent
++                        = io.papermc.paper.adventure.PaperAdventure.asAdventure(input);
```
jpenilla: I think once we convert to adventure all we need to do is pass the result into the relevant adventure Codec, like is done above. Then we don't need to make #translated public either.

patches/server/0010-Adventure.patch
jpenilla: Maybe we should just modify the #localizedCodec methods to check for null before using the Map and return the base/non-localized Codec instead

jpenilla (approved):

jpenilla:
I think this misses some cases where ComponentSerialization.CODEC is used directly instead of passing the self codec, like show_text hover events. Solving that is more complicated though, so I think it's fine to leave for later and not block this improvement.

booky10:
> This misses some cases where ComponentSerialization.CODEC is used directly instead of passing the self codec, like show_text hover events. Solving that is more complicated though, so I think it's fine to leave for later and not block this improvement.

Solving this properly would probably require to re-create every codec requiring the Component codec for every locale. A quick workaround for this would be to just pass everything (which has show_text hover events, etc.) through the adventure codec.

jpenilla:
Yeah, both options there are not very pretty, which is why I think we should just leave it for when/if someone finds a use case that justifies it. The only thing I can think of would be translating datapack messages with a plugin, but that feels really niche, and if someone needs that to work they can open an issue.

PR #9895, opened 2023-10-31
Remove "fix-curing-zombie-villager-discount" exploit option

booky10:
MC-181190 was fixed in vanilla with 23w31a and the paper patch doesn't fix anything anymore.


Machine-Maker:
I don't think we have to increment the config version and create a custom transformation to remove the old config. We have a big file full of config settings to remove. See `RemovedConfigurations` and add the correct path for the setting to that instead.

booky10:
> I don't think we have to increment the config version and create a custom transformation to remove the old config. We have a big file full of config settings to remove. See `RemovedConfigurations` and add the correct path for the setting to that instead.

Changed, I didn't knew `RemovedConfiguations` existed :+1:

Machine-Maker (approved):

Phoenix616:
Docs should be adjusted accordingly then: https://github.com/PaperMC/docs/blob/main/docs/paper/admin/reference/configuration/world-configuration.mdx#L633-L636

PR #9898, opened 2023-11-01
add PlayerBrushBlockEvent

### LINKED ISSUE #9774: Add PlayerBrushBlockEvent

As it currently exists, the BlockDropItemEvent is fired after the player has completed brushing a BrushableBlock. This event only controls the dropped items but occurs after the block has already changed from a 'suspicious <type>' to a normal <type> block. A new event which controlled this change and controlled whether the BlockDropItemEvent would provide more customization to the brushing feature introduced in 1.20.

Introduce a new event - PlayerBrushBlockEvent.
This new event can extend the BlockEvent class and have a variable for the Player.

In the BrushableBlockEntity class, move the `this.dropContent(player)` and `this.level.setBlock(this.worldPosition, block1defaultBlockState(), 3);` into an if-statement to check if the brushable block event is cancelled. 

See the attached picture for the possible solution. I attempted to contribute to Paper directly but got a little lost following the Contributing ReadMe documentation for Git and GitHub and gave up.
image

As of right now, the only solution is to listen to the BlockItemDropEvent and check if the block was a suspicious type, using a BukkitRunnable set to a few ticks to "reset" the block if we want the mechanic to be cancelled. A new event to represent this situation is a more clean solution and provides more flexibility for developers.

### Other

_No response_

chmodsayshello:
This is a small pull request adding an event that gets called whenever a player brushes a block.
Closes #9774

### DIFF
--- a/patches/api/0443-add-PlayerBrushBlockEvent.patch
+++ b/patches/api/0443-add-PlayerBrushBlockEvent.patch
@@ -1,0 +1,91 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: chmodsayshello <chmodsayshello@hotmail.com>
+Date: Wed, 1 Nov 2023 13:24:22 +0100
+Subject: [PATCH] add PlayerBrushBlockEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/player/PlayerBrushBlockEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerBrushBlockEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..11a46a23e5f82162401698ebd9142170413b4a7b
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/player/PlayerBrushBlockEvent.java
+@@ -0,0 +1,79 @@
++package io.papermc.paper.event.player;
++
++import org.bukkit.Material;
++import org.bukkit.block.Block;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.entity.Player;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public class PlayerBrushBlockEvent extends PlayerEvent implements Cancellable {
++    private static final HandlerList handlerList = new HandlerList();
++    private boolean isCancelled = false;
++    private final Block block;
++    private ItemStack drop;
++
++    public PlayerBrushBlockEvent(final @NotNull Player player, @NotNull final Block block, @NotNull final ItemStack drop){
++        super(player);
++        this.block = block;
++        this.drop = drop;
++    }
++
++    /**
++     * Get the block the player tries to brush
++     *
++     * @return The block being brushed
++     */
++    @NotNull
++    public Block getBlock() {
++        return this.block;
++    }
++
++    /**
++     * Gets the drop from the block
++     *
++     * @return The Itemstack of the drop
++     */
++    @NotNull
++    public ItemStack getDrop() {
++        return this.drop;
++    }
++
++    /**
++     * Sets the drop for the suspicious block
++     *
++     * @param drop The Itemstack to be dropped
++     */
++    public void setDrop(@NotNull final ItemStack drop) {
++        this.drop = drop;
++    }
++
++    @Override
++    public boolean isCancelled() {
++        return this.isCancelled;
++    }
++
++    /**
++     * Cancel the Event
++     *
++     * @param cancel if true, nothing will drop, and suspicious blocks will not turn into normal ones
++     */
++    @Override
++    public void setCancelled(final boolean cancel) {
++        this.isCancelled = cancel;
++    }
++
++    @Override
++    @NotNull
++    public HandlerList getHandlers() {
++        return getHandlerList();
++    }
++
++    @NotNull
++    public static HandlerList getHandlerList() {
++        return handlerList;
++    }
++}
--- a/patches/server/1044-add-PlayerBrushBlockEvent.patch
+++ b/patches/server/1044-add-PlayerBrushBlockEvent.patch
@@ -1,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: chmodsayshello <chmodsayshello@hotmail.com>
+Date: Wed, 1 Nov 2023 14:01:08 +0100
+Subject: [PATCH] add PlayerBrushBlockEvent
+
+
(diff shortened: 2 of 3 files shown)


patches/server/1044-add-PlayerBrushBlockEvent.patch
```diff
@@ -0,0 +1,51 @@
...
+         }
+ 
+         return InteractionResult.CONSUME;
+@@ -86,8 +94,7 @@ public class BrushItem extends Item {
+                             BlockEntity var18 = world.getBlockEntity(blockPos);
+                             if (var18 instanceof BrushableBlockEntity) {
+                                 BrushableBlockEntity brushableBlockEntity = (BrushableBlockEntity)var18;
+-                                boolean bl2 = brushableBlockEntity.brush(world.getGameTime(), player, blockHitResult.getDirection());
+-                                if (bl2) {
++                                if (brushableBlockEntity.brush(world.getGameTime(), player, blockHitResult.getDirection())) { // Paper - remove "bl2" boolean (already defined)
```
chmodsayshello: To clarify: even though I did not change this function, Paper didn't compile due to there being another variable within the same scope, also called "bl2". Thats why I had to make this change...

Warriorrrr: Decompile fixes can go into the mc dev fixes patch

chmodsayshello: done, thanks!

lynxplay:
Welcome to paper :wave: thank you for your first PR!

I am a tad unsure if your current implementation does actually fix the related issue.
Right now, this mostly looks like an event that is pretty much just a duplicate of the PlayerInteractEvent.

If I interpret the issue correctly, the event should be called when the player is basically done brushing the block instead of the start of the brushing process.

chmodsayshello:
> Welcome to paper 👋 thank you for your first PR!

Thanks, but this isn't my first PR, have some old ones I have to clean up some day

> If I interpret the issue correctly, the event should be called when the player is basically done brushing the block instead of the start of the brushing process.

Yeah, my bad, sorry, the event will no longer be cancel-able though, however, then, I can expose the dropped itemstack

lynxplay:
Ah sorry, just saw "first time contributor" :sweat: 
I mean, plugins can infer a not cancelled "PlayerBrushBlockEvent" that is called exactly how the issue describes it (cancellable, does not expose item drops) and the following BlockDropItem event stuff.

lynxplay:
As a follow up, warrior suggested internally that maybe an event called each time the block is "partially" brushed in the BrushItem useOnTick method would be useful.

Would work both as a "player starts brushing" and a progress event to do stuff with.

chmodsayshello:
> As a follow up, warrior suggested internally that maybe an event called each time the block is "partially" brushed in the BrushItem useOnTick method would be useful.

What would be the use of that? The server cannot stop the client from showing the animation to my knowledge

chmodsayshello:
I'm sorry for this force push mess, had quite some trouble reverting my changes to the decompile-fixes patch... I'll squash the PR a final time soon

lynxplay:
No worries, we squash on merge xD Go wild with the commits.
I mean, I guess the "per brush tick" event is not really needed if brush would properly call events when it converts the block types to the "more brushed" state

Machine-Maker (changes requested):
I do not think this event is needed at all. You can accomplish the same things by listening to the BlockDropItemEvent which, contrary to the linked issue, is fired **before** the block is changed allowing you to change the dropped items.

You can then just add a call to the EntityChangeBlockEvent before the blockstate is changed when the brushing is completed which you can cancel to prevent the brushable block from changing.

chmodsayshello:
> I do not think this event is needed at all. You can accomplish the same things by listening to the BlockDropItemEvent which, contrary to the linked issue, is fired **before** the block is changed allowing you to change the dropped items.
> 
> You can then just add a call to the EntityChangeBlockEvent before the blockstate is changed when the brushing is completed which you can cancel to prevent the brushable block from changing.

In my opinion, having control over this in a singular event is better than having to listen to these two, as then, finding out if they were triggered due to that same block, same player, same time would be slightly harder.
Also, couldn't a lot of events just be done plugin side using the PlayerInteractionEvent?

Anyways, feel free to close this PR as "won't add" if you disagree with me :)

patches/api/0443-add-PlayerBrushBlockEvent.patch
```diff
@@ -0,0 +1,91 @@
...
++    }
++
++    /**
++     * Gets the drop from the block
++     *
++     * @return The Itemstack of the drop
++     */
++    @NotNull
++    public ItemStack getDrop() {
++        return this.drop;
```
yannicklamprecht, line 54: Should return a copy instead, so that setType / setMetaData doesn't change the drops. Instead devs should be encouraged to use setDrops afterwards.

patches/api/0443-add-PlayerBrushBlockEvent.patch
```diff
@@ -0,0 +1,91 @@
...
++import org.bukkit.Material;
++import org.bukkit.block.Block;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.entity.Player;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
```
yannicklamprecht, line 24: Java doc missing.

patches/server/1044-add-PlayerBrushBlockEvent.patch
```diff
@@ -0,0 +1,29 @@
...
+     private void brushingCompleted(Player player) {
+         if (this.level != null && this.level.getServer() != null) {
++            // Paper start - PlayerBrushBlockEvent
++            io.papermc.paper.event.player.PlayerBrushBlockEvent event = new io.papermc.paper.event.player.PlayerBrushBlockEvent(
++                (org.bukkit.entity.Player) player.getBukkitEntity(),
++                org.bukkit.craftbukkit.block.CraftBlock.at(this.getLevel(), this.getBlockPos()),
++                item.asBukkitCopy()
++            );
++            event.callEvent();
++            if(event.isCancelled()) {
```
yannicklamprecht, line 22: Diff can be reduced to the following. callEvent() returns a boolean, by looking into the docs you'll see that it does what you did in 3 statements.
```suggestion
+            if(new io.papermc.paper.event.player.PlayerBrushBlockEvent((org.bukkit.entity.Player) player.getBukkitEntity(), org.bukkit.craftbukkit.block.CraftBlock.at(this.getLevel(), this.getBlockPos()), item.asBukkitCopy()).callEvent()) {
```

yannicklamprecht (changes requested):

codebycam:
Hey @chmodsayshello, we're going to close this PR for the reason Machine-Maker mentioned [above](https://github.com/PaperMC/Paper/pull/9898#pullrequestreview-1726114004). 
Please join the [Discord](https://discord.com/invite/papermc) if you need any assistance with the API!

PR #9901, opened 2023-11-02
Add more API to LingeringPotionSplashEvent

### LINKED ISSUE #9363: LingeringPotionSplashEvent should have a "force spawn" option

In vanilla, an "empty" lingering potion, or a lingering potion without any effects will not produce an area effect cloud. Upstream breaks this, and Paper fixes it. But it would certainly be useful to be able to force an empty area effect cloud to exist.

I'm not sure of the best approach, but here is some options.

The **best** option from an API design standpoint would be to fire the LingeringPotionSplashEvent as cancelled when the effects are empty, and uncancelling it would create an area effect cloud. However, this is probably too large of an API break.

The other option is to just add some "forceCreate" boolean field to the event to force the cloud to be created if its empty. I'm not a huge fan of this, cause this "forceCreate" option would only come into play if there were no effects, which is a minority of times this event fires.

Just manually creating the effect cloud yourself when listening to the LingeringPotionSplashEvent. Kind of annoying to do this, as you have to rely on hardcoded impl details about the default parameters for the cloud, like size, age, etc. 

### Other

_No response_

notTamion:
resolves #9363 
This allows to completely change if the AreaEffectCloud gets created, not only force it getting created
not sure if how i implemented this is good i am hoping for some feedback. i think the method names and jdocs are gonna need some work so i am definitely also waiting on feedback for that.


patches/api/0353-More-Projectile-API.patch
```diff
@@ -417,3 +417,49 @@ index c8015ff610e3c1222cb368ea1d8a0c2f3785d9c7..02584eced96944a551140f8150c65a7c
...
  // Paper end
+diff --git a/src/main/java/org/bukkit/event/entity/LingeringPotionSplashEvent.java b/src/main/java/org/bukkit/event/entity/LingeringPotionSplashEvent.java
+index 1584c6c41c3ca51b8ab7f48efb140b0fe31f535b..7ec304fc2a543d7b73c58c6b2f7a50b5e7bfa870 100644
+--- a/src/main/java/org/bukkit/event/entity/LingeringPotionSplashEvent.java
++++ b/src/main/java/org/bukkit/event/entity/LingeringPotionSplashEvent.java
+@@ -17,6 +17,7 @@ public class LingeringPotionSplashEvent extends ProjectileHitEvent implements Ca
+     private static final HandlerList handlers = new HandlerList();
+     private boolean cancelled;
+     private final AreaEffectCloud entity;
++    private boolean createAreaEffectCloud;
```
Doc94: comments?

notTamion: ah woops forgot those

notTamion:
I hope the method name i fine

Machine-Maker (approved):

PR #9907, opened 2023-11-04
Include packet class name in packet encoding error messages

Warriorrrr:
9/10 times the class name will be in the stacktrace when this happens, but I figured it would be useful to also include it in the error message itself, as it currently only mentions the packet's id


Machine-Maker (approved):

Machine-Maker:
I moved the change into the patch that adds the ObfHelper so its not completely out-of-order.

lynxplay (approved):

PR #9908, opened 2023-11-04
Remove duplicate water-sensitivity damage for unaware mobs

Lulu13022002:
The drown damage is already dealt in another method even for unaware mobs, i guess previously it was dealt in that same method causing the return to break that.


Warriorrrr (approved):

Machine-Maker (approved):

Machine-Maker:
To clarify, this is actually the damage that water sensitive mobs take when exposed to water. So its not caused by "drowning" despite what the damage cause is. It's for like blazes and snow golems.

PR #9909, opened 2023-11-04
Fix duplicate egg throw sound

Lulu13022002:
This fix an issue with the egg throw sound being played twice (slightly noticeable with the volume of the sound compared to other projectile) and the enderpearl that didn't reset its cooldown clientside when the PlayerLaunchProjectileEvent is cancelled.

### DIFF
--- a/patches/server/0215-PlayerElytraBoostEvent.patch
+++ b/patches/server/0215-PlayerElytraBoostEvent.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] PlayerElytraBoostEvent
 
 
 diff --git a/src/main/java/net/minecraft/world/item/FireworkRocketItem.java b/src/main/java/net/minecraft/world/item/FireworkRocketItem.java
-index 7c627d27300247db9122ab2081049345ef306073..6e6c9477b70b0e1ab36bca9fa91806a283ebedd1 100644
+index 7c627d27300247db9122ab2081049345ef306073..badd3ff4b8e716c0ac255397c8592b72d8a6e4db 100644
 --- a/src/main/java/net/minecraft/world/item/FireworkRocketItem.java
 +++ b/src/main/java/net/minecraft/world/item/FireworkRocketItem.java
-@@ -61,12 +61,19 @@ public class FireworkRocketItem extends Item {
+@@ -61,8 +61,17 @@ public class FireworkRocketItem extends Item {
              if (!world.isClientSide) {
                  FireworkRocketEntity fireworkRocketEntity = new FireworkRocketEntity(world, itemStack, user);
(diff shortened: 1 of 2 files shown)


Machine-Maker:
In all the cases where the launch projectile event is fired... shouldn't we uncomment the `user.awardStat(Stats.ITEM_USED.get(this));` line at the bottom and remove the one wee add? It still would be correct and less vanilla diff. The only thing we would need to handle above was the itemstack shrink. You could also just leave that at the bottom if you added a new boolean at the top that determined if the stack should shrink. I think that is cleaner.

Also, for `EggItem` specifically, we can leave the playSound at the bottom too since that will fire after the event anyways.

patches/server/0223-PlayerLaunchProjectileEvent.patch
```diff
@@ -5,267 +5,255 @@ Subject: [PATCH] PlayerLaunchProjectileEvent
...
-+                    ((net.minecraft.server.level.ServerPlayer) user).getBukkitEntity().updateInventory();
-+                }
-+
-+                world.playSound((Player) null, user.getX(), user.getY(), user.getZ(), SoundEvents.ENDER_PEARL_THROW, SoundSource.NEUTRAL, 0.5F, 0.4F / (net.minecraft.world.entity.Entity.SHARED_RANDOM.nextFloat() * 0.4F + 0.8F));
-+                user.awardStat(Stats.ITEM_USED.get(this));
-+                user.getCooldowns().addCooldown(this, 20);
-+            } else {
++            com.destroystokyo.paper.event.player.PlayerLaunchProjectileEvent event = new com.destroystokyo.paper.event.player.PlayerLaunchProjectileEvent((org.bukkit.entity.Player) user.getBukkitEntity(), org.bukkit.craftbukkit.inventory.CraftItemStack.asCraftMirror(itemstack), (org.bukkit.entity.Projectile) entityenderpearl.getBukkitEntity());
++            if (!event.callEvent() || !world.addFreshEntity(entityenderpearl)) {
++                user.getCooldowns().removeCooldown(this);
```
Machine-Maker: Why do we need to remove the cooldown here? I don't think the client adds to its own cooldown.

Lulu13022002: The client seems to predict the cooldown and still run it at least visually.

Machine-Maker: I wonder if we should just send the packet itself then, instead of relying on that logic. Like what if we put an event call inside removeCooldown?

Lulu13022002: That's fair, updated to just send the packet

PR #9910, opened 2023-11-05
Add HiddenPotionEffect API

### LINKED ISSUE #9811: Get Hidden Weaker Potion Effects

notTamion:
supersedes #9831 
This implements the HiddenPotionEffect API that Machine proposed on Discord voice chat. This didn't seem like it would fit any existing patch so i created new ones
closes #9811


Machine-Maker (approved):
Looks good, still need to test in game. Also, its possible there is an existing patch this can be merged into, not sure.

notTamion:
currently when creating a new PotionEffect that has a hidden PotionEffect and adding that to a PotionMeta the hidden effect won't get added to the PotionMeta cause afaik there is no NBT for hidden effects or anything like that on a Potion. I don't think documenting this is sufficient. Perhaps just adding the hidden effects (if there are any) as normal custom effects to the PotionMeta is a solution? Please leave some feedback on this

Owen1212055 (approved):

Machine-Maker (approved):

Machine-Maker (approved):
Tested to ensure serialization works with the new hidden effect. Also tested that the conversion properly works by giving myself stacked effects.

lynxplay (dismissed):

lynxplay (approved):

PR #9916, opened 2023-11-07
Configurable memory condition tick rate

ruViolence:
This setting will help limit the tick rate of OneShot entity tasks because "tick-rate.behaviour" does not limit OneShot tasks.

I did this because players were creating villager farms and the existing settings were not helping me at all:
img

### DIFF
--- a/patches/server/1047-Configurable-memory-condition-tick-rate.patch
+++ b/patches/server/1047-Configurable-memory-condition-tick-rate.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Configurable memory condition tick rate
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java b/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
-index 6f323026b2d6fb5151e8b9dd9466e96eb026ccbd..6aeb432d7e79cbf5c64da934bd6d4d63a7a35e44 100644
+index 6f323026b2d6fb5151e8b9dd9466e96eb026ccbd..4be6a59a7752c35b3a86c62db53735464e57e393 100644
 --- a/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
 +++ b/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
 @@ -333,8 +333,19 @@ public class BehaviorBuilder<E extends LivingEntity, M> implements App<BehaviorB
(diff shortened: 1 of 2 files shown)


patches/server/1047-Configurable-memory-condition-tick-rate.patch
```diff
@@ -0,0 +1,30 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java b/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
+index 6f323026b2d6fb5151e8b9dd9466e96eb026ccbd..6aeb432d7e79cbf5c64da934bd6d4d63a7a35e44 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
++++ b/src/main/java/net/minecraft/world/entity/ai/behavior/declarative/BehaviorBuilder.java
+@@ -333,8 +333,19 @@ public class BehaviorBuilder<E extends LivingEntity, M> implements App<BehaviorB
+     static final class PureMemory<E extends LivingEntity, F extends K1, Value> extends BehaviorBuilder<E, MemoryAccessor<F, Value>> {
+         PureMemory(final MemoryCondition<F, Value> query) {
+             super(new BehaviorBuilder.TriggerWithResult<E, MemoryAccessor<F, Value>>() {
++                // Paper start - memory condition tick rate
++                private final String configKey = query.memory().toString().toLowerCase(java.util.Locale.ROOT);
```
Machine-Maker: I think using `Locale.ENGLISH` is better here? We don't want it changing across systems which I *think* ROOT might?

Machine-Maker (changes requested):

patches/server/1047-Configurable-memory-condition-tick-rate.patch
Sytm: To quote the [Javadocs of the root locale](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Locale.html#ROOT):

> This is regarded as the base locale of all locales, and is used as the language/country neutral locale for the locale sensitive operations.

So it should be preferred for operations like this

ruViolence: I copied this from that commit. Are you sure I need to change it to `ENGLISH`?
https://github.com/PaperMC/Paper/commit/12942dc94ba42043461a4d5b5c049a6e53f7c2ee#diff-3e8d15a07d2e1a33d90f0ab38c54c0457d715330c7e478cb55e1ac794d8526a5R134

Machine-Maker: Ok, sounds like I was under the wrong impression of what it was. `ROOT` is fine.

Machine-Maker: Ok, looking at the impl here, I don't really like using the toString() method here. Just call the getKey() method on `BuiltInRegistries.MEMORY_MODULE_TYPE` directly. Best to toString the ResourceLocation directly as that has less chance of changing in the future.

Machine-Maker (changes requested):

PR #9921, opened 2023-11-08
Add ChunkStatusChangeEvent

Cryptite:
Chunk Statuses are a wily thing, but it is nice to know when the state of one changes for concept like lazy-loading, etc.

I'm not sure this event should ever be anything other than read-only but this basic version has worked well for me for awhile.


patches/api/0446-ChunkStatusChangeEvent.patch
```diff
@@ -0,0 +1,71 @@
...
++    public static HandlerList getHandlerList() {
++        return handlers;
++    }
++
++    public enum ChunkStatus {
++        INACCESSIBLE,
++        FULL,
++        BLOCK_TICKING,
++        ENTITY_TICKING;
++    }
```
Owen1212055: Cant we use Chunk$LoadLevel?

electronicboy: kinda, but they messed up the order of them so it breaks the nice ordinal thing

patches/api/0446-ChunkStatusChangeEvent.patch
```diff
@@ -0,0 +1,71 @@
...
+From: Cryptite <cryptite@gmail.com>
+Date: Wed, 8 Nov 2023 09:23:23 -0600
+Subject: [PATCH] ChunkStatusChangeEvent
+
+
+diff --git a/src/main/java/com/destroystokyo/paper/event/chunk/ChunkStatusChangeEvent.java b/src/main/java/com/destroystokyo/paper/event/chunk/ChunkStatusChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..9a556fe4321ad1b4809c0870a484124284239a7f
+--- /dev/null
++++ b/src/main/java/com/destroystokyo/paper/event/chunk/ChunkStatusChangeEvent.java
```
notTamion: new events should usually go into a sub-package of io.papermc.paper.event

patches/api/0446-ChunkStatusChangeEvent.patch
Cryptite: Yeah, was nice for the isUpgrade() method

patches/api/0446-ChunkStatusChangeEvent.patch
```diff
@@ -0,0 +1,71 @@
...
+From: Cryptite <cryptite@gmail.com>
+Date: Wed, 8 Nov 2023 09:23:23 -0600
+Subject: [PATCH] ChunkStatusChangeEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/chunk/ChunkStatusChangeEvent.java b/src/main/java/io/papermc/paper/chunk/ChunkStatusChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..7d446b6219f425aabdb6a7fbb587f89e8a0f23d8
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/chunk/ChunkStatusChangeEvent.java
```
notTamion: wrong package. you are missing a /event. i think this could go into io.papermc.paper.event.world or a sub-package of that

Spottedleaf:
The event is fired incorrectly. The chunk system changes the chunk status incrementally, and carefully handles concurrent requests to lower it. It is possible that the overall result of the operation is to lower the status, even if initially it was requested to be higher. You can see this by how the function `updateCurrentState` works.

You should look more carefully at the operations done in each status change to determine where to place the event. In general this area of code is the worst part of the chunk system and needs to be refactored, so it may not be possible to actually place an event in there without causing issues.

patches/api/0446-ChunkStatusChangeEvent.patch
Cryptite: Whoops

Cryptite:
> The event is fired incorrectly. The chunk system changes the chunk status incrementally, and carefully handles concurrent requests to lower it. It is possible that the overall result of the operation is to lower the status, even if initially it was requested to be higher. You can see this by how the function `updateCurrentState` works.
> 
> You should look more carefully at the operations done in each status change to determine where to place the event. In general this area of code is the worst part of the chunk system and needs to be refactored, so it may not be possible to actually place an event in there without causing issues.

I do see lower on that `nextState` is captured for subsequent potential status changes. Might be sufficient to just lower the event to after that that `nextState.isOrAfter(currState)` check? `nextState` isn't even used in that method otherwise...

PR #9924, opened 2023-11-10
Add API to retrieve an attribute modifier from a UUID

TonytheMacaroni:
Adds a `getModifier(UUID)` method to `AttributeInstance` to get an attribute modifier using its `UUID`.

Question: Should I merge this diff into the transient modifier patches?


Machine-Maker:
Is it worth exposing the full map somehow? Like an immutable copy of the UUID, AttributeModifier map in the nms AttributeInstance? That would provide more functionality, including getting the modifier by the UUID. 

But if its not useful since there is already a method to get all the modifiers, then we don't need it.

TonytheMacaroni:
Added an additional `AttributeInstance#removeModifier(UUID)` method. As for exposing the full map as an immutable copy, I'm not sure how useful it'd be, especially since you can just get a modifier's UUID from the modifier itself.

patches/server/1047-Add-UUID-attribute-modifier-API.patch
```diff
@@ -0,0 +1,32 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/attribute/CraftAttributeInstance.java
++++ b/src/main/java/org/bukkit/craftbukkit/attribute/CraftAttributeInstance.java
+@@ -44,6 +44,21 @@ public class CraftAttributeInstance implements AttributeInstance {
+         return result;
+     }
+ 
++    // Paper start
++    @Override
++    public AttributeModifier getModifier(java.util.UUID uuid) {
++        Preconditions.checkArgument(uuid != null, "uuid");
```
Lulu13022002: This error message is not helpful, same below

Lulu13022002 (commented):
I think an override is needed in the UnmodifiableAttributeInstance class for the removeModifier

patches/server/1047-Add-UUID-attribute-modifier-API.patch
TonytheMacaroni: Changed the error message.

patches/server/1047-Add-UUID-attribute-modifier-API.patch
```diff
@@ -4,8 +4,32 @@ Date: Thu, 9 Nov 2023 20:34:44 -0500
...
+     }
+ 
++    @Override
++    public void removeModifier(UUID uuid) {
++        throw new UnsupportedOperationException("Cannot modify default attributes");
++    }
++
+     @Override
+     public void addModifier(AttributeModifier modifier) {
+         throw new UnsupportedOperationException("Cannot modify default attributes");
```
TonytheMacaroni: I added the diff to `UnmodifiableAttributeInstance` within this patch instead of the original [default attributes patch](https://github.com/TonytheMacaroni/Paper/blob/master/patches/server/0659-Get-entity-default-attributes.patch#L9-L44) to match the [transient modifiers patch](https://github.com/TonytheMacaroni/Paper/blob/master/patches/server/0943-Add-transient-modifier-API.patch#L7-L22) - should I move it to the original?

patches/server/1047-Add-UUID-attribute-modifier-API.patch
```diff
@@ -15,14 +39,14 @@ index 0029412b1242879deb898524001bb4cc7550fa78..42917d79e20163a44f41c77f8a3d5e24
 +    // Paper start
 +    @Override
 +    public AttributeModifier getModifier(java.util.UUID uuid) {
-+        Preconditions.checkArgument(uuid != null, "uuid");
++        Preconditions.checkArgument(uuid != null, "UUID cannot be null");
```
Sytm:
```suggestion
+        Preconditions.checkNotNull(uuid, "uuid");
```
This would also suffice as a not null check

TonytheMacaroni: Used `Preconditions#checkArgument` to keep in style with the other methods in the class. Also, not a big deal, but `Preconditions#checkNotNull` throws a different exception (`NullPointerException` vs. `IllegalArgumentException`).

Sytm: I checked some other CB classes and it is indeed a mix between `Preconditions.checkArgument(x != null, "x");` and `Preconditions.checkNotNull(x, "x cannot be null");`, so yeah that makes sense to keep in style with the surrounding code

Machine-Maker: Yeah, I think checkArgument is better because we want the illegal argument exception to be thrown.

patches/server/1047-Add-UUID-attribute-modifier-API.patch
Machine-Maker: No, I think it is fine to add here. We do a similar thing with another method we add to AttributeInstance.

Machine-Maker:
I'm just remembering that there are 2 kinds of modifiers, the transient and permanent ones. Do we need to do anything special to handle that with this? Currently the `removeModifier` method will remove either (since they still have to have unique UUIDs).

TonytheMacaroni:
Don't really see the need to distinguish? It's only 'permanent' in the sense that it gets saved, and removal shouldn't be affected by that.

Machine-Maker (approved):

Owen1212055 (approved):

PR #9929, opened 2023-11-11
Add hand to fish event for all player interactions

booky10:
The hand was previously only specified when casting the fishing rod. This is useful for plugins which may want to modify the fishing rod item when reeling in.


patches/server/1047-Add-hand-to-fish-event-for-all-player-interactions.patch
```diff
@@ -0,0 +1,72 @@
...
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java b/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java
+index a9eaa079a43bc8a5e81deaf6df5ce2f9c53cb319..1c8138900c45ed1e547befc40456370638a16521 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java
+@@ -475,6 +475,12 @@ public class FishingHook extends Projectile {
+     public void readAdditionalSaveData(CompoundTag nbt) {}
+ 
+     public int retrieve(ItemStack usedItem) {
```
Machine-Maker: Add `@Deprecated` and `@DoNotUse` to this method. It looks like you changed the only place its used, and we want to catch any new uses of it in the future.

Machine-Maker (changes requested):

Machine-Maker (approved):

lynxplay (approved):

PR #9934, opened 2023-11-11
Fix Mushroom cow stew api

Owen1212055:
Hugely jank, we need to prolly remove those methods asap.

### DIFF
--- a/patches/api/0300-Missing-Entity-API.patch
+++ b/patches/api/0300-Missing-Entity-API.patch
@@ -78,17 +78,18 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
 +}
 diff --git a/src/main/java/io/papermc/paper/potion/SuspiciousEffectEntry.java b/src/main/java/io/papermc/paper/potion/SuspiciousEffectEntry.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..d2e1b9d867394e46e9491bf9e9cfa53c843ef16b
+index 0000000000000000000000000000000000000000..c8446678e39e777bd2c9992d5c577f4c7606ce15
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/potion/SuspiciousEffectEntry.java
-@@ -0,0 +1,36 @@
+@@ -0,0 +1,37 @@
 +package io.papermc.paper.potion;
 +
 +import org.bukkit.potion.PotionEffectType;
++import org.jetbrains.annotations.Contract;
 +import org.jetbrains.annotations.NotNull;
 +
 +/**
-+ * Represents an effect paired with a duration.
++ * Represents a {@link PotionEffectType} paired with a duration.
 + */
 +public sealed interface SuspiciousEffectEntry permits SuspiciousEffectEntryImpl {
 +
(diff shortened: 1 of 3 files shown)


patches/api/0299-Missing-Entity-API.patch
```diff
@@ -76,6 +76,58 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
...
+@@ -0,0 +1,46 @@
++package io.papermc.paper.potion;
++
++import org.bukkit.potion.PotionEffectType;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents an effect paired with a duration.
++ */
++public interface PotionEffectPair {
```
kennytv: This should get a less generic name, it sounds like it _holds_ one or a pair of potion effects (not a type), when really it is just a bare bones potion effect. 

But I also don't know what 😎

kennytv (approved):
Otherwise looks good

Machine-Maker (approved):

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -76,6 +76,60 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
...
++
++    /**
++      * Gets the duration for this effect instance.
++      *
++      * @return duration in ticks
++      */
++    int duration();
++
++    /**
++     * Gets a new instance of PotionEffectPair.
```
Lulu13022002: This interface has been renamed to SuspiciousEffectEntry

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -76,6 +76,60 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
...
++     *
++     * @param effectType effect type
++     * @param duration duration
```
Lulu13022002: The unit should be said somewhere

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -693,34 +747,71 @@ index 939a3dbfcf38f38e4e39d28973ef723157ce0a50..738d547d2a6966122cb2f9f6e94263ee
...
++    default void setStewEffectDuration(int duration) {
++        throw new UnsupportedOperationException("Mushroom cows can now hold multiple effects. Use #setStewEffects");
++    }
 +
 +    /**
 +     * Gets the type of effect applied to stew
 +     * from this mushroom cow is.
 +     *
 +     * @return effect type, or null if an effect is currently not set
++     * @deprecated Mushroom cows can now hold multiple effects
```
Lulu13022002: This message should say which method use instead like in the error thrown below. Same for the other methods

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -693,34 +747,71 @@ index 939a3dbfcf38f38e4e39d28973ef723157ce0a50..738d547d2a6966122cb2f9f6e94263ee
...
++     *
++     * @return immutable effect pair collection
++     */
++    @NotNull
++    java.util.List<io.papermc.paper.potion.SuspiciousEffectEntry> getStewEffects();
++
++    /**
++     * Sets effects applied to stew items for this mushroom cow.
++     *
++     * @param effects effect pair collection
```
Lulu13022002: The class has been renamed, same above

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -76,6 +76,60 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
...
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/potion/SuspiciousEffectEntry.java
+@@ -0,0 +1,36 @@
++package io.papermc.paper.potion;
++
++import org.bukkit.potion.PotionEffectType;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Represents an effect paired with a duration.
```
Lulu13022002: The doc should be more specific that effect here means effectType for all your cases

patches/server/0624-Missing-Entity-API.patch
```diff
@@ -861,32 +862,47 @@ index 6152f4d3d58f4b598d5bd92dbd7c5428c5ff8bc5..a5d3845acc607f640dace05a85f00896
...
      }
  
 +    // Paper start
 +    @Override
-+    public int getStewEffectDuration() {
-+        throw new UnsupportedOperationException(); // TODO https://github.com/PaperMC/Paper/issues/9742
-+    }
++    public java.util.List<io.papermc.paper.potion.SuspiciousEffectEntry> getStewEffects() {
++        if (this.getHandle().stewEffects == null) {
++            return java.util.List.of();
```
Lulu13022002, line 889: The mutability is not consistent, this will be unmodifiable but the other case will be modifiable without any effect on the underlying list.

patches/api/0300-Missing-Entity-API.patch
```diff
@@ -76,6 +76,60 @@ index 0000000000000000000000000000000000000000..39ad7d283609d7e427a2ab35b6fad839
...
++
++    /**
++     * Gets a new instance of PotionEffectPair.
++     *
++     * @param effectType effect type
++     * @param duration duration
++     * @return new instance
++     */
++    @NotNull
++    static SuspiciousEffectEntry suspiciousEffectEntry(PotionEffectType effectType, int duration) {
```
Machine-Maker: So I know I said to name it this for static imports... but I feel like I changed my own mind. We don't do that anywhere else in the API really for static factory-type methods. Do we want to here? Its just kinda long instead of `.create` or something.

kennytv: I like create or of better as well

lynxplay (approved):

lynxplay (approved):

PR #9937, opened 2023-11-12
Fix missing event call for entity teleport API

booky10:
Previously a plugin couldn't cancel or modify an entity-teleport of another plugin.


Owen1212055:
Generally calling events in api methods is a bit iffy?

booky10:
Teleporting players through API also causes an event to be called. Adding an event here is the only way to prevent other plugins from moving an entity, other than teleporting it back every tick if it moved.

Sytm:
If the event should get added for this method, I just want to add that this should then get merged into the patch where this method was added without the surrounding paper comments

booky10:
> If the event should get added for this method, I just want to add that this should then get merged into the patch where this method was added without the surrounding paper comments

Which patch are you referring to? The "More Teleport API" patch is modifying this method, but the initial method is from CraftBukkit/Spigot.

Sytm:
If I am not mistaken this is the teleport method taking optional teleport flags as arguments, and that method has been added by the "More Teleport API" patch

booky10:
Hmmm. Yes, its not the best solution, especially as there is currently no teleport cause at all for entities.

Machine-Maker (approved):

Owen1212055 (approved):

Owen1212055:
In the future this event could be added onto through the teleport API addition patch.

xsmeths:
This appears to prevent Asynchronously teleporting entities, is this intended or an unforeseen side effect?

Owen1212055:
> This appears to prevent Asynchronously teleporting entities without an error being printed in console, is this intended or an unforeseen side effect?

Using our API or through NMS? Async teleportation that you may be doing here is not supported, but please show stacktrace.

xsmeths:
> > This appears to prevent Asynchronously teleporting entities without an error being printed in console, is this intended or an unforeseen side effect?
> 
> Using our API or through NMS? Async teleportation that you may be doing here is not supported, but please show stacktrace.


thank you for your reply, this is no longer an issue, I appreciate you taking your time to reply to me

Brokkonaut:
This new event call without a cause in the event is really an problem. Now it is not possible to allow plugin based teleports while disabling natural teleports without doing a stack walk..

PR #9939, opened 2023-11-12
Use ? super in Consumer/Predicate API

Machine-Maker:
Also reformats any patches that had incorrect Consumer and Predicate params to follow a better codestyle. Eventually we are going to want to do this to the whole API, when we have upstream in source in repo, but for API that we add, we want it to be a part of the patch so we don't clutter up the git history any more than needed.

### DIFF
--- a/patches/api/0099-Additional-world.getNearbyEntities-API-s.patch
+++ b/patches/api/0099-Additional-world.getNearbyEntities-API-s.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] Additional world.getNearbyEntities API's
 Provides more methods to get nearby entities, and filter by types and predicates
 
 diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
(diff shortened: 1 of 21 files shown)


patches/api/0111-Add-getNearbyXXX-methods-to-Location.patch
```diff
@@ -24,24 +24,23 @@ index 5a219f132747eb62204e98de4cb850748a43c9b4..6693e3d8dc2519facb12db981a6b6325
...
 +    /**
 +     * Returns a list of entities within a bounding box centered around a Location.
-+     *
++     * <p>
 +     * Some implementations may impose artificial restrictions on the size of the search bounding box.
 +     *
-+     * @param x 1/2 the size of the box along x axis
-+     * @param y 1/2 the size of the box along y axis
-+     * @param z 1/2 the size of the box along z axis
++     * @param x 1/2 the size of the box along then x axis
```
Lulu13022002: then?

Machine-Maker: should be `the`.

electronicboy (approved):

Owen1212055 (approved):

PR #9941, opened 2023-11-12
Fix MC-172801

### LINKED ISSUE #9827: Bees aren't affected by setting generic fly speed
### Expected behavior

Bees should fly faster or lower depending on the attribute set of generic fly speed but they don't.

### Observed/Actual behavior

There is a bug on the mojang bug website since 1.15 but it never got fixed.

### Steps/models to reproduce

Create a bee entity, change the attribute fly speed and the bee keeps the default value. 

You can test also with the command: `/summon minecraft:bee ~ ~ ~ {Attributes:[{Name:"generic.flying_speed",Base:3.0}]}`

### Plugin and Datapack List

Just a plugin to test this. No datapacks.

### Paper version

[13:09:30 INFO]: Checking version, please wait...
[13:09:30 INFO]: This server is running Paper version git-Paper-234 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: f613437)

### Other

This the link to the mojang bug post [link](https://bugs.mojang.com/browse/MC-172801)

TeakSlack:
Fixes MC-172801

This patch properly applies the generic.flying_speed attribute to bees and parrots. If entities do not have this attribute, they are unaffected. The flying_speed attribute is multiplied by 0.049999999254942 in order to have bees and parrots with the default value of flying_speed be unaffected, having the vanilla 0.02 value.

Closes #9827

### DIFF
--- a/patches/server/1051-Fixes-MC-172801.patch
+++ b/patches/server/1051-Fixes-MC-172801.patch
@@ -1,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Teak Slack <teakislack@gmail.com>
+Date: Sun, 12 Nov 2023 14:38:52 -0500
+Subject: [PATCH] Fixes MC-172801
+
+This patch intends to fix MC-172801, an issue in which the attribute "generic.flying_speed" would not have any effect. This attribute is intended to change the flying speed of parrots and bees,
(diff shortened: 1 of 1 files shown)


kennytv:
I'm pretty sure this is a behavioral change, since most mobs already define their own attributes that differ from the default

PR #9946, opened 2023-11-14
Implement tick rate options for OneShot behaviors

### LINKED ISSUE #9925: Behavior Tick Rates aren't operating as designed
### Expected behavior

The behaviours tick rates configured to the configurable tick rate in the Paper World Config.

unknown_2023 11 10-13 03
(Recorded using Paper-1.19 Latest with same configuration from 1.20.2)

### Observed/Actual behavior

The behaviours tick rates aren't configuring to the configurable tick rate set in the Paper World Config.

unknown_2023 11 10-12 43

### Steps/models to reproduce

1. Set configuration with the following settings in `paper-world-defaults-yml`
```yaml
tick-rates:
  behavior:
    villager:
      validatenearbypoi: 300
      aquirepoi: 300
```
2. Start up server and make a similar build to build shown in the videos.
3. Villager is changing POI as soon as lever is pressed rather than 300 ticks after.

### Plugin and Datapack List

- Essentials
- LuckPerms
- PlugManX
- ViaBackwards
- ViaVersion

### Paper version

This server is running Paper version git-Paper-280 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 4675152)

### Other

_No response_

froobynooby:
TL;DR - 1.19.3 broke Paper's behaviour tick rate settings, this restores the settings' pre-1.19.3 functionality

Some entity AI behaviours were changed in 1.19.3 from having their own classes to becoming instances of the `OneShot `behaviour, including `ValidateNearbyPoi`. This caused the tick rate options to stop working for those behaviours, as the options use the class names as config keys.

This PR re-adds tick rate option support for the behaviours that were affected by the update.

All `OneShot` behaviours are created by a call to the `BehaviorBuilder.create()` method, which happens in a static method in that behaviour's class, for example `ValidateNearbyPoi`:

```java
public class ValidateNearbyPoi {
    private static final int MAX_DISTANCE = 16;

    public static BehaviorControl<LivingEntity> create(Predicate<Holder<PoiType>> poiTypePredicate, MemoryModuleType<GlobalPos> poiPosModule) {
        return BehaviorBuilder.create((context) -> {
            // stuff
        }
    }
}
```

To get the config key for all the `OneShot` behaviours, I've modified the `BehaviorBuilder.create()` method to grab the calling class of the method from the stack trace, which is then used as the config key. In the case of the snippet above, that would be the `ValidateNearbyPoi` class. This maintains compatibility with configs before 1.19.3. Kind of an ugly way of doing it, not sure if I'm missing an easier way.

Fixes #9925

### DIFF
--- a/patches/server/0631-Rate-options-and-timings-for-sensors-and-behaviors.patch
+++ b/patches/server/0631-Rate-options-and-timings-for-sensors-and-behaviors.patch
@@ -28,37 +28,40 @@ index 4bd813161a5d76a83cdbd0a9209b9ea9e60ffe1b..e2764186bd6b838ed5cd86c15597a08d
       * Get a named timer for the specified tile entity type to track type specific timings.
       * @param entity
 diff --git a/src/main/java/net/minecraft/world/entity/ai/behavior/Behavior.java b/src/main/java/net/minecraft/world/entity/ai/behavior/Behavior.java
-index 4ea437253539e3ee533ca9da77a337cbf4d1e807..57ef7fbba3028c28231abf7b7ae78aa019323536 100644
+index 4ea437253539e3ee533ca9da77a337cbf4d1e807..07adec6ce15f9bf37863f881b4ed209520f339a2 100644
 --- a/src/main/java/net/minecraft/world/entity/ai/behavior/Behavior.java
 +++ b/src/main/java/net/minecraft/world/entity/ai/behavior/Behavior.java
-@@ -13,6 +13,10 @@ public abstract class Behavior<E extends LivingEntity> implements BehaviorContro
+@@ -13,6 +13,18 @@ public abstract class Behavior<E extends LivingEntity> implements BehaviorContro
      private long endTimestamp;
      private final int minDuration;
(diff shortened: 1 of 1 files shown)


lynxplay:
Near duplicate of #9946 unless I am understanding the diff wrong, just implemented on a different layer ?

froobynooby:
I could be reading #9916 wrong, but I think it is adding another setting to let you rate limit behaviours that require certain memories: jobsite, bed location, meeting point, etc. As any number of behaviours can depend on a given memory, this would rate limit a large of number of behaviours.

This PR just restores the pre-1.19.3 functionality of the existing behaviour rate limit settings, which let you rate limit individual behaviours

lynxplay:
Hmmm, possibly a nicer solution then. I mean the other PR just rate limits the try trigger on the pure memory rather than the consuming behaviour.

I'll have a longer look over this once home 👍 the mobile diff viewer isn't particularly great

froobynooby:
Rebased for 1.20.4

PR #9949, opened 2023-11-17
Fix EntityShootBowEvent/ProjectileLaunchEvent consuming arrows when cancelled

### LINKED ISSUE #7702: Cancelling the PlayerReadyArrowEvent causes inventory desync of the arrow item stack
### Expected behavior

When canceling the PlayerReadyArrowEvent the arrow item stack should not be used.

### Observed/Actual behavior

When the event is cancelled the item stack's amount will still decrement until the player clicks on the item / updates the inventory.

### Steps/models to reproduce

1. Cancel the PlayerReadyArrowEvent.
2. Shoot an arrow in survival mode.

### Plugin and Datapack List

Just a single test plugin doing nothing but cancelling the event.

### Paper version

This server is running Paper version git-Paper-277 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 87e11bf)
Previous version: git-Paper-216 (MC: 1.18.1)

### Other

_No response_

### LINKED ISSUE #9919: Canceling archery causes the player's inventory to become out of sync
### Expected behavior

Cancel EntityShootBowEvent or PlayerReadyArrowEvent, the number of arrows in the player's inventory will not be reduced

### Observed/Actual behavior

The number of arrows will decrease until you click on the arrow in the inventory to synchronize

### Steps/models to reproduce

```
    @EventHandler
    public void onEntityShootBow(final @NotNull EntityShootBowEvent e) {
        e.setCancelled(true); // or e.setConsumeItem(false);
    }
```
```
    @EventHandler
    public void onPlayerReadyArrow(final @NotNull PlayerReadyArrowEvent e) {
        e.setCancelled(true);
        // It should not be necessary for plugins to use this method. If it is required for some reason, it is probably a bug.
        // e.getPlayer().updateInventory();
    }
```

### Plugin and Datapack List

Blank

### Paper version

This server is running Paper version git-Paper-280 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 4675152)

### Other

_No response_

### LINKED ISSUE #10665: cancelling bow shoot event still consumes the arrow on 1.20.5
### Expected behavior

shouldn't consume arrow

### Observed/Actual behavior

it consumes it

### Steps/models to reproduce

cancel entity bow shoot event

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version git-Paper-"0db72ac" (MC: 1.20.6) (Implementing API version 1.20.6-R0.1-SNAPSHOT) (Git: 0db72ac)
You are 5 version(s) behind
Previous version: git-Paper-"8f7ac62" (MC: 1.20.5)

### Other

_No response_

### LINKED ISSUE #11113: ProjectileLaunchEvent Bug in Paper 1.20.6: Item Removed on Cancel
### Expected behavior

When the ProjectileLaunchEvent is canceled, the inventory item responsible for the projectile should remain in the player's inventory and not be removed. This behavior should be consistent with previous versions, such as 1.20.4, where the item is not removed if the event is canceled.
Here are videos demonstrating this issue on a 1.20.6 server:

https://github.com/user-attachments/assets/c44196da-b61e-41d0-b15b-96337aafb080

And with the same code on a 1.20.4 server:

https://github.com/user-attachments/assets/c4fa7892-7998-4498-97d5-8bf20f0be4da

### Observed/Actual behavior

In Paper version 1.20.6, the inventory item responsible for the projectile is removed even if the ProjectileLaunchEvent is canceled. Additionally, the item appears to be removed before the event is called, which differs from the expected behavior seen in earlier versions like 1.20.4.

### Steps/models to reproduce

**Steps to Reproduce:**

1. Set up a Paper server running version 1.20.6.
2. Implement the following code to handle the ProjectileLaunchEvent:
 ```java
public class ProjectileLaunchListener implements Listener {
    
    @EventHandler
    public void onProjectileLaunch(ProjectileLaunchEvent event) {
        event.setCancelled(true);
    }
} 
```

3. Cancel the ProjectileLaunchEvent within the code.
4. Launch a projectile and observe the behavior of the inventory item.

### Plugin and Datapack List

**Plugins:**
```
Server Plugins (2):
Paper Plugins:
 - ProjectileTest
Bukkit Plugins:
 - ViaVersion
 ```

**DataPacks:**
```
```

### Paper version

This server is running Paper version 1.20.6-148-ver/1.20.6@20f5165 (2024-07-02T15:37:33Z) (Implementing API version 1.20.6-R0.1-SNAPSHOT)

### Other

_No response_

notTamion:
closes #9919 (inventory desync doesn't seem to be an issue anymore since the arrow just gets consumed even when cancelled)
closes #7702 (same here)
closes #11113 
closes #10665

---
Download the paperclip jar for this pull request: [paper-9949.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2005820922.zip)

### DIFF
--- a/patches/api/0108-EntityShootBowEvent-consumeArrow-and-getArrowItem-AP.patch
+++ b/patches/api/0108-EntityShootBowEvent-consumeArrow-and-getArrowItem-AP.patch
@@ -6,7 +6,7 @@ Subject: [PATCH] EntityShootBowEvent consumeArrow and getArrowItem API
 Adds ability to get what arrow was shot, and control if it should be consumed.
 
 diff --git a/src/main/java/org/bukkit/event/entity/EntityShootBowEvent.java b/src/main/java/org/bukkit/event/entity/EntityShootBowEvent.java
-index 1a8366f6757431baaba4f3d48abea3cf0ec1f1ad..e4efcd757ca51a4edd396f55c812dbe89ffb68c7 100644
+index 1a8366f6757431baaba4f3d48abea3cf0ec1f1ad..1da003ded82f943eff0438126227d1d84599f681 100644
 --- a/src/main/java/org/bukkit/event/entity/EntityShootBowEvent.java
 +++ b/src/main/java/org/bukkit/event/entity/EntityShootBowEvent.java
 @@ -22,7 +22,32 @@ public class EntityShootBowEvent extends EntityEvent implements Cancellable {
@@ -28,12 +28,12 @@ index 1a8366f6757431baaba4f3d48abea3cf0ec1f1ad..e4efcd757ca51a4edd396f55c812dbe8
 +    public ItemStack getArrowItem() {
 +        return this.getConsumable();
 +    }
-+
+ 
 +    @Deprecated
 +    public EntityShootBowEvent(@NotNull final LivingEntity shooter, @Nullable final ItemStack bow, @NotNull final Entity projectile, final float force) {
 +        this(shooter, bow, new ItemStack(org.bukkit.Material.AIR), projectile, force);
 +    }
- 
++
 +    @Deprecated
 +    public EntityShootBowEvent(@NotNull final LivingEntity shooter, @Nullable final ItemStack bow, @NotNull ItemStack arrowItem, @NotNull final Entity projectile, final float force) {
 +        this(shooter, bow, arrowItem, projectile, EquipmentSlot.HAND, force, true);
@@ -42,3 +42,13 @@ index 1a8366f6757431baaba4f3d48abea3cf0ec1f1ad..e4efcd757ca51a4edd396f55c812dbe8
      public EntityShootBowEvent(@NotNull final LivingEntity shooter, @Nullable final ItemStack bow, @Nullable final ItemStack consumable, @NotNull final Entity projectile, @NotNull final EquipmentSlot hand, final float force, final boolean consumeItem) {
          super(shooter);
          this.bow = bow;
+@@ -112,9 +137,7 @@ public class EntityShootBowEvent extends EntityEvent implements Cancellable {
+      * consumed).
+      *
+      * @param consumeItem whether or not to consume the item
+-     * @deprecated not currently functional
+      */
+-    @Deprecated
+     public void setConsumeItem(boolean consumeItem) {
+         this.consumeItem = consumeItem;
+     }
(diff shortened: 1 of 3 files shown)


notTamion:
due to mojang moving logic around the arrow itemstack will now get removed before the event is even called. easiest way to deal with this seems to just give the player the itemstack back. This pr also still fixes to desync happening with PlayerReadyBowEvent and EntityShootBowEvent

patches/api/0108-EntityShootBowEvent-consumeArrow-and-getArrowItem-AP.patch
```diff
@@ -42,3 +42,13 @@ index 1a8366f6757431baaba4f3d48abea3cf0ec1f1ad..e4efcd757ca51a4edd396f55c812dbe8
...
+@@ -112,9 +137,7 @@ public class EntityShootBowEvent extends EntityEvent implements Cancellable {
+      * consumed).
+      *
+      * @param consumeItem whether or not to consume the item
+-     * @deprecated not currently functional
+      */
+-    @Deprecated
+     public void setConsumeItem(boolean consumeItem) {
+         this.consumeItem = consumeItem;
+     }
```
granny, line 54: extra diff

notTamion: what do you mean by extra diff? make it into a new patch?

granny: Disregard; I didn't realize the removal of the deprecation notice & annotation was intentional

notTamion: ah ok

notTamion:
We could either add some checks regarding those enchantments (not sure about the state changes though) or we do some refactoring and remove the item after the event is called. i will take a second look at moving that removal to after the event call later

lynxplay:
Yea I think I may be in favour of removing item post event call. How to track that nicely is going to be a fun challenge. We also need to make sure that nothing there runs that could be affected by not removing items. 

Regarding removal anyway, if the diff ends up with some form of tracking wrapper around Item stack, we may also be able to remove and precisely readd.

notTamion:
Problem: EntityShootBowEvent is called for Crossbows. Is that intended behaivour? If it is what are we supposed to do when the event is cancelled/consumeItem is set to false? just leaving the arrow in the crossbow would basically allow for rapid fire with consumeItem on false

lynxplay:
In 1.20.4, that field was simply ignored for crossbows. So it would always "consume".

Jiangyuluo:
not fully fix.

notTamion:
How is that?

FN-FAL113:
up, broken on 1.20.6

Lulu13022002 (commented):
It seems this patch introduce a new dupe for arrow. Crossbow doesn't consume the arrow when the item is used in survival but the player can still launched the projectile.

And some diff around EntityShootBowEvent#setConsumeItem deprecation has been removed during the rebase.

Jiangyuluo:
This issue still exists in the latest version. The build version provided by 9949 may still become invalid in some cases (which I cannot accurately explain)

lynxplay:
> which I cannot accurately explain
Can you try at least? Or show it via screenshots/video?
If we missed something in the latest build of the PR we'd obviously like to fix that before merging this.

lynxplay:
Resolved your comments lulu, now with a cuter looking entity load crossbow event impl.

Lulu13022002 (commented):
Arrows with the intangible_projectile component launched in survival with a bow are not consumed.

PR #9951, opened 2023-11-18
Finish adding FluidState API

viciscat:
This pull request adds the FluidState API started by MachineMaker in PR #8609.
It adds everything in the PR's todo list. I tested the methods and they returned everything correctly. I don't really know what more I could say. Feel free to ask any questions!


patches/api/0447-Add-FluidState-API.patch
```diff
@@ -0,0 +1,171 @@
...
++    public io.papermc.paper.block.fluid.FluidData getFluidAt(int x, int y, int z);
++
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the given {@link Location}
++     *
++     * @param location The location of the fluid
++     * @return The fluid data at the given location
++     */
++    @NotNull
++    public io.papermc.paper.block.fluid.FluidData getFluidAt(Location location);
```
Sytm: Maybe replace both of these methods with a single one taking a [Position](https://jd.papermc.io/paper/1.20/io/papermc/paper/math/Position.html)?

You then can also use `MCUtil.toBlockPos` to convert the position into a NMS BlockPosition

patches/server/1051-Add-FluidState-API.patch
```diff
@@ -0,0 +1,242 @@
...
++    public PaperFluidData clone() {
++        try {
++            return (PaperFluidData) super.clone();
++        } catch (CloneNotSupportedException ex) {
++            throw new AssertionError("Clone not supported", ex);
++        }
++    }
++
++    @Override
++    public Vector getFlowDirection(World world, Location location) {
```
Sytm: Remove the separate world argument, because a location contains a world

patches/server/1051-Add-FluidState-API.patch
```diff
@@ -0,0 +1,242 @@
...
++        return CraftVector.toBukkit(this.state.getFlow(
++            ((CraftWorld) world).getHandle(),
++            new BlockPos(
++                location.getBlockX(),
++                location.getBlockY(),
++                location.getBlockZ()
++        )));
```
Sytm: Use MCUtil to convert the Location to a blockpos

Sytm (changes requested):
Also where did the methods on the BlockState go that the original PR had?

patches/api/0447-Add-FluidState-API.patch
viciscat: Why not use Location just like all other methods like getBlockAt?

patches/server/1051-Add-FluidState-API.patch
viciscat: Location can not contain a World. Which would mean i would have to return null or make it throw an exception

viciscat:
> Also where did the methods on the BlockState go that the original PR had?
Owen said on [Discord](https://discord.com/channels/289587909051416579/925530366192779286/1135657985205620836) that we probably shouldn't couple liquid with blocks. That's the main reason for the fact it's not there. I could still add it tho. If do add it, should I make it a new patch or edit the existing one?

patches/api/0447-Add-FluidState-API.patch
Sytm: Position is a new API that is used where a world from the location object does not get used, as methods on the World / RegionAccessor discard the world inside the location because it uses itself

patches/server/1051-Add-FluidState-API.patch
Sytm: You could add an not null check using `Preconditions.checkNotNull(location.getWorld())` or something similar, because calling this method like this seems a bit silly:
```java
fluid.getFlowDirection(location.getWorld(), location);
```

Sytm:
> Owen said on [Discord](https://discord.com/channels/289587909051416579/925530366192779286/1135657985205620836) that we probably shouldn't couple liquid with blocks. That's the main reason for the fact it's not there. I could still add it tho. If do add it, should I make it a new patch or edit the existing one?

If Owen said that you can leave it like it currently is, he is more knowledgeable around this stuff after all.

patches/api/0447-Add-FluidState-API.patch
viciscat: Fair enough. I sill think a method that takes x, y, z is always nice to have.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,171 @@
...
++     * @return true if the fluid is water, false otherwise
++     */
++    boolean isWater();
++
++    /**
++     * Returns whether this fluid is lava.
++     *
++     * @return true if the fluid is lava, false otherwise
++     */
++    boolean isLava();
```
Owen1212055: IMO, these should be utilities on Fluid rather than on FluidData

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,171 @@
...
++     */
++    float getHeight();
++
++    /**
++     * Returns whether this fluid is the same fluid as the one passed in the parameter
++     *
++     * @param fluid a fluid
++     * @return true if they are the same, false otherwise
++     */
++    boolean is(Fluid fluid);
```
Owen1212055: This should just be a default implementation of getType() == ...

Or not exist at all.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,171 @@
...
++package io.papermc.paper.block.fluid;
++
++import io.papermc.paper.math.Position;
++import org.bukkit.Fluid;
++import org.bukkit.Location;
++import org.bukkit.World;
++import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
++
++public interface FluidData extends Cloneable {
```
Owen1212055: This is tricky here.

The whole "blockdata" api is based on the idea that most methods can exist as the block is independent.

However fluid is strictly tied to the world due to its inherently unfinished implementation.

I think that we should call this FluidState, to not only align with vanilla but to make a clear distinction between block data here.

Because in truth, FluidState represents a block in the world.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,171 @@
...
++     * @return the fluid type
++     */
++    @NotNull Fluid getFluidType();
++
++    /**
++     * Returns a copy of this FluidData.
++     *
++     * @return a copy of the fluid data
++     */
++    @NotNull FluidData clone();
```
Owen1212055: (This would need to be removed)

patches/server/1047-Add-FluidState-API.patch
```diff
@@ -0,0 +1,239 @@
...
++    }
++
++    /* Registry */
++
++    static {
++        //<editor-fold desc="PaperFluidData Registration" defaultstate="collapsed">
++        register(LavaFluid.Source.class, PaperFallingFluid::new);
++        register(WaterFluid.Source.class, PaperFallingFluid::new);
++        register(LavaFluid.Flowing.class, PaperFlowing::new);
++        register(WaterFluid.Flowing.class, PaperFlowing::new);
```
Owen1212055: Where exactly does this handle air? Kinda confused.

Owen1212055 (changes requested):
In general I think this api is very helpful, but I think we need to move a bit in a different direction here. More specifically, you're sort of mashing the BlockData concept of being "wordless" and BlockStates "world" idea.

Since you're introducing methods that require a world, but also some that don't.

cc @Lulu13022002

patches/api/0446-Add-FluidState-API.patch
viciscat: something along the lines of `Fluid.isWater(Fluid fluid)`?

patches/api/0446-Add-FluidState-API.patch
viciscat: I put it there because it was in FluidState, but i guess i can remove it

patches/server/1047-Add-FluidState-API.patch
viciscat: Didn't do that, this was in MachineMaker's original PR and to be honest I have no idea how it works. If i get the FluidData of an air block, and do getFluidType() it returns Fluid.EMPTY so uh it works i guess?

patches/api/0446-Add-FluidState-API.patch
viciscat: This was the original name in MachineMaker's PR. And I guess it makes sense, but shouldn't it be named differently to not confuse it with the minecraft one?

viciscat:
> you're sort of mashing the BlockData concept of being "wordless" and BlockStates "world" idea.
Since you're introducing methods that require a world, but also some that don't.

So should I do methods that require a world or not? I would say they don't require a world since you get the FluidData from the World.

patches/api/0446-Add-FluidState-API.patch
Owen1212055: Yes, (assuming not static here)

patches/api/0446-Add-FluidState-API.patch
Owen1212055: Well…. We aren’t going to confuse it with the Minecraft name. Mayching MC names is ideal.

patches/server/1047-Add-FluidState-API.patch
Owen1212055: Okay cool

patches/api/0446-Add-FluidState-API.patch
Machine-Maker: I do not agree. This isn't the equivalent of a bukkit BlockState, which is what it will be confused with if we call it FluidState. It's the equivalent of an nms BlockState which the api calls BlockData. 

nms FluidState does not represent a block in the world, all the methods on it require a Level.

patches/server/1047-Add-FluidState-API.patch
Machine-Maker: I didn't include empty here because it's handled below in the getOrDefault call on the map when creating an instance. the empty just uses the regular PaperFluidData type.

patches/api/0446-Add-FluidState-API.patch
Owen1212055: Okay, that's fair. discard.

patches/api/0446-Add-FluidState-API.patch
Owen1212055: Do note here that this may be better to be covered FluidTag in future, but this would be future stuff, as the tag API won't work nicely with fluids since it's an enum currently.

patches/api/0446-Add-FluidState-API.patch
Machine-Maker: I think its fair to have a copy/clone/whatever method to create a duplicate of it, since it doesn't have any world state, copying a fluiddata and applying it to some other location is valid.

patches/api/0446-Add-FluidState-API.patch
viciscat: Is this good or needlessly excessive ?
<details>
  <summary>Code</summary>
  
```java
public enum Fluid implements Keyed {
   // ...
    WATER{
        @Override
        public boolean isWater() {
            return true;
        }
    },
    /**
     * Flowing water.
     */
    FLOWING_WATER{
        @Override
        public boolean isWater() {
            return true;
        }
    },
    /**
     * Stationary lava.
     */
    LAVA{
        @Override
        public boolean isLava() {
            return true;
        }
    },
    // ...
}
// ...
public boolean isWater() {
        return false;
    }
    
    public boolean isLava() {
        return false;
    }
```
</details>

viciscat:
(Continuing on my previous comment)
The thing is that some methods in `FluidState` just don't need a Level and position like `getAmount` and `getHeight` since it can just read it from the block properties (you know, the data you can see when looking at a fluid with F3). `getFlowDirection` really is just the only one that needs a position and level

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,201 @@
...
+     }
++    // Paper start
++    public boolean isWater() {
++        return false;
++    }
++
++    public boolean isLava() {
++        return false;
++    }
++    // Paper end
```
Machine-Maker: These are incorrect. I don't think *any* isWhatever methods should be added. These are already set by tags and changeable by datapacks. Those tags are already exposed to the API, I don't think we need an alternative way to do it.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,201 @@
...
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the specified position.
++     *
++     * @param x The x-coordinate of the position
++     * @param y The y-coordinate of the position
++     * @param z The z-coordinate of the position
++     * @return The {@link io.papermc.paper.block.fluid.FluidData} at the specified position
++     */
++    @NotNull
++    public io.papermc.paper.block.fluid.FluidData getFluidAt(int x, int y, int z);
```
Machine-Maker:
```diff
-    public io.papermc.paper.block.fluid.FluidData getFluidAt(int x, int y, int z);
+    public io.papermc.paper.block.fluid.FluidData getFluidData(int x, int y, int z);
```

To be consistent with getBlockData method names

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,201 @@
...
++
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the given position
++     *
++     * @param position The location of the fluid
++     * @return The fluid data at the given location
++     */
++    @NotNull
++    public io.papermc.paper.block.fluid.FluidData getFluidAt(io.papermc.paper.math.Position position);
```
Machine-Maker:
```diff
-    public io.papermc.paper.block.fluid.FluidData getFluidAt(io.papermc.paper.math.Position position);
+    public io.papermc.paper.block.fluid.FluidData getFluidData(io.papermc.paper.math.Position position);
```

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,201 @@
...
++     */
++    float getHeight();
++
++    /**
++     * Returns whether this fluid is the same fluid as the one passed in the parameter
++     *
++     * @param fluid a fluid
++     * @return true if they are the same, false otherwise
++     */
++    default boolean is(Fluid fluid) {
```
Machine-Maker: I don't see the purpose of this event. We already have a getFluidType which does this check, not really needed and we don't have similar methods on other similar types.

patches/server/1047-Add-FluidState-API.patch
```diff
@@ -0,0 +1,222 @@
...
++}
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java b/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
+index 1a17875426468b287c8ea3f559ea516d0218f7a0..0bd63e98501fcfbfddadda9bd5b7e9340dbf3950 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
+@@ -202,6 +202,7 @@ import org.bukkit.inventory.ItemStack;
+ import org.bukkit.potion.PotionData;
+ import org.bukkit.potion.PotionType;
+ import org.bukkit.util.Vector;
++import org.jetbrains.annotations.NotNull;
```
Machine-Maker: Inline the import (or just don't use it at all, not needed)

Machine-Maker (changes requested):

patches/api/0446-Add-FluidState-API.patch
Machine-Maker: Should be removed as I mentioned in my own review

patches/api/0446-Add-FluidState-API.patch
viciscat: Ok! Will be removed

patches/server/1047-Add-FluidState-API.patch
viciscat: Oh yeah oops forgot about fully qualified imports

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,147 @@
...
++}
+diff --git a/src/main/java/org/bukkit/RegionAccessor.java b/src/main/java/org/bukkit/RegionAccessor.java
+index cbb51dde78f792db4ddac43f144d23ff4b12d25f..938737b2f7b343fec814f6c9d5386b2161cf25f3 100644
+--- a/src/main/java/org/bukkit/RegionAccessor.java
++++ b/src/main/java/org/bukkit/RegionAccessor.java
+@@ -102,6 +102,28 @@ public interface RegionAccessor extends Keyed { // Paper
+      */
+     @NotNull
+     BlockState getBlockState(int x, int y, int z);
++    // Paper Start - FluidState API
```
Lulu13022002: The second part of the paper comments should be in lowercase (start/end), same for other occurences. The next newline doesn't need to be in comment and can be left before.

patches/server/1047-Add-FluidState-API.patch
```diff
@@ -0,0 +1,214 @@
...
++    /* Registry */
++
++    static {
++        //<editor-fold desc="PaperFluidData Registration" defaultstate="collapsed">
++        register(LavaFluid.Source.class, PaperFallingFluid::new);
++        register(WaterFluid.Source.class, PaperFallingFluid::new);
++        register(LavaFluid.Flowing.class, PaperFlowing::new);
++        register(WaterFluid.Flowing.class, PaperFlowing::new);
++        //</editor-fold>
++    }
```
Lulu13022002: This part is over engineered and can be simplified by simple conditions (isEmpty = PaperFluidData, isSource = PaperFallingFluid, else PaperFlowing)

patches/server/1047-Add-FluidState-API.patch
```diff
@@ -0,0 +1,214 @@
...
+         return CraftBlock.at(this.getHandle(), new BlockPos(x, y, z)).getState();
+     }
++    // Paper Start - FluidState API
++    @Override
++    public io.papermc.paper.block.fluid.FluidData getFluidData(final int x, final int y, final int z) {
++        return io.papermc.paper.block.fluid.PaperFluidData.createData(getHandle().getFluidState(new BlockPos(x, y, z)));
++    }
++
++    @Override
++    public io.papermc.paper.block.fluid.FluidData getFluidData(final io.papermc.paper.math.Position position) {
```
Lulu13022002: This can be a default method

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,147 @@
...
++     * Returns the flowing direction the liquid as a vector
++     *
++     * @return the flow direction vector
++     */
++    Vector getFlowDirection(Location location);
++
++    /**
++     * Returns the amount of liquid is in the fluid, as in the number of layers.
++     *
++     * @return the amount as an integer, between 0 and 8
```
Lulu13022002: If you know the range you can use the Range annotation

patches/server/1047-Add-FluidState-API.patch
```diff
@@ -0,0 +1,214 @@
...
++        try {
++            return (PaperFluidData) super.clone();
++        } catch (CloneNotSupportedException ex) {
++            throw new AssertionError("Clone not supported", ex);
++        }
++    }
++
++    @Override
++    public Vector getFlowDirection(Location location) {
++        Preconditions.checkNotNull(location.getWorld());
```
Lulu13022002: I think checkArgument would be better here with a proper message

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,147 @@
...
++     * @return The {@link io.papermc.paper.block.fluid.FluidData} at the specified position
++     */
++    @NotNull
++    io.papermc.paper.block.fluid.FluidData getFluidData(int x, int y, int z);
++
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the given position
++     *
++     * @param position The location of the fluid
++     * @return The fluid data at the given location
```
Lulu13022002: This method now takes a position not a location

patches/server/1047-Add-FluidState-API.patch
Machine-Maker: The reason I had it this way in my original PR was so it was similar to CraftBlockData. We can simplify it, but there is something to be said for setting it up the same way.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,147 @@
...
++     * @return a copy of the fluid data
++     */
++    @NotNull FluidData clone();
++
++    /**
++     * Returns the flowing direction the liquid as a vector
++     *
++     * @return the flow direction vector
++     */
++    Vector getFlowDirection(Location location);
```
Lulu13022002: The parameter is not documented

patches/server/1047-Add-FluidState-API.patch
viciscat: I will not touch it because I don't wanna break anything

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,153 @@
...
++     */
++    @Range(from = 0, to = 8)
++    int getAmount();
++
++    /**
++     * Returns the height of the fluid in blocks
++     *
++     * @return the height as a float value, representing the height in blocks
++     */
++    float getHeight();
```
Lulu13022002: I'm not sure about the use case of this method, people just have to divide the amount by 9 instead? Otherwise the doc should be clear that this is the height on only one block and not in block**s** that might be confusing.

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,153 @@
...
++package io.papermc.paper.block.fluid.type;
++
++public interface Flowing extends FallingFluid {
++
++    /**
++     * Get the level of the flowing liquid.
++     *
++     * @return the level
++     */
++    int getLevel();
```
Lulu13022002: This is kinda a duplicate with getAmount() ?

patches/api/0446-Add-FluidState-API.patch
```diff
@@ -0,0 +1,153 @@
...
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the given position
++     *
++     * @param position The location of the fluid
++     * @return The fluid data at the given position
++     */
++    @NotNull
++    default io.papermc.paper.block.fluid.FluidData getFluidData(io.papermc.paper.math.Position position) {
++        return getFluidData(position.blockX(), position.blockY(), position.blockZ());
++    };
```
Lulu13022002: This semicolon is not needed

patches/api/0446-Add-FluidState-API.patch
Lulu13022002: This is not completely resolved the parameter still say location

patches/api/0449-Add-FluidState-API.patch
```diff
@@ -0,0 +1,146 @@
...
++     */
++    @NotNull FluidData clone();
++
++    /**
++     * Returns the direction of the flow of the liquid at the given location as a vector.
++     *
++     * @param location - the location to check the liquid flow
++     * @return the flow direction vector at the given location
++     */
++    Vector getFlowDirection(Location location);
```
Machine-Maker: Missing annotation on the return type **and** parameter

patches/api/0449-Add-FluidState-API.patch
```diff
@@ -0,0 +1,146 @@
...
+diff --git a/src/main/java/org/bukkit/RegionAccessor.java b/src/main/java/org/bukkit/RegionAccessor.java
+index 43dd6c59cceba12f27e6b265acc3ad97eea37abd..9be3d603908f22514a507887d5150fe487e0a186 100644
+--- a/src/main/java/org/bukkit/RegionAccessor.java
++++ b/src/main/java/org/bukkit/RegionAccessor.java
+@@ -102,6 +102,31 @@ public interface RegionAccessor extends Keyed { // Paper
+     @NotNull
+     BlockState getBlockState(int x, int y, int z);
+ 
++    // Paper start - FluidState API
++
```
Machine-Maker: Extra newline

patches/api/0449-Add-FluidState-API.patch
```diff
@@ -0,0 +1,146 @@
...
++    io.papermc.paper.block.fluid.FluidData getFluidData(int x, int y, int z);
++
++    /**
++     * Gets the {@link io.papermc.paper.block.fluid.FluidData} at the given position
++     *
++     * @param position The location of the fluid
++     * @return The fluid data at the given position
++     */
++    @NotNull
++    default io.papermc.paper.block.fluid.FluidData getFluidData(io.papermc.paper.math.Position position) {
```
Machine-Maker: Missing annotation on the parameter

Machine-Maker (changes requested):

Machine-Maker (approved):

Owen1212055 (approved):

PR #9964, opened 2023-11-23
Un-experimentalize Entity TP APIs

Machine-Maker:
The Relative Teleport/Look At APIs has been mature for a while now with no breaking changes needed. I think we can safely remove the `Experimental` annotation from them. I don't want to just end up never removing these annotations.

### DIFF
--- a/patches/api/0370-More-Teleport-API.patch
+++ b/patches/api/0370-More-Teleport-API.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] More Teleport API
 
 diff --git a/src/main/java/io/papermc/paper/entity/LookAnchor.java b/src/main/java/io/papermc/paper/entity/LookAnchor.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..c8312691c27ae436029ec5011ddf073582b12cba
+index 0000000000000000000000000000000000000000..544eec787ea837f7d29df6519255840d6fe087d7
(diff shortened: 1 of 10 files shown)


Owen1212055 (approved):

emilyy-dev:
Would async variants of the new teleport methods be added in a separate PR? Since this is just taking them out of the experimental phase

Machine-Maker:
I think those can be added whenever really. I don't think those teleport async methods have any server implementation, it's just loading the chunk async and then teleporting within the future that is returned.

imDaniX:
Might be an annoying question, but what about Location? The whole Position API seems to be immutable, but Location, which implements FinePosition, just isn't. I feel like there might be a problem when a plugin has positions stored that change over time. Also, Location already has *a bunch* of mutating offsetting methods, and now it has more, but without a single mark that those are creating an immutable Position that isn't a Location.
I'd say either Location's and Positions' javadoc should describe the mutability status, or there being some mechanism to determine whether or not a Position is mutable *without* explicitly checking if a Position is an instance of Location.

Machine-Maker:
Ok yeah, there has been some more discussion about Position on the discord. Possibly we can just remove FinePosition from being implemented by Location and just have an asPosition method. Either way, removed that from this patch, now its just the teleport API.

kennytv (approved):

PR #9968, opened 2023-11-25
Configurable Dry and Wet Farmland Tick Rates

MrPowerGamerBR:
If your server has a lot of farms, the `isNearWater` calls gets *very* expensive.

To workaround this, we can throttle the tick rate of farmlands, because after all, you don't *really* need to check if the farmland is near a source of water that often, right? It is not like the water will magically vanish if they are in a player made farm.

This helps a lot, on my server farm lands were taking up ~8% of CPU time, so throttling the tick rate of wet farmlands helped to bring it down while not having a negative gameplay impact.

There are two options: `dry-farmland` and `wet-farmland`, the reason it is separate is because the performance impact of dry farmlands isn't that big, but has a big gameplay impact where players may be confused about why the farmland isn't wet after placing water near it. Yes, dry farmlands have the same `isNearWater` checks as wet farmlands, but dry farmlands do end up turning to regular dirt after a while, so it is like they are "naturally" throttled by the server.

The code for the tick rate throttle was copied from the "Configurable Grass Spread Tick Rate" patch.

(The reason the name is "Farmland" and not "Farm Land" is because that's what the game calls the block)


Machine-Maker (commented):
It's almost worth creating a new wrapper type for tick rates cause now we have like 5 of those big ugly long if statements that are identical (FarmBlock and SpreadingSnowyDirtBlock)

MrPowerGamerBR:
> It's almost worth creating a new wrapper type for tick rates cause now we have like 5 of those big ugly long if statements that are identical (FarmBlock and SpreadingSnowyDirtBlock)

If it is easier, maybe creating a simple static function `shouldSkip(int tickRate, int hashCode)` could also help clean up those pesky if statements without requiring too much work.

Machine-Maker:
> If it is easier, maybe creating a simple static function shouldSkip(int tickRate, int hashCode) could also help clean up those pesky if statements without requiring too much work.

It's probably fine for now. I will probably end up changing it to something custom wrapper type during one of my periodic config cleanup PRs.

Machine-Maker (approved):
Looks good.

This pr adds the 2nd and 3rd blockstates which have their own tick rates (along with mob spawners). Depending on how many more are useful to add, might be nice to have some sort of way to set the random tick rate of any blockstate, but that can happen later if needed (and if its a good idea at all)

lynxplay (approved):

PR #9969, opened 2023-11-25
Lazily create LootContext for criterions

MrPowerGamerBR:
For each player on each tick, enter block triggers are invoked, and these create loot contexts that are promptly thrown away since a lot of the times the predicate doesn't match. (in vanilla, it should only pass for the end gateway, or when the player is in water to get the boat's recipes, and iirc none of these requires a LootContext)

To avoid this, we now lazily create the LootContext if the criterion passes the predicate AND if any of the listener triggers require a loot context instance.

While the performance increase ain't that big (on a random profile with ~55 players on  my server, the `createContext` call was using 0.10% according to spark), it would be nice to just *avoid* that useless allocation anyway, and theorically if you have a lot of players moving around, the performance impact of this call would be bigger.

### DIFF
--- a/patches/server/1053-Lazily-create-LootContext-for-criterions.patch
+++ b/patches/server/1053-Lazily-create-LootContext-for-criterions.patch
@@ -1,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MrPowerGamerBR <git@mrpowergamerbr.com>
+Date: Tue, 21 Nov 2023 12:16:39 -0300
+Subject: [PATCH] Lazily create LootContext for criterions
+
+For each player on each tick, enter block triggers are invoked, and these create loot contexts that are promptly thrown away since the trigger doesn't pass the predicate
+
+To avoid this, we now lazily create the LootContext if the criterion passes the predicate AND if any of the listener triggers require a loot context instance
+
+diff --git a/src/main/java/net/minecraft/advancements/critereon/SimpleCriterionTrigger.java b/src/main/java/net/minecraft/advancements/critereon/SimpleCriterionTrigger.java
+index f0367a9cce13ef576fbb7023c0aba6eb48963606..bbdbf62a7696640d92a6bb758052abaaa6f86edd 100644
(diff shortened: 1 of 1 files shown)


patches/server/1053-Lazily-create-LootContext-for-criterions.patch
```diff
@@ -0,0 +1,33 @@
...
+             for(CriterionTrigger.Listener<T> listener : set) {
+                 T simpleInstance = listener.trigger();
+                 if (predicate.test(simpleInstance)) {
+                     Optional<ContextAwarePredicate> optional = simpleInstance.playerPredicate();
++                    // Paper start - lazily create LootContext for criterions
++                    if (lootContext == null && optional.isPresent()) {
++                        lootContext = EntityPredicate.createContext(player, player);
++                    }
++                    // Paper end
+                     if (optional.isEmpty() || optional.get().matches(lootContext)) {
```
Machine-Maker: I'm not a super huge fan of the way this logic is structured. To be clear, it seems to work, but the `optional.isPresent()` check on the earlier if statement to counter the `optional.isEmpty()` check below might not be the most clear way to do this.

As much as I don't like using the return value of variable assignment, perhaps this is more clear?
```diff
- if (optional.isEmpty() || optional.get().matches(lootContext)) {
+ if (optional.isEmpty() || optional.get().matches(lootContext = (lootContext == null ? EntityPredicate.createContext(player, player) : lootContext))) {
```

Just a suggestion, if we like what is currently here, prob just add a comment on the isPresent check about it.

MrPowerGamerBR: Here's the patch with your suggestions, if other people also agree that this is better, I will commit it directly to the branch :3 

[1053-Lazily-create-LootContext-for-criterions.patch](https://github.com/PaperMC/Paper/files/13464884/1053-Lazily-create-LootContext-for-criterions.patch)

If it would be better to be more clear to read, then maybe a `boolean requiresLootContext = optional.isPresent()` could be better tho, since it makes it clear that the `optional` predicate is how we determine if we need the loot context. But sadly, that would be a larger diff.

Machine-Maker (approved):

Owen1212055 (approved):

PR #9970, opened 2023-11-25
Optimize block entity removal by using a ArrayList with bulk remove by index

MrPowerGamerBR:
The `blockEntityTickers` list was replaced with a custom list based on fastutil's `ObjectArrayList` with a small yet huge change for us: A method that allows us to remove a list of indexes from the list.

This is WAY FASTER than using `removeAll` with a list of entries to be removed, because we don't need to calculate the identity of each block entity to be removed, and we can jump directly to where the search should begin, giving a performance boost for small removals (because we don't need to loop thru the entire list to find what element should be removed, and we can quit after removing all the indexes) and a performance boost for big removals (no need to calculate the identity of each block entity), ESPECIALLY if the server has a lot of block entities, or if you are unloading a lot of block entities.

While I don't have nice and fancy and fluffy benchmarks on hand, here's some results from a quick and dirty benchmark test...
* If you have 10k block entities and attempt to remove the first block entity...
   * Current implementation: ~6333.6 op/s
   * New implementation: ~33745.2 op/s
* If you have 10k block entities and attempt to remove the last block entity...
   * Current implementation: ~6726.4 op/s
   * New implementation: ~34487.8 op/s
* If you have 10k block entities and attempt to remove block entities from 900 to 915 (simulating what happens in a chunk unload)
   * Current implementation: ~5073.6 op/s
   * New implementation: ~31476.6 op/s
* If you have 10k block entities and attempt to remove block entities from it % 2 == 0 (this is like, the worst case scenario for us)
   * Current implementation: ~1372.2 op/s
   * New implementation: ~2895.6 op/s

**Notes:**
* The patch is a bit messy, but I can clean up after the idea is "approved".
* Yes, this is faster than if you manually used `remove(int)` for each block entity, since that would create a arraycopy for each removal, and that makes us have [the big sad](https://pbs.twimg.com/media/EAmr-PAWsAEoiWR.jpg). (see patch below, that's how Minecraft originally works and Paper replaced it with a better implementation)
* I'm not 100% sure if the `removeAllByIndex` is good enough: I have tested it a lot and run the server with both implementations (old and the new) running with `equals` checking both of them to see if the order of both lists matched, and it seems to be fine.
* The `removeAllByIndex` is somewhat based on `removeElements` from fastutil, originally I was attempting to "remove block entities by range" by using `removeElements` but that wasn't good if the ranges were scattered across the list, which can happen if multiple chunks with low block entity counts unload.
* ~~*Maybe* this patch should be removed beforehand? Or maybe that patch should be edited? They both edit the same section attempting to improve removal performance. https://github.com/PaperMC/Paper/blob/master/patches/server/0164-Fix-MC-117075-TE-Unload-Lag-Spike.patch~~
  * This patch now replaces the old patch
* I think that the list doesn't need to be named `BlockEntityTickersList`, considering that it doesn't have any specific block entity behavior, however I don't know a good name for it. `ObjectArrayListWithBulkRemovalByIndex` seems wonky.
* A super hacky patch is here if you want to compare the performance between the old behavior and the new behavior, it logs to the console what implementation was fastest to remove, requires some changes to make it work on Paper tho: https://github.com/SparklyPower/SparklyPaper/blob/b96d6f8f3895d944b3c6f52fb4e0f8d706ba3f3a/patches/removed/server/0019-BlockEntityTickersList-optimization-tests.patch#L1


Machine-Maker (changes requested):
This patch should just completely replace `Fix MC-117075: TE Unload Lag Spike`. That is the patch that changes the removal logic from vanilla's removal 1 at a time to using the removeAll. No point in having a patch that is completely reverted in a later patch.

MrPowerGamerBR:
> This patch should just completely replace `Fix MC-117075: TE Unload Lag Spike`. That is the patch that changes the removal logic from vanilla's removal 1 at a time to using the removeAll. No point in having a patch that is completely reverted in a later patch.

I've now replaced the old patch with the new code :3

I also changed the code a bit: Now the list itself tracks what indexes and where to start the search, instead of throwing the code in the `tickBlockEntities()` method, which makes the `Level` diff a bit cleaner. The `tickBlockEntities()` marks indexes to be removed with `markAsRemoved(int index)` and, after all block entities are ticked, `removeMarkedEntries()` cleans up the "marked as removed" entries.

Machine-Maker (approved):
Approving as the concept and implementation seem fine. I don't know if someone else like @Spottedleaf might have a different preferred way to remove the tickers. 

I also think the final product should use better param/local names as single letters here can just be super confusing.

MrPowerGamerBR:
> I don't know if someone else like Spottedleaf might have a different preferred way to remove the tickers.

Another way that I thought that it could be optimized even further beyond is if you `System.arraycopy` instead of `a[j++] = a[i];` for each element that is not in the list. This way, indexes that are far apart (example: index 100 and 8000) could avoid the looping thru all the elements just to get from index 100 to 8000.

So, using our example: The search starts at index 100, then the elements from 101 to 7999 are `System.arraycopy`'d to index 100, so on and so forth.

However this is not bullet proof, since for our worst case (if we are removing block entities with index `it % 2 == 0`) the performance drops drastically, way more than the current ArrayList implementation. Of course, in a real life scenario things something like that would never happen, so *maybe* it is worth a shot later to experiment if we could make it even faster.

About the variables: Yeah they really need better names lmao, those short variables names are actually from fastutil's `removeElements` method, since I was modifying that method to do what I wanted.

MrPowerGamerBR:
Updated the patch to change the variable names to be more descriptive and easy to understand

patches/server/0164-Fix-MC-117075-TE-Unload-Lag-Spike.patch
```diff
@@ -1,42 +1,158 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: mezz <tehgeek@gmail.com>
-Date: Wed, 9 Aug 2017 17:51:22 -0500
+From: MrPowerGamerBR <git@mrpowergamerbr.com>
+Date: Sun, 26 Nov 2023 13:02:16 -0300
 Subject: [PATCH] Fix MC-117075: TE Unload Lag Spike
```
kennytv: Not related to the changes, but the patch should be renamed to say block entities instead of TE/tile entities

patches/server/0164-Fix-MC-117075-TE-Unload-Lag-Spike.patch
```diff
@@ -1,42 +1,158 @@
...
++        super(c);
++    }
++
++    /**
++     * Marks an entry as removed
++     *
++     * @param index the index of the item on the list to be marked as removed
++     */
++    public void markAsRemoved(final int index) {
++        // The block entities list always loop starting from 0, so we only need to check if the startSearchFromIndex is -1 and that's it
```
kennytv: While true, I don't like assuming this over code that can change or be extended at any time and possibly go by unnoticed, and adding the extra check isn't expensive, same for a bounds check

patches/server/0164-Fix-MC-117075-TE-Unload-Lag-Spike.patch
```diff
@@ -1,42 +1,158 @@
...
++        this.startSearchFromIndex = -1; // Reset the start search index
++    }
++
++    /**
++     * Removes elements by their index.
++     *
++     * @param startSearchFromIndex from where the search should start
++     * @param indexesToBeRemoved list of indexes to be removed
++     */
++    private void removeAllByIndex(final int startSearchFromIndex, final IntOpenHashSet indexesToBeRemoved) { // can't use Set<Integer> because we want to avoid autoboxing when using contains
```
kennytv: IntSet

kennytv (dismissed):

lynxplay:
Looking over this patch, I wonder if a reverse approach would be faster.
Given we have a sorted list of indices to remove from the list we would be able to do determine a set of "ranges" of elements to be kept.

E.g. with `toRemove = [0, 3, 5]` and a list of `[a, b, c, d, e, f, g]` we can immediately conclude that items `1` and `2` have to be moved by one index to the left, leading to `[b, c, c, d, e, f, g]`. We can also conclude item `4` has to be moved to the left by two indices, leading to `[b, c, d, d, e, f, g]`. Further more we know that item `6` has to be moved by three indices leaving us with a final array of `[b, c, d,  g, e, f, g]`. We removed three indices so we array copy the last three to null and we are done.

This way we a) can avoid a set and potential collisions. b) have a terribly easy setup for System.arraycopy optimisation of larger moves down the line (e.g. interval [1-2] has two items, if that is above threshold `n` we use arraycopy, otherwise we use a plain for-loop to shift by whatever diff, in the ranges case one.

I might be missing something, but I feel like this would potentially benefit the data structure.

MrPowerGamerBR:
Paper 1.20.4 now uses a `ReferenceOpenHashSet` instead of a `ObjectOpenCustomHashSet` for the `toRemove` list, I still need to benchmark if this patch is still needed *but* after a brief smol benchmark that I've made, I think this patch still provides a performance improvement.

MrPowerGamerBR:
I still need to update this PR to rebase it to 1.20.6, but here's the same patch on my own fork: https://github.com/SparklyPower/SparklyPaper/blob/ver/1.20.6/patches/server/0015-Fix-MC-117075-TE-Unload-Lag-Spike.patch

lynxplay:
I'll implement my suggestion too and run benchmarks over night :+1:

Issue #9458, opened 2023-07-07
Priority of MC-Command is overridden by the same Paper-Command


Leahcimkrob:
### Expected behavior

When I run the /locate command, I want to know which biome or structure is nearby (MC-CMD /locate)

### Observed/Actual behavior

With /locate the MC command is not executed as it was up to 1.19.4, but since 1.20 and 1.20.1 the command worldguard:loacte

### Steps/models to reproduce

Install paper and Worldguard an you can reproduce it

### Plugin and Datapack List

[19:25:43 INFO]: PM | Plugins (20):
[19:25:43 INFO]:  - AutoRestart 4.2.1
[19:25:43 INFO]:  - ChatFree 1.2
[19:25:43 INFO]:  - ChestCommands 4.0.4
[19:25:43 INFO]:  - Citizens 2.0.32-SNAPSHOT (build 3135)
[19:25:43 INFO]:  - CMI 9.6.1.0
[19:25:43 INFO]:  - CMILib 1.3.1.2
[19:25:43 INFO]:  - FastAsyncWorldEdit 2.6.5-SNAPSHOT-511;a669be2
[19:25:43 INFO]:  - HeadDatabase 4.17.3
[19:25:43 INFO]:  - HolographicDisplays 3.0.3-SNAPSHOT-b257
[19:25:43 INFO]:  - ImageMaps 1.0.7
[19:25:43 INFO]:  - LuckPerms 5.4.89
[19:25:43 INFO]:  - Multiverse-Core 4.3.2-SNAPSHOT-b870
[19:25:43 INFO]:  - PacketsBookFix 1.0
[19:25:43 INFO]:  - PlaceholderAPI 2.11.3
[19:25:43 INFO]:  - PluginManager 2.8.1
[19:25:43 INFO]:  - ProtocolLib 5.0.0-SNAPSHOT-b618
[19:25:43 INFO]:  - SignColorPlus 1.2
[19:25:43 INFO]:  - TAB-Bridge 3.1.0
[19:25:43 INFO]:  - Vault 1.7.3-CMI
[19:25:43 INFO]:  - WorldGuard 7.0.9-SNAPSHOT+2241-c0cbe27

### Paper version

[19:26:18 INFO]: This server is running Paper version git-Paper-66 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT)

### Other

After updating from 1.19.4 to 1.20.1 the /locate cmd just converted to locating players using compass
[...]

PR #9971, opened 2023-11-25
Add config file for better command aliases

### LINKED ISSUE #9214: Aliases for commands defined in commands.yml don't properly TAB complete
### Expected behavior

I'm making this issue since #1733 has long since gone stale, and with the recent changes to paper plugins and brigadier-backed plugin commands, I think it's relevant again.

Since the addition of TAB completion in Minecraft 1.13, the commands.yml aliases system hasn't been updated to properly respect the TAB completion providers of commands.

A side effect of paper plugins being introduced is that the load order of plugins has changed, in part as developers seek to close the loops of circular dependencies. In turn, this means that the order commands are registered has changed between plugins. The official solution to this problem, aside from developers implementing configuration options to outright disable commands, is to define overriding aliases in `commands.yml`, redirecting the standard command to the namespace-backed fallback version of a command for the plugin you wish to take precedence.

For instance, say two plugins define a /home command: Plugin1 and Plugin2. Plugin2 happens to register its command before Plugin1, but the user would like it so that Plugin1's home command is the one effectively used when the user types /home. 

So, the user defines an alias in commands.yml ([relevant bukkit docs](https://bukkit.fandom.com/wiki/Commands.yml)):
```yaml
aliases:
  home:
    - "plugin1:home $1-"
```
This alias tells the server that when /home is executed, to instead execute /plugin1:home with all the arguments passed, starting with the first argument (this system is 1-indexed). Executing the alias with any amount of arguments now correctly executes plugin1's home command.

### Observed/Actual behavior

The problem, however, is now when the user attempts to TAB complete when using an alias.

Instead of displaying the TAB suggestions provided by plugin1's home command, the suggestion provider instead always returns the list of every online player as the suggested input for every argument.

The same is true if you define an alias that forwards a different number or range of passed command arguments; the online player list is always returned instead of the correct TAB completion.

### Steps/models to reproduce

* Install a plugin that provides TAB suggestions for a command
* Define an alias for it in commands.yml that passes all arguments to the command
* Enter the /installedplugin:command that you defined an alias for; observe TAB suggestions working normally
* Observe that when you enter the alias you defined, instead of the target command's TAB suggestions being shown as before, the local player list is instead displayed for every possible argument, for some reason.

### Plugin and Datapack List

n/a &mdash; Though for testing with a command override, you could try HuskHomes and Essentials, which both provide a /home command, among others.

### Paper version

1.19.4-534

### Other

I'm not familiar with the (presumably, mostly ancient CraftBukkit) code that handles this system, but I suspect given the fact the alias system supports logic such as argument ranges, an implementation of this system would be fairly complex to pull off, particularly as paper commands don't necessarily handle TAB completion based on the context of how many arguments have been passed&mdash;or use something else to handle this, such as Lucko's Commodore library.

The solution I'd like to see is, if a user is TAB completing with an aliased command, that Paper would _check if their current input is within the range of the supported aliases_, and, if so, grabs the TAB suggestions from the aliases' target command TAB suggestion provider (if there is one); otherwise does not suggest any arguments; and _not_ the list of online players.

Machine-Maker:
Adds a new file, `command-aliases.yml` into our config folder for defining simple aliases for commands that **work** with tab completion and permissions. Schema is as follows
```yaml
alias1: some-command

alias2:
  - some
  - command

alias3:
  target: some-command
  permission: some.permission

alias4:
  target:
    - some
    - command
  permission: some.permission
```

The default permission behavior is to combine all of the permission requirements at all of the nodes along the target path and assign that to the alias node. But that can be changed to a custom permission in the config.

----

Closes https://github.com/PaperMC/Paper/issues/9214
Related https://github.com/PaperMC/Paper/issues/9458

----

Will be merged into the paper config patch just before merging.

### DIFF
--- a/patches/server/1053-Add-config-file-for-better-command-aliases.patch
+++ b/patches/server/1053-Add-config-file-for-better-command-aliases.patch
@@ -1,0 +1,395 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 25 Nov 2023 12:28:05 -0800
+Subject: [PATCH] Add config file for better command aliases
+
+
+diff --git a/src/main/java/io/papermc/paper/command/subcommands/ReloadCommand.java b/src/main/java/io/papermc/paper/command/subcommands/ReloadCommand.java
+index bd68139ae635f2ad7ec8e7a21e0056a139c4c62e..06d955772a66d29cda5cf486421d975d12859889 100644
+--- a/src/main/java/io/papermc/paper/command/subcommands/ReloadCommand.java
++++ b/src/main/java/io/papermc/paper/command/subcommands/ReloadCommand.java
+@@ -26,6 +26,7 @@ public final class ReloadCommand implements PaperSubcommand {
+ 
+         MinecraftServer server = ((CraftServer) sender.getServer()).getServer();
+         server.paperConfigurations.reloadConfigs(server);
++        server.paperConfigurations.getAliasesConfig().reloadAliases(server.getCommands().getDispatcher(), ((CraftServer) sender.getServer()).getCommandMap());
+         server.server.reloadCount++;
+ 
+         Command.broadcastCommandMessage(sender, text("Paper config reload complete.", GREEN));
+diff --git a/src/main/java/io/papermc/paper/configuration/ConfigurationLoaders.java b/src/main/java/io/papermc/paper/configuration/ConfigurationLoaders.java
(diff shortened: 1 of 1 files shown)


YanisBft:
Great work, will be very helpful! Do you plan on deprecating Bukkit command aliases? Or still support it officially

Machine-Maker:
> Do you plan on deprecating Bukkit command aliases?

I don't think we have to, its pretty barebones as-is, and just won't get updates (like it hasn't gotten updates in years anyways).

Huge-mistake:
Bump, is there still a chance to merge this PR?   (╯°□°）

PR #9973, opened 2023-11-25
Add required-plugins config option

### LINKED ISSUE #8916: Config option to stop server if plugins failed to load

If a critical plugin fails to load, such as a claim protection plugin, but the server continues to start anyway and allows players to join this would put the server in a state where players are able to do things they otherwise wouldn't

A "required-plugins" config option which has a list of plugin names which the server will stop in the event they fail to load
Additionally it would be nice to just specify "*" to require that all plugins load

I have attempted to put together a PR for this but I have ran into issues with the config not yet being available during paper plugin bootstrapping, if someone wants to help with this, that would be great

### Other

_No response_

notTamion:
resolves #8916 
i was unsure where to add checks but i think they fit pretty nicely where they are now.


Owen1212055 (approved):

notTamion:
i have been using the hasStarted boolean from Watchdog since that kinda behaves like a isreloading boolean. however the watchdog thread is not started until after the first tick is done. so a required plugin could have done this:
image
and the server would not get stopped.

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
+index 9c7552968b8c017c71a7a77557a66a03ed89f125..20bf5892408dbd94417f48ae069ecbae7709c43f 100644
+--- a/src/main/java/io/papermc/paper/plugin/manager/PaperPluginInstanceManager.java
++++ b/src/main/java/io/papermc/paper/plugin/manager/PaperPluginInstanceManager.java
+@@ -210,6 +210,13 @@ class PaperPluginInstanceManager {
+         if (!(plugin instanceof JavaPlugin javaPlugin)) {
+             throw new IllegalArgumentException("Only expects java plugins.");
+         }
++        // Paper start
++        if (!this.server.isStopping() && !this.server.isReloading() && io.papermc.paper.configuration.GlobalConfiguration.get().misc.requiredPlugins.contains(plugin.getName())) {
++            this.server.getLogger().severe("The required Plugin " + plugin.getName() + " got disabled. Stopping Server.");
```
kennytv:
```suggestion
+            this.server.getLogger().severe("The required plugin " + plugin.getName() + " was disabled. Stopping Server.");
```

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
+index 1c9742ad81f04052d2c3bc18c7636f45b2fc5160..beab25cd9ed84135547db62d52636d7ee5b83525 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -275,6 +275,13 @@ public class DedicatedServer extends MinecraftServer implements ServerInterface
+         // CraftBukkit end
+ 
+         // Paper start
++        List<String> missingPlugins = new java.util.ArrayList<>(io.papermc.paper.configuration.GlobalConfiguration.get().misc.requiredPlugins);
++        missingPlugins.removeAll(java.util.Arrays.stream(this.server.paperPluginManager.getPlugins()).filter(org.bukkit.plugin.Plugin::isEnabled).map(org.bukkit.plugin.Plugin::getName).toList());
++        if(!missingPlugins.isEmpty()) {
```
kennytv: Missing space

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
++            DedicatedServer.LOGGER.error("The following required Plugins aren't enabled: " + String.join(", ", missingPlugins) + ". Stopping Server.");
```
kennytv: Lowercase plugins

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
+diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+index 1c9742ad81f04052d2c3bc18c7636f45b2fc5160..beab25cd9ed84135547db62d52636d7ee5b83525 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -275,6 +275,13 @@ public class DedicatedServer extends MinecraftServer implements ServerInterface
+         // CraftBukkit end
+ 
+         // Paper start
++        List<String> missingPlugins = new java.util.ArrayList<>(io.papermc.paper.configuration.GlobalConfiguration.get().misc.requiredPlugins);
++        missingPlugins.removeAll(java.util.Arrays.stream(this.server.paperPluginManager.getPlugins()).filter(org.bukkit.plugin.Plugin::isEnabled).map(org.bukkit.plugin.Plugin::getName).toList());
```
kennytv: Collect enabled plugins into a set and use containsAll to make this cleaner

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
+     private final io.papermc.paper.datapack.PaperDatapackManager datapackManager; // Paper
+     public static Exception excessiveVelEx; // Paper - Velocity warnings
+@@ -1009,7 +1010,10 @@ public final class CraftServer implements Server {
+ 
+     @Override
+     public void reload() {
+-        org.spigotmc.WatchdogThread.hasStarted = false; // Paper - Disable watchdog early timeout on reload
++        // Paper start
++        this.isReloading = true;
++        org.spigotmc.WatchdogThread.hasStarted = false; // Disable watchdog early timeout on reload
```
kennytv: Don't change this line, add the isReloading one with its own Paper comment

patches/server/1054-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,92 @@
...
+         this.reloadCount++;
+         this.configuration = YamlConfiguration.loadConfiguration(this.getConfigFile());
+         this.commandsConfiguration = YamlConfiguration.loadConfiguration(this.getCommandsConfigFile());
+@@ -1100,10 +1104,11 @@ public final class CraftServer implements Server {
+         this.enablePlugins(PluginLoadOrder.STARTUP);
+         this.enablePlugins(PluginLoadOrder.POSTWORLD);
+         this.getPluginManager().callEvent(new ServerLoadEvent(ServerLoadEvent.LoadType.RELOAD));
+-        org.spigotmc.WatchdogThread.hasStarted = true; // Paper - Disable watchdog early timeout on reload
++        // Paper start
++        org.spigotmc.WatchdogThread.hasStarted = true; // Disable watchdog early timeout on reload
```
kennytv: Same here

kennytv (changes requested):

patches/api/0449-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,44 @@
...
++++ b/src/main/java/org/bukkit/Bukkit.java
+@@ -985,6 +985,15 @@ public final class Bukkit {
+     public static @Nullable ItemStack createExplorerMap(@NotNull World world, @NotNull Location location, @NotNull org.bukkit.generator.structure.StructureType structureType, @NotNull org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored) {
+         return server.createExplorerMap(world, location, structureType, mapIcon, radius, findUnexplored);
+     }
++
++    /**
++     * Checks if the Server is currently Reloading
++     *
++     * @return if the Server is Reloading
```
powercasgamer:
```suggestion
+     * Checks if the server is currently reloading
+     *
+     * @return if the server is reloading
```
Might need a better explanation

patches/api/0449-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,44 @@
...
++++ b/src/main/java/org/bukkit/Server.java
+@@ -839,6 +839,13 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
+      *      org.bukkit.generator.structure.StructureType, int, boolean)
+      */
+     @Nullable ItemStack createExplorerMap(@NotNull World world, @NotNull Location location, @NotNull org.bukkit.generator.structure.StructureType structureType, @NotNull org.bukkit.map.MapCursor.Type mapIcon, int radius, boolean findUnexplored);
++
++    /**
++     * Checks if the Server is currently Reloading
++     *
++     * @return if the Server is Reloading
```
powercasgamer:
```suggestion
+     * Checks if the server is currently reloading
+     *
+     * @return if the server is reloading
```
Might need a better explanation

powercasgamer (changes requested):

patches/server/1054-Add-required-plugins-config-option.patch
Lulu13022002: You can use parameterized logging here

patches/server/1054-Add-required-plugins-config-option.patch
notTamion: i might be understanding this wrong. but that would mean that we can't give the user information on what plugins actually aren't enabled and on a server with a lot of plugins and a bunch of logging that information can be pretty useful so you don't have to dig through logs.

patches/api/0449-Add-required-plugins-config-option.patch
notTamion: i mean what do you suggest? "Checks if the server is currently reloading." seems self explanatory

patches/server/1055-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,82 @@
...
+     private final List<CraftPlayer> playerView;
++    public boolean isReloading; // Paper
+     public int reloadCount;
+     private final io.papermc.paper.datapack.PaperDatapackManager datapackManager; // Paper
+     public static Exception excessiveVelEx; // Paper - Velocity warnings
+@@ -1009,6 +1010,7 @@ public final class CraftServer implements Server {
+ 
+     @Override
+     public void reload() {
++        this.isReloading = true; // Paper
```
Machine-Maker: So this only covers 1 type of reload. There is the very much supported reload caused by `/minecraft:reload` which reloads datapacks, recipes, tags, advancements, etc. I think that should be covered by this as well.

patches/api/0449-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,44 @@
...
+     }
++
++    /**
++     * Checks if the server is currently reloading.
++     *
++     * @return if the server is reloading
++     */
++    public static boolean isReloading() {
++        return server.isReloading();
++    }
```
Machine-Maker: Create a separate `// Paper start - isReloading API` comment block instead of adding to an existing one. This is something I've been starting to do because it makes it more clear which patch diff comes from without having to look at the blame. Same in `Server` as well.

patches/server/1055-Add-required-plugins-config-option.patch
```diff
@@ -0,0 +1,82 @@
...
++++ b/src/main/java/io/papermc/paper/plugin/manager/PaperPluginInstanceManager.java
+@@ -211,6 +211,13 @@ class PaperPluginInstanceManager {
+         if (!(plugin instanceof JavaPlugin javaPlugin)) {
+             throw new IllegalArgumentException("Only expects java plugins.");
+         }
++        // Paper start
++        if (!this.server.isStopping() && !this.server.isReloading() && io.papermc.paper.configuration.GlobalConfiguration.get().misc.requiredPlugins.contains(plugin.getName())) {
++            this.server.getLogger().severe("The required plugin " + plugin.getName() + " got disabled. Stopping Server.");
++            this.server.shutdown();
++            return;
```
Machine-Maker: Is any weirdness caused by starting the shutdown here when this disable call could be in disablePlugins which is a loop over all the plugins, calling disable on each of them? I haven't tested it, it just might be a concern.

Machine-Maker (changes requested):

patches/server/1055-Add-required-plugins-config-option.patch
notTamion: after thinking about it for some time i think instead of adding the isReloading API i should just add an internal boolean called isReloadingPlugins. and then move the check from the initServer method into the last if statement of the CraftServer#enablePlugins.

patches/server/1055-Add-required-plugins-config-option.patch
notTamion: honestly i think just removing the shutdown here is best. maybe logging that there was an attempt to disable the plugin or making the method return a boolean depending on if the plugin was disabled succesfully

patches/server/0005-Paper-config-files.patch
```diff
@@ -784,6 +784,7 @@ index 0000000000000000000000000000000000000000..73a34b4e378e6012a01c8ac8b0922482
 +        public boolean useDimensionTypeForCustomSpawners = false;
 +        public boolean strictAdvancementDimensionCheck = false;
 +        public IntOr.Default compressionLevel = IntOr.Default.USE_DEFAULT;
++        public List<String> requiredPlugins = List.of();
```
Owen1212055: Set

notTamion: i just added it as an import cause maybe its gonna be needed in the future

notTamion:
perhaps some kind of Plugin#isRequired API could be added so pluginmanagers or whatever know not to disable that plugin. Though JavaPlugin is in the API and not Server so it can't access the GlobalConfiguration

PR #9976, opened 2023-11-26
[ci skip] Correct Windows requiring path to gradlew

notTamion:
with powershell being the "new" default terminal for windows (which does not load commands from the current directory by default) i think this should be mentioned

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,7 +67,7 @@ Assuming you have already forked the repository:
 
 1. Clone your fork to your local machine;
 2. Type `./gradlew applyPatches` in a terminal to apply the changes from upstream.
-In the Windows Command Prompt, leave out the `./` at the beginning for all `gradlew` commands;
+On Windows, replace the `./` with `.\` at the beginning for all `gradlew` commands;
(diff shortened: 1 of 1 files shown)


Owen1212055:
I do not get the point of this change, this seems obvious. I think that the people using this project will be aware of what terminal to use.

notTamion:
since powershell is the default and i think thats also what most people use the wording "On Windows" isn't correct imo. while yes people using the project will probably know that. It is still incorrect.

Owen1212055:
How is "powershell" the default? Opening intellij and pretty much any major IDE, you get a standard command terminal. 

I just think this is negligible.

notTamion:
wait huh. then i don't remember changing the default. what version of windows are you using?

Owen1212055:
We'll see what others think, perhaps.

emilyy-dev:
Could just say that on Windows you should use `.\gradlew.bat`, that'll work for both the old conhost and powershell

MiniDigger:
Pretty sure both intellij and windows terminal default to PowerShell nowadays, I agree with Emily's comment

Owen1212055 (approved):

lynxplay (approved):

PR #9981, opened 2023-11-28
Add MaterialTags for Armor Sets

thelooter:
Adds MaterialTags for the Following Sets

- Leather Armor
- Chainmail Armor
- Iron Armor
- Golden Armor
- Diamond Armor
- Netherite Armor


Machine-Maker:
I'm skeptical about adding more of these right now. I think we are planning on deprecating this whole API when upstream deprecates the Material enum.

They'll be replaced by a better system for adding custom tags. Of course if that happens in 2 years, its fine to add these now, but if its 1.21, then I'd rather not.

Owen1212055:
Yes, I prefer that we move more towards using future registry API to register custom tags instead of adding to this.

Thank you for your contribution, regardless. 😄

PR #9982, opened 2023-11-29
MergedProperties POC for sharing properties between projects

jpenilla:
This is an alternate way to share dependency versions between api & server to version catalogs (#9974), which have the problem where any catalog change that impacts the generated accessor ABI will cause the entire paperweight pipeline to run again (which makes it impractical for our use case)

If we decide to go forward with this, will need to merge https://github.com/PaperMC/paperweight/pull/226 and make a paperweight release first.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
     java
     `maven-publish`
     id("com.github.johnrengelman.shadow") version "8.1.1" apply false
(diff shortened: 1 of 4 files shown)


MiniDigger:
I like this idea, pretty much version catalogs but without the codegen stuff, so all the same advantages but no real disadvantages (other than lost type safety which seems fine)

PR #9984, opened 2023-11-29
Skip Folia's EntityScheduler executeTick checks if the scheduler hasn't been used at least once

MrPowerGamerBR:
This PR is mostly for "hey maybe there is a better solution for this", since I don't expect this PR to be accepted since it is mostly a hacky hack that probably has concurrency issues (accessing `hasScheduledAtLeastOneTask` outside of a synchronized block) and probably has a better solution.

Folia's EntityScheduler `executeTick` loop in `MinecraftServer`'s `tickChildren` is slooooow if you have a lot of entities (example: for a server with 15k entities, it uses ~1.7ms every tick, even if none of your plugins are using Folia's EntityScheduler)

To workaround this, we just don't attempt to do any of the checks if the scheduler hasn't been used at least once. Yes, I know, the scheduler increments the `tickCount`, but that doesn't matter since all the scheduled tasks are "relative" to the current tick count, not bound to the server's tick count.

The big performance boost comes from not needing to do all the checks `executeTick` does. Most server entities won't ever use a entity scheduler anyway, just think about it: Are plugins really going to add a scheduler to each item frame, each armor stand, each dropped item, each... you get my point.

Another way of doing this is changing the scheduler to store entities that have pending tasks into a global list to avoid getting all entities from all worlds, this is what I have done in my fork, and maybe this is a better less hacky solution, BUT that will probably make Folia explode, this needs to be tested further in a Folia environment instead of a Paper environment (on a Paper environment, I have been using this patch for one week+ without any issues): https://github.com/SparklyPower/SparklyPaper/blob/ver/1.20.2/patches/server/0010-Skip-EntityScheduler-s-executeTick-checks-if-there-i.patch (See PR #9985)

If anyone wants to reproduce the performance impact: Just spawn a bunch of entities, I used Armor Stands to test this.

(In the profiler, the loop is `net.minecraft.server.MinecraftServer.lambda$tickChildren$15()`)

**Profiler Before the Changes:** https://spark.lucko.me/eRvT1QqnI0
**Profiler After the Changes:** https://spark.lucko.me/DDAk1JQC0u

### DIFF
--- a/patches/server/0975-Folia-scheduler-and-owned-region-API.patch
+++ b/patches/server/0975-Folia-scheduler-and-owned-region-API.patch
@@ -18,10 +18,10 @@ the schedulers depending on the result of the ownership check.
 
 diff --git a/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java b/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5d5fd906d
+index 0000000000000000000000000000000000000000..f397d3a3dc1b8d8dd66982232e6e6a6542e9be96
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java
-@@ -0,0 +1,181 @@
+@@ -0,0 +1,186 @@
 +package io.papermc.paper.threadedregions;
 +
 +import ca.spottedleaf.concurrentutil.util.Validate;
@@ -69,6 +69,7 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +    private final Long2ObjectOpenHashMap<List<ScheduledTask>> oneTimeDelayed = new Long2ObjectOpenHashMap<>();
 +
 +    private final ArrayDeque<ScheduledTask> currentlyExecuting = new ArrayDeque<>();
++    private boolean hasScheduledAtLeastOneTask = false;
 +
 +    public EntityScheduler(final CraftEntity entity) {
 +        this.entity = Validate.notNull(entity);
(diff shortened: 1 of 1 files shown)


patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -162,6 +164,9 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +     * @throws IllegalStateException If the scheduler is retired.
 +     */
 +    public void executeTick() {
++        if (!this.hasScheduledAtLeastOneTask)
```
kennytv: Add brackets

patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -69,6 +69,7 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +    private final Long2ObjectOpenHashMap<List<ScheduledTask>> oneTimeDelayed = new Long2ObjectOpenHashMap<>();
 +
 +    private final ArrayDeque<ScheduledTask> currentlyExecuting = new ArrayDeque<>();
++    private boolean hasScheduledAtLeastOneTask = false;
```
kennytv: I'd maybe call this `awaitingFirstTask = true`

kennytv (changes requested):

MrPowerGamerBR:
I have made another PR that also solves the same issue, but that one is based off the patch I mentioned in this PR's description:  #9985

I decided to open a new PR instead of editing this one because I think both approaches are nice: 
* This one requires less changes, but if a EntityScheduler has been used at least once, then the scheduler will continue being ticked until retired.
* The other one requires more changes, but is cleaner and better overall.

kennytv:
Closing this in favor of https://github.com/PaperMC/Paper/pull/9985

PR #9985, opened 2023-11-29
Store entities with scheduled Folia EntityScheduler tasks in a set to avoid unnecessary checks and unnecessary iterations thru all entities

MrPowerGamerBR:
This is an alternative implementation of #9984, based on the patch that I'm running on my server. https://github.com/SparklyPower/SparklyPaper/blob/ver/1.20.2/patches/server/0010-Skip-EntityScheduler-s-executeTick-checks-if-there-i.patch

In my opinion this PR is better than other PR because it is cleaner and better, but it is more complex than the other one. I haven't tested on Folia tho to see if there is any concurrency issues BUT I think there wouldn't be any, all checks are made within the `stateLock` synchronized block.

Folia's EntityScheduler `executeTick` loop in `MinecraftServer`'s `tickChildren` is slooooow if you have a lot of entities (example: for a server with 15k entities, [it uses ~1.7ms every tick](https://spark.lucko.me/eRvT1QqnI0), even if none of your plugins are using Folia's EntityScheduler)

To workaround this, we keep all entities with scheduled tasks into a concurrent set, and we loop that concurrent set instead of looping thru all world's entities. Yes, I know, the scheduler increments the `tickCount`, but that doesn't matter since all the scheduled tasks are "relative" to the current tick count, not bound to the server's tick count. As long as we keep ticking the tick count when there are pending tasks, it works fine.

The biggest performance boost comes from not needing to do all the checks `executeTick` does. Most server entities won't ever use a entity scheduler anyway, just think about it: Are plugins really going to add a scheduler to each item frame, each armor stand, each dropped item, each... you get my point. This also has a nice performance boost compared to #9984 because we don't need to iterate thru all entities that exist in all worlds, we only need to iterate the entities that we know that have pending tasks.

About set vs list: *Maybe* it would be better to use a list, I have used a set because it needed to be concurrent, but I could change to a list (it would require some synchronization checks tho)

If anyone wants to reproduce the performance impact: Just spawn a bunch of entities, I used Armor Stands to test this, but any entity should work.

**Profiler Before the Changes:** https://spark.lucko.me/eRvT1QqnI0
**Profiler After the Changes:** *haven't done it yet but just pretend it is the same profiler result but with 0% lag because we are iterating a set with 0 entries*

### DIFF
--- a/patches/server/0851-Folia-scheduler-and-owned-region-API.patch
+++ b/patches/server/0851-Folia-scheduler-and-owned-region-API.patch
@@ -45,15 +45,16 @@ index d2dee700f2c5cc7d6a272e751a933901fe7a55b6..834b85f24df023642f8abf7213fe578a
          } catch (Throwable ex) {
 diff --git a/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java b/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5d5fd906d
+index 0000000000000000000000000000000000000000..ba5ff3af4d3af6f8ec4a47262aafb50ff1348549
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/threadedregions/EntityScheduler.java
-@@ -0,0 +1,181 @@
+@@ -0,0 +1,191 @@
 +package io.papermc.paper.threadedregions;
 +
 +import ca.spottedleaf.concurrentutil.util.Validate;
 +import io.papermc.paper.util.TickThread;
 +import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
++import net.minecraft.server.MinecraftServer;
 +import net.minecraft.world.entity.Entity;
 +import org.bukkit.craftbukkit.entity.CraftEntity;
 +
@@ -87,6 +88,7 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +     * The Entity. Note that it is the CraftEntity, since only that class properly tracks world transfers.
 +     */
 +    public final CraftEntity entity;
++    public final MinecraftServer server;
 +
 +    private static final record ScheduledTask(Consumer<? extends Entity> run, Consumer<? extends Entity> retired) {}
 +
@@ -97,7 +99,8 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +
 +    private final ArrayDeque<ScheduledTask> currentlyExecuting = new ArrayDeque<>();
 +
-+    public EntityScheduler(final CraftEntity entity) {
++    public EntityScheduler(final MinecraftServer server, final CraftEntity entity) {
++        this.server = Validate.notNull(server);
 +        this.entity = Validate.notNull(entity);
 +    }
 +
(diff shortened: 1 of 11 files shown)


patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -1121,26 +1131,30 @@ index 0000000000000000000000000000000000000000..d306f911757a4d556c82c0070d4837db
...
-index 9fff00b3706a97ba71a80f2ba39577b229325e02..1bac9f7894321b5710f5475f6598ec9f3ccd94d3 100644
+index 9fff00b3706a97ba71a80f2ba39577b229325e02..27d134de7788886c4d9cb614e0206661fe53dd34 100644
 --- a/src/main/java/net/minecraft/server/MinecraftServer.java
 +++ b/src/main/java/net/minecraft/server/MinecraftServer.java
-@@ -1466,6 +1466,20 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+@@ -305,6 +305,7 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+     public volatile Thread shutdownThread; // Paper
+     public volatile boolean abnormalExit = false; // Paper
+     public boolean isIteratingOverLevels = false; // Paper
++    public final Set<org.bukkit.craftbukkit.entity.CraftEntity> entitiesWithScheduledTasks = java.util.concurrent.ConcurrentHashMap.newKeySet(); // Paper - Folia schedulers
```
Spottedleaf: This needs to be the NMS entity to avoid issues with the CB entity being mapped to by multiple NMS entities.

patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -167,6 +172,11 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +        TickThread.ensureTickThread(thisEntity, "May not tick entity scheduler asynchronously");
 +        final List<ScheduledTask> toRun;
 +        synchronized (this.stateLock) {
++            // Do we *really* have scheduled tasks tho?
++            if (this.currentlyExecuting.isEmpty() && this.oneTimeDelayed.isEmpty()) { // Check if we have any pending tasks and, if not, skip!
++                this.server.entitiesWithScheduledTasks.remove(entity);
```
Spottedleaf: Inappropriate formatting, when accessing field variables you must use the `this` qualifier

patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -148,6 +152,7 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +            if (this.tickCount == RETIRED_TICK_COUNT) {
 +                return false;
 +            }
++            this.server.entitiesWithScheduledTasks.add(entity);
```
Spottedleaf: Inappropriate formatting, when accessing field variables you must use the `this` qualifier to clearly indicate the access is for a field

patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -90,6 +93,7 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +                throw new IllegalStateException("Already retired");
 +            }
 +            this.tickCount = RETIRED_TICK_COUNT;
++            this.server.entitiesWithScheduledTasks.remove(entity);
```
Spottedleaf: Inappropriate formatting, when accessing field variables you must use the `this` qualifier to clearly indicate the access is for a field

patches/server/0975-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -167,6 +172,11 @@ index 0000000000000000000000000000000000000000..62484ebf4550b05182f693a3180bbac5
 +        TickThread.ensureTickThread(thisEntity, "May not tick entity scheduler asynchronously");
 +        final List<ScheduledTask> toRun;
 +        synchronized (this.stateLock) {
++            // Do we *really* have scheduled tasks tho?
++            if (this.currentlyExecuting.isEmpty() && this.oneTimeDelayed.isEmpty()) { // Check if we have any pending tasks and, if not, skip!
++                this.server.entitiesWithScheduledTasks.remove(thisEntity);
++                return;
++            }
```
kennytv: This should be moved below the retired check

kennytv (approved):
Otherwise looks good

electronicboy (approved):

MrPowerGamerBR:
I still need to update this PR to rebase it to 1.20.6, but here's the same patch on my own updated 1.20.6 fork: https://github.com/SparklyPower/SparklyPaper/blob/ver/1.20.6/patches/server/0009-Skip-EntityScheduler-s-executeTick-checks-if-there-i.patch

MrPowerGamerBR:
Updated to 1.21

MrPowerGamerBR:
Superseded by https://github.com/PaperMC/Paper/commit/7c90c7c4970b44b8bcaebd1b3f4f124e9757dc3c :3

PR #9989, opened 2023-11-30
Chunk serialization events

derverdox:
This patch adds new Chunk events that are called during the serialization / generation process of chunks.

The use case of such events is that heavy saving/loading tasks from cache to the chunk nbt container or vice versa are executed asynchronously as well but in the same thread.


Owen1212055:
I like the principal here, and think that yes it would be very useful to be able to hook onto the chunk loading async in order to add your own data. However, the issue I see here is the PDC really thread safe here? We may need to offer some sort of alternate implementation to ensure that it is safely mutable async. Unsure though, since if the chunk loading is blocked until the async event completion is done theoretically it shouldn't have any duel-thread mutations?

PR #9990, opened 2023-11-30
Adding chunk serialization events

derverdox:
This patch adds new Chunk events that are called during the serialization / generation process of chunks.
A problem some developers may face while using the ChunkLoad or ChunkUnload events is the fact that they are called synchronously.
These proposed events are called during the process.

The use case of such events is that heavy saving / loading tasks from cache to the chunk nbt container or vice versa are executed asynchronously aswell but in the same thread
that saves / loads / generates the chunk.

### DIFF
--- a/patches/api/0449-Adding-ChunkSerialization-and-Generation-events.patch
+++ b/patches/api/0449-Adding-ChunkSerialization-and-Generation-events.patch
@@ -1,0 +1,204 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: derverdox <mail.ysp@web.de>
+Date: Thu, 30 Nov 2023 01:22:24 +0100
+Subject: [PATCH] Adding ChunkSerialization and Generation events.
+
+This patch adds new Chunk events that are called during the serialization / generation process of chunks.
+A problem some developers may face while using the ChunkLoad or ChunkUnload events is the fact that they are called synchronously.
+These proposed events are called during the process.
+
+The use case of such events is that heavy saving / loading tasks from cache to the chunk nbt container or vice versa are executed asynchronously aswell but in the same thread
+that saves / loads / generates the chunk. 
+
+diff --git a/src/main/java/io/papermc/paper/event/chunk/ChunkDataEvent.java b/src/main/java/io/papermc/paper/event/chunk/ChunkDataEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..73c94926c975360f9b7c79f910dd09a0f3346058
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/chunk/ChunkDataEvent.java
+@@ -0,0 +1,61 @@
++package io.papermc.paper.event.chunk;
++
++import org.bukkit.Bukkit;
++import org.bukkit.Chunk;
++import org.bukkit.World;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.bukkit.persistence.PersistentDataContainer;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public abstract class ChunkDataEvent extends Event {
++    private static final HandlerList handlers = new HandlerList();
++    private final World world;
++    private final Chunk chunk;
++    private final int chunkX;
++    private final int chunkZ;
++    private final PersistentDataContainer persistentDataContainer;
++
++    public ChunkDataEvent(@NotNull World world, @Nullable Chunk chunk, int chunkX, int chunkZ, @NotNull PersistentDataContainer chunkPersistentDataContainer) {
++        super(!Bukkit.isPrimaryThread());
++        this.world = world;
++        this.chunk = chunk;
++        this.chunkX = chunkX;
++        this.chunkZ = chunkZ;
++        this.persistentDataContainer = chunkPersistentDataContainer;
++    }
++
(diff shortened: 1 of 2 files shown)


patches/server/1055-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,226 @@
...
+Subject: [PATCH] Adding ChunkSerialization and Generation events.
+
+
+diff --git a/src/main/java/de/verdox/mccreativelab/worldgen/ChunkDataUtil.java b/src/main/java/de/verdox/mccreativelab/worldgen/ChunkDataUtil.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..682f4ac0708ba8bff474f8dedb14df03dd8e2189
+--- /dev/null
++++ b/src/main/java/de/verdox/mccreativelab/worldgen/ChunkDataUtil.java
+@@ -0,0 +1,37 @@
++package de.verdox.mccreativelab.worldgen;
```
powercasgamer: Why is this in another package? Also looks like the exact same methods are in CraftEventFactory?

derverdox: Oh I thought I removed it. This was a patch from a private fork.

patches/api/0449-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,204 @@
...
++        return chunkX;
++    }
++
++    public int getChunkZ() {
++        return chunkZ;
++    }
++
++    @Nullable
++    public Chunk getChunk() {
++        return this.chunk;
```
kennytv, line 57: This should get docs as to when the chunk is null or not null

patches/api/0449-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,204 @@
...
++import org.bukkit.Chunk;
++import org.bukkit.World;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.bukkit.persistence.PersistentDataContainer;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++public abstract class ChunkDataEvent extends Event {
++    private static final HandlerList handlers = new HandlerList();
```
kennytv, line 31: Uppercase name for static final fields

patches/api/0449-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,204 @@
...
++
++/**
++ * Called when a chunk is loaded from disk.
++ * You can use this event to manipulate the persistent data container of the chunk after it is loaded.
++ * Since by default chunk loading is asynchronously on paper this might be an alternative to the ChunkLoadEvent in some cases.
++ * This event might get called asynchronously.
++ */
++public class ChunkDeserializeEvent extends ChunkDataEvent {
++    private static final HandlerList handlers = new HandlerList();
++    public ChunkDeserializeEvent(@NotNull final World world, @Nullable final Chunk chunk, final int chunkX, final int chunkZ, @NotNull final PersistentDataContainer chunkPersistentDataContainer) {
```
kennytv, line 103: Extra line between the field and constructor

patches/server/1055-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,182 @@
...
+--- a/src/main/java/io/papermc/paper/chunk/system/scheduling/NewChunkHolder.java
++++ b/src/main/java/io/papermc/paper/chunk/system/scheduling/NewChunkHolder.java
+@@ -1817,7 +1817,9 @@ public final class NewChunkHolder {
+         public void run() {
+             final CompoundTag toSerialize;
+             try {
+-                toSerialize = ChunkSerializer.saveChunk(this.world, this.chunk, this.asyncSaveData);
++                // Paper start - add ChunkDataEvents
++                toSerialize = ChunkSerializer.saveChunk(this.world, this.chunk, this.asyncSaveData, false);
++                // Paper end - add ChunkDataEvents
```
kennytv, line 17: Single line changes just get a ` // Paper - ...` comment at the end of the line

patches/server/1055-Adding-ChunkSerialization-and-Generation-events.patch
```diff
@@ -0,0 +1,182 @@
...
++import net.minecraft.world.level.chunk.ChunkAccess;
++import net.minecraft.world.level.chunk.LevelChunk;
+ import net.minecraft.world.level.storage.loot.LootContext;
+ import net.minecraft.world.level.storage.loot.LootTable;
+ import net.minecraft.world.level.storage.loot.parameters.LootContextParams;
+@@ -242,6 +247,7 @@ import org.bukkit.inventory.EquipmentSlot;
+ import org.bukkit.inventory.InventoryView;
+ import org.bukkit.inventory.Recipe;
+ import org.bukkit.inventory.meta.BookMeta;
++import org.bukkit.persistence.PersistentDataContainer;
```
kennytv, line 151: Especially in this class it's going to be ugly, but please remove the added imports and instead fully qualify them when used in code to reduce diff

kennytv (changes requested):

PR #9991, opened 2023-11-30
Adding CustomModelChoice to bukkit recipe system

derverdox:
CustomModelChoice is used to create recipes that not only compare for materials but also CustomModelData.
Using the ExactChoice alternative is not a good idea since items can have different lores or attributes.
This is a great way to implement recipes with the new possibilities of Minecraft custom items.

### DIFF
--- a/patches/api/0449-Adding-CustomModelChoice.patch
+++ b/patches/api/0449-Adding-CustomModelChoice.patch
@@ -138,7 +138,7 @@ index 523818cbb0d6c90481ec97123e7fe0e2ff4eea14..f4cd027cb9adb23ae31dacba7d5205e4
 +
 +        @Override
 +        public @NotNull ItemStack getItemStack() {
-+            var choice = choices.get(0);
++            CustomItemData choice = choices.get(0);
 +            return choice.createStack();
 +        }
 +
--- a/patches/api/0449-Adding-CustomModelChoice.patch
+++ b/patches/api/0449-Adding-CustomModelChoice.patch
@@ -1,0 +1,185 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: derverdox <mail.ysp@web.de>
+Date: Thu, 30 Nov 2023 13:20:23 +0100
+Subject: [PATCH] Adding CustomModelChoice
+
+CustomModelChoice is used to create recipes that not only compare for materials but also CustomModelData.
+Using the ExactChoice alternative is not a good idea since items can have different lores or attributes.
+This is a great way to implement recipes with the new possibilities of Minecraft custom items.
+
+
+diff --git a/src/main/java/org/bukkit/inventory/CustomItemData.java b/src/main/java/org/bukkit/inventory/CustomItemData.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..3adbb5dbe9d53522b1ef528ba4c86d43f1dcef16
+--- /dev/null
++++ b/src/main/java/org/bukkit/inventory/CustomItemData.java
+@@ -0,0 +1,52 @@
++package org.bukkit.inventory;
++
++import org.bukkit.Material;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * CustomItemData representing a vanilla material and a customModelData value.
++ * Is used in {@link RecipeChoice.CustomModelChoice}
++ * @param material The vanilla material
++ * @param customModelData the customModelData value of the item.
++ */
++public record CustomItemData(@NotNull Material material, int customModelData) {
++    /**
++     * Creates an item stack with the material and the customModelData of the CustomItemDataObject
++     * @return the new item stack
++     */
++    public @NotNull ItemStack createStack() {
++        var stack = new ItemStack(material);
++        stack.editMeta(meta -> meta.setCustomModelData(customModelData));
++        return stack;
++    }
++
++    /**
++     * Checks if the provided item stack is equal based on its material and customModelData
++     * @param stackToCheck The item stack to check
++     * @return true if they are equal
++     */
++    public boolean isSame(@Nullable ItemStack stackToCheck) {
++        if(stackToCheck == null)
++            return false;
++        if (!stackToCheck.getType().equals(material))
++            return false;
++        return customModelData == ItemStack.getCustomModelData(stackToCheck);
++    }
++
++    /**
++     * Reads the CustomItemData from an item stack
++     * @param stack The item stack
++     * @return a new CustomItemData object
++     */
++    public static CustomItemData fromItemStack(@NotNull ItemStack stack){
++        return new CustomItemData(stack.getType(), ItemStack.getCustomModelData(stack));
++    }
++    @Override
++    public String toString() {
++        return "CustomItemData{" +
++            "material=" + material +
++            ", customModelData=" + customModelData +
++            '}';
++    }
++}
+diff --git a/src/main/java/org/bukkit/inventory/ItemStack.java b/src/main/java/org/bukkit/inventory/ItemStack.java
+index 0af73cc04edb93b9772136d4d808f657ea40e733..184cd662885f3ec96308315532631b9740f0a7f7 100644
+--- a/src/main/java/org/bukkit/inventory/ItemStack.java
++++ b/src/main/java/org/bukkit/inventory/ItemStack.java
+@@ -5,6 +5,7 @@ import com.google.common.collect.ImmutableMap;
+ import java.util.LinkedHashMap;
+ import java.util.List; // Paper
+ import java.util.Map;
++import java.util.Objects;
+ import org.bukkit.Bukkit;
+ import org.bukkit.Material;
+ import org.bukkit.Translatable;
+@@ -1005,4 +1006,32 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
+         return type.isAir() || amount <= 0;
+     }
+     // Paper end
++
++    // Paper start - Adding CustomModelChoice to bukkit recipes
++    /**
++     * Helper method to set customModelData of an ItemStack
++     * @param stack the ItemStack
++     * @param customModelData the new customModelData
++     */
++    public static void setCustomModelData(@NotNull ItemStack stack, int customModelData){
++        Objects.requireNonNull(stack);
++        stack.editMeta(meta -> meta.setCustomModelData(customModelData));
++    }
++
++    /**
++     * Helper method to get the customModelData of an ItemStack
++     * @param stack the ItemStack
++     * @return the customModelData
++     */
++
++    public static int getCustomModelData(ItemStack stack){
++        if(stack == null || stack.getType().isAir())
++            return 0;
++        if(!stack.hasItemMeta())
++            return 0;
++        if(!stack.getItemMeta().hasCustomModelData())
++            return 0;
++        return stack.getItemMeta().getCustomModelData();
++    }
++    // Paper end - Adding CustomModelChoice to bukkit recipes
+ }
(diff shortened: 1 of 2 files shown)


patches/api/0449-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,185 @@
...
++        Objects.requireNonNull(stack);
++        stack.editMeta(meta -> meta.setCustomModelData(customModelData));
++    }
++
++    /**
++     * Helper method to get the customModelData of an ItemStack
++     * @param stack the ItemStack
++     * @return the customModelData
++     */
++
```
powercasgamer:
```suggestion
```

patches/server/1055-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,110 @@
...
+-        } else {
++        } // Paper start - Adding CustomModelChoice to bukkit recipes
++        else if (bukkit instanceof RecipeChoice.CustomModelChoice choice) {
++            stack = new Ingredient(choice.getChoices().stream()
++                                         .map(customItemData -> CraftItemStack.asNMSCopy(customItemData.createStack()))
++                                         .map(Ingredient.ItemValue::new)
++            );
++            stack.isCustomModelChoice = true;
++        }
++        else { // Paper end - Adding CustomModelChoice to bukkit recipes
```
powercasgamer:
```suggestion
+        } else { // Paper end - Adding CustomModelChoice to bukkit recipes
```

patches/server/1055-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,110 @@
...
+-        } else {
++        } // Paper start - Adding CustomModelChoice to bukkit recipes
++        else if(list.isCustomModelChoice){
++            List<CustomItemData> choices = new ArrayList<>(list.itemStacks.length);
++            for (net.minecraft.world.item.ItemStack i : list.itemStacks) {
++                choices.add(CustomItemData.fromItemStack(CraftItemStack.asBukkitCopy(i)));
++            }
++            return new RecipeChoice.CustomModelChoice(choices);
++        }
++        else { // Paper end - Adding CustomModelChoice to bukkit recipes
```
powercasgamer:
```suggestion
+        } else { // Paper end - Adding CustomModelChoice to bukkit recipes
```

patches/server/1055-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,110 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
+@@ -23,6 +23,19 @@ import org.bukkit.material.MaterialData;
+ @DelegateDeserialization(ItemStack.class)
+ public final class CraftItemStack extends ItemStack {
+ 
++    // Paper start - Adding CustomModelChoice to bukkit recipes
++    public static void setCustomModelData(net.minecraft.world.item.ItemStack stack, int customModelData){
++        stack.getOrCreateTag().putInt("CustomModelData", customModelData);
++    }
++    public static int getCustomModelData(net.minecraft.world.item.ItemStack stack){
```
powercasgamer:
```suggestion
+    }
+
+    public static int getCustomModelData(net.minecraft.world.item.ItemStack stack){
```

patches/api/0449-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,185 @@
...
++    /**
++     * Checks if the provided item stack is equal based on its material and customModelData
++     * @param stackToCheck The item stack to check
++     * @return true if they are equal
++     */
++    public boolean isSame(@Nullable ItemStack stackToCheck) {
++        if(stackToCheck == null)
++            return false;
++        if (!stackToCheck.getType().equals(material))
++            return false;
```
powercasgamer:
```suggestion
+        if (!stackToCheck.getType().equals(material)) return false;
```

patches/api/0449-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,185 @@
...
++ * @param material The vanilla material
++ * @param customModelData the customModelData value of the item.
++ */
++public record CustomItemData(@NotNull Material material, int customModelData) {
++    /**
++     * Creates an item stack with the material and the customModelData of the CustomItemDataObject
++     * @return the new item stack
++     */
++    public @NotNull ItemStack createStack() {
++        var stack = new ItemStack(material);
```
powercasgamer:
```suggestion
+        ItemStack stack = new ItemStack(material);
```

patches/api/0449-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,185 @@
...
++            this.choices = Collections.unmodifiableList(List.of(customItemData));
++        }
++
++        public CustomModelChoice(Material material, int customModelData) {
++            this.choices = Collections.unmodifiableList(List.of(new CustomItemData(material, customModelData)));
++        }
++
++        @Override
++        public @NotNull ItemStack getItemStack() {
++            var choice = choices.get(0);
```
powercasgamer: Don't use var

patches/server/1055-Adding-CustomModelChoice.patch
```diff
@@ -0,0 +1,110 @@
...
+     public static final Codec<Ingredient> CODEC = Ingredient.codec(true);
+     public static final Codec<Ingredient> CODEC_NONEMPTY = Ingredient.codec(false);
+ 
+@@ -84,6 +86,13 @@ public final class Ingredient implements Predicate<ItemStack> {
+ 
+                     continue;
+                 }
++                else if(this.isCustomModelChoice){ // Paper start - Adding CustomModelChoice to bukkit recipes
++                    var material = itemstack.getBukkitStack().getType();
++                    var customModelData = CraftItemStack.getCustomModelData(itemstack);
```
powercasgamer: Don't use var

powercasgamer (changes requested):

lynxplay (commented):
This feels rather arbitrary IMO.
Preferably we just match specific nbt tags.

Custom model data feels somewhat randomly selected from the plethora of ways plugin developers can identify custom items. Given the fact that custom model data is not even the preferred way of identifying custom items (the PDC is) I am really not sure if I like a slapped on server side recipe choice that makes use of the custom model data value.

The fact that mojang does not support nbt matching recipes has been a long standing issue and I 100% can see the *technical* usage for this, however I think *if* we end up trying to slap a weaker version of ExactRecipe choice at it, then we should be looking into a more generally applicable implementation rather than grabbing a display value such as customModelData.

derverdox:
> This feels rather arbitrary IMO. Preferably we just match specific nbt tags.
> 
> Custom model data feels somewhat randomly selected from the plethora of ways plugin developers can identify custom items. Given the fact that custom model data is not even the preferred way of identifying custom items (the PDC is) I am really not sure if I like a slapped on server side recipe choice that makes use of the custom model data value.
> 
> The fact that mojang does not support nbt matching recipes has been a long standing issue and I 100% can see the _technical_ usage for this, however I think _if_ we end up trying to slap a weaker version of ExactRecipe choice at it, then we should be looking into a more generally applicable implementation rather than grabbing a display value such as customModelData.

I see your point. However, the idea of it is that we actively ignore the rest of the information in the pdc. I agree that we might need a more generalized approach but there are some things ExactChoice can't do for us.

Consider the following scenario:
I used to build a concept where every item had a description—based on where it came from it had a different description. They also had custom textures. So only comparing them by materials would not make sense. Instead, I would like to compare them using their custom model data and their material.

Let's say you have 20 different apples from different regions. They are all the same "item" but differ in their description based on the biome they were collected from.
I think it's not a good idea to create 20 different ExactChoices for every description. Besides, it would not be possible to change the description of the custom item dynamically because there would be no exact choice for that.
So comparing for customModelData as well makes sense in my head or am I missing something?

lynxplay:
What do you mean the *rest" of the PDC ? CustomModelData isn't in the pdc. I 100% get what you are going for here. I am not saying it would be useless.

I am more arguing that this change picks an arbitrary way to identify a custom item and promotes it to the API with a recipe choice when plenty of other ways exist.

It's in a way fixing a symptom of a way deeper issue, Mojang not permitting recipe ingredients via compound tag.

If I am identifying my custom items via a pdc tag, e.g. I have a saord that changes custom model data due to some in-game event but remains the same item because it's just a visual change, this PR fails to give me the ability to havw it as a recipe choice.

If I (for whatever reason) use the display name, same concept applies.
This approach is imo just flawed and now generic enough for what it is trying to achieve.

derverdox:
> What do you mean the *rest" of the PDC ? CustomModelData isn't in the pdc. I 100% get what you are going for here. I am not saying it would be useless.
> 
> I am more arguing that this change picks an arbitrary way to identify a custom item and promotes it to the API with a recipe choice when plenty of other ways exist.
> 
> It's in a way fixing a symptom of a way deeper issue, Mojang not permitting recipe ingredients via compound tag.
> 
> If I am identifying my custom items via a pdc tag, e.g. I have a saord that changes custom model data due to some in-game event but remains the same item because it's just a visual change, this PR fails to give me the ability to have it as a recipe choice.
> 
> If I (for whatever reason) use the display name, same concept applies. This approach is imo just flawed and generic enough for what it is trying to achieve.

I get your point. 
The fact that it solves problems vanilla recipes have with custom model data is a pleasant consequence.
A more generalized approach would be to have something like an ExactChoice that compares only to what is there.

Like a "predicate choice" where your input is a template ItemStack, the Choice compares every existing data in the nbt while ignoring what is not there.
If the template item has customModelData it compares for that. If it has a lore it compares with the lore.
If it has a custom display name it compares to that.
The difference with exactchoice is that it requires a replica. The "predicate choice" would only compare for data in the item stack nbt and ignore everything the template does not have.

lynxplay:
Yea pretty much. A "here is this compound tag of nbt, only compare the inputs values if the provided tag even has that value" recipe choice would be nice but rn I am unsure if we have nice enough API to model that kind of item nbt 

I presume whenever the properties API makes it in, that might be nicer.

Machine-Maker:
Do you know if this works with shapeless recipes? My initial guess is that it does not.

derverdox:
> Do you know if this works with shapeless recipes? My initial guess is that it does not. The initial implementation of ExactChoice also did not work with shapeless recipes which upstream just decided to ignore putting in the javadoc that it didn't work with them. I don't really want to go back to that adding more RecipeChoice that doesn't work in all recipes.

Ive been using this implementation for quite a while now but I am willing to do a full testing session to make clear it works on every recipe! :)

derverdox:
> Do you know if this works with shapeless recipes? My initial guess is that it does not. The initial implementation of ExactChoice also did not work with shapeless recipes which upstream just decided to ignore putting in the javadoc that it didn't work with them. I don't really want to go back to that adding more RecipeChoice that doesn't work in all recipes.

Is there a specific reason why we are still using StackingContents and RecipePicker instead of building a whole new system? I was thinking about PredicateChoice. MaterialChoice and ExactChoice would extend PredicateChoice. However I don't know if we can get same performance achieved by Bitset dfs etc...

electronicboy:
The nature of ExactChoice being flawed is more an issue with how bukkit represents itemstacks and would generally stand a chance of being fixed once we get the property API
Replacing entire vanilla systems generally needs a strong reason when plugins have been delivering this within their own code for years now, especially given the expectations that we'd be expected to maintain it, while mojang is working on stuff which would generally make the custom model data irrelevant for a good chunk of its usecases

derverdox:
> The nature of ExactChoice being flawed is more an issue with how bukkit represents itemstacks and would generally stand a chance of being fixed once we get the property API Replacing entire vanilla systems generally needs a strong reason when plugins have been delivering this within their own code for years now, especially given the expectations that we'd be expected to maintain it, while mojang is working on stuff which would generally make the custom model data irrelevant for a good chunk of its usecases

I agree. However, I think we need a more generalized approach to implementing custom recipes.

PR #9992, opened 2023-12-02
Updates documentation in EntityEquipment.java

### LINKED ISSUE #9867: EntityEquipment drop chances docs are wrong or missing API
The docs for the EntityEquipment drop chance-related methods all say 
> A drop chance of 1.0F will always drop

which is misleading/false. Setting the chance to 1.0F will make it always drop if a player is responsible for killing the entity. Setting it to 1.00001F (or anything above 1) will make it so that it will drop when killed by anything, not just a player. This either needs to be documented or API added to set guaranteed drops.

waterlily1:
Updates documentation in EntityEquopment.java to reflect behavior mentioned in issue #9867


electronicboy:
If these are all methods from upstream, it should probably be inside of the upstream javadoc fixes patch rather than its own

Machine-Maker (approved):
Looks good. Can be moved into the upstream javadocs fixes patch either by the author or the merger.

PR #9996, opened 2023-12-03
Adding PredicateChoice to Paper API

derverdox:
3 days ago I posted a pull request #9991 about adding CustomModelChoice. However, we came to the conclusion that we need a more generalized approach.
As a result, I began working on what is called PredicateChoice.
To make it work the StackedContents logic had to be edited.
In the ExtraContentsMap we don't store the exact ingredient items of every recipe ingredient. Instead, whenever we account a new stack in StackedContents logic we compare it to the provided predicates and store them as "ingredient items" if they fulfill at least one predicate.

We act like the items in the inventory of the player are the exact choices when any ingredient predicate returns true.

It gets more apparent when looking at the changes.
The first tests showed that it works like a charm.

Pros
- With predicate choice one can implement his own version of ItemPredicate. This allows for the programmer to make the recipe check for whatever he wants.
- They still work as intended if you want to use MaterialChoice or ExactChoice.

Cons
- If you use PredicateChoice the predicate tests Bukkit ItemStacks. Consequently, NMS stacks need to instantiate their Craftbukkit mirror to make the checks happen.

Let me know your thoughts! :) Maybe I have missed something.


patches/api/0449-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,187 @@
...
+Subject: [PATCH] Adding PredicateChoice to Paper API
+
+
+diff --git a/src/main/java/de/verdox/predicatechoice/ItemPredicate.java b/src/main/java/de/verdox/predicatechoice/ItemPredicate.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..cbddd2e8f5d8a1c19c56029af5608ed8f2e8cc6c
+--- /dev/null
++++ b/src/main/java/de/verdox/predicatechoice/ItemPredicate.java
+@@ -0,0 +1,95 @@
++package de.verdox.predicatechoice;
```
Machine-Maker: Wrong package

patches/api/0449-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,187 @@
...
++import java.util.ArrayList;
++import java.util.Arrays;
++import java.util.List;
++import java.util.function.Predicate;
++import java.util.stream.Collectors;
++
++public interface ItemPredicate extends Predicate<ItemStack> {
++    List<ItemStack> recipeBookExamples();
++
++    class MaterialPredicate implements ItemPredicate {
```
Machine-Maker: I do not think we should be re-implementing Material and Exact recipe choices as predicate choices. Those already work as is, we should be able to add predicate choices on top.

patches/api/0449-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,187 @@
...
++
++import com.google.common.base.Preconditions;
++import org.bukkit.inventory.ItemStack;
++import org.bukkit.inventory.RecipeChoice;
++import org.jetbrains.annotations.NotNull;
++import java.util.Collections;
++import java.util.List;
++import java.util.Objects;
++
++public record PredicateChoice(ItemPredicate itemPredicate) implements RecipeChoice {
```
Machine-Maker: No public records exposed to the API. PredicateChoice should be an interface with static factory methods to create new instances of it that take recipe book examples and a `Predicate<ItemStack>`. Make the interface sealed that only permits the package-private record to implement it.

patches/api/0449-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,187 @@
...
++        if (itemPredicate.recipeBookExamples().size() > 1) {
++            stack.setDurability(Short.MAX_VALUE);
++        }
++
++        return stack;
++    }
++
++    @Override
++    public PredicateChoice clone() {
++        return new PredicateChoice(itemPredicate);
```
Machine-Maker: Asking a non rhetorical question, does this violate the clone method contract? (reusing the predicate instance)

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,210 @@
...
+         final int existing = this.exactChoiceIds.getInt(exactChoice);
+         if (existing > -1) {
+             return existing;
+         }
+         final int id = this.idCounter.getAndIncrement();
++        boolean isPotentialIngredient = false;
++        for (final Ingredient ingredient : this.recipe.getIngredients()) {
++            if (ingredient.test(exactChoice)) {
++                isPotentialIngredient = true;
++                final IntList idList = extraStackingIds.computeIfAbsent(ingredient, ingredient1 -> new IntArrayList());
```
Machine-Maker: Should already be in the map right? I think it should throw if its not (or leave this, and don't pre-populate in the initialize method)

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,210 @@
...
+-            }
++        this.recipe = recipe;
++        for (final Ingredient ingredient : recipe.getIngredients()) {
++            if(ingredient.isEmpty())
++                continue;
++            extraStackingIds.put(ingredient, new IntArrayList());
+         }
+     }
+ 
+     private int registerExact(final ItemStack exactChoice) {
```
Machine-Maker: Method name, param name, need to change. Not actually what it does anymore.

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,210 @@
...
++        if(exactChoice.isEmpty())
++            return -1;
+         final int existing = this.exactChoiceIds.getInt(exactChoice);
+         if (existing > -1) {
+             return existing;
+         }
+         final int id = this.idCounter.getAndIncrement();
++        boolean isPotentialIngredient = false;
++        for (final Ingredient ingredient : this.recipe.getIngredients()) {
++            if (ingredient.test(exactChoice)) {
```
Machine-Maker: Instead of testing against all the ingredients, why not leave all the non-predicate ingredients alone, and then only test against the predicate ingredients?

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,210 @@
...
+             this.accountStack(stack);
+         }
+@@ -39,7 +40,8 @@ public class StackedContents {
+         if (!stack.isEmpty()) {
+             int i = getStackingIndex(stack);
+             int j = Math.min(maxCount, stack.getCount());
+-            if (this.extrasMap != null && stack.hasTag() && this.extrasMap.accountStack(stack, j)) return; // Paper - if an exact ingredient, don't include it
++            // We always want to use extra map!
++            if (this.extrasMap != null /*&& stack.hasTag()*/ && this.extrasMap.accountStack(stack, j)) return; // Paper - if an exact ingredient, don't include it
+             this.put(i, j);
```
Machine-Maker: The extras map is now just completely replacing the logic for accounting stacks. Probably good to mention that, rename the field accordingly.

Machine-Maker (changes requested):
Missing all sorts of `// Paper` comments, but that's not a big deal.

Overall, this works with my initial testing. I haven't yet done any spark profiles yet to see if the performance is significantly worse or anything.

My main concern with this is the unnecessary replacement of MaterialChoice and ExactChoice. We should just be able to add a 3rd (PredicateChoice) to that lineup without needing to replace the others.

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
derverdox: Makes more sense i agree

derverdox:
> [This is just a quick mockup](https://github.com/Machine-Maker/Paper/commit/8efaaaafeed89ae92c03ca850f87ea706ce55f33) of what I had in mind. Ignore the API, it is just quicker to make it not an interface+record. I didn't exhaustively test it, but the test in the test plugin work so far.

Looks good to me! :) My PR was more of a proof of concept. I am fine with your implementation.

patches/api/0449-Adding-PredicateChoice-to-Paper-API.patch
derverdox: I think it depends on the actual implementation of itemPredicate. Maybe it would be better if the interface provided a cloning method as well.

patches/server/1055-Adding-PredicateChoice-to-Paper-API.patch
derverdox: Yes.

Machine-Maker:
You can go ahead and fixup yours

derverdox:
> You can go ahead and fixup yours

Allright

derverdox (commented):
I created two new patches. Your implementation heavily inspires them. It is way cleaner to use the logic when a predicate is used.
Even fewer code changes than before. I tested things myself, and it works like a charm :)

Since we know that the extra map logic works, we use working code and add more item contents to its storage. In this case, the items that match any predicate choice.

It is a simple solution that brings many new possibilities to the table. 
Hope there is no fatal flaw in this idea.

patches/server/1056-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,212 @@
...
+     }
+ 
+     public boolean accountStack(final ItemStack stack, final int count) {
++        // We are adding items that pass the predicate test.
++        for(final Ingredient predicateChoice : this.predicateChoices){
++            if(predicateChoice.test(stack)){
++                boolean isStackTypeRegistered = this.exactChoiceIds.getInt(stack) > -1;
++                final int id = this.registerExact(stack);
++                // We only want to add the stacking id to the list one time
++                if(!isStackTypeRegistered)
```
Machine-Maker: You should just be able to check the `id` returned by `registerExact` if it is -1, it's already been added to the exactChoice map.

patches/api/0462-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,75 @@
...
++
++        @Override
++        public boolean equals(final Object o) {
++            if (this == o) return true;
++            if (o == null || getClass() != o.getClass()) return false;
++            PredicateChoice that = (PredicateChoice) o;
++            return Objects.equals(itemPredicate, that.itemPredicate);
++        }
++
++        public static interface ItemPredicate extends Predicate<ItemStack> {
```
Machine-Maker: This type shouldn't exist

patches/api/0462-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,75 @@
...
++++ b/src/main/java/org/bukkit/inventory/RecipeChoice.java
+@@ -233,4 +233,64 @@ public interface RecipeChoice extends Predicate<ItemStack>, Cloneable {
+             return "ExactChoice{" + "choices=" + choices + '}';
+         }
+     }
++
++    /**
++     * Represents a choice that matches when the item predicate is fulfilled.
++     */
++    public static class PredicateChoice implements RecipeChoice {
```
Machine-Maker: This should be an interface that is backed by a package-private record implementation with static methods to create instances of it.

patches/server/1056-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,212 @@
...
+ import net.minecraft.world.item.ItemStack;
+ import net.minecraft.world.level.ItemLike;
++import org.bukkit.craftbukkit.inventory.CraftItemStack;
++import org.bukkit.inventory.RecipeChoice;
+ 
+ public final class Ingredient implements Predicate<ItemStack> {
+ 
+     public static final Ingredient EMPTY = new Ingredient(Stream.empty());
+     private final Ingredient.Value[] values;
++    @Nullable private RecipeChoice.PredicateChoice.ItemPredicate itemPredicate; // Paper
```
Machine-Maker: Just have a *public* field for Predicate<ItemStack>

patches/server/1056-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,212 @@
...
+     public static final Codec<Ingredient> CODEC = Ingredient.codec(true);
+     public static final Codec<Ingredient> CODEC_NONEMPTY = Ingredient.codec(false);
+-
++    // Paper start - Adding PredicateChoice
++    public Ingredient(RecipeChoice.PredicateChoice.ItemPredicate itemPredicate) {
++        List<org.bukkit.inventory.ItemStack> bukkitChoices = itemPredicate.recipeBookExamples();
++        this.itemPredicate = itemPredicate;
++        this.values = bukkitChoices.stream().map(CraftItemStack::asNMSCopy).map(ItemValue::new).toArray(Value[]::new);
++    }
++    // Paper end - Adding PredicateChoice
```
Machine-Maker: No need for a custom constructor, do all the logic in CraftRecipe

patches/server/1056-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,212 @@
...
+@@ -269,4 +283,10 @@ public final class Ingredient implements Predicate<ItemStack> {
+             return Collections.singleton(this.item);
+         }
+     }
++    // Paper start - Adding PredicateChoice
++    @Nullable
++    public RecipeChoice.PredicateChoice.ItemPredicate getItemPredicate() {
++        return itemPredicate;
++    }
++    // Paper end - Adding PredicateChoice
```
Machine-Maker: with a public field, no need for a getter

patches/server/1056-Adding-PredicateChoice-to-Paper-API.patch
```diff
@@ -0,0 +1,212 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftRecipe.java
+@@ -30,7 +30,13 @@ public interface CraftRecipe extends Recipe {
+         } else if (bukkit instanceof RecipeChoice.ExactChoice) {
+             stack = new Ingredient(((RecipeChoice.ExactChoice) bukkit).getChoices().stream().map((mat) -> new net.minecraft.world.item.crafting.Ingredient.ItemValue(CraftItemStack.asNMSCopy(mat))));
+             stack.exact = true;
+-        } else {
++        }
++        // Paper start - Adding PredicateChoice
++        else if(bukkit instanceof RecipeChoice.PredicateChoice predicateChoice){
++            stack = new Ingredient(predicateChoice.getItemPredicate());
```
Machine-Maker: Just use the regular constructor and pass in the recipe book examples as itemstacks and then assign the predicate field on the next line.

Machine-Maker (changes requested):

partydev:
Upon testing this found that clicking the recipe book item does not move the items in your inventory to the crafting grid.
image
Ingredients have custom model data, lore, and attributes set.

JakeGBLP:
Is this still being worked on?

derverdox:
> Is this still being worked on?

Hello! I was working on my bachelor thesis lately so i was not able to fully work on this.
Will continue soon!

derverdox:
> Upon testing this found that clicking the recipe book item does not move the items in your inventory to the crafting grid. image Ingredients have custom model data, lore, and attributes set. The recipe does work, just not the recipe books behavior. This may be more to blame on the recipe book patch than this patch.

Hmmm. I will look into this!

derverdox:
> Upon testing this found that clicking the recipe book item does not move the items in your inventory to the crafting grid. image Ingredients have custom model data, lore, and attributes set. The recipe does work, just not the recipe books behavior. This may be more to blame on the recipe book patch than this patch.

Hey. I just uploaded a fix to the recipe book problem. The issue seems to be resolved. It would be nice if you could test as well! :)

patches/api/0469-Adding-PredicateChoice.patch
```diff
@@ -0,0 +1,94 @@
...
++        if (choices.size() > 1) {
++            stack.setDurability(Short.MAX_VALUE);
++            return stack;
++        }
++        return stack;
++    }
++
++    @Override
++    public @NotNull RecipeChoice clone() {
++        return new PredicateChoiceImpl(predicate, recipeBookExamples());
```
Machine-Maker, line 38: This also has to clone each itemstack in the recipe book examples because ItemStack is a mutable type.

patches/api/0469-Adding-PredicateChoice.patch
```diff
@@ -0,0 +1,94 @@
...
++            return new PredicateChoiceImpl(predicate, List.of(recipeBookExamples));
++        }
++
++        static PredicateChoice create(@NotNull Predicate<ItemStack> predicate, java.util.Collection<ItemStack> recipeBookExamples){
++            Objects.requireNonNull(predicate, "The item predicate cannot be null!");
++            Objects.requireNonNull(predicate, "The mustHaveRecipeBookExample cannot be null!");
++            if(recipeBookExamples.isEmpty())
++                throw new IllegalArgumentException("Please provide at least one recipe book example item!");
++            return new PredicateChoiceImpl(predicate, List.copyOf(recipeBookExamples));
++        }
```
Machine-Maker, line 79: I would move these methods up 1 level to be on RecipeChoice itself and named `createPredicateChoice`

An alternative would be to just make PredicateChoice a top-level class next to RecipeChoice and have `create` methods in it. Not a fan of the static factory methods inside an inner class.

Machine-Maker (changes requested):
Some of the diff in the server patch can be moved back to the patch that created it, such as simple renames, like the hasExactIngredients but that doesn't have to be done now, can be done before merging.

We also need to make it clear that *mutating* the stack within the Predicate provided in the `create` method does not support mutating the stack in an way. I don't want to open that can of worms, and I also don't want to copy the itemstack each time that predicate might be called.

patches/api/0469-Adding-PredicateChoice.patch
```diff
@@ -0,0 +1,94 @@
...
+     }
++    // Paper start - Adding PredicateChoice
++    /**
++     * Represents a choice that matches when the item predicate is fulfilled.
++     */
++
++    interface PredicateChoice extends RecipeChoice {
++        static PredicateChoice create(@NotNull Predicate<ItemStack> predicate, ItemStack... recipeBookExamples){
++            Objects.requireNonNull(predicate, "The item predicate cannot be null!");
++            Objects.requireNonNull(predicate, "The mustHaveRecipeBookExample cannot be null!");
```
dawon, line 67: This line checks the exactly same thing as the one above... I think here should be recipeBookExamples?

patches/api/0469-Adding-PredicateChoice.patch
```diff
@@ -0,0 +1,94 @@
...
++            if(recipeBookExamples.length == 0)
++                throw new IllegalArgumentException("Please provide at least one recipe book example item!");
++            return new PredicateChoiceImpl(predicate, List.of(recipeBookExamples));
++        }
++
++        static PredicateChoice create(@NotNull Predicate<ItemStack> predicate, java.util.Collection<ItemStack> recipeBookExamples){
++            Objects.requireNonNull(predicate, "The item predicate cannot be null!");
++            Objects.requireNonNull(predicate, "The mustHaveRecipeBookExample cannot be null!");
```
dawon, line 75: This line checks the exactly same thing as the one above... I think here should be recipeBookExamples?

dawon (changes requested):

JakeGBLP:
@derverdox please get back to this at some point 💪

lynxplay:
This was actually overtaken in https://github.com/PaperMC/Paper/pull/12017

JakeGBLP:
> This was actually overtaken in #12017

My bad I missed it

PR #9998, opened 2023-12-04
Add EntityCollideWithEntityEvent

### LINKED ISSUE #8729: Entity with Entity collide event

I am currently writing a plugin that simulates a cavalry "charge" of sorts. Basically, while riding a horse, if you run into an entity, that entity gets damaged and knocked back a bit.

I found the VehicleEntityCollisionEvent but apparently that only fires for boats and minecarts.

Either:
- Expanding the VehicleEntityCollisionEvent to fire for horses
- A new EntityCollideWithEntityEvent (not too sure on the naming here lol)

Currently, for this I have to use the EntityMoveEvent and check the bounding boxes there.

### Other

_No response_

notTamion:
resolves #8729 
i decided to store the entities in a set since its not like there is one entity pushing another it just depends on which entity gets ticked first and then pushes both entities in opposite directions.
post hardfork #11906

### DIFF
--- a/patches/api/0365-Collision-API.patch
+++ b/patches/api/0365-Collision-API.patch
@@ -3,7 +3,68 @@ From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
 Date: Wed, 6 Oct 2021 20:10:36 -0400
 Subject: [PATCH] Collision API
 
+Co-authored-by: Tamion <70228790+notTamion@users.noreply.github.com>
 
+diff --git a/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java b/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..0505e5479526bcfb757e6078d6bbec423498c303
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
+@@ -0,0 +1,54 @@
++package io.papermc.paper.event.entity;
++
++import org.bukkit.entity.Entity;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++import java.util.Set;
++
++/**
++ * Fired when two entities collide with each other.
++ * If cancelled, the entities won't get pushed away from each other.
++ */
++public class EntityCollideWithEntityEvent extends Event implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private boolean cancelled;
++    private final Set<Entity> entities;
++
++    @ApiStatus.Internal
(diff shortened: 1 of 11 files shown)


patches/api/0373-Collision-API.patch
```diff
@@ -3,7 +3,64 @@ From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
...
++
++public class EntityCollideWithEntityEvent extends Event implements Cancellable {
++
++    private static final HandlerList handlers = new HandlerList();
++    private boolean canceled;
++    private final Set<Entity> entities = new HashSet<>();
++
++    public EntityCollideWithEntityEvent(@NotNull Entity entity1, @NotNull Entity entity2) {
++        this.entities.add(entity1);
++        this.entities.add(entity2);
```
Sytm: Perhaps use `Set.of(entity1, entity2)` instead to create an immutable set, because being able to change it does not really make sense for this implementation

patches/api/0373-Collision-API.patch
```diff
@@ -3,7 +3,64 @@ From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
...
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++import java.util.HashSet;
++import java.util.Set;
++
++public class EntityCollideWithEntityEvent extends Event implements Cancellable {
++
++    private static final HandlerList handlers = new HandlerList();
++    private boolean canceled;
```
Sytm: Should be called `cancelled`

patches/api/0373-Collision-API.patch
```diff
@@ -3,7 +3,63 @@ From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
...
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++import java.util.HashSet;
++import java.util.Set;
++
++public class EntityCollideWithEntityEvent extends Event implements Cancellable {
++
++    private static final HandlerList handlers = new HandlerList();
```
Machine-Maker: `HANDLER_LIST`. Best to use the standards for naming static finals (no matter that upstream doesn't do that)

patches/api/0373-Collision-API.patch
```diff
@@ -3,7 +3,63 @@ From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
...
++    private boolean cancelled;
++    private final Set<Entity> entities;
++
++    public EntityCollideWithEntityEvent(@NotNull Entity entity1, @NotNull Entity entity2) {
```
Machine-Maker: Add `ApiStatus.Internal` annotation to the constructor

Machine-Maker (changes requested):

patches/api/0373-Collision-API.patch
imDaniX: Isn't this naming style is used only for constants? `HandlerList` is changing when (un)registering listeners, so it should not be considered a constant.

electronicboy: convention within bukkit is to call it `HANDLER_LIST`

Warriorrrr (approved):

kennytv (approved):

PR #10002, opened 2023-12-06
Add proper version check to startup

notTamion:
removes the shitty "version check" which basically just checks if the build is more than 21 days old.
I put the new check right after the git information is printed, which is probably where it fits the nicest without having to change anything about how the PaperVersionFetcher and the methods it calls work.

Unsure if this should be merged into the Implement PaperVersionFetcher patch


lynxplay:
Given there is no inherit value in this information and administrators that want to "skip" the current "out-of-date" message can just use the property, I am closing this PR.

I see the technical usage of it, but I don't think the extra web call to either downloads/github API are useful for anyone,

PR #10006, opened 2023-12-08
Set true custom payload channel size limit

### LINKED ISSUE #10003: Channel name length restriction is too short
### Expected behavior

Channel names can be allowed to be longer than 64 characters. Plugins channels are for interop with client mods, if modded frameworks such as NeoForge or Fabric don't enforce a limit, Paper shouldn't either. 

### Observed/Actual behavior

I got this bug report about my mod unable to connect to a Paper server on a Velocity server. I was told to report this as this network channel name limit is quite odd and very short.
https://github.com/TelepathicGrunt/Bumblezone/issues/315

Log from the user: https://pastebin.com/xyM0sFfB

> org.bukkit.plugin.messaging.ChannelNameTooLongException: Attempted to send a Plugin Message to a channel that was too large. The maximum length a channel may be is 64 chars (attempted 75 - 'the_bumblezone:networking/the_bumblezone/queen_randomize_trades_sync_packet.
>         at org.bukkit.plugin.messaging.StandardMessenger.validateAndCorrectChannel(StandardMessenger.java:500) ~[paper-api-1.20.1-R0.1-SNAPSHOT.jar:?]

### Steps/models to reproduce

Use 1.20.1 Bumblezone to connect to a Velocity server running Paper.

### Plugin and Datapack List

N/A

### Paper version

N/A (user did not say but I'll bring them to this issue report)

### Other

_No response_

EpicPlayerA10:
According to [Plugin Message packet](https://wiki.vg/Protocol#Serverbound_Plugin_Message_.28play.29) the channel itself does not have a limit, but `Identifier` has a limit. If we look up here we can see that `Identifier` can hold text of length `32767` bytes. So the max possible channel size is that `32767` bytes.

I wonder why spigot left there so small channel size limit. Maybe it's because minecraft used to have a limit set on the channel name but now it doesn't. 
EDIT: Yeah it would match: https://wiki.vg/index.php?title=Plugin_channels&diff=14556&oldid=14514

Related issue: #10003

### DIFF
--- a/patches/api/0180-Set-true-custom-payload-channel-size-limit.patch
+++ b/patches/api/0180-Set-true-custom-payload-channel-size-limit.patch
@@ -3,7 +3,7 @@ From: Shane Freeder <theboyetronic@gmail.com>
 Date: Fri, 18 Oct 2019 17:39:05 +0100
 Subject: [PATCH] Set true custom payload channel size limit
 
-This fixes compatibility with some mods that are sending very long channel names. Also gives developers the ability to send logner channel names.
+This fixes compatibility with some mods that are sending very long channel names. Also gives developers the ability to send longer channel names.
 
 diff --git a/src/main/java/org/bukkit/plugin/messaging/Messenger.java b/src/main/java/org/bukkit/plugin/messaging/Messenger.java
 index 9d2c68c826f3b867d407e7f13c6394a899cc8ee8..ca6442ebe6a01a546e2f6d5c79583189ed0b0292 100644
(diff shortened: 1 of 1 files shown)


kennytv:
I'd keep the system property in case someone wants to lower it

EpicPlayerA10:
> I'd keep the system property in case someone wants to lower it for whatever reason

Alright

lynxplay (approved):
Merger pls use java.lang.Short.MAX_VALUE

Owen1212055 (approved):

PR #10007, opened 2023-12-09
fix more incorrect array nullability annotations

Machine-Maker:
Fixes incorrect array nullability annotations. For reference for anyone not reading this, arrays require 2 annotations (1 if the element type is primitive). Like
```java
void someMethod(@NotNull String @Nullable [] array);
```
means that you can pass null or an array with no null elements.


kennytv:
Not sure I like that amount of diff over what's *effectively* a cosmetic change, I think that should be left for later

PR #10008, opened 2023-12-09
Update README.md

Luispupuis:


lynxplay (commented):
The project is called paper, the readme is correct here.

PR #10016, opened 2023-12-10
fix NPE on EntityTeleportEvent getTo

Machine-Maker:
Almost everywhere EntityTeleportEvent (or a subclass) was used, the nullability of getTo was never checked despite it being marked as nullable (as well as setTo). This is really dumb, as what is null supposed to do? So I just made it act as being cancelled.

Also in one spot, the firing of the event for /teleport, the to location wasn't even respected if you called setTo.


kennytv:
Why not just cancel the event if passed null if the param should stay nullable?

kennytv:
Or honestly, just make it notnull if it's broken anyways lol

Machine-Maker:
> Or honestly, just make it notnull if it's broken anyways lol

I considered that, but I would also have to change it for all the subclasses of EntityTeleportEvent which have constructors that are marked with `@Nullable`. Specifically for the `EntityPortalEvent` which *has* always respected null getTo.

Owen1212055 (approved):

PR #10019, opened 2023-12-10
fix `PlayerInteractEvent#getClickedBlock()` returns a wrong block in adventure mode.

LemonCaramel:
Fixes a bug where clicking on the SOUTH, EAST, or UP face of a block returns an incorrect block position.
This bug was introduced in #9211.


kennytv (approved):
Thank you

PR #10021, opened 2023-12-11
Catch async usage of playsound

Owen1212055:
This causes the world random to scream due to multi-thread acces

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -3309,7 +3309,7 @@ index 23bdb77690ba15bcbbfb0c70af23336d08ac7752..8f144a357174bbe096ac9b38a5e67a61
                  }
                  collection = icons;
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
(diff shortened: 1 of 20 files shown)


Machine-Maker:
Do we also need catchers for the adventure sound methods? Should be added to the adventure patch if so.

Machine-Maker (approved):
Looks good. This was expanded to catch any use of World#playSound off the main thread as none of the 3 methods are thread-safe. It does more specific checks for Server#playSound methods as there are some cases where the use is thread-safe.

kennytv (approved):

Machine-Maker:
> .playSound is just sending a packet

Very much not true on World and Server. On `Player`, playSound is as simple as sending a packet. But not on those other types. They either iterate over a thread-unsafe list of players in a world or use the vanilla packet broadcasting system based on location which contains thread-unsafe logic. 

The "world random" aspect is just one part of this. If a sound isn't provided a seed, a seed must be generated from the world's random instance which is a thread-unsafe instance.

As to your second post, I'm not sure what you are "wat"ing? THat diff is just changing some hunk line numbers and index hashes.

Machine-Maker:
> It appears the line
lighning.isSilent = isSilent
was removed, I could be mistaken though.

That's correct. That line was removed. But not as a part of this PR. That is just part of an existing patch that removes that line.
image
These 3 spots are the lines in that patch that changed as a result of this PR.

PR #10026, opened 2023-12-11
Fix long loading screen when refreshing skins

AlexProgrammerDE:
The main necessary change was to send:
```java
new ClientboundGameEventPacket(ClientboundGameEventPacket.LEVEL_CHUNKS_LOAD_START, 0.0F)
```
to do this I call `sendLevelInfo`, which is called by respawn and dimension teleport and has also this packet as a side-effect.

I also made the respawn packet code smaller.


AlexProgrammerDE:
Now it also resends the right permission level set in ops.json for the few people that actually use that "feature".

Machine-Maker (approved):

envizar:
Merge this please, this long loading screen is so annoying

Fedeasd:
it happens to me

Owen1212055 (approved):

Xemorr:
Is there a known way to completely mitigate the loading screen like pre 1.20.4 behaviour?

PR #10028, opened 2023-12-12
Fix deprecated bungee APIs still serializing to JSON in packets

jpenilla:


Machine-Maker (approved):
Absolutely no **undeprecated** API still uses this right? We def want to discourage the double conversion.

PR #10033, opened 2023-12-13
send sound and particle packets immediately even if off main

Machine-Maker:
@bergerkiller reported that on Paper specifically, smooth sounds were getting messed up and I think this is the culprit. He'll test and confirm. Paper's network optimizations include a bit that makes packets dispatched async transfer to the main thread with the reasoning that 3rd party packet listeners like from protocollib mutate world state in those listeners and this makes that safer.

I'm curious if people think we should continue to do that? Packet listeners should understand that some packets are dispatched off the main thread and they may need to sync back to the main thread to read/write world state.

---
Download the paperclip jar for this pull request: [paper-10033.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1120124594.zip)

### DIFF
--- a/patches/server/0297-Optimize-Network-Manager-and-add-advanced-packet-sup.patch
+++ b/patches/server/0297-Optimize-Network-Manager-and-add-advanced-packet-sup.patch
@@ -28,7 +28,7 @@ and then catch exceptions and close if they fire.
 Part of this commit was authored by: Spottedleaf, sandtechnology
 
 diff --git a/src/main/java/net/minecraft/network/Connection.java b/src/main/java/net/minecraft/network/Connection.java
-index 25881c890c643ce90bdcda6b094d912bafb0ed75..1931db6936773657bd43b9b16de950cb3e7a2303 100644
+index 25881c890c643ce90bdcda6b094d912bafb0ed75..93d3f29b08c79479c27d3f39e6c799c705e69902 100644
 --- a/src/main/java/net/minecraft/network/Connection.java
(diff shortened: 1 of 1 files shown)


bergerkiller:
This fixes the problem for me

Cryptite:
Could this be, dunno, toggleable? I can see lots of reasons why we'd want packets to stay async the entire time without gumming up the MT.

electronicboy:
that dispatching thing was generally done because we kinda got fed up of plugins killing the server doing stupid stuff in packet listeners; I do semi still agree with running them there, people seem to assume that they can still touch stuff because the server has some areas where mojang, in limited and isolationionary contexts, where they do

Ideally we'd ensure that both can be faciliated, but, that is moreso on plugin devs, we just wanna generally ensure that there is a safe way to do that crud, within some level of effort

bergerkiller:
Any reason why there can't be some API that protocollib itself can talk to to tell paper that a particular packet should be sync? So, as soon as someone listens for sound packets sync, it runs it on the main thread? Cant see why it cant be dynamic...

bergerkiller:
I could be wrong, but I believe theres also a sound packet to play a (registry) sound for an entity. To make a cow moo. I think the benefit was that the sound moves with the entity, unlike the normal sound packet.

Maybe that one should be included too.

Machine-Maker:
> I could be wrong, but I believe theres also a sound packet to play a (registry) sound for an entity.

I included that one already. the sound entity packet.

kennytv (approved):

PR #10036, opened 2023-12-16
Scoreboard objective number format api 

davidmayr:
This PR introduces an easily accessible number format API (using adventure styles and components) for scoreboard objectives for a feature introduced in 1.20.3. 

Setting a number format can be as easy as this:
```java
//Remove numbers
objective.numberFormat(NumberFormat.blank());

// Style numbers in green and bold
objective.numberFormat(NumberFormat.styled(NamedTextColor.GREEN, TextDecoration.BOLD));

//Replace numbers
objective.numberFormat(NumberFormat.fixed(Component.text("Lol")));
```

As for the implementation: The conversion between the Adventure style and Mojangs implementation is a bit messy.  The only other way I can think of to implement this is manually converting every style property. As paper doesn't even do that with chat components either, I have decided to just serialize and deserialize again.

### DIFF
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -99,3 +99,27 @@ index a193ffabb05160b462dee1ba8f687fdbc84405b6..3aba9fc61378421ebc2d2ab46b555fe7
 +    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
 +    //Paper end
  }
+diff --git a/src/main/java/org/bukkit/scoreboard/Score.java b/src/main/java/org/bukkit/scoreboard/Score.java
+index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..9e4bff64ebb7025e8b2af4fbdfb3e4f6883b035d 100644
+--- a/src/main/java/org/bukkit/scoreboard/Score.java
++++ b/src/main/java/org/bukkit/scoreboard/Score.java
+@@ -83,4 +83,19 @@ public interface Score {
+      */
+     void resetScore() throws IllegalStateException;
+     // Paper end
++
++    //Paper start - number format
++    /*
++     * Gets the number format for this objective's entries (scores)
++     * @return this objective's number format
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++
++    /*
++     * Gets the number format for this objective's entries (scores)
++     * @param format the number format to set
++     */
++    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
++    //Paper end
++
+ }
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -135,7 +135,7 @@ index 0000000000000000000000000000000000000000..2489f63707bce74ecd8045948de73d52
 +
 +record StyledFormatImpl(Style style) implements StyledFormat {}
 diff --git a/src/main/java/org/bukkit/scoreboard/Objective.java b/src/main/java/org/bukkit/scoreboard/Objective.java
-index a193ffabb05160b462dee1ba8f687fdbc84405b6..703b7788fbed7aff7f8274f40df3d4b6bfa30804 100644
+index a193ffabb05160b462dee1ba8f687fdbc84405b6..4f92140eb3512adce6e15b10565d7b30414912ac 100644
 --- a/src/main/java/org/bukkit/scoreboard/Objective.java
 +++ b/src/main/java/org/bukkit/scoreboard/Objective.java
 @@ -175,4 +175,20 @@ public interface Objective {
@@ -149,18 +149,18 @@ index a193ffabb05160b462dee1ba8f687fdbc84405b6..703b7788fbed7aff7f8274f40df3d4b6
 +     * @return this objective's number format
 +     * @throws IllegalStateException if this objective has been unregistered
 +     */
-+    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat();
 +
 +    /*
 +     * Sets the number format for this objective's entries (scores)
 +     * @param format the number format to set, pass null to reset format to default
 +     * @throws IllegalStateException if this objective has been unregistered
 +     */
-+    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
++    void numberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
 +    //Paper end
  }
 diff --git a/src/main/java/org/bukkit/scoreboard/Score.java b/src/main/java/org/bukkit/scoreboard/Score.java
-index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..01c250162985db9a882a30e3bb12e7a3e9246e0d 100644
+index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..340177ffb4ada77a2b09e9dce34f201771ddf839 100644
 --- a/src/main/java/org/bukkit/scoreboard/Score.java
 +++ b/src/main/java/org/bukkit/scoreboard/Score.java
 @@ -83,4 +83,23 @@ public interface Score {
@@ -175,7 +175,7 @@ index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..01c250162985db9a882a30e3bb12e7a3
 +     * @throws IllegalStateException if the associated objective has been
 +     *     unregistered
 +     */
-+    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat();
 +
 +    /*
 +     * Sets the number format for this score
@@ -183,7 +183,7 @@ index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..01c250162985db9a882a30e3bb12e7a3
 +     * @throws IllegalStateException if the associated objective has been
 +     *     unregistered
 +     */
-+    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
++    void numberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
 +    //Paper end
 +
  }
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -77,7 +77,7 @@ index 0000000000000000000000000000000000000000..73c1e5990ca8af18dcdea4bdd275ee3c
 +
 +}
 diff --git a/src/main/java/org/bukkit/scoreboard/Objective.java b/src/main/java/org/bukkit/scoreboard/Objective.java
-index a193ffabb05160b462dee1ba8f687fdbc84405b6..3aba9fc61378421ebc2d2ab46b555fe7cd99474a 100644
+index a193ffabb05160b462dee1ba8f687fdbc84405b6..ad2fe248ebbd37b22a04a12a509d4399f62daf2e 100644
 --- a/src/main/java/org/bukkit/scoreboard/Objective.java
 +++ b/src/main/java/org/bukkit/scoreboard/Objective.java
 @@ -175,4 +175,18 @@ public interface Objective {
@@ -93,14 +93,14 @@ index a193ffabb05160b462dee1ba8f687fdbc84405b6..3aba9fc61378421ebc2d2ab46b555fe7
 +    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
 +
 +    /*
-+     * Gets the number format for this objective's entries (scores)
-+     * @param format the number format to set
++     * Sets the number format for this objective's entries (scores)
++     * @param format the number format to set, pass null to reset format to default
 +     */
 +    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
 +    //Paper end
  }
 diff --git a/src/main/java/org/bukkit/scoreboard/Score.java b/src/main/java/org/bukkit/scoreboard/Score.java
-index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..9e4bff64ebb7025e8b2af4fbdfb3e4f6883b035d 100644
+index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..86f70362a8852d06a378b5dd827818d330f2a496 100644
 --- a/src/main/java/org/bukkit/scoreboard/Score.java
 +++ b/src/main/java/org/bukkit/scoreboard/Score.java
 @@ -83,4 +83,19 @@ public interface Score {
@@ -110,14 +110,14 @@ index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..9e4bff64ebb7025e8b2af4fbdfb3e4f6
 +
 +    //Paper start - number format
 +    /*
-+     * Gets the number format for this objective's entries (scores)
-+     * @return this objective's number format
++     * Gets the number format for this score
++     * @return this score's number format
 +     */
 +    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
 +
 +    /*
-+     * Gets the number format for this objective's entries (scores)
-+     * @param format the number format to set
++     * Sets the number format for this score
++     * @param format the number format to set, pass null to reset format to default
 +     */
 +    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
 +    //Paper end
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -164,7 +164,7 @@ index 0000000000000000000000000000000000000000..c1d703b8fb949d9a7ea3b78ead0f031e
 +
 +}
 diff --git a/src/main/java/org/bukkit/scoreboard/Objective.java b/src/main/java/org/bukkit/scoreboard/Objective.java
-index a193ffabb05160b462dee1ba8f687fdbc84405b6..7c7cebc06c7d16025a8a12421a79fbe50674049d 100644
+index a193ffabb05160b462dee1ba8f687fdbc84405b6..8bd2c0366181aa0260575a7564e6e0270f4a23d9 100644
 --- a/src/main/java/org/bukkit/scoreboard/Objective.java
 +++ b/src/main/java/org/bukkit/scoreboard/Objective.java
 @@ -175,4 +175,22 @@ public interface Objective {
@@ -172,9 +172,9 @@ index a193ffabb05160b462dee1ba8f687fdbc84405b6..7c7cebc06c7d16025a8a12421a79fbe5
      @NotNull Score getScoreFor(@NotNull org.bukkit.entity.Entity entity) throws IllegalArgumentException, IllegalStateException;
      // Paper end - improve scoreboard entries
 +
-+    // Paper start - number format
++    // Paper start - number format api
 +    /*
-+     * Gets the number format for this objective's entries (scores)
++     * Gets the number format for this objective's scores.
 +     *
 +     * @return this objective's number format
 +     * @throws IllegalStateException if this objective has been unregistered
@@ -182,16 +182,16 @@ index a193ffabb05160b462dee1ba8f687fdbc84405b6..7c7cebc06c7d16025a8a12421a79fbe5
 +    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat();
 +
 +    /*
-+     * Sets the number format for this objective's entries (scores)
++     * Sets the number format for this objective's scores.
 +     *
 +     * @param format the number format to set, pass null to reset format to default
 +     * @throws IllegalStateException if this objective has been unregistered
 +     */
 +    void numberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
-+    // Paper end
++    // Paper end - number format api
  }
 diff --git a/src/main/java/org/bukkit/scoreboard/Score.java b/src/main/java/org/bukkit/scoreboard/Score.java
-index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..722f6f920045f6f85b347270bbb4e08232db1b9d 100644
+index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..b1dc16b8a149f46e153b6394fcc70e38c770d61a 100644
 --- a/src/main/java/org/bukkit/scoreboard/Score.java
 +++ b/src/main/java/org/bukkit/scoreboard/Score.java
 @@ -83,4 +83,25 @@ public interface Score {
@@ -199,9 +199,9 @@ index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..722f6f920045f6f85b347270bbb4e082
      void resetScore() throws IllegalStateException;
      // Paper end
 +
-+    // Paper start - number format
++    // Paper start - number format api
 +    /*
-+     * Gets the number format for this score or null if the score has not been set yet
++     * Gets the number format for this score or null if the score has not been set yet.
 +     *
 +     * @return this score's number format
 +     * @throws IllegalStateException if the associated objective has been
@@ -210,13 +210,13 @@ index 1eaa9a93f8eff5f18a6cce2d74f21eb19db273c8..722f6f920045f6f85b347270bbb4e082
 +    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat();
 +
 +    /*
-+     * Sets the number format for this score
++     * Sets the number format for this score.
 +     *
 +     * @param format the number format to set, pass null to reset format to default
 +     * @throws IllegalStateException if the associated objective has been
 +     *     unregistered
 +     */
 +    void numberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
-+    // Paper end
++    // Paper end - number format api
 +
  }
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -6,14 +6,14 @@ Subject: [PATCH] add number format api
 
 diff --git a/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java b/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..708b931d4d4a630b0d562bc426c6cf7117b29528
+index 0000000000000000000000000000000000000000..486da6ebe0137bb3280e8b33c8e35e309507f118
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java
 @@ -0,0 +1,5 @@
 +package io.papermc.paper.scoreboard.numbers;
 +
 +record BlankFormatImpl() implements NumberFormat {
-+    public static BlankFormatImpl INSTANCE = new BlankFormatImpl();
++    public static final BlankFormatImpl INSTANCE = new BlankFormatImpl();
 +}
 diff --git a/src/main/java/io/papermc/paper/scoreboard/numbers/FixedFormat.java b/src/main/java/io/papermc/paper/scoreboard/numbers/FixedFormat.java
 new file mode 100644
--- a/patches/api/0449-add-number-format-api.patch
+++ b/patches/api/0449-add-number-format-api.patch
@@ -135,10 +135,10 @@ index 0000000000000000000000000000000000000000..2489f63707bce74ecd8045948de73d52
 +
 +record StyledFormatImpl(Style style) implements StyledFormat {}
 diff --git a/src/main/java/org/bukkit/scoreboard/Objective.java b/src/main/java/org/bukkit/scoreboard/Objective.java
-index a193ffabb05160b462dee1ba8f687fdbc84405b6..ad2fe248ebbd37b22a04a12a509d4399f62daf2e 100644
+index a193ffabb05160b462dee1ba8f687fdbc84405b6..703b7788fbed7aff7f8274f40df3d4b6bfa30804 100644
 --- a/src/main/java/org/bukkit/scoreboard/Objective.java
 +++ b/src/main/java/org/bukkit/scoreboard/Objective.java
-@@ -175,4 +175,18 @@ public interface Objective {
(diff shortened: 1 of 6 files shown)


yannicklamprecht (approved):
Like that. Yes, we're sealed know. 🥳

456dev:
In vanilla, you can set it per each score holder, right? How is that done here?

MenschenToaster:
> In vanilla, you can set it per each score holder, right? How is that done here?
> 
> Unless I'm misunderstanding what the "objective" is

I totally missed that that's possible in vanilla. You can set it for the entire objective (e.g. your entire sidebar) or apparently per entry as well (which I missed)

I will add that asap

MenschenToaster:
That's implemented now. With the latest commit, you can change the formatting of individual scores or the whole objective

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,125 @@
...
++
++    //Paper start - number format
++    /*
++     * Gets the number format for this objective's entries (scores)
++     * @return this objective's number format
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++
++    /*
++     * Gets the number format for this objective's entries (scores)
```
yannicklamprecht:
```suggestion
+     * Sets the number format for this objective's entries (scores)
```

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,125 @@
...
++     * @param format the number format to set
```
yannicklamprecht:
```suggestion
+     * @param format the number format to set, pass null to reset format to default
```

yannicklamprecht (changes requested):

patches/api/0449-add-number-format-api.patch
davidmayr: Good catch, fixed now ^^

patches/api/0449-add-number-format-api.patch
davidmayr: Added in the newest commit

jpenilla:
should use our codecs for conversions, not json, see the minecraft round trip style test in AdventureCodecsTest for an example.

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,175 @@
...
+ import net.minecraft.network.chat.MutableComponent;
++import net.minecraft.network.chat.Style;
+ 
+ public class FixedFormat implements NumberFormat {
+     public static final NumberFormatType<FixedFormat> TYPE = new NumberFormatType<FixedFormat>() {
+@@ -38,6 +39,7 @@ public class FixedFormat implements NumberFormat {
+     public MutableComponent format(int number) {
+         return this.value.copy();
+     }
++    public Component getValue() { return this.value; } //Paper - make number formats accessible
```
Machine-Maker: Use [Access Transformers](https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#access-transformers) (ATs) to make the field public instead of adding a getter.

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,175 @@
...
+diff --git a/src/main/java/net/minecraft/network/chat/numbers/StyledFormat.java b/src/main/java/net/minecraft/network/chat/numbers/StyledFormat.java
+index 6a0700e2e88ebdd91a731084a3cf83b50f768730..6c1b586477ca6f86a05a53b2fc3ef6f345d8ab0d 100644
+--- a/src/main/java/net/minecraft/network/chat/numbers/StyledFormat.java
++++ b/src/main/java/net/minecraft/network/chat/numbers/StyledFormat.java
+@@ -43,7 +43,7 @@ public class StyledFormat implements NumberFormat {
+     public MutableComponent format(int number) {
+         return Component.literal(Integer.toString(number)).withStyle(this.style);
+     }
+-
++    public Style getStyle() { return this.style; } //Paper - make number formats accessible
```
Machine-Maker: Use ATs

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,175 @@
...
++}
+diff --git a/src/main/java/net/minecraft/network/chat/numbers/FixedFormat.java b/src/main/java/net/minecraft/network/chat/numbers/FixedFormat.java
+index 2b9f393062444a663dcd25c78f9471beda65ccf0..eb0c12016a5ef1221a3d312b71c160f03124acaf 100644
+--- a/src/main/java/net/minecraft/network/chat/numbers/FixedFormat.java
++++ b/src/main/java/net/minecraft/network/chat/numbers/FixedFormat.java
+@@ -5,6 +5,7 @@ import net.minecraft.network.FriendlyByteBuf;
+ import net.minecraft.network.chat.Component;
+ import net.minecraft.network.chat.ComponentSerialization;
+ import net.minecraft.network.chat.MutableComponent;
++import net.minecraft.network.chat.Style;
```
Machine-Maker: Don't add this import, just use the fully qualified name inline where needed.

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,175 @@
...
++            com.google.gson.JsonElement json = net.minecraft.Util
++                .getOrThrow(
++                    net.minecraft.network.chat.Style.Serializer.CODEC
++                        .encodeStart(com.mojang.serialization.JsonOps.INSTANCE, styled.getStyle()),
++                    com.google.gson.JsonParseException::new
++                );
++
++            net.kyori.adventure.text.format.Style adventureStyle =
++                net.kyori.adventure.text.serializer.gson.GsonComponentSerializer.gson()
++                    .serializer().fromJson(json, net.kyori.adventure.text.format.Style.class);
```
Machine-Maker: As `jpenilla` said, use the style codec from AdventureCodec to convert between the two.

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,125 @@
...
++/*
++ * A scoreboard number format that applies a custom formatting to the score number.
++ */
++public record StyledFormat(Style style) implements NumberFormat {
++
++    public static final StyledFormat NO_STYLE = new StyledFormat(Style.empty());
++
++    public static final StyledFormat SIDEBAR_DEFAULT = new StyledFormat(Style.empty().color(NamedTextColor.RED));
++
++    public static final StyledFormat PLAYER_LIST_DEFAULT = new StyledFormat(Style.empty().color(NamedTextColor.YELLOW));
```
Machine-Maker: Tests should be added on the server side to ensure the equivalence of these fields with the vanilla fields.

Machine-Maker (changes requested):
All there records should be package-private and not exposed to the API. Instead, have interfaces with a method for each record parameter. The field for the blank format can be on the NumberFormat interface.

patches/server/1052-add-number-format-api.patch
davidmayr: That wasn't intentional and probably caused by my IDE 😅

patches/api/0449-add-number-format-api.patch
davidmayr: Good call, added

MenschenToaster:
> Should use our codecs for conversions, not the GsonComponentSerializer, see the minecraft round trip style test in AdventureCodecsTest for an example. Also put the conversion methods at PaperAdventure.asVanilla/asMinecraft to be consistent.

So just we are on the same path: I should add a asVanilla method for the **style conversion** to PaperAdventure? Or do you mean the entire number formatting? (Probably not, but just making sure)

Aaaand should I update the adventure patch? Because as far as I can tell, everything inside that class comes from that patch and if I would put it in this one it could break multiple patches if this one would be removed (e.g. temporarily while upgrading), right?

MenschenToaster:
> All there records should be package-private and not exposed to the API. Instead, have interfaces with a method for each record parameter. The field for the blank format can be on the NumberFormat interface.

Sorry, but I don't think I really get what you are asking me to do. 😅

As far as I can tell, you want me to either:
Package private the entire number format implementations and add static methods to the base interface for creation. But that would be an issue for accessing the number format value, wouldn't it? 

You could theoretically also have meant to only package private the constructors, but I don't think that's the case because private constructors don't exist in records, if I recall correctly? - sorry if I'm wrong, I'm using Kotlin most of the time 😅

Or you want me to:
Create interfaces for every type and make the implementations private (but still in the API because I don't know of any provider service to construct it?) But I don't really get the point of that? Wouldn't that just be useless class clutter?

Machine-Maker:
> Aaaand should I update the adventure patch? Because as far as I can tell, everything inside that class comes from that patch and if I would put it in this one it could break multiple patches if this one would be removed (e.g. temporarily while upgrading), right?

Yeah, you can make asVanilla/asAdventure methods inside PaperAdventure that use the codec's and they should be in the adventure patch.

> Or you want me to:
Create interfaces for every type and make the implementations private (but still in the API because I don't know of any provider service to construct it?) But I don't really get the point of that? Wouldn't that just be useless class clutter?

Yes, exactly this. interfaces for NumberFormat, StyledFormat, FixedFormat (no need for blank format, should just use an equality check to know that). And then package-private records that implement those interfaces with static methods on **NumberFormat** to create instances of each one.

There are a couple reasons for this. 1) javadocs. can have better, separated javadocs for the fluent getters and factory methods and types if you separate out the record from the API interface. 2) future changes. Changing records and making them maintain compat is super ugly if something changes in the future. If the constructor is public, you have to always keep the old one and make new ones and you have to have old components... its just not fun. Look at StructuresLocateEvent$Result and observe the reason why its not good to do that. Better to have static methods to deprecate and we can just change whatever we want in the record.

yannicklamprecht:
After some thoughts I probably would be against the nullability in the setter and just add a Reset record that is also sealed.
What do you think of @Machine-Maker, @jpenilla, @MenschenToaster ?

Machine-Maker:
If we didn't want to have nullability in the setter, I would prefer a clear method instead. I don't think adding a type for resetting is a good idea.

MenschenToaster:
> no need for blank format, should just use an equality check to know that

Internally it makes sense, but for accessibility when querying the current format it can get confusing and just more difficult to check, right?

If somebody wants to know if it's a blank format, they would have to do:
`if(!(format instanceof StyledFormat) && !(format instanceof FixedFormat)) return;`
to know if it's a blank format or not. And if there would be a new format in the future, their code would be broken

Machine-Maker:
By "equality check" I meant you can do
```java
if (NumberFormat.blank().equals(format)) {}
```

MenschenToaster:
> By "equality check" I meant you can do
> 
> ```java
> if (NumberFormat.blank().equals(format)) {}
> ```

That makes way more sense. Sure, then I'll do that

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,194 @@
...
++/*
++ * A scoreboard number format that replaces the score number with a chat component.
++ */
++public interface FixedFormat extends NumberFormat {
++
++    /*
++     * The component shown instead of the number for a score
++     * @return the chat component
++     */
++    Component getComponent();
```
Machine-Maker: name it `component` so you don't have to add a method to the record.

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,194 @@
...
++
++    StyledFormat SIDEBAR_DEFAULT = NumberFormat.styled(Style.empty().color(NamedTextColor.RED));
++
++    StyledFormat PLAYER_LIST_DEFAULT = NumberFormat.styled(Style.empty().color(NamedTextColor.YELLOW));
++
++    /*
++     * The style that is being applied to the number in the score
++     * @return the style to apply
++     */
++    Style getStyle();
```
Machine-Maker: Name it `style` so you don't need the method on the record

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,194 @@
...
++/*
++ * A scoreboard number format that applies a custom formatting to the score number.
++ */
++public interface StyledFormat extends NumberFormat {
++
++    StyledFormat NO_STYLE = NumberFormat.styled(Style.empty());
++
++    StyledFormat SIDEBAR_DEFAULT = NumberFormat.styled(Style.empty().color(NamedTextColor.RED));
++
++    StyledFormat PLAYER_LIST_DEFAULT = NumberFormat.styled(Style.empty().color(NamedTextColor.YELLOW));
```
Machine-Maker: I think we should make these static methods that return these constants that can then be declared inside the package private record. similar to the blank instance.

patches/server/0010-Adventure.patch
```diff
@@ -1621,6 +1623,24 @@ index 0000000000000000000000000000000000000000..c10c0ffa29332d73328f088935a0b914
...
++            .parse(net.minecraft.nbt.NbtOps.INSTANCE, encoded), IllegalStateException::new);
++    }
++
++    public static Style asAdventure(net.minecraft.network.chat.Style style) {
++        net.minecraft.nbt.Tag encoded = Util.getOrThrow(net.minecraft.network.chat.Style.Serializer.CODEC
++            .encodeStart(net.minecraft.nbt.NbtOps.INSTANCE, style), IllegalStateException::new);
++
++        return Util.getOrThrow(AdventureCodecs.STYLE_MAP_CODEC.codec()
++            .parse(net.minecraft.nbt.NbtOps.INSTANCE, encoded), IllegalStateException::new);
++    }
```
Machine-Maker: I think we are preferring to use JavaOps.INSTANCE for this instead of NbtOps.INSTANCE, as it just converts to a native java map instead of an nbt tag.

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,162 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
++++ b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
+@@ -152,6 +152,30 @@ final class CraftObjective extends CraftScoreboardComponent implements Objective
+     }
+     // Paper end
+ 
++    //Paper start - add number format
++    @Override
++    public io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat() {
++        net.minecraft.network.chat.numbers.NumberFormat vanilla = this.objective.numberFormat();
```
Machine-Maker: This should have a `this.checkState` at the top to ensure the objective is still registered. Those exist on the other getters like getRenderType and getDisplaySlot

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,162 @@
...
++            return null;
++        }
++
++        return io.papermc.paper.util.PaperScoreboardFormat.asPaper(vanilla);
++    }
++
++
++    @Override
++    public void setNumberFormat(io.papermc.paper.scoreboard.numbers.NumberFormat format) {
++        if(format == null) {
```
Machine-Maker: check state

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,194 @@
...
++     * @return this score's number format
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++
++    /*
++     * Sets the number format for this score
++     * @param format the number format to set, pass null to reset format to default
++     */
++    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
++    //Paper end
```
Machine-Maker: Each of these should have `@throws IllegalStateException if this objective has been unregsitered`.

Machine-Maker (changes requested):

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,162 @@
...
++        }
++
++        return io.papermc.paper.util.PaperScoreboardFormat.asPaper(vanilla);
++    }
++
++
++    @Override
++    public void setNumberFormat(io.papermc.paper.scoreboard.numbers.NumberFormat format) {
++        final net.minecraft.world.scores.ScoreAccess access = this.objective.checkState()
++            .board.getOrCreatePlayerScore(this.entry, this.objective.getHandle());
```
Machine-Maker: Looking into the logic further... This seems not ideal. What do we expect setNumberFormat to do on a score that doesn't exist? Is it right that it creates a score for this entry? I'm not sure it is. One possible solution is to throw an exception on these setters/getters if the score isn't set, can link to Score#isScoreSet to check that first or something. 

Curious for opinions.

patches/api/0449-add-number-format-api.patch
davidmayr: I'm kind of against that. Having it like that shows the implementation directly to the devs that use this api. 

Although these constants might change in order to stay consistent with Mojangs implementation, I don't think they ever will. So I'd prefer people to see what it actually does

Machine-Maker: They can still see them if they goto the source file. Package-private api source is still visible in the IDE or whatever, there isn't an issue with visibility. I think it is just cleaner if its in the impl instead of the interface. 

Incidentally that's another reason for the interface+record pattern is keeping the interface as straightforward as possible.

davidmayr: Well, you're right, but I just think it's a bit easier to follow along when it's in the interface itself

patches/server/1052-add-number-format-api.patch
davidmayr: I think throwing an exception is the best solution. Since that happens when the objective is unregistered as well

Machine-Maker: So my one thing against that is that the /scoreboard command does create an empty score if one doesn't exist if you only change the number format.

davidmayr: Sooo we should just leave it like I have it here? How about querying? Do we make that only accessible if there is a score? (so we throw instead of returning null?)

patches/api/0449-add-number-format-api.patch
lynxplay: If those values ever end up depending on internals, it is a lot easier for us to change things around if the values are exposed via methods vs fields.

Grabbing a different method result (via e.g. ServiceProvider) is a lot nicer than redefining these fields to a result of a ServiceProvider provided instance's method (e.g. `ServiceProvider.get(ScoreboardStyles.class).blablabla()` )

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,185 @@
...
++++ b/src/main/java/org/bukkit/scoreboard/Objective.java
+@@ -175,4 +175,20 @@ public interface Objective {
+      */
+     @NotNull Score getScoreFor(@NotNull org.bukkit.entity.Entity entity) throws IllegalArgumentException, IllegalStateException;
+     // Paper end - improve scoreboard entries
++
++    //Paper start - number format
++    /*
++     * Gets the number format for this objective's entries (scores)
++     * @return this objective's number format
```
lynxplay: Might be worth to comment the current default of this value (just the number) in the javadocs.

lynxplay: Might need more comments on the `@Nullable` vs `@NotNull` given we expose NO_STYLE in the API now.

patches/server/1052-add-number-format-api.patch
Machine-Maker: Ok, so I think the best solution is as follows...
Setters on Score should add the score if it doesn't exist. If we don't do that, we introduce behavior where calling the same methods in different orders behaves differently (setScore, setNumberFormat vs setNumberFormat, setScore).
Getters on Score should not add the score, but return some default value, null, false, 0, the default values for that type. 

And this behavior can be documented in the javadocs

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,185 @@
...
++     * @throws IllegalStateException if this objective has been unregistered
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++
++    /*
++     * Sets the number format for this objective's entries (scores)
++     * @param format the number format to set, pass null to reset format to default
++     * @throws IllegalStateException if this objective has been unregistered
++     */
++    void setNumberFormat(@Nullable io.papermc.paper.scoreboard.numbers.NumberFormat format);
```
lynxplay: I think we are better off sticking to fluent naming here.
While spigot has not released their API yet, potentially clashing with them seems like a thing we might want to avoid.

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,162 @@
...
++}
+diff --git a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
+index 3157f3d2f9ce7af4a763203672817a7f5c7bd4fb..6fde24cb6427f6edacaf372ffb72514d393c75b2 100644
+--- a/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
++++ b/src/main/java/org/bukkit/craftbukkit/scoreboard/CraftObjective.java
+@@ -152,6 +152,34 @@ final class CraftObjective extends CraftScoreboardComponent implements Objective
+     }
+     // Paper end
+ 
++    //Paper start - add number format
```
Lulu13022002: Missing a space between // and Paper, applies for other comments as well

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,162 @@
...
++    @Override
++    public io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat() {
++        this.checkState();
++
++        net.minecraft.network.chat.numbers.NumberFormat vanilla = this.objective.numberFormat();
++        
++        if(vanilla == null) {
```
Lulu13022002: Missing a space between `if` and `(`, same for other conditions

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,185 @@
...
++    /*
++     * Gets the number format for this objective's entries (scores)
++     * @return this objective's number format
++     * @throws IllegalStateException if this objective has been unregistered
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat getNumberFormat();
++
++    /*
++     * Sets the number format for this objective's entries (scores)
++     * @param format the number format to set, pass null to reset format to default
```
Lulu13022002: Missing a new line between the description and the parameter/return explanation, same for others

patches/api/0449-add-number-format-api.patch
davidmayr: > I think we are better off sticking to fluent naming here. While spigot has not released their API yet, potentially clashing with them seems like a thing we might want to avoid.

With fluent naming, you mean numberFormat() as a getter and numberFormat(param) as a setter, like in the adventure APIs?

yannicklamprecht: Jup, that is fluent naming.

patches/server/1052-add-number-format-api.patch
davidmayr: So basically leave it like I already had it? I implemented a fix (pushing shortly) because it would throw a npe if the score had not been registered, but other than that, what you're suggesting is what it's already like?

patches/api/0449-add-number-format-api.patch
davidmayr: Should I add these methods to the StyledFormat interface, or would you guys prefer to have that in the number format interface along the other methods?

davidmayr: > Should I add these methods to the StyledFormat interface, or would you guys prefer to have that in the number format interface along the other methods?

@Machine-Maker

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,195 @@
...
+diff --git a/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java b/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..708b931d4d4a630b0d562bc426c6cf7117b29528
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/scoreboard/numbers/BlankFormatImpl.java
+@@ -0,0 +1,5 @@
++package io.papermc.paper.scoreboard.numbers;
++
++record BlankFormatImpl() implements NumberFormat {
++    public static BlankFormatImpl INSTANCE = new BlankFormatImpl();
```
kashike:
```suggestion
+    public static final BlankFormatImpl INSTANCE = new BlankFormatImpl();
```

patches/api/0449-add-number-format-api.patch
davidmayr: > Might be worth to comment the current default of this value (just the number) in the javadocs.

Well in server terms there is no default value as far as I can tell. That is fully up to the client what to style that like. It would be hard to describe how it would look like in game, since objectives and their scores can be more than just the scoreboard sidebar...

I would suggest a wording like that maybe?
For score (getter):
`The number format might be null if the score has not been set yet or when the objective default (if set) 
or the client default number format is used`
For score (setter):
`@param format the number format to set, pass null to reset the format to the objective default (if set) or to the client default`

Objective(getter): 
`@return this objective's number format or null if the client default is being used`
Objective(setter):
`@param format the number format to set, pass null to reset format to default`

patches/api/0449-add-number-format-api.patch
davidmayr: I added them to the styled format interface for now. Let me know if I should change it

patches/api/0449-add-number-format-api.patch
davidmayr: Oops, fixed now

patches/api/0449-add-number-format-api.patch
Machine-Maker: I think we stick with nullable getter/setter. I don't think we wanna make the assumption that noStyle StyledFormat will always be the same as the default format because @MenschenToaster is right, that is a client implementation detail.

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,222 @@
...
+@@ -83,4 +83,25 @@ public interface Score {
+      */
+     void resetScore() throws IllegalStateException;
+     // Paper end
++
++    // Paper start - number format
++    /*
++     * Gets the number format for this score or null if the score has not been set yet
++     *
++     * @return this score's number format
```
Machine-Maker: change to smth like
`@return this score's number format, or null if not set or the score doesn't exist

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,222 @@
...
++     * @throws IllegalStateException if the associated objective has been
++     *     unregistered
++     */
++    @Nullable io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat();
++
++    /*
++     * Sets the number format for this score
```
Machine-Maker: Add `.`.

Generally the main block of javadoc should be proper english with the param/return/throws tags allowing phrases

patches/server/1052-add-number-format-api.patch
```diff
@@ -0,0 +1,171 @@
...
++            if (scoreInfo != null) {
++                net.minecraft.network.chat.numbers.NumberFormat vanilla = scoreInfo.numberFormat();
++
++                if (vanilla == null) {
++                    return null;
++                }
++
++                return io.papermc.paper.util.PaperScoreboardFormat.asPaper(vanilla);
++            }
++        }
```
Machine-Maker: This logic is wrong and is actually a bug fixed in another PR. Specifically the part about doing a `contains` check on `getTrackedPlayers`. Just use the null returned by getPlayerScoreInfo as a way of knowing if the score exists.

patches/server/1052-add-number-format-api.patch
Machine-Maker: Yes, this is now correct.

Machine-Maker (changes requested):

patches/api/0449-add-number-format-api.patch
davidmayr: I changed it now, but I also incorporated the other possible reasons for it being null, like @lynxplay mentioned in his review before. Let me know if this is good or if I need to make changes

lynxplay (approved):
Requires testing and javadoc addition regarding the creation of the score on the numberFormat setter, otherwise LGTM.

Given upstream has yet to release even a PR for this, I think it is fine to move forwards with this sooner than later 👍

MenschenToaster:
> Requires testing

Did some basic testing. I only tested the number format on the score and not on objectives. But the original code in my first commit was tested on the objective in another plugin as well. (Because that code is used in a paper fork of mine as of right now) I would assume that works.

Getting (and therefor converting the format back) works as well (which can be sen by the output in the console)
Some further testing might be required but as far as I can see everything works as intended

grafik

MenschenToaster:
Any updates?

yannicklamprecht (approved):

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,224 @@
...
++++ b/src/main/java/io/papermc/paper/scoreboard/numbers/FixedFormat.java
+@@ -0,0 +1,17 @@
++package io.papermc.paper.scoreboard.numbers;
++
++import net.kyori.adventure.text.Component;
++
++/*
++ * A scoreboard number format that replaces the score number with a chat component.
++ */
++public interface FixedFormat extends NumberFormat {
```
kezz: Can implement `ComponentLike`

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,224 @@
...
++        return new StyledFormatImpl(style);
++    }
++
++    /*
++     * Creates a scoreboard number format that replaces the score number with a chat component.
++     *
++     * @param component the component to replace the number with
++     * @return a fixed number format
++     */
++    static FixedFormat fixed(@NotNull Component component) {
```
kezz: Might be nice to have a `ComponentLike` overload

patches/api/0449-add-number-format-api.patch
```diff
@@ -0,0 +1,224 @@
...
+@@ -0,0 +1,45 @@
++package io.papermc.paper.scoreboard.numbers;
++
++import net.kyori.adventure.text.format.NamedTextColor;
++import net.kyori.adventure.text.format.Style;
++
++/*
++ * A scoreboard number format that applies a custom formatting to the score number.
++ */
++public interface StyledFormat extends NumberFormat {
```
kezz: Can implement `StyleBuilderApplicable`

kezz (approved):
Lovely PR, left a few minor additions that could make it a bit more adventure, but otherwise very nice

lynxplay:
Upstream has proposed https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/1344/overview on upstream, it is probably best to hold out for that.

lynxplay (approved):

Machine-Maker (approved):

Cryptite:
Might be nice to update the OP which provides examples of the API.

For example, fixed number format is now `objective.numberFormat(NumberFormat.fixed(Component.text("Lol"));`

lynxplay:
Done :+1:

MenschenToaster:
> Done 👍
We both edited the description at the same time 🙃
Didn't even know maintainers could edit the PRS description 😅

PR #10037, opened 2023-12-16
Fix and add new scoreboard API

Machine-Maker:
Following up on https://github.com/PaperMC/Paper/pull/10036 (which adds API for the new number format stuff), this adds API for per-score display names and trigger locking as well as per-objective auto updating display names.

This also fixes several critical issues with the scoreboard API caused by the 1.20.3/4 update. The new ScoreHolder type threw off logic expectations in the craft-impls.

### DIFF
--- a/patches/api/0450-Add-more-scoreboard-API.patch
+++ b/patches/api/0450-Add-more-scoreboard-API.patch
@@ -1,0 +1,90 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 16 Dec 2023 14:45:46 -0800
+Subject: [PATCH] Add more scoreboard API
+
+
+diff --git a/src/main/java/org/bukkit/scoreboard/Objective.java b/src/main/java/org/bukkit/scoreboard/Objective.java
+index a193ffabb05160b462dee1ba8f687fdbc84405b6..bd4d84cbf220ab02f09ece97873bbf0bdf7a45ba 100644
+--- a/src/main/java/org/bukkit/scoreboard/Objective.java
++++ b/src/main/java/org/bukkit/scoreboard/Objective.java
+@@ -175,4 +175,24 @@ public interface Objective {
+      */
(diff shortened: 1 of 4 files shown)


Machine-Maker:
I'm not confident in a few of the method names, the display auto updating and entry custom name ones.

I'm also not confident in the approach currently implemented for handling setting locked or custom name state on a score if the score doesn't exist. Right now it prevents creating an empty score, but that's not what vanilla commands do. The vanilla scoreboard commands will just create a blank one if you set a custom name or number format without first setting a score value.

patches/api/0449-Add-more-scoreboard-API.patch
```diff
@@ -0,0 +1,86 @@
...
++     */
++    void setLocked(boolean locked);
++    /**
++     * Get the custom name for this entry.
++     *
++     * @return the custom name or null if not set
++     * @throws IllegalStateException if the score isn't set
++     * @throws IllegalStateException if the associated objective has been unregistered
++     */
++    @Nullable net.kyori.adventure.text.Component entryCustomName();
```
kennytv: The name is a bit weird, what about just `displayName` or `customName` to also stay consistent with other api?

patches/api/0449-Add-more-scoreboard-API.patch
```diff
@@ -0,0 +1,86 @@
...
++     * Sets if this score is locked and can
++     * be used by the {@code /trigger} command
++     * executed by the owner of this score.
++     *
++     * @param locked true to lock, false to unlock
++     * @throws IllegalStateException if the score isn't set
++     * @throws IllegalStateException if the associated objective has been unregistered
++     */
++    void setLocked(boolean locked);
++    /**
```
kennytv: Empty line

kennytv (approved):
Otherwise good. Behavior doesn't need to reflect commands, so I think the current behavior makes sure people don't accidentally do something they possibly don't want

patches/api/0449-Add-more-scoreboard-API.patch
Machine-Maker: Went with "customName"

kennytv (approved):

PR #10045, opened 2023-12-18
Adventure 4.15

Machine-Maker:
```[tasklist]
- [x] removeResourcePacks(UUID, UUID...)
- [x] clearResourcePacks
- [x] sendResourcePacks(ResourcePackRequest)
- [x] request callbacks
- [x] Translator#hasAnyTranslations (improve perf for rendering translatables)
- [ ] mark other send methods on Player as Obsolete/Deprecated?
- [x] bundle packet
- [x] ~~options for default serializers~~not required, old way still works
- [x] TranslationArgument migration
```


Machine-Maker:
So we need to decide what happens if someone registers a callback for a UUID that already has one registered, and when exactly those callbacks are fired.

kennytv:
I dont think the others need marking, they stand fine as convenience methods for now. The callbacks should be called before bukkit events

Machine-Maker:
Ok, removed all the TODO comments added and left comments explaining those decisions. Should be all ready.

kennytv (approved):

PR #10047, opened 2023-12-19
add itemmetas for Paintings and ItemFrames

456dev:
in progress.
I have also added a fromIndex method to Rotation, lmk if i should revert it / split it into another PR or smth.

since this is my first time delving into itemmeta, please tell me if I'm doing something wrong / miss anything

there are still TODO comments littered around for me to fix, but I might as well create the PR now for feedback
I wanted to put the CraftMetas in a paper package, but the need to extend the package private base class

is it possible to use inheritance with the CraftItemMetas?
ie: if i wanted to have the TileXYZ on a parent class, since the implementation is the same for both.

i would assume so, but noticed similar duplicate handling with the entitytag, and wanted to check first incas there was a reason

### DIFF
--- a/patches/api/0449-Move-index-rotation-conversion-to-the-rotation-enum.patch
+++ b/patches/api/0449-Move-index-rotation-conversion-to-the-rotation-enum.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Isaac - The456 <the456@the456gamer.dev>
+Date: Tue, 19 Dec 2023 00:19:05 +0000
+Subject: [PATCH] Move index - rotation conversion to the rotation enum
+
+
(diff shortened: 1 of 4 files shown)


Owen1212055:
Correct, this is meant to be covered by a sort of “entity” property api.

456dev:
ok, should i stop work on this then?
ik there are no etas, but it seems a little while off, so some api might be better than none,
esp since from what i read [here](https://github.com/PaperMC/Paper/pull/8711#issue-1507269666) it doesnt replace itemmeta

Owen1212055:
Yeah so, we are most likely going to close this here. In general, we really want to prevent adding onto the item meta system due to what was specified above, but also because it's tricky because introducing new types introduces some serialization differences between spigot/paper that we don't want to run into now.

Thank you for your PR regardless. 😄

456dev:
No worries!

PR #10060, opened 2023-12-20
Bytecode Modification Framework

Machine-Maker:
Groundwork for a better-designed bytecode modification framework for the larger structural changes we need to do. The utilities currently implemented are what was needed to split Position out of Location and manage return types and param types. 

This framework includes generated classes with generated static methods and it's done all auto-magically. It is also versioned. It changes the `api-version` field for paper plugins to be specific to a minor version and won't run modifications not applicable for that version.

### DIFF
--- a/patches/server/1056-Bytecode-Modification-Framework.patch
+++ b/patches/server/1056-Bytecode-Modification-Framework.patch
@@ -1,0 +1,175 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 12 Nov 2023 16:56:03 -0800
+Subject: [PATCH] Bytecode Modification Framework
+
+
+diff --git a/build.gradle.kts b/build.gradle.kts
+index 3588770a9ea6ee0a9508b218758650f43d994715..d9f18b872e906290ea9cfe7c6ef5b508561f70f7 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -56,10 +56,11 @@ dependencies {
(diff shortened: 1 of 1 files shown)


patches/server/1052-Bytecode-Modification-Framework.patch
```diff
@@ -0,0 +1,1252 @@
...
++
++    @Override
++    public @Nullable String superClassName() {
++        return this.superClassName;
++    }
++
++    protected static Class<?> defineGeneratedDelegates(final Class<? extends VersionedClassloaderBytecodeModifier> versionedClass, final RewriteRule finalRule) {
++        final String className = versionedClass.getName() + "_GeneratedDelegates";
++        final ClassWriter writer = new ClassWriter(ClassWriter.COMPUTE_FRAMES | ClassWriter.COMPUTE_MAXS);
++        writer.visit(Opcodes.V1_8, Opcodes.ACC_PUBLIC, className.replace('.', '/'), null, Type.getInternalName(Object.class), null);
```
MiniDigger: no asm expert, but why 1.8 and not 17?

Machine-Maker: I just copied that from our `ASMEventExecutorGenerator` when I was still figuring out how to generate classes. We can change it if we want. Idk if it matters or not.

MiniDigger: idk if this would break if you use records or whatever

Machine-Maker:
Ok, I think this in a place where it covers most of our immediate needs. It can be expanded for more expansive tasks if needed. Just see the 2 PRs targeting this branch for how it's used, https://github.com/PaperMC/Paper/pull/10061 and https://github.com/PaperMC/Paper/pull/10062.

PR #10061, opened 2023-12-20
Split Position from Location type hierarchy

Machine-Maker:
This is how the bytecode modification framework from https://github.com/PaperMC/Paper/pull/10060 is used to ~~replace some parameter types with Position and~~ make Location not implement Position anymore. The test plugin is setup to compile against current paper api and show that there are no compat breaks.

### DIFF
--- a/patches/api/0487-Split-Position-from-Location-type-hierarchy.patch
+++ b/patches/api/0487-Split-Position-from-Location-type-hierarchy.patch
@@ -1,0 +1,51 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Tue, 19 Dec 2023 23:39:49 -0800
+Subject: [PATCH] Split Position from Location type hierarchy
+
+
+diff --git a/src/main/java/org/bukkit/Location.java b/src/main/java/org/bukkit/Location.java
+index 65618b6b3c950fb27707f243a766511d6cd3aab4..0c82941c045ddede59da69f865f27b3e4a0f5a12 100644
+--- a/src/main/java/org/bukkit/Location.java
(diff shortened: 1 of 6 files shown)


patches/api/0449-Split-Position-from-Location-type-hierarchy.patch
```diff
@@ -0,0 +1,126 @@
...
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index 567a36a4887da8994c9170e2885aa8cc357efa0c..f678d4d3b91a8ab592580284bb9ceccfb4551b89 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -119,7 +119,7 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+      * @return Block at the given location
+      */
+     @NotNull
+-    public Block getBlockAt(@NotNull Location location);
++    public Block getBlockAt(@NotNull io.papermc.paper.math.Position location);
```
MiniDigger: all the param names should be adjusted too, or did you skip them to minimize diff (since you need to touch javadocs too)
also, some of them are paper methods, is it worth folding that back to the OG patch? or do we wait till softspoon?

Machine-Maker: I just chose some random methods to demo this for. So I didn't go the full distance and rename params or javadocs.

We could start by just doing all methods that paper adds, but we really want to do it in large chunks so that there isn't a point where half the codebase is using one and not the other. That gets annoying real quick.

Machine-Maker:
So I did a bit more converting methods and as you can see the diff has grown quite a bit. I'm starting to get hesitant to do this for more than methods with Location that we add. 

One downside of not doing it all at once is it makes it slightly more annoying to work with the API before its all converted since its 2 different types.

PR #10066, opened 2023-12-21
Add Registry#getKey

Machine-Maker:
Some types that have registries support existing without being in a registry and therefore have no key. This is part one of supporting that, adding a Registry#getKey(Object) method to be able to get the key for an object so we can remove Keyed from such objects.

This also deprecates Keyed#getKey on TrimMaterial and TrimPattern as these are the 2 offending types currently in the api.

----

Part 1 of https://github.com/PaperMC/Paper/pull/10062

### DIFF
--- a/patches/api/0451-Improve-Registry.patch
+++ b/patches/api/0451-Improve-Registry.patch
@@ -1,0 +1,114 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 20 Dec 2023 02:03:10 -0800
+Subject: [PATCH] Improve Registry
+
+Adds Registry#getKey(Object) which should be the
+primary way people get the key for an object. Registry
+items need to exist without having a key and so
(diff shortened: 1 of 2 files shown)


patches/api/0449-Improve-Registry.patch
```diff
@@ -0,0 +1,117 @@
...
+     Registry<LootTables> LOOT_TABLES = new SimpleRegistry<>(LootTables.class);
+     /**
+      * Server materials.
+@@ -307,6 +313,22 @@ public interface Registry<T extends Keyed> extends Iterable<T> {
+     @Nullable
+     T get(@NotNull NamespacedKey key);
+ 
++    // Paper start - improve Registry
++    /**
++     * Get the key for this object.
```
kennytv: Should probably clarify here that getting null would be the exception, and give an example

Machine-Maker: I'm also thinking there might be a use for a getKeyOrThrow method? Sometimes I hate nullable methods when I **know** it exists.

Could also make getKey return an optional, idk.

kennytv (approved):

PR #5678, opened 2021-05-18
Add drops to shear events

Machine-Maker:
Closes #5672

PR #10071, opened 2023-12-23
Enhanced PlayerShearEntityEvent: Customizable Drops

twisti-dev:
Introducing a modification to PlayerShearEntityEvent that allows customization of drops from sheared entities.


Warriorrrr:
Thank you for your contribution, unfortunately this is a duplicate of the already existing PR #5678, so I'm going to have to close this in favour of that one.

PR #10076, opened 2023-12-24
Fix MC-111729

viciscat:
The more I worked on this, the sillier it became. So this is a draft for now.

The point of this PR is to fix [MC-111729](https://bugs.mojang.com/browse/MC-111729) (and a remap issue, doesn't affect the pr so is removable).
Did some testing, works when creating the objective with the scoreboard command and with a plugin.

The reason this is a draft is I really don't know how to deal with fake players created by plugins, or just big servers with a lot of players. I think having the command do it all the time is fine but maybe add a boolean to the methods that registers a new objective to enable or disable the init of scores? I dunno.

This bug isn't really a big issue but would still be nice to fix it!


viciscat:
I'm just gonna close this. No intentions to continue it.

PR #10088, opened 2023-12-28
full support for UnixDomainSocket

Coloryr:
full support for UnixDomainSocketAddress
**It must work with [this](https://github.com/PaperMC/Velocity/pull/1173)**
It change the api, but retained original support
{4E839E59-23CF-4fa3-B0F9-2C75F679304E}
image

```
server-ip=unix\:/tmp/paperserver.sock
```

```
[servers]
# Configure your servers here. Each key represents the server's name, and the value
# represents the IP address of the server to connect to.
lobby = "unix:/tmp/paperserver.sock"
```

### DIFF
--- a/patches/api/0453-full-support-for-UnixDomainSocketAddress.patch
+++ b/patches/api/0453-full-support-for-UnixDomainSocketAddress.patch
@@ -1,0 +1,639 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Color_yr <402067010@qq.com>
+Date: Thu, 28 Dec 2023 20:21:48 +0800
+Subject: [PATCH] full support for UnixDomainSocketAddress
+
+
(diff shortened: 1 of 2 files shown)


electronicboy:
except it doesn't retain original support, it just means that that stuff will blow up, and so, this cannot be merged until a major

also, patches generally should not modify earlier patches

Owen1212055:
Seems to dupe ? https://github.com/PaperMC/Paper/pull/9086

Coloryr:
> except it doesn't retain original support, it just means that that stuff will blow up, and so, this cannot be merged until a major
> 
> also, patches generally should not modify earlier patches

No, if UDS is not used, the original API is supported
And I need to delete the patch before the modification?

Coloryr:
> Seems to dupe ? #9086

The modified content is the same, but the modification method and API processing are different

Coloryr:
My modification is for the `group server`

PR #10090, opened 2023-12-28
Fix some component bugs in login disconnect packet

LemonCaramel:
Fixes an issue in `ClientboundLoginDisconnectPacket` where Adventure translate components are not rendered
and Serialized components having a maximum length of 32767.

TestPlugin.java
```java

    @Override
    public void onEnable() {
        final GlobalTranslator translator = GlobalTranslator.translator();
        final TranslationRegistry registry = TranslationRegistry.create(Key.key("paper", "test-plugin"));
        registry.register("paper.kittens", Locale.US, new MessageFormat("Kittens!!!!!!"));
        translator.addSource(registry);

    }

    @EventHandler
    void onAsyncPlayerPreLogin(final AsyncPlayerPreLoginEvent event) {
        // 35096 lengths (Translated: 33596 lengths)
        event.disallow(AsyncPlayerPreLoginEvent.Result.KICK_OTHER, MiniMessage.miniMessage().deserialize(
            "<rainbow>" + "<click:run_command:kittens><lang:paper.kittens> ".repeat(300)
        ));
    }
}
```

### DIFF
--- a/patches/api/0006-Adventure.patch
+++ b/patches/api/0006-Adventure.patch
@@ -3227,10 +3227,21 @@ index 9c68c3f2d61500479f48b80264f625aaae2f3204..399afcd19fcb6acd24857ed6ab48cf0d
      private static final HandlerList handlers = new HandlerList();
      private boolean cancel = false;
 diff --git a/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java b/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java
-index 9866c07c999f46cb585709804aaad710c3031d5a..c7c45e2de8cca1bf8b8e12752e08db62403efa6a 100644
+index 9866c07c999f46cb585709804aaad710c3031d5a..3432872303aa0df97b5d9090fe98b269ef3cb9f4 100644
 --- a/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java
 +++ b/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java
-@@ -14,7 +14,7 @@ import org.jetbrains.annotations.NotNull;
+@@ -10,11 +10,18 @@ import org.jetbrains.annotations.NotNull;
+  * Stores details for players attempting to log in.
+  * <p>
+  * This event is asynchronous, and not run using main thread.
++ * <p>
++ * When this event is fired, the player's locale is not
++ * available. Therefore, any translatable component will be
++ * rendered with the default locale, {@link java.util.Locale#US}.
++ * <p>
++ * Consider rendering any translatable yourself with {@link net.kyori.adventure.translation.GlobalTranslator#render}
++ * if the client's language is known.
+  */
  public class AsyncPlayerPreLoginEvent extends Event {
      private static final HandlerList handlers = new HandlerList();
      private Result result;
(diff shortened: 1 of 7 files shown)


patches/server/0010-Adventure.patch
```diff
@@ -2605,6 +2605,26 @@ index 5a541e14caccaca97759879fd081becd0e8c6af2..13859822feebb65f14973aa688a54fcb
...
+--- a/src/main/java/net/minecraft/network/protocol/login/ClientboundLoginDisconnectPacket.java
++++ b/src/main/java/net/minecraft/network/protocol/login/ClientboundLoginDisconnectPacket.java
+@@ -17,7 +17,14 @@ public class ClientboundLoginDisconnectPacket implements Packet<ClientLoginPacke
+ 
+     @Override
+     public void write(FriendlyByteBuf buf) {
+-        buf.writeUtf(Component.Serializer.toJson(this.reason));
++        // Paper start - Adventure
++        if (this.reason instanceof io.papermc.paper.adventure.AdventureComponent adv) {
++            buf.writeUtf(io.papermc.paper.adventure.PaperAdventure.asJsonString(adv.adventure$component(), buf.adventure$locale), FriendlyByteBuf.MAX_COMPONENT_STRING_LENGTH);
```
electronicboy: I would generally like to note that afaict, that locale field will always be null

MiniDigger: is that still the case with the config phase being added? I dont have an overview anymore at what point that event is fired

LemonCaramel: > I would generally like to note that afaict, that locale field will always be null

In the login phase locale field is always null, but rendered with the default locale.
Would it be better to change the locale argument to null? (since the result will be the same anyway)

electronicboy: > is that still the case with the config phase being added?

Yes, because this packet only exists within login

I was earlier thinking that i'd be somewhat inclined to explictly pass null and leave a note as to why, headache comes into the expectations of events and such using this, maybe be inclined to add a javadoc there, but 🤷

LemonCaramel: Okay, I added Javadoc in API and changed the locale argument to null.

patches/api/0006-Adventure.patch
```diff
@@ -3255,11 +3255,13 @@ index 9866c07c999f46cb585709804aaad710c3031d5a..c7c45e2de8cca1bf8b8e12752e08db62
...
-      *
-      * @param message New kick message
-      */
--    public void setKickMessage(@NotNull final String message) {
++    /**
++     * Sets the kick message to display if getResult() != Result.ALLOWED
++     *
++     * @param message New kick message. At the point when this event is fired,
++     *     the server does not have the player's locale, so the translate component
++     *     will render with default locale({@link java.util.Locale#US}).
```
Machine-Maker: The longer description should be in the main body of the javadoc.

Also it should be, 
```
When this event is fired, the player's locale is not
available. Therefore, any translatable component will be
rendered with the default locale, {@link java.util.Locale#US}.
<p>
Consider rendering any translatable yourself with {@link GlobalTranslator#render}
if the client's language is known.
```

patches/server/0010-Adventure.patch
```diff
@@ -2605,6 +2605,27 @@ index 5a541e14caccaca97759879fd081becd0e8c6af2..13859822feebb65f14973aa688a54fcb
...
+-        buf.writeUtf(Component.Serializer.toJson(this.reason));
++        // Paper start - Adventure
++        if (this.reason instanceof io.papermc.paper.adventure.AdventureComponent adv) {
++            // In the login phase, buf.adventure$locale field is always null
++            buf.writeUtf(io.papermc.paper.adventure.PaperAdventure.asJsonString(adv.adventure$component(), null), FriendlyByteBuf.MAX_COMPONENT_STRING_LENGTH);
++            return;
++        }
++        // Paper end - Adventure
++
++        buf.writeUtf(Component.Serializer.toJson(this.reason), FriendlyByteBuf.MAX_COMPONENT_STRING_LENGTH); // Paper - Adventure (Fix vanilla serialized component length)
```
Machine-Maker: This also needs to render non-adventure components. We started doing that elsewhere and I think its probably good to cover all the bases. 

See ComponentSerialization#LOCALIZED_CODECS and just use the Locale.US one from there.

Machine-Maker (changes requested):
The javadoc comment applies to the other modified javadocs from the API

lynxplay (approved):

Machine-Maker (approved):

PR #10094, opened 2023-12-29
full support for UnixDomainSocketAddress

Coloryr:
reopen with https://github.com/PaperMC/Paper/pull/10088


MachineBreaker:
Imo you should merge the patch in 596

Coloryr:
> Imo you should merge the patch in 596

API or Sevrer?

Coloryr:
> I don't know if there's a patch for Unix support in the API module but on server there's one so you should merge into it (IMO)

So what do you mean is to modify that patch instead of creating a new one?

PR #10105, opened 2023-12-31
Call EntityChangeBlockEvent for cake eating

Machine-Maker:
Call EntityChangeBlockEvent for eating cakes or adding a candle to the cake

### DIFF
--- a/patches/server/0834-Fire-EntityChangeBlockEvent-in-more-places.patch
+++ b/patches/server/0834-Fire-EntityChangeBlockEvent-in-more-places.patch
@@ -112,7 +112,7 @@ index db507638a97b5a33df712c54daff35b21922c0dd..2e75fd06e9e379eb95ebfe55086ffc32
                  CriteriaTriggers.ITEM_USED_ON_BLOCK.trigger((ServerPlayer)player, blockPos, itemStack);
              }
 diff --git a/src/main/java/net/minecraft/world/item/EnderEyeItem.java b/src/main/java/net/minecraft/world/item/EnderEyeItem.java
-index 1fddc9025a9e5079f05dd6c86f6ca43c5d51d9c3..6d45e3ebea6721b9c168fea572e9bd610eb54431 100644
+index 1977e702f6af39ebf100c1f2f2edc2d1c4d003b0..cfcd1778b5ae66395400221879dde3575591b23d 100644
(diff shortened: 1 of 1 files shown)


lynxplay (commented):
Should this also include calls for CandleCakeBlock being ignited and extinguished ?

patches/server/0834-Fire-EntityChangeBlockEvent-in-more-places.patch
```diff
@@ -198,6 +198,35 @@ index 32995cb5efdad0bc34ecacacb78cccd21220ba8d..21212462e6b415e96536a27b2c009d15
...
++                // Paper end - call change block event
+                 world.playSound((Player) null, pos, SoundEvents.CAKE_ADD_CANDLE, SoundSource.BLOCKS, 1.0F, 1.0F);
+                 world.setBlockAndUpdate(pos, CandleCakeBlock.byCandle(block));
+                 world.gameEvent((Entity) player, GameEvent.BLOCK_CHANGE, pos);
+@@ -107,6 +112,12 @@ public class CakeBlock extends Block {
+             int i = (Integer) state.getValue(CakeBlock.BITES);
+ 
+             world.gameEvent((Entity) player, GameEvent.EAT, pos);
++            // Paper start - call change block event
++            final BlockState newState = i < 6 ? state.setValue(CakeBlock.BITES, i + 1) : world.getFluidState(pos).createLegacyBlock();
```
Owen1212055: Do we need the fluid legacy check even if the block isn't waterloggable?

Machine-Maker: We do that in other cases so it's consistent since that is what removeBlock/destroyBlock does.

Owen1212055: 👍

kennytv: 6 -> CakeBlock.MAX_BITES

Machine-Maker:
> Should this also include calls for CandleCakeBlock being ignited and extinguished ?

I think this can be left for another PR. It's not super clear if/where an event for *igniting* should be called as bukkit already has the BlockIgniteEvent. Perhaps a counterpart BlockExtinguishEvent instead? Either way, not clear, can be done later.

lynxplay (approved):

PR #10110, opened 2024-01-01
Fix untrack event not being called for all 'untracks'

### LINKED ISSUE #10107: `PlayerUntrackEntityEvent` is no longer called if untracked player/entity teleports to another world
### Expected behavior

The event will be called in any case, including if `event.getPlayer()` will untrack `event.getEntity()` (other `Player`), as before in Paper 1.20.1

### Observed/Actual behavior

The event is called only if `event.getEntity()` (other `Player`) was untracked due to movement/teleport far away from `event.getPlayer()` but **they are both in the same world**.

### Steps/models to reproduce

Join the game from two accounts (or ask a friend to help) and place both players in the same location, then teleport one of them to another world.

Test code:
```java
public void handleEvent(PlayerUntrackEntityEvent theEvent) {
	Entity vEntity =
		theEvent.getEntity();

	if (vEntity.getType() == EntityType.PLAYER) {
		System.out.println("getPlayer: " + theEvent.getPlayer().getName());
		System.out.println("getEntity: " + vEntity.getName());
		System.out.println();
	}
}
```

### Plugin and Datapack List

Only test plugin w/o data packs

### Paper version

b365 (1.20.4)

### Other

I don't know if this issue occurs only if `event.getEntity().getType()` == `PLAYER` , or with any other type too

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/10107

Moves the untrack event call back to the Entity#stopSeenByPlayer method. It was originally added there, then moved when the track event became cancellable. But the untrack event wasn't (and can't) be made cancellable so it should've been left.

### DIFF
--- a/patches/api/0394-Player-Entity-Tracking-Events.patch
+++ b/patches/api/0394-Player-Entity-Tracking-Events.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Player Entity Tracking Events
 
 diff --git a/src/main/java/io/papermc/paper/event/player/PlayerTrackEntityEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerTrackEntityEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..db0a011815daf2690deb4ad2aff08227664fc6b8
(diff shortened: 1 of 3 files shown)


yannicklamprecht (approved):
LGTM

kennytv (approved):

PR #10118, opened 2024-01-03
add method to set creeper swell dir

Gameoholic:
Currently you cannot "unignite" a creeper or change its swell speed without NMS.
This PR allows you to do that.
Let me know if I did anything wrong that needs changing :)

### DIFF
--- a/patches/api/0457-add-method-to-set-creeper-swell-dir.patch
+++ b/patches/api/0457-add-method-to-set-creeper-swell-dir.patch
@@ -1,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Game <viner.atmic@gmail.com>
+Date: Wed, 3 Jan 2024 22:00:48 +0200
+Subject: [PATCH] add method to set creeper swell dir
+
+
+diff --git a/src/main/java/org/bukkit/entity/Creeper.java b/src/main/java/org/bukkit/entity/Creeper.java
+index e144f618702122bf28ebedc5cb8ce0f6ef27c107..5135c63e56e8e5a8e1fd5c2d0bda29fd814b2e39 100644
+--- a/src/main/java/org/bukkit/entity/Creeper.java
++++ b/src/main/java/org/bukkit/entity/Creeper.java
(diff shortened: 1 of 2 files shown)


Machine-Maker:
I'm concerned that it seems this value could be reset every tick in the SwellGoal#tick method. So if you set the swell dir to 1 and the creeper didn't have a target, it would immediately get set back to -1. Can you outline some examples or usecases where the API setting the value does have an effect? If there are such examples, then the behavior should probably be more documented in the javadocs.

TonytheMacaroni:
As mentioned in the PR description, calling `Creeper#setIgnited(false)` doesn't really work, as the swell direction isn't changed. So shouldn't that method be modified to set it so that unigniting actually works?

Gameoholic:
You're right about the value being reset every tick in the goal, it may be better to just have a fix for `Creeper#setIgnited(false)` instead of allowing the swelldir to be changed. The reason I hadn't noticed was because I wrote my own custom creeper goal.


Haven't tested it yet, but something like this?
```
    @Override
    public void setIgnited(boolean ignited) {
        getHandle().setIgnited(ignited);
        if (getHandle().getSwellDir() > 0 && !ignited)
            getHandle().setSwellDir(-1);
    }
```

Would it also be beneficial to have a field for the creeper's swell speed (>0) , that you *can* change through a setter, and just use that value in the goal? So you could make creepers swell up faster for example

electronicboy:
if the sole intent for exposing a method is to fix the behavior of another method, it would make more sense to just fix the OG method, imho; if this had some unexpected side-effect, that would be a different matter, but, that doesn't sound to be the case?

I'm not 100% aware of all the implications here, however, nothing against also exposing a value even if its only real use would be inside of custom AI or whatever

Gameoholic:
I'll just fix the og method then.

Owen1212055:
please merge into missing entity api patch as well

Gameoholic:
Should I create a new PR and close this one? Seeing as I'm doing something different (fixing an existing method)

electronicboy:
I'd generally defer this to owen as he is more aware of the mob AI API type stuff, but, imho, we should do both, add the method and make the old method work as expected

PR #10120, opened 2024-01-03
Fix EntityChangePoseEvent being called during worldgen

### LINKED ISSUE #10119: EntityPoseChangeEvent called when generating structures with camels, causing chunk system error
### Expected behavior

The chunk to generate/load without issues.

### Observed/Actual behavior

Chunk system error when generating the chunk: https://pastes.dev/H3ujZPDylz

### Steps/models to reproduce

1. Download and add the attached datapack. It generates a world full of structures with a camel with `LastPoseChange` of less than 0: [cameeelll.zip](https://github.com/PaperMC/Paper/files/13823769/cameeelll.zip)

2. Start the server.

3. Server will crash as soon as it tries to generate the structure. 

### Plugin and Datapack List

Not relevant.

### Paper version

Not relevant.

### Other

When a camel is loaded and readAdditionalSaveData is called, it will always set the pose if the value of `LastPoseChange` is zero or less. A camel being loaded as part of a structure, for example in desert village, will lead to that logic being called off the main thread. Because the EntityPoseChangeEvent can only be called on the main thread, an exception will be thrown leading to a chunk system error that in turn leads to a crash. 

In theory the same issue should occur if a camel is loaded again and not as part of worldgen.

### LINKED ISSUE #10123: Entities spawned via `/place structure` still have generation flag set
### Expected behavior

Running `/place structure ...` with some structure that spawns entities like a village should spawn valid villagers.

### Observed/Actual behavior

The villagers spawned still have some generation flag set restricting the valid API to use on them.

### Steps/models to reproduce

1. Create a test plugin with 2 event handlers setup as below. You need to add one passthrough `EntityTransformer`
```java

@EventHandler
public void onEvent(ChatEvent event) {
    final Collection<Villager> villagers = event.getPlayer().getLocation().getNearbyEntitiesByType(Villager.class, 10, 10, 10);
    villagers.forEach(v -> v.playEffect(EntityEffect.VILLAGER_HEART));
}

@EventHandler
        return true;
    });
}
```
2. Run `/structure place ...` with some village.
3. say something in chat near the spawned villagers and see the exceptions in the log being unable to play an effect for generation entities.

### Plugin and Datapack List

N/A

### Paper version

1fa48d140c5469d894b436a7ba0cb5cb6f819ee0

### Other

This is caused by the incorrect assumption by upstream that all structure generation happens via `WorldGenRegion` which then serializes the entities to NBT and re-loads them before adding them to `ServerLevel`, at which point the generation flag will be reset. But the `/place` command generates via the current `ServerLevel` and adds entities directly which means the flag isn't reset.

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/10119
Fixes https://github.com/PaperMC/Paper/issues/10123

---
Download the paperclip jar for this pull request: [paper-10120.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1148601399.zip)


Owen1212055 (approved):
Donno if we need to have a diff on the event call? But lgtm

Machine-Maker:
Ok, finally got around to testing this. Need to actually set the generation bool to true on the entity instead much earlier before nbt loading happened. I also removed 2 pairs of overrides of `addFreshEntityWithPassengers` so the delegated addFreshEntity is actually called for each passenger.

electronicboy (approved):

Owen1212055 (approved):

PR #10122, opened 2024-01-03
Fix CraftStructureTransformer from deleting generated entities

Machine-Maker:
`CraftLimitedRegion` loads and stores entities in the protochunk and doesn't parse the entities from NBT each time. This *requires* that a new instance of `CraftLimitedRegion` is created each time before passing it to an api consumer and then saving/clearing it afterwards.

`CraftStructureTransformer` was creating a single instance for an entire structure place which was causing entities to be deleted so no structures would spawn entities.

Demo to cause the issue. Just add this listener and teleport around to newly generated villages.
```java
@EventHandler
        System.out.println(region.getEntities());
        return true;
    });
}
```


Machine-Maker:
Alternative fix would be to add the entity to the list stored in CraftLimitedRegion if the entities were loaded. That way you can re-use the instances and they will stay up-to-date.

PR #10125, opened 2024-01-04
Properly fix allowing World for Block/EntityTransformer

### LINKED ISSUE #10124: AsyncStructureGenerateEvent transformers provide incorrect region interface
### Expected behavior

AsyncStructureGenerateEvent has transfomers for modifying added entities or placed blocks in `EntityTransformer` and `BlockTransformer`. The world interface provided should be `RegionAccessor` to reflect that structure generation can happen to `WorldGenRegion` and `ServerLevel`.

### Observed/Actual behavior

These functional interfaces provide a `LimitedRegion` to interact with the world during that process. This reflects an incorrect implementation by upstream which thought that this all happened to `WorldGenRegion`. If a `ServerLevel` becomes wrapped in a `CraftLimitedRegion` all sorts of assumptions in the implementation of `CraftLimitedRegion` fail, like assumed casts to ProtoChunk.

### Steps/models to reproduce

1. Setup an event handler in a test plugin.
```java
@EventHandler
		// execute a single call to getEntities which will throw CCE (except its hidden because debug logging for commands is disabled but the command will show an error message
		System.out.println(region.getEntities());
        return true;
    });
}
```
2. Run a `/place structure ...` command with a village. Structure generation via `/place` operates on a `ServerLevel` whereas normal chunk generation places structures via `WorldGenRegion`. 
3. Observe the command failing. There won't be an explicit error message unless you change the Commands.java slf4j logger to have debug enabled. You can add a breakpoint yourself to view the actual exception which is a CCE from `LevelChunk` to `ProtoChunk`.

### Plugin and Datapack List

N/A

### Paper version

https://github.com/PaperMC/Paper/commit/1fa48d140c5469d894b436a7ba0cb5cb6f819ee0

### Other

Similar to https://github.com/PaperMC/Paper/issues/10123, this is caused by the incorrect assumption that it will always be a `WorldGenRegion` when running the transformers set by the API. Unfortunately, I think this requires an **API** break to fix. Changing the type of region in `BlockTransformer` and `EntityTransformer` from `LimitedRegion` to `RegionAccessor`. It should never have been `LimitedRegion` in the first place and having `LimitedRegion` is broken functionality. 

I think breaking API here is completely fine because any plugins which based their logic on this assumption are also incorrect and should be required to fix it. This API is also relatively new and so hopefully not too many plugins will be forced to fix and re-compile.

EDIT: I'm now realizing that its marked as Experimental API by upstream so we can easily break it.

Machine-Maker:
Block/EntityTransformer for AsyncStructureGenerateEvent need to accept World and LimitedRegion instances because structure generation can happen to both ServerLevel and WorldGenRegion instances.

Fixes https://github.com/PaperMC/Paper/issues/10124

I also changed the constructor param type for CraftLimitedRegion so this doesn't happen again. We will catch this issue at compile time then.

### DIFF
--- a/patches/api/0457-Fix-allowing-World-for-Block-EntityTransformer.patch
+++ b/patches/api/0457-Fix-allowing-World-for-Block-EntityTransformer.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 3 Jan 2024 20:31:07 -0800
+Subject: [PATCH] Fix allowing World for Block/EntityTransformer
+
+Block/EntityTransformer for AsyncStructureGenerateEvent need to
+accept World and LimitedRegion instances because structure generation
+can happen to both ServerLevel and WorldGenRegion instances.
+
+diff --git a/src/main/java/org/bukkit/util/BlockTransformer.java b/src/main/java/org/bukkit/util/BlockTransformer.java
+index 7f430519c021a68ba718921db20be8442082be5d..ef5438cab5f8beff9846a7128cf5eb343722432e 100644
+--- a/src/main/java/org/bukkit/util/BlockTransformer.java
++++ b/src/main/java/org/bukkit/util/BlockTransformer.java
+@@ -58,5 +58,5 @@ public interface BlockTransformer {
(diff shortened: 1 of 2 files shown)


YanisBft:
Thank you, I'm pretty sure this was the error I linked on discord a few days ago

electronicboy (approved):

PR #10127, opened 2024-01-04
Add velocity forwarding secret env override

oliverjanka:
Previously, the velocity forwarding secret could only be configured via
the configuration option in the global paper configuration.
This makes configuring/passing such a value rather difficult for
containerized/orchestrated servers as these configuration files are
usually part of the server data volume itself and hence cannot be
sourced from a secret.

This commit enables administrators to define the `PAPER_VELOCITY_SECRET`
environment variable, which will override any potentially configured
velocity secret.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -487,10 +487,10 @@ index 0000000000000000000000000000000000000000..d9502ba028a96f9cc846f9ed428bd806
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..45ab7ff5497ceccb85944d2e8d354d9c860abf3b
+index 0000000000000000000000000000000000000000..b92bfd89e32becde2e7630c6116c16f8a4f6614a
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
-@@ -0,0 +1,324 @@
+@@ -0,0 +1,331 @@
 +package io.papermc.paper.configuration;
 +
 +import co.aikar.timings.MinecraftTimings;
@@ -594,7 +594,14 @@ index 0000000000000000000000000000000000000000..45ab7ff5497ceccb85944d2e8d354d9c
 +
 +            @PostProcess
 +            private void postProcess() {
-+                if (this.enabled && this.secret.isEmpty()) {
++                if (!this.enabled) return;
++
++                final String environmentSourcedVelocitySecret = System.getenv("PAPER_VELOCITY_SECRET");
++                if (environmentSourcedVelocitySecret != null && !environmentSourcedVelocitySecret.isEmpty()) {
++                    this.secret = environmentSourcedVelocitySecret;
++                }
++
++                if (this.secret.isEmpty()) {
 +                    LOGGER.error("Velocity is enabled, but no secret key was specified. A secret key is required. Disabling velocity...");
 +                    this.enabled = false;
 +                }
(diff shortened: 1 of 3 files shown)


oliverjanka:
FYI: I only changed the files GlobalConfiguration, VelocityProxy and PaperConfigurations I don't know why `rebuildPatches` also changed the other patches.

electronicboy:
Being frank, this has kinda been on the todo list for a while, but if we wanna do this, we would want to maintain consistency with velocities current expected behavior, i.e. the secret if already set should probably be migrated to the file, and we should follow the file naming of velocity so that people can just copy the file over and save confusion

oliverjanka:
@electronicboy I added functionality to migrate in-config secrets to file automatically. I hope this fits what you meant.

jorishaenseler15:
I really need this feature to run a minecraft server securely on k8s. So i'm looking forward to see the merged pullrequest. @electronicboy

Nevah5:
Bump

Machine-Maker (changes requested):
If the goal is to replace the `secret` setting with a `secretFile`, then there should be a proper migration/transformation for this and then we remove the `secret` setting. If the goal is to offer both an in-config secret, and a secret file, I think there should be a setting that sets where the secret is coming from.

Basically a lot of that logic added in `postProcess` should either be moved to a proper migration, or use a third setting to determine where to read the secret from.

oliverjanka:
@Machine-Maker what's the status of this issue?

patches/server/0005-Paper-config-files.patch
```diff
@@ -3543,6 +3545,76 @@ index 0000000000000000000000000000000000000000..90071acf589fb2fa1c3480883165b429
...
++    private V30_VelocitySecretFile() {
++    }
++
++    public static void apply(ConfigurationTransformation.VersionedBuilder builder) {
++        builder.addVersion(VERSION, ConfigurationTransformation.builder().addAction(PATH, INSTANCE).build());
++    }
++    @Override
++    public Object @Nullable [] visitPath(final NodePath path, final ConfigurationNode value) throws ConfigurateException {
++        if (value.hasChild(LEGACY_SECRET_PATH)) {
++            final String secretFile = getSecretFile(value);
```
Machine-Maker: I think we can assume that since this versioned migration is running, the secret file path isn't set in the config right?

patches/server/0005-Paper-config-files.patch
```diff
@@ -3543,6 +3545,76 @@ index 0000000000000000000000000000000000000000..90071acf589fb2fa1c3480883165b429
...
++                migrateSecret(secretFile, legacySecret.getString());
++            }
++            value.removeChild(LEGACY_SECRET_PATH);
++        }
++        return null;
++    }
++
++    private void migrateSecret(final String secretFile, final String secret) {
++        try {
++            Files.writeString(Path.of(secretFile), secret, StandardOpenOption.CREATE_NEW);
```
Machine-Maker: The path should be relative to the `config` folder, *I think*.

Machine-Maker (changes requested):

lynxplay:
I moved this PR over to an environment variable, which should be a lot easier to use than a volume mount from a secret.
The PR will still respect a potentially configured value to enable people on restrictive hosts etc to continue to define their secrets via a config, but once the configured env var is set, it'll take precedent.

The PR is gonna be up for more in-depth review tomorrow, if you have time till then to give feedback on my changes, that would be great so we can merge (would not wanna merge without an okay from you @oliverjanka, given the drastic changes).

If we wanna revert to your previous state, 4d85ab46001bded20077c2b7bcb8763783211588 was the previous HEAD.

oliverjanka:
Actually I'm not in favour of providing secret data through environment variables due to security concerns. This is one of the reasons I opted for this approach (along with the fact that Velocity works similarly).
However, I'm currently very busy and do not have time to work on my previous or an alternative approach and I'd like to see this feature implemented soon.

Therefore, I won’t stand in the way of you proceeding with this method 🙂.

lynxplay:
Thanks for your quick feedback.
While yea, an environment variable is accessible by anything in the pod, so is the velocity secret file given the user running the server process does need read permissions, and the server internal field storing said secret.

I'll proceed with the environment variable just to get *some* way of doing it in, and if we care enough, the above ref can always be used to potentially revisit a file based approach as well as the env variable.

Machine-Maker (approved):

lynxplay (approved):

PR #10135, opened 2024-01-05
Shulker duplicate event

### LINKED ISSUE #10129: Add `ShulkerDuplicateEvent`

Yes! We want to make than one shulker can't summon more than 3 other shulkers when a shulker bullet hits him. But currently there's no way to get the "parent" shulker and, for example, allow or cancel spawning of new shulker based on parent's PDC value.

Add cancellable `ShulkerDuplicateEvent`, that contain a `getParentShulker` method and, if it's easy to do, `getShulkerBullet`.

"Storing information about collisions"? (c) cat in Paper Discord

### Other

Sure you can use other names for the event and methods. I just used the first ones that came to mind.

Thanks!

chasertw123:
Add a cancellable shulker duplicate event. Provides both the parent and child in the event.

Fixes #10129


patches/server/1062-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,26 @@
...
++++ b/src/main/java/net/minecraft/world/entity/monster/Shulker.java
+@@ -485,6 +485,15 @@ public class Shulker extends AbstractGolem implements VariantHolder<Optional<Dye
+                     entityshulker.setVariant(this.getVariant());
+                     entityshulker.moveTo(vec3d);
+                     this.level().addFreshEntity(entityshulker, org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason.BREEDING); // CraftBukkit - the mysteries of life
++                    // Paper start - Shulker spawn event
++                    if (!entityshulker.isRemoved()) {
++                        io.papermc.paper.event.entity.ShulkerDuplicateEvent event = new io.papermc.paper.event.entity.ShulkerDuplicateEvent(entityshulker.getBukkitEntity(), this.getBukkitEntity());
++                        event.callEvent();
++                        if (event.isCancelled()) {
```
MiniDigger: this can all be just new Event().call I think, also, generally events are fired before something is added, why do you call it after addFreshEntity?

patches/api/0457-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,78 @@
...
++ * Is called when a {@link org.bukkit.entity.Shulker} duplicates.
++ */
++public class ShulkerDuplicateEvent extends EntityEvent implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    private final Entity parent;
++    private boolean cancelled;
++
++    public ShulkerDuplicateEvent(@NotNull Entity child, @NotNull Entity parent) {
```
MiniDigger: seems strange to accept entity here when you always cast anyways

Machine-Maker: Missing `@ApiStatus.Internal`

Machine-Maker: I agree with mini, just make it accept Shulker

Machine-Maker (changes requested):

chasertw123:
Resolved requested changes. Only thing I was unsure of was if I need to discard the entity but I assumed I did.

patches/api/0457-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,78 @@
...
++        this.parent = parent;
++    }
++
++    /**
++     * Gets the shulker that is duplicating.
++     *
++     * @return The Child
++     */
++    @NotNull
++    public Shulker getEntity() {
```
Sytm, line 48: This needs an Override annotation

patches/api/0457-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++
++    /**
++     * Gets the shulker that is duplicating.
++     *
++     * @return The Child
++     */
++    @Override
++    @NotNull
++    public Shulker getEntity() {
++        return (Shulker) entity;
```
MiniDigger, line 49: you forgot to remove the cast

Sytm: This is the protected entity field from the EntityEvent class, that needs to be casted

patches/api/0457-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,79 @@
...
++    public ShulkerDuplicateEvent(@NotNull Shulker child, @NotNull Shulker parent) {
++        super(child);
++        this.parent = parent;
++    }
++
++    /**
++     * Gets the shulker that is duplicating.
++     *
++     * @return The Child
++     */
```
Machine-Maker: I would further clarify that this is the new entity being created, and that it does not exist in the world yet.

patches/server/1062-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,23 @@
...
+index b73dac8f68041f8a2e0752d70cc9d08b5cfd1cde..e02c9c9abbde9b18fad1398d3096961430e21810 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/Shulker.java
++++ b/src/main/java/net/minecraft/world/entity/monster/Shulker.java
+@@ -484,6 +484,12 @@ public class Shulker extends AbstractGolem implements VariantHolder<Optional<Dye
+                 if (entityshulker != null) {
+                     entityshulker.setVariant(this.getVariant());
+                     entityshulker.moveTo(vec3d);
++                    // Paper start - Shulker duplicate event
++                    if (!new io.papermc.paper.event.entity.ShulkerDuplicateEvent((org.bukkit.entity.Shulker) entityshulker.getBukkitEntity(), (org.bukkit.entity.Shulker) this.getBukkitEntity()).callEvent()) {
++                        entityshulker.discard();
```
Machine-Maker: This is interesting to me... Does this happen elsewhere for events relating to the spawning of entities? If it does, then you can leave the discard, I just don't remember exactly how the handling of this is done elsewhere.

Machine-Maker (changes requested):

patches/server/1062-Add-ShulkerDuplicateEvent.patch
chasertw123: That was why originally I did it after the add fresh entity. I could not find any other examples similar to this scenario. Not discarding it seemed wrong since the entity would remain unmarked yet also unadded to the world.

TonytheMacaroni: `EntityBreedEvent` has similar logic, but does not have a discard.

chasertw123: Thanks for some reason didn't see entity breed event had the child.

patches/api/0457-Add-ShulkerDuplicateEvent.patch
```diff
@@ -36,7 +36,8 @@ index 0000000000000000000000000000000000000000..eb32542031088b1735fd0cc90fe350d0
 +    }
 +
 +    /**
-+     * Gets the shulker that is duplicating.
++     * Gets the shulker that is duplicating. The new entity that is being created,
++     * and that does not exist in the world yet.
 +     *
 +     * @return The Child
```
molor: (maybe should write "child" with a small letter, similar to "parent" in the corresponding method just below?)

Warriorrrr (approved):

patches/server/1062-Add-ShulkerDuplicateEvent.patch
```diff
@@ -0,0 +1,22 @@
...
++++ b/src/main/java/net/minecraft/world/entity/monster/Shulker.java
+@@ -484,6 +484,11 @@ public class Shulker extends AbstractGolem implements VariantHolder<Optional<Dye
+                 if (entityshulker != null) {
+                     entityshulker.setVariant(this.getVariant());
+                     entityshulker.moveTo(vec3d);
++                    // Paper start - Shulker duplicate event
++                    if (!new io.papermc.paper.event.entity.ShulkerDuplicateEvent((org.bukkit.entity.Shulker) entityshulker.getBukkitEntity(), (org.bukkit.entity.Shulker) this.getBukkitEntity()).callEvent()) {
++                        return;
++                    }
++                    // Paper end
```
kennytv: Same comment after - on the end comment

kennytv (approved):
Also see the above comment, otherwise approved

Machine-Maker (approved):

PR #10142, opened 2024-01-07
Add FallingBlockDropAsItemEvent

Esophose:
Added a FallingBlockDropAsItemEvent to allow modifying or cancelling the item that is dropped when a falling block either expires or is unable to form a block.

As far as I can tell it isn't possible through any other means to associate the FallingBlock with the item that is being spawned, hence this event.


lynxplay:
Unless I am missing something here but https://jd.papermc.io/paper/1.20/org/bukkit/event/entity/EntityDropItemEvent.html exists 😅

Esophose:
That somehow evaded the minds of 3 different people https://canary.discord.com/channels/289587909051416579/555462289851940864/1193064729321685072

PR #10145, opened 2024-01-07
TNT setblock API

Owen1212055:


patches/api/0300-Missing-Entity-API.patch
```diff
@@ -1128,6 +1128,31 @@ index a52a7af219633d575dcbe8ac4b219834bfd4d4d2..1e839b247182af6873a4d74b236d6412
...
+         return this.getOrigin();
+     }
++    // Paper start
++    /**
++     * Gets the data of the block shown by this TNT entity.
++     *
++     * @return BlockData of the shown block
++     */
++    @org.jetbrains.annotations.NotNull
++    org.bukkit.block.data.BlockData getBlock();
```
lynxplay, line 1146: Might need better naming.

tnt.getBlock() I'd expect to return a block somewhere, idk maybe the block the tnt originated ?
This is more like the displayed block data ?

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/10146
(names are better anyways, and has preconditions)

Owen1212055:
The stars aligned and we managed to PR the exact same thing at the almost exact same time, very spooky. But, they did it better. 😔

PR #10149, opened 2024-01-08
Add removal reason to EntityRemoveFromWorldEvent

TonytheMacaroni:
Adds a removal reason to `EntityRemoveFromWorldEvent`. The `RemovalReason` enum mirrors the NMS `Entity.RemovalReason`, except that the methods `shouldDestroy`/`shouldSave` are renamed to `willDestroy`/`willSave`. Due to the chunk system rewrite, unloaded entities are not set as removed with `RemovalReason.UNLOAD_TO_CHUNK`, so currently if the removal reason is not set and the entity is inaccessible, it'll default to `RemovalReason.UNLOAD_TO_CHUNK` as the removal reason.

### DIFF
--- a/patches/api/0031-Entity-AddTo-RemoveFrom-World-Events.patch
+++ b/patches/api/0031-Entity-AddTo-RemoveFrom-World-Events.patch
@@ -59,13 +59,14 @@ new file mode 100644
 index 0000000000000000000000000000000000000000..5ad5632d4d47d8b42e4f2af19c0fe6cf94ac5643
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/entity/EntityRemoveFromWorldEvent.java
-@@ -0,0 +1,42 @@
+@@ -0,0 +1,169 @@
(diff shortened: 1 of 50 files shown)


kennytv (dismissed):

NoahvdAa (dismissed):

lynxplay:
Given upstream has merged into an entity remove from world event, can you expand this PR to include the new reasons?

Would be fixing #10280

prototype464:
Yes please!

TonytheMacaroni:
Updated the PR to include the causes upstream added. Also moved the primary event call to `CraftEventFactory#callEntityRemoveEvent`, as using `ServerLevel.EntityCallbacks#onTrackingEnd` didn't provide the additional cause context, and that method isn't called on entity unload. The previous event call there was made to use a new `INACCESSIBLE` cause, for when the chunk an entity is in is made inaccessible (but prior to fully unloading).

TonytheMacaroni:
Rebased. Also removed some unneeded diff.

PR #10155, opened 2024-01-11
Update patching script: added an echo statement for better script exe…

KhwajaYousuf:
…cution visibility


electronicboy:
This adds nothing of functional note and just blasts the git history of the file in the name of tabs vs spaces

PR #10156, opened 2024-01-11
Update test plugin

KhwajaYousuf:
Add player join event handler in TestPlugin

Enhance the TestPlugin by adding an event handler for player join events. Now, whenever a player joins the server, a welcome message is sent to the player.

This improves the functionality and interaction of the TestPlugin with player events.


Owen1212055:
Thank you for your contribution, however, changes like these aren’t really necessary. Take a look at our issue page for possible good first issues for contribution.

Thank you! 😊

PR #10158, opened 2024-01-11
Make spawn egg colours visible

luisc99:
Adds an unsafe method for getting the underlying colour definitions for spawn eggs, based on the entity type. I recognise this is probably a fairly niche thing, but it can be useful to have a consistent way of getting some colours for entities when doing certain visual effects.

I've added this to UnsafeValues as it's accessing data directly from the SpawnEggItem class rather than reading data from an individual ItemStack's meta, although I'm happy for it to be moved somewhere more appropriate if desired.

The code is based on ItemFactory#getSpawnEgg, [as found in this patch](https://github.com/PaperMC/Paper/blob/a80d31b1bd84a56f969634c39d78dfa1c437aa9f/patches/server/0655-Add-ItemFactory-getSpawnEgg-API.patch).


patches/api/0458-Make-spawn-egg-color-visible.patch
```diff
@@ -0,0 +1,28 @@
...
++    /**
++     * Obtains the underlying color informating for a spawn egg of a given
++     * entity type, or -1 if the entity passed does not have a spawn egg.
++     * Spawn eggs have two colors - the background layer (0), and the
++     * foreground layer (1)
++     * @param entityType The entity type to get the color for
++     * @param layer The texture layer to get a color for
++     * @return The color of the layer for the entity's spawn egg
++     */
++    int getSpawnEggLayerColor(org.bukkit.entity.EntityType entityType, int layer);
```
yannicklamprecht: Does the int represent a RGB or RGBA color value or is it just a index for an enum?
Please either clarify in JavaDoc or return a `@Nullable Color` instead.

yannicklamprecht (changes requested):

patches/api/0458-Make-spawn-egg-color-visible.patch
kennytv: Yeah it should be a Color object, not the raw int

yannicklamprecht:
Please squash commits in each submodule and rebuildPatches.

luisc99:
Whoops, didn't notice it made a new patch, fixed

yannicklamprecht (approved):

Warriorrrr (approved):
I think it might be nice to add getters for the background & highlight color to spawn egg meta, but currently also lgtm

patches/server/1060-Make-spawn-egg-color-visible.patch
```diff
@@ -0,0 +1,28 @@
...
++    // Paper start - spawn egg color visibility
++    @Override
++    public org.bukkit.Color getSpawnEggLayerColor(org.bukkit.entity.EntityType entityType, int layer) {
++        String typeId = entityType.getKey().toString();
++        net.minecraft.resources.ResourceLocation typeKey = new net.minecraft.resources.ResourceLocation(typeId);
++        net.minecraft.world.entity.EntityType<?> nmsType = net.minecraft.core.registries.BuiltInRegistries.ENTITY_TYPE.get(typeKey);
++        net.minecraft.world.item.SpawnEggItem eggItem = net.minecraft.world.item.SpawnEggItem.byId(nmsType);
++        return eggItem == null ? null : org.bukkit.Color.fromRGB(eggItem.getColor(layer));
++    }
++    // Paper end
```
yannicklamprecht:
```suggestion
+    // Paper end - spawn egg color visibility
```

yannicklamprecht (changes requested):

patches/api/0458-Make-spawn-egg-color-visible.patch
```diff
@@ -0,0 +1,28 @@
...
++     * Obtains the underlying color informating for a spawn egg of a given
++     * entity type, or null if the entity passed does not have a spawn egg.
++     * Spawn eggs have two colors - the background layer (0), and the
++     * foreground layer (1)
++     * @param entityType The entity type to get the color for
++     * @param layer The texture layer to get a color for
++     * @return The color of the layer for the entity's spawn egg
++     */
++    @Nullable org.bukkit.Color getSpawnEggLayerColor(org.bukkit.entity.EntityType entityType, int layer);
++    // Paper end
```
yannicklamprecht:
```suggestion
+    // Paper end - spawn egg color visibility
```
Same here.

yannicklamprecht (changes requested):

Machine-Maker (approved):
If this wasn't in Unsafe, I would have some comments, but I think since it is, and therefore API compat isn't guaranteed, I think its fine. 

Perhaps some more stable API can be added once we have subtypes of ItemType and BlockType in the API.

PR #10160, opened 2024-01-11
Add BlockStateMeta#clearBlockState

### LINKED ISSUE #10159: `ItemMeta` keeping default NBT values
### Expected behavior

For `ItemMeta` implmentations such as `BlockStateMeta` to not add default NBT data/have a way to clear that data.

Ideally `BlockStateMeta`/`ArmorStandMeta`/whatever other implementations have this issue could automatically exclude the NBT value if it's the default, but assuming that's impossible to do without breaking backwards-compatibility in some way, making the setters nullable/some `clear` method (maybe even one to automatically check and remove default values?) would also be helpful.

### Observed/Actual behavior

Certain `ItemMeta` implementations (such as `BlockStateMeta`/`ArmorStandMeta`) leave behind NBT data, even if it's the default values.
Which means that if you, for example, add some items into a chest item using `BlockStateMeta` then clear it to make it a "normal" chest again, it still won't stack with other chests.

### Steps/models to reproduce

Simple event handler to reproduce the issue:
```java
    @EventHandler
    public void on(PlayerInteractEvent event) {
        ItemStack chest = event.getItem();
        if (!(chest.getItemMeta() instanceof BlockStateMeta stateMeta)) {
            return;
        }
        Chest state = (Chest) stateMeta.getBlockState();
        switch (event.getAction()) {
            case LEFT_CLICK_AIR -> state.getInventory().addItem(new ItemStack(Material.STONE));
            case RIGHT_CLICK_AIR -> state.getInventory().clear();
        }
        stateMeta.setBlockState(state);
        chest.setItemMeta(stateMeta);
        event.getPlayer().getInventory().setItemInMainHand(chest);
    }
```
Left click with a chest a few times, then right click to clear, and it will still have the default block state left in it's NBT: `{BlockEntityTag:{Items:[],id:"minecraft:chest",x:0,y:0,z:0}}`.

### Plugin and Datapack List

```
> plugins
[19:29:09 INFO]: Server Plugins (2):
[19:29:09 INFO]: Bukkit Plugins:
[19:29:09 INFO]:  - Debuggery, PaperTestPlugin
```
```
> datapack list
[19:29:57 INFO]: There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
[19:29:57 INFO]: There are no more data packs available
```

### Paper version

```
> version
[19:30:11 INFO]: Checking version, please wait...
[19:30:12 INFO]: This server is running Paper version git-Paper-382 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: a80d31b)
Previous version: git-Paper-350 (MC: 1.20.4)
```

### Other

I know the item property API is coming soon which will probably allow directly clearing this, but seeing as the `ItemMeta` API is staying & the item property API is still a little while away, would be nice to have some solution for this.
See also some [related discussion](https://discord.com/channels/289587909051416579/555462289851940864/1190312492396720278) on Discord.

Machine-Maker:
Resolves https://github.com/PaperMC/Paper/issues/10159

---- 

An alternative would be to make `BlockStateMeta#setBlockState` nullable. I don't know if we want to make that change, especially since the interface is already setup so the getter is nonnull and there is a `has` boolean method.


tal5:
Not sure how big of a PR do you want to make this, but as mentioned in the original issue there's a few other places with a similar problem:
- `ArmorStandMeta`'s values are all `boolean` (i.e. can't be cleared)
- `AxolotlBucketMeta`'s variant can't be cleared
- `CompassMeta`'s `LodestoneTracked` is a `boolean` and will always remain in NBT, etc.

electronicboy (approved):

Machine-Maker:
Yeah, I agree with @electronicboy here. Those point to **specific** properties that can be reset. The property API is the way to handle that. Other similar clear methods pertain to whole groups of data, like PotionMeta#clearCustomEffects, FireworkMeta#clearEffects, and more, all exist and I think this falls into that category.

Owen1212055 (approved):

PR #10161, opened 2024-01-12
Use a ConcurrentHashMap for enum mappings for BlockData

Machine-Maker:
This API should be thread-safe as there is no world state

Reported in https://forums.papermc.io/threads/concurrentmodificationexception-on-chunkgenerator.1022/


Machine-Maker:
I think ideally, the mapping between enums would be just all created during startup, and that is certainly something we could do to fix this, just this is smaller diff.

electronicboy (approved):

PR #10163, opened 2024-01-13
fill command event

yannicklamprecht:
**TODO:**

- [x] cleanup diff -> formatting is messed

Disscuss if we want to expose the filter predicate that way or doing it the "lazy" way using enum and nullable predicate at event level.

### DIFF
--- a/patches/api/0455-fill-command-event.patch
+++ b/patches/api/0455-fill-command-event.patch
@@ -1,0 +1,240 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Yannick Lamprecht <yannicklamprecht@live.de>
+Date: Sat, 13 Jan 2024 20:46:18 +0100
+Subject: [PATCH] fill command event
+
+
+diff --git a/src/main/java/io/papermc/paper/event/command/FillCommandEvent.java b/src/main/java/io/papermc/paper/event/command/FillCommandEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..1e6c4a8809e052a7e68e3ffc3b33eeb07d111f6a
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/command/FillCommandEvent.java
+@@ -0,0 +1,95 @@
++package io.papermc.paper.event.command;
++
++import org.bukkit.block.BlockState;
++import org.bukkit.command.CommandSender;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.Event;
++import org.bukkit.event.HandlerList;
++import org.bukkit.util.BoundingBox;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Triggers when fill command is executed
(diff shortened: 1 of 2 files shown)


Owen1212055:
I think we need to decide the usefulness of this event. I generally think that this doesn’t have much use, as most server owners will not give players access to fill / setblock due to not only how primitive it is.. but also like it’s not as performant as what world edit and such may offer.

Machine-Maker:
My reason for thinking this is a good thing to have, is it allows client-side mods to take advantage of the vanilla commands for pasting schematics and such.

patches/api/0458-fill-command-event.patch
```diff
@@ -0,0 +1,220 @@
...
++        return fillMode;
++    }
++
++    /**
++     * Specifies the bounding box of the two opposing corner blocks of the region to be filled (the "fill region").
++     *
++     * @return the selection bounding box
++     */
++    public BoundingBox getBoundingBox() {
++        return boundingBox;
```
Machine-Maker: Need to clone as its mutable

Machine-Maker (changes requested):

Owen1212055:
> My reason for thinking this is a good thing to have, is it allows client-side mods to take advantage of the vanilla commands for pasting schematics and such.

Schematic paster mods typically already have this logic for by placing blocks, which is now gonna be easier with the block interaction attribute.

Just donno, I seriously do not know servers that would even use setblock/fill over FAWE, which has been the standard in general for stuff like this.

Machine-Maker:
> Schematic paster mods typically already have this logic for by placing blocks, which is now gonna be easier with the block interaction attribute.

They didn't yesterday in chat when someone was asking about this for Litematica support. I don't think the printer functionality is built-in anymore. Plus, printing is still different from pasting all at once.

Machine-Maker:
> I seriously do not know servers that would even use setblock/fill over FAWE

This doesn't replace worldedit by any means, and having one enable doesn't disable the other.

patches/api/0456-fill-command-event.patch
```diff
@@ -0,0 +1,239 @@
...
++     * Specifies the bounding box of the two opposing corner blocks of the region to be filled (the "fill region").
++     *
++     * @return the selection bounding box
++     */
++    public @NotNull BoundingBox getBoundingBox() {
++        return boundingBox.clone();
++    }
++
++    /**
++     * Is the state tha will be set by the fill command
```
yannicklamprecht: Missing t

patches/api/0456-fill-command-event.patch
```diff
@@ -0,0 +1,239 @@
...
+@@ -0,0 +1,126 @@
++package io.papermc.paper.event.command;
++
++import java.util.function.Predicate;
++import org.bukkit.Location;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++
++/**
++ * The placement mode of the fill command see
```
yannicklamprecht: Needs a complete sentence.

yannicklamprecht:
Any updates here?

PR #10166, opened 2024-01-14
Add CropBlock API

Leguan16:
This pull request exposes the `getBaseSeedId` methods from crop blocks. 
This makes it possible to get the corresponding seed to its block representation. 
E.g. `Material.POTATO` for `minecraft:potatoes`

I don't know about the name `getBaseSeedId`. I named it like that cause its the same as the name from the server. Maybe there is a better name for it tho.

### DIFF
--- a/patches/api/0457-CropBlock-API.patch
+++ b/patches/api/0457-CropBlock-API.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: leguan <longboard.noah@gmail.com>
+Date: Sun, 14 Jan 2024 21:26:35 +0100
+Subject: [PATCH] CropBlock API
+
+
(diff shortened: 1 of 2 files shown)


Machine-Maker:
I'm not a fan of this. This type of API should wait until Material is split up into ItemType and BlockType, and then a subtype of BlockType can be made for CropBlockType that has such a method. 

Putting it on BlockData just seems like a hack to work around that. No nms BlockState or API BLockData is required to make the mapping.

Leguan16:
I mean I totally agree with you on that. I didn't like putting it on `BlockData` either but the `CraftCropBlock` extends it so it just made sense as it was the easiest way to do it.

Owen1212055:
In general, we are waiting for the BlockType API. This would allow us to better expose API for stuff like this rather than *hacking* it into the current API we already have.

Feel free to reopen it once that API has gone through. Thank you for your contribution!

PR #10172, opened 2024-01-17
add back optimise non-flush packet sending patch

wuangg:
There are reasons I can think of why the original patch was dropped:

1. Mojang made some code changes in order to optimize the networking performance in 1.20.2, as some of these changes were similar to a collection of patches had already existed in Paper, resulting in the removal of these patches.
2. `AbstractEventExecutor.LazyRunnable` had been deprecated since the release of Netty 4.1.92 (see https://github.com/netty/netty/issues/13335)

Although there is an alternative mechanism provided from the API reference of `AbstractEventExecutor`, the solution in this PR only requires a small amount of code changes.


patches/server/1048-Optimise-non-flush-packet-sending.patch
```diff
@@ -0,0 +1,55 @@
...
+
+diff --git a/src/main/java/net/minecraft/network/Connection.java b/src/main/java/net/minecraft/network/Connection.java
+index 2ae08b21c63490bbf8cd870f9585d82ed131f815..435df4a71e19b84706d318d73cb11c1f82955251 100644
+--- a/src/main/java/net/minecraft/network/Connection.java
++++ b/src/main/java/net/minecraft/network/Connection.java
+@@ -148,6 +148,7 @@ public class Connection extends SimpleChannelInboundHandler<Packet<?>> {
+         this.stopReadingPackets = true;
+     }
+     // Paper end - packet limiter
++    private io.netty.channel.SingleThreadEventLoop executor; // Paper - optimise packets that are not flushed
```
electronicboy: This diff is likely not needed

electronicboy: Actually, no, we need to at least move over to AbstractEventLoop, though, it seems that all of the event executors we'd see all implement SingleThreadEventLoop

electronicboy (approved):

electronicboy:
it's a small change in which packets which *need* to get out fast are generally marked as such, and this is working fine here, however, I do test behind a bungee proxy

wuangg:
> Has this been tested? I can't connect to the server.

Oops.

I was testing @electronicboy 's suggestion and accidentally pushed that code change as I was out of focus for a little bit.
Now this PR should work normally again just like from the beginning.

wuangg:
> Has this been tested? I can't connect to the server.

Oops.

Yes this has been tested and indeed was working, but I was testing out @electronicboy 's suggestion (which causes the player cannot connect to the server) and accidentally pushed that code change as I was out of focus for a bit.

Now this PR should work fine just like from the beginning and I have to be careful next time. Thank you for catching me up!

patches/server/1048-Optimise-non-flush-packet-sending.patch
wuangg: Moving from `SingleThreadEventLoop` to `AbstractEventLoop` could cause `ClassCastException` to be thrown, rendering the server cannot be joined by the players.

wuangg:
> This patch broke the plugin that reads and modifies chat packets with ProtocolLib.

Are there any log about this?

electronicboy:
https://github.com/dmulloy2/ProtocolLib/blob/80aa42009981b505370b00dd71af6d4e41726f38/src/main/java/com/comphenix/protocol/injector/netty/channel/NettyEventLoopProxy.java

This is basically down to how they hook the event loop stuff

wuangg:
It seems like I have to open a PR on ProtocolLib to make their event loop hook extending `AbstractEventLoop`

wuangg:
Close this until I came up with a suitable solution.

Issue #10133, opened 2024-01-05
BlockDropItemEvent is very poorly (and incorrectly) documented


Machine-Maker:
The javadoc for `BlockDropItemEvent` is at best misleading, at worst just plain wrong.
> Called if a block broken by a player drops an item.

No, it's also called for just breaking glass with your fist (with an empty item list)

> This event will also be called if the player breaks a multi block structure, for example a torch on top of a stone. Both items will have an event call.

No, all items broken because of immediate block updates will have their drops contained in this event's list.

I don't know if these are recent behavioral changes from the documentation or if the documentation has been wrong from the beginning. I feel its more likely we should fix the docs instead of try and adjust the behavior of the event to match the docs.

PR #10175, opened 2024-01-18
[ci skip] Fix incorrect documentation for BlockDropItemEvent

rowan-vr:
Fixes the incorrect Javadoc comment of the BlockDropItemEvent as described in #10133 

I'm not entirely sure if the `// Paper end - Fix incorrect documentation` is in the right place as it's between the Javadoc comment and the class definition.


Machine-Maker:
Generally we don’t even add paper comments for Javadoc changes. Also, this change can be merged in with an existing patch called something like “fix javadocs” or “fix upstream javadocs”.

rowan-vr:
Thank you, I'll do that in the morning

Machine-Maker (approved):

lynxplay (approved):

Issue #10112, opened 2024-01-02
Call EntityChangeBlockEvent for candle extinguish (normal/cake) or lighting


lynxplay:
### Is your feature request related to a problem?

N/A

### Describe the solution you'd like.

Call EntityChangeBlockEvent when players extinguish a candle or a candle cake.

### Describe alternatives you've considered.

N/A

PR #10176, opened 2024-01-18
Call EntityChangeBlockEvent for extinguishing (cake) candles

rowan-vr:
Calls the EntityChangeBlockEvent for when candles, normal ones and the ones on cakes, are extinguished as described in #10112


kennytv:
Thank you for the PR! There's already a patch for adding tons of missing calls, please merge it into that: `Fire EntityChangeBlockEvent in more places` (patch number 0824 currently)

kennytv (dismissed):
(otherwise approved)

Owen1212055 (changes requested):
This does not handle cases where the player is null. Such as extinguishing with potions.

Machine-Maker (changes requested):
Looking at how extinguishing happens, I no longer think this event is appropriate for it. Extinguishing candles can happen due to non-entity related interactions, like water flowing. We need a solution that covers all these cases, not just some of them.

rowan-vr:
Maybe a blockextinguishevent and a playerblockextinguish event? Both events could extend some event to be able to listen to it
  
  
  
  
  
  
  

  
  
>

PR #10180, opened 2024-01-18
Disable memory reserve allocating

Warriorrrr:
As far as I can tell, the memory that this method reserves is only ever released by the server in one spot, which is unused in paper. By not allocating it we save about 10,5 MB of memory that can then be used elsewhere


electronicboy:
This is 100% intentional, this memory will pretty much end up in old gen and not be a real factor towards anything until you hit "oh shit", at which point, it's freed in order to pray that they can close the JVM in a safe and effective™ manner

electronicboy:
Then again, the place its used is dumb, I'm not 100% sure on removing this, it's pretty much a pointless slab of memory in which is effectively useless, so it's not a big deal in any manner, not sure, either we remove it or we figure out a manner of effectively using it if needs be

Owen1212055 (approved):
Yeah no, I think that this is fine to have removed. It makes sense on the client as it's used to give some wiggle room for loading world lists and such, which isn't an issue on the server.

lynxplay (approved):

lynxplay (approved):

Machine-Maker (approved):

PR #10183, opened 2024-01-20
Add world to Entity AddTo/RemoveFrom Events

1stGlitch:
When a plugin listens to the EntityAddToWorld and EntityRemoveFromWorld events, I don't believe there is currently any method of directly obtaining which world the entity was actually added to/removed from. Using event.getEntity().getWorld() works in many cases, but not all. Specifically, when an entity is teleported from one world to another, the location of the entity is updated prior to the removal event being called. This means that when an entity goes through a nether/end portal or is teleported between worlds with a command, a plugin listening to the EntityRemoveFromWorldEvent has no way of determining which world the entity was actually removed from (without relying on other events).

To resolve this, I've added the world as a field in the events along with a getter to retrieve it. I also removed an unused import and made the documentation more clear on the event behaviour when chunks load/unload.

### DIFF
--- a/patches/api/0033-Entity-AddTo-RemoveFrom-World-Events.patch
+++ b/patches/api/0033-Entity-AddTo-RemoveFrom-World-Events.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Entity AddTo/RemoveFrom World Events
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/entity/EntityAddToWorldEvent.java b/src/main/java/com/destroystokyo/paper/event/entity/EntityAddToWorldEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..bf5fec2451244ed863f86a113a7e1de94c39c03a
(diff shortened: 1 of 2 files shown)


kennytv (approved):

Machine-Maker (commented):
Should the javadocs include a note that it might not be the same as the entity's world? or is that not worth mentioning.

Leguan16:
I think it doesn't hurt to add it.

1stGlitch:
Sure, it certainly doesn't hurt to clarify things. I added a note to the javadoc in the EntityRemoveFromWorld event. Though I don't believe it's necessary to note it in the EntityAddToWorld event. While I did add a getWorld() method to the EntityAddToWorld event as a precaution, this issue mainly pertains to the EntityRemoveFromWorld event, and I can't think of a scenario where the world would differ from the entity's world when the EntityAddToWorld event is being fired.

Owen1212055 (approved):

PR #10185, opened 2024-01-22
itemstack expose tooltip lines

### LINKED ISSUE #9484: Expose ItemStack#getTooltipLines

I could see some use for getting the full list of Components that make up the tooltip for an item, maybe if you want to render the itemstack somewhere else which supports Components.

Just expose the method on ItemStack with options for showing the advanced and creative tooltips

Using NMS

### Other

It's tricky though, cause the order and contents of the Components should not be considered API. Perhaps this would be more useful if each item type's specific tooltip lines were exposed, but then you lose out on the defined order.

yannicklamprecht:
Resolves https://github.com/PaperMC/Paper/issues/9484

I made the decision to not expose stuff like asCreative() and instead added a separate enum value for that.
Feel free to decide otherwise.

### DIFF
--- a/patches/api/0460-ItemStack-Tooltip-API.patch
+++ b/patches/api/0460-ItemStack-Tooltip-API.patch
@@ -1,0 +1,153 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Yannick Lamprecht <yannicklamprecht@live.de>
+Date: Mon, 22 Jan 2024 13:27:18 +0100
+Subject: [PATCH] ItemStack Tooltip API
+
+
+diff --git a/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContext.java b/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContext.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..c965e9c98b083ce712f5d16bc081cb45c11a4d6b
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContext.java
+@@ -0,0 +1,86 @@
++package io.papermc.paper.inventory.tooltip;
++
++import org.bukkit.entity.Player;
++import org.jetbrains.annotations.Contract;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Context for computing itemstack tooltips via
++ * {@link org.bukkit.inventory.ItemStack#computeTooltipLines(TooltipContext, Player)}
++ */
++public interface TooltipContext {
++
++    /**
++     * Creates a new context with the given advanced and creative
++     * mode settings.
++     *
++     * @param advanced whether the context is for advanced tooltips
++     * @param creative whether the context is for the creative inventory
++     * @return a new context
++     */
++    @Contract("_, _ -> new")
++    static @NotNull TooltipContext create(final boolean advanced, final boolean creative) {
++        return new TooltipContextImpl(advanced, creative);
++    }
++
++    /**
++     * Creates a new context that is neither advanced nor creative
++     * mode settings.
++     *
++     * @return a new context
++     */
++    @Contract("-> new")
++    static @NotNull TooltipContext create() {
++        return new TooltipContextImpl(false, false);
++    }
++
++    /**
++     * Returns whether the context is for the creative
++     * mode inventory.
++     * <p>
++     * Creative tooltips are shown by default when a player is
++     * in the creative inventory.
++     * <p>
++     * <b>Disclaimer:</b>
++     * <p>
++     * Be aware that this can change based on the Minecraft version used.
++     * We do not guarantee that all the values will always be present.
++     *
++     * @return true if for creative mode inventory
++     */
++    boolean isCreative();
++
++    /**
++     * Returns whether the context is for advanced
++     * tooltips.
++     * <p>
++     * Advanced tooltips are shown by default
++     * when a player has {@code F3+H} enabled.
++     * <p>
++     * <b>Disclaimer:</b>
++     * <p>
++     * Be aware that this can change based on the Minecraft version used.
++     * We do not guarantee that all the values will always be present.
++     *
++     * @return true if for advanced tooltips
++     */
++    boolean isAdvanced();
++
++    /**
++     * Returns a new context with {@link #isCreative()}
++     * set to true.
++     *
++     * @return a new context
++     */
++    @Contract("-> new")
++    @NotNull TooltipContext asCreative();
++
++    /**
++     * Returns a new context with {@link #isAdvanced()}
++     * set to true.
++     *
++     * @return a new context
++     */
++    @Contract("-> new")
++    @NotNull TooltipContext asAdvanced();
++}
+diff --git a/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContextImpl.java b/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContextImpl.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..1d9bed6691f581529c53b577b26f1d0f902ccb0d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/inventory/tooltip/TooltipContextImpl.java
+@@ -0,0 +1,16 @@
++package io.papermc.paper.inventory.tooltip;
++
++import org.jetbrains.annotations.NotNull;
++
++record TooltipContextImpl(boolean isCreative, boolean isAdvanced) implements TooltipContext {
++
++    @Override
++    public @NotNull TooltipContext asCreative() {
++        return new TooltipContextImpl(true, this.isAdvanced);
++    }
++
++    @Override
++    public @NotNull TooltipContext asAdvanced() {
++        return new TooltipContextImpl(this.isCreative, true);
++    }
++}
+diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
+index 890c07cfc2e64a52752e96d518578b5eb1afbd19..c6222fe754dee53f269944c7578dde8106377f20 100644
+--- a/src/main/java/org/bukkit/UnsafeValues.java
++++ b/src/main/java/org/bukkit/UnsafeValues.java
+@@ -281,4 +281,6 @@ public interface UnsafeValues {
+     @org.jetbrains.annotations.ApiStatus.Internal
+     io.papermc.paper.plugin.lifecycle.event.LifecycleEventManager<org.bukkit.plugin.Plugin> createPluginLifecycleEventManager(final org.bukkit.plugin.java.JavaPlugin plugin, final java.util.function.BooleanSupplier registrationCheck);
+     // Paper end - lifecycle event API
++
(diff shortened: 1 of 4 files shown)


kennytv (approved):

patches/server/1044-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,30 @@
...
++    public List< net.kyori.adventure.text.Component> tooltipLines(ItemStack itemStack, final org.bukkit.entity.Player player, final org.bukkit.inventory.TooltipOptions options) {
++        Preconditions.checkArgument(options != null, "options == null");
++        net.minecraft.world.item.TooltipFlag flag = switch (options){
++            case NORMAL -> net.minecraft.world.item.TooltipFlag.NORMAL;
++            case ADVANCED -> net.minecraft.world.item.TooltipFlag.ADVANCED;
++            case NORMAL_CREATIVE -> net.minecraft.world.item.TooltipFlag.NORMAL.asCreative();
++            case ADVANCED_CREATIVE -> net.minecraft.world.item.TooltipFlag.ADVANCED.asCreative();
++        };
++        List<net.minecraft.network.chat.Component> lines = CraftItemStack.asNMSCopy(itemStack).getTooltipLines(player == null ? null: ((org.bukkit.craftbukkit.entity.CraftPlayer)player).getHandle(), flag);
++        return lines.parallelStream().filter(java.util.Objects::nonNull).map(io.papermc.paper.adventure.PaperAdventure::asAdventure).toList();
```
kennytv: No need to parallelstream here, the returned value also shouldnt contain any null components, so just map directly

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,61 @@
...
++ * Item tooltips
++ * When hovering over an item in the inventory with the mouse cursor, a tooltip with the item’s name will appear next to the cursor.
++ * Some items display extra information on their tooltips, such as enchantments and shulker box contents.
++ * <p>
++ * Basic item tooltips are also visible in the "Items" section of the statistics menu when hovering over a block or item on the left side of the screen.
++ * In Java Edition, advanced tooltips, including item IDs and the exact remaining durability of tools, can be displayed by pressing F3 + H.
++ * <p>
++ * <a href="https://minecraft.wiki/w/Tooltip">Tooltip</a>
++ */
++public enum TooltipOptions {
```
Machine-Maker: Shouldn't be plural.

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,61 @@
...
++   NORMAL, ADVANCED, NORMAL_CREATIVE, ADVANCED_CREATIVE
```
Machine-Maker: I think some javadocs for each of these would be warranted.

patches/server/1044-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,30 @@
...
++    @Override
++    public List< net.kyori.adventure.text.Component> tooltipLines(ItemStack itemStack, final org.bukkit.entity.Player player, final org.bukkit.inventory.TooltipOptions options) {
++        Preconditions.checkArgument(options != null, "options == null");
++        net.minecraft.world.item.TooltipFlag flag = switch (options){
++            case NORMAL -> net.minecraft.world.item.TooltipFlag.NORMAL;
++            case ADVANCED -> net.minecraft.world.item.TooltipFlag.ADVANCED;
++            case NORMAL_CREATIVE -> net.minecraft.world.item.TooltipFlag.NORMAL.asCreative();
++            case ADVANCED_CREATIVE -> net.minecraft.world.item.TooltipFlag.ADVANCED.asCreative();
++        };
++        List<net.minecraft.network.chat.Component> lines = CraftItemStack.asNMSCopy(itemStack).getTooltipLines(player == null ? null: ((org.bukkit.craftbukkit.entity.CraftPlayer)player).getHandle(), flag);
```
Machine-Maker: You should use `CraftItemStack#unwrap` here instead of copy. There is no need to copy the stack as there shouldn't be any mutation of the stack going on (I think).

Machine-Maker (changes requested):

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,88 @@
...
++    NORMAL,
++    /**
++     * In addition to the values displayed with option {@link #NORMAL},
++     * it adds the following values
++     * <ul>
++     *     <li>item ID e.g. 'minecraft:stone'</li>
++     *     <li>the amount of NBT tags applied onto the stack</li>
++     *     <li>the exact durability of tools e.g. 'Durability: 1560 / 1561'</li>
++     * </ul>
++     *  and the exact remaining durability of tools
```
yannicklamprecht: Missed removing that

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,88 @@
...
++     */
++    ADVANCED,
++    /**
++     * Adds creative category to the tool tip
```
yannicklamprecht: .. the ..

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,88 @@
...
++     */
++    NORMAL_CREATIVE,
++    /**
++     * Displays all from {@link #ADVANCED} and {@link #NORMAL_CREATIVE}
```
yannicklamprecht: ... tooltip option

patches/server/1044-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,30 @@
...
+index 509f8487a170f3dc84b091acf16df26e42391189..c0f95f569feb0ed05085e6cb6f8bacafe8d2c5a8 100644
+--- a/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
++++ b/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
+@@ -641,6 +641,19 @@ public final class CraftMagicNumbers implements UnsafeValues {
+         return org.bukkit.craftbukkit.CraftStatistic.getNMSStatistic(statistic).getName();
+     }
+     // Paper end
++    // Paper start - expose itemstack tooltip lines
++    @Override
++    public List< net.kyori.adventure.text.Component> tooltipLines(ItemStack itemStack, final org.bukkit.entity.Player player, final org.bukkit.inventory.TooltipOption options) {
```
lynxplay: Space after `<`.
The rest of this method also seems to have some formatting issues (see missing space after switch statement, missing space after (...CraftPlayer) cast)

yannicklamprecht: Should be addressed.

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,87 @@
...
++     */
++    public @NotNull List<net.kyori.adventure.text.Component> tooltipLines(@Nullable org.bukkit.entity.Player player, @NotNull TooltipOption tooltipOption) {
++        return Bukkit.getUnsafe().tooltipLines(this, player, tooltipOption);
++    } // Paper end - expose itemstack tooltip lines
+ }
+diff --git a/src/main/java/org/bukkit/inventory/TooltipOption.java b/src/main/java/org/bukkit/inventory/TooltipOption.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..513ab14c5d8c2b22fe6c8c0c0a95d1af5ba2144b
+--- /dev/null
++++ b/src/main/java/org/bukkit/inventory/TooltipOption.java
```
Lulu13022002: I think this should go in the paper package

patches/api/0455-itemstack-expose-tooltip-lines.patch
```diff
@@ -0,0 +1,87 @@
...
++     * Passing the player maps additional data that are player
++     * or player world specific.
++     *
++     * @param player the player is optional, but triggers additional mapping.
++     * @param tooltipOption these options describe as whom the tooltips should be viewed
++     * @return a list of components or a empty list if non-available
++     */
++    public @NotNull List<net.kyori.adventure.text.Component> tooltipLines(@Nullable org.bukkit.entity.Player player, @NotNull TooltipOption tooltipOption) {
++        return Bukkit.getUnsafe().tooltipLines(this, player, tooltipOption);
++    } // Paper end - expose itemstack tooltip lines
```
Lulu13022002: Paper end should be on another line

patches/api/0455-itemstack-expose-tooltip-lines.patch
yannicklamprecht: Addressed.

patches/api/0455-itemstack-expose-tooltip-lines.patch
yannicklamprecht: Addressed.

patches/server/1044-itemstack-expose-tooltip-lines.patch
Machine-Maker: I changed it back to asNMSCopy because apparently there is some possibility of a mutation in the tooltip lines (when an exception is caught parsing the JSON for display name and lore, it removes the display name and lore).

lynxplay (dismissed):
Needs consideration for

a) CREATIVE and CREATIVE_ADV only apply to *inside* the creative inventory. not normal player inventory of a player that is in creative.
b)  no creative category nor tags are rendered.pain
c) "unify that advanced is kinda 


Scratch that ---
computeTooltipLines(Player, TooltipFlag...)
with Flags CREATIVE and ADVANCE.

Callers may pass none, one or both to combine the two.

yannicklamprecht:
Gone for the Adventure style or similar way. Hopefully that's what you meant @lynxplay.
JavaDoc is missing now, will add it later. Brainstorming welcome. 

Sus stew.
Normal, Normal Advanced, Creative, Creative Advanced.

Bildschirmfoto 2024-01-28 um 16 26 14

```java
 @EventHandler
    public void onPlayerInteract(PlayerInteractEvent interactEvent) {
        if (interactEvent.getAction() == Action.RIGHT_CLICK_BLOCK || interactEvent.getAction() == Action.LEFT_CLICK_BLOCK) {
            return;
        }
        var player = interactEvent.getPlayer();
        var itemStack = interactEvent.getItem();
        if (itemStack != null && !itemStack.getType().isAir()) {
            var flagConfig = TooltipFlagConfiguration.normal();
            if (player.getGameMode() == GameMode.CREATIVE) {
                flagConfig = flagConfig.asCreative();
            }
            if (interactEvent.getAction() == Action.RIGHT_CLICK_AIR) {
                flagConfig = flagConfig.asAdvanced();
            }
            final List<Component> components = itemStack.computeTooltipLines(flagConfig, player);
            components.forEach(player::sendMessage);
        } else {
            player.sendRichMessage("<red>Something unexpected happened.");
        }
    }
```

yannicklamprecht:
@Machine-Maker fine with the simplification. We can complicate it when needed e.g. Mojang introduces one that doesn't have one or the other.

electronicboy (approved):

lynxplay (approved):

Machine-Maker (approved):

PR #10190, opened 2024-01-25
Properly check if a loot table exists

### LINKED ISSUE #10189: Lootable#setLootTable not modifying Entity loot tables
### Expected behavior

Calling an Entity's setLootTable should... well, set the loot table.

### Observed/Actual behavior

I am spawning an Entity that I immediately call `setLootTable(LootTables.EMPTY.getLootTable());` on. Spigot docs say I should set to null, but this doesn't seem to have any impact on the drops of the entity, while setting to empty does.

On Spigot, this works as I expect, the entities spawned do not drop anything. 

On Paper, with the same configuration (theyre launching from the same folder), this is not the case, they have standard vanilla drop tables. This happens when I set to null as well. 

### Steps/models to reproduce

1. Spawn an entity
2. Modify its loot table
3. Kill it, and see standard drops

### Plugin and Datapack List

SystemSecurity, WorldEdit, WorldGuard
There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]

SystemSecurity is my plugin, who's development found this bug. My localhost test server is spigot, while my hosted test server for me + helpers is paper. Confirmed this happened on paper within my local setup as well, but is absent on spigot. 

### Paper version

This server is running Paper version git-Paper-398 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: b700460)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/10189

When the loot table registry system was changed, this wasn't update to actually check if the loot table didn't exist to return null. As a result, LootTables.EMPTY#getLootTable returned null when it shouldn't have.

### DIFF
--- a/patches/server/0008-CB-fixes.patch
+++ b/patches/server/0008-CB-fixes.patch
@@ -87,17 +87,15 @@ diff --git a/src/main/java/org/bukkit/craftbukkit/CraftServer.java b/src/main/ja
 index 5f12fce84e0ec001dc43523753883a098434fcb6..d6a1b9bbf9737ed884ecf4af31e1521f46807405 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftServer.java
 +++ b/src/main/java/org/bukkit/craftbukkit/CraftServer.java
-@@ -2454,7 +2454,13 @@ public final class CraftServer implements Server {
+@@ -2454,7 +2454,11 @@ public final class CraftServer implements Server {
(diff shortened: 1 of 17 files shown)


kennytv (approved):

lynxplay (approved):

MrJetby:
Не работает жирный цвет, почините пж

PR #10193, opened 2024-01-27
Add Arrow/Stinger Removal Time API

### LINKED ISSUE #9371: `ArrowBodyCountChangeEvent` is called every tick

`ArrowBodyCountChangeEvent` is called every tick if entity has > 30 arrows in his body because of "hard limit" for 30 arrows in `LivingEntity.java -> tick()`. This continues until arrow count becomes =< 30.

I doubt it has a bad effect on server performance, but I think it's bad anyway.

- Add methods like `LivingEntity#setArrowsInBodyLimit(integer)`/`integer getArrowsInBodyLimit()` (30 by default) so `removeArrowTime` will be calculated correctly if arrow count is > 30 and <= limit.

or...
- Add methods like `LivingEntity#setRemoveNextArrowInBodyTime(integer)`/`integer getRemoveNextArrowInBodyTime()` which (exposes `removeArrowTime` and) can be used, for example, in `ArrowBodyCountChangeEvent` to "delay" its next call.

or... add both?

Manually calculate `removeArrowTime` every tick in `ArrowBodyCountChangeEvent` and just cancel the event if we think that it's too early to remove next arrow.

notTamion:
resolves #9371

### DIFF
--- a/patches/server/0062-Add-methods-for-working-with-arrows-stuck-in-living-.patch
+++ b/patches/server/0062-Add-methods-for-working-with-arrows-stuck-in-living-.patch
@@ -7,10 +7,10 @@ Upstream added methods for this, original methods are now
 deprecated
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
-index 5825f942db3b9870631ff093708dee0e930fccc8..b93d03c1af8b82c009c3f6007cc30e888f6764e0 100644
+index 5825f942db3b9870631ff093708dee0e930fccc8..c5cc6beca76884fdf8eb7bacb619b4d5dedbd31a 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
-@@ -266,9 +266,15 @@ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
+@@ -266,10 +266,29 @@ public class CraftLivingEntity extends CraftEntity implements LivingEntity {
      }
  
      @Override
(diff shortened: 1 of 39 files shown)


patches/server/0062-Add-methods-for-working-with-arrows-stuck-in-living-.patch
```diff
@@ -6,11 +6,32 @@ Subject: [PATCH] Add methods for working with arrows stuck in living entities
...
 
+diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+index 3eeb40c2176a80b9e2a472d43671ae0fe087d9e7..370ba23256cfbfe591c13c9052cef25ee74b6974 100644
+--- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
++++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+@@ -188,6 +188,7 @@ public abstract class LivingEntity extends Entity implements Attackable {
+     public InteractionHand swingingArm;
+     public int swingTime;
+     public int removeArrowTime;
++    public int maximumArrowsInBody; // Paper - Add methods for working with arrows stuck in living entities
```
molor: Isn't it necessary to define a vanilla-default value here (`30`)?
Because now, if I understood correctly, it will be `0` and this PR will prevent having arrows in body by default at all...
Same for stingers

notTamion: woops forgot to set that lol

lynxplay:
The behaviour here is kind of weird.
This is not really a max, barely a soft maximum.
It is a weird way to set one part of the computation algorithm used to determine the ticks till next arrow removal.

I think we'd be better off if we expose a setter for said tick time till next removal.
Plugins wanting to change the value for e.g. 40 arrows can do that in the removal event by setting said value to e.g. always 20 ticks or 1 second.
Or compute their own logic.

We'd need to look into exposing a similar "stinger change event" so plugins can hook there, at which point moving things up in the ArrowRemovalEvent to a common "SomethingRemovalEvent" (name patent pending) might be needed.

notTamion:
yeah thats probably a better way to do it. what would be a good name for the setters? something like setTicksUntilNextArrowRemoved?

notTamion:
Alright the method naming might need some changes but i think the current adventure style naming doesn't sound too bad.

patches/api/0459-Conduit-API.patch
Lulu13022002, line 1: Something messed with git seems like you merged two branches from two different pr

lynxplay (approved):

Machine-Maker (approved):

electronicboy (approved):

PR #10195, opened 2024-01-27
Add Conduit API

notTamion:
Adds a little Conduit API to the basically not existing one lol


patches/server/1045-Conduit-API.patch
```diff
@@ -0,0 +1,35 @@
...
++
++    // Paper start - Conduit API
++    @Override
++    public boolean isActive() {
++        return this.getTileEntity().isActive();
++    }
++
++    @Override
++    public int getRange() {
++        return this.getTileEntity().effectBlocks.size() / 7 * 16;
```
Owen1212055: Can we have diff on change stuff here? These magic numbers are scary.

notTamion: what do you mean by diff? i mean thats just simple vanilla logic for conduits. for every 7 prisma blocks you get 16 blocks of range

lynxplay (changes requested):
Needs diff on change e.g.
```patch
diff --git a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
index 963a59615..4f31dfa64 100644
--- a/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
+++ b/src/main/java/net/minecraft/world/level/block/entity/ConduitBlockEntity.java
@@ -191,7 +191,7 @@ public class ConduitBlockEntity extends BlockEntity {

     private static void applyEffects(Level world, BlockPos pos, List<BlockPos> activatingBlocks) {
         int i = activatingBlocks.size();
-        int j = i / 7 * 16;
+        int j = i / 7 * 16; // Paper - Conduit API; diff on change
         int k = pos.getX();
         int l = pos.getY();
         int i1 = pos.getZ();
```

electronicboy (approved):

lynxplay (approved):

PR #10203, opened 2024-01-29
Fix WaterBottleSplashEvent not forwarding hit result

Lulu13022002:
This fix an issue where getHitBlock, getHitEntity and getHitBlockFace are always null for this event unlike its parent.

### DIFF
--- a/patches/server/0720-More-Projectile-API.patch
+++ b/patches/server/0720-More-Projectile-API.patch
@@ -20,7 +20,7 @@ public net.minecraft.world.entity.projectile.Projectile canHitEntity(Lnet/minecr
 Co-authored-by: Nassim Jahnke <nassim@njahnke.dev>
 
 diff --git a/src/main/java/net/minecraft/world/entity/projectile/ThrownPotion.java b/src/main/java/net/minecraft/world/entity/projectile/ThrownPotion.java
-index bb116b0c75a311d0dc65c032a7727598890ef942..bc003f35251c119db095b73c951bf03d8cb7ba8c 100644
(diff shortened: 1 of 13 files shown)


patches/server/0720-More-Projectile-API.patch
```diff
@@ -559,6 +568,35 @@ index 25133298b26523bd36300ab03e216200c915386b..af7e3acde2eb521ac809b4c5980ff6ad
              hitEntity = ((EntityHitResult) position).getEntity().getBukkitEntity();
          }
  
+@@ -875,19 +875,21 @@ public class CraftEventFactory {
+     }
+ 
+     // Paper start - Fix potions splash events
+-    public static io.papermc.paper.event.entity.WaterBottleSplashEvent callWaterBottleSplashEvent(net.minecraft.world.entity.projectile.ThrownPotion potion, HitResult hitResult, Map<LivingEntity, Double> affectedEntities, java.util.Set<LivingEntity> rehydrate, java.util.Set<LivingEntity> extinguish) {
++    public static io.papermc.paper.event.entity.WaterBottleSplashEvent callWaterBottleSplashEvent(net.minecraft.world.entity.projectile.ThrownPotion potion, @Nullable HitResult hitResult, Map<LivingEntity, Double> affectedEntities, java.util.Set<LivingEntity> rehydrate, java.util.Set<LivingEntity> extinguish) {
```
kennytv: I'd just make this nullable in the patch the method is added to avoid diff over diff

kennytv (approved):

Machine-Maker (approved):
I moved some more of the logic relating to nullable HitResult to the previous patch, and then tested it. Works as it should.

PR #10206, opened 2024-01-31
Fixes Incorrect Collision Behaviour with Block VoxelShapes

### LINKED ISSUE #9807: Incorrect Collision Behaviour with Pistons
### Expected behavior

When i piston pushes a TNT entity into a solid block, and then retracts, the TNT should not be misaligned, and a piston one block away cannot push the TNT

### Observed/Actual behavior

When i piston pushes a TNT entity into a solid block, and then retracts, the TNT gets misaligned, and a piston one block away can to push the TNT

### Steps/models to reproduce

Here is a video showing the Problem

https://streamable.com/8kl784

### Plugin and Datapack List

No Plugins/ Datapacks

### Paper version

 This server is running Paper version git-Paper-217 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: cfe311d)

### Other

Misaligned TNT as described/shown in the video is something that can happen in vanilla, however in normal minecraft it requires a slightly more complex setup, the contraption shown in the video only works on servers running paper

DanielSchmerber:
The function that checks wich blocks should be included in a collision has a special case with blocks that have a "Block" Shape.
If it is a Block, the check wether the block should be included in the collision does not subtract a collision-Epsilon from the bounding box
The Optimized Implementation of paper did not have this special case, wich intruduced non-vanilla behaviour

This PR fixes https://github.com/PaperMC/Paper/issues/9807


Spottedleaf:
The code that Vanilla uses is likely a poor copy of optimisations Paper previously had years ago. The problem with this code is that Vanilla is not adhering to the collision system's principles, as the epsilon exists to fix collision bugs. The collision patch also fixes issues with the epsilon in hard entity collisions (i.e boats) as well, which are not going to be reverted.

Really, this should be reported to Mojang.

DanielSchmerber:
The PR changes the code that is responsible for checking wich blocks to collide with, not collision itself

This is only an problem on Paper currently, and causes noticable non vanilla behaviours as shown in the confirmed issue (https://streamable.com/dibq02)

the collision without epsilon is only used on Blocks with a "Block" Shapes, and would not revert any other collision that follows the epsilon rule

PR #10212, opened 2024-02-01
Add BlockPressChangeEvent

dawon:
Adds Event to detect Presses and Releases of Buttons&Pressure Plates. 

I know button press can be detected via InteractEvents, but my main motivation was the Release part, which does not have any event. So if you deem it to be too much, I can rework it into something like BlockReleasedEvent or BlockRevertedToDefaultEvent...


TonytheMacaroni:
Not really commenting on the merits of this event, but you can check for the pressing and releasing of pressure plates and buttons using `BlockRedstoneEvent`, can't you?

dawon:
> Not really commenting on the merits of this event, but you can check for the pressing and releasing of pressure plates and buttons using `BlockRedstoneEvent`, can't you?

Yes, you can... Again, the pressing Part is there for sole parity/symmetry of the event... also as opposed to BlockRedstoneEvent, which is called for dozens other cases, this is only for the presses...

PR #10216, opened 2024-02-03
Fix NotNull locales

Moulberry:
Two methods returning locales are annotated @NotNull, despite being able to return null

### DIFF
--- a/patches/api/0006-Adventure.patch
+++ b/patches/api/0006-Adventure.patch
@@ -3586,7 +3586,7 @@ index 2f6ca42330675733b2b4132cbb66e433788d05d5..997b06c19a5277656521e0e298f2958c
  
      @NotNull
 diff --git a/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java b/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
-index 36b436e145a7215682b692a87ab894df25752c1d..c5e8bffef736365881953161ff812c98adf93a80 100644
+index 36b436e145a7215682b692a87ab894df25752c1d..dc6b41950570c3a8b02415dd9017b2336e6e7f0c 100644
 --- a/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
 +++ b/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
 @@ -12,17 +12,31 @@ public class PlayerLocaleChangeEvent extends PlayerEvent {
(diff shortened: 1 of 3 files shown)


patches/api/0006-Adventure.patch
```diff
@@ -3608,7 +3608,10 @@ index 36b436e145a7215682b692a87ab894df25752c1d..c5e8bffef736365881953161ff812c98
      public PlayerLocaleChangeEvent(@NotNull Player who, @NotNull String locale) {
          super(who);
          this.locale = locale;
-+        this.adventure$locale = net.kyori.adventure.translation.Translator.parseLocale(locale); // Paper
++        // Paper start
++        java.util.Locale javaLocale = net.kyori.adventure.translation.Translator.parseLocale(locale);
++        this.adventure$locale = javaLocale != null ? javaLocale : java.util.Locale.US;
```
Leguan16: Why don't you use `requireNonNullElse` here like at the other spots?

electronicboy (approved):

imDaniX:
Is there a reason why the default fallback is the US locale? I'd say if player's locale cannot be parsed, it'd better be something like `Locale.ROOT`, so we can check for that and act accordingly, like warning a player.

olijeffers0n:
> Is there a reason why the default fallback is the US locale? I'd say if player's locale cannot be parsed, it'd better be something like `Locale.ROOT`, so we can check for that and act accordingly, like warning a player.

Personally, I would say it is very standard to default to English. I am biased, but defaulting to the server's root Locale seems like it could cause issues

imDaniX:
> Personally, I would say it is very standard to default to English. I am biased, but defaulting to the server's root Locale seems like it could cause issues

Well, there's also `Locale.ENGLISH` which is used fairly often too, hence the question. Also, the root locale isn't "server's root locale", but rather a neutral one, as stated in its javadoc.

patches/api/0006-Adventure.patch
Moulberry: I find requireNonNullElse to be rather unwieldy. It was used at the other location because otherwise 200+ patches would have their line numbers changed.

If the maintainers would prefer
`this.adventure$locale = java.util.Objects.requireNonNullElse(net.kyori.adventure.translation.Translator.parseLocale(locale), java.util.Locale.US);`
here then I'll make that change

Moulberry:
> Is there a reason why the default fallback is the US locale? I'd say if player's locale cannot be parsed, it'd better be something like `Locale.ROOT`, so we can check for that and act accordingly, like warning a player.

Locale.US is the fallback locale used elsewhere:
https://github.com/PaperMC/Paper/blob/c5d168cef92148a6799141832fdb5f95732c1a38/patches/server/0010-Adventure.patch#L2763
https://github.com/PaperMC/Paper/blob/c5d168cef92148a6799141832fdb5f95732c1a38/patches/server/0010-Adventure.patch#L1476

I don't mind Locale.ROOT, but consistency with the existing code is more important

patches/api/0006-Adventure.patch
Leguan16: It's just weird in terms of consistency.

electronicboy (approved):

PR #10218, opened 2024-02-03
Fix entity desync

### LINKED ISSUE #9958: ArmorStand teleportation desync
### Expected behavior

Im currently developing a vehicle plugin and I am teleporting the seats every tick to their offset position from the main armorstand.
I expect them to be like in this image.

### Observed/Actual behavior

However, when driving out of the render distance of a player and entering it again, the armorstands are desynced with the player like in this picture.

After a few seconds the armorstands fix themself.

### Steps/models to reproduce

1. Teleport an armorstand slowly out of the render distance, 
2. Teleport it back slowly

### Plugin and Datapack List

Citizens, FAWE, spark

### Paper version

```
> version
[22:41:40 INFO]: Checking version, please wait...
[22:41:40 INFO]: This server is running Paper version git-Paper-297 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 250388d)
Previous version: git-Paper-"250388d" (MC: 1.20.2)
```

### Other

This [spigot post](https://www.spigotmc.org/threads/armor-stand-position-desync-when-teleporting-into-and-out-of-render-distance.574237/) actually shows up the same issue, in case my explanation was a bit unclear.

felixb8:
resolves #9958

### DIFF
--- a/patches/server/1045-Fix-entity-desync.patch
+++ b/patches/server/1045-Fix-entity-desync.patch
@@ -1,0 +1,18 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Felix <29801805+ImGxrke@users.noreply.github.com>
+Date: Sat, 3 Feb 2024 20:53:53 +0100
+Subject: [PATCH] Fix entity desync
+
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerEntity.java b/src/main/java/net/minecraft/server/level/ServerEntity.java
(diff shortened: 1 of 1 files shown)


Dimiikou:
lgtm!

Owen1212055 (changes requested):
This is not a proper fix for this. This is an issue with relative teleporting which will need to be looked at, avoiding relative teleportation completely is not the solution.

Owen1212055:
Closing as per the top comment.

PR #10225, opened 2024-02-06
add rich message component support to configuration

### LINKED ISSUE #5449: Add Component-Methods to FileConfigurations

The problem is that with the new Component-System, Adventure, that has been put into the Paper-API, everything's pretty much fine, but FileConfigurations, like the default config you can add to your plugin, does currently not have any methods related to Components.

It would be great to have methods for FileConfigurations regarding Components.
Such as:
getComponent(String path), which returns a Component,
getComponentList(String path) , which returns a List consisting of Components,
and maybe also getText and get TextList, for TextComponents, but that one is just QoL.

Because getString and getStringList both exist, it would be great to have the same methods for Components, without having to convert to one from a string or List<String>.

It is possible to get a plain String out of the config and then convert that one to a Component or TextComponent, 
and it is also possible, easiest with an ArrayStream, to convert a List<String> to a List<Component>, but if Adventure is now the only way to go, and all the methods containing Strings are deprecated now, it would be great to have simple methods for the new Types.

### Agreements

I have searched for and ensured there isn't already an open issue regarding this.
I have ensured the feature I'm requesting isn't already in the latest supported Paper build.

yannicklamprecht:
resolves #5449 

TODOS:

- [x] Wait for adventure 4.16.0 release
- [x] remove mavenLocal() and migrate to version 4.16.0
- [x] add changes to adventure patch

### DIFF
--- a/patches/api/0006-Adventure.patch
+++ b/patches/api/0006-Adventure.patch
@@ -5,16 +5,17 @@ Subject: [PATCH] Adventure
 
 Co-authored-by: zml <zml@stellardrift.ca>
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
+Co-authored-by: Yannick Lamprecht <yannicklamprecht@live.de>
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index b3df8cf8d32a4cc99a3a3950f146cdab65826e1b..b577114c2b89fe2053123d1a542d37dff7fa8d5a 100644
+index b3df8cf8d32a4cc99a3a3950f146cdab65826e1b..9011451c6a60db08dcc255a84470f8ce92f7b2a4 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -11,12 +11,28 @@ java {
  
  val annotationsVersion = "24.0.1"
  val bungeeCordChatVersion = "1.20-R0.2"
-+val adventureVersion = "4.15.0"
++val adventureVersion = "4.16.0"
 +val apiAndDocs: Configuration by configurations.creating {
 +    attributes {
 +        attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category.DOCUMENTATION))
@@ -2053,6 +2054,109 @@ index fcc34b640265f4dccb46b9f09466ab8e1d96043e..5c813ac024f675951159a59d88d8baa0
 +        return this.getCaller();
 +    }
 +    // Paper end
+ }
+diff --git a/src/main/java/org/bukkit/configuration/ConfigurationSection.java b/src/main/java/org/bukkit/configuration/ConfigurationSection.java
+index b6b00af08f12f838411845e4f4e29e62826dfc7f..d168b1e58b4b4ad00466fab60232d516551668e0 100644
+--- a/src/main/java/org/bukkit/configuration/ConfigurationSection.java
++++ b/src/main/java/org/bukkit/configuration/ConfigurationSection.java
+@@ -1058,4 +1058,98 @@ public interface ConfigurationSection {
+      * one line.
+      */
+     public void setInlineComments(@NotNull String path, @Nullable List<String> comments);
++
++    // Paper start - add rich message component support to configuration
++    /**
++     * Gets the requested MiniMessage formatted String as Component by path.
++     * <p>
++     * If the Component does not exist but a default value has been specified,
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @return Requested Component.
++     */
++    default net.kyori.adventure.text.@Nullable Component getRichMessage(final @NotNull String path) {
++        return this.getRichMessage(path, null);
++    }
++
++    /**
++     * Gets the requested MiniMessage formatted String as Component by path.
++     * <p>
++     * If the Component does not exist but a default value has been specified,
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param fallback component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, !null -> !null")
++    default net.kyori.adventure.text.@Nullable Component getRichMessage(final @NotNull String path, final net.kyori.adventure.text.@Nullable Component fallback) {
++        return this.getComponent(path, net.kyori.adventure.text.minimessage.MiniMessage.miniMessage(), fallback);
++    }
++
++    /**
++     * Sets the specified path to the given value.
++     * <p>
++     * If value is null, the entry will be removed. Any existing entry will be
++     * replaced, regardless of what the new value is.
++     *
++     * @param path Path of the object to set.
++     * @param value New value to set the path to.
++     */
++    default void setRichMessage(final @NotNull String path, final net.kyori.adventure.text.@Nullable Component value) {
++        this.setComponent(path, net.kyori.adventure.text.minimessage.MiniMessage.miniMessage(), value);
++    }
++
++    /**
++     * Gets the requested formatted String as Component by path deserialized by the ComponentDecoder.
++     * <p>
++     * If the Component does not exist but a default value has been specified,
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param decoder ComponentDecoder instance used for deserialization
++     * @return Requested Component.
++     */
++    default <C extends net.kyori.adventure.text.Component> @Nullable C getComponent(final @NotNull String path, final net.kyori.adventure.text.serializer.@NotNull ComponentDecoder<? super String, C> decoder) {
++        return this.getComponent(path, decoder, null);
++    }
++
++    /**
++     * Gets the requested formatted String as Component by path deserialized by the ComponentDecoder.
++     * <p>
++     * If the Component does not exist but a default value has been specified,
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param decoder ComponentDecoder instance used for deserialization
++     * @param fallback component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, _, !null -> !null")
++    default <C extends net.kyori.adventure.text.Component> @Nullable C getComponent(final @NotNull String path, final net.kyori.adventure.text.serializer.@NotNull ComponentDecoder<? super String, C> decoder, final @Nullable C fallback) {
++        java.util.Objects.requireNonNull(decoder, "decoder");
++        final String value = this.getString(path);
++        return decoder.deserializeOr(value, fallback);
++    }
++
++    /**
++     * Sets the specified path to the given value.
++     * <p>
++     * If value is null, the entry will be removed. Any existing entry will be
++     * replaced, regardless of what the new value is.
++     *
++     * @param path Path of the object to set.
++     * @param encoder the encoder used to transform the value
++     * @param value New value to set the path to.
++     */
++    default <C extends net.kyori.adventure.text.Component> void setComponent(final @NotNull String path, final net.kyori.adventure.text.serializer.@NotNull ComponentEncoder<C, String> encoder, final @Nullable C value) {
++        java.util.Objects.requireNonNull(encoder, "encoder");
++        this.set(path, encoder.serializeOrNull(value));
++    }
++    // Paper end - add rich message component support to configuration
  }
 diff --git a/src/main/java/org/bukkit/conversations/Conversable.java b/src/main/java/org/bukkit/conversations/Conversable.java
 index b7d8dd30360a38dbdc7bbce40c8e6ced7261f833..0817f2395c2b18828565435568ce651f5ba99a99 100644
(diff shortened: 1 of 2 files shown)


Machine-Maker (changes requested):
We should probably allow the ability to pass custom MiniMessage instances for this for custom tags and whatnot.

imDaniX:
> We should probably allow the ability to pass custom MiniMessage instances for this for custom tags and whatnot.

Having an option to pass custom `ComponentSerializer` in general would be helpful too.

As a side note, this will currently throw an exception if a user uses `§` in their string. I'm just thinking that there should be some safe railings for this case (like returning a def/null component), or at least some javadoc info mentioning the behavior.

yannicklamprecht:
> We should probably allow the ability to pass custom MiniMessage instances for this for custom tags and whatnot.

Done.

lynxplay (dismissed):
I think it would be great if we split this into 

getRichMessage(path)
getRichMessage(path, defaultMini)

getComponent(path, ComponentSerialiser<T, String>, defaultStr)

that way, we also enable easy fetching for alternative formats down the line.

patches/api/0459-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,119 @@
...
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param component component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, !null -> !null")
++    @Nullable
++    default net.kyori.adventure.text.Component getRichMessage(@NotNull String path, @Nullable net.kyori.adventure.text.Component component){
```
yannicklamprecht: @lynxplay should I yeet this?

yannicklamprecht: As discussed in voice, we instead yank the String default one.

> We have String Templates in one Java version anyways. So we just have a MiniMessage StringProcessor.

patches/api/0460-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,85 @@
...
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param componentSerializer componentSerializer instance used for deserialization
++     * @param component component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, _, !null -> !null")
++    @Nullable
++    net.kyori.adventure.text.Component getComponent(@NotNull String path, @NotNull net.kyori.adventure.text.serializer.ComponentSerializer<net.kyori.adventure.text.Component, net.kyori.adventure.text.Component, String> componentSerializer, @Nullable net.kyori.adventure.text.Component component);
```
Machine-Maker: This method should use generics to support serializers that output other Component subtypes

prob smth like `<C extends Component> C getComponent(..., ComponentSerializer<C, ?, String> serializer, ...)`

kennytv (approved):

patches/api/0460-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,85 @@
...
++    /**
++     * Gets the requested Component by path.
++     * <p>
++     * If the Component does not exist but a default value has been specified,
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @return Requested Component.
++     */
```
emilyy-dev: I think it's worth noting that the component at the given path is a MiniMessage string, the same for the (potential) default value and the other getRichMessage method

patches/api/0460-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,85 @@
...
++     * this will return the default value. If the Component does not exist and no
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param component component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, !null -> !null")
++    @Nullable
++    default net.kyori.adventure.text.Component getRichMessage(@NotNull String path, @Nullable net.kyori.adventure.text.Component component){
```
emilyy-dev: rename `component` parameter to `fallback`/`def`, same with the `getComponent` method

patches/api/0460-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,85 @@
...
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param componentSerializer componentSerializer instance used for deserialization
++     * @param component component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, _, !null -> !null")
++    @Nullable
++    <C extends net.kyori.adventure.text.Component> C getComponent(@NotNull String path, @NotNull net.kyori.adventure.text.serializer.ComponentSerializer<C, C, String> componentSerializer, @Nullable C component);
```
emilyy-dev: Do we care about a fallback-lacking overload? `getComponent(String, ComponentSerializer)`

patches/api/0460-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,85 @@
...
+     }
+ 
++    // Paper start - add rich message component support to configuration
++    @Nullable
++    @Override
++    public <C extends net.kyori.adventure.text.Component> C getComponent(@NotNull final String path, @NotNull final net.kyori.adventure.text.serializer.ComponentSerializer<C, C, String> componentSerializer, @Nullable final C component) {
++        final String value = getString(path);
++        return value == null ? component : componentSerializer.deserialize(value);
++    }
++    // Paper end - add rich message component support to configuration
```
emilyy-dev: this can be promoted to `ConfigurationSection` as a `default` method

emilyy-dev (commented):
Generally looks good, my couple concerns are regarding documentation mostly

lynxplay:
General "lets wait and see" for https://github.com/KyoriPowered/adventure/issues/1034, as we could use ComponentDecoder over ComponentSerialiser if possible.

I'd give it a couple days, otherwise we'll commodore it down the line.

emilyy-dev:
Also maybe the patch could/should be promoted to the adventure patch, would like to know if anyone has strong opinions against that

yannicklamprecht:
> General "lets wait and see" for [KyoriPowered/adventure#1034](https://github.com/KyoriPowered/adventure/issues/1034), as we could use ComponentDecoder over ComponentSerialiser if possible.
> 
> I'd give it a couple days, otherwise we'll commodore it down the line.

Waiting another week or month will not kill it. So just wait for the decoder.

patches/api/0460-add-rich-message-component-support-to-configuration.patch
yannicklamprecht: Hopefully addressed like you expected it.

patches/api/0461-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,68 @@
...
++     * @return Requested Component.
++     */
++    @Contract("_, !null -> !null")
++    @Nullable
++    default net.kyori.adventure.text.Component getRichMessage(@NotNull String path, @Nullable net.kyori.adventure.text.Component fallback) {
++        return getComponent(path, net.kyori.adventure.text.minimessage.MiniMessage.miniMessage(), fallback);
++    }
++
++    /**
++     * Gets the requested MiniMessage formatted String as Component by path.
```
Machine-Maker: Not always `MiniMessage` now that it takes any serializer

patches/api/0461-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,68 @@
...
++     * default value was specified, this will return null.
++     *
++     * @param path Path of the Component to get.
++     * @param componentSerializer componentSerializer instance used for deserialization
++     * @param fallback component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, _, !null -> !null")
++    @Nullable
++    default <C extends net.kyori.adventure.text.Component> C getComponent(@NotNull String path, @NotNull net.kyori.adventure.text.serializer.ComponentSerializer<C, C, String> componentSerializer, @Nullable C fallback) {
```
Machine-Maker: I think the first generic param should be `?` cause we technically don't care at all about that input type, we only care about the output type.

Machine-Maker (changes requested):
Should we add `set` methods as well? I don't think there are for other specific types, but since we aren't hooking Component into the snakeyaml type system, maybe some set methods would be good?

patches/api/0461-add-rich-message-component-support-to-configuration.patch
emilyy-dev: > I think the first generic param should be `?` cause we technically don't care at all about that input type, we only care about the output type.

Although if we wait for ComponentDecoder to be a thing this wouldn't be a concern

yannicklamprecht:
> Should we add `set` methods as well? I don't think there are for other specific types, but since we aren't hooking Component into the snakeyaml type system, maybe some set methods would be good?

Addressed with 72031ae5a6b410c505cd0285a1003897d1b31449

patches/api/0461-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,114 @@
...
++     *
++     * @param path Path of the Component to get.
++     * @param decoder ComponentDecoder instance used for deserialization
++     * @param fallback component that will be used as fallback
++     * @return Requested Component.
++     */
++    @Contract("_, _, !null -> !null")
++    @Nullable
++    default <C extends net.kyori.adventure.text.Component> C getComponent(@NotNull String path, @NotNull net.kyori.adventure.text.serializer.ComponentSerializer<?, C, String> decoder, @Nullable C fallback) {
++        com.google.common.base.Preconditions.checkArgument(decoder != null, "Decoder needs to be present");
```
kashike:
```suggestion
+        java.util.Objects.requireNonNull(decoder, "decoder");
```

patches/api/0461-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,114 @@
...
++     * <p>
++     * If value is null, the entry will be removed. Any existing entry will be
++     * replaced, regardless of what the new value is.
++     *
++     * @param path Path of the object to set.
++     * @param encoder the encoder used to transform the value
++     * @param value New value to set the path to.
++     */
++    default <I extends net.kyori.adventure.text.Component> void setComponent(@NotNull String path, @NotNull net.kyori.adventure.text.serializer.ComponentEncoder<I, String> encoder, @Nullable I value) {
++        com.google.common.base.Preconditions.checkArgument(encoder != null, "Encoder needs to be present");
```
kashike:
```suggestion
+        java.util.Objects.requireNonNull(encoder, "encoder");
```

patches/api/0461-add-rich-message-component-support-to-configuration.patch
```diff
@@ -0,0 +1,114 @@
...
+diff --git a/src/main/java/org/bukkit/configuration/MemorySection.java b/src/main/java/org/bukkit/configuration/MemorySection.java
+index 90751edd742d9b8d2171c4b16891aa009e4b8dfe..e057b03671f091eccc47ae9961b0ceb37d40504a 100644
+--- a/src/main/java/org/bukkit/configuration/MemorySection.java
++++ b/src/main/java/org/bukkit/configuration/MemorySection.java
+@@ -189,6 +189,12 @@ public class MemorySection implements ConfigurationSection {
+     @Override
+     public void set(@NotNull String path, @Nullable Object value) {
+         Preconditions.checkArgument(!Strings.isNullOrEmpty(path), "Cannot set to an empty path");
++        // Paper start - add rich message component support to configuration
++        if(value instanceof net.kyori.adventure.text.Component component) {
```
kashike:
```suggestion
+        if (value instanceof final net.kyori.adventure.text.Component component) {
```

kashike (dismissed):

patches/api/0460-add-rich-message-component-support-to-configuration.patch
yannicklamprecht: Addressed in https://github.com/PaperMC/Paper/commit/88393c02489850bbf8d582bf0da56bdea82fd408#diff-641780de4795f475f2b5c979f77ab04d7be650885ee99dd34b574ea6a0209983R74-R76

patches/api/0461-add-rich-message-component-support-to-configuration.patch
yannicklamprecht: Addressed by using a local version of https://github.com/KyoriPowered/adventure/pull/1035 see https://github.com/PaperMC/Paper/pull/10225/commits/45b2e17f141e167b861f15b21dc9c85277fb5786

yannicklamprecht:
Should we fixup this onto the adventure commit?

lynxplay:
Yea :+1: add yourself as coauthor but yea

yannicklamprecht:
> Yea 👍 add yourself as coauthor but yea

Done.

electronicboy (approved):

emilyy-dev (approved):

Owen1212055 (approved):

Machine-Maker (approved):

PR #5075, opened 2021-01-17
Add BlockPreDispenseEvent

me4502:
This PR adds an event that is called prior to a Dispenser firing anything. This allows full access to the pre-fire state of the dispenser inventory, and therefore clean cancelling of the event.

Main use case here is for plugins that want to either entirely cancel a dispenser event, or want to perform an action based off the contents of the dispenser. Due to the way the normal dispense event works, neither of these are currently possible.

This also allows working around an old Spigot bug, https://hub.spigotmc.org/jira/browse/SPIGOT-4162

PR #10226, opened 2024-02-07
Fire BlockPreDispenseEvent for droppers

rymiel:
The BlockPreDispenseEvent, added in #5075, is only fired for dispensers, and not droppers. I see this as an inconsistency in the API, as BlockDispenseEvent is called for both droppers and dispensers.

What I'm trying to achieve is being able to cancel droppers doing anything while also avoiding the sound and particles. This is possible for dispensers through this event. Cancelling BlockDispenseEvent works fine for droppers, but it still plays the click sound.

I saw prior discussion in the Discord (many months ago) which mentions that there is no reason the event isn't called for droppers, and that it really should.

### DIFF
--- a/patches/server/0504-Add-BlockPreDispenseEvent.patch
+++ b/patches/server/0504-Add-BlockPreDispenseEvent.patch
@@ -17,17 +17,17 @@ index 5593a0aa9e618071b6521b213dde0f628348c3dc..644e64850479cea20a98b8a65503ccf3
                      tileentitydispenser.setItem(i, idispensebehavior.dispense(sourceblock, itemstack));
                  }
 diff --git a/src/main/java/net/minecraft/world/level/block/DropperBlock.java b/src/main/java/net/minecraft/world/level/block/DropperBlock.java
-index 1d13f8a1009d6eda351c697052d499d594a6aaa8..7cc7ee9d9cb0dd1b785e1b48a1296fd914ba14a4 100644
+index 1d13f8a1009d6eda351c697052d499d594a6aaa8..9a8a0fb958e8ec782111507bae957f854b2aac72 100644
 --- a/src/main/java/net/minecraft/world/level/block/DropperBlock.java
 +++ b/src/main/java/net/minecraft/world/level/block/DropperBlock.java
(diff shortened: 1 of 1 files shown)


patches/server/0504-Add-BlockPreDispenseEvent.patch
```diff
@@ -16,6 +16,18 @@ index 5593a0aa9e618071b6521b213dde0f628348c3dc..644e64850479cea20a98b8a65503ccf3
...
                  }
+diff --git a/src/main/java/net/minecraft/world/level/block/DropperBlock.java b/src/main/java/net/minecraft/world/level/block/DropperBlock.java
+index 1d13f8a1009d6eda351c697052d499d594a6aaa8..7cc7ee9d9cb0dd1b785e1b48a1296fd914ba14a4 100644
+--- a/src/main/java/net/minecraft/world/level/block/DropperBlock.java
++++ b/src/main/java/net/minecraft/world/level/block/DropperBlock.java
+@@ -65,6 +65,7 @@ public class DropperBlock extends DispenserBlock {
+                 ItemStack itemstack = tileentitydispenser.getItem(i);
+ 
+                 if (!itemstack.isEmpty()) {
++                    if (!org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockPreDispenseEvent(world, pos, itemstack, i)) return; // Paper - Add BlockPreDispenseEvent
```
Machine-Maker: This would fire the event when the dropper is moving the item into a different inventory. Do we want that? Or instead should this be fired directly before it's actually dispensed (which happens a few lines below this one).

I'm not 100% sure, but either way, it should be clarified.

Machine-Maker (changes requested):

patches/server/0504-Add-BlockPreDispenseEvent.patch
emilyy-dev: Yeah, I think it's sensible to call the pre dispense event right before the dispense behaviour is invoked. For the rest of the logic between inventories there is the inventory move item event

Machine-Maker (approved):

Issue #10223, opened 2024-02-05
Server JAR detection in plugins folder


kashike:
We should properly detect server JARs that have been put in the `plugins` folder and ignore loading + warn that they are there.

jstN0body:
What would be a good way to do this? My first thought was to look at the JAR manifest, but I don't know if that's the best strategy.

jpenilla:
The vanilla server jar & paper bundler/paperclip both have a `version.json` file at the root of the jar. NeoForge & Fabric installers won't have those but not sure if that's in scope of the issue anyways.

jstN0body:
That makes sense, but any thoughts on bukkit or spigot server jars?

molor:
Why not just load only `jar`s that contain `(paper-)plugin.yml` file?

jstN0body:
Paper already does that by default. However, there is no system to detect server jars in specific

jstN0body:
![image](https://github.com/PaperMC/Paper/assets/49841276/df86d051-afb4-4f6d-9527-5d768bb6f00e)
Output when first starting the server.
![image](https://github.com/PaperMC/Paper/assets/49841276/f0fb0b1b-afe6-41d9-9190-f7c76b78ba56)
Output when reloading the server. (notice that worldedit reloads with no issue)

Warriorrrr:
Merged in #10228

PR #10228, opened 2024-02-07
Check for server JARs in plugin directory

micah-rao:
Simple detection logic to address the issue #10223


patches/server/1045-Check-for-server-JARs-in-plugin-directory.patch
```diff
@@ -0,0 +1,27 @@
...
++                context + " is a server JAR and should be removed from the plugins folder."
++            );
++
++            throw new IllegalArgumentException(context + " is a server JAR");
++        }
++        // Paper end - Check for server JARs in plugin directory
++
+         PluginFileType<?, ?> type = PluginFileType.guessType(file);
+         if (type == null) {
+             throw new IllegalArgumentException(source + " does not contain a " + String.join(" or ", PluginFileType.getConfigTypes()) + "! Could not determine plugin type, cannot load a plugin from it!");
```
electronicboy: The logic this PR adds should probably be inside of here

micah-rao: Sorry if I'm just missing the obvious, but when you say "here", where exactly are you referencing?

jstN0body:
image
Output when first starting the server.
image
Output when reloading the server. (notice that worldedit reloads with no issue)

patches/server/1045-Check-for-server-JARs-in-plugin-directory.patch
electronicboy: inside of the `type == null` body, that way we only blow up just because a plugin bundles a version.json file inside of its meta folder

patches/server/1046-fixup-Check-for-server-JARs-in-plugin-directory.patch
```diff
@@ -0,0 +1,39 @@
...
+-
+         PluginFileType<?, ?> type = PluginFileType.guessType(file);
+         if (type == null) {
++            // Paper start - Check for server JARs in plugin directory
++            if (file.getEntry("META-INF/versions.list") != null) {
++                java.util.logging.Logger.getLogger("Minecraft").warning(
++                    context + " is a server JAR and should be removed from the plugins folder."
++                );
++
++                throw new IllegalArgumentException(context + " is a server JAR");
```
kashike, line 33: This should be non-fatal, and ignore the server JAR.

kashike (changes requested):

electronicboy (approved):
approving, but this should be merged into the paper plugins patch
maybe some better tongtangulation for the message too

electronicboy (approved):

PR #10229, opened 2024-02-07
async StoredUserList saving

wuangg:
Offload JSON serializing and file writing to another thread so as not to put a strain on the main thread performance when interacting with user lists that contain a lot of list entries.

### DIFF
--- a/patches/server/1046-Async-StoredUserList-saving.patch
+++ b/patches/server/1046-Async-StoredUserList-saving.patch
@@ -1,0 +1,200 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tran Nguyen Ngoc Quang <3d7777456@gmail.com>
+Date: Thu, 8 Feb 2024 04:44:10 +0700
+Subject: [PATCH] Async StoredUserList saving
+
+
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index d06185566b447c432d4dc2e3ba04d121bcdbc71b..4c460a5605b282b6ace7216128d1d6b231404d8e 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -979,6 +979,7 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         this.isSaving = true;
+         if (this.playerList != null) {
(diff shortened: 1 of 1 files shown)


lynxplay (changes requested):
Shooting off saves onto a thread pool with > 1 threads is a somewhat dangerous territory where you could potentially end up  with two threads trying to write to the whitelist.json file.

The SaveData IO async patch pulled prior created a single one thread large pool for this.
Moving this to a similar setup would be better imo.
Obviously then with proper logic to shutdown said new pool.

wuangg:
I have acknowledged about using a single thread executor for this right after I created this PR. I will do the required changes once I have free time.

electronicboy:
I would just use a shared executor between all of them, no need for them to each have their own

wuangg:
Yes, I should have gone with that solution as I wasn't thoughtful enough earlier. Thanks for the head up.

electronicboy (approved):

patches/server/1046-Async-StoredUserList-saving.patch
```diff
@@ -0,0 +1,192 @@
...
+             StoredUserList.LOGGER.warn("Could not save the list after removing a user.", ioexception);
+         }
+@@ -102,8 +109,14 @@ public abstract class StoredUserList<K, V extends StoredUserEntry<K>> {
+         return this.map.values();
+     }
+ 
+-    public void save() throws IOException {
++    // Paper start - async StoredUserList saving
++    static void shutdownExecutor() {
++        StoredUserList.SAVE_EXECUTOR.shutdown();
```
lynxplay: shutdown() does not await running tasks, we'll need to look into e.g. awaitTermination with a *plausible* timeout.

@electronicboy suggests waiting for 30 seconds.

wuangg: done.

patches/server/1046-Async-StoredUserList-saving.patch
```diff
@@ -0,0 +1,199 @@
...
+-    public void save() throws IOException {
++    // Paper start - async StoredUserList saving
++    static void shutdownExecutor() {
++        final java.util.concurrent.ExecutorService executor = StoredUserList.SAVE_EXECUTOR;
++
++        executor.shutdown();
++        try {
++            executor.awaitTermination(30L, java.util.concurrent.TimeUnit.SECONDS);
++        } catch (java.lang.InterruptedException ex) {
++            executor.shutdownNow();
```
lynxplay, line 164: You'll want to re-raise the interrupt flag here via
`Thread.currentThread().interrupt();` after the shutdownNow() call.

PR #10235, opened 2024-02-10
Propagate failed to bind to port error

Warriorrrr:
Throws a new IllegalStateException with the failed to bind to port exception as the cause, this will make it so a more helpful stacktrace is included in the crash report which is helpful when someone comes for support.


patches/server/1049-Propagate-failed-to-bind-to-port-error.patch
```diff
@@ -0,0 +1,18 @@
...
+
+diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+index 6de6dad36203479677a29ad61e21bc369d4e5513..2ecf79086e9e8069f43120e9146e98962a75eca8 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -265,6 +265,7 @@ public class DedicatedServer extends MinecraftServer implements ServerInterface
+             DedicatedServer.LOGGER.warn("**** FAILED TO BIND TO PORT!");
+             DedicatedServer.LOGGER.warn("The exception was: {}", ioexception.toString());
+             DedicatedServer.LOGGER.warn("Perhaps a server is already running on that port?");
++            if (true) throw new IllegalStateException("Failed to bind to port", ioexception); // Paper - Propagate failed to bind to port error
```
Leguan16: Just out of curiosity, why do you check `if (true)` here?

Warriorrrr: To avoid it from complaining about the return being unreachable, I thought it'd be better to do it like so to make it only 1 line of diff but it probably doesn't matter much

Leguan16: Ohh yeah that makes sense. Thanks for clarification :)

Machine-Maker (approved):

electronicboy:
this one has been on the eternal todo list for eons

electronicboy (approved):

lynxplay (approved):

PR #10237, opened 2024-02-10
Add RayTraceConfiguration

### LINKED ISSUE #9942: Allow changing the raySize of LivingEntity#rayTraceEntities

Allow changing the raySize of LivingEntity#rayTraceEntities
This will never include Displays because Displays have a hitbox of 0.

Allowing the definition of the raySize so developers can include Displays.

Requiring Spigot's long method
```
result = origin.getWorld().rayTraceEntities(origin.getEyeLocation(), location.getDirection(), 100, 1.0D, entity -> !entity.equals(origin));
```

### Other

I told Spigot about this, and they're adding a JavaDoc to `World#rayTraceEntities` to tell developers to define the raySize for Displays. Rather than the method that inputs 0.0D for the raySize.

notTamion:
closes #9942 
Adds a RayTraceBuilder which can be constructed using a starting Location, direction and max distance.
I am currently still unsure of how to integrate the raytrace methods on Block. I would definitely love to hear some feedback on that.
post hardfork #11907


Machine-Maker (changes requested):
I think this should be setup in a way similar to adventure's JoinConfiguration. interfaces+builder that are then used in a method on World rather than calling some method on the builder itself to get the ray trace result.

notTamion:
Could you give me an example explanation specifically for RayTrace in mind? (i don't exactly understand what you mean) do you mean like a RayTrace class that can be constructed which holds all the info about the direction etc. and then gets passed into the raytrace method on World, LivingEntity etc.?

notTamion:
what should be done regarding LivingEntity or Block? those interfaces only have methods for raytracing with the target of there own type

notTamion:
Alright i am really not sure about this. I feel like the kind of RayTraceBuilder or RayTraceExecutor we had in the beginnning just feels better. Using a configuration where stuff like starting point, distance, direction all need to be adjusted for each use specificly doesn't seem very convenient. I guess in the example of the JoinConfiguration that just works better cause not that many variables need to be adjusted for each use.

PR #10241, opened 2024-02-11
[ci skip] Update Contributing.md

notTamion:
Since Windows Defender fucking with Intellij and WSL performance is a very common issue, i think the fix which worked for a lot of people (including me) should be mentioned somewhere. Perhaps we should also add an warning afterwards telling people to be careful with what they add to there AV exclusions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,18 +4,18 @@ PaperMC is happy you're willing to contribute to our projects. We are usually
 very lenient with all submitted PRs, but there are still some guidelines you
 can follow to make the approval process go more smoothly.
 
-## Use a Personal Fork and not an Organization
+## Branch Protection
 
 Paper will routinely modify your PR, whether it's a quick rebase or to take care
 of any minor nitpicks we might have. Often, it's better for us to solve these
 problems for you than make you go back and forth trying to fix them yourself.
 
-Unfortunately, if you use an organization for your PR, it prevents Paper from
-modifying it. This requires us to manually merge your PR, resulting in us
-closing the PR instead of marking it as merged.
+Unfortunately, if you use an organization or the master branch for your PR, 
+it prevents Paper from modifying it. This requires us to manually merge your
+PR, resulting in us closing the PR instead of marking it as merged.
 
 We much prefer to have PRs show as merged, so please do not use repositories
-on organizations for PRs.
+on organizations or the master branch for PRs.
 
 See <https://github.com/isaacs/github/issues/1681> for more information on the
 issue.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -485,13 +488,17 @@ using an out of date version of Windows 10, update your system with the
 To set up WSL 2, follow the information here:
 <https://docs.microsoft.com/en-us/windows/wsl/install>
 
+To avoid issues with Intellij and WSL 2 you may also want to add the processes `fsnotifier-wsl`
```
Doc94:
```suggestion
To avoid issues with IntelliJ and WSL 2 you may also want to add the processes `fsnotifier-wsl` 
```

Doc94:
Maybe the new info about the Windows Defender can be in a "info" box before the caution about mnt? also mention this only apply for people using Windows Defender and not has all this disabled.

MiniDigger:
doesnt intellij have a popup that prompts you to do this (and offers to do it automatically if you are a priviledged user)

notTamion:
it does have a popup to add the project directory to the exclusions however in that case that is not sufficient

QuanTrieuPCYT (approved):

Doc94 (approved):

CONTRIBUTING.md
```diff
@@ -4,18 +4,18 @@ PaperMC is happy you're willing to contribute to our projects. We are usually
...
+## Branch Protection
 
 Paper will routinely modify your PR, whether it's a quick rebase or to take care
 of any minor nitpicks we might have. Often, it's better for us to solve these
 problems for you than make you go back and forth trying to fix them yourself.
 
-Unfortunately, if you use an organization for your PR, it prevents Paper from
-modifying it. This requires us to manually merge your PR, resulting in us
-closing the PR instead of marking it as merged.
+Unfortunately, if you use an organization or the master branch for your PR,
```
jpenilla: We close PRs from master as it complicates the dev workflow, not related to the write issue with organization forks.

notTamion: Been a while since i took a look at this pr. what exactly was the deal with master branch again? iirc you guys have to change something in gh settings to be a able to push a forks master? could you or anyone else please elaborate on that.

electronicboy: I think github used to create issues with it, no idea if that's still an issue or not; but, a common issue is that it often creates issues for people who are unfamiliar with git when stuff diverges, things like managing more than 2 PRs, etc

notTamion: does something like "you also shouldn't create PRs from the master branch as this is bad practice and may cause problems in the dev workflow" (not final just a rough idea) sound good?

notTamion:
should be more accurate now

Doc94:
> should be more accurate now
> 
> EDIT: Also i think it would be worth adding an index to the Contributing.md as especially new contributors might want to search up specific topics mulitple times

Github show a index buy you need click an icon for that if im not wrong

notTamion:
> Github show a index buy you need click an icon for that if im not wrong

oh yeah forgot that existed

PR #10245, opened 2024-02-12
add overloads to use suspicious effect entry to mushroom cow and suspicious stew meta

### LINKED ISSUE #9749: SuspiciousStewMeta does not respect other values on PotionEffect
### Expected behavior

The type used in SuspiciousStewMeta (`PotionEffect`) accurately conveys the information stored in the meta 

### Observed/Actual behavior

`PotionEffect` has 4 additional pieces of information that are not persisted in the itemstack nbt, only the type and duration are. This is not mentioned in the docs, but beyond that I don't think the use of PotionEffect is correct. See below.

### Steps/models to reproduce

N/A

### Plugin and Datapack List

N/A

### Paper version

9624451ffd9f9db8840f538e1a926e02d64dce48

### Other

I don't think adding documentation mentioning this is sufficient. I think deprecating existing methods with PotionEffect and using a separate type is appropriate. Whatever the solution is, the same thing should then be used for fixing the Mushroom Cow multiple effects API. That has the same issue. Perhaps the type can be an interface that PotionEffect extends or something, but usually the parameter type is an interface that best encapsulates the data being used and when `PotionEffect` is used elsewhere, all parts are used.

yannicklamprecht:
The existing method with PotionEffect suggests that all attributes are used. In fact, only the PotionEffectType and the duration are used.

resolves #9749 

TODO:

- [x] expand on StewMeta if not done already
- [x] deprecate other PotionEffect methods
- [x] write a chunk full of JavaDoc
- [x] add Paper comments


kennytv:
get and set already exist at the bottom of the class, so their naming should be kept and the upstream ones deprecated

yannicklamprecht:
> get and set already exist at the bottom of the class, so their naming should be kept and the upstream ones deprecated

Yeah, for the get call I just forgot to remove it. Realized it ans stopped implementing the set call.
Names are kept. Would either migrate the remove calls to using `Predicate<SuspiciousEffectEntry> ` to support only filtering using one or the other. Currently the original PotionEffectType overload filters by PotionEffectType and the SuspicousEffectEntry one by PotionEffect and Duration.
After the Predicate one is implemented we can deprecate upstream one for remove.

patches/api/0463-add-overloads-to-use-suspicious-stew-entry-to-mushro.patch
```diff
@@ -0,0 +1,49 @@
...
+@@ -40,8 +42,13 @@ public interface MushroomCow extends Cow, io.papermc.paper.entity.Shearable { //
+      * @return true if the effects to be applied to the suspicious stew changed
+      * as a result of this call
+      */
++    @Deprecated
+     boolean addEffectToNextStew(@NotNull PotionEffect effect, boolean overwrite);
+ 
++    boolean addEffectToNextStew(@NotNull SuspiciousEffectEntry suspiciousEffectEntry, boolean overwrite);
++
++    boolean hasEffectForNextStew(@NotNull SuspiciousEffectEntry suspiciousEffectEntry);
```
Machine-Maker: Do we need a `has` method for the SuspiciousEffectEntry? I think the one with PotionEffectType is enough. This is just a helper for getEffects().contains().

patches/api/0463-add-overloads-to-use-suspicious-stew-entry-to-mushro.patch
```diff
@@ -0,0 +1,49 @@
...
++
+     /**
+      * Removes a custom potion effect from being applied to the next suspicious
+      * stew received from milking this {@link MushroomCow}.
+@@ -52,6 +59,8 @@ public interface MushroomCow extends Cow, io.papermc.paper.entity.Shearable { //
+      */
+     boolean removeEffectFromNextStew(@NotNull PotionEffectType type);
+ 
++    boolean removeEffectFromNextStew(@NotNull SuspiciousEffectEntry suspiciousEffectEntry);
```
Machine-Maker: I think a similar thing applies here, where we don't necessarily need a remove method that takes an entry.

yannicklamprecht: Depends on the requirement. Do we ever want to remove a potion effect type with a specific duration and keep the same potion effect types with different duration?

Owen1212055 (approved):

PR #10253, opened 2024-02-16
Add server version util class

masmc05:
Currently, it's impossible to get the information about the server from bootstrap, as `Bukkit.getServer()` is null.

The goal of this pr is to allow an easy access to information about the server, even from bootstrap.

Here's an example usage:
```java
    // import static io.papermc.paper.util.ServerInfo.serverInfo;
    @Override
        var log = context.getLogger();
        if (serverInfo().is("1.20.4")) {
            log.info("Server is running 1.20.4");
        } else {
            log.info("Server is not running 1.20.4");
        }
        if (serverInfo().is("1.20.3")) {
            log.info("Server is running 1.20.3");
        } else {
            log.info("Server is not running 1.20.3");
        }
        if (serverInfo().isAtLeast("1.20.3")) {
            log.info("Server is running 1.20.3 or newer");
        } else {
            log.info("Server is not running 1.20.3 or newer");
        }
        if (serverInfo().isAtLeast("1.20.6")) {
            log.info("Server is running 1.20.6 or newer");
        } else {
            log.info("Server is not running 1.20.6 or newer");
        }
        if (serverInfo().isAtLeast("1.20.7")) {
            log.info("Server is running 1.20.7 or newer");
        } else {
            log.info("Server is not running 1.20.7 or newer");
        }
        if (serverInfo().isAtLeast("1.20")) {
            log.info("Server is running 1.20 or newer");
        } else {
            log.info("Server is not running 1.20 or newer");
        }
        if (serverInfo().isAtLeast("1.21")) {
            log.info("Server is running 1.21 or newer");
        } else {
            log.info("Server is not running 1.21 or newer");
        }
        if (serverInfo().isImplementing(Key.key("papermc", "folia"))) {
            log.info("Server implements folia");
        } else {
            log.info("Server does not implement folia");
        }
        //Logically, this should always return true
        if (serverInfo().isImplementing(Key.key("papermc", "paper"))) {
            log.info("Server implements paper");
        } else {
            log.info("Server does not implement paper");
        }
        log.info("Server version: {}", serverInfo().version());
        log.info("Server version name: {}", serverInfo().versionName());
        log.info("Bukkit version: {}", serverInfo().apiVersion());
        log.info("Server brand: {}", serverInfo().serverBrand());
    }
```
which results in those outputs (Bukkit version remained the same since it's unknown how it will be in reality for a snapshot etc, just changed the version.json file to ones provided in other versions)
For 1.20.6: 
```css
[22:46:04 INFO]: [Paper-Test-Plugin] Server is not running 1.20.4
[22:46:04 INFO]: [Paper-Test-Plugin] Server is not running 1.20.3
[22:46:04 INFO]: [Paper-Test-Plugin] Server is running 1.20.3 or newer
[22:46:04 INFO]: [Paper-Test-Plugin] Server is running 1.20.6 or newer
[22:46:04 INFO]: [Paper-Test-Plugin] Server is not running 1.20.7 or newer
[22:46:04 INFO]: [Paper-Test-Plugin] Server is running 1.20 or newer
[22:46:04 INFO]: [Paper-Test-Plugin] Server is not running 1.21 or newer
[22:46:04 INFO]: [Paper-Test-Plugin] Server does not implement folia
[22:46:04 INFO]: [Paper-Test-Plugin] Server implements paper
[22:46:04 INFO]: [Paper-Test-Plugin] Server version: 1.20.6
[22:46:04 INFO]: [Paper-Test-Plugin] Server version name: 1.20.6
[22:46:04 INFO]: [Paper-Test-Plugin] Bukkit version: 1.20.6-R0.1-SNAPSHOT
[22:46:04 INFO]: [Paper-Test-Plugin] Server brand: Paper
```


TheMeinerLP:
Wouldn't it make more sense to transfer this Check Logic into the PaperLib and rather return Try Version String and Protocol Version into the bootstrap context ?

So this is just a suggestion

Sytm:
I think this stuff fits best into the server, because only the server can reliably tell what it implements. The checking for classes (what PaperLib does) might not properly reflect everything depending on what is checked

TheMeinerLP:
> I think this stuff fits best into the server, because only the server can reliably tell what it implements. The checking for classes (what PaperLib does) might not properly reflect everything depending on what is checked

Well, if PaperLib can test on the basis of given things. Bootstrap context or other things. You don't have to do anything with Reflections if the correct values are given. I see such a version check more as an ultitiy than that it should be part of a server software. But that's more my perspective.
The other side of it would be that developers would have it easier.
But is it worth the maintainability?

masmc05:
> I believe the concept here is great, I am currently unsure if the impl and api methods are the best tho.
> 
> A) int varargs seems weird. This would lock the API to only support major versions, which I am unsure is something we want to commit to. B) I like `implement` (tho it should be `implements`). This seems especially useful for forks as you pointed out, e.g. purpur or folia. Is a single string there enough or should we force some form of namespace? I presume the chances of a second fork named purpur that does not actually supply purpur apiare slim to none, but yea, food for thought.

A) I will think how I could change that

B) Sadly, "implements" is a reserved word by java, that's why it's "implement". And what are the guarantees that that other fork won't use the same namespace? If someone named their fork exactly the same as other fork in my opinion it's their fight, I made this method mostly for popular forks, like folia, since if a plugin will make a special support for a fork, it's really not likely it will make it for a small unknown fork.

lynxplay:
kek, yea you are right. flame my tired brain. `implement` is still not nice. `offers` ? `providesAPI`.

masmc05:
> But is it worth the maintainability?

You can look at the implementation, it's based on 3 information from the nms
1. **The version**. From 1.13, nms has a class named SharedConstants, which always had the version, but they may often change how to get it from that class, if they change it a bit, it won't be hard to find the new method in the SharedConstants to get the version, but it breaks compatibility for the plugins using reflection
2. **The api version**. Well, this depends on Versioning class, which didn't change for the last 12 years, i don't think this will bring any trouble
3. **The server name**. I don't think someone ever will want to touch this... And it's based on an own class, so it's independent from future changes, also here goes the check for implemented apis
4. **The UnsafeValues**. It's the same unsafe as in Bukkit#getUnsafe, if people depend on it, they're on their own, so this doesn't have any maintainability burden at all

Also this pr doesn't modify any classes, just adds some new, so it's reaaaalllyyy cheap to maintain.


> Wouldn't it make more sense to transfer this Check Logic into the PaperLib

Why should I use paper lib if I only support paper, in a paper plugin, just to check the version.....

masmc05:
> kek, yea you are right. flame my tired brain. `implement` is still not nice. `offers` ? `providesAPI`.

Maybe then `hasAPI`?

Leguan16:
What about `isImplementingAPI`?

masmc05:
> What about `isImplementingAPI`?

I think `if (PaperServerInfo.isImplementing("Folia"))` sounds as the best solution

Leguan16:
But I also think that some sort of name spacing wouldn't be bad. It looks weird to provide a String imo.

vadcx:
I'm pleased to see the discussion taking place here, let me respond to respond points first:

> The server name. I don't think someone ever will want to touch this...

But isn't Paper as much Spigot (the server, not the plugin), as it is CraftBukkit?

> Sadly, "implements" is a reserved word

The point here is feature-based interaction and enablement. I strongly believe this should be called **.supports()** or similar. Supports Folia? Yes. Suports X-Y patch from 1 year ago? No. Supports a sub-feature of a patch from 9 months ago? Yes.

> should we force some form of namespace?

If the API operates on strings, idk how do you want to enforce it? Change the API to explicitly require a parent namespace key + child key? This is little different from a convention calling features like "paper.versioningapi". This then will be similar to how browsers implement experimental/proprietary CSS features by adding a -webkit or -moz prefix.

Versioning of the Features? If this idea evolves to a feature-support API, at some point it'll be useful and necessary to retrieve a feature's individual version number. The game/server revision *could* be used for it, but it's more work and less reliable than knowing a certain patches' exact version.

"Just add a more specific key to a patch" is not the complete answer, because there can be a patch gap between a behavioral change/feature's implementation and the understanding that it warrants a new, more specific feature key.

---

Although I'm surprised by vararg version ints too, my question is different. How to work with and represent weekly snapshots? And the absolute edge case, special editions like April 1st (ok this is over the top).

For "isVersion" I would add CAPS docs to instead use greater/lower than for versioning (especially if this PR moves toward capability-based API queries). Because isVersion should be understood as hardcoding an exact version in the majority of cases.

----

I find this is a better (and more gentle) solution for a seamless multi-version and multi-vendor support. It'd be better placed in upstream (Spigot).

patches/server/1046-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,102 @@
...
++                return a > b;
++            }
++        }
++        return true;
++    }
++
++    @Override
++    public boolean implement(final @NotNull String api) {
++        Preconditions.checkNotNull(api, "The API cannot be null");
++        return this.implementedAPIs.contains(api.toUpperCase(Locale.ROOT));
```
vadcx: I suggest to turn it lower case instead. Most will type in lower case and it's more likely to align with potential duplicate strings for interning. Not even important in this case, but I wanted to comment still.

patches/server/1046-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,102 @@
...
++    }
++
++    @Override
++    public boolean isAtLeast(final int @NotNull ... version) {
++        int maxLen = Math.max(this.version.length, version.length);
++        for (int i = 0; i < maxLen; i++) {
++            int a = i < this.version.length ? this.version[i] : 0;
++            int b = i < version.length ? version[i] : 0;
++            if (a != b) {
++                return a > b;
```
vadcx: I would rename a,b to "ourVer", "theirVer” / ”requestedVer" for extra clarity.

masmc05:
> Versioning of the Features? If this idea evolves to a feature-support API, at some point it'll be useful and necessary to retrieve a feature's individual version number. The game/server revision could be used for it, but it's more work and less reliable than knowing a certain patches' exact version.

I think this will bring too much trouble to maintain, all changes to features are made to have backwards compatibility, so you may just need to check for the version + fork api presence to see if the new feature capability is here

> But isn't Paper as much Spigot (the server, not the plugin), as it is CraftBukkit?

As I see it, very soon the hard fork will happen, then it will be very hard to say that the server is implementing the spigot/bukkit api, adding something with the scope of removing it very soon is not a bright idea in my opinion

> The point here is feature-based interaction and enablement.

The point is the implemented fork api, as novice developers often may forget to store the value if the server is folia, meaning they will have often Class#forName calls, which are expensive, this new method should be safe to recommend in docs and other places for new developers. Also this is a way easier method than to find the class and copy it for class check.

> Although I'm surprised by vararg version ints too, my question is different. How to work with and represent weekly snapshots? And the absolute edge case, special editions like April 1st (ok this is over the top).

I thought about changing PaperServerInfo#is to a string which just checks for equals with current version and PaperServerInfo#isAtLeast will just mean if the current version appeared after the specified release version or if it's the current release version

> I find this is a better (and more gentle) solution for a seamless multi-version and multi-vendor support than reflection/exceptions. It'd be better placed in upstream (Spigot).

Personally, I don't want to care about the spigot api, if people want to still support spigot, then they may continue to just deal with the missing api like this, I didn't support spigot in my plugins for a long time and recommend others to do the same (even though I understand that private plugins are a lot easier to make without spigot support than public/paid plugins)

masmc05:
Made a few changes which we discussed here, like using a key to check for implemented api and support for non-release versions.

The support for snapshots/pre-releases could be made better but I don't think that it's an environment for which we may want to spend too much effort to give the best support possible. But at least this pr already defines a possible future behaviour that developers may expect, and which can be further discussed here.

DerEchtePilz:
I really like this concept but I am unsure if the class should be named `PaperServerInfo`. Why not name it `ServerInfo`?

masmc05:
> I really like this concept but I am unsure if the class should be named `PaperServerInfo`. Why not name it `ServerInfo`?

Idk, PaperServerInfo sounds better for me, but if someone else agrees then i can change that to ServerInfo

DerEchtePilz:
> Idk, PaperServerInfo sounds better for me, but if someone else agrees then i can change that to ServerInfo

Well, I think for Paper it's fine either way but I was much more thinking of forks for example. Sure, they can patch that name again but naming it ServerInfo is much more generic than PaperServerInfo.
I don't know, these are just my thoughts about this..

Leguan16:
I agree with DerEchtePilz on this one. `PaperServerInfo` seems more like a implementation name than an API name. The whole bukkit API is meant to be generic so I would not stop here and start naming things after Paper. Though maybe a different name then `ServerInfo` can be found. IMO it doesn't really fit. Since it mostly is about the version maybe call it `VersionInfo` or generally just `Version`.

masmc05:
> I agree with DerEchtePilz on this one. `PaperServerInfo` seems more like a implementation name than an API name. The whole bukkit API is meant to be generic so I would not stop here and start naming things after Paper. Though maybe a different name then `ServerInfo` can be found. IMO it doesn't really fit. Since it mostly is about the version maybe call it `VersionInfo` or generally just `Version`.

"VersionInfo" or "Version" already does not go well with isImlementing, also limits future possibilities to add utility methods

Who knows what other information about the server people may want to get in the bootstrap, naming it "ServerInfo" will give green light to a lot of  read only operations that may be invoked before CraftServer exists, I just added mostly only version methods because that's the most used thing right know but which is inaccessible in bootstrap, and don't think other methods may fit well in this pr

I'll rename the class to "ServerInfo" these days when I'll have a bit of free time

patches/server/1048-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,211 @@
...
++        private final String id = SharedConstants.getCurrentVersion().getId();
++        private final List<String> versionList;
++        private NonStableVersionInfo() {
++            HttpClient client = HttpClient.newHttpClient();
++            HttpRequest request = HttpRequest.newBuilder()
++                .uri(URI.create(manifestURL))
++                .build();
++            List<String> versionList = new ArrayList<>();
++            try {
++                HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
```
yannicklamprecht: Http requests should not be done in a ctor as it is a relativly heavy operation. It should be extracted into a factory method or better generated during build time as the currently running unstable version can only support version < current unstable version that has been built.


So a check of isAtleast("1.20.4") against 24w09a would succeed cause 1.20.4 is in the json array.
isAtLeast("1.20.5") doesn't succeed against 24w09a as the value is not contained in the built json array.
It basically doesn't make sense to call the endpoint on every startup if the minecraft version doesn't change without replacing the jar.

yannicklamprecht (changes requested):

patches/server/1048-Add-paper-version-util-class.patch
masmc05: As I said before, it's mostly to give idea how it will work on snapshots, i don't want to go deeper and change the build script of the server to work with this, and don't think that paper supporting snapshots will come soon either

And even if it does, snapshot version will most probably be unstable anyways, and mostly for devs to start updating their plugins, a snapshot paper shouldn't ever be on production, so this code won't be on a production server

masmc05: When paper will support snapshots and prereleases, only then it will really possible to look at how this could be made properly, but now it's too unknown how that snapshot support from paper will even look like to be able to think how can this be better

patches/api/0464-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++import net.kyori.adventure.key.Key;
++import net.kyori.adventure.util.Services;
++import org.bukkit.UnsafeValues;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * A utility class to get information about the server
++ * Works even before Bukkit is initialized (e.g. on bootstrap)
++ */
++public class ServerInfo {
```
Leguan16: Is there a reason why you don't just make the class abstract?

patches/server/1048-Add-paper-version-util-class.patch
yannicklamprecht: Then at least move the data fetching into a static factory method, instantiate the unstable versioninfo using the fetched data and return that.

patches/api/0464-Add-paper-version-util-class.patch
masmc05: Why would I?

patches/server/1048-Add-paper-version-util-class.patch
masmc05: > Then at least move the data fetching into a static factory method, instantiate the unstable versioninfo using the fetched data and return that.

Could you show a pseudo code of what you mean? I just don't understand the difference from what it's now, as we still will need to block the object initialization to get the version data. But I highly suspect that I just don't think correctly of what you mean with that change.

patches/api/0464-Add-paper-version-util-class.patch
Leguan16: I don't know. I just thought it would look cleaner since you block creating a new instance anyway. Maybe I am just misunderstanding something.

patches/server/1048-Add-paper-version-util-class.patch
yannicklamprecht:
```suggestion
diff --git a/src/main/java/io/papermc/paper/util/misc/ServerInfoProviderImpl.java b/src/main/java/io/papermc/paper/util/misc/ServerInfoProviderImpl.java
index 1364afc507..bacc415fbe 100644
--- a/src/main/java/io/papermc/paper/util/misc/ServerInfoProviderImpl.java
+++ b/src/main/java/io/papermc/paper/util/misc/ServerInfoProviderImpl.java
@@ -43,7 +43,7 @@ public class ServerInfoProviderImpl implements ServerInfoProvider {
         } else {
             // Snapshot, pre-release, etc
             // Supported the best we can
-            this.version = new NonStableVersionInfo();
+            this.version = NonStableVersionInfo.create();
         }
     }
     @Override
@@ -147,7 +147,12 @@ public class ServerInfoProviderImpl implements ServerInfoProvider {
         private final String version = SharedConstants.getCurrentVersion().getName();
         private final String id = SharedConstants.getCurrentVersion().getId();
         private final List<String> versionList;
-        private NonStableVersionInfo() {
+
+        private NonStableVersionInfo(List<String> versionList) {
+            this.versionList = ImmutableList.copyOf(versionList);
+        }
+
+        private static NonStableVersionInfo create() {
             HttpClient client = HttpClient.newHttpClient();
             HttpRequest request = HttpRequest.newBuilder()
                 .uri(URI.create(manifestURL))
@@ -162,8 +167,9 @@ public class ServerInfoProviderImpl implements ServerInfoProvider {
                 }
             } catch (IOException | InterruptedException e) {
                 LOGGER.error("Failed to fetch version manifest, plugins depending on version check may not work correctly", e);
+                return NonStableVersionInfo(List.of());
             }
-            this.versionList = ImmutableList.copyOf(versionList);
+            return new NonStableVersionInfo(versionList);
         }
         @Override
         public boolean is(final int @NotNull ... version) {

```

masmc05: Ok, will change to that

patches/api/0464-Add-paper-version-util-class.patch
masmc05: Yeah, seems like you misunderstood something

masmc05:
Rebased on 1.20.5

Leguan16:
Wouldn't this be a great addition to the removal of relocation to provide a straightforward way of getting the version? I mean it would only work from 1.20.(5/6)+ but I guess it's good to include both in the same version so there is a nice and modern alternative.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++import net.kyori.adventure.key.Key;
++import net.kyori.adventure.util.Services;
++import org.bukkit.UnsafeValues;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * A utility class to get information about the server
++ * Works even before Bukkit is initialized (e.g. on bootstrap)
++ */
++public class ServerInfo {
```
lynxplay: Given how little this class should be used in normal developer workflow, I think it is fine to make this an interface,
expose a single static `instance()` method or whatever, and then return.

Saves us from double declaring everything.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++    private ServerInfo() {
++        throw new UnsupportedOperationException("This class cannot be instantiated");
++    }
++    private static final ServerInfoProvider provider = Services.service(ServerInfoProvider.class).orElseThrow();
++
++    /**
++     * Get the version of the server
++     * @return the version of the server (e.g. "1.20.4", "1.20.2 Pre-release 2", "23w31a")
++     */
++    public static @NotNull String version() {
```
lynxplay: I think #version should yield the id, not the name.
The name is a lot less usable for any programmatic purpose given its spaces and uncertain format.
The id seems like a more useful version string to default on.

We can certainly expose a #versionName()  or something.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++     */
++    public static @NotNull String apiVersion() {
++        return provider.apiVersion();
++    }
++
++    /**
++     * Get the name of the server
++     * @return the name of the server (e.g. "Paper")
++     */
++    public static @NotNull String serverName() {
```
lynxplay: Is this needed, given we have the isImplementing logic?

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++        return provider.serverName();
++    }
++
++    /**
++     * Returns true if the server is a minecraft release version,
++     * false is it's a snapshot, pre-release, etc
++     * @return if the server is on a release version
++     */
++    public static boolean isRelease() {
```
lynxplay: Not needed for non-snapshot version of this PR.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++        return provider.isRelease();
++    }
++
++    /**
++     * Returns the unsafe values for the server for unsafe version values
++     * @return the unsafe values for the server
++     */
++    @Deprecated
++    @NotNull
++    public static UnsafeValues unsafe() {
```
lynxplay: I think there is little point in exposing this.
Instead, expose protocol and data version on this (soon) interface.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++    }
++
++    /**
++     * Checks if the server runs exactly the specified version
++     * <p>
++     * Note: Will always return false on non release versions
++     * @param version the version to check (e.g. 1, 20, 4)
++     * @return true if the server runs exactly the specified version
++     */
++    public static boolean is(int @NotNull... version) {
```
lynxplay: See below, I don't think this is needed.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++    public static boolean is(@NotNull String version) {
++        return provider.is(version);
++    }
++
++    /**
++     * Checks if the server runs a version which is after this version or the same
++     * @param version the version to check (e.g. 1, 20, 4)
++     * @return true if the server runs on this version or a newer version
++     */
++    public static boolean isAtLeast(int @NotNull... version) {
```
lynxplay: The int varargs locks the `is` method into only being useful for releases.
Given we might want to release dev builds for snapshot, this seems unfortunate.
I think it is better to accept a String here and do the parsing ourselfs.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++        return provider.is(version);
++    }
++
++    /**
++     * Checks if the server runs exactly the specified version
++     * @param version the version to check (e.g. "1.20.4", "1.20.2 Pre-release 2", "1.20.2-pre2", "23w31a")
++     * @return true if the server runs exactly the specified version
++     */
++    public static boolean is(@NotNull String version) {
++        return provider.is(version);
```
lynxplay: I think we are fine with just taking the id here.
The name of a version is not really a format I want to rely on for a method contract.

lynxplay (changes requested):
So after a bit of discussion on this, I think we'd be (for now) best of to only
pull changes regarding full release versions.

Snapshot handling is complex, specifically with isAtLeast.
The commits in this PR will keep the logic around, so when we do get to releasing paper for snapshots they can serve as starting point.

Beyond that, see below for changes for request.
Again, thank you for the pr :partying_face:

patches/api/0477-Add-paper-version-util-class.patch
masmc05: This still could be used in log messages and other information dumps for further debugging

patches/api/0477-Add-paper-version-util-class.patch
masmc05: Ok, will change that

Just I felt that protocol and data versions are a bit just magic values that plugins may find some use for them even on bootstrap, but shouldn't be encouraged to use and should find alternatives, that's why chosed to expose them trough unsafe

patches/api/0477-Add-paper-version-util-class.patch
lynxplay: Hmmm, yea convincing argument 🙏

patches/api/0477-Add-paper-version-util-class.patch
lynxplay: Alternatively we don't expose protocol version here but simply expose a static getting on the unsave instance or something. E.g. UnsaveValues.get().protocol version()

Exposing it here just feels a bit off, but I agree that I guess protocol and data version might be pretty magic.

patches/api/0477-Add-paper-version-util-class.patch
masmc05: supporting strings here as arguments makes the logic as complex as it was for snapshot server handling, as if we are on 1.20.4, it's harder to say if it is older than random 99w99a, the same difficulty when it was vice versa

I don't think plugins will keep support for multiple snapshots or very old snapshot versions even when paper will release dev builds for snapshots, and if they will keep support for one snapshot version, they have #is(String) method to check for it

lynxplay: Yea I agree it'll be complex down the line to support snapshots, the parsing logic and manifest comparisons will be a bit of work but I don't think it's worth locking us out of this method working for snapshot comparison.

Even without snapshot support, int varags for this is a less than ideal parameter as `isAtLeast()` would then be valid.

As said in the general review, we are not currently supporting snapshots and hence reducing this pr to release-only while keeping us open to one day add snapshot support would be the goal.

For this, I think parsing a string passed here should be fine with the restriction that this method only takes stringified release versions, e.g 1.20.6 but not 1.20.6-rc1/24w18a.

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++    public static @NotNull String apiVersion() {
++        return provider.apiVersion();
++    }
++
++    /**
++     * Get the name of the server
++     * @return the name of the server (e.g. "Paper")
++     */
++    public static @NotNull String serverName() {
++        return provider.serverName();
```
kashike:
```suggestion
+     * Get the brand of the server
+     * @return the brand of the server (e.g. "Paper")
+     */
+    public static @NotNull String serverBrand() {
+        return provider.serverBrand();
```

patches/api/0477-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,188 @@
...
++     * Get the api version of the server
++     * @return the api version of the server (e.g. "1.20.4-R0.1-SNAPSHOT")
++     */
++    @NotNull String apiVersion();
++
++    /**
++     * Get the name of the server
++     * @return the name of the server (e.g. "Paper")
++     */
++    @NotNull String serverName();
```
kashike:
```suggestion
+     * Get the brand of the server
+     * @return the brand of the server (e.g. "Paper")
+     */
+    @NotNull String serverBrand();
```

patches/server/1046-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,215 @@
...
++    }
++
++    @Override
++    public @NotNull String apiVersion() {
++        return this.bukkitVersion;
++    }
++
++    @Override
++    public @NotNull String serverName() {
++        return this.serverName;
```
kashike:
```suggestion
+    public @NotNull String serverBrand() {
+        return this.serverBrand;
```

patches/server/1046-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,215 @@
...
++import java.util.Objects;
++import java.util.Optional;
++import java.util.Set;
++import java.util.concurrent.CompletableFuture;
++import java.util.stream.Stream;
++
++public class ServerInfoProviderImpl implements ServerInfoProvider {
++    private final String bukkitVersion = Versioning.getBukkitVersion();
++    private final Set<Key> implementedAPIs;
++    private final String serverName;
```
kashike:
```suggestion
+    private final String serverBrand;
```

patches/server/1046-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,215 @@
...
++    private final VersionInfo version;
++    public ServerInfoProviderImpl() {
++        ServerImplementationInfo info = new ServerImplementationInfo();
++        this.serverName = info.latestName;
```
kashike:
```suggestion
+        this.serverBrand = info.latestBrand;
```

kashike (changes requested):

patches/api/0477-Add-paper-version-util-class.patch
masmc05: Will expose in a separate pr

masmc05:
Applied all requested changes

patches/api/0479-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,86 @@
...
++    /**
++     * Get the brand of the server
++     * @return the brand of the server (e.g. "Paper")
++     */
++    @NotNull String serverBrand();
++
++    /**
++     * Checks if the server runs exactly the specified version
++     * @param version the version to check (e.g. "1.20.4", "1.20.2-pre2", "23w31a")
++     * @return true if the server runs exactly the specified version
```
Machine-Maker, line 69: Do an `@see` to the `#version` method to say the 2 values are "linked"

masmc05:
Fixed the javadoc which failed the workflow

patches/api/0479-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,87 @@
...
++    /**
++     * Get the server info instance
++     * @return the server info instance
++     */
++    static @NotNull ServerInfo serverInfo() {
++        class Holder { // Hide the field from api in a local class
++            private static final Supplier<ServerInfo> instance = Services.service(ServerInfo.class)::orElseThrow;
++        }
++        return Holder.instance.get();
++    }
```
Machine-Maker: Have the static method at the top of the source so people looking at the source for the file can immediately see how to obtain the instance.

patches/server/1049-Add-paper-version-util-class.patch
```diff
@@ -0,0 +1,119 @@
...
++import java.util.Locale;
++import java.util.Set;
++import java.util.regex.Pattern;
++import java.util.stream.Stream;
++
++public class ServerInfoImpl implements ServerInfo {
++    private final String bukkitVersion = Versioning.getBukkitVersion();
++    private final String version = SharedConstants.getCurrentVersion().getId();
++    private final int[] versionArr = this.parseReleaseVersion(this.version);
++    private final Pattern release = Pattern.compile("[0-9](\\.[0-9]+)+");
```
Machine-Maker: Is it a good idea to leave the length of this open ended? the `+` means that `1.2.3.4.5.6.7.8.9` is a valid version here. I think it should be more strict.

Machine-Maker (changes requested):

patches/server/1049-Add-paper-version-util-class.patch
masmc05: It still could have worked and would have been parsed and correctly checked, but made it more strict

masmc05:
Applied all requested changes

kashike:
Hey @masmc05! Thanks for your work on this pull request! ❤️ We've merged (most) of your work into #10729 where we've done some other changes related to versioning, and we'll be continuing this pull request over there. Feel free to comment over there.

Note: there are some minor differences, and there are one or two methods missing in #10729 currently.

PR #10256, opened 2024-02-16
Fix changes on item in EntityPickupItemEvent not reflected #10065

### LINKED ISSUE #10065: Changes on EntityPickupItemEvent not reflected
### Expected behavior

Changes done in the event to the item stack associated with the event should be reflected on the actual picked up item.

### Observed/Actual behavior

The changes are not reflected and are ignored.

### Steps/models to reproduce

Example code: https://code.lewds.de/BQ7OpU

Output showing that even though the item was changed it didnt end up in the inventory
image

The whole process is:
- I drop a diamond to a zombie
- When picking up an item I set a key inside the pdc.
- I check whether the key is actually there. It is.
- I check after two ticks if the item in the hand. It is not.
- The zombie gets killed after 20 ticks
- The zombie drops a diamond without any pdc keys set.
- I pick the diamond up. The key is added again and present in my inventory.
- I drop the now tagged item to a zombie again.
- The zombie picks up my item and the key is now actually there

### Plugin and Datapack List

Only the example plugin above.

### Paper version

1This server is running Paper version git-Paper-318 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 9271ee7)
Previous version: git-Paper-196 (MC: 1.20.1)

### Other

_No response_

rainbowdashlabs:
First time UwU pls be gentle

Fixes #10065 by adding a new overload for the EntityPickupItemEvent event factory and event constructor.
Changes were implemented as proposed in the mentioned issue.

### DIFF
--- a/patches/api/0463-Fix-changes-on-item-in-EntityPickupItemEvent-not-ref.patch
+++ b/patches/api/0463-Fix-changes-on-item-in-EntityPickupItemEvent-not-ref.patch
@@ -5,14 +5,14 @@ Subject: [PATCH] Fix changes on item in EntityPickupItemEvent not reflected
 
 
 diff --git a/src/main/java/org/bukkit/event/entity/EntityPickupItemEvent.java b/src/main/java/org/bukkit/event/entity/EntityPickupItemEvent.java
-index 94ee5a3354722aa5d825da727b7b7071fdc6bacc..a907b0e49d27ac49a834e6be1c416f04f378da16 100644
+index 94ee5a3354722aa5d825da727b7b7071fdc6bacc..64447122b6f917fced3bc10dd8e80ecaee983015 100644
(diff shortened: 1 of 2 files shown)


lynxplay:
After some initial discussion, this change is going to be a lot more involved than it at least seems to be rn.

The *easiest* move here would be to make the yielded ItemStack "immutable" by cloning it in the "getter".

This way, plugins would still need to edit the Item's ItemStack in order to mutate the then used item. Not ideal but I guess doable.
If we want plugins to modify *only* the taken up item, things become a lot harder as we now have to decide what mutation takes precedent. A potentially mutated Item or the ItemStack.

Additionally, the current patch obviously does not properly pass the ItemStack that is consumed. Its count is wrong in a couple of places.
Will certainly need more discussion.

PR #10263, opened 2024-02-19
Adding a PlayerPreTeleportEvent event called before the teleportation…

Will33ELS:
Hello,

I would like to propose the addition of a new event to Paper that would be called before a player's teleportation attempt.

In my current scenario, I have a cosmetic plugin that adds a passenger to the player (a TextDisplay), and all teleportations are blocked because the player has at least one passenger, with no event being called to remove it. This forces the use of NMS (which is discouraged today).

The PlayerPreTeleportEvent event would be invoked before all checks and would allow plugin developers the opportunity to ensure compliance with checks before teleportation occurs.


Leguan16:
I already opened a PR that adds this event: #10173

Will33ELS:
@Leguan16 Well noted, thank you. Do you know when your PR will be processed?

Leguan16:
Nope. Can be tomorrow, can be in 2 weeks. But you can leave feedback if you want :)

PR #10266, opened 2024-02-19
Configurable max block/fluid ticks

Gerrygames:
This allows to limit the maximum fluid and blocks ticks per server tick. I used this option limit the performance impact of large redstone contraptions; Specifically large glow berry farms with a lot of flowing water.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -15,7 +15,7 @@ public net.minecraft.server.dedicated.DedicatedServerProperties reload(Lnet/mine
 public net.minecraft.world.level.NaturalSpawner SPAWNING_CATEGORIES
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 2df1cae62cff433a7f3f55f561f70719bb6a745b..65e9d5918d46b123fb4f8122344a7d3863aec758 100644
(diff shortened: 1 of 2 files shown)


patches/server/0005-Paper-config-files.patch
```diff
@@ -1899,6 +1899,8 @@ index 0000000000000000000000000000000000000000..385ca2c1022e0985550a5cc2bbff953f
 +        public int dryFarmland = 1;
 +        public Table<EntityType<?>, String, Integer> sensor = Util.make(HashBasedTable.create(), table -> table.put(EntityType.VILLAGER, "secondarypoisensor", 40));
 +        public Table<EntityType<?>, String, Integer> behavior = Util.make(HashBasedTable.create(), table -> table.put(EntityType.VILLAGER, "validatenearbypoi", -1));
++        public int maxFluidTicks = 65536;
```
electronicboy: I don't think that putting this in tick-rates is a really good idea, as everything else in here is a "how often per tick", not a "how many per tick" deal

electronicboy (dismissed):

patches/server/0005-Paper-config-files.patch
Gerrygames: Yeah I thought the same thing. But I don't think any other category fits either. Any recommendations?

Owen1212055 (approved):

Machine-Maker (approved):

PR #10267, opened 2024-02-19
Add entity heal API

TotalledZebra:
I haven't contributed to Paper yet so I'm not sure if this is all correct. I feel like this is useful since healing an entity is a pretty common use case and there already is a heal method in NMS.


TotalledZebra:
I've added the missing annotations required by the tests. However after some discussion on Discord I believe the healing either needs to happen without calling an event or there needs to be some sort of flag to control whether or not the EntityRegainHealthEvent is called.

patches/api/0465-Add-entity-heal-API.patch
```diff
@@ -0,0 +1,41 @@
...
+      */
+     void setHealth(double health);
+ 
++    /**
++     * Heal this entity by the given amount. This will call {@link org.bukkit.event.entity.EntityRegainHealthEvent}
++     *
++     * @param amount The amount to heal this entity by
++     */
++    void heal(double amount);
++
```
notTamion: Missing paper start and end comments here.

notTamion (changes requested):
I think its fine to call the event here considering that Damageable#damage also calls the event iirc

kennytv (approved):

lynxplay (approved):

PR #10268, opened 2024-02-19
Fix corrupted plugin.yml breaking plugin loading

TheMolkaPL:
Fix corrupted plugin.yml file in one plugin (or more) not loading any plugins by the server.

The bug puts the server in a dangerous state where vanilla worlds (overworld, nether and the end) are loaded, server is accessible to players, but no plugins are loaded and enabled. To reproduce the bug 1) create `update` directory inside `plugins`, 2) put a valid and a corrupted plugin JAR in `plugins`. By "corrupted" I mean plugin that is eg. missing the `plugin.yml` file, or the file has the wrong YAML syntax. 3) Start the server

The issue has been discovered by one of our clients, we reproduced it, fixed and sent to you to merge it to upstream.

### DIFF
--- a/patches/server/0013-Paper-Plugins.patch
+++ b/patches/server/0013-Paper-Plugins.patch
@@ -5552,10 +5552,10 @@ index 0000000000000000000000000000000000000000..f30fcf60b64150e381c7b813016aa9dd
 +}
 diff --git a/src/main/java/io/papermc/paper/plugin/provider/source/FileProviderSource.java b/src/main/java/io/papermc/paper/plugin/provider/source/FileProviderSource.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..5b58df8df7efca0f67e3a14dd71051dfd7a26079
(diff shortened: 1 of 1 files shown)


Owen1212055:
Future plans, we certainly need to stricten server starting anyways. As plugins get more and more important for the server lifecycle, data loss will be possible in the future.

I think this should be fine for now.

lynxplay:
Wouldn't the better place to catch this be in the `DirectoryProviderSource`'s prepareContext ?

This is treating a single symptom instead of the fact that we currently do not catch exceptions in walkFiles on there.

Owen1212055:
Yes the general catching here could prolly be expanded.

lynxplay:
Implemented by https://github.com/PaperMC/Paper/pull/10279, thank you!

For future PRs, please open them from a user account.
Org owned forks cannot be modified by maintainers so we had to open a new PR instead of being able to reuse yours.

PR #10275, opened 2024-02-22
Prioritise direct entity for Damager in damage event and avoid redundant causing entity

### LINKED ISSUE #10273: Incorrect damager returned for firework damage
### Expected behavior

The `EntityDamageByEntityEvent` called for when a firework explodes and deals damage should return the firework for `#getDamager`.

### Observed/Actual behavior

The owner of the exploding firework is used for `#getDamager` if it exists, and an `EntityDamageEvent` is fired instead if there is no owner.

### Steps/models to reproduce

Fire a firework rocket, record the damager. Tested using this listener:

```java
import org.bukkit.Bukkit;
import org.bukkit.event.entity.EntityDamageEvent;
import org.bukkit.event.entity.EntityDamageByEntityEvent;

    @EventHandler
    public void onDamage(EntityDamageEvent event) {
        if (event instanceof EntityDamageByEntityEvent) return;
        Bukkit.broadcastMessage("Damage event:\n    %s".formatted(event.getEntity()));
    }

    @EventHandler
    public void onDamage(EntityDamageByEntityEvent event) {
        Bukkit.broadcastMessage("Damage by entity event:\n    %s\n    %s".formatted(event.getEntity(),
            event.getDamager()));
    }

}
```

### Plugin and Datapack List

Just the test plugin.

### Paper version

This server is running Paper version git-Paper-430 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: 4939f87)
Previous version: git-Paper-424 (MC: 1.20.4)

### Other

Caused by the upstream damage changes introduced in build 423 (#10242). The firework is properly returned as the damager in build 422. The issue is caused by a missing call to `DamageSource#customCausingEntity` in the following two lines of `FireworkRocketEntity#dealExplosionDamage`:

Line 242:
```java
this.attachedToEntity.hurt(this.damageSources().fireworks(this, this.getOwner()), 5.0F + (float) (nbttaglist.size() * 2));
```

Line 269:
```java
entityliving.hurt(this.damageSources().fireworks(this, this.getOwner()), f1);
```

Doc94:
This fix #10273 making the damager selector use the same logic than Projectile for every case, directEntity is who cause the damage (the damager).

if any can enable the jar for allow users test another bahaviours.. i test the basic based in my upstream PR for make sure this change not break any thing.

---
Download the paperclip jar for this pull request: [paper-10275.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1267366590.zip)

### DIFF
--- a/patches/server/1048-Fix-CB-behaviours-in-DamageSource-for-DamageEvents.patch
+++ b/patches/server/1048-Fix-CB-behaviours-in-DamageSource-for-DamageEvents.patch
@@ -24,7 +24,7 @@ index d677759ac6b6d3cfe5a2af76dc1f0034b216ac2d..620dc20bda7a1959ff7c562e57e8b422
      private static EntityDamageEvent handleEntityDamageEvent(Entity entity, DamageSource source, Map<DamageModifier, Double> modifiers, Map<DamageModifier, Function<? super Double, Double>> modifierFunctions, boolean cancelled) {
          CraftDamageSource bukkitDamageSource = new CraftDamageSource(source);
          Entity damager = source.getCausingEntity();
-+        // Paper start - make damager priorice the direct entity of damagesource
++        // Paper start - make damager prioritise the direct entity of damagesource
 +        if (bukkitDamageSource.isIndirect() && source.getDirectEntity() != null) {
 +            damager = source.getDirectEntity();
 +        }
--- a/patches/server/1048-Fix-CB-behaviours-in-DamageSource-for-DamageEvents.patch
+++ b/patches/server/1048-Fix-CB-behaviours-in-DamageSource-for-DamageEvents.patch
@@ -1,9 +1,21 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Doc <nachito94@msn.com>
 Date: Thu, 22 Feb 2024 10:53:16 -0300
-Subject: [PATCH] Priorice direct entity for damager in DamageEvents
+Subject: [PATCH] Fix CB behaviours in DamageSource for DamageEvents
 
 
+diff --git a/src/main/java/net/minecraft/world/damagesource/DamageSource.java b/src/main/java/net/minecraft/world/damagesource/DamageSource.java
+index 1561b85a45f52a8162f43553f8485bfe084b8f1f..adb9be0c1cbfd524e26e2ff49e37c58fb3ef8751 100644
+--- a/src/main/java/net/minecraft/world/damagesource/DamageSource.java
++++ b/src/main/java/net/minecraft/world/damagesource/DamageSource.java
+@@ -61,6 +61,7 @@ public class DamageSource {
+     }
+ 
+     public DamageSource customCausingEntity(Entity entity) {
++        if (this.customCausingEntity != null || this.directEntity == entity || this.causingEntity == entity) return this; // Paper - avoid this method broke a few CB behaviours
+         DamageSource damageSource = this.cloneInstance();
+         damageSource.customCausingEntity = entity;
+         return damageSource;
 diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
 index d677759ac6b6d3cfe5a2af76dc1f0034b216ac2d..620dc20bda7a1959ff7c562e57e8b422650ce1d7 100644
 --- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
--- a/patches/server/1048-Priorice-direct-entity-for-damager-in-DamageEvents.patch
+++ b/patches/server/1048-Priorice-direct-entity-for-damager-in-DamageEvents.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Thu, 22 Feb 2024 10:53:16 -0300
+Subject: [PATCH] Priorice direct entity for damager in DamageEvents
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index d677759ac6b6d3cfe5a2af76dc1f0034b216ac2d..620dc20bda7a1959ff7c562e57e8b422650ce1d7 100644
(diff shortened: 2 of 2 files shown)


TonytheMacaroni:
This does indeed fix the problem with fireworks.

The issue with the causing entity for some other explosions (i.e. TNT) being wrong is still a thing, though. As far as I can tell, the usage of `DamageSource#customCausingEntity` here:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/world/level/Explosion.patch#42-43
Is unnecessary as a result of this patch, as the direct entity and the set custom causing entity here are always the same, after checking the usages of `Explosion`. Whenever the `damagesource` is non-null, its direct source is always the source `entity` passed to the constructor, and `world.damageSources().explosion(this)` uses the `Explosion`'s source entity for the direct entity. Reverting this line should fix that issue.

Doc94:
> This does indeed fix the problem with fireworks.
> 
> The issue with the causing entity for some other explosions (i.e. TNT) being wrong is still a thing, though. As far as I can tell, the usage of `DamageSource#customCausingEntity` here: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/world/level/Explosion.patch#42-43 Is unnecessary as a result of this patch, as the direct entity and the set custom causing entity here are always the same, after checking the usages of `Explosion`. Whenever the `damagesource` is non-null, its direct entity is always the source entity passed to the constructor, and `world.damageSources().explosion(this)` uses the `Explosion`'s source entity for the direct entity. Reverting this line should fix that issue.

Yeah i notice that in a few test... I remember set that for a reason.. and i forgot that reason now xd
My approach is more edit the method where the custom causing entity is set and make ignore the set if the entity passed is the direct entity, or the customCausingEntity is already set or if the entity passed is null for avoid mess in the CB bahaviour

Doc94:
@TonytheMacaroni PR updated for avoid the customCausingMethod set the value if.
- The customCausingEntity is already set
- The causingEntity is the same than the entity passed
- The directEntity is the same than the entity passed

this can solve the most of issues related to this auxiliar field...

Doc94:
> That would fix the issue with TNT, but still, the call to `customCausingEntity` isn't needed in `Explosion`? That's the only area that that method causes this specific issue.
> 
> And the behavior of `customCausingEntity` is still quite suspect. Take evoker fangs, if it has an owner, the direct entity is the fangs and the owner is the causing entity. But without an owner, `customCausingEntity` makes the fangs the causing entity and doesn't set a direct entity. Similar issue with its usage with ender pearls and lightning - they each have their own owners that should be the causing entity, and the direct entity is also never set in these cases. The inconsistency wouldn't really matter if `net.minecraft.world.damagesource.DamageSource#getCausingEntity` was only used for the damage events, but it's also exposed in the API as `org.bukkit.damage.DamageSource#getCausingEntity`.

The WindCharge make a few things to the damagesource... i point to the upstream PR where talking with Machine suggest a form to handle the static field in the explosion. https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/1180/overview?commentId=12404

TonytheMacaroni:
Just to note, I don't have an account to view the upstream comments you're referring to.

Doc94:
> Just to note, I don't have an account to view the upstream comments you're referring to.

oh...

- The customCausingEntity was make for allow the old static damager set using a static field in CB then need handle that in all places where was placed.. in the case of explosions in the constructor or in another case pass a custom damagesource for that, for this i rewrite a little the method where set the custom causing entity for avoid mistakes
- This customCausingEntity is taken over the causingEntity from vanilla for make DamageSource in bukkit side reflect the custom changes in CB/Paper and avoid things like DamageSource show WindCharge with not causing or direct entity and CB set the causing entity for that

i hope all this can help... i make more tests and the results for projectiles and another explosions looks valid.

JasperLorelai:
Minor spelling issue: `priorice` > `prioritise`/`prioritize`. Present both in [comment](https://github.com/PaperMC/Paper/pull/10275/files#diff-e7daef737934ff92ad2e55ec883170e33ddaa94f42a33bcd7861add5107356c8R27), PR title, and commit message, though only the first might matter for future context.

kennytv (approved):

Owen1212055 (approved):

patches/server/1048-Fix-CB-behaviours-in-DamageSource-for-DamageEvents.patch
```diff
@@ -0,0 +1,34 @@
...
+         return damageSource;
+diff --git a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+index d677759ac6b6d3cfe5a2af76dc1f0034b216ac2d..620dc20bda7a1959ff7c562e57e8b422650ce1d7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -1082,6 +1082,11 @@ public class CraftEventFactory {
+     private static EntityDamageEvent handleEntityDamageEvent(Entity entity, DamageSource source, Map<DamageModifier, Double> modifiers, Map<DamageModifier, Function<? super Double, Double>> modifierFunctions, boolean cancelled) {
+         CraftDamageSource bukkitDamageSource = new CraftDamageSource(source);
+         Entity damager = source.getCausingEntity();
++        // Paper start - make damager prioritise the direct entity of damagesource
```
u9g, line 27:
```suggestion
+        // Paper start - make damager prioritize the direct entity of damagesource
```

u9g: Same mistake in the pr's title

u9g: https://grammarist.com/spelling/prioritize-vs-prioritise/

JasperLorelai: British spelling is a mistake?

DragonsAscent: Very much so (joking). 

Realistically you don't see the game naming it armour > armor so I think it makes sense

Owen1212055:
Superseded by https://github.com/PaperMC/Paper/pull/10307

PR #10284, opened 2024-02-24
add biome change per player

### LINKED ISSUE #9889: Change biomes per-player in chunk/biome packets

There isn't a way to set biomes to be per-player

In the `PlayerChunkLoadEvent` or a new superevent of that, there should be a way to change the biome data sent to the client in order for each player to be able to get a different set of biomes.

I think this has to be a new superevent of PlayerChunkLoadEvent because biomes can change independent of chunks being sent to the client with the new fillbiome command. There is also a specific packet for changing a chunk biomes.

N/A

### Other

I'm not sure of the correct API structure for representing chunk biomes. Generally, I think it would be beneficial if plugins could create chunk biome data arbitrarily to they aren't creating new PalettedContainers for every player. Having some way plugins could create biome data, cache it, and then use it for specific players would be ideal.

ChunkSnapshot is a good place to start. It wraps around 2 arrays of PalettedContainers for blockstates and biomes. Having a supertype of that which just wraps biomes seems like a good idea.

Something to note is that the array of PalettedContainer for biomes has a different length depending on the height of the world. So care would have to be taken to ensure that the sizes matched.

yannicklamprecht:
resolves #9889 

# TODO:

- [ ] Implement and call event
- [ ] finish implementing BiomesSnapshot
- [ ] add a handle to Chunk
- [ ] test api with plugin
- [ ] test compatibility with plugins using PlayerChunkLoadEvent
- [ ] reduce diff
- [ ] add paper comments

### DIFF
--- a/patches/api/0495-add-biome-change-per-player.patch
+++ b/patches/api/0495-add-biome-change-per-player.patch
@@ -1,0 +1,275 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Yannick Lamprecht <yannicklamprecht@live.de>
+Date: Sat, 24 Feb 2024 15:29:32 +0100
+Subject: [PATCH] add biome change per player
+
+
+diff --git a/src/main/java/io/papermc/paper/event/packet/PlayerBiomesLoadEvent.java b/src/main/java/io/papermc/paper/event/packet/PlayerBiomesLoadEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..cf8e0d297144e12c64cc7a9036ca896c1f4ac7af
+--- /dev/null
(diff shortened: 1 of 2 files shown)


Owen1212055:
We have a more generalized idea for something like this via our registry api, prolly better to wait.

Would allow us to make any registry modifiable.

Machine-Maker:
I'm thinking it would still be useful to get the biomes being sent by vanilla. Right now, the event is only dealing with any overrides. Implementing this might also be way nicer if this event was implemented on the server side. So the creation of BiomeSnapshots could be done in a lazy fashion to avoid creating snapshots for everything just to call the event.

yannicklamprecht:
post hard fork pr: https://github.com/PaperMC/Paper/pull/11799

PR #10285, opened 2024-02-25
check if itemstack is stackable first

granny:
Avoids an issue where unstackable items are checked against themselves when an inventory is full of items.

For more context to paper devs who are in the purpur discord: https://discord.com/channels/685683385313919172/697248751575761057/1211147576213119006

This can and should be merged into a pre-existing patch - I'm not sure which one so I've left it as it's own for now.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -63,4 +63,5 @@ Ollie <69084614+olijeffers0n@users.noreply.github.com>
 Oliwier Miodun <naczs@blueflow.pl>
 aerulion <aerulion@gmail.com>
 Lukas Planz <lukas.planz@web.de>
+granny <contact@granny.dev>
 ```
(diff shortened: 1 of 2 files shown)


electronicboy (approved):
I believe that there is already a semi-related patch which does something similar elsewhere that this could be merged in-to as said

DxrkCat:
> I believe that there is already a semi-related patch which does something similar elsewhere that this could be merged in-to as said

This patch only partially fixes the problem, because the server will be overloaded with sending notifications that will show up as broadcastChanges() in spark.lucko.me.

electronicboy:
sending/dealing with a lot of nested data is always going to suck, theres not too much that can be done about that unless you have full control over stuff

Owen1212055 (approved):

Machine-Maker (approved):

PR #10288, opened 2024-02-28
Add BlockSearchPlayersEvent

### LINKED ISSUE #10137: Add `BlockApplyEffectsEvent`

Yes, I am trying to apply potion effects to players outside of water when there is a nearby conduit in water, but currently there is no good way to test for an active conduit trying to give effects.

A `BlockApplyEffects` event, for beacons and conduits, containing a `getBlock()` method and a `getPower()` method for the size of the conduit prismarine structure or beacon pyramid.

This event would be called when a beacon or conduit is about to iterate over players to apply effects, firing once each time regardless of if no players are going to be given the effect.

Looping through nearby blocks for an active conduit or beacon, however this would not work as it would have to often loop through large amounts of blocks, likely causing the server to crash or slow down by a lot as it wouldn't be fast enough to handle searching through so many nearby blocks

### Other

Another name for the event might be better, as `BlockApplyEffects` may suggest that effects are actually being applied rather than the game being about to try applying them, I suggested it with that name since the method for trying to apply effects in the game source code is `applyEffects()`

notTamion:
resolves #10137 
no idea if there are any other blocks that grant effects, conduits and beacons are the only ones i know of


molor:
`BlockSearchPlayerEvent` sounds better IMO

notTamion:
I agree, however i will wait since i am sure maintainers are gonna have something to say about the naming too

kennytv:
I'd say BlockSearchPlayersEvent

PR #10290, opened 2024-02-29
Deprecate ItemStack#setType

Owen1212055:
This leads to discussion in regards to `ItemStack#setType`, which currently is not supported by vanilla, but is reintroduced by spigot.

This method possibly allows for unsafe behavior, as illegal item types can be passed in places that may occur after item validation.

This is most evident in event calls, where it is possible to induce server crashes in some cases due to developers putting illegal item types in ItemStacks in the wrong places.

### DIFF
--- a/patches/api/0124-PlayerElytraBoostEvent.patch
+++ b/patches/api/0124-PlayerElytraBoostEvent.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] PlayerElytraBoostEvent
 
 diff --git a/src/main/java/com/destroystokyo/paper/event/player/PlayerElytraBoostEvent.java b/src/main/java/com/destroystokyo/paper/event/player/PlayerElytraBoostEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..42375d44ed0db457b96165414e9b45a557bac1ab
+index 0000000000000000000000000000000000000000..9ebac41c22b8866df616020d409e5e1a49cddca5
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/event/player/PlayerElytraBoostEvent.java
-@@ -0,0 +1,90 @@
+@@ -0,0 +1,104 @@
 +package com.destroystokyo.paper.event.player;
 +
 +import org.bukkit.entity.Firework;
@@ -17,6 +17,7 @@ index 0000000000000000000000000000000000000000..42375d44ed0db457b96165414e9b45a5
 +import org.bukkit.event.Cancellable;
 +import org.bukkit.event.HandlerList;
 +import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.inventory.EquipmentSlot;
 +import org.bukkit.inventory.ItemStack;
 +import org.jetbrains.annotations.ApiStatus;
 +import org.jetbrains.annotations.NotNull;
@@ -31,14 +32,17 @@ index 0000000000000000000000000000000000000000..42375d44ed0db457b96165414e9b45a5
 +    @NotNull private final ItemStack itemStack;
 +    @NotNull private final Firework firework;
 +    private boolean consume = true;
++    @NotNull
++    private final EquipmentSlot hand;
 +
 +    private boolean cancelled;
 +
 +    @ApiStatus.Internal
-+    public PlayerElytraBoostEvent(@NotNull Player player, @NotNull ItemStack itemStack, @NotNull Firework firework) {
++    public PlayerElytraBoostEvent(@NotNull Player player, @NotNull ItemStack itemStack, @NotNull Firework firework, @NotNull EquipmentSlot hand) {
 +        super(player);
 +        this.itemStack = itemStack;
 +        this.firework = firework;
++        this.hand = hand;
 +    }
 +
 +    /**
@@ -77,6 +81,16 @@ index 0000000000000000000000000000000000000000..42375d44ed0db457b96165414e9b45a5
 +     */
 +    public void setShouldConsume(boolean consume) {
 +        this.consume = consume;
++    }
++
++    /**
++     * Gets the hand holding the firework used for boosting this player.
++     *
++     * @return interaction hand
++     */
++    @NotNull
++    public EquipmentSlot getHand() {
++        return this.hand;
 +    }
 +
 +    @Override
--- a/patches/api/0469-Deprecate-ItemStack-setType.patch
+++ b/patches/api/0469-Deprecate-ItemStack-setType.patch
@@ -1,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Owen1212055 <23108066+Owen1212055@users.noreply.github.com>
+Date: Thu, 29 Feb 2024 17:54:26 -0500
+Subject: [PATCH] Deprecate ItemStack#setType
+
+
+diff --git a/src/main/java/org/bukkit/inventory/ItemStack.java b/src/main/java/org/bukkit/inventory/ItemStack.java
+index 245a730a54c4b241a9a67eccceefafd2763bd238..7414b4fa690d393a8e9557cc1fd1ce12fa426940 100644
+--- a/src/main/java/org/bukkit/inventory/ItemStack.java
++++ b/src/main/java/org/bukkit/inventory/ItemStack.java
+@@ -134,8 +134,18 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
+      * {@link Material#isItem()} returns false.</b>
+      *
+      * @param type New type to set the items in this stack to
++     * @deprecated <b>Setting the material type of ItemStacks is no longer supported.</b>
++     * <p>
++     * This method is deprecated due to potential illegal behavior that may occur
++     * during the context of which this ItemStack is being used, allowing for certain item validation to be bypassed.
++     * It is recommended to instead create a new ItemStack object with the desired
++     * Material type, and if possible, set it in the appropriate context.
++     *
++     * Using this method in ItemStacks passed in events will result in undefined behavior.
++     * @see ItemStack#withType(Material)
+      */
+     @Utility
++    @Deprecated // Paper
+     public void setType(@NotNull Material type) {
+         Preconditions.checkArgument(type != null, "Material cannot be null");
+         this.type = type;
+@@ -148,6 +158,24 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
+             this.data = null;
+         }
+     }
++    // Paper start
++    /**
++     * Creates a new ItemStack with the specified Material type, where the item count and item meta is preserved.
++     *
++     * @param type The Material type of the new ItemStack.
++     * @return A new ItemStack instance with the specified Material type.
++     */
++    @NotNull
++    @org.jetbrains.annotations.Contract(value = "_ -> new", pure = true)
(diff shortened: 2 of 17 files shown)


imDaniX:
Any chance for something like `ItemStack#withType` which will return a new ItemStack with desired type and applicable ItemMeta?

Owen1212055:
withType has been added.

patches/api/0464-Deprecate-ItemStack-setType.patch
```diff
@@ -0,0 +1,54 @@
...
++     *
++     * @param type The Material type of the new ItemStack.
++     * @return A new ItemStack instance with the specified Material type.
++     */
++    @NotNull
++    @org.jetbrains.annotations.Contract(value = "_, -> new", pure = true)
++    public ItemStack withType(@NotNull Material type) {
++        ItemStack itemStack = new ItemStack(type, this.amount);
++        if (this.hasItemMeta()) {
++            itemStack.setItemMeta(this.getItemMeta());
```
yannicklamprecht: Does setItemMeta check that only type compatible meta is set?

electronicboy: it checks that the ItemMeta impl class is valid for that type, if not, it will fail

yannicklamprecht: Should that be documented on that method or should we expect that knowledge from the devs?

https://jd.papermc.io/paper/1.20/org/bukkit/inventory/ItemFactory.html#isApplicable(org.bukkit.inventory.meta.ItemMeta,org.bukkit.inventory.ItemStack)

Owen1212055: > Should that be documented on that method or should we expect that knowledge from the devs?
> 
> https://jd.papermc.io/paper/1.20/org/bukkit/inventory/ItemFactory.html#isApplicable(org.bukkit.inventory.meta.ItemMeta,org.bukkit.inventory.ItemStack)
> 

I was considering it, but idk, it's kinda just behavior that is expected when using set type anyways

electronicboy: I would imagine it should be noted as a trap, as the expectation here is likely that "oh, it should at least copy the lore" type of deal

patches/server/1056-fixup-Add-BlockPreDispenseEvent.patch
```diff
@@ -0,0 +1,57 @@
...
++++ b/src/main/java/net/minecraft/world/level/block/DispenserBlock.java
+@@ -104,7 +104,11 @@ public class DispenserBlock extends BaseEntityBlock {
+                 DispenseItemBehavior idispensebehavior = this.getDispenseMethod(itemstack);
+ 
+                 if (idispensebehavior != DispenseItemBehavior.NOOP) {
+-                    if (!org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockPreDispenseEvent(world, pos, itemstack, i)) return; // Paper - Add BlockPreDispenseEvent
++                    // Paper start - BlockPreDispenseEvent
++                    io.papermc.paper.event.block.BlockPreDispenseEvent preDispenseEvent = org.bukkit.craftbukkit.event.CraftEventFactory.handleBlockPreDispenseEvent(world, pos, itemstack, i);
++                    if (!preDispenseEvent.callEvent()) return;
++                    itemstack = org.bukkit.craftbukkit.inventory.CraftItemStack.asNMSCopy(preDispenseEvent.getItemStack());
```
Machine-Maker: I think a utility method to prevent doing a copy if, in fact, the itemstack instance wasn't changed via a setItemStack call in the event. Just saves us a copy for what I think will be most cases.

patches/server/1058-Item-Mutation-Fixes.patch
```diff
@@ -0,0 +1,30 @@
...
+index 343f44db579839eb61376f876b5eff2e615dc2e5..7b9015bdc3f888056a75f16046b3a19a00fc1465 100644
+--- a/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java
++++ b/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java
+@@ -251,11 +251,17 @@ public class EnchantmentMenu extends AbstractContainerMenu {
+                             return;
+                         }
+                         // CraftBukkit end
+-                        boolean flag = itemstack.is(Items.BOOK);
++                        // Paper start
++                        itemstack2 = org.bukkit.craftbukkit.inventory.CraftItemStack.unwrap(event.getItem());
```
kennytv: itemstack2 is a good name

kennytv (approved):
(make sure to handle the fixups)

Machine-Maker (changes requested):
* InventoryBlockStartEvent should have a setItemStack along with a fix for `CampfireStartEvent`'s call to use the split(1) stack instead of the stack from hand.

patches/api/0469-Deprecate-ItemStack-setType.patch
```diff
@@ -0,0 +1,54 @@
...
+     }
++    // Paper start
++    /**
++     * Creates a new ItemStack with the specified Material type, where the item count and item meta is preserved.
++     *
++     * @param type The Material type of the new ItemStack.
++     * @return A new ItemStack instance with the specified Material type.
++     */
++    @NotNull
++    @org.jetbrains.annotations.Contract(value = "_, -> new", pure = true)
```
kashike:
```suggestion
+    @org.jetbrains.annotations.Contract(value = "_ -> new", pure = true)
```

kashike (dismissed):

patches/api/0124-PlayerElytraBoostEvent.patch
```diff
@@ -79,6 +83,16 @@ index 0000000000000000000000000000000000000000..42375d44ed0db457b96165414e9b45a5
 +        this.consume = consume;
 +    }
 +
++    /**
++     * Gets the hand holding the firework used for boosting this player.
++     *
++     * @return hand
```
Lulu13022002: Missing description of this parameter

patches/api/0469-Item-Mutation-Fixes.patch
```diff
@@ -0,0 +1,71 @@
...
+     private final Map<Enchantment, Integer> enchants;
+@@ -72,6 +72,17 @@ public class EnchantItemEvent extends InventoryEvent implements Cancellable {
+         return item;
+     }
+ 
++    // Paper start
++    /**
++     * Sets the item to be enchanted
++     *
++     * @param item item
```
Lulu13022002: Missing description of this parameter and this should be moved after the getItem() method

patches/server/1055-Item-Mutation-Fixes.patch
```diff
@@ -0,0 +1,72 @@
...
+diff --git a/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java b/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java
+index 343f44db579839eb61376f876b5eff2e615dc2e5..b4a63e8d04b60d5a0c6c2d407d777c4142cee2a6 100644
+--- a/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java
++++ b/src/main/java/net/minecraft/world/inventory/EnchantmentMenu.java
+@@ -229,7 +229,7 @@ public class EnchantmentMenu extends AbstractContainerMenu {
+                 return false;
+             } else if (this.costs[id] > 0 && !itemstack.isEmpty() && (player.experienceLevel >= j && player.experienceLevel >= this.costs[id] || player.getAbilities().instabuild)) {
+                 this.access.execute((world, blockposition) -> {
+-                    ItemStack itemstack2 = itemstack;
++                    ItemStack itemstack2 = itemstack; // Paper -- diff on change
```
Lulu13022002: Duplicate `-` before the comment

patches/server/1055-Item-Mutation-Fixes.patch
```diff
@@ -0,0 +1,72 @@
...
+                     // CraftBukkit start
+@@ -251,11 +251,19 @@ public class EnchantmentMenu extends AbstractContainerMenu {
+                             return;
+                         }
+                         // CraftBukkit end
+-                        boolean flag = itemstack.is(Items.BOOK);
++                        // Paper start
++                        itemstack2 = org.bukkit.craftbukkit.inventory.CraftItemStack.getOrCloneOnMutation(item, event.getItem());
++                        if (itemstack2 != itemstack) {
++                         this.enchantSlots.setItem(0, itemstack2);
```
Lulu13022002: Bad indentation

patches/server/1055-Item-Mutation-Fixes.patch
```diff
@@ -0,0 +1,72 @@
...
++                        }
++                        boolean flag = itemstack2.is(Items.BOOK);
++                        // Paper end
+ 
+                         if (flag) {
++                            // Paper start - reoder
```
Lulu13022002: typo and this comment can be on a single line after the moved var

Lulu13022002 (commented):
I got this error when i try to test the EnchantItemEvent: https://pastes.dev/b5KQ0vJydM seems paper doesn't check if the itemmeta is null but it shouldn't be null in the first place for the item i tested (a sword) seems the amount is zero in that case. This is done using the new withType method on the ItemStack. You should probably override this method in the Craft impl too to avoid converting a CraftStack into an api one with this method.
The code is just: e.setItem(e.getItem().withType(Material.BAKED_POTATO))

Owen1212055:
Resolved, thanks lulu.

patches/server/1055-Deprecate-ItemStack-setType.patch
```diff
@@ -0,0 +1,26 @@
...
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
+@@ -711,4 +711,15 @@ public final class CraftItemStack extends ItemStack {
+     static boolean hasItemMeta(net.minecraft.world.item.ItemStack item) {
+         return !(item == null || item.getTag() == null || item.getTag().isEmpty());
+     }
++    // Paper start - with type
++    @Override
++    public ItemStack withType(final Material type) {
++        ItemStack itemStack = asNewCraftStack(CraftItemType.bukkitToMinecraft(type), this.getAmount());
```
Lulu13022002: This goes into a double conversion (Material -> Item (internal) -> Material) for no reason? Also you can set the item meta directly using the regular constructor.

Machine-Maker (approved):

patches/server/1055-Deprecate-ItemStack-setType.patch
```diff
@@ -0,0 +1,36 @@
...
+     }
++    // Paper start - with type
++    @Override
++    public ItemStack withType(final Material type) {
++        if (type == Material.AIR) return CraftItemStack.asCraftMirror(null);
++
++        final net.minecraft.world.item.ItemStack copy = new net.minecraft.world.item.ItemStack(
++            CraftItemType.bukkitToMinecraft(type), this.getAmount()
++        );
++        if (this.handle != null) copy.setTag(this.handle.getTag());
```
kashike: Should we be copying the tag as well?

lynxplay: Yea mb, talked about it in the call yesterday 💀

lynxplay (approved):

PR #10302, opened 2024-03-06
Expose power on fireballs

### LINKED ISSUE #10186: Fireball setVelocity incorrectly sets power
See https://github.com/PaperMC/Paper/pull/9815#discussion_r1461306045

The patch should probably be reverted and maybe a new setPower method added

TonytheMacaroni:
Fixes #10186.

~~Patch previously modified the power of the projectile, which actually controls how fast it accelerates. It now uses `AbstractHurtingProjectile#setDirection`, which also sets the power, but uses the default magnitude of `0.1`.~~

~~Considered also directly exposing the power, but was unsure exactly on the best approach of implementing it. `AbstractHurtingProjectile#setDirection` assumes a magnitude of `0.1`, which would need changing if power could be changed. Could either calculate the magnitude from the current power on setting direction, which would need some consideration for when the projectile is created, or including a field for it that's set when the power is modified.~~

Now entirely reverts the original patch, and adds a javadoc comment to `Fireball#setVelocity` to note that fireballs accelerate in the direction of their power. Also added `Fireball#setPower` and `Fireball#getPower` to control the power of the projectile. Currently `Fireball#getDirection` and `Fireball#getPower` do the same thing - should I remove `Fireball#getPower` or deprecate `Fireball#getDirection` for its misleading name?


TonytheMacaroni:
On second thought, is setting the direction when you set the velocity even a sensible thing to do? Perhaps just a javadoc note on `Fireball#setVelocity` is necessary?

patches/api/0466-Expose-power-for-fireball-projectiles.patch
```diff
@@ -0,0 +1,41 @@
...
++    public void setPower(@NotNull Vector power);
++
++    /**
++     * Gets the power of a fireball. The power determines the direction and magnitude of its acceleration.
++     *
++     * @return the power
++     */
++    @NotNull
++    public Vector getPower();
++
```
maxcom1: Missing `// Paper start` and `// Paper end` comments - https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md

patches/server/0948-Expose-power-on-fireball-projectiles.patch
```diff
@@ -1,29 +1,28 @@
...
++        this.getHandle().yPower = power.getY();
++        this.getHandle().zPower = power.getZ();
++    }
++
++    @Override
++    public Vector getPower() {
++        return new Vector(this.getHandle().xPower, this.getHandle().yPower, this.getHandle().zPower);
 +    }
-+    // Paper end - set direction without normalizing
 +
```
maxcom1: Missing new `// Paper start` and `// Paper end` comments - https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md

maxcom1 (changes requested):

Machine-Maker (approved):

PR #10306, opened 2024-03-08
Fire EntityChangeBlockEvent on beehive nectar deposit

maxcom1:
Fire `EntityChangeBlockEvent` after a Bee is released from a Beehive, when bee deposits nectar to the beehive.

**Can be used to e.g.:**
* change amount of deposited nectar, modify final hive's honey level after deposition (deposited nectar amount is normally 1 or 2, we can set the value to e.g. 5)
* use in various mechanisms
* just cancel deposit of nectar to the hive


patches/api/0466-Add-BeehiveNectarDepositEvent.patch
```diff
@@ -0,0 +1,127 @@
...
++import org.bukkit.event.block.BlockEvent;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Fired when Bee's nectar is deposited into the Beehive.
++ * <p>
++ * The event is fired right after the bee is released, only if there is a space in the hive for new nectar
++ */
++public class BeehiveNectarDepositEvent extends BlockEvent implements Cancellable {
```
lynxplay: Generally, this entire event is not needed.
EntityChangeBlockEvent already exposes
 a) the changing block in its old state and hence a way to get the old honey level
 b) the block data of the block after the change, which can be used to get and set the new honey level
 c) point B also allows to compute the deposited nectar
 d) the bee is available as the entity of the entity change block event.
 
The API addition here is hence not really needed and we can instead just call EntityChangeBlockEvent.

The API patch can hence be completely dropped and the server patch changed to call EntityChangeBlockEvent.
If you are comfortable enough, you can also merge the server patch changes into the
`Fire EntityChangeBlockEvent in more places` patch.
Make sure to add yourself as a co-author there :+1:

maxcom1:
So I started testing

I changed Paper-server code of BeehiveBlockEntity as follows:
```java
// Paper start - Fire EntityChangeBlockEvent in more places
BlockState newBlockState = iblockdata.setValue(BeehiveBlock.HONEY_LEVEL, i + j);

if (org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(entitybee, blockposition, newBlockState)) {
    world.setBlockAndUpdate(blockposition, newBlockState);
}
// Paper end - Fire EntityChangeBlockEvent in more places
```
and modified my test plugin:
```java
@EventHandler
public void onNectarDeposit(EntityChangeBlockEvent e) {
    if (!(e.getBlockData() instanceof Beehive newBeeHive)) return;
    if (!(e.getEntity() instanceof Bee bee)) return;

    Beehive oldBeeHive = (Beehive) e.getBlock().getBlockData();

    int oldHoneyLevel = oldBeeHive.getHoneyLevel();
    int newHoneyLevel = newBeeHive.getHoneyLevel();
    int difference = newHoneyLevel-oldHoneyLevel;

    Location beehiveLoc = e.getBlock().getLocation();

    String debug = "[NectarDepositListener] <yellow>Bee " + e.getEntity().getUniqueId() +
            " <white>dropped <gold>" + difference +
            " nectars <white>into "+
            " <yellow>Beehive at x: " + beehiveLoc.getBlockX() + ", y: " + beehiveLoc.getBlockY() + ", z: " + beehiveLoc.getBlockZ() + "<white>." +
            " The <aqua>old honey level <white>is <aqua>" + oldHoneyLevel + "<white>, <light_purple>new honey level <white>is <light_purple>" + newHoneyLevel;

    Bukkit.getServer().sendMessage(MiniMessage.miniMessage().deserialize(debug));

    ((Beehive) e.getBlockData()).setHoneyLevel(5);
}
```

The output from plugin (debug message) is sent correctly - it shows values *before* modifying:
image

But the honey level of the hive does not change, it increases as if its value is not changed. Are you sure that the new block state is mutable?

lynxplay (changes requested):
After a bit more consideration I don't think it is worth to introduce a new event.
For now, we will have to live without the block data being mutable.
To set the new honey level, people can still just cancel the event and update the block themselves.

I think it makes more sense for us to look into later expanding EntityChangeBlockEvent to allow for a `setBlockData` that is potentially respected instead of littering our API with event types that basically add 0 new info beyond well, a `setBlockData` on EntityChangeBlockEvent.

Can you try to for now, just implement a normal EntityChangeBlockEvent for this?

maxcom1:
Sure, I understand - I was planning to change this to fire EntityChangeBlockEvent, just will look into it when I have some time :>

maxcom1:
Done

lynxplay (approved):

Machine-Maker (approved):

PR #10307, opened 2024-03-09
Fix DamageSource API

Machine-Maker:
Uses the correct entity in the EntityDamageByEntity event Returns the correct entity for API's DamageSource#getCausingEntity

Superceeds https://github.com/PaperMC/Paper/pull/10275

---
Download the paperclip jar for this pull request: [paper-10307.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1332142312.zip)

### DIFF
--- a/patches/api/0466-Fix-DamageSource-API.patch
+++ b/patches/api/0466-Fix-DamageSource-API.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sat, 16 Mar 2024 11:21:14 -0700
+Subject: [PATCH] Fix DamageSource API
+
+
+diff --git a/src/main/java/org/bukkit/event/entity/EntityDamageByEntityEvent.java b/src/main/java/org/bukkit/event/entity/EntityDamageByEntityEvent.java
+index 6b24d1281cb8f0253430c9c1a1323e2670bb9c93..8ea4be529400b34df3d31b0f17c2d145345523d9 100644
+--- a/src/main/java/org/bukkit/event/entity/EntityDamageByEntityEvent.java
++++ b/src/main/java/org/bukkit/event/entity/EntityDamageByEntityEvent.java
+@@ -60,6 +60,20 @@ public class EntityDamageByEntityEvent extends EntityDamageEvent {
+     }
+     // Paper end
+ 
++    // Paper start
++    /**
++     * {@inheritDoc}
++     * <p>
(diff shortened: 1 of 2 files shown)


Doc94:
In a fast check i wanna ask.. this make the thing i mention in upstream about now DamageEvent#getDamager is not the same than DamageSource#getEntity? if that is the case this make a inconsistence beetween plugins running in Paper and Spigot not? (well that its not relevant...)

patches/server/1054-Fix-DamageSource-API.patch
```diff
@@ -0,0 +1,166 @@
...
+         this.withSweep = true;
+@@ -56,13 +57,15 @@ public class DamageSource {
+         return this.poison;
+     }
+ 
+-    public Entity getCausingEntity() {
+-        return (this.customCausingEntity != null) ? this.customCausingEntity : this.causingEntity;
++    // Paper start - fix DamageSource API
++    public @Nullable Entity getCustomEventDamager() {
++        return (this.customEventDamager != null) ? this.customEventDamager : this.directEntity;
```
Doc94: this now ignore the causingEntity?

Machine-Maker: Yeah, because the Event#getDamager should always return the direct entity if a custom one hasn't been set.

lynxplay: That is the whole point.
You would not want the EntityDamageByEntityEvent to have the player lighting the tnt as the event damager.
You'd want the tnt itself as damager.

Doc94: yeah the thing dont remember correctly if the directEntity can be null.

Doc94:
For explosions looks is working fine.
For tests cases where now is used customCausingEntity the DamageSource has null direct/causing entity causing like i said this sounds inconsistent.. and make a user better handle in the event for damager rather than try to use DamageSource for that info

lynxplay (approved):

patches/api/0466-Fix-DamageSource-API.patch
```diff
@@ -0,0 +1,31 @@
...
+@@ -60,6 +60,20 @@ public class EntityDamageByEntityEvent extends EntityDamageEvent {
+     }
+     // Paper end
+ 
++    // Paper start
++    /**
++     * {@inheritDoc}
++     * <p>
++     * The {@link DamageSource#getDirectEntity()} may be different from the {@link #getDamager()}
++     * if the Minecraft damage source did not originally include an damager entity, but one was included
```
NonSwag, line 20: small typo here: an damager -> should be a damager

Issue #10309, opened 2024-03-10
Make whitelist enabled by default


MrGeoTech:
### Is your feature request related to a problem?

Recently, more and more people have become aware of server scanning and using it to find servers. As a result, it has become shockingly easy to find server that are open and ripe for griefing. One of the easiest ways to prevent this is by enabling a whitelist. For a lot of new server owners, a whitelist only really comes to light after they have already been griefed. I feel that if the whitelist was enabled by default and maybe even a warning message when the server whitelist is turned off (and when the server is started and the whitelist is disabled), new server owners will recognize the importance of this issue. Up until now, simply not sharing your ip has been enough but we are at a point in time where more is needed and I think something as simple will drastically help with the issue at hand.

### Describe the solution you'd like.

Enable whitelist by default, warn the command sender about the dangers of disabling the whitelist when disabled, and a warning when starting the server with the whitelist disabled.

### Describe alternatives you've considered.

There could be and argument that just documenting enabling a whitelist in the setup docs would be enough but I feel that is not enough, even if it is documented as something required to do when first setting up a server.

### Other

I realize that this should be a Mojang issue but as they are not doing anything about it themselves, I feel as the most popular Minecraft server software, Paper has the responsibility to keep its users safe from dangers they might not even be aware of. To me, it seems like a simple thing to do that will have a large impact on the security of Paper's users while also not being too much of a hassle for server owners who don't want the whitelist enabled. I'm also hoping that this will set a precedent that other server software and potentially even Mojang will recognize.

TetraTheta:
I think this kind of request should go to Mojang directly, rather than PaperMC.

Mojang's decision is setting `white-list`(enable whitelist) and `enforce-whitelist`(kick players not on whitelist) to `false`. I don't think PaperMC should interfere with it.

If Paper enforces whitelist by default, that said 'new server owners' won't be able to figure out how to allow people to get in their server.

If a server is griefed, server owner is also responsible for the incident. Server owner must know how their server is configured, by carefully reading the documentation. PaperMC already offers [detailed documentation](https://docs.papermc.io/paper/reference/configuration). That's enough for PaperMC. Anything further should be in server owner's responsibility.

Leguan16:
In the other hand, this would probably cause people flood the help chats to ask why they cannot join their fresh server.

lynxplay:
I want to close this feature request as not planned, however I'll leave this open for other people to chime in on.

[...]

PR #10312, opened 2024-03-10
Add onboarding message for initial server start

Leguan16:
This is a PR as a response to #10309.

What i have done here is by no means final and just an attempt to find a way to check if it is the "initial" start of a server. 
~~I decided to go for the `server.properties` file because this is the first file that came to my mind that you wouldnt really delete.~~ I am now using the global configuration to determine if it's the initial start. 

Basically, I am checking if the file exists and set a boolean if it doens't. After the server has been started, I check that boolean and print a message if it's the initial start. 
Obviously, you could just delete the file and get that message again so it wouldn't be the "initial" start anymore. So, if you have any other idea how to handle this, please leave feedback. 

The GUI is also modified to contain a message and links to the onboarding page. 
The message is also not final. 

This would also allow for a `InitialServerStartEvent` later on once this is merged.

### DIFF
--- a/patches/server/1054-Add-onboarding-message-for-initial-server-start.patch
+++ b/patches/server/1054-Add-onboarding-message-for-initial-server-start.patch
@@ -1,0 +1,103 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: leguan <longboard.noah@gmail.com>
+Date: Sun, 10 Mar 2024 20:10:41 +0100
+Subject: [PATCH] Add onboarding message for initial server start
+
+
+diff --git a/src/main/java/io/papermc/paper/configuration/Configurations.java b/src/main/java/io/papermc/paper/configuration/Configurations.java
+index c01b4393439838976965823298f12e4762e72eff..218bf89fd7583d6db9f64754c4db8fcce5415bdb 100644
+--- a/src/main/java/io/papermc/paper/configuration/Configurations.java
++++ b/src/main/java/io/papermc/paper/configuration/Configurations.java
+@@ -126,6 +126,7 @@ public abstract class Configurations<G, W> {
+         if (Files.notExists(configFile)) {
+             node = CommentedConfigurationNode.root(loader.defaultOptions());
+             node.node(Configuration.VERSION_FIELD).raw(this.globalConfigVersion());
++            GlobalConfiguration.isFirstStart = true;
+         } else {
+             node = loader.load();
+             this.verifyGlobalConfigVersion(node);
+diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+index b53b6cc4463675096b061b3b65f14a4695c742e2..f2e30d8743a97c0541808dc5c03300c0d881b843 100644
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -25,6 +25,7 @@ public class GlobalConfiguration extends ConfigurationPart {
+     private static final Logger LOGGER = LogUtils.getLogger();
+     static final int CURRENT_VERSION = 29; // (when you change the version, change the comment, so it conflicts on rebases): <insert changes here>
+     private static GlobalConfiguration instance;
++    public static boolean isFirstStart = false;
+     public static GlobalConfiguration get() {
+         return instance;
+     }
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index d06185566b447c432d4dc2e3ba04d121bcdbc71b..469eaf3a65793edd8f40d40fb476c05913032047 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -1155,6 +1155,16 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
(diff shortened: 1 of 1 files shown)


masmc05:
I think most of the time people that are starting the server for the first time see the eula message, I think it may be good enough to just add the additional info with the link to documentation after/before the eula message

Leguan16:
> I think most of the time people that are starting the server for the first time see the eula message, I think it may be good enough to just add the additional info with the link to documentation after/before the eula message

I think that would be too early.

MiniDigger:
I like the idea of improving the onboarding process

Malfrador:
I'm not 100% sure about storing Paper things in the server.properties file. This would be the first time for Bukkit, Spigot and Paper to add a new value there.
While it's not really gonna cause issues, this doesn't seem like the proper thing to do. IMO just using the paper global configuration would be fine - yes you get the message again if you delete your config, but it's just a log message, not a big deal, and idk if the event really is needed. There's a good chance users have no plugins installed during the first startup anyways. 

I do agree that the EULA is too early though. Some panels also realise EULA acceptance with a checkbox or similar, so not everyone will see it.

Leguan16:
> I'm not 100% sure about storing Paper things in the server.properties file.

I'm not sure if I'm actually storing something in the server.properties. Besides that,  thinking further this might still be too early. Ideally the message would be shown after the first successful server start. So if a plugin crashes the server the message won't be shown at all since at the next startup the server.properties file already exists.

Leguan16:
You are absolutely right. This is generally for people who set up a server for their first time. They won't have plugins if they do it for their first time. I will change it to the global config tomorrow. 

Another thing: We could send the same message to the first person joining the server. Or would that be too much?

patches/server/1054-Add-onboarding-message-on-initial-start.patch
```diff
@@ -0,0 +1,52 @@
...
+     }
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index d06185566b447c432d4dc2e3ba04d121bcdbc71b..035f3dc291f75128e913c4bc940fa8c4a38df0cf 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -1155,6 +1155,17 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+             long tickSection = Util.getNanos();
+             long currentTime;
+             // Paper end - further improve server tick loop
++            // Paper start - Add info if it is the initial start
```
codebycam: "Onboarding message for initial server start"?

patches/server/1054-Add-onboarding-message-on-initial-start.patch
```diff
@@ -0,0 +1,52 @@
...
++            if (io.papermc.paper.configuration.GlobalConfiguration.get().isFirstStart) {
++                LOGGER.info("##############################################################");
++                LOGGER.info("It looks like this is the first time you start this server.");
++                LOGGER.info("If this is the case then it is recommended you check out the");
++                LOGGER.info("documentation about the next step after the first startup.");
++                LOGGER.info("For more information visit <insert link>");
++                LOGGER.info("##############################################################");
```
codebycam: This could probably be condensed a little?
```
This is the first time you're starting this server. 
It's recommended you read our 'Getting Started' documentation for guidance.
View this and more helpful information here: <link>
```

codebycam (changes requested):

CodeByCam:
> Another thing: We could send the same message to the first person joining the server. Or would that be too much?
Yeah I definitely wouldn't be sending the message to the first person joining. Just having it in console is fine in my opinion.

patches/server/1054-Add-onboarding-message-on-initial-start.patch
Leguan16: Sounds great!

Leguan16:
> If we're going to add an onboarding message, we really should have an onboarding docs page. Just linking to a category is not really accessible, IMHO, and just leaves more "where do I go" questions than it answers;

That was the plan anyway I think. 

> We might also want to consider adding a header message to the GUI app too, in order to provide a clickable link for people not using a terminal

Good idea.

Leguan16:
I addressed the comments and added a message in the GUI. It is centered by default, idk if we leave it that way. Also, should we display it always? IMO yeah. 
image

olijeffers0n:
Looks good GUI wise - is the link clickable? It can stay for the entire of the first launch I think. 

Once I have made my page you can use that as a link - PR blocked by docs for now.

Leguan16:
> It can stay for the entire of the first launch I think.

I am more talking of a permanent component of this GUI. Like it doesn't really take up that much space and I think the GUI is used by relatively new people anyways. 

> is the link clickable?

yes.

> Once I have made my page you can use that as a link - PR blocked by docs for now.

Obviously, I just used it as a placeholder.

granny:
1. add a colon after "you can visit"   `...you can visit: <link>`
1. make it obvious that the link is clickable, e.g. underlined and set to the color blue

> I am more talking of a permanent component of this GUI. Like it doesn't really take up that much space and I think the GUI is used by relatively new people anyways.

Wouldn't hurt to permanently have it available in the GUI. It could also be abstracted into some kind of banner with rotating text.

olijeffers0n:
Docs have been merged :) - https://docs.papermc.io/paper/next-steps

Leguan16:
Thanks Ollie. 
I changed the link, made it clear that its clickable (decided against a blue link since i don't think it would fit in the GUI if everything else is just black) and added more catches.  
image

olijeffers0n:
Yeah I like that, nice work

lynxplay (approved):

Machine-Maker (approved):

PR #10314, opened 2024-03-10
Fix World.getWorldType not detecting amplified and large biomes

LeonTG:
I fixed this in my Paper fork, I figured this was a good enough change to PR upstream so other people can enjoy the fix!

It works by accessing the settings of the NoiseBasedChunkGenerator and checking if it has amplified or large biome settings! and I've tested it to make sure it works too

First time making a PR to a project like this, so hopefully I did it correctly!

### DIFF
--- a/patches/api/0466-Fix-World.getWorldType-not-detecting-amplified-and-l.patch
+++ b/patches/api/0466-Fix-World.getWorldType-not-detecting-amplified-and-l.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: LeonTG <leontg77@gmail.com>
+Date: Sun, 10 Mar 2024 20:52:52 +0100
+Subject: [PATCH] Fix World.getWorldType not detecting amplified and large
+ biomes
+
+
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index c5fe36050eeaff80cfb989fe2f38370215af6fe5..851e79257cf765a389a6c712f417fea05c453c8e 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -2745,12 +2745,9 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+      * Gets the type of this world.
+      *
+      * @return Type of this world.
+-     * @deprecated world type is only used to select the default word generation
+-     * settings and is not stored in Vanilla worlds, making it impossible for
+-     * this method to always return the correct value.
+      */
+     @Nullable
+-    @Deprecated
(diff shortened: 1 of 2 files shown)


patches/api/0466-Fix-World.getWorldType-not-detecting-amplified-and-l.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index c5fe36050eeaff80cfb989fe2f38370215af6fe5..851e79257cf765a389a6c712f417fea05c453c8e 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -2745,12 +2745,9 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
+      * Gets the type of this world.
+      *
+      * @return Type of this world.
+-     * @deprecated world type is only used to select the default word generation
```
electronicboy, line 16: This should be left deprecated, the concept of a WorldType no longer exists, and the enum itself is even outdated because of the fact that this no longer exists.

While we can match the existing setting templates for now, this is not a mechanism which makes sense in the long run, IMHO

LeonTG: Should I remove that part of the patch entirely or is it worth making it change the wording of the java doc deprecation explaination instead?

LeonTG: I just thought of another idea, would methods like `isFlat()`, `isAmplified()` and `isLargeBiomes()` be a better alternative for the API portion of this?

Machine-Maker: I disagree with keeping it deprecated. While I agree that `WorldType` is not a good representation of how worlds work now, there isn't an alternative yet and WorldType itself is not deprecated. This method was only deprecated because of the bug that this PR fixes, so un-deprecated it seems appropriate. It can be re-deprecated with a different message when alternative API is added.

Machine-Maker:
After some more discussion, I think the immediate solution is to add this fix, but at the same time, also add the missing options to WorldType, which include the missing NoiseGeneratorSettings types, like caves, and floating islands.

LeonTG:
> After some more discussion, I think the immediate solution is to add this fix, but at the same time, also add the missing options to WorldType, which include the missing NoiseGeneratorSettings types, like caves, and floating islands.

Alright I can do that

What about my idea of also adding `isFlat()`, `isAmplified()` methods? should I add that too or leave it be for you guys to figure out a better plan?

Machine-Maker:
> What about my idea of also adding isFlat(), isAmplified() methods?

No, I don't think so.

PR #10327, opened 2024-03-14
Implemented EntityDamageDoorEvent

LB45440078L:
# EntityDamageDoorEvent: Handle Door Damage Interactions

## Description
I've introduced a new event called `EntityDamageDoorEvent` in the PaperMC server software. This event is triggered when an entity (currently limited to Zombies and Pillagers by default) damages a door. It provides essential information about the damaged door, the break time, and the total door break time.

This implementation takes into account differences in breaking time between zombies (240 ticks) and pillagers (60 ticks).

## Event Details
- **Event Name**: `EntityDamageDoorEvent`
- **Trigger**: When an entity damages a door
- **Purpose**: To allow developers to handle door damage interactions

## Key Components
1. **Entity**: The damaging entity (e.g., Zombie, Pillager).
2. **Door**: The door being damaged.
3. **Break Time**: The ticks spent breaking the door.
4. **Door Break Time**: The total ticks required to break the door.

## Usage Example
```java
@EventHandler
public void onEntityDamageDoor(EntityDamageDoorEvent event) {
    Entity damagingEntity = event.getEntity();
    Door damagedDoor = event.getDamagedDoor();
    int breakTime = event.getBreakTime();
    int totalBreakTime = event.getDoorBreakTime();
    
    float percentage = ((float) breakTime / totalBreakTime) * 100f;
    if (percentage > 50.0f) {
        e.getNearbyEntities(5, 3, 5).stream()
            .filter(Player.class::isInstance)
            .map(Player.class::cast)
            .forEach(p -> p.sendMessage("Your home door is halfway broken, watch out dude :("));
    }
    // Additional logic...
}
```


patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,44 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/entity/ai/goal/BreakDoorGoal.java b/src/main/java/net/minecraft/world/entity/ai/goal/BreakDoorGoal.java
+index a85885ee51df585fa11ae9f8fcd67ff2a71c5a18..0d2de8e8747795393631f851cb941b268aa0a4b0 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/goal/BreakDoorGoal.java
++++ b/src/main/java/net/minecraft/world/entity/ai/goal/BreakDoorGoal.java
+@@ -1,10 +1,12 @@
+ package net.minecraft.world.entity.ai.goal;
+ 
+ import java.util.function.Predicate;
++import io.papermc.paper.event.entity.EntityDamageDoorEvent;
```
lynxplay: Please checkout our contribution guidelines. Inline these imports.

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,44 @@
...
+ public class BreakDoorGoal extends DoorInteractGoal {
+ 
+@@ -55,6 +57,20 @@ public class BreakDoorGoal extends DoorInteractGoal {
+     @Override
+     public void tick() {
+         super.tick();
++
++        // Paper start
++        EntityDamageDoorEvent event = new EntityDamageDoorEvent(this.mob.getBukkitEntity(), new CraftDoor(this.mob.level().getBlockState(this.doorPos)), this.breakTime, this.doorBreakTime);
++        event.callEvent();
```
lynxplay: callEvent() yields a boolean indicating if the event was cancelled. You can use that immediately instead of calling isCancelled again.

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,44 @@
...
+ 
+ public class BreakDoorGoal extends DoorInteractGoal {
+ 
+@@ -55,6 +57,20 @@ public class BreakDoorGoal extends DoorInteractGoal {
+     @Override
+     public void tick() {
+         super.tick();
++
++        // Paper start
++        EntityDamageDoorEvent event = new EntityDamageDoorEvent(this.mob.getBukkitEntity(), new CraftDoor(this.mob.level().getBlockState(this.doorPos)), this.breakTime, this.doorBreakTime);
```
lynxplay: The event is called every tick of the goal, which break your own set up API contract that states this event is only called when a new  progress level is called.

lynxplay: Creating a CraftDoor like this is pretty meh. paper offers a cached nms.BlockState#createCraftBlockData that we can just yield back under the BlockData type.

The BreakDoorEvent does not offer the block data pre-casted to the Door interface, I am unsure if we want to do this here.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++    }
++
++    /**
++     * Gets the cancellation state of this event. A cancelled event will not
++     * be executed in the server, but will still pass to other plugins
++     *
++     * @return true if this event is cancelled
++     */
++    @Override
++    public boolean isCancelled() {
```
lynxplay: No need to copy-paste the javadocs of cancelled here.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++        return breakTime;
++    }
++
++    /**
++     * Set the current progress in ticks that
++     * this entity has spent damaging the door.
++     * @param breakTime The ticks spent damaging the door.
++     */
++    public void setBreakTime(final int breakTime) {
++        Preconditions.checkArgument(breakTime < 0, "The breaking progress must be equal or greater than zero!");
```
lynxplay: Incorrect precondition.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++    private int doorBreakTime;
++
++    public EntityDamageDoorEvent(final @NotNull Entity what, final @NotNull Door damagedDoor, final int breakTime, final int doorBreakTime) {
++        super(what);
++        this.damagedDoor = damagedDoor;
++        this.breakTime = breakTime;
++        this.doorBreakTime = doorBreakTime;
++    }
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
```
lynxplay: Move the static member to the top of the class as every other event does pls :+1:

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++    public int getBreakTime() {
++        return breakTime;
++    }
++
++    /**
++     * Set the current progress in ticks that
++     * this entity has spent damaging the door.
++     * @param breakTime The ticks spent damaging the door.
++     */
++    public void setBreakTime(final int breakTime) {
```
lynxplay: The method, in combination with setDoorBreakTime, feels dangerous.
A quick glance at the internals you are modifying shows that, at least rn, the server does not consider the case where `breakTime > doorBreakTime`.

Plugins could potentially modify this, leading to the zombie never breaking the door.

I am unsure if that is something we want to allow.
Might be worth preventing this via preconditions too.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++        Preconditions.checkArgument(breakTime < 0, "The breaking progress must be equal or greater than zero!");
++        this.breakTime = breakTime;
++    }
++
++    /**
++     * Get the total amount of ticks required to break
++     * the door that the entity is currently damaging.
++     * @return The amount of ticks necessary for the entity to break the door.
++     */
++    public int getDoorBreakTime() {
```
lynxplay: Can you please add jetbrains' Range annotation to these methods.
We don't need to expose the internal value, just a positive range from 0 to Interger.MAX_VALUE suffices.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,124 @@
...
++     * @return The breaking progress value.
++     */
++    public int getBreakTime() {
++        return breakTime;
++    }
++
++    /**
++     * Set the current progress in ticks that
++     * this entity has spent damaging the door.
++     * @param breakTime The ticks spent damaging the door.
```
lynxplay: Missing empty line before the param description. (see other methods too, e.g. before (at)return too)

lynxplay (changes requested):

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Hello, I've been fixing all other comments, however I never heard of this.

How can I access this method?

lynxplay: `this.mob.level().getBlockState(this.doorPos).createCraftBlockData()`

in this case. It yields CraftBlockData, not door, but as mentioned previously, I am unsure if we want to force the Door type into the API there anyway, it might be better to just have `BlockData getDamagedDoor()`

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Fixed for all.

I've added preconditions for both, so one isn't greater than the other, and other isn't smaller than other.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Fixed.

patches/api/0466-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Removed. Fixed.

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: I changed to this. I'll wait for other's opinion on it.

LB45440078L: Made the event api explaination javadoc accurate according to implementation.

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Used this. Fixed.

patches/server/1054-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
LB45440078L: Removed imports, used inlines. Fixed.

lynxplay (changes requested):
Please merge the patches, instead of committing your changes as a new patch.
See our contribution.md for instructions.

Moving break time up also seems a bit meh in terms of diff.
What is stopping us from just passing this.breakTime + 1 to the event?

Looking through the code a bit more, I'd also add a notice in the javadocs that this even will be fired even if breaking the door is disallowed by the mob_griefing gamerule as that may be wrongly assumed to not be the case.

You also missed the `import org.bukkit.craftbukkit.block.impl.CraftDoor;` line in your import inlining.

Beyond that, is it worth to expose the "swing arm" part of the logic to the event?
E.g. allow plugins to define/overide if the entity will swing their arms in the damage tick.

Leguan16:
If you don't know what lynxplay is talking about: Read the contributing guidelines here: https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md

CMarcoo:
yes I noticed I forgot to merge my patches :(

lynxplay:
Are you going to fix the patch situation or do you want me to do that?

CMarcoo:
> Are you going to fix the patch situation or do you want me to do that?

Yes please. I truly don't know how to do it. I tried merging my patches and it became even worse.

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,123 @@
...
++import org.jetbrains.annotations.Range;
++
++/**
++ * Called for every tick an entity is damaging a door.
++ * <p>
++ * This event is only called when a new level of progress
++ * has been reached by the damaging entity.
++ */
++public class EntityDamageDoorEvent extends EntityEvent implements Cancellable {
++
```
maxcom1: So the event is called every tick OR only when new damage progress level is displayed? This is a bit unclear, at least for me

LB45440078L: Every tick that the door is being progressively damaged to its destruction.

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,132 @@
...
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private final BlockData damagedDoor;
++    private boolean cancelled;
++    private int breakTime;
++    private final int doorBreakTime;
++    private boolean playEffect;
++
++    public EntityDamageDoorEvent(final @NotNull Entity what, final @NotNull BlockData damagedDoor, final int breakTime, final int doorBreakTime, boolean playEffect) {
```
Lulu13022002: Missing internal annotation and what is not a good name here

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,132 @@
...
++    }
++
++    /**
++     * Set the current progress in ticks that
++     * this entity has spent damaging the door.
++     *
++     * @param breakTime The ticks spent damaging the door.
++     */
++    public void setBreakTime(@Range(from = 0, to = Integer.MAX_VALUE) final int breakTime) {
++        Preconditions.checkArgument(breakTime >= 0, "The breaking progress must be at least -1!");
```
Lulu13022002: This message doesn't match that check and doesn't allow -1

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,132 @@
...
++
++/**
++ * Called for every tick an entity is damaging a door.
++ * <p>
++ * This event is only called when a new level of progress
++ * has been reached by the damaging entity.
++ */
++public class EntityDamageDoorEvent extends EntityEvent implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
```
Lulu13022002: Missing a new line after this field

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,132 @@
...
++import org.bukkit.event.entity.EntityEvent;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Range;
++
++/**
++ * Called for every tick an entity is damaging a door.
++ * <p>
++ * This event is only called when a new level of progress
++ * has been reached by the damaging entity.
++ */
```
Lulu13022002: That's not true? the event is called each tick but the mob will take more than 10 ticks to break the doors

patches/server/1056-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,31 @@
...
++++ b/src/main/java/net/minecraft/world/entity/ai/goal/BreakDoorGoal.java
+@@ -55,7 +55,19 @@ public class BreakDoorGoal extends DoorInteractGoal {
+     @Override
+     public void tick() {
+         super.tick();
+-        if (this.mob.getRandom().nextInt(20) == 0) {
++        // Paper start - damage door event
++        boolean playEffect = this.mob.getRandom().nextInt(20) == 0;
++        io.papermc.paper.event.entity.EntityDamageDoorEvent event = new io.papermc.paper.event.entity.EntityDamageDoorEvent(this.mob.getBukkitEntity(), this.mob.level().getBlockState(this.doorPos).createCraftBlockData(), this.breakTime, this.doorBreakTime, playEffect);
++        if (event.callEvent()) {
```
Lulu13022002: This is wrong and break the vanilla behavior like the event was always cancelled

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,132 @@
...
++        this.breakTime = breakTime;
++    }
++
++    /**
++     * Get the total amount of ticks required to break
++     * the door that the entity is currently damaging.
++     *
++     * @return The amount of ticks necessary for the entity to break the door.
++     */
++    @Range(from = 0, to = Integer.MAX_VALUE)
```
Lulu13022002: This return -1 for zombies so the current range is wrong but i think it would be better to call the getDoorBreakTime() that fallback to the default value of 240 ticks for the other constructor.

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
Owen1212055: Not allowing -1

patches/api/0469-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,140 @@
...
++
++/**
++ * Called for every tick an entity is damaging a door.
++ */
++public class EntityDamageDoorEvent extends EntityEvent implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    private final BlockData damagedDoor;
++    private boolean cancelled;
```
Lulu13022002, line 33: This field should be at the bottom of this group

patches/api/0469-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,140 @@
...
++    @Range(from = 0, to = Integer.MAX_VALUE)
++    public int getCurrentBreakTicks() {
++        return this.currentBreakTicks;
++    }
++
++    /**
++     * Set the current progress in ticks that
++     * this entity has spent damaging the door.
++     * <p>
++     * To instantly break this door, supply {@link EntityDamageDoorEvent#getRequiredBreakTicks()} - 1 into this
```
Lulu13022002, line 74: No need to reference the event itself in that link since you're already in that class

patches/api/0469-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,140 @@
...
++
++    @Override
++    public boolean isCancelled() {
++        return this.cancelled;
++    }
++
++    /**
++     * Cancelling this event will reset the break progress of this door.
++     *
++     * @param cancel true if you wish to cancel this event
```
Lulu13022002, line 124: This is redundant with the interface

patches/api/0469-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
```diff
@@ -0,0 +1,140 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: CMarcoo <cmarco.org@gmail.com>
+Date: Thu, 14 Mar 2024 17:04:43 +0100
+Subject: [PATCH] Added EntityDamageDoorEvent and implemented into Server.
```
Lulu13022002, line 4: The second part of the name is not needed and implicit and no dots are required at the end

patches/api/0468-Added-EntityDamageDoorEvent-and-implemented-into-Ser.patch
Lulu13022002: This is not resolved yet

Issue #10274, opened 2024-02-21
Expand Hopper BlockState API


Fusezion:
### Is your feature request related to a problem?

Hoppers have an nbt tag TransferCooldown which is the amount of ticks left between item transfers.
There's a spigot config setting to change the default behavior but no provided method to change it within the bukkit api.

### Describe the solution you'd like.

Expand the existing Hopper blockstate class to include a `Hopper#setTransferCooldown(int)` and `Hopper#getTransferCooldown()`

### Describe alternatives you've considered.

I've only been able to achieve this currently by changing the nbt tag manually, which is not a very efficient method in larger scales

lynxplay:
Resolved by 99a641690abd3d3669249ce10243e0739a13af33

PR #10328, opened 2024-03-15
Expand Hopper BlockState API

viciscat:
This is a PR for the issue right below
--> #10274 <--

A quick PR that adds 2 new methods to the `Hopper` interface:
`setTransferCooldown()` and `getTransferCooldown()`
They change the transfer cooldown of the hopper before it does an item transfer or sucks in an item.


Machine-Maker (approved):

asdanjer:
just wantet to say very nice that this was added. really wantet that for a bit now.

PR #10333, opened 2024-03-16
Clone mutable types in events when changes are discarded 

Machine-Maker:


Machine-Maker:
Updated this PR to clone mutable types like Location, Vector, and BlockData in event getters in cases where any mutations to them are of no consequence to the server. This prevents event listeners from accidentally mutating the value that subsequent event listeners are trying to access.

kennytv (approved):

lynxplay (approved):

PR #10335, opened 2024-03-17
Add permission check for spawn protection

TheMeinerLP:
Problem:
If you want to allow builders or other members to build at the spawn but don't want to give them an operator. So Vanilla does not currently offer any possibility. 

Solution:
Without doing anything hacky, the PR adds a configuration option to Misc on Paper which overrides this problem in combination with a permission.

Test possibility:
- 2 players join the server
- One of two gets OP, player 2 can no longer mine within the spawn protection

Here is the JoinEvent test how to test this:
```java
    @EventHandler
    public void onJoin(PlayerJoinEvent event) {
        PermissionAttachment permissionAttachment = event.getPlayer().addAttachment(this);
        permissionAttachment.setPermission("bukkit.environment.spawnprotection.ignore", true);
    }
```

### DIFF
--- a/patches/server/1057-Add-permission-check-for-spawn-protection.patch
+++ b/patches/server/1057-Add-permission-check-for-spawn-protection.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheMeinerLP <p.glanz@madfix.me>
+Date: Sun, 17 Mar 2024 20:53:42 +0100
+Subject: [PATCH] Add permission check for spawn protection
+
+
+diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+index 68d268b6fff126e8645b6deec3fb549ea2286b77..2c923159a4acca65f637f8dd93731bf2213589c8 100644
+--- a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
++++ b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
+@@ -575,6 +575,8 @@ public class DedicatedServer extends MinecraftServer implements ServerInterface
+             return false;
+         } else if (this.getPlayerList().isOp(player.getGameProfile())) {
+             return false;
++        } else if (io.papermc.paper.configuration.GlobalConfiguration.get().misc.allowOverrideOpBehaviorForSpawnProtection && player.getBukkitEntity().hasPermission("paper.environment.spawnprotection.bypass")) { // Paper start - Add permission check for spawn protection
++            return false; // Paper end - Add permission check for spawn protection
+         } else if (this.getSpawnProtectionRadius() <= 0) {
+             return false;
+         } else {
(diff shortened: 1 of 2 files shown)


lynxplay:
What is the gain of this over just disabling spawn protection and e.g. locking it down with world guard?

TheMeinerLP:
To keep the vanilla style and not be dependent on 2 plugins

patches/server/1057-Add-permission-check-for-spawn-protection.patch
```diff
@@ -0,0 +1,33 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+index f2e30d8743a97c0541808dc5c03300c0d881b843..408c9d15275a1b035ea4e30f2e9a4be95bdec28e 100644
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -332,6 +332,9 @@ public class GlobalConfiguration extends ConfigurationPart {
+         @Comment("See https://luckformula.emc.gs")
+         public boolean useAlternativeLuckFormula = false;
+         public boolean useDimensionTypeForCustomSpawners = false;
++
```
yannicklamprecht: nit: remove empty line.

electronicboy:
Not fond of the perm node, we're not bukkit, it's it's a bypass, not an ignore
Configuration file changes go into the configuration patch
and the comment needs updating

I really dislike the notion of having a 1 liner patch for based around mechanism we generally do not encourage people to use, bleh

lynxplay:
Yea I mean "vanilla does not have that ability" I guess is true, but this isn't vanilla and spawn protection is far from a best practice to use when trying to limit build access. It doesn't have a flag system like world guard nor can you deploy it in different regions.

It is also easily bypassed by pistons, had weird quirks like not being active without an operator etc.
Mob griefing also iirc works in there? It's far far from a good solution for servers. A bandaid permission that people can use when the proper way is a protection plugin seem more like a false sense of security than anything.

TheMeinerLP:
The goal is not to be the "best solution", but simply to give players the opportunity to build on the spawn without having to come straight to them with operators. 
Operators could then ban players or do other things. In this sense, a permission here would make more sense than coming around the corner with the operator or the whole plugin horde.

TheMeinerLP:
I would like to point out that either nobody is allowed to have OP or the spawn protection must be set to 0. There is no in-between. This would be a way to allow this intermediate path and keep the server vanilla as much as possible. Because WorldEdit/FAWE and WorldGuard on top only to allow people to build on spawn via "flags". I think that's excessive from my admin and dev point of view

lynxplay:
I mean, this way you are adding a permissions plugin into the mix.
Something that may also not be present on the "vanilla" server. The plugin "horde" is now 1 plugin instead of 2. Still with the same previously stated issues and now you install one permissions plugin instead of two (WG + WE) to allow builders to make use of a barely functioning protection mechanism.

I digress however, while I personally don't care for this patch, other members might so we can leave this up for discussion.

lynxplay:
With a heavy heart, I'll close this PR after some more discussion.
The spawn protection logic exists, but is by far enough to protect your spawn.

If you want builders to be the only ones building and changing your spawn, you'll need a proper plugin like WG anyway.
The fact that WG requires WorldEdit is, to that degree, annoying and I am unsure if you can lobby them to bundle some form of "minimal region selection" bs, but it doesn't change the fact that this permission flag is not useful for anything beyond the false hope that spawn protection is a proper region protection concept.

This patch makes sense in forks that care for such a niche usecase on their specific server setup, but on the broad spectrum that is paper I don't think it is worth adding :+1:

TheMeinerLP:
Thanks for the feedback. I will hold it in my own fork

PR #10346, opened 2024-03-21
Move invisible setting up to entities

FireInstall:
This moves setting an entity invisible from living entities up to all entities (since vanilla supports it) 

Why?
Even for default "invisible" entities like interaction or item_display the Hitbox and line of sight is always visible when pressing F3+B. This can be a problem if they should be hidden.


Owen1212055 (approved):

patches/api/0469-Move-invisible-setting-up-to-entities.patch
```diff
@@ -0,0 +1,65 @@
...
+     void lockFreezeTicks(boolean locked);
+     // Paper end - Freeze Tick Lock API
+ 
++    // Paper start - move invisible setting up to entities
++    /**
++     * Sets whether the entity is invisible or not.
++     *
++     * @param invisible If the entity is invisible
++     */
++    public void setInvisible(boolean invisible);
```
kennytv: Remove the explicit public modifiers

patches/server/1056-Move-invisible-setting-up-to-entities.patch
```diff
@@ -0,0 +1,59 @@
...
++    // Paper start - move invisible setting up to entities
++    @Override
++    public boolean isInvisible() {
++        return this.getHandle().isInvisible();
++    }
++
++    @Override
++    public void setInvisible(boolean invisible) {
++        this.getHandle().persistentInvisibility = invisible;
++        this.getHandle().setSharedFlag(5, invisible);
```
kennytv: Use `FLAG_INVISIBLE`

kennytv (dismissed):

Owen1212055:
(Note for PR author, this is being resolved by the merger)

patches/server/0586-Missing-Entity-API.patch
```diff
@@ -667,6 +669,28 @@ index fc0f0e841dc974d080e1abb9bbafb5165801131f..d657fd2c507a5b215aeab0a5f3e9c2ee
...
++    // Paper start - move up invisibility
++    @Override
++    public boolean isInvisible() {
++        return this.getHandle().isInvisible();
++    }
++
++    @Override
++    public void setInvisible(boolean invisible) {
++        this.getHandle().persistentInvisibility = invisible;
++        this.getHandle().setSharedFlag(net.minecraft.world.entity.Entity.FLAG_INVISIBLE, invisible);
```
Lulu13022002: The fqn is not needed

Machine-Maker (approved):

kashike (approved):

PR #10298, opened 2024-03-04
Add causes to EntityKnockbackByEntityEvent

LoliColleen:
[https://github.com/PaperMC/Paper/issues/10296](url)

PR #10353, opened 2024-03-24
Add cause to knockback events

Leguan16:
A redo of #10298 since the original author doesn't have enough time to work on their PR. 
closes #10598

### DIFF
--- a/patches/api/0114-Add-EntityKnockbackByEntityEvent-and-EntityPushedByE.patch
+++ b/patches/api/0114-Add-EntityKnockbackByEntityEvent-and-EntityPushedByE.patch
@@ -5,10 +5,11 @@ Subject: [PATCH] Add EntityKnockbackByEntityEvent and
  EntityPushedByEntityAttackEvent
 
 Co-authored-by: aerulion <aerulion@gmail.com>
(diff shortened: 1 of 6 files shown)


patches/api/0114-Add-EntityKnockbackByEntityEvent-and-EntityPushedByE.patch
```diff
@@ -127,6 +139,8 @@ index 0000000000000000000000000000000000000000..404bec776244fd776566c81f671f1009
 +        this.acceleration = acceleration.clone();
 +    }
 +
++
++
```
Lulu13022002: This change is not needed

Leguan16: removed it xd

kennytv (approved):

Machine-Maker:
Handled a different way in https://github.com/PaperMC/Paper/pull/10831. We opted to just add our own EntityKnockbackEvent and our own reason there and a bunch of other hierarchy changes that went beyond what this PR was doing.

PR #10358, opened 2024-03-26
Allow player-list API to self un/list

electronicboy:
This is supported behavior by the client, there is no real reason, to my understanding, to prevent this


maxcom1:
Maybe we could add a configuration option for that?

kennytv (approved):

Machine-Maker (approved):

lynxplay (approved):

PR #10359, opened 2024-03-26
Ignore minecart in activation range

Earthcomputer:
Allow Minecarts and boats to be ticked from any distance


Leguan16:
Proper Paper comments are missing. 
See the formatting section in the contributing guidelines. 
Your stream was still fun to watch xd.

molor:
Maybe should add a configuration option for this?

kennytv (approved):
Comments aren't strictly necessary since this is editing a Spigot class. A config option would definitely be nice, but can also be added outside of this RP

lynxplay (approved):

PR #10378, opened 2024-04-01
[ci skip] sync up asm versions

Machine-Maker:
We had different versions of asm on the classpath which is just annoying when searching for stuff.


kennytv:
Closing since you have already (have to) bump it in the upstream pr

PR #10380, opened 2024-04-01
Add Option to unfix tripwire disarming

notTamion:
since people on the discord have been asking about this frequently and it doesn't require more maintenance i think it is fine to add a config option to allow players to enable this.


kennytv:
Doing a double steal to make the original PR author the author of this commit

notTamion:
i am going to charge you for theft lol

PR #10375, opened 2024-03-30
Fix invalid block entities created during world gen

QuickGlare:
This patch fixes https://github.com/PaperMC/Paper/issues/10022 and https://github.com/sachingorkar102/Lootin-plugin/issues/24

When using WorldGenRegion#setBlock on a block position where a BlockEntity exists it doesn't remove the old one, this creates a broken state in the world and makes the clients crash due to an internal server error.

To fix these issue, I'm removing the BlockEntity when WorldGenRegion#setBlock is set (to fix new worlds) and I've changed the LevelChunk#tick method to remove the invalid BlockEntity (to fix the existing worlds)

PR #10383, opened 2024-04-02
Make sure block entities are valid when loading them

QuickGlare:
This PR fixes these 2 issues (https://github.com/PaperMC/Paper/issues/10022 and https://github.com/sachingorkar102/Lootin-plugin/issues/24) but for existing worlds

When a chunk loads it checks all the existing block entities and makes sure they are valid, if they are not the block entities are removed from the world.

### DIFF
--- a/patches/server/1055-make-sure-block-entities-are-valid-when-loading-them.patch
+++ b/patches/server/1055-make-sure-block-entities-are-valid-when-loading-them.patch
@@ -1,0 +1,41 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Pierpaolo Coletta <p.coletta@glyart.com>
+Date: Tue, 2 Apr 2024 20:56:19 +0200
+Subject: [PATCH] make sure block entities are valid when loading them
+
+
+diff --git a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
+index 465458e8a7dbaf9afb32709a71c7b2620d1e1fd2..b0068071c0b377edddf246bd4e5c3534d1e36c61 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
++++ b/src/main/java/net/minecraft/world/level/chunk/LevelChunk.java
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Owen1212055:
This seems more like a bandaid fix, where exactly are these block entities being added from? I would think it would be better to patch where these are coming from.

QuickGlare:
> This seems more like a bandaid fix, where exactly are these block entities being added from? I would think it would be better to patch where these are coming from.

These block entities are coming from this other issue (https://github.com/PaperMC/Paper/pull/10375), the other PR fixes it only for new worlds and this one is a fix for existing ones

Owen1212055:
Is this only an issue for old worlds that use the custom generator override stuff but then load those chunks?

QuickGlare:
> Is this only an issue for old worlds that use the custom generator override stuff but then load those chunks?

Yes, because they would have a Sculk Sensor block entity where there is a Sculk Catalyst block

Owen1212055:
Then maybe we should have this under a config. Not many plugins use this system, and i'd hate to keep this workaround forever.

QuickGlare:
> adding a config option generally offers a level of commitment beyond just having something sitting around for a while, otherwise, we'd need to investigate putting it as part of the forceUpgrade pathway so that we can tell people to just run that before we remove it, but, that seems meh

I didn't check how the forceUpdate flag works in the code but that sounds like a good idea, it's something that doesn't run everytime a chunk loads and it's a thing that servers with corrupted worlds would have to run only once

lynxplay:
After a bit of discussion, I'll have a look at simply porting this check to a plugin that can perform this check via command.
Maintaining this fix in paper for a single issue caused by a shitty spigot api implementation is rather annoying, so a plugin for people affected by this seems more friendly.

I'll leave this open in case my timeline tomorrow explodes and I don't find the time for this soon.

lynxplay:
Threw together a quick plugin to iterate and process all chunks to find and remove these invalid block entities: 
 
https://github.com/lynxplay/block-entity-bin

People that had their worlds corrupted by the respective bug may use the plugin to fix their chunk data.
Any form of issue regarding the plugin is to be opened on the plugin's github page, not paper's repository.
As with any form of world data shenanigans, backup your data before running this.

PR #10387, opened 2024-04-04
Add ItemStack array serialization methods

kennytv:
Serializing multiple items to bytes isn't trivial, so often beginners fall into the trap of using output IO streams and Bukkit serialization instead, so these provide helpful extra methods to deal with item arrays.

Also kept the serialization methods exposed in API as a sort of textbook example/nudge in the right direction of what should be used instead of object streams

### DIFF
--- a/patches/api/0182-Add-Raw-Byte-ItemStack-Serialization.patch
+++ b/patches/api/0182-Add-Raw-Byte-ItemStack-Serialization.patch
@@ -22,7 +22,7 @@ index 6e67fdb091a006d2d13bc2d93db4d55348af4c8f..e41d5d9b810c8816cd0d1eba5fd1ea56
      // Paper end
  }
 diff --git a/src/main/java/org/bukkit/inventory/ItemStack.java b/src/main/java/org/bukkit/inventory/ItemStack.java
-index e9c29fc1db686b80bc2477d78ec2b361b8600b9e..5707d038bdc41448bca7d6ebf4ef5b378809f5a8 100644
+index e9c29fc1db686b80bc2477d78ec2b361b8600b9e..dda3a421953761dfca5f2f6498905c43fb6fe37d 100644
 --- a/src/main/java/org/bukkit/inventory/ItemStack.java
 +++ b/src/main/java/org/bukkit/inventory/ItemStack.java
 @@ -661,6 +661,117 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
@@ -112,7 +112,7 @@ index e9c29fc1db686b80bc2477d78ec2b361b8600b9e..5707d038bdc41448bca7d6ebf4ef5b37
 +     * @return ItemStack array migrated to this version of Minecraft if needed
 +     * @see #deserializeBytes(byte[])
 +     */
-+    public static @Nullable ItemStack @NotNull [] deserializeItemsFromBytes(final byte @NotNull [] bytes) {
++    public static @NotNull ItemStack @NotNull [] deserializeItemsFromBytes(final byte @NotNull [] bytes) {
 +        try (final java.io.ByteArrayInputStream inputStream = new java.io.ByteArrayInputStream(bytes)) {
 +            final java.io.DataInputStream input = new java.io.DataInputStream(inputStream);
 +            final byte version = input.readByte();
--- a/patches/api/0182-Add-Raw-Byte-ItemStack-Serialization.patch
+++ b/patches/api/0182-Add-Raw-Byte-ItemStack-Serialization.patch
@@ -22,10 +22,10 @@ index 6e67fdb091a006d2d13bc2d93db4d55348af4c8f..e41d5d9b810c8816cd0d1eba5fd1ea56
      // Paper end
  }
 diff --git a/src/main/java/org/bukkit/inventory/ItemStack.java b/src/main/java/org/bukkit/inventory/ItemStack.java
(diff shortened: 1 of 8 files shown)


patches/api/0189-Add-Raw-Byte-ItemStack-Serialization.patch
```diff
@@ -50,7 +60,123 @@ index fd3d4dd231d756d51db0155a4c3ad970c4f456ed..1f31ff5b85217a1c631f05f43c5a6583
...
++     * @return bytes representing the items in NBT
++     * @see #serializeAsBytes()
++     */
++    public static byte @NotNull [] serializeItemsAsBytes(java.util.@NotNull Collection<ItemStack> items) {
++        try (final java.io.ByteArrayOutputStream outputStream = new java.io.ByteArrayOutputStream()) {
++            final java.io.DataOutput output = new java.io.DataOutputStream(outputStream);
++            output.writeByte(ARRAY_SERIALIZATION_VERSION);
++            output.writeInt(items.size());
++            for (final ItemStack item : items) {
++                if (item == null) {
```
electronicboy: this should handle empty stacks too

electronicboy (approved):

patches/api/0189-Add-Raw-Byte-ItemStack-Serialization.patch
```diff
@@ -50,7 +60,136 @@ index fd3d4dd231d756d51db0155a4c3ad970c4f456ed..1f31ff5b85217a1c631f05f43c5a6583
...
+--- a/src/main/java/org/bukkit/util/io/BukkitObjectInputStream.java
++++ b/src/main/java/org/bukkit/util/io/BukkitObjectInputStream.java
+@@ -3,8 +3,10 @@ package org.bukkit.util.io;
+ import java.io.IOException;
+ import java.io.InputStream;
+ import java.io.ObjectInputStream;
++import java.util.Collection;
+ import org.bukkit.configuration.serialization.ConfigurationSerializable;
+ import org.bukkit.configuration.serialization.ConfigurationSerialization;
++import org.bukkit.inventory.ItemStack;
```
Owen1212055: stop kenny

Owen1212055 (dismissed):

patches/api/0189-Add-Raw-Byte-ItemStack-Serialization.patch
kennytv: L+ratio https://discord.com/channels/289587909051416579/925530366192779286/1225463811859812382

lynxplay (changes requested):
Brought up in a merge party in VC, but iirc the general notion was to move away from null item stacks in the API as they require a double check of `item != null && !item.isEmpty()`.

Would it make more sense for this method to yield back ItemStack#empty() instead of null if 0 is read?

kennytv:
While the api output still provides null items, I would not enforce air, otherwise you either produce inconsistent results or break easy use from api input

bridgelol:
Something nice for this might be to allow developers to directly invoke input streams upon deserialization and have a method that outputs a ByteArrayOutputStream as well, making it easier to apply compression

Kinqdos:
So whats the consensus now?

patches/api/0187-Add-Raw-Byte-ItemStack-Serialization.patch
```diff
@@ -50,7 +60,117 @@ index d758cea1d8e88937678dbfd0ac72d49b6c160fe0..066f99a1f4cc42cf0e87d495f97a0685
...
++            final ItemStack[] items = new ItemStack[count];
++            for (int i = 0; i < count; i++) {
++                final int length = input.readInt();
++                if (length == 0) {
++                    // Empty item, keep entry as null
++                    continue;
++                }
++
++                final byte[] itemBytes = new byte[length];
++                input.read(itemBytes);
```
gepron1x: consider using readNBytes or readFully? I used the similiar method for deserializing item arrays and, for long byte sequences, read() failed. I fixed the issue by doing
```java
final byte[] itemBytes = input.readNBytes(length);
```

lynxplay (approved):

patches/api/0187-Add-Raw-Byte-ItemStack-Serialization.patch
lynxplay: Resolved under the premise  that we are using a ByteArrayInputStream which does have full access to all data that will ever be read and hence will not require blocking/waiting for the nth byte to be available.

patches/api/0182-Add-Raw-Byte-ItemStack-Serialization.patch
```diff
@@ -50,7 +52,124 @@ index e9c29fc1db686b80bc2477d78ec2b361b8600b9e..4b4c364dad70126dee17aabca2c3da9f
...
++    /**
++     * Deserializes this itemstack from raw NBT bytes.
++     * <p>
++     * If you need a string representation to put into a file, you can for example use {@link java.util.Base64} encoding.
++     *
++     * @param bytes bytes representing an item in NBT
++     * @return ItemStack array migrated to this version of Minecraft if needed
++     * @see #deserializeBytes(byte[])
++     */
++    public static @Nullable ItemStack @NotNull [] deserializeItemsFromBytes(final byte @NotNull [] bytes) {
```
kennytv: Not nullable

lynxplay: True

PR #10388, opened 2024-04-04
Deprecate Bukkit#getLogger

powercasgamer:
https://discord.com/channels/289587909051416579/925530366192779286/1225482929333997609


mbax:
Perhaps note in the deprecation message *why* this is a bad idea to use?

MiniDigger:
> I'd propose additionally wiring it into the SysoutCatcher or similar, so that it nags authors to stop using it. As demonstrated by the SysoutCatcher, an effective way to get plugin authors to stop using something is to get half their userbase to nag them about it.

not instantly, that would be meh, at least give people one cycle to notice the deprecation

powercasgamer:
> Perhaps note in the deprecation message _why_ this is a bad idea to use?

Do you have a preferred message?

Lulu13022002:
This could be merged in the patch "Fix Spigot annotation mistakes"

patches/api/0468-Deprecate-Bukkit-getLogger.patch
```diff
@@ -0,0 +1,36 @@
...
+
+diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
+index db46e3233edf18bd91a6da8612ccc9f0080d5c17..ab532ab39a373cc0e832e23528eb95597f66f334 100644
+--- a/src/main/java/org/bukkit/Bukkit.java
++++ b/src/main/java/org/bukkit/Bukkit.java
+@@ -1056,8 +1056,10 @@ public final class Bukkit {
+      * Returns the primary logger associated with this server instance.
+      *
+      * @return Logger associated with this server
++     * @deprecated This logger doesn't have a plugin prefix and therefore makes it hard for people to debug. Use {@link org.bukkit.plugin.Plugin#getLogger()}
```
electronicboy: Somebody who can tonguetangulate better than me would be appreciated, but, generally;
This logger is owned by the Minecraft server software, for logging plugin messages, you should use a Logger from within your own contexts, i.e. Plugin#getLogger(), or create a logger for your Class (i.e. slf4j) in order to retain contextual information which is used for identifying the source of messages

electronicboy (approved):

lynxplay (approved):

PR #10392, opened 2024-04-06
Add listing API for PlayerProfile

Flo0:
This enables listing, unlisting and updating any of [latyncy, gamemode, displayname] for PlayerProfiles in the tablist.

Burden of tracking those listings is on the caller.

Extends the ClientboundPlayerInfoUpdatePacket class and its nested Entry class for ease of use.

[Test video](https://i.imgur.com/Ust0U15.mp4)


yannicklamprecht (changes requested):
Fixed imports patch should be squashed into into original patch.

yannicklamprecht (approved):
nit: spaces between the packet factory methods  like updateLatency and so on. Probably a non issue for the Paper team.

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
```diff
@@ -0,0 +1,80 @@
...
++     * Burden of tracking the profile is on the caller.
++     * Using profiles of online players is an invalid operation.
++     * Use {@link #listPlayer(Player)} instead.
++     *
++     * @param profile The {@link com.destroystokyo.paper.profile.PlayerProfile} to list.
++     * @param latency The latency of the profile in milliseconds.
++     * @param gameMode The game mode of the profile.
++     * @param displayName The display name of the profile.
++     */
++    void listProfile(@NotNull com.destroystokyo.paper.profile.PlayerProfile profile, int latency, @NotNull GameMode gameMode, @Nullable net.kyori.adventure.text.Component displayName);
```
Machine-Maker, line 27: I'm not satisfied with the name of this method. I think it should be `send` something like the other "packet-only" methods we have in the API. Something like `sendPlayerTabListProfile` or some combinations of those words.

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
```diff
@@ -0,0 +1,80 @@
...
++     * Lists the {@code profile} in the tablist.
++     * Burden of tracking the profile is on the caller.
++     * Using profiles of online players is an invalid operation.
++     * Use {@link #listPlayer(Player)} instead.
++     *
++     * @param profile The {@link com.destroystokyo.paper.profile.PlayerProfile} to list.
++     */
++    default void listProfile(@NotNull com.destroystokyo.paper.profile.PlayerProfile profile) {
++        listProfile(profile, 0, GameMode.SURVIVAL, null);
++    }
```
Machine-Maker, line 39: I'm not sure having this method makes much sense here. These defaults are a little arbitrary. Like 0 ping? why survival? what if its a creative server? I'm not sure we need anything except the full method with all the options.

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
Machine-Maker, line 1: I don't think `listed` is a descriptive enough word to be used alone to describe this kind of API. Along with my msg about the `listProfile` method name, the various `update` names should be changed to be like `sendTabListLatency`, `sendTabListDisplayName`, etc. 

Additionally, there is no method to *remove* a profile, only disable it's listed status. These things are separate and should be controlled separately. Send a fake Profile but also have a separate option for listing it or not. And have methods to change the listed nature of that profile, but also have a method for removing that profile.

All of the methods *except* the ones that sends the initial ClientboundPlayerInfoUpdatePacket do not require the full PlayerProfile, only the UUID, although we may not care and just require that you pass the full PlayerProfile instead, unsure on that.

patches/server/1055-Add-listing-API-for-PlayerProfile.patch
```diff
@@ -0,0 +1,133 @@
...
++        Preconditions.checkNotNull(profile, "profile cannot be null");
++        Preconditions.checkArgument(isOfflineProfile(profile), "using the profile of an online player is not allowed");
++        if (this.getHandle().connection == null) return;
++
++        this.getHandle().connection.send(ClientboundPlayerInfoUpdatePacket.updateDisplayName(profile.getId(), io.papermc.paper.adventure.PaperAdventure.asVanilla(displayName)));
++    }
++
++    @Override
++    public void unlistProfile(@NotNull com.destroystokyo.paper.profile.PlayerProfile profile) {
++        Preconditions.checkNotNull(profile, "profile cannot be null");
```
Machine-Maker, line 118: We prefer using `checkArgument(profile != null)` because that will throw an `IllegalArgumentException` instead of an `NullPointerException`.

Machine-Maker (changes requested):

patches/server/1055-Add-listing-API-for-PlayerProfile.patch
Flo0: Im also not a fan of this check. I was just copying this from another paper method.

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
Flo0: Those default values where taken from the internal EntryBuilder of the ClientboundPlayerInfoUpdatePacket.
Its honestly just a quol method. Would be nice to have a second opinion of this ^^

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
Flo0: Yes makes sense. Ill change those.

patches/api/0468-Add-listing-API-for-PlayerProfile.patch
Flo0: Yes, separating adding and listing makes sense.
I would argue against using UUIDs as it carries some ambiguity (and possible confusion) with it.

Issue #10211, opened 2024-02-01
Reference `InventoryDragEvent` in `InventoryClickEvent` javadoc


imDaniX:
### Is your feature request related to a problem?

There's no info in `InventoryClickEvent` javadoc about `InventoryDragEvent` being called instead when player clicks and drags their cursor. Many developers with no experience with inventory API just don't know this behavior and there's no info about it in the documentation.

### Describe the solution you'd like.

Reference `InventoryDragEvent` in `InventoryClickEvent` javadoc mentioning the behavior. The `@see` tag would also be appreciated.

### Describe alternatives you've considered.

I'm just mentioning the event myself when there are any related issues.

Machine-Maker:
Yeah, this is probably a good thing to do. Especially since craftbukkit breaks the drag behavior making the drag event fire if you move your mouse even one pixel between mousedown and mouseup.

PR #10395, opened 2024-04-07
Referenced InventoryDragEvent in documentation of InventoryClickEvent

seercle:
Hello,
As requested by #10211 , I added some reference about ```InventoryDragEvent``` in ```InventoryClickEvent``` javadoc. I chose to describe the behaviour between Drag and Click events a bit, instead of just referecing with an ```@see``` tag. Is this ok regarding the overall convention about paper javadoc, or is this unnecessary?


patches/api/0468-Reference-InventoryDragEvent-in-InventoryClickEvent.patch
```diff
@@ -0,0 +1,21 @@
...
+
+diff --git a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+index ac67a4e321f43d1ede09dafe2daa1f07de8a923f..7126f3a9f2fd9b529be8063490f2d0bf0839aac8 100644
+--- a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+@@ -16,6 +16,9 @@ import org.jetbrains.annotations.Nullable;
+ /**
+  * This event is called when a player clicks in an inventory.
+  * <p>
++ * In case of a drag action within an inventory, InventoryClickEvent is only called at the beggining of the drag.
```
Leguan16:
```suggestion
+ * In case of a drag action within an inventory, InventoryClickEvent is only called at the beginning of the drag.
```

Leguan16 (commented):
Could maybe be merged into [Fix-upstream-javadocs.patch](https://github.com/PaperMC/Paper/blob/master/patches/api/0056-Fix-upstream-javadocs.patch)

patches/api/0468-Reference-InventoryDragEvent-in-InventoryClickEvent.patch
seercle: Thanks for the typo, also great idea about merging patches, didn't know about this.

patches/api/0468-Reference-InventoryDragEvent-in-InventoryClickEvent.patch
```diff
@@ -0,0 +1,21 @@
...
+index ac67a4e321f43d1ede09dafe2daa1f07de8a923f..7126f3a9f2fd9b529be8063490f2d0bf0839aac8 100644
+--- a/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
++++ b/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+@@ -16,6 +16,9 @@ import org.jetbrains.annotations.Nullable;
+ /**
+  * This event is called when a player clicks in an inventory.
+  * <p>
++ * In case of a drag action within an inventory, InventoryClickEvent is only called at the beginning of the drag.
++ * {@link InventoryDragEvent} is then called at the end of the drag.
++ * <p>
```
Machine-Maker: Half of this isn't true. If its a drag, the click event isn't fired at all and the drag is fired at the end.

Machine-Maker (changes requested):
Can be merged into the javadoc fixes, but that can be done by the person who merges it, doesn't have to be done by OP

Machine-Maker (approved):

PR #10396, opened 2024-04-07
Add CartographyItemEvent and get/setResult for CartographyInventory

### LINKED ISSUE #8749: Event for when a player crafts inside a cartography table

The is already a `SmithItemEvent`, so why not a `CartographyItemEvent` :) . It would be very useful to detect when a player crafts a map / multiple maps at once and get the map data from it.

It would be great to detect when a player crafts a map + get the `MapView` data. It would also be nice to be able to detect every single item that has been crafted using shift + click.

Currently this must be done with a lot of checks using the `InventoryClickEvent` and feels wrong to do it like this. You can also not check the result of multiple crafted maps using shift + click

### Other

_No response_

sowelipililimute:
Fixes #8749.


TheMeinerLP:
Tested with: 
```java
@EventHandler
public void onCartographEvent(CartographyItemEvent event) {
    final CartographyInventory inventory = event.getInventory();
    inventory.setResult(new ItemStack(Material.ACACIA_BOAT));
    event.setCancelled(true);
    event.setResult(Event.Result.DENY);
}
 ```
Result:
Result cannot not be taken from inventory.
image
---
Tested with: 
```java
@EventHandler
public void onCartographEvent(CartographyItemEvent event) {
    event.setCurrentItem(new ItemStack(Material.ACACIA_BOAT));
}
```

Result:
Result is not changing with that code.
image

Is this an expected behaviour that `setCurrentItem` does not change the item?

patches/server/1055-Implement-get-setResult-for-CartographyInventory.patch
```diff
@@ -0,0 +1,31 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventoryCartography.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventoryCartography.java
+index af455b3a4eb74339580dc3ec723c832791c555f0..f26cb926103db5b9e6c80aeccbed1249beee1f7c 100644
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventoryCartography.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventoryCartography.java
+@@ -4,8 +4,19 @@ import net.minecraft.world.Container;
+ import org.bukkit.inventory.CartographyInventory;
+ 
+ public class CraftInventoryCartography extends CraftResultInventory implements CartographyInventory {
+-
```
Machine-Maker: extraneous diff

patches/api/0469-Add-CartographyItemEvent.patch
```diff
@@ -0,0 +1,42 @@
...
++/**
++ * Called when the recipe of an Item is completed inside a cartography table.
++ */
++public class CartographyItemEvent extends InventoryClickEvent {
++
++    public CartographyItemEvent(@NotNull InventoryView view, @NotNull InventoryType.SlotType type, int slot, @NotNull ClickType click, @NotNull InventoryAction action) {
++        super(view, type, slot, click, action);
++    }
++
++    public CartographyItemEvent(@NotNull InventoryView view, @NotNull InventoryType.SlotType type, int slot, @NotNull ClickType click, @NotNull InventoryAction action, int key) {
```
Machine-Maker: Add `@ApiStatus.Internal` to both these constructors. Also, I don't know if other events of this type have them, but we can add Preconditions I think to the constructors to ensure that the Inventory passed in is a CartographyInventory instead of a potential error when someone uses the getter and it throws a ClassCastException

patches/server/1055-Implement-get-setResult-for-CartographyInventory.patch
```diff
@@ -0,0 +1,31 @@
...
++
++    // Paper begin - add getResult/setResult implementation for CartographyInventory
++    @Override
++    public org.bukkit.inventory.ItemStack getResult() {
++        return this.getItem(2);
++    }
++
++    @Override
++    public void setResult(org.bukkit.inventory.ItemStack item) {
++        this.setItem(2, item);
```
Machine-Maker: For the `2` here, use the public static field `CartographyTableMenu.RESULT_SLOT`

patches/server/1056-Implement-CartographyItemEvent.patch
```diff
@@ -0,0 +1,30 @@
...
+diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+index 860d33ac56680babd87358b1d7d2216dac3f7865..5eb8bcb43c019384bfdf21677446deb770ee87db 100644
+--- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
++++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -3101,6 +3101,19 @@ public class ServerGamePacketListenerImpl extends ServerCommonPacketListenerImpl
+                             }
+                         }
+ 
++                        // Paper start - cartography item event
++                        if (packet.getSlotNum() == 2 && top instanceof org.bukkit.inventory.CartographyInventory) {
```
Machine-Maker: Same for the `2` here. You can also use the Java 14 instanceof (e.g. `top instanceof CartographyInventory cartInv`)

Machine-Maker (changes requested):

pontaoski:
poke

lynxplay (approved):

Machine-Maker (approved):

PR #10398, opened 2024-04-08
Added chunk view API

Flo0:
This exposes some functionality of the underlying ChunkTrackingView.

When adding certain objects to chunks, it is often desirable to have actual knowledge about who is currently seeing this chunk, in order to selectively send informations to those viewers.

This can be useful for holograms and similar functionality.


patches/server/1055-Added-API-implementation.patch
```diff
@@ -0,0 +1,61 @@
...
++        });
++        return chunks;
++    }
++
++    @Override
++    public boolean isViewingChunk(org.bukkit.Chunk chunk) {
++        net.minecraft.world.level.ChunkPos chunkPos = new net.minecraft.world.level.ChunkPos(chunk.getX(), chunk.getZ());
++        return this.getHandle().getChunkTrackingView().contains(chunkPos);
++    }
++    // Paper end
```
Cryptite: You can use `this.getHandle().chunkLoader.sentChunks.contains(chunk.getChunkKey());` to access the already-existing list of ChunkKeys stored by the server, rather than constructing a new ChunkPos.

`sentChunks` may need to be made public however, that's what I did on my fork (for nms access reasons)

electronicboy:
I don't see anywhere that the chunk tracking view is actually maintained?

Flo0:
> I don't see anywhere that the chunk tracking view is actually maintained?

Im also just finding out that the chunkTrackingView is pretty much unused? The getter and setter for this field are never called.
I should change the backbone to the RegionizedPlayerChunkLoader...

patches/server/1055-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,58 @@
...
+     // Paper end
+ 
++    // Paper start - Add chunk view API
++    @Override
++    public Set<java.lang.Long> getSentChunkKeys() {
++        return this.getHandle().chunkLoader.getSentChunks();
++    }
++
++    @Override
++    public boolean isChunkSent(org.bukkit.Chunk chunk) {
```
Vrganj: Why not use the chunk key here instead of Chunk, if you opted for using keys above

Cryptite: Or at least provide an extra overload which accepts a key.

Cryptite (changes requested):

patches/server/1055-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,63 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/chunk/system/RegionizedPlayerChunkLoader.java b/src/main/java/io/papermc/paper/chunk/system/RegionizedPlayerChunkLoader.java
+index 1b090f1e79b996e52097afc49c1cec85936653e6..627c54afcb7b1c1e70333b31ee45e3ad75f85e31 100644
+--- a/src/main/java/io/papermc/paper/chunk/system/RegionizedPlayerChunkLoader.java
++++ b/src/main/java/io/papermc/paper/chunk/system/RegionizedPlayerChunkLoader.java
+@@ -33,6 +33,7 @@ import org.bukkit.craftbukkit.entity.CraftPlayer;
+ import org.bukkit.entity.Player;
+ import java.lang.invoke.VarHandle;
+ import java.util.ArrayDeque;
++import java.util.Set;
```
lynxplay: Inline the import please.

patches/api/0469-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,42 @@
...
++     */
++    boolean isChunkSent(org.bukkit.Chunk chunk);
++
++    /**
++     * Checks if the player has been sent a specific chunk.
++     *
++     * @param chunkKey the chunk key to check
++     * @return true if the player has been sent the chunk, false otherwise
++     */
++    boolean isChunkSent(java.lang.Long chunkKey);
```
lynxplay: What is the point of a boxed long here?
The implementation requires a primitive long in the first place. Unless I am mistakes it should be using a fastutil long set.

patches/api/0469-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,42 @@
...
+     void resetIdleDuration();
+     // Paper end
+ 
++    // Paper start - Add chunk view API
++    /**
++     * Gets the set of chunks that have been sent to the player.
++     *
++     * @return an immutable set of chunk keys
++     */
++    java.util.Set<java.lang.Long> getSentChunkKeys();
```
lynxplay: Can be marked with @org.jetbrains.annotations.Unmodifiable.
Additionally, the NotNull annotation.

patches/api/0469-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,42 @@
...
++
++    /**
++     * Checks if the player has been sent a specific chunk.
++     *
++     * @param chunk the chunk to check
++     * @return true if the player has been sent the chunk, false otherwise
++     */
++    boolean isChunkSent(org.bukkit.Chunk chunk);
```
lynxplay: Missing NotNull annotation.
This can also just be a default method tbh.

patches/api/0469-API-for-checking-sent-chunks.patch
```diff
@@ -0,0 +1,42 @@
...
++
++    /**
++     * Checks if the player has been sent a specific chunk.
++     *
++     * @param chunkKey the chunk key to check
```
lynxplay: I'd link to Chunk#getChunkKey here via a `@see` or something as the concept of a chunk key, or well, its creation is at least somewhat explained there.

lynxplay (changes requested):

patches/api/0469-API-for-checking-sent-chunks.patch
Flo0: No use here, changed to primitive.

patches/api/0469-API-for-checking-sent-chunks.patch
Flo0: Unmodifiable has a retention policy of class only i think

lynxplay: Correct. It isn't needed at runtime, but Intellij can infer more about the returned collection via said annotation.

Vrganj:
I think you should return a collection of chunks rather than the keys. If you use the getChunkAt method and pass `false` as the second argument, it'll just construct the chunk without generating it.

Cryptite:
> I think you should return a collection of chunks rather than the keys. If you use the getChunkAt method and pass `false` as the second argument, it'll just construct the chunk without generating it.

At least optionally. Keys is best overall but the overload could be nice.

Flo0:
> I think you should return a collection of chunks rather than the keys. If you use the getChunkAt method and pass `false` as the second argument, it'll just construct the chunk without generating it.

I specifically chose Set because it conveys an important property: There are no duplicates in the returned collection.
An overload method returning Chunks wont be added again. I initially had it but had a convo that made me remove it.

Vrganj:
If you're referring to the conversation with electroniccat, he probably missed the boolean parameter.
Using it that way just constructs a chunk, nothing else. It's pretty much a tuple of chunk coordinates
with the added benefit of a world - which is still much cleaner than a set of magic numbers.

Cryptite:
They aren't magic numbers, they are explicitly derived values that correspond with Chunk Coordinates and commonly represent a lighter way to store data about chunks, (longs vs a Chunk object). 

Granted, a `CraftChunk` is still a pretty small class, but there is at least a non-zero performance impact of the mapping and allocations of `sentChunks` from the long->chunks and (and players usually have many hundreds of chunks loaded). 

I'd still advocate for having both options, but perhaps the higher-ups should weigh in.

electronicboy:
I did clarify on the conversation that I did miss the change to the behavior of getChunkAt, but at least, IMHO, that is part of the headache here, I guess most of my concerns are somewhat mitigated by the fact that we can generally assure that it's not going to create some fun and interesting issues due to world switching if you put the CraftChunk alloc in the correct place

Vrganj:
> They aren't magic numbers, they are explicitly derived values that correspond with Chunk Coordinates and commonly represent a lighter way to store data about chunks, (longs vs a Chunk object).
> 
> Granted, a `CraftChunk` is still a pretty small class, but there is at least a non-zero performance impact of the mapping and allocations of `sentChunks` from long->chunk (and players usually have many hundreds of chunks loaded).
> 
> I'd still advocate for having both options, but perhaps the higher-ups should weigh in.

That's what I thought about block keys until they changed world height and got rid of them...

You're already returning the copy as a set of boxed longs, might as well convert them to Chunks while you're at it.

Vrganj:
Honestly the last thing I'd change is to avoid copying the long set when converting to a chunk set. Looks great otherwise.

Machine-Maker (approved):

Vrganj:
I commited some changes here relating to the cloning & boxing, seems like that's gone ignored.
https://github.com/Flo0/Paper/pull/1

Vrganj:
I made a few changes in https://github.com/Flo0/Paper/pull/1 specifically to avoid unnecessary cloning and boxing, now you've made the Set both immutable and a clone..?

lynxplay:
I made those changes as we were preparing this for a merge today.

We need to clone both, your proposed changes would end up with an unmodifiable view returned by getSentChunkKeys (one that is updated when the player moves) while getSentChunks is a normal clone.

Additional, a view is a dangerous thing given the fact that the server maintained set is not save for concurrent access. Callers would have to clone the returned view when trying to use it off the main thread, otherwise they'll run into terribly obscure errors when trying to access the unmodifiable view from getChunkKeysSent while the server mutates the underlying set as fast util collections don't have extensive error messages for concurrent modifications.

molodador:
> I made those changes as we were preparing this for a merge today.
> 
> We need to clone both, your proposed changes would end up with an unmodifiable view returned by getSentChunkKeys (one that is updated when the player moves) while getSentChunks is a normal clone.
> 
> Additional, a view is a dangerous thing given the fact that the server maintained set is not save for concurrent access. Callers would have to clone the returned view when trying to use it off the main thread, otherwise they'll run into terribly obscure errors when trying to access the unmodifiable view from getChunkKeysSent while the server mutates the underlying set as fast util collections don't have extensive error messages for concurrent modifications.

I disagree. It being an unmodifiable view is clearly documented, and storing such a view for future (especially asynchronous) access is bad practice, and should be avoided precisely because of the pitfalls you describe. If you need the collection later, be it on the main server thread or on another, you clone it; the point of using a view is to leave agency over that to the caller. This is why it is documented as such.

One could argue that exposing a view introduces a greater risk of programmer error. But I would argue that eliminating such risk is, generally, not what we are here to do, nor is it possible in its entirety. The semantics and behaviour of the returned collection are documented and it is on the caller to act in accordance with those semantics -- just as it is expected that one should not try modifying an unmodifiable collection.

lynxplay:
Did I miss anything in the java docs regarding this being supposed to be a view?

It wasn't even a view in the initially proposed pr given it used Set.copyOf, which constructs an array based set.

This would also still not fix the issue that youd have getChunkKeysSent returning a view while getChunksSent would not. We could work around this with a transforming collection, but I honestly do not see the point of this being a view in the first place?

You'd save yourself a single internal clone of the hash table, and that is *only* the case if you need to iterate over all sent chunks as the *isChunkSent* methods do not need the clone in the first place.

Your argument that we are not here to eliminate risks in exposed API is confusing to me. I'd give it to you if the exposed view would throw a proper exception if it is incorrectly used down the line that points out the offending plugin, but this will generally not be the case with fastutil exceptions.
Idk what documentation you referred to, nothing in this PR at least goes indepth of the dangers that using a view on a different thread poses.

molodador:
> Did I miss anything in the java docs regarding this being supposed to be a view?
> 
> It wasn't even a view in the initially proposed pr given it used Set.copyOf, which constructs an array based set.
>
> Idk what documentation you referred to, nothing in this PR at least goes indepth of the dangers that using a view on a different thread poses.

To clarify, I am referring to the modifications proposed by @Vrganj, in his [PR](https://github.com/Flo0/Paper/pull/1) wherein the methods are annotated with the documenting annotations, [@UnmodifiableView](https://javadoc.io/doc/org.jetbrains/annotations/20.1.0/org/jetbrains/annotations/UnmodifiableView.html) and @Unmodifiable, respectively. This should be sufficient. If it is not, the semantics of a view can be explicitly noted on the method javadoc, but this defeats the point of using annotations for that purpose.

> This would also still not fix the issue that youd have getChunkKeysSent returning a view while getChunksSent would not. We could work around this with a transforming collection, but I honestly do not see the point of this being a view in the first place?

I do not see how this is an issue.

> You'd save yourself a single internal clone of the hash table, and that is _only_ the case if you need to iterate over all sent chunks as the _isChunkSent_ methods do not need the clone in the first place.

Given how the returned collections are immutable, there is not much else to be done with them apart from calling ``Set::contains`` or iterating over them. As you correctly pointed out, ``Set::contains`` is already covered by ``isChunkSent``, so ``getChunkKeysSent`` should be optimized for iteration in turn, as that is what it will be primarily, perhaps exclusively, used for. I agree that we could use a transforming collection view for ``getChunksSent`` and in fact I would quite prefer this.

> Your argument that we are not here to eliminate risks in exposed API is confusing to me. I'd give it to you if the exposed view would throw a proper exception if it is incorrectly used down the line that points out the offending plugin, but this will generally not be the case with fastutil exceptions. Idk what documentation you referred to, nothing in this PR at least goes indepth of the dangers that using a view on a different thread poses.

You are right in that many of the fastutil iterators are not fail-fast, but keep in mind that sensible CME errors are in no way guaranteed by fail-fast iterators, either, and iterating over an asynchronously modified (non-thread-safe) collection is inherently unpredictable. Also, note that the 'offending plugin' will be pointed out by any exception still thrown, as fastutil's reading methods are pure, and therefore cannot cause exceptions on our end; only calls made by the holder of the view can result in exceptions, which will in turn point out the offending plugin.

Exposing views is nothing new or out of the norm, and, in my opinion, a certain degree of competence and reviewing of the documentation should be expected of users. The only exception to this expectation that I will admit to is the integrity of the internal server state: exposing the long set as a mutable collection, for example, would be unacceptable. Exposing it as an unmodifiable view, however, poses no risk to internal correctness, and threfore by my measure it is acceptable.

Responsibility for the correctness of user code, provided that easily-accessible documentation exists to ensure ahead of time confidence in it, rests on the user writing it.

Finally, said PR also eliminates an accidental case of long boxing in the chunk key -> chunk conversion. Even if you do not go through with the view, I implore you to use a primitive iterator as in the PR.

lynxplay:
I'll bring up the use of view again in todays merge party regarding this.

Especially in regards to server state, a plugin incorrectly using this method and potentially screwing over the server terribly feels, at least to me, worth the "cost" of  copying the open hash set once.

The long boxing is a good spot, will be included.

lynxplay:
Alright, talked it over with the view vs non view thing with some of the team.
Generally, my argument "protecting someone from getting the set, storing it and then using it off thread" is somewhat weak, agree with you on that end.

The argument of "api precedent" however, is what we are going to base the decision of copying the set on.
There is little to no views, as said early, on the API that directly reference internal server state (as in java Collection views, Entity interface being a "view" is separate).
As such, introducing a view would again be a potential unfriendly move towards developers, no just because of the inability of iterating off thread (even tho that argument is weak) but also regarding the behaviour of the returned set, specifically its mutability.

I went ahead and avoided the long boxing as well as moved to a faster clone method that directly clones the long[] instead of running an addAll() over the contents of the source set.

While I understand that this might not please you, I hope you understand where I am coming from on this decision.
Thank you again for your input on this. The discussion was certainly worth having and reconsidering the usage of a view in depth was not a waste of time, even if it did not end up being used :+1:

lynxplay (approved):

molodador:
Sounds reasonable. Discourse over the matter is good enough for me. It's not that significant of a difference in terms of performance for most users, anyway. ~and I can always reflect/nms into it if i need to, har har har~ I just wanted to bring up the topic to make sure it gets its due consideration before us committing to any given API contract.

PR #10400, opened 2024-04-08
Added API to get player's proxy address

nostalfinals:
## Introduction

This patch added an API to get the player's proxy address.
It will be useful if the server uses multiple proxy instances that support HAProxy Protocol (frp, nginx, etc.) and wants to know which proxy instance the player uses.

## How?

I added a method, which is `Player#getHAProxyAddress`. This method returns an `InetSocketAddress`, if the server didn't enable Proxy Protocol, or the player didn't connect with an HAProxy instance, then it will return null.

## Example

```kotlin
@EventHandler
fun playerJoinEvent(event: PlayerJoinEvent) {
    val player = event.player
    val proxyAddr = player.haproxyAddress

    if (proxyAddr == null) {
        logger.info("Player ${player.name} didn't connect with a proxy!")
        return
    }

    logger.info("Player ${player.name} is using proxy: ${proxyAddr.hostString}:${proxyAddr.port}")
}
```


patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,79 @@
...
+         this.receiving = side;
+diff --git a/src/main/java/net/minecraft/server/network/ServerConnectionListener.java b/src/main/java/net/minecraft/server/network/ServerConnectionListener.java
+index 586521a2cbb1d4dcfb912029f65e4363ec7674a7..44d26da371bd0b9ba970c24c85c62b5640120cd5 100644
+--- a/src/main/java/net/minecraft/server/network/ServerConnectionListener.java
++++ b/src/main/java/net/minecraft/server/network/ServerConnectionListener.java
+@@ -28,6 +28,7 @@ import io.netty.util.Timeout;
+ import io.netty.util.Timer;
+ import java.io.IOException;
+ import java.net.InetAddress;
++import java.net.InetSocketAddress;
```
powercasgamer:
```suggestion

```

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,79 @@
...
+@@ -140,10 +141,17 @@ public class ServerConnectionListener {
+                                         String realaddress = message.sourceAddress();
+                                         int realport = message.sourcePort();
+ 
++                                        // Paper start - Add API to get player's proxy address
++                                        String proxyAddress = message.destinationAddress();
++                                        int proxyPort = message.destinationPort();
++
+                                         SocketAddress socketaddr = new java.net.InetSocketAddress(realaddress, realport);
++                                        SocketAddress proxyAddr = new InetSocketAddress(proxyAddress, proxyPort);
```
powercasgamer:
```suggestion
+                                        SocketAddress proxyAddr = new java.net.InetSocketAddress(proxyAddress, proxyPort);
```

powercasgamer (changes requested):
imports

patches/api/0469-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,27 @@
...
+ 
++    // Paper start - Add API to get player's proxy address
++    /**
++    * Gets the socket address of this player's proxy
++    *
++    * @return the player's proxy address, null if the server didn't enable Proxy Protocol, or the player didn't connect with an HAProxy instance
++    */
++    @Nullable
++    public InetSocketAddress getProxyAddress();
++    // Paper end
```
notTamion: end comments also have identifiers

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,79 @@
...
+--- a/src/main/java/net/minecraft/network/Connection.java
++++ b/src/main/java/net/minecraft/network/Connection.java
+@@ -148,6 +148,10 @@ public class Connection extends SimpleChannelInboundHandler<Packet<?>> {
+         this.stopReadingPackets = true;
+     }
+     // Paper end - packet limiter
++    // Paper start - Add API to get player's proxy address
++    @Nullable
++    public SocketAddress proxyAddress;
++    // Paper end
```
notTamion: same here

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,79 @@
...
++                                        String proxyAddress = message.destinationAddress();
++                                        int proxyPort = message.destinationPort();
++
+                                         SocketAddress socketaddr = new java.net.InetSocketAddress(realaddress, realport);
++                                        SocketAddress proxyAddr = new InetSocketAddress(proxyAddress, proxyPort);
+ 
+                                         Connection connection = (Connection) channel.pipeline().get("packet_handler");
+                                         connection.address = socketaddr;
++                                        connection.proxyAddress = proxyAddr;
++                                        // Paper end
```
notTamion: same here + you can move your stuff below all of mojangs code so you don't also have all that vanilla code between your comments.

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,79 @@
...
++
++        SocketAddress addr = this.getHandle().connection.connection.proxyAddress;
++
++        if (addr instanceof InetSocketAddress) {
++            return (InetSocketAddress) addr;
++        }
++
++        return null;
++    }
++    // Paper end
```
notTamion: same here

notTamion (changes requested):

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
notTamion: i think i should also just add here that we probably don't necessarily need to create all those variables to match mojangs style here.

patches/api/0469-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,27 @@
...
++++ b/src/main/java/org/bukkit/entity/Player.java
+@@ -248,6 +248,16 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+     @Nullable
+     public InetSocketAddress getAddress();
+ 
++    // Paper start - Add API to get player's proxy address
++    /**
++    * Gets the socket address of this player's proxy
++    *
++    * @return the player's proxy address, null if the server didn't enable Proxy Protocol, or the player didn't connect with an HAProxy instance
```
powercasgamer:
```suggestion
+    * @return the player's proxy address, null if the server doesn't have Proxy Protocol enabled, or the player didn't connect to an HAProxy instance
```
Something like this?

patches/server/1055-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,69 @@
...
+--- a/src/main/java/net/minecraft/network/Connection.java
++++ b/src/main/java/net/minecraft/network/Connection.java
+@@ -148,6 +148,10 @@ public class Connection extends SimpleChannelInboundHandler<Packet<?>> {
+         this.stopReadingPackets = true;
+     }
+     // Paper end - packet limiter
++    // Paper start - Add API to get player's proxy address
++    @Nullable
++    public SocketAddress proxyAddress;
++    // Paper end - Add API to get player's proxy address
```
powercasgamer:
```suggestion
+    @Nullable
+    public SocketAddress proxyAddress; // Paper - Add API to get player's proxy address
```

powercasgamer (changes requested):

patches/api/0469-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,27 @@
...
+     public InetSocketAddress getAddress();
+ 
++    // Paper start - Add API to get player's proxy address
++    /**
++    * Gets the socket address of this player's proxy
++    *
++    * @return the player's proxy address, null if the server doesn't have Proxy Protocol enabled, or the player didn't connect to an HAProxy instance
++    */
++    @Nullable
++    public InetSocketAddress getProxyAddress();
```
electronicboy: This should generally be more declarative to what this is actually exposing, as it generally feels detached from what it actually exposes

lynxplay (commented):
The naming here feels a bit off.
I don#t have a better suggestion at hand, but given how rare the usage of HA proxies infront of servers are, I feel like the name `proxyAddress` does not properly separate that this method is exlusivly for HA proxies and not normal "proxies" as the minecraft community would understand (e.g. velo/waterfall/bungee).

patches/api/0469-Added-API-to-get-player-s-proxy-address.patch
yannicklamprecht: getProxyJoinedWithAddress() or is that even worse?

nostalfinals: I'm considering if just name it as `getHAProxyAddress` directly will be better.

lynxplay: I like it too 👍

nostalfinals:
I've already changed the method name and updated this to 1.20.5.

patches/server/1046-Added-API-to-get-player-s-proxy-address.patch
```diff
@@ -0,0 +1,67 @@
...
++    // Paper start - Add API to get player's proxy address
++    @Override
++    public @Nullable InetSocketAddress getHAProxyAddress() {
++        if (this.getHandle().connection == null) {
++            return null;
++        }
++
++        SocketAddress addr = this.getHandle().connection.connection.proxyAddress;
++
++        if (addr instanceof InetSocketAddress) {
```
kennytv: pattern matching

kennytv (approved):

patches/server/1046-Added-API-to-get-player-s-proxy-address.patch
yannicklamprecht:
```java
return this.getHandle().connection.connection.proxyAddress instanceof InetSocketAddress inetSocketAddress ? inetSocketAddress : null;
```


or alternative keeping the GuardClause.

```java
    @Override
    public @Nullable InetSocketAddress getHAProxyAddress() {
        if (this.getHandle().connection == null || !(this.getHandle().connection.connection.proxyAddress instanceof InetSocketAddress inetSocketAddress)) {
            return null;
        }
        return inetSocketAddress;
    }

 ```

lynxplay (approved):

Machine-Maker (approved):

PR #10411, opened 2024-04-12
Remove dead code (LegacyResult)

brickmonster:
I've had this patch to remove dead code since June, is there any interest in it? If there is I'll merge it in properly and submit it.

### DIFF
--- a/patches/server/0010-Adventure.patch
+++ b/patches/server/0010-Adventure.patch
@@ -14,7 +14,7 @@ Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
 
 diff --git a/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java b/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..215ccf169d4e8522da5495768d33cf3e5cd92690
+index 0000000000000000000000000000000000000000..9090b0ebcd3ceecb14f83b274ad364d6ea17d1f4
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
 @@ -0,0 +1,418 @@
@@ -416,7 +416,7 @@ index 0000000000000000000000000000000000000000..215ccf169d4e8522da5495768d33cf3e
 +        });
 +    }
 +
-+    static @Nullable String tryCollapseToString(final Component component) {
++    public static @Nullable String tryCollapseToString(final Component component) {
 +        if (component instanceof final TextComponent textComponent) {
 +            if (component.children().isEmpty() && component.style().isEmpty()) {
 +                return textComponent.content();
@@ -613,10 +613,10 @@ index 0000000000000000000000000000000000000000..23bd6d2d8fed5a3491e856f8b875456d
 +}
 diff --git a/src/main/java/io/papermc/paper/adventure/ChatProcessor.java b/src/main/java/io/papermc/paper/adventure/ChatProcessor.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..7a2e7a55a3f6b24c7a3b8c838b66151a820f5261
+index 0000000000000000000000000000000000000000..65ad76dd43c9a8f96333572f795967dc603064c1
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/adventure/ChatProcessor.java
-@@ -0,0 +1,391 @@
+@@ -0,0 +1,379 @@
 +package io.papermc.paper.adventure;
 +
 +import io.papermc.paper.chat.ChatRenderer;
@@ -638,6 +638,7 @@ index 0000000000000000000000000000000000000000..7a2e7a55a3f6b24c7a3b8c838b66151a
 +import net.kyori.adventure.audience.ForwardingAudience;
 +import net.kyori.adventure.key.Key;
 +import net.kyori.adventure.text.Component;
++import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
 +import net.minecraft.Optionull;
 +import net.minecraft.Util;
 +import net.minecraft.core.registries.Registries;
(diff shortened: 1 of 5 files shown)


yannicklamprecht (commented):
Original patch should be adapted instead.
I doubt that it is unused.

brickmonster:
Yeah I'll adapt the original patch if there's interest. It's just that I didn't want to touch the Adventure patch if there's no interest.

I've had a look over it again, the only reason this code exists is to support chat preview, which was removed after a very brief existence in minecraft. I don't understand why AsyncPlayerChatPreviewEvent is still being fired if the feature was removed, it's a nonsense event.

Machine-Maker:
Does upstream still fire that event? The concept of chat decoration is still very much a thing, even if previews aren't. Notice that our events are called "ChatDecorate" instead of "ChatPreview". If upstream isn't firing their event anymore during that process, I am fine removing our calling of it.

brickmonster:
Looks like they don't fire it. I checked out the base tag in Paper-Server and grep doesn't find any reference to the event.

brickmonster:
I'll update this pull request tonight/tomorrow with a proper commit.

By the way it looks like github built the change before there was a review, is that supposed to happen? If not then it might be because this is a draft.

Machine-Maker:
Yeah, I can't find any reference to it either in their source, so we can remove our calls to it.

patches/server/0010-Adventure.patch
```diff
@@ -2201,9 +2196,19 @@ index 887b9fd625aa23c4ec828a175d63695f915232d3..136ef0b360e29e736add1e6801831d24
 +
 +    record MessagePair(net.kyori.adventure.text.Component component, String legacyMessage) { }
 +
-+    record ModernResult(Component component, boolean hasNoFormatting, boolean modernized) implements Result {
-+        public ModernResult(net.kyori.adventure.text.Component component, boolean hasNoFormatting, boolean modernized) {
-+            this(io.papermc.paper.adventure.PaperAdventure.asVanilla(component), hasNoFormatting, modernized);
++    record ModernResult(Component component) implements Result {
```
brickmonster, line 2199: `ModernResult` and `Result` can probably be removed with a larger refactor

brickmonster:
All of the changes in ChatDecorator.java (ModernResult and Result) can probably be removed with a larger refactor.

Not sure how to rebase on the latest changes, it's very big.

Machine-Maker:
> All of the changes in ChatDecorator.java (ModernResult and Result) can probably be removed with a larger refactor. "Inline constant arguments" is a move in that direction but can be ignored if you want.

Yeah, I think we should go all the way here, not partially remove stuff

brickmonster:
This is the merged commit, you can see the individual refactoring steps [here](https://github.com/brickmonster/Paper-remove-LegacyFormat/tree/unmerged).

patches/server/0010-Adventure.patch
```diff
@@ -1213,6 +1039,72 @@ index 0000000000000000000000000000000000000000..d496a9a6ad229e42f1c44e31eafa6974
...
++    @Override
++    public CompletableFuture<Component> decorate(@Nullable ServerPlayer sender, @Nullable CommandSourceStack commandSourceStack, Component message) {
++        return decorate(server, sender, commandSourceStack, message);
++    }
++
++    private static CompletableFuture<Component> decorate(final MinecraftServer server, final @Nullable ServerPlayer player, final @Nullable CommandSourceStack commandSourceStack, final Component originalMessage) {
++        return CompletableFuture.supplyAsync(() -> {
++            // Make a copy so that if the event if cancelled we can guarantee the original is being returned.
++            final net.kyori.adventure.text.Component originalMessageCopy = PaperAdventure.asAdventure(originalMessage.copy());
++            final net.kyori.adventure.text.Component initialResult = PaperAdventure.asAdventure(originalMessage.copy());
```
Machine-Maker: These two calls to `.copy()` are not needed. Calling `PaperAdventure#asAdventure` already copies it. In addition, adventure's components are immutable, so you don't need to create 2 of these, just pass the same instance in for both params of the event.

patches/server/0010-Adventure.patch
```diff
@@ -1213,6 +1039,72 @@ index 0000000000000000000000000000000000000000..d496a9a6ad229e42f1c44e31eafa6974
...
++                event = new AsyncChatDecorateEvent(true, craftPlayer, originalMessageCopy, initialResult);
++            }
++
++            server.server.getPluginManager().callEvent(event);
++
++            if (!event.isCancelled() && !event.result().equals(PaperAdventure.asAdventure(originalMessage))) {
++                return PaperAdventure.asVanilla(event.result());
++            }
++
++            return originalMessage;
```
Machine-Maker: I think this can be simplified to be
If the event is cancelled, return `originalMessage`, otherwise, return `PaperAdventure.asVanilla(event.result())`. No need for this extra equals check.

patches/server/0010-Adventure.patch
```diff
@@ -2452,25 +2292,25 @@ index e2def0ca552343143e495736d533b3334686fd62..c87b708c368713a23a10ad97704575ee
...
-+    public PlayerChatMessage withResult(net.minecraft.network.chat.ChatDecorator.Result result) {
-+        final PlayerChatMessage msg = this.withUnsignedContent(result.component());
-+        return new PlayerChatMessage(msg.link, msg.signature, msg.signedBody, msg.unsignedContent, msg.filterMask, result);
++    public PlayerChatMessage withResult(Component result) {
++        final PlayerChatMessage msg = this.withUnsignedContent(result);
++        return new PlayerChatMessage(msg.link, msg.signature, msg.signedBody, msg.unsignedContent, msg.filterMask, io.papermc.paper.adventure.PaperAdventure.asAdventure(result));
 +    }
-+    public net.minecraft.network.chat.ChatDecorator.Result requireResult() {
++    public net.kyori.adventure.text.Component requireResult() {
 +        return Objects.requireNonNull(this.result, "Requires a decoration result to be set here");
```
Machine-Maker: I think we can get rid of the extra parameter all together. The result of the decoration should just be set as the unsigned content of the PlayerChatMessage like vanilla does it.

patches/server/0010-Adventure.patch
```diff
@@ -846,17 +693,13 @@ index 0000000000000000000000000000000000000000..57081e84d50f7eca7ea11540b550da82
 +         */
 +        this.message = message;
 +        this.async = async;
-+        if (this.message.requireResult().modernized()) {
-+            this.craftbukkit$originalMessage = this.message.requireResult().message().legacyMessage();
-+        } else {
-+            this.craftbukkit$originalMessage = message.signedContent();
-+        }
++        this.craftbukkit$originalMessage = message.signedContent();
```
Machine-Maker: It might not be appropriate to have this anymore? And just pass the legacy serialized result from ChatDecorator (or the plain string if there isn't one).

Machine-Maker (changes requested):

patches/server/0010-Adventure.patch
brickmonster: Leaving that out of this pull request as discussed.

Machine-Maker (approved):

lynxplay (approved):

PR #10413, opened 2024-04-12
Change online mode default for Velocity configuration

gecko10000:
Changing this default will not affect anything unless `proxies.velocity.enabled` is also true, and discouraging piracy is always nice.


patches/server/1055-Change-online-mode-default-for-Velocity-configuratio.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+index 0cc2307636279915c1f8529e62174cc696e185ee..30fe1c0645a07d663b08c0f988a1ab3a750bf7c4 100644
+--- a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
+@@ -127,7 +127,7 @@ public class GlobalConfiguration extends ConfigurationPart {
+ 
+         public class Velocity extends ConfigurationPart {
+             public boolean enabled = false;
+-            public boolean onlineMode = false;
++            public boolean onlineMode = true;
```
electronicboy: This should modify the actual patch in which this was added

electronicboy (approved):

lynxplay (approved):

PR #10415, opened 2024-04-12
[ci skip] Update book page/char limit for book meta doc

Lulu13022002:
The old bungee component methods still use the old limit but i'm not sure if it's worth to fix them here.


Machine-Maker (approved):

lynxplay (approved):

electronicboy:
Page limits? Where we're going, we don't need *page limits*

PR #10416, opened 2024-04-12
Include DamageSource in death event

Machine-Maker:


lynxplay (approved):

lynxplay:
Implemented upstream.

PR #10423, opened 2024-04-16
"Downgrade" Vineflower to 1.10.1 release

jpenilla:
The main thing we lose from the 1.11 snapshots is a recent fix for redundant casts.


patches/server/1005-Optimize-Pathfinder-Remove-Streams-Optimized-collect.patch
```diff
@@ -56,19 +56,20 @@ index 3e27c9a74c38f11a436df8cd413f9f5f16fd8c3e..c4d6b145d54aaf415dd4b08811c02425
...
              node.closed = true;
  
 -            for (Target target : set) {
+-                if (node.distanceManhattan(target) <= (float)distance) {
 +            // Paper start - optimize collection
 +            for (int i1 = 0; i1 < positions.size(); i1++) {
 +                final Map.Entry<Target, BlockPos> entry = positions.get(i1);
 +                Target target = entry.getKey();
-                 if (node.distanceManhattan(target) <= distance) {
++                if (node.distanceManhattan(target) <= distance) {
```
kennytv: Could readd the cast to keep the original line unmodified, but probably doesn't matter/would stay like this in a later update anyways

kennytv (approved):

PR #10436, opened 2024-04-19
Fix inventory desync with PlayerLeashEntityEvent

### LINKED ISSUE #10432: The lead turns into a ghost item when cancelling PlayerLeashEntityEvent
### Expected behavior

That the lead do not turn into a ghost item.

### Observed/Actual behavior

That the item doesn't turn into a ghost item and having to update player inventory after cancelling the event in order for the item to not render as a ghost item.

### Steps/models to reproduce

Simply add this listener and just remove the if() clause. 

```
    @EventHandler
    public void onLeash(PlayerLeashEntityEvent event) {
        ItemStack itemStack = player.getInventory().getItemInMainHand();
        if(itemStack.hasItemMeta() && hookHandler.isHook(itemStack)) {
            event.setCancelled(true);
            //player.updateInventory();
        }
    }
```

### Plugin and Datapack List

Codex, MobArmory, Skyline

### Paper version

[00:03:02 INFO]: Checking version, please wait...
[00:03:02 INFO]: This server is running Paper version git-Paper-451 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: ab1afb0)
You are 34 version(s) behind
Previous version: git-Paper-485 (MC: 1.20.4)

### Other

_No response_

notTamion:
fixes #10432

### DIFF
--- a/patches/server/0926-Fix-inventory-desync.patch
+++ b/patches/server/0926-Fix-inventory-desync.patch
@@ -4,6 +4,18 @@ Date: Wed, 23 Aug 2023 13:22:09 -0700
 Subject: [PATCH] Fix inventory desync
 
 
+diff --git a/src/main/java/net/minecraft/world/entity/Mob.java b/src/main/java/net/minecraft/world/entity/Mob.java
+index d1833dcc92b1570e62a1941bbea6d66ff0b2ee5f..5da7ffee71bf6d82a53b3c2f99e2bd898d0b1799 100644
+--- a/src/main/java/net/minecraft/world/entity/Mob.java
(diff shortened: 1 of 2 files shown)


kennytv (approved):
Might be nice to only send what actually needs sending, but otherwise

Machine-Maker:
> Might be nice to only send what actually needs sending, but otherwise

I have always advocated against this, due to client-side mods that deal with inventory management. Imagine a scenario where a hotbar slot is auto filled if a stack runs out, only sending that specific slot wouldn't update the rest of the inventory.

Machine-Maker (approved):

PR #10440, opened 2024-04-20
Add Configuration for vertical Despawn Ranges

### LINKED ISSUE #10177: Add a separate Y-value check for "despawn-ranges"

When the number of players reaches dozens of people, the high simulation distance cannot be afforded. My simulation distance is 5.
And then Monsters outside of simulation distance do not disappear randomly.
Peper has an option for this<Despawn Range (hard)>
Hopefully it would be nice to separate the Y-value checks separately.
For example, monsters can be configured to disappear outside of 128 from the player's Y value.The X and Z values disappear beyond 80.
(It is guaranteed that there is no need to modify the afk spots of the biological farm)
image

This request is a bit like "tracking-range-y".
More granular configuration options will help a lot of people.
Hopefully someone will be able to create a PR for this.

Add a separate Y-value check for "despawn-ranges"
despawn-ranges-y

https://hangar.papermc.io/Kyle/harderdespawn
I'm using this plugin right now.but only when the chunk is unloaded, the mob disappears. It helped me somewhat, but not much(It is even abused by players, and when the player encounters a monster fight, he quits the game and rejoins, and the monsters are all gone).
It would be nice if "despawn-ranges" could be configured with a Y value separately.
"despawn-ranges-y"

### Other

https://github.com/PaperMC/Paper/issues/9476
Thanks to electronicboy suggestion, I closed the previous Feature Request.
The previous Feature Request was a bit weird

notTamion:
resolves #10177 
config naming/structure might need some changes, i didn't wanna break existing configs so i just did it like this.

currently doesn't work for Raiders as those have extra checking logic in the removeWhenFarAway method, so i am not quite sure what to do about those

EDIT: also wanted to add that i am not sure if the normal despawn range should still impact the y axis. since your x axis is going to be higher then your y axis 99 % of the time anyways i will just have it keep affecting that aswell

Docs-pr at: PaperMC/docs#465

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -2466,10 +2466,10 @@ index 0000000000000000000000000000000000000000..9c339ef178ebc3b0251095f320e4a7a3
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/serializer/DespawnRangeSerializer.java b/src/main/java/io/papermc/paper/configuration/serializer/DespawnRangeSerializer.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..04909c5b5d4b70538ad2bc8b29c6471acd4310f7
+index 0000000000000000000000000000000000000000..0e3a7e28f79bc29ee051463d32ea25b864ca6b81
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/serializer/DespawnRangeSerializer.java
-@@ -0,0 +1,77 @@
+@@ -0,0 +1,78 @@
 +package io.papermc.paper.configuration.serializer;
 +
 +import io.papermc.paper.configuration.WorldConfiguration.Entities.Spawning.DespawnRange;
@@ -2540,10 +2540,11 @@ index 0000000000000000000000000000000000000000..04909c5b5d4b70538ad2bc8b29c6471a
 +    }
 +
 +    private int ifPresent(ConfigurationNode node, int or) {
-+        if (node.virtual())
++        if (node.virtual()) {
 +            return or;
-+        if (node.raw() instanceof Integer i)
++        } else if (node.raw() instanceof Integer i) {
 +            return i;
++        }
 +        throw new RuntimeException(new SerializationException(node, Integer.class, "Expected an integer, got \"" + node.raw() + "\" instead"));
 +    }
 +}
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -854,10 +854,10 @@ index 0000000000000000000000000000000000000000..69add4a7f1147015806bc9b63a8340d1
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/PaperConfigurations.java b/src/main/java/io/papermc/paper/configuration/PaperConfigurations.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..783eac6e458c6f1a0584301fb84a2fe341868f34
+index 0000000000000000000000000000000000000000..1cee7ba88f8cd03009d9a3a01c1128eff598833c
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/PaperConfigurations.java
-@@ -0,0 +1,466 @@
+@@ -0,0 +1,463 @@
 +package io.papermc.paper.configuration;
 +
 +import com.google.common.base.Suppliers;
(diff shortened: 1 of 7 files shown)


Machine-Maker:
I'm not sold on the format for the config with this change. There are several options that I can think of that I'd prefer
```yaml
despawn-ranges:
    monster:
        hard:
            xz: 30 # or "horizontal"
            y: 5 # or "vertical"
        soft: 20 # just putting a numerical value would mean it applies to both (as it does now)
```
or
```yaml
despawn-ranges:
    monster:
        hard: 30
        soft: 20
        soft-y: 5 # basically the `-y` value wouldn't even be in the config if it wasn't being used
```

notTamion:
is there an existing way i can add optional configurations like that?

lynxplay:
As you already said, this does not currently "work", as the distanceSqrt still takes into account the y diff.
Given we already have changes to basically *every* line in this section, I think it would be fair to just completely rewrite this logic
in a new paper block without consideration of keeping old code around/minimizing diff (given that this isn't minimal diff either, every line is touched beyond the two discard calls).

This would also *heavily* improve the variable names, making this whole mess a lot easier to read.

notTamion:
alright i will probably be working on the custom serializer and the rewrite tomorrow

notTamion:
Implemented the requested changes

BangXi:
bump
Is there a chance to merge before 1.21 arrives

notTamion:
if someone could comment on whether or not the config layout is ok that would be great, then i can quickly writeup a docs pr so that's also ready.

new config layout (as proposed by machine):
```yaml
despawn-ranges:
    monster:
        hard:
            horizontal: 30
            vertical: 5
        soft: 20 # just putting a numerical value would mean it applies to both (as it does now)
```

BangXi:
> if someone could comment on whether or not the config layout is ok that would be great, then i can quickly writeup a docs pr so that's also ready.
> 
> new config layout (as proposed by machine):
> 
> ```yaml
> despawn-ranges:
>     monster:
>         hard:
>             horizontal: 30
>             vertical: 5
>         soft: 20 # just putting a numerical value would mean it applies to both (as it does now)
> ```

I think this config format layout is great
Clear and straightforward

CatBang:
Sorry, another bump, the stable version 1.21 has been released for some time now, can this feature be merged now? I really need it.

patches/server/0005-Paper-config-files.patch
```diff
@@ -2447,6 +2450,90 @@ index 0000000000000000000000000000000000000000..9c339ef178ebc3b0251095f320e4a7a3
...
++        if (xzSoftLimit == ySoftLimit) {
++            soft.set(xzSoftLimit);
++        } else {
++            soft.node("horizontal").set(xzSoftLimit);
++            soft.node("vertical").set(ySoftLimit);
++        }
++    }
++
++    private int ifPresent(ConfigurationNode node, int or) {
++        if (node.virtual())
```
kennytv: braces

patches/server/0032-Add-configurable-entity-despawn-distances.patch
```diff
@@ -5,23 +5,44 @@ Subject: [PATCH] Add configurable entity despawn distances
...
++                // Paper start - Configurable despawn distances
++                io.papermc.paper.configuration.WorldConfiguration.Entities.Spawning.DespawnRange despawnRanges = this.level().paperConfig().entities.spawning.despawnRanges.get(this.getType().getCategory());
  
-                 if (d0 > (double) j && this.removeWhenFarAway(d0)) {
-                     this.discard(EntityRemoveEvent.Cause.DESPAWN); // CraftBukkit - add Bukkit remove cause
-                 }
+-                if (d0 > (double) j && this.removeWhenFarAway(d0)) {
+-                    this.discard(EntityRemoveEvent.Cause.DESPAWN); // CraftBukkit - add Bukkit remove cause
+-                }
++                double xzDistance = entityhuman.distanceToSqr(this.getX(), entityhuman.getY(), this.getZ());
```
kennytv: Add a squared suffix where applicable here and below, else the variables can be confused easily

kennytv (approved):

lynxplay (dismissed):
LGTM, however we will need to come up with a docs PR that correctly represents this more dynamic syntax.

lynxplay:
Brought this up internally, but I am less and less convinced that this "dynamic" syntax is that nice for us.
Configs would now potentially change outside of a version migration because someone defined 
```
hard:
  xy: 10
  y: 10
```

Would obviously be interested in other peoples input here, but a single migration and simply always using the more verbose layout would imo work just as fine.

notTamion:
Would configs changing node structure have any kind of impact though? configurate still only sees that Map<MobType, DespawnRange> and all (de)serialization regarding the underlying nodes is done by the serializer. Do you have any example of how this might have an bad impact?

lynxplay:
I mean, mostly that the file changes without a migration notice in the _version field.

Now that I look at it, we'd also instead want to make it use `default` instead of the current versions magic values.
And current layout brings a couple of annoying side effects.
When you only define `hard: default`, the config makes sense. All ranges for hard despawning are defaulted.
If you insert
```yaml
hard:
  horizontal: 10
```
then vertical is *implicitly* set to default, even without it being present in the config.
So now we are introducing 
  a) paper changing the config file without the need to do so, potentially breaking/messing with server state managers (at least mine lol)
  b) it computes an *implicit* value.

Beyond the fact that we need to change this to using "default" values, need to drop the `ConfigSerializable` annotation and also need to register the deserializer in the first place, I think it may be worth to (at least) look into preserving the config structure if people defined the same value for horizontal/vertical.

notTamion:
I can definitely change it so it doesn't doesn't create a node with a default value if it hasn't been defined in the config and change it so it uses default instead of the normal numbers. _version is there to make sure configs aren't loaded on incompatible versions (which i just realized i also have to up considering old servers won't play nice with this) not to track config history. I will just wait on this pr for some time and allow people to give their thoughts on it.

lynxplay:
Yea probably best, I'll through it at my list to discuss with machine whenever we next sit down.
But yea, if you could register the serialiszer and yank the config seria annotation, that would be great. Just so that no one forgets that xD

notTamion:
Already did that yesterday, see the latest commit.

lynxplay:
Ah sweet, I must have pulled an out of date version then when I was looking over it today, sorry!

lynxplay:
Going further this PR should

a) emit xz, y fields if specified in config, even if their values match. For this DespawnRange needs to track if the val was defined in long or short syntax. (Preferably new type that is Range(xz, y, longSyntax)

b) require defining vertical if horizonal was defined. No "partial" long syntax.

PR #10448, opened 2024-04-23
Adjust large packet handler to run when above protocol limit

Spottedleaf:
Previously, PacketEncoder assumed that a packet value larger
than the protocol limit would be compressed to become smaller
than the protocol limit. However, not all packets will compress
below the protocol limit.

To try to better handle this, we will run the large packet handler
when the packet size is above the protocol limit when the packet
has a large packet fallback to avoid a case where the packet
does not compress below protocol limit (at which point, it is
too late to run the large packet handler).


kennytv (approved):

electronicboy (approved):

Spottedleaf:
https://github.com/PaperMC/Paper/compare/f4c7d373e4a1...7ac24a18940d

PR #10454, opened 2024-04-25
readd itemstack data sanitization

Machine-Maker:
Just an initial idea for how to do the itemstack sanitization with the new stream codec system. Needs to be tested once compile issues are fixed.


Machine-Maker:
Rebased and tested, and yes, it does sanitize the data component values correctly where needed. If this format of doing this is O.K., we can add the rest of the missing sanitization functionality.

lynxplay (approved):

Issue #10354, opened 2024-03-24
Certain item-validation config options don't seem to work.


Joko991:
### Expected behavior

Changing the maximum length of the page, and max number of bytes in a page to -1 should allow for the player to use these books without these limits.

### Observed/Actual behavior

The "Book too large!" kick seems to activate no matter what the config options are set to, page-max seems to work with all positive numbers up to 2147483647 so it is set to that, but the maximum length of page seems to be unchanged no matter what is in the config.

### Steps/models to reproduce

1.Have a paper server with default options.
2. Set "page" and "page-max" under item-validationto 2147483647, -1, or some other very large number.
3. If page-max is set to -1, books are disabled. No matter what "page" is set to the kick always occurs at around the same length.

### Plugin and Datapack List

Plugins:
Bukkit Plugins:
[22:13:17 INFO]:  
- 
CoreProtect
, 
floodgate
, 
Geyser-Spigot
, 
SkinsRestorer
, 
ViaVersion
Datapacks:
[22:13:22 INFO]: There are 2 data pack(s) enabled: 
[
vanilla (built-in)
]
, 
[
file/bukkit (world)
]
[22:13:22 INFO]: There are no more data packs available

### Paper version

[22:13:25 INFO]: 
Checking version, please wait...
[22:13:25 INFO]: 
This server is running Paper version git-Paper-461 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: b600140)
You are running the latest version
Previous version: git-Paper-459 (MC: 1.20.4)

### Other

Page-max has been set to 2147483647 on my server, "page" has been tested on 2147483647, 16777216, default configuration, and -1, all of them provided the same result from what I saw.
total-multiplier is set to 1.0, so that should not be an issue.
The player is not an operator, and is on Java edition.

electronicboy:
This is basically a feature request, not a bug report

Joko991:
Ah, I thought this would be considered a bug report. From what I saw in console the server doesn't start and gives you an error if any of the values are too high/low, so I thought the values I have currently would be fine. Would it be possible for if it is set to a negative number or just -1, for the check to be disabled?

aHalliday13:
Has there been any updates on this? I run a small paper server, and it's infuriating to get kicked making books sometimes as small as six pages, and not being able to modify the configuration in any meaningful way to fix it. I agree with joko that there should be some way to just disable the check.

PR #10457, opened 2024-04-26
Allow server administrators to disable book size checks (on the right branch this time)

aHalliday13:
Re-creating this pull request because it was on the wrong branch last time and it was automatically closed. I run a small Minecraft server, and we've been running into issues with the book size check. After struggling to find a suitable configuration, I've made this patch, which allows an admin to disable the book size check by setting page-max in paper-global.yml to -1. This should also fix issue https://github.com/PaperMC/Paper/issues/10354.


Leguan16:
Should be merged into [patches/server/Book-Size-Limits.patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0276-Book-Size-Limits.patch). Although i am not sure if this setting will be available after the update to 1.20.6 since minecraft also removes it.

aHalliday13:
Looks like that patch you linked no longer exists? I'll make the necessary changes to this PR, just want to get clarification on what changes to make. Also what exactly are you referring to for 1.20.6? Thanks!

Leguan16:
It still exists. Just search for Book-size-limit.patch in patches/server. The number in front can change.

kennytv (approved):

NoahvdAa (approved):

PR #10462, opened 2024-04-27
Preliminary API for 'new' brains AI system for mobs

### LINKED ISSUE #10443: API for controlling new mob behaviours

Probably related to #10442.
Related to #6862.

As Minecraft moves into the future, more and more of the new mobs are mostly controlled via the new systems instead of the old ones that the current Bukkit Goals API abstracts over, and the inability to control them from the Paper API is more glaring.

A new abstraction (or perhaps retrofitting the existing one) for controlling mob behaviours

Using NMS in order to prod at the underlying Minecraft stuff myself. Whilst doable, NMS usage makes cross-version support an individual plugin issue instead of a communal solution in Paper

### Other

I guess it would be nice to have some sort of indicator in the existing API docs for what they don't support currently, maybe a `UnsupportedOperationException` to make failures easier to catch instead of silently not working/only returning `false`

sowelipililimute:
Would fix #10443 if done.

This is a rough proposal for abstracting over the brains AI system for mobs for plugins to make use of, similar to the Goals API. I don't think it's feasible to attempt to make the new AI conform to the old interfaces.

At the moment I understand the rough skeleton and arteries of the system, though I'm still trying to piece together what does what in user-facing terms. 'get the mob to walk somewhere' has been unsuccessful so far.

As far as API stability concerns, as long as the exposed set of tasks isn't too 1:1 with what Mojang has, then we should be fine (until the next Big Rewrite:tm: whenever that is) considering the general stability of the system across versions since its introductions.

There are still some unanswered questions in the API design:
- do events make sense to have here? I'd say no considering the amount of lambdas in the Minecraft implementation that make any real introspection API unfeasible.
- do we care to expose sensors/make them modifiable? It could probably be done in a followup PR to make custom tasks & sensors a thing, as I'm mostly concerned about exposing what exists in Vanilla at the moment.
- is the arrangement of the entrypoints to the new API sensical?

Test plugin is included in PR to ease testing of the new API until it's ready to ship.

### DIFF
--- a/paper-api-generator/generated/io/papermc/paper/registry/keys/ActivityKeys.java
+++ b/paper-api-generator/generated/io/papermc/paper/registry/keys/ActivityKeys.java
@@ -1,0 +1,217 @@
+package io.papermc.paper.registry.keys;
+
+import static net.kyori.adventure.key.Key.key;
+
+import io.papermc.paper.entity.ai.Activity;
+import io.papermc.paper.generated.GeneratedFrom;
+import io.papermc.paper.registry.RegistryKey;
+import io.papermc.paper.registry.TypedKey;
+import net.kyori.adventure.key.Key;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Vanilla keys for {@link RegistryKey#ACTIVITY}.
+ *
+ * @apiNote The fields provided here are a direct representation of
+ * what is available from the vanilla game source. They may be
+ * changed (including removals) on any Minecraft version
+ * bump, so cross-version compatibility is not provided on the
+ * same level as it is on most of the other API.
+ */
+@SuppressWarnings({
+        "unused",
+        "SpellCheckingInspection"
+})
+@GeneratedFrom("1.21.1")
+@ApiStatus.Experimental
+public final class ActivityKeys {
+    /**
+     * {@code minecraft:admire_item}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> ADMIRE_ITEM = create(key("admire_item"));
+
+    /**
+     * {@code minecraft:avoid}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> AVOID = create(key("avoid"));
+
+    /**
+     * {@code minecraft:celebrate}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> CELEBRATE = create(key("celebrate"));
+
+    /**
+     * {@code minecraft:core}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> CORE = create(key("core"));
+
+    /**
+     * {@code minecraft:dig}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> DIG = create(key("dig"));
+
+    /**
+     * {@code minecraft:emerge}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> EMERGE = create(key("emerge"));
+
+    /**
+     * {@code minecraft:fight}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> FIGHT = create(key("fight"));
+
+    /**
+     * {@code minecraft:hide}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> HIDE = create(key("hide"));
+
+    /**
+     * {@code minecraft:idle}
+     *
+     * @apiNote This field is version-dependant and may be removed in future Minecraft versions
+     */
+    public static final TypedKey<Activity> IDLE = create(key("idle"));
+
+    /**
+     * {@code minecraft:investigate}
+     *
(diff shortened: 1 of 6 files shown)


patches/server/1060-Implement-preliminary-API-for-working-with-new-mob-A.patch
```diff
@@ -0,0 +1,861 @@
...
++    private final Map<Integer, Map<Activity, Set<BehaviorControl<? super E>>>> availableBehaviorsByPriority = Maps.newTreeMap();
++    private Schedule schedule = Schedule.EMPTY;
++    private final Map<Activity, Set<Pair<MemoryModuleType<?>, MemoryStatus>>> activityRequirements = Maps.newHashMap();
++    private final Map<Activity, Set<MemoryModuleType<?>>> activityMemoriesToEraseWhenStopped = Maps.newHashMap();
++    private Set<Activity> coreActivities = Sets.newHashSet();
++    private final Set<Activity> activeActivities = Sets.newHashSet();
++    private Activity defaultActivity = Activity.IDLE;
++    private long lastScheduleUpdate = -9999L;
++
++    public static <E extends LivingEntity> Brain.Provider<E> provider(
```
yannicklamprecht: Import it how you should. See contribution.md on how to handle NMS changes.

patches/server/1060-Implement-preliminary-API-for-working-with-new-mob-A.patch
```diff
@@ -0,0 +1,861 @@
...
++++ b/src/main/java/io/papermc/paper/entity/ai/PaperActivities.java
+@@ -0,0 +1,111 @@
++package io.papermc.paper.entity.ai;
++
++import org.jetbrains.annotations.NotNull;
++import net.minecraft.world.entity.schedule.Activity;
++
++public class PaperActivities implements io.papermc.paper.entity.ai.Activities {
++    @Override
++    public @NotNull io.papermc.paper.entity.ai.Activity getAdmireItem() {
```
yannicklamprecht: I'm totally against exposing that via method that way.
I personally prefer a key based approach, where you can use a paper api-generator module to generate these keys automatically. You can even generate the mapping for the keys automatically.


```suggestion
+    public @NotNull io.papermc.paper.entity.ai.Activity get(PaperActivityType activity) {
```

Better naming should be suggested.

patches/api/0474-Add-preliminary-API-for-working-with-new-mob-AI-syst.patch
```diff
@@ -0,0 +1,175 @@
...
+index 0000000000000000000000000000000000000000..d03d81256a8ba9a07e1a6d3990f1d3654ea608d2
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/ai/Activities.java
+@@ -0,0 +1,32 @@
++package io.papermc.paper.entity.ai;
++
++import org.jetbrains.annotations.NotNull;
++
++public interface Activities {
++    public @NotNull Activity getAdmireItem();
```
yannicklamprecht: See below, use api-generator and key based approach.

yannicklamprecht (changes requested):

Owen1212055:
Feel free to reference https://github.com/PaperMC/Paper/pull/6968

This should support creating own sensors/memories via the bootstrap api.

patches/api/0474-Add-preliminary-API-for-working-with-new-mob-AI-syst.patch
sowelipililimute: ... how do i use api-generator? i don't see any documentation in the repo

sowelipililimute: fixed

patches/server/1060-Implement-preliminary-API-for-working-with-new-mob-A.patch
sowelipililimute: done

patches/server/1060-Implement-preliminary-API-for-working-with-new-mob-A.patch
sowelipililimute: removed

pontaoski:
how does the way tasks are exposed look?

pontaoski:
need to figure out generating the SensorTypeKeys in a more type-safe manner

patches/api/0480-Add-preliminary-API-for-working-with-new-mob-AI-syst.patch
```diff
@@ -0,0 +1,236 @@
...
++/**
++ * A brain is the central AI for some modern Minecraft entities such as
++ * villagers and sniffers.
++ */
++public interface Brain<E extends LivingEntity> {
++    /**
++     * If the tasks in the given `activity` can be run, the brain will switch to
++     * the given `activity`.
++     */
++    public void useActivityIfPossible(@NotNull Activity activity);
```
kashike: This applies to all additions - interfaces do not need methods to specify `public`.

```suggestion
+    void useActivityIfPossible(@NotNull Activity activity);
```

kashike (changes requested):

patches/api/0480-Add-preliminary-API-for-working-with-new-mob-AI-syst.patch
sowelipililimute: fixed

pontaoski:
can someone explain the CI failure to me I don't really get it and also I can't reproduce it locally

electronicboy:
> GeneratedFrom("1.20.6")

yannicklamprecht:
> can someone explain the CI failure to me I don't really get it and also I can't reproduce it locally

As @electronicboy said in the short form. You need to regenerate the classes with the updated version.
So just remove the comment in paper-api-generator project, so that it will be recognized as a project.
Then run the main method in that project to regenerate the classes. It'll automatically update the code and of course the annotation.

PR #10478, opened 2024-04-28
disguise api

yannicklamprecht:
relates to https://github.com/PaperMC/Paper/discussions/10584

Adds a simple disguise api that also makes sure that invalid entity data, not belonging to the type that is faked, is filtered.
It also adds a server generator module to reduce maintenance of the entity data filter.

tldr; The API provides a way to disguise any entity as another or a player with having the possibility to interact with the entity like with any others. The big advantage is that you've the whole vanilla logic and Mob Goal and Pathfinding included without hassle.

Testing the changes can be done using the following snippet:
https://gist.github.com/yannicklamprecht/639acba127a54f5617e1281f099b7aff

TODO:

- [x] **Add JavaDoc**
- [x] remove test plugin code
- [x] comment paper-server-generator
- [x] ~~figure out how to get teams working with real teams, client side ones worked in my fork~~ Non issue just use the `Entity#getName()` instead of the `Entity#getScoreboardEntryName()` 
- [x] remove player profile if player disguise on ServerEntiry#removePairing(player) to avoid client disconnects when moving away and approaching again. see https://github.com/PaperMC/Paper/pull/10478#issuecomment-2105299573
- [x] wait until dev/1.21 is merged into master
- [x] When player is disguised as entity his attributes are propagated to the client, which isn't supported by certain entities
- [x] When player is disguised as entity that has no ability to interact/build, the server should not propagate the players action to the rest of the players. Otherwise the other player gets kicked.
- [x] When changing disguise while in visible range from e.g. Boat to a Living entity disguise, attributes will be sent again. -> on disguise change => auto hide and show entity to players.

Effective loc:
- 600 loc in patches
- 3k+ loc generated filter classe
- 400 loc generator code copied from api-generator
- 100 loc for generating the filter class.


patches/server/1047-add-disguise-api.patch
```diff
@@ -0,0 +1,261 @@
...
+--- a/src/main/java/net/minecraft/server/level/ServerEntity.java
++++ b/src/main/java/net/minecraft/server/level/ServerEntity.java
+@@ -308,7 +308,11 @@ public class ServerEntity {
+         Packet<ClientGamePacketListener> packet = this.entity.getAddEntityPacket();
+ 
+         this.yHeadRotp = Mth.floor(this.entity.getYHeadRot() * 256.0F / 360.0F);
+-        sender.accept(packet);
++        // Paper start - disguise api
++        if(!io.papermc.paper.disguise.DisguiseUtil.tryDisguise(player, entity, (net.minecraft.network.protocol.game.ClientboundAddEntityPacket) packet)){
++            sender.accept(packet);
```
yannicklamprecht: unindent.

patches/api/0473-add-disguise-api.patch
```diff
@@ -0,0 +1,158 @@
...
+index 62e3793903905b94eb1a120345015149abb33713..d4e425bcab43bd129b5559dc7fb9fad87f617656 100644
+--- a/src/main/java/org/bukkit/entity/Entity.java
++++ b/src/main/java/org/bukkit/entity/Entity.java
+@@ -1155,4 +1155,8 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
+      */
+     @NotNull String getScoreboardEntryName();
+     // Paper end - entity scoreboard name
++    // Paper start - disguise api
++    @NotNull io.papermc.paper.disguise.DisguiseData getDisguiseData();
++    void setDisuiseData(@NotNull io.papermc.paper.disguise.DisguiseData disguiseData);
```
yannicklamprecht: Fix setter.
```suggestion
+    void setDisguiseData(@NotNull io.papermc.paper.disguise.DisguiseData disguiseData);
```

patches/api/0473-add-disguise-api.patch
```diff
@@ -0,0 +1,158 @@
...
+@@ -0,0 +1,25 @@
++package io.papermc.paper.disguise;
++
++import com.destroystokyo.paper.profile.PlayerProfile;
++import org.bukkit.entity.EntityType;
++import org.jetbrains.annotations.NotNull;
++
++public sealed interface DisguiseData permits DisguiseData.OriginalDisguise, EntityTypeDisguise, PlayerDisguise {
++    static @NotNull DisguiseData original() {
++        return new OriginalDisguise();
```
yannicklamprecht: Just delegate.
```suggestion
+        return reset();
```

paper-server-generator.settings.gradle.kts
```diff
@@ -0,0 +1,2 @@
+// Uncomment to enable the 'paper-server-generator' project
+ include(":paper-server-generator")
```
yannicklamprecht:
```suggestion
 // include(":paper-server-generator")
```

patches/server/1047-add-disguise-api.patch
```diff
@@ -0,0 +1,261 @@
...
++
++public final class DisguiseUtil {
++    private DisguiseUtil(){}
++
++    public static boolean tryDisguise(ServerPlayer player, Entity entity, ClientboundAddEntityPacket clientboundAddEntityPacket) {
++            return switch (entity.getBukkitEntity().getDisguiseData()) {
++                case DisguiseData.OriginalDisguise disguise -> false;
++                case io.papermc.paper.disguise.EntityTypeDisguise(var type) -> {
++                    player.connection.send(new net.minecraft.network.protocol.game.ClientboundAddEntityPacket(
++                        clientboundAddEntityPacket.getId(),
```
yannicklamprecht: Move creation to methods to avoid duplication.

patches/api/0475-add-disguise-api.patch
```diff
@@ -0,0 +1,158 @@
...
+--- a/src/main/java/com/destroystokyo/paper/SkinParts.java
++++ b/src/main/java/com/destroystokyo/paper/SkinParts.java
+@@ -17,4 +17,23 @@ public interface SkinParts {
+     boolean hasHatsEnabled();
+ 
+     int getRaw();
++
++    interface Builder {
++        @org.jetbrains.annotations.NotNull
++        Builder withCape();
```
kashike: Should be formatted like so:
```suggestion
+        @org.jetbrains.annotations.NotNull Builder withCape();
```

patches/api/0475-add-disguise-api.patch
```diff
@@ -0,0 +1,158 @@
...
++        @org.jetbrains.annotations.NotNull
++        Builder withJacket();
```
kashike: These should all accept a `boolean` argument.

kashike (changes requested):

patches/server/1047-add-disguise-api.patch
```diff
@@ -0,0 +1,274 @@
...
++            return switch (entity.getBukkitEntity().getDisguiseData()) {
++                case DisguiseData.OriginalDisguise disguise -> false;
++                case io.papermc.paper.disguise.EntityTypeDisguise(var type) -> {
++                    player.connection.send(create(clientboundAddEntityPacket, org.bukkit.craftbukkit.entity.CraftEntityType.bukkitToMinecraft(type)));
++                    yield  true;
++                }
++                case io.papermc.paper.disguise.PlayerDisguise(var playerProfile, var listed, var skinParts) -> {
++                    PlayerProfile adapted = Bukkit.createProfile(entity.getUUID(), entity.hasCustomName()? entity.getBukkitEntity().getCustomName(): playerProfile.getName());
++                    adapted.setProperties(playerProfile.getProperties());
++                    net.minecraft.network.protocol.game.ClientboundPlayerInfoUpdatePacket.Entry playerUpdate =
```
MiniDigger: some of the classes here are imported and some use the fully qualified name (even if they are imported)
just use imports for everything, its a class we add so no risk of dum conflicts

patches/server/1047-add-disguise-api.patch
```diff
@@ -0,0 +1,274 @@
...
+                     datawatcher_item.setDirty(false);
++                    if (io.papermc.paper.disguise.DisguiseUtil.shouldSkip((net.minecraft.world.entity.Entity) entity, datawatcher_item.getAccessor())) continue; // Paper - disguise api
+                     list.add(datawatcher_item.value());
+                 }
+             }
+@@ -116,7 +117,7 @@ public class SynchedEntityData {
+ 
+         for (int j = 0; j < i; ++j) {
+             SynchedEntityData.DataItem<?> datawatcher_item = adatawatcher_item[j];
+-
```
MiniDigger: I would keep the empty line here and below to have nicer looking diff

patches/server/1047-add-disguise-api.patch
```diff
@@ -0,0 +1,274 @@
...
++                            clientboundAddEntityPacket.getId(),
++                            List.of(new SynchedEntityData.DataItem<>(Player.DATA_PLAYER_MODE_CUSTOMISATION, (byte) skinParts.getRaw()).value())
++                        ));
++                    }
++                    yield  true;
++                }
++            };
++    }
++
++    private static ClientboundAddEntityPacket create(ClientboundAddEntityPacket clientboundAddEntityPacket, EntityType<?> entityType) {
```
MiniDigger: MAJOR nit pick but if you name the arugment packet the method looks a whole lot nicer, lol

yannicklamprecht:
Tested with:
https://gist.github.com/yannicklamprecht/639acba127a54f5617e1281f099b7aff

Only open part is the discussion for a common generator module

yannicklamprecht:
We should probably move the common generator classes into a separate module to reduce duplications.

MiniDigger:
this is complicated enough that we prolly want a lil docs page with an example, but that can wait till after its merged and we know the final api design

yannicklamprecht:
> this is complicated enough that we prolly want a lil docs page with an example, but that can wait till after its merged and we know the final api design

Thought of adding a Java 18+ snippet to the javadoc but a docs page will be better.

MiniDigger:
I meant the api itself

yannicklamprecht:
Addressed issue with experience orbs by filtering them out. Mentioned that exclusion in the JavaDoc.

Additionally found an issue with ElderGuardian when disguising them as a Player on the client side. That happens only when the ElderGuardian is in the water. First thought is the ElderGuardian stream sucking health from other entities. To be sure I need to deobf the client stacktrace.

```java
Failed to handle packet adm@395c0068
java.lang.ClassCastException: class gct cannot be cast to class ckb (gct and ckb are in unnamed module of loader 'app')
	at fxy.a(SourceFile:1068)
	at adm.a(SourceFile:41)
	at adm.a(SourceFile:13)
	at zz.a(SourceFile:25)
	at bpm.d(SourceFile:162)
	at bpq.d(SourceFile:23)
	at bpm.A(SourceFile:136)
	at bpm.bA(SourceFile:121)
	at ffh.c(SourceFile:1275)
	at ffh.f(SourceFile:888)
	at net.minecraft.client.main.Main.main(SourceFile:265)
```

kennytv:
That's ClientPacketListener handleEntityEvent

yannicklamprecht:
> That's ClientPacketListener handleEntityEvent, Player can't be cast to Guardian

Found the reason why. I basically moved away from the entity, the chunk or at least the entities got unloaded.
Due to the state that the disguise data is not persisted on the entity, when loading the entity again it published itself to the client without the disguise info. The client knows the entity uuid only as the disguised type but gets the original one.

I updated the test plugin snippet to reflect the needed change - cache cleaning on death excluded.

It would be way better to persist the disguise data in NBT to avoid manual caching per plugin.

yannicklamprecht:
Addressed the last known issue by removing the profile info for a player disguised entity on ServerEntity#removePairing(player).
https://github.com/PaperMC/Paper/pull/10478/files#diff-1b00ff7ef0678fae6f893a77a2ed93bece7dbbd375a03115590b25304423ab1fR161-R170

https://github.com/PaperMC/Paper/pull/10478/files#diff-1b00ff7ef0678fae6f893a77a2ed93bece7dbbd375a03115590b25304423ab1fR252

yannicklamprecht:
Update to 1.21

XxGoldenbluexX:
good job!

patches/server/1057-add-disguise-api.patch
```diff
@@ -0,0 +1,360 @@
...
++    /*
++     * Is used to skip entity meta that doesn't fit the disguised type.
++     * e.g. Player having a float at index 15 (additional hearts) and the server side entity is an Armorstand
++     * that has a byte at that index.
++     */
++    public static boolean shouldSkip(Entity entity, EntityDataAccessor<?> dataAccessor) {
++        return switch (entity.getBukkitEntity().getDisguiseData()) {
++            case DisguiseData.OriginalDisguise original -> false;
++            case EntityTypeDisguise entityTypeDisguise -> !io.papermc.paper.entity.meta.EntityMetaWatcher.isValidForClass(
++                CraftEntityType.bukkitToMinecraft(entityTypeDisguise.entityType()).getBaseClass(),
```
Lulu13022002: This will always return Entity.class, so the generated content is almost useless and all the metadata are skipped.
Generally i don't think the new module is really needed and this can be done with regular reflection like https://pastes.dev/FNvwYGnwgS

yannicklamprecht: I'm against using reflection for that. It can hide potential bugs occuring during runtime. When generating everything during update time we can find issues way faster.

I addressed your concern regarding the base class issue by generating a mapping.

https://github.com/PaperMC/Paper/pull/10478/files#diff-a2c5f4e52ea4c2f48ddb4719a5c436f964a8e49beb30c3c90f099d4b8f94e18a

https://github.com/PaperMC/Paper/pull/10478/files#diff-e150aa8e1c1a506433452cd2d30284e9d7bba7a54f13d8c0faef70e715d816a4R150

patches/server/1061-add-disguise-api.patch
```diff
@@ -0,0 +1,440 @@
...
++    // Paper start - disguise api
++    private io.papermc.paper.disguise.DisguiseData disguiseData = io.papermc.paper.disguise.DisguiseData.original();
++    @Override
++    public @org.jetbrains.annotations.NotNull io.papermc.paper.disguise.DisguiseData getDisguiseData() {
++        return disguiseData;
++    }
++
++    @Override
++    public void setDisguiseData(@org.jetbrains.annotations.NotNull io.papermc.paper.disguise.DisguiseData disguiseData) {
++        getHandle().moonrise$getTrackedEntity().moonrise$clearPlayers();
```
yannicklamprecht: The following lines ensure that a force refresh is triggered for all players that can see the entity.

MiniDigger:
Hey @yannicklamprecht, since I see you rebasing this PR all the time: you really don't need to do that for this particular one.
While we do like this API in principle, it opened up a lot of discussion internally and we are not really sure yet on *how* we want to land it. We do have an idea on how it could look like post hard fork, but for now you don't need to waste time updating it, unless you need it for your own uses of course.
Thanks for sticking with us!

yannicklamprecht:
> Hey @yannicklamprecht, since I see you rebasing this PR all the time: you really don't need to do that for this particular one. While we do like this API in principle, it opened up a lot of discussion internally and we are not really sure yet on _how_ we want to land it. We do have an idea on how it could look like post hard fork, but for now you don't need to waste time updating it, unless you need it for your own uses of course. Thanks for sticking with us!



1. It's a no brainer to update the pr.
2. I need the feature in my own fork and I'm to lazy to workaround the fact that code generation for forks isn't working that easy after a code gen change somewhere this year.

yannicklamprecht:
Post hardfork MR: https://github.com/PaperMC/Paper/pull/11793

PR #10608, opened 2024-04-28
Rewrite reflection in library loader jars

jpenilla:


jpenilla:
I think this might fail on versioned/multi-release classes?

kennytv (approved):

PR #10612, opened 2024-04-28
Validate manifest provided mapping namespace

lynxplay:
With plugins adopting paperweights mapping namespaces in their manifest file in and after paper 1.20.5, incorrectly spelled namespaces previously were simply implied as the spigot namespace.

This commit properly ensures plugin authors are using a known mapping namespace in their manifest, avoiding down the line confusion on incorrect remapping behaviour.

The commit also fixes two incorrect log parameters.


jpenilla (approved):
Thanks

PR #10613, opened 2024-04-28
Data Components

Owen1212055:
This is like the item property api but split up into a lot of smaller chunks.

In this case, we replace item meta internally by a property map held on item stacks.

Opinions wanted:
- names
- serialization logic


patches/api/0474-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,433 @@
...
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/datakey/types/Unbreakable.java
+@@ -0,0 +1,15 @@
++package io.papermc.paper.datakey.types;
++
++import org.jetbrains.annotations.Contract;
++
++public interface Unbreakable {
++
++    static Unbreakable of(boolean showInTooltip) {
```
powercasgamer: this looks ugly

patches/server/1047-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,588 @@
...
++        public final DataComponentPatch.Builder builder = DataComponentPatch.builder(); // Paper
++        public void skullCallback(com.mojang.authlib.GameProfile gameProfile) {} // Paper - support updating profile after resolving it
+ 
+         <T> Applicator put(ItemMetaKeyType<T> key, T value) {
+             this.builder.set(key.TYPE, value);
+@@ -756,8 +756,13 @@ class CraftMetaItem implements ItemMeta, Damageable, Repairable, BlockDataMeta {
+         return result;
+     }
+ 
++    // Paper
```
powercasgamer:
```suggestion
+    // Paper start
```

patches/server/1047-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,588 @@
...
++
++    public static <NMS, API> void register(DataComponentType<NMS> type, DataKey<API> dummy, Class<API> apiTypeValidation, Function<API, NMS> nmsConverter, Function<NMS, API> apiConverter) {
++        ADAPTERS.put(BuiltInRegistries.DATA_COMPONENT_TYPE.getKey(type), new ComponentAdapter<>(type, apiTypeValidation, nmsConverter, apiConverter));
++    }
++
++    public record ComponentAdapter<API, NMS>(DataComponentType<NMS> type, Class<API> apiTypeValidation, Function<API, NMS> nmsConverter, Function<NMS, API> apiConverter) {
++
++    }
++
++
```
powercasgamer:
```suggestion
```

patches/api/0474-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,433 @@
...
+diff --git a/src/main/java/org/bukkit/Material.java b/src/main/java/org/bukkit/Material.java
+index 82d009c0bbe4b3026a535e02d6e0ed20c7bd525d..14de901564c8f8d7049047593bd16bc4d88505b2 100644
+--- a/src/main/java/org/bukkit/Material.java
++++ b/src/main/java/org/bukkit/Material.java
+@@ -128,7 +128,7 @@ import org.jetbrains.annotations.Nullable;
+ @SuppressWarnings({"DeprecatedIsStillUsed", "deprecation"}) // Paper
+ public enum Material implements Keyed, Translatable, net.kyori.adventure.translation.Translatable { // Paper
+     //<editor-fold desc="Materials" defaultstate="collapsed">
+-    AIR(9648, 0),
++    AIR(9648, 64), // Paper - air techncially stacks to 64
```
powercasgamer:
```suggestion
+    AIR(9648, 64), // Paper - air technically stacks to 64
```

patches/api/0474-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,433 @@
...
+new file mode 100644
+index 0000000000000000000000000000000000000000..f5c8c3c4f6989119934e47f82a120e486108c699
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/datakey/PatchedDataKeyMap.java
+@@ -0,0 +1,31 @@
++package io.papermc.paper.datakey;
++
++import io.papermc.paper.datakey.patch.DataKeyMapPatch;
++import org.jetbrains.annotations.NotNull;
++import javax.annotation.Nullable;
```
powercasgamer:
```suggestion
+import org.jetbrains.annotations.Nullable;
```

patches/api/0474-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,433 @@
...
++}
+diff --git a/src/main/java/io/papermc/paper/datakey/DataKeyMap.java b/src/main/java/io/papermc/paper/datakey/DataKeyMap.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..648a235377199c82df3896bf5fa6781dd5cce27d
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/datakey/DataKeyMap.java
+@@ -0,0 +1,26 @@
++package io.papermc.paper.datakey;
++
++import javax.annotation.Nullable;
```
powercasgamer:
```suggestion
+import org.jetbrains.annotations.Nullable;
```

powercasgamer (changes requested):

patches/api/0474-WIP-DataKey-API.patch
yannicklamprecht: Do you prefer something like that?

```suggestion
+    static Unbreakable.Builder builder() {
```

and usage would be smth like that:

```java
Unbreakable.builder().showInTooltip()....build();
```

Malfrador:
Not using vanilla names but "Datakey" instead is really confusing. Mojang officially (in changelogs) calls them "Item Stack Components" or "Data Components" and the Paper API should use a similar name - everything else is just confusing to users.

`itemStack.getDataKeyMap().remove(DataKeys.HIDE_ADDITIONAL_TOOLTIP);` 
is a lot more confusing than just
`itemStack.getComponents().remove(ItemComponent.HIDE_ADDITIONAL_TOOLTIP);`

I don't think adventure components existing justifies deviating from vanilla naming so much. The distinction between a _Text_ Component and a _Item_ Component is clear enough. 

Like, imagine someone wanting to look up a component they found in the API on the Minecraft Wiki, or the other way around. They wouldn't know what they are looking for.
This is also gonna get worse if Mojang decides to adapt the same component system for other parts of the game, such as entities or block entities.

patches/api/0474-WIP-DataKey-API.patch
kashike: `showInTooltip` should accept a `boolean`.

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
+index 0000000000000000000000000000000000000000..a226c4d1d901ea6086ac49263a0f21867328e36f
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/datakey/PatchedDataComponentHolder.java
+@@ -0,0 +1,21 @@
++package io.papermc.paper.datakey;
++
++import io.papermc.paper.datakey.map.DataKeyMap;
++import org.jetbrains.annotations.Nullable;
++
++public interface PatchedDataComponentHolder {
```
kennytv: This shouldn't be necessary, it'd be nicer to be able to have *everything* on ItemStack through implementing PatchedDataKeyMap directly (and renaming the methods to get/setData etc. for clarity)

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
+@@ -0,0 +1,25 @@
++package io.papermc.paper.datakey.map;
++
++import io.papermc.paper.datakey.DataKey;
++import org.jetbrains.annotations.Nullable;
++import java.util.Set;
++
++public interface DataKeyMap {
++
++    static DataKeyMap empty() {
```
kennytv: Is there any use for this?

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++
++public interface PatchedDataKeyMap extends DataKeyMap {
++
++    static PatchedDataKeyMap empty() {
++        return DataKeyMapBridge.Holder.BRIDGE.of(DataKeyMapBridge.Holder.BRIDGE.empty());
++    }
++
++    static PatchedDataKeyMap of(DataKeyMap map) {
++        return DataKeyMapBridge.Holder.BRIDGE.of(map);
++    }
```
kennytv: What's the use of these two? clearing should be done via an explicit `clearPatchedData`/`resetData` method or similar

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++    <T> T set(DataKey.Valued<T> type, @Nullable T value);
++
++    // Returns if this value was previously set
++    boolean set(DataKey.NonValued type);
++
++    @Nullable
++    <T> T remove(DataKey.Valued<T> type);
++
++    // Returns if this value was previously set
++    boolean remove(DataKey.NonValued type);
```
kennytv: Remove needs further discussion for how to allow resetting to a default vs. removing a default
Current nice ideas are
- removePath vs. remove
- remove vs. reset
- ???

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++
++    @NotNull
++    DataKeyMapPatch asPatch();
++
++    void applyPatch(@NotNull DataKeyMapPatch patch);
```
kennytv: This should just accept/return PatchedDataKeyMap directly and possibly be moved to ItemStack or a separate interface

kennytv (commented):
General ideas

patches/api/0475-WIP-DataKey-API.patch
Owen1212055: Do we not want some kind of middle man object? So for example people could apply certain changes to multiple item stacks? (I would allow people to make their own patch sets)

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++
++    PatchedDataKeyMap of(DataKeyMap keyMap);
++
++    DataKeyMap empty();
++
++    DataKeyMap itemStack(Material material);
++
++    PatchedDataKeyMap meta(Material material, ItemMeta itemMeta);
++
++    ItemMeta toMeta(Material material, PatchedDataKeyMap dataKeyMap);
```
kashike:
```suggestion
+    ItemMeta toItemMeta(Material material, PatchedDataKeyMap dataKeyMap);
```

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++@ApiStatus.Internal
++public interface DataKeyMapBridge {
++
++    PatchedDataKeyMap of(DataKeyMap keyMap);
++
++    DataKeyMap empty();
++
++    DataKeyMap itemStack(Material material);
++
++    PatchedDataKeyMap meta(Material material, ItemMeta itemMeta);
```
kashike:
```suggestion
+    PatchedDataKeyMap fromItem(Material material, ItemMeta itemMeta);
```

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++import org.jetbrains.annotations.ApiStatus;
++
++@ApiStatus.Internal
++public interface DataKeyMapBridge {
++
++    PatchedDataKeyMap of(DataKeyMap keyMap);
++
++    DataKeyMap empty();
++
++    DataKeyMap itemStack(Material material);
```
kashike:
```suggestion
+    DataKeyMap fromItem(Material material);
```

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++public interface Unbreakable {
++
++    static Unbreakable of(boolean showInTooltip) {
++        return ComponentTypesBridge.Holder.BRIDGE.unbreakable(showInTooltip);
++    }
++
++    boolean isShownInTooltip();
++
++    @Contract("_ -> new")
++    Unbreakable showInTooltip(boolean hideInTooltip);
```
kashike:
```suggestion
+    Unbreakable showInTooltip(boolean showInTooltip);
```

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++
++import org.jetbrains.annotations.Contract;
++
++public interface Unbreakable {
++
++    static Unbreakable of(boolean showInTooltip) {
++        return ComponentTypesBridge.Holder.BRIDGE.unbreakable(showInTooltip);
++    }
++
++    boolean isShownInTooltip();
```
kashike:
```suggestion
+    boolean showInTooltip();
```

kashike (changes requested):

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++    }
++
++    @Nullable
++    <T> T set(DataKey.Valued<T> type, @Nullable T value);
++
++    // Returns if this value was previously set
++    boolean set(DataKey.NonValued type);
++
++    @Nullable
++    <T> T remove(DataKey.Valued<T> type);
```
masmc05: This feels a little bit too ambiguous. Is it removing the patch? Is it removing it entirely? Even if specified in javadoc, it will be often forgotten.

Maybe instead of `remove` method, it could be:
- set(Key, null) - forces the absence of the data ("!minecraft:tool")
- reset(Key) - removes the patch/resets to vanilla (default) value
- unset(DataKey.NonValued) - forces the absence of the key (no unit value in nms)

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -0,0 +1,560 @@
...
++
++    // Returns if this value was previously set
++    boolean remove(DataKey.NonValued type);
++
++    @NotNull
++    DataKeyMapPatch asPatch();
++
++    void applyPatch(@NotNull DataKeyMapPatch patch);
++
++    PatchedDataKeyMap clone();
```
kennytv: clone bad, should be copySomething

patches/api/0475-WIP-DataKey-API.patch
Owen1212055: So you can wrap around a patched map

patches/api/0475-WIP-DataKey-API.patch
Owen1212055: There's no clearing here? This is used for declaring itemstacks.

patches/api/0475-WIP-DataKey-API.patch
Owen1212055: handled via conversation

patches/api/0475-WIP-DataKey-API.patch
Owen1212055: Remove has been split into some more methods

Owen1212055:
## General changes:
- DataKey -> DataComponentType (legit just went with vanilla names, we can trim names if need be)
- `PatchedDataComponentMap` no longer has `remove`, but has `unset`, and `reset`. Which I think make the most sense in this case.
    - Reset will remove the patch present on the map
    - Unset will unset it (so, like `!minecraft:tool`)
- ItemStack now implements `PatchedDataKeyMapHolder` which contains methods like `setData` which can be accessed on ItemStack directly, rather than needing to call `components()` evey time

## Builder Types:
   - Unbreakable: `Unbreakable.unbreakable().showInTooltip(false).build()`
   - ItemLore: `ItemLore.lore().lore(List.of(Component.text("h"))).build()`

It should be noted that there are builder for most components for not-simple future-proofing. Advanced objects that might hold a bit more context to them...

Additionally it helps gain a little more power in the data component API (such has being able to get the styled tooltip from the lore)

patches/api/0475-WIP-DataKey-API.patch
```diff
@@ -293,6 +293,23 @@ index 0000000000000000000000000000000000000000..325a597b8d3e0c574b7bc80b7c9d0dee
...
+new file mode 100644
+index 0000000000000000000000000000000000000000..7fe764cde13fcac1f3ceadd55bbb84a6692d866a
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/datakey/types/AbstractShownInTooltip.java
+@@ -0,0 +1,11 @@
++package io.papermc.paper.datakey.types;
++
++import org.jetbrains.annotations.Contract;
++
++public interface AbstractShownInTooltip<T> {
```
kennytv: No likey the "Abstract" prefix. Though in any case, I would move this to a different package

Owen1212055:
Added new pot decoration / charged projectiles type.

Owen1212055:
Superceded by https://github.com/PaperMC/Paper/pull/10845 to allow for easier contribution by other paper team members.

PR #10632, opened 2024-05-01
Add EquipmentSlot#getOppositeHand

SoSeDiK:
Adds convenience method to get an opposite hand.
Allows this:
```java
EquipmentSlot oppositeHand = hand.getOppositeHand();
```
Instead of this:
```java
var oppositeHand = hand == EquipmentSlot.HAND ? EquipmentSlot.OFF_HAND : EquipmentSlot.HAND;
```

### DIFF
--- a/patches/api/0376-Add-EquipmentSlot-convenience-methods.patch
+++ b/patches/api/0376-Add-EquipmentSlot-convenience-methods.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] Add EquipmentSlot convenience methods
 
 
 diff --git a/src/main/java/org/bukkit/inventory/EquipmentSlot.java b/src/main/java/org/bukkit/inventory/EquipmentSlot.java
-index 5642d8af60b6649497aba9b0f6ab7bba7702b9ee..adf0a94f98eb66b3957d4009d83ad5f741e0aa79 100644
+index 5642d8af60b6649497aba9b0f6ab7bba7702b9ee..a10d56b797e58b56bd4cef6de40691f7eac9b5b1 100644
 --- a/src/main/java/org/bukkit/inventory/EquipmentSlot.java
 +++ b/src/main/java/org/bukkit/inventory/EquipmentSlot.java
-@@ -33,4 +33,27 @@ public enum EquipmentSlot {
+@@ -33,4 +33,42 @@ public enum EquipmentSlot {
      public EquipmentSlotGroup getGroup() {
          return group.get();
      }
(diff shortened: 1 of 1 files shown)


kennytv (approved):

Owen1212055 (approved):

patches/api/0376-Add-EquipmentSlot-convenience-methods.patch
```diff
@@ -24,6 +24,21 @@ index 5642d8af60b6649497aba9b0f6ab7bba7702b9ee..adf0a94f98eb66b3957d4009d83ad5f7
...
++     *
++     * @return the opposite hand
++     * @throws IllegalStateException if this equipment slot is not a hand
++     * @see #isHand()
++     */
++    public @NotNull EquipmentSlot getOppositeHand() {
++        return switch (this) {
++            case HAND -> OFF_HAND;
++            case OFF_HAND -> HAND;
++            default -> throw new IllegalStateException("Can't get an opposite hand for not a hand");
```
Leguan16: Maybe something like?
```suggestion
+            default -> throw new IllegalStateException("Unable to determine opposite hand for hand: " + name());
```

kennytv: It's not gonna be a hand, so the end of the sentence doesn't work. Though also: that should be IllegalArgumentException or something, the state one implies the application is in the wrong state to accept a query

Leguan16:
```suggestion
+            default -> throw new IllegalArgumentException("Unable to determine opposite hand for equipment slot: " + name());
```
Something like this?

Leguan16: On a second thought is it necessary to throw an exception? Can't we just return null? Exception seems a little bit overkill, no?

electronicboy: returning null is generally crass, you're performing an operation which is enacting on an illegal argument to itself

SoSeDiK: Added name() to the output and changed to IllegalArgumentException.
Initially went with IllegalStateException due to exceptions' javadocs interpretation, they are somewhat vague.

Leguan16: "For not a hand" still sounds weird. Why don't we replace it with "for equipment slot"?

Machine-Maker (approved):

PR #10634, opened 2024-05-01
Extend fishing API

SoSeDiK:
Adds a missing fishing state when the fish is lured and fires an event for it.
Also adds a way to control the fish swimming time towards the bobber.


kennytv (approved):

patches/server/1045-Add-missing-fishing-event-state.patch
```diff
@@ -0,0 +1,26 @@
...
+index 35299affb699d745804a3b60cb78b9323f3e068e..bca9b5e9c7e57a81f4b9691a22e44d485cd223c1 100644
+--- a/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java
++++ b/src/main/java/net/minecraft/world/entity/projectile/FishingHook.java
+@@ -410,6 +410,15 @@ public class FishingHook extends Projectile {
+                     this.fishAngle = Mth.nextFloat(this.random, this.minLureAngle, this.maxLureAngle);
+                     this.timeUntilHooked = Mth.nextInt(this.random, this.minLureTime, this.maxLureTime);
+                     // CraftBukkit end
++                    // Paper start - Add missing fishing event state
++                    if (this.getPlayerOwner() != null) {
++                        var playerFishEvent = new PlayerFishEvent((Player) this.getPlayerOwner().getBukkitEntity(), null, (FishHook) this.getBukkitEntity(), PlayerFishEvent.State.LURED);
```
Owen1212055: no var please, use a qualified class name

SoSeDiK: Is there a specific policy regarding `var`s? I don't see anything in contributing.md and Paper uses it in a bunch of other places similar to this one.
It's a constructor, so it should be obvious what's returned?

patches/api/0475-Add-missing-fishing-event-state.patch
```diff
@@ -0,0 +1,23 @@
...
+          * in.
+          */
+         BITE
++        // Paper start - Add missing fishing event state
++        /**
++         * Called when a bobber was lured, and is now waiting to be hooked
++         * (when a "fish" starts to swim toward the bobber to bite it).
++         */
++        , LURED
++        // Paper end - Add missing fishing event state
```
Machine-Maker: Did we check that the javadocs actually work here? The comma being in front on the same line might mess with the javadoc generation.

SoSeDiK: Indeed, nice catch!
(*I should stop writing the code in the expense of sleep*)

lynxplay:
We did some heavy refactor during a merge session, could you please give your input on our changes, thank you :)

SoSeDiK:
Looks fine, I like it 👍

PR #10635, opened 2024-05-01
Add a way to check whether the chest is blocked

SoSeDiK:
Adds a way to check whether the chest is blocked (due to the block above / a sitting cat).

`CraftEnderChest`'s upper block check is hardcoded, so this PR uses `ChestBlock`'s one instead (they are identical and the implementation makes no sense to change in the future), but if preferred, the check could be extracted into a separated method and used instead.


kennytv (approved):

Machine-Maker (approved):

patches/server/1045-More-Chest-Block-API.patch
```diff
@@ -0,0 +1,50 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/block/CraftEnderChest.java
+@@ -58,4 +58,11 @@ public class CraftEnderChest extends CraftBlockEntityState<EnderChestBlockEntity
+         return getTileEntity().openersCounter.opened;
+     }
+     // Paper end - More Lidded Block API
++
++    // Paper start - More Chest Block API
++    @Override
++    public boolean isBlocked() {
++        return isPlaced() && net.minecraft.world.level.block.ChestBlock.isBlockedChestByBlock(getWorldHandle(), getPosition());
```
Machine-Maker: Add a comment explaining where this logic comes from, cause it's not super straight forward, in fact EnderChestBlock doesn't even use this method even though it does the same logic. I would add a `// Paper - diff on change` to the link in EnderChestBlock to catch any changes.

patches/server/1045-More-Chest-Block-API.patch
```diff
@@ -0,0 +1,50 @@
...
++        }
++        if (ChestBlock.isChestBlockedAt(getWorldHandle(), getPosition())) {
++            return true;
++        }
++        if (ChestBlock.getBlockType(this.data) == net.minecraft.world.level.block.DoubleBlockCombiner.BlockType.SINGLE) {
++            return false;
++        }
++        net.minecraft.core.Direction direction = ChestBlock.getConnectedDirection(this.data);
++        return ChestBlock.isChestBlockedAt(getWorldHandle(), getPosition().relative(direction));
++    }
```
Machine-Maker: Add a comment explaining the general area this logic is taken from.

Machine-Maker (changes requested):

patches/server/1045-More-Chest-Block-API.patch
SoSeDiK: Added comments. I'm not the best at explaining things, so suggestions welcome

patches/server/1045-More-Chest-Block-API.patch
SoSeDiK: Agree, noted.
I've added sanity checks used in source, and the method grew a bit. I'm not using chest's exact check as it's also somewhat hardcoded in a sense that redundant data is fetched & throw-away objects are created, so instead the method mimics it using the same checks & logic. Not sure if that's acceptable, or if something else is preferred.

Machine-Maker (approved):

PR #10637, opened 2024-05-01
Allow getting/setting the sign's editor uuid

SoSeDiK:


kennytv (approved):

patches/server/0860-More-Sign-Block-API.patch
```diff
@@ -1,48 +1,59 @@
...
-@@ -198,6 +198,14 @@ public class CraftSign<T extends SignBlockEntity> extends CraftBlockEntityState<
+@@ -198,6 +198,24 @@ public class CraftSign<T extends SignBlockEntity> extends CraftBlockEntityState<
      }
      // Paper end
  
-+    // Paper start - side facing API
++    // Paper start - More Sign Block API
++    @Override
++    public java.util.UUID getAllowedEditorUniqueId() {
++        return getTileEntity().getPlayerWhoMayEdit();
```
Owen1212055: and here?

~~Also, this is technically duplicate API with `getAllowedEditor`. Should we deprecate the old one here?~~

patches/server/0860-More-Sign-Block-API.patch
```diff
@@ -1,48 +1,59 @@
...
++    }
++
++    @Override
++    public void setAllowedEditorUniqueId(java.util.UUID uuid) {
++        getTileEntity().setAllowedPlayerEditor(uuid);
```
Owen1212055: Should we be calling `this.ensureNoWorldGeneration();` here? It seems to be called in getAllowedEditor.

SoSeDiK: Sure, won't hurt

Machine-Maker (approved):

lynxplay:
While the code generally looks good, could you elaborate a bit on what the usecase for such a setter is?
At least a setter for anything that isn't just `null`?

Having a bit of a hard time thinking of a usecase that would have a developer set the allowed uuid but not open the sign editor, given this value is terribly fleeting and is reset when the set uuid player is too far gone.

SoSeDiK:
> setter

It's quite niche, but is used exactly for that: either to reset the current editor, or to set one without immediately opening the sign screen

> getter

Player instance is obtained from uuid, and in some cases only the uuid is needed (e.g. if you just care whether the sign is being edited, or want to get value from Map by uuid). Getting Player isn't that bad for those cases, but since there's setter, getter doesn't hurt

lynxplay:
> or to set one without immediately opening the sign screen.

I am still unsure on what that would useful for. Any way to open said sign via the API would override the allowed editor, so would the player interacting with it. Which kinda only leaves packets, at which point, why have this in the API.

SoSeDiK:
> open said sign via the API would override the allowed editor

Correct, and is expected. Plugins should check whether there's already an editor if they want to avoid overrides.

> so would the player interacting with it

I just checked, and other players can't interact with the sign (open it / dye) if there's an editor set, and said editor is within range.

___
Setting the allowed player softly "notifies" other plugins that this sign is occupied without the need to actually open the sign on the same tick, or at all (e.g. you might want to open chest menu instead, and on inventory close, set the editor back no null).

The other way to handle the given example would be to cancel all interact events while you handle your own sign editing, but in my case I do not want to cancel literally all interacts since not all interactions interfere with editor, and other plugins can't know who's occupying the sign.

___
You brought up a good point about the range reset, though. While all my sign interactions are handled within the range, I think a note about this could be added to javadocs? Or even a way to toggle the range check? (but not whether the player's online one)

kennytv:
Just a note should be fine

SoSeDiK:
Rebased, clarified javadocs.

Owen1212055 (approved):

lynxplay (approved):

PR #10638, opened 2024-05-01
Expose #hasColor to leather armor

SoSeDiK:
`#hasColor` makes little sense in API context since leather armor always has a color, hence `#isDyed`

Fixed a missing leather horse armor in javadoc alongside

### DIFF
--- a/patches/api/0056-Fix-upstream-javadocs.patch
+++ b/patches/api/0056-Fix-upstream-javadocs.patch
@@ -1449,6 +1449,21 @@ index 18c2864c99d4dfae16cdb35143486aeebb9a6fd6..d66857825528ee772219440dffa28ad8
       * {@link EquipmentSlot} will be returned. This is because
       * AttributeModifiers without a slot are active in any slot.<br>
       * If there are no attributes set for the given slot, an empty map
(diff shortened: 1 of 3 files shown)


kennytv (approved):

Owen1212055:
This method is technically flawed, in that if someone manages to dye the armor with the exact color of `0xA06540` it'll appear as undyed. Is this issue or does it not really matter.

kennytv:
_If you dye grass green, is it truly dyed?_

Leguan16:
I mean it 1:1 reflects `hasColor` so I wouldn't say it's an issue here. Just my opinion.

Owen1212055:
Yes but it's craftbukkit's hasColor, which is not really great.

Regardless, it's probably fine. But just something to note.

SoSeDiK:
It's already noted in the javadoc:
> has a color different from {@link ItemFactory#getDefaultLeatherColor()}

#setColor also notes that having "no color" is same as having the default one:
> Setting it to null is equivalent to setting it to {@link ItemFactory#getDefaultLeatherColor()}.

Machine-Maker (approved):

PR #10639, opened 2024-05-01
Add Halloween API

SoSeDiK:
Allow checking / forcing Halloween Season/Day via API.
Also makes a tiny optimization over vanilla by using the month/day ints directly (they are stored in fields) instead of fetching from enum.

### DIFF
--- a/patches/server/1045-Add-Halloween-API.patch
+++ b/patches/server/1045-Add-Halloween-API.patch
@@ -1,0 +1,109 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: SoSeDiK <mrsosedik@gmail.com>
+Date: Wed, 1 May 2024 16:26:13 +0300
+Subject: [PATCH] Add Halloween API
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/ambient/Bat.java b/src/main/java/net/minecraft/world/entity/ambient/Bat.java
+index dc27ddf5131e7398a5390a5187261d4c7fb6ccaa..8eef60c1a3e6376955fde95a75a888b0b2028d77 100644
+--- a/src/main/java/net/minecraft/world/entity/ambient/Bat.java
++++ b/src/main/java/net/minecraft/world/entity/ambient/Bat.java
+@@ -240,11 +240,7 @@ public class Bat extends AmbientCreature {
+     }
+ 
+     private static boolean isHalloween() {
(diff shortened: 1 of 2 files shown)


patches/server/1045-Add-Halloween-API.patch
```diff
@@ -0,0 +1,109 @@
...
+@@ -240,11 +240,7 @@ public class Bat extends AmbientCreature {
+     }
+ 
+     private static boolean isHalloween() {
+-        LocalDate localdate = LocalDate.now();
+-        int i = localdate.get(ChronoField.DAY_OF_MONTH);
+-        int j = localdate.get(ChronoField.MONTH_OF_YEAR);
+-
+-        return j == 10 && i >= 20 || j == 11 && i <= 3;
++        return org.bukkit.Bukkit.isHalloweenSeason(); // Paper - Add Halloween API
```
electronicboy, line 20: I would much rather we not redirect mojang calls into replacements from the bukkit singleton

SoSeDiK: Implementation is the same, plus forced enable/disable. The patch should fail in case vanilla changes anything.
Not overriding these kinda makes half of this API (forcing/disabling Halloween) obsolete.

electronicboy: I generally in the long run want to start moving away from the Bukkit/Server "god" class, and so would generally prefer that we don't move implementation aspects to it

SoSeDiK: What would be the alternative?
Without it, plugins won't have a common ground to communicate, and in this case it makes no sense to enable/disable Halloween and yet have no according behavior changes.

electronicboy: Create another class and proxy the calls to that; maybe others on the team won't care as much, however

kashike:
Is this really needed in the API? This can be re-created in a plugin easily.

Owen1212055:
Agreed

SoSeDiK:
Depends. You need to duplicate the perfectly working vanilla system (having another time check, changing equipment on some entities, controlling bat's spawn rules and keeping track of changes in the future) while also not having any communication with other plugins.

I'd love for plugins to be able to add features during Halloween while also providing a ground to control it from one place (Paper's a good source to depend on, unlike some obscure plugin). I don't think maintaining this is much of a pain.
Feel free to close if doomed unwanted though.

PR #10646, opened 2024-05-02
Updated Upstream (CraftBukkit)

Machine-Maker:
Upstream has released updates that appear to apply and compile correctly. This update has not been tested by PaperMC and as with ANY update, please do your own testing

CraftBukkit Changes:
66fd94322 SPIGOT-7652: Remove remap for SPELL_MOB_AMBIENT which no longer exists
ecfa4f973 SPIGOT-7654: ItemStack#isSimilar does not work with empty BlockStateMeta
4460ecc49 SPIGOT-7655: ItemMeta#addItemFlags(ItemFlag.HIDE_ATTRIBUTES) not working when no attribute modifiers set
5d84f48a4 SPIGOT-7653: Update ApiVersion.CURRENT with latest version and include tests


Machine-Maker:
Ok, fixed the issue this introduced. We are going to keep the behavior of ItemFlag's not saving on itemstacks unless data that they hide is also present.

kennytv (approved):

lynxplay (approved):

PR #10651, opened 2024-05-03
fix: prevent profile lookups when unnecessary

### LINKED ISSUE #10648: Inefficient gameprofile lookups while using velocity's modern forwarding
### Timings or Profile link

N/A

### Description of issue

Upon examining (after getting rate limited **lol**), I noticed paper nodes are making mojang api calls upon every player join which was interesting to me as velocity should forward all of this info.

Upon digging deeper I realized this is completely redundant.

ServerLoginPacketListenerImpl#callPlayerPreLoginEvents(GameProfile) gets called on all occasions, upon forwarding using velocity, or just generic standalone paper.

What's interesting is that the this call always gets executed even when it shouldn't need to.

image
image

This call ultimately results in either;
#1 The usercache needing to lookup this gameprofile again
#2 A mojang api call (this is the main issue, however saving some CPU usage in case #1 is also nice).

If you have 3 joint sub-servers on your velocity network that all run on the same public ip, it is pretty easy to get rate limited, as I discovered the hard way 👍 

What I recommend (and I will be creating a PR for this, however I'd first like some feedback to confirm I'm not entirely delusional and am missing a use case for this call while using velocity):

If this call was made with the origin of velocity modern forwarding **and** the AsyncPlayerPreLoginEvent didn't modify the player profile, we no longer execute this call.

### Plugin and Datapack List

not important

### Server config files

Not important

### Paper version

All

### Other

As you can see here https://github.com/PaperMC/Velocity/blob/368d1a7c1250b4138f683e64ebb6aadfe7a4b1b3/proxy/src/main/java/com/velocitypowered/proxy/connection/PlayerDataForwarding.java#L73

All necessary gameprofile properties are being forwarded

bridgelol:
fixes #10648

### DIFF
--- a/patches/server/0140-Basic-PlayerProfile-API.patch
+++ b/patches/server/0140-Basic-PlayerProfile-API.patch
@@ -16,10 +16,10 @@ public org.bukkit.craftbukkit.profile.CraftPlayerProfile setProperty(Ljava/lang/
 
 diff --git a/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java b/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..68c99e29450e318d2b6eb83099e3e4012b6f0c01
+index 0000000000000000000000000000000000000000..a750ec2cde2c6b3942806241b2d972cef57f3d44
 --- /dev/null
 +++ b/src/main/java/com/destroystokyo/paper/profile/CraftPlayerProfile.java
-@@ -0,0 +1,412 @@
+@@ -0,0 +1,416 @@
 +package com.destroystokyo.paper.profile;
 +
 +import com.mojang.authlib.yggdrasil.ProfileResult;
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Owen1212055:
How does this affect game profiles with outdated info? Haven’t read this but does this prevent game profiles with outdated names / skins from being correctly updated?

bridgelol:
> How does this affect game profiles with outdated info? Haven’t read this but does this prevent game profiles with outdated names / skins from being correctly updated?

Really all this does is make it so gameprofile checks are fully handled on velocity (having paper do them as well is redundant)

electronicboy:
outdated info is, afaik, generally not updated unless something goes out of it's way to do that. There is generally no manner in which this method is called in which we're not just calling complete because we're expecting a new game profile, or that we're trying to recycle something old; I don't think that this will generally have an impact on anything reasonable

bridgelol:
Am also running this in a production environment with hundreds of concurrent players without any issues so far

kennytv (approved):

lynxplay (approved):

PR #10669, opened 2024-05-06
Add dispenserSpawnEntity event (1.19.4)

DerToaster98:
Adds a cancellable event that gets fired whenever a dispenser would spawn a entity (mainly useful for projectiles and primed tnt).

Workarounds for this would involve cancelling the entire event per item stack which in my opinion is not the best solution. Another workaround would be to use ASM or similar to manually adjust the source code. Or one could use NMS to manually override the dispense behaviors, which would be more effort over all though.

EDIT: Realized i made this on the 1.19.4 branch, will make a new PR for the master one

### DIFF
--- a/patches/api/0418-Add-DispenserSpawnEntityEvent.patch
+++ b/patches/api/0418-Add-DispenserSpawnEntityEvent.patch
@@ -1,0 +1,73 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DerToaster98 <38782719+DerToaster98@users.noreply.github.com>
+Date: Mon, 6 May 2024 12:33:00 +0200
+Subject: [PATCH] Add DispenserSpawnEntityEvent
+
+
+diff --git a/src/main/java/org/bukkit/event/block/DispenserSpawnEntityEvent.java b/src/main/java/org/bukkit/event/block/DispenserSpawnEntityEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..2207e9a3601ff67f3b9c13c6c23ad7c3d2a710aa
+--- /dev/null
++++ b/src/main/java/org/bukkit/event/block/DispenserSpawnEntityEvent.java
+@@ -0,0 +1,61 @@
++package org.bukkit.event.block;
++
++import javax.annotation.Nullable;
++
++import org.bukkit.block.Block;
++import org.bukkit.entity.Entity;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++
++/**
++ * 
++ * @author DerToaster98
++ *
++ * Called whenever a entity will be spawned by a dispenser.
++ * This includes:
++ *  - projectiles
++ *  - primed tnt
++ *  - spawn eggs
++ *  - boats
++ *  - minecarts
++ *  
++ *  Cancelling will cancel the entire action => no item shrinking etc.
++ *
++ */
++public class DispenserSpawnEntityEvent extends BlockEvent implements Cancellable {
++	
++	private static final HandlerList handlers = new HandlerList();
++	private boolean cancelled = false;
++	private final @Nullable Entity entity;
++	
++	public DispenserSpawnEntityEvent(@NotNull Block block, Entity entity) {
++		super(block);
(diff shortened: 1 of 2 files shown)


NonSwag:
I think a new event for that would be unnecessary 
The item dispense event is enough in my opinion 
You can easily check whether an item is a spawn egg or a projectile or any other kind of entity when dispensed

DerToaster98:
Yes, i *could* do that, but i will only have access to exactly the item using that approach. And for what i want to do i need access to the entity created when dispensed *and* the dispenser itself. Correct me if i'm wrong, but ItemDispenseEvent only gives access to the item. 

And i would need to replicate the entire logic for entity spawning through dispensed items (e.g. eggs, tnt, arrow and so on) for it to be accurate.

The other way that i'm aware of is using ASM or mixins or similar to achieve this or to use NMS and to re-register all the dispenseBehaviors with altered code.

NonSwag:
you could listen to the entity spawn event caused by dispensers

DerToaster98:
Iirc there is just an entity spawn event, no? Haven't seen one specific for dispensers.  And iirc the entity spawn event lacks a "reason" or context what spawned the entity or am i missing something here?

notTamion:
i didn't look at the implementation of this so i am not sure if this is possible but perhaps you could expand on the BlockDispenseEvent by adding a getEntity or something similar

NonSwag:
> Iirc there is just an entity spawn event, no? Haven't seen one specific for dispensers. And iirc the entity spawn event lacks a "reason" or context what spawned the entity or am i missing something here?
Entit#getEntitySpawnReason()
image

DerToaster98:
ok, that gets me the reason, but for what i want to do i need to have access to the dispenser too. ANd having to scan every of the 14 blocks around the entity is **not** an option

DerToaster98:
> i didn't look at the implementation of this so i am not sure if this is possible but perhaps you could expand on the BlockDispenseEvent by adding a getEntity or something similar
> 
> EDIT: also you can just relocate the current pr to the master branch you don't to create a new one

Hm, good idea

Machine-Maker:
I think this duplicates a lot of what is proposed in https://github.com/PaperMC/Paper/pull/7377.

DerToaster98:
> I think this duplicates a lot of what is proposed in #7377.

Looks like it. Will still keep this open though, cause the PR you mentioned hasn't been touched in quite some time.

NonSwag:
The idea of adding an entity to the existing dispense event is actually pretty good.
It could return a nonnull entity in every case because the item entity could also be returned.

Machine-Maker:
> EDIT 2: Haven't looked in detail but for what i want i need access to the dispenser BE that fired the projectile or spawned the tnt, not having it doesn't help me here.

The [projectile source](https://jd.papermc.io/paper/1.20.6/org/bukkit/entity/Projectile.html#getShooter()) **is** on the Projectile entity.

DerToaster98:
> > EDIT 2: Haven't looked in detail but for what i want i need access to the dispenser BE that fired the projectile or spawned the tnt, not having it doesn't help me here.
> 
> The [projectile source](https://jd.papermc.io/paper/1.20.6/org/bukkit/entity/Projectile.html#getShooter()) **is** on the Projectile entity with https://jd.papermc.io/paper/1.20.6/org/bukkit/projectiles/BlockProjectileSource.html.

Interesting, i will take a look, thanks!

DerToaster98:
> > EDIT 2: Haven't looked in detail but for what i want i need access to the dispenser BE that fired the projectile or spawned the tnt, not having it doesn't help me here.
> 
> The [projectile source](https://jd.papermc.io/paper/1.20.6/org/bukkit/entity/Projectile.html#getShooter()) **is** on the Projectile entity with https://jd.papermc.io/paper/1.20.6/org/bukkit/projectiles/BlockProjectileSource.html.

Ok, took a look, I can use that for any projectile, but i still need something for the tnt being dispensed. I know of workarounds but using those is not an option. Guess I'll adjust my solution here and wait for the (probably unlikely) case that it will be merged. Otherwise I need to find a way to compile the jar on my own.

Machine-Maker:
> Ok, took a look, I can use that for any projectile, but i still need something for the tnt being dispensed.

Yes, the mentioned PR does add an event for handling TNT which can't have a projectile source as its not a projectile.

PR #10679, opened 2024-05-09
Use internals for getting block/entity counts

electronicboy:
For a long time I've been meaning to move some of this logic internally as this would allow us to avoid hitting systems like block state snapshots which can create issues as many of the spigot implementations of this stuff are increasingly broken, leading to unexpected crashes during ticking, even if the API cannot properly interact with these such states/items, it's generally more preferable to not crash the server in the course, and just let those interactions fail more gracefully.


patches/server/0011-Use-TerminalConsoleAppender-for-console-improvements.patch
```diff
@@ -259,6 +260,39 @@ index 8323f135d6bf2e1f12525e05094ffa3f2420e7e1..a143ea1e58464a3122fbd8ccafe417bd
...
  }
+diff --git a/src/main/java/net/minecraft/server/Main.java b/src/main/java/net/minecraft/server/Main.java
+index 5b4ac7b4fd0077e900e9f788963f1613bbc9a5d0..544354eec56b884e322185dc0c6d43f3fa090f15 100644
+--- a/src/main/java/net/minecraft/server/Main.java
++++ b/src/main/java/net/minecraft/server/Main.java
+@@ -152,6 +152,28 @@ public class Main {
+                 return;
+             }
+ 
++            // Paper start - Use TCA for console - hack for broken Windows Java versions
```
kennytv: Wrong base here

patches/api/0011-Timings-v2.patch
```diff
@@ -2897,18 +2826,25 @@ index 3bf7db7eac81e3cc6f5c6700637d10d1b4b7a47b..77f8b0889cd7039bf041fc052fba33b6
...
 diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
-index 6e46302c272b468375f2de3f7f992f55f13805b8..01e796e487cc16710f51b457466a37ba70e1e665 100644
+index 6e46302c272b468375f2de3f7f992f55f13805b8..d54cd072b6d7f448122097996870eb6a0025caf7 100644
 --- a/src/main/java/org/bukkit/UnsafeValues.java
 +++ b/src/main/java/org/bukkit/UnsafeValues.java
-@@ -40,6 +40,7 @@ public interface UnsafeValues {
+@@ -1,5 +1,6 @@
+ package org.bukkit;
+ 
++import co.aikar.timings.TimingHistory;
```
Leguan16: Import?

electronicboy:
forced pushed to better deal with the wrong base so I can reduce the diff noise induced

kennytv (approved):

Machine-Maker:
I think we can probably just close this right? All this is timings-related diff and we're just gonna remove timings.

lynxplay:
With timings disabled by default now and deprecated for future removal in #11128, I'll close this PR, sorry cat </3

PR #10689, opened 2024-05-10
Added option to ignore entities when lightning strikes

Jakush:
Hey, in short, this patch adds option to ignore specific entity (or type of entity) from being damaged by lightning when lightning strikes.

### DIFF
--- a/patches/server/1050-Added-option-to-ignore-entities-when-lightning-is-st.patch
+++ b/patches/server/1050-Added-option-to-ignore-entities-when-lightning-is-st.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Added option to ignore entities when lightning is striking
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/LightningBolt.java b/src/main/java/net/minecraft/world/entity/LightningBolt.java
-index 4f701788bd21b61cad251a3a88f9bc416fb99051..cb755f5e017e003d740f9944bb3617af5d839019 100644
+index 4f701788bd21b61cad251a3a88f9bc416fb99051..ab5e83e3bb04967e57f7c2f3fe9a60c879243d71 100644
 --- a/src/main/java/net/minecraft/world/entity/LightningBolt.java
 +++ b/src/main/java/net/minecraft/world/entity/LightningBolt.java
 @@ -6,8 +6,10 @@ import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


lynxplay:
Hi, thank you for your PR :tada: 

Could you please elaborate on the usage of this API over simply listening to the respective events (Combust, HangingBreak, DamageEvent) which all expose the lightning strike entity?

This kind of API seems pretty counter active to the event driven design of the bukkit API.

Machine-Maker:
I would be more in favor of an event fired right before the lighting searches for an entity to hit to configure the parameters for that search or even to specify a specific entity before the search even happens. That seems more useful to me.

lynxplay:
Yea, then go with Machines way :+1: 
Adding this kind of state to the entity feels very off, an event prior to the lightnings actual strike sounds like the API friendlier approach.

lynxplay:
> Hello, I am trying to rebase my commit as in method 1, though I am not able to select anything, could you help me please?

Would probably be best if you ask for help on our [discord](https://discord.gg/papermc), github isn't a good place for that.

patches/server/1050-Added-option-to-ignore-entities-when-lightning-is-st.patch
```diff
@@ -0,0 +1,59 @@
...
+ import javax.annotation.Nullable;
++import io.papermc.paper.event.weather.PreLightningSelectionEvent;
+ import net.minecraft.advancements.CriteriaTriggers;
+ import net.minecraft.core.BlockPos;
+ import net.minecraft.core.Direction;
+@@ -31,6 +33,7 @@ import net.minecraft.world.phys.AABB;
+ import net.minecraft.world.phys.Vec3;
+ // CraftBukkit start
+ import org.bukkit.craftbukkit.event.CraftEventFactory;
++import org.bukkit.entity.LightningStrike;
```
notTamion: Please use fully qualified imports as you are modifying a non-paper file

notTamion (changes requested):
Adding methods for ignoring entities seems wrong. Why not call the event right after the server looks for entities to hit? then you can pass in the entities to the event and let the user add/remove entities

patches/server/1050-Added-option-to-ignore-entities-when-lightning-is-st.patch
Jakush: Should be resolved

notTamion:
> For me it is better to use more "direct" way to somehow ignore other entities than to check with EntityDamageByEntityEvent, of course it could be done just with this event.

why exactly do you not want to use EntityDamageByEntityEvent? this really seems like a very simple and intended way of doing this. we also don't have an event that gets called for when a zombie is checking for players they can attack, so why add one for this?

notTamion:
> You could use this event as a more direct way for filtering entities
I feel like that event is already as direct as you need it to be
> I could also imagine more features in this event in the future as configuring selection radius which you wouldn't want in EntityDamageByEntityEvent
Sure though i believe such a thing could be added to the LightningStrikeEvent or LightningStrike Entity. though you could also manually get the Entities in a specific area and damage them which gives you even more control over it.
> Event for zombie checking for players which he could attack does not exist though there is more general event for this EntityTargetEvent
EntityTargetEvent is called when a Mob starts targeting a player not when he checks if they are in reach to attack

notTamion:
oh i would have waited for other peoples opinion on this. Perhaps someone else has a different opinion on it

notTamion:
Thanks for your work on this anyways!

PR #10703, opened 2024-05-12
Fix CraftBukkit drag system

### LINKED ISSUE #9915: PaperMC breaks the intended Vanilla behavior of emptying bundles.
### Expected behavior

When bundles have items inside them, and you right click drag to empty the bundle, after dragging your cursor away from the slot that was right clicked and then releasing RMB, the newest item added to the bundle's inventory will be placed. This is the behaviour using the vanilla Minecraft Server Jar.

### Observed/Actual behavior

Holding right click and dragging your cursor while emptying a bundle causes the bundle itself to be dropped into a slot.
This makes emptying bundles more difficult, as even the slightest movement while right clicking to remove an item from the bundle

### Steps/models to reproduce

Using vanilla Minecraft:
Obtain a bundle in any way, you can use /give commands: "/give @s minecraft:bundle 1"
Put any item inside, or any amount of items inside
Select the bundle with LMB
Try to hold right click on any inventory slot, then drag your cursor while holding RMB.
Release RMB. The newest item should be placed in the inventory slot that was selected.

To replicate the issue using PaperMC, repeat the steps above, and you should find that the bundle itself is dropped, and not the newest item, which is NOT the intended behaviour of Minecraft's bundles.

### Plugin and Datapack List

No plugins or datapacks applicable. Base PaperMC version was used in 1.20.2.

### Paper version

[09:40:31 INFO]: This server is running Paper version git-Paper-280 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 4675152)

### Other

_No response_

notTamion:
resolves #9915

### DIFF
--- a/patches/server/1049-Fix-CraftBukkit-drag-system.patch
+++ b/patches/server/1049-Fix-CraftBukkit-drag-system.patch
@@ -5,58 +5,69 @@ Subject: [PATCH] Fix CraftBukkit drag system
 
 == AT ==
 public net.minecraft.world.inventory.AbstractContainerMenu quickcraftSlots
+public net.minecraft.world.inventory.AbstractContainerMenu quickcraftStatus
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index c450447585af4c8cdc87abe871c229ff895c3e53..49a3012064ee3d73cf4d5fa1a9e1713f955d8404 100644
+index c450447585af4c8cdc87abe871c229ff895c3e53..e2f9a442878d03d664f48a6df0ef185d3cf622ea 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-@@ -3118,6 +3118,51 @@ public class ServerGamePacketListenerImpl extends ServerCommonPacketListenerImpl
+@@ -3118,6 +3118,61 @@ public class ServerGamePacketListenerImpl extends ServerCommonPacketListenerImpl
                              }
                              break;
                          case QUICK_CRAFT:
 +                            // Paper start - Fix CraftBukkit drag system
-+                            if (AbstractContainerMenu.getQuickcraftHeader(packet.getButtonNum()) == 2 && this.player.containerMenu.quickcraftSlots.size() == 1) {
-+                                if (packet.getButtonNum() == 6) {
-+                                    click = ClickType.RIGHT;
-+                                    action = InventoryAction.PLACE_ONE;
-+                                } else {
-+                                    click = ClickType.LEFT;
-+                                    Slot slot = this.player.containerMenu.quickcraftSlots.iterator().next();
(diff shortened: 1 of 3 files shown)


patches/server/1049-Fix-CraftBukkit-drag-system.patch
```diff
@@ -0,0 +1,263 @@
...
+     private SlotAccess createCarriedSlotAccess() {
+diff --git a/src/main/java/net/minecraft/world/inventory/Slot.java b/src/main/java/net/minecraft/world/inventory/Slot.java
+index c39c773112fb8b534b926f2f2b47fe6fbb69fcb2..6c8a4d8ba0d27e6e5e58ec1f39b03ac6cd1f17ae 100644
+--- a/src/main/java/net/minecraft/world/inventory/Slot.java
++++ b/src/main/java/net/minecraft/world/inventory/Slot.java
+@@ -7,6 +7,7 @@ import net.minecraft.resources.ResourceLocation;
+ import net.minecraft.world.Container;
+ import net.minecraft.world.entity.player.Player;
+ import net.minecraft.world.item.ItemStack;
++import org.bukkit.craftbukkit.inventory.CraftItemStack; // Paper - Fix CraftBukkit drag system
```
Leguan16: why is this import necessary?

notTamion: cause i would have to do a fqi like 5 times in each class

Machine-Maker:
I feel like despite what CB's comment says about being unable to call the click event, we could call it by just mimicking some of the checks in the `case QUICK_CRAFT`.
```java
int currentStatus = this.player.containerMenu.quickcraftStatus;
int newStatus = AbstractContainerMenu.getQuickcraftHeader(packet.getButtonNum());
if ((currentStatus != 1 || newStatus != 2 && currentStatus != newStatus)) {
} else if (this.player.containerMenu.getCarried().isEmpty()) {
} else if (newStatus == 0) {
} else if (newStatus == 1) {
} else if (newStatus == 2) {
    if (!this.player.containerMenu.quickcraftSlots.isEmpty()) {
        if (this.player.containerMenu.quickcraftSlots.size() == 1) {
            // fire "click" event
        }
    }
}
```
Just put this right before the call to `AbstractContainerMenu#clicked`

notTamion:
I am having a bit of trouble understanding what you mean exactly. but if i am interpreting your comment correctly that wouldn't work because we don't have access to the new Items that early in the stack. i should probably also mention that cb's comment is wrong (assuming we want to call InventoryDragEvent if the mouse is moved in just the same slot, this way we don't break the existing calls for InventoryDragEvent and this also allows plugins to check whether or not the single item was dragged into the slot or clicked which might be useful for some developers)

notTamion:
@Machine-Maker i was able to construct a packet that seems to be calling the event correctly aswell as not breaking vanilla logic (i think) like you proposed yesterday.

Machine-Maker (approved):

lynxplay (approved):
LGTM.
Current drawback being that single slot "draws" do no longer
trigger the draw event. E.g. what is just a click event but with
a held down right click in a single slot.

Up for a core/maintainer to decide if the minimal break from spigot is fine
here, imo it is.

lynxplay (approved):

PR #10710, opened 2024-05-12
Add an 'empty' RecipeChoice for certain recipes and ingredient slots

Machine-Maker:
Adds `RecipeChoice#empty` which is allowed in specific recipes and ingredient slots.

---
Download the paperclip jar for this pull request: [paper-10710.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1519357910.zip)

### DIFF
--- a/patches/api/0477-Fix-issues-with-recipe-API.patch
+++ b/patches/api/0477-Fix-issues-with-recipe-API.patch
@@ -1,11 +1,20 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Jake Potrebic <jake.m.potrebic@gmail.com>
 Date: Sun, 12 May 2024 10:42:42 -0700
-Subject: [PATCH] Improve Recipe validation
+Subject: [PATCH] Fix issues with recipe API
 
+Improves the validation when creating recipes
+and RecipeChoices to closer match what is
+allowed by the Codecs and StreamCodecs internally.
+
+Adds RecipeChoice#empty which is allowed in specific
+recipes and ingredient slots.
+
+Also fixes some issues regarding mutability of both ItemStack
+and implementations of RecipeChoice.
 
 diff --git a/src/main/java/org/bukkit/inventory/CookingRecipe.java b/src/main/java/org/bukkit/inventory/CookingRecipe.java
-index f7fa79393aef40027446b78bac8e9490cfafd8bc..a64dae15db89038f3eb599f2a41edf45990f4fda 100644
+index f7fa79393aef40027446b78bac8e9490cfafd8bc..07906ca1a9b39fcc6774870daa498402f7f37917 100644
 --- a/src/main/java/org/bukkit/inventory/CookingRecipe.java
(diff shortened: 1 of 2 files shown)


tal5:
There's still an issue with the `Preconditions.checkArgument(!result.isEmpty(), "Recipe cannot have an empty result.")` check in the `SmithingRecipe` constructor failing for some vanilla recipes, which can be replicated by just going over them, e.g. `getServer().recipeIterator().forEachRemaining(recipe -> getLogger().info(""))`.
Not super familiar with the recipe system, but from what I can see only `SmithingTransformRecipe`s actually have a result, and `SmithingTrimRecipe`s just pass in an air `ItemStack` which triggers that check (I.e. can probably just move the check into `SmithingTransformRecipe`?)

Machine-Maker:
> There's still an issue with the Preconditions.checkArgument(!result.isEmpty(), "Recipe cannot have an empty result.")

Yeah, fixed that directly on master.

It did make me think about another possible error, and it turns out that upstream was just gonna insert `null` into RecipeChoice parameters in constructors if the nms Ingredient was empty. And that was gonna error. Now no vanilla recipes do that, but recipes added via datapacks surely could. So I've updated this PR to correctly handle converting an nms Ingredient back into an empty RecipeChoice. I also added a test that just runs the conversion for all recipes on the server which should catch changes like this in the future.

patches/api/0479-Fix-issues-with-recipe-API.patch
```diff
@@ -109,14 +182,14 @@ index db8bcc66bdc4bedfffb4705db6338eda4c0ad29a..c143b0d8b97d514566bac8413d0346cf
...
-@@ -232,5 +244,15 @@ public interface RecipeChoice extends Predicate<ItemStack>, Cloneable {
+@@ -232,5 +294,15 @@ public interface RecipeChoice extends Predicate<ItemStack>, Cloneable {
          public String toString() {
              return "ExactChoice{" + "choices=" + choices + '}';
          }
 +
 +        // Paper start - check valid ingredients
 +        @Override
-+        public @NotNull RecipeChoice validate() {
++        public @NotNull RecipeChoice validate(final boolean allowEmpty) {
```
kennytv: I'd rename this to allowEmptyRecipe to avoid ambiguity with empty items within the recipe

patches/api/0479-Fix-issues-with-recipe-API.patch
```diff
@@ -78,24 +87,88 @@ index 39f9766a03d420340d79841197f75c8b1dd49f4a..4e59f5176fd6cf92457ad750081c253a
...
++    /**
++     * An "empty" recipe choice. Only valid as a recipe choice in
++     * specific places. Check the javadocs of a method before using it
++     * to be sure it's valid for that recipe and ingredient type.
++     *
++     * @return the empty recipe choice
++     */
++    static RecipeChoice empty() {
++        //<editor-fold desc="holder" defaultstate="collapsed">
++        final class Holder {
```
kennytv: I personally think having the usual package private class is much cleaner than this, at least in this case

kennytv (approved):

PR #10714, opened 2024-05-13
Allow empty ingredients for smithing recipes

dawon:
As per discussion on discord.

not yet tested, will test hopefully later on today


Leguan16:
Without taking a look at impl isn't this implemented here with #10710?

dawon:
> Without taking a look at impl isn't this implemented here with #10710?

Completely overlooked Machine already did it... You are right...

PR #10719, opened 2024-05-13
Extend Player Interact cancellation to cover Jigsaw blocks

### LINKED ISSUE #10709: Jigsaw Interactions Cannot Be Cancelled With PlayerInteractionEvent
### Expected behavior

When cancelling a PlayerInteractionEvent caused by an opped player in creative mode interacting with a Jigsaw block, one would expect the Jigsaw block GUI to not open on the player client's screen, as the event was cancelled.

### Observed/Actual behavior

Instead, when the PlayerInteractionEvent related to the player (in creative mode) interacting with the jigsaw block is cancelled, the Jigsaw block's interface is still opened on the client. This is completely abnormal from what is expected when a PlayerInteractEvent is cancelled. For example:

- When a PlayerInteractEvent relating to a player opening a furance is cancelled, the client does not open the furnace GUI
- When a PlayerInteractEvent relating to a player opening a chest is cancelled, the client does not open the chest GUI
- When a PlayerInteractEvent relating to a (creative mode) player opening a command block is cancelled, the client does not open the command block GUI

### Steps/models to reproduce

1. Create a small plugin that implements a Listener similar to the one below to cancel all PlayerInteractEvents
    @EventHandler
    public void onPlayerInteractEvent(PlayerInteractEvent event) {
        event.setCancelled(true);
    }
2. Throw that plugin onto a server and boot it. While the server is running, every single PlayerInteractEvent should be cancelled.
3. Make sure you are listed as an operator and in creative mode.
4. Attempt to interact with a Jigsaw block.
5. Note that the Jigsaw block interface is still opened (even though all PlayerInteractEvents are being cancelled)

### Plugin and Datapack List

I am running no plugins other than the one listed above for canceling all PlayerInteractEvents.
I am using no data packs.

### Paper version

This server is running Paper version git-Paper-318 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: 9271ee7)
Previous version: git-Paper-307 (MC: 1.19.2)

### Other

_No response_

eteryi:
Fixes #10709

### DIFF
--- a/patches/server/0172-Extend-Player-Interact-cancellation.patch
+++ b/patches/server/0172-Extend-Player-Interact-cancellation.patch
@@ -10,7 +10,7 @@ Flower pots are also not updated on the client when interaction is cancelled, th
 also resolves this.
 
 diff --git a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
-index f4bf927477af23365a2d4618a1634195bd2054c8..5f2dec1917f1c1c3bb69446832321f3fc21dc129 100644
+index f4bf927477af23365a2d4618a1634195bd2054c8..e2d8bb349591d103606b158bc116b09d333c87ce 100644
 --- a/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
 +++ b/src/main/java/net/minecraft/server/level/ServerPlayerGameMode.java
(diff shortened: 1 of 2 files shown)


ChillyTheEevee:
This definitely seems to fix the problem! Easy solution

Machine-Maker (approved):

kennytv (approved):

electronicboy (approved):

lynxplay (approved):

PR #10725, opened 2024-05-14
Add ItemInHands API to Allay

notTamion:


patches/api/0480-Add-ItemInHands-API-to-Allay.patch
```diff
@@ -0,0 +1,32 @@
...
+     @Nullable
+     public Location getJukebox();
++
++    // Paper start - Add ItemInHands API to Allay
++    /**
++     * Sets the item the allay is currently holding.
++     *
++     * @param item the item to hold
++     */
++    void setItemInHands(@Nullable org.bukkit.inventory.ItemStack item);
```
lynxplay, line 22: I mean, `allay.getEquipment().setItemInMainHand(item);`

We can a) merge this into more entity api patch and b, probably just forward this call?
Or just not have this method in the first place, its how you'd set any other entities item in hand.

notTamion: oh welp, i opened this pr because yesterday someone on the discord was complaining they had to use nms to set the item and emily said she would open an issue (i guess she saw that method later aswell since she didn't open the issue yet) so after i couldn't find a method for it on the Inventory i just took their word for it and opened the pr lol. well ima close this then

kashike, line 22: `setItemInHand`

Issue #10713, opened 2024-05-13
Libraries loaded using `addURL(URL)` don't get remapped


frengor:
### Expected behavior

Libraries loaded calling `addURL(URL)` on the plugin class loader (for example, using a library like [Libby](https://github.com/AlessioDP/libby)) should be remapped if the plugin is spigot-mapped.

### Observed/Actual behavior

The classes of such libraries are not remapped. Consequently, the loaded libraries cannot access CraftBukkit or NMS classes.

### Steps/models to reproduce

I've created a test plugin available at <https://github.com/frengor/AddURLTest>. Steps to reproduce are present in the `README` together with the already built plugin.

### Plugin and Datapack List

None apart the plugin mentioned before.

### Paper version

[05:38:14 INFO]: This server is running Paper version git-Paper-71 (MC: 1.20.6) (Implementing API version 1.20.6-R0.1-SNAPSHOT) (Git: 11c3963)
You are running the latest version
Previous version: git-Paper-64 (MC: 1.20.6)

### Other

The test plugin correctly works on Spigot servers, but not on Paper servers.

This issue was originally discovered while testing the spigot mapped version of [UltimateAdvancementAPI](https://github.com/frengor/UltimateAdvancementAPI) on Paper 1.20.6. UltimateAdvancementAPI uses [Libby](https://github.com/AlessioDP/libby) to load [CommandAPI](https://github.com/JorelAli/CommandAPI) at runtime, which then cannot access NMS and CraftBukkit classes since it isn't remapped.

kennytv:
At that stage it is already too late to properly remap something via the normal system. libby could redirect to Spigot's maven dependency resolver where present to be properly integrated into the whole server ecosystem, or if it really needs to, call the remapper itself

frengor:
The problem with using the `libraryLoader` class loader is that it is present (i.e. not null) only if the plugin declares libraries in its `plugin.yml`, however plugins using Libby usually don't declare any. On top of that, when using Libby in a library you have no control on libraries declared by the plugins which will depend on you.

From `LibraryLoader.createLoader(PluginDescriptionFile desc)`:
```java
if ( desc.getLibraries().isEmpty() )
{
    return null;
}
```

Could always initializing the `libraryLoader` field to a non-null class loader be enough? (in which case I could make a PR)

PR #10726, opened 2024-05-14
Always return a non-null library loader from `LibraryLoader.createLoader(PluginDescriptionFile)`

frengor:
When there are no libraries declared in the `plugin.yml` of a plugin, the `PluginClassLoader.libraryLoader` field is initialized to a null value. Since it's marked as final, it's impossible to use the library loader (via reflection, for example using a library like [Libby](https://github.com/AlessioDP/libby)) in such case.

This PR fixes this by making `LibraryLoader.createLoader(PluginDescriptionFile)` always return a non-null class loader.

Also see https://github.com/PaperMC/Paper/issues/10713#issuecomment-2108086463

### DIFF
--- a/patches/api/0480-Always-return-a-non-null-library-loader.patch
+++ b/patches/api/0480-Always-return-a-non-null-library-loader.patch
@@ -1,0 +1,34 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: fren_gor <goro@frengor.com>
+Date: Tue, 14 May 2024 17:57:54 +0200
+Subject: [PATCH] Always return a non-null library loader
+
+Always returning a non-null class loader makes it possible to use the library loader (via reflection) even when there are no libraries declared in the plugin.yml of a plugin.
+
+diff --git a/src/main/java/org/bukkit/plugin/java/LibraryLoader.java b/src/main/java/org/bukkit/plugin/java/LibraryLoader.java
+index 8e1b6be2462aaa692efa1f72986921a6dc357196..b026d66102c58a0a7d497f28f69dea13fe1b1ed0 100644
+--- a/src/main/java/org/bukkit/plugin/java/LibraryLoader.java
++++ b/src/main/java/org/bukkit/plugin/java/LibraryLoader.java
+@@ -81,12 +81,20 @@ public class LibraryLoader
+         this.repositories = repository.newResolutionRepositories( session, Arrays.asList( new RemoteRepository.Builder( "central", "default", "https://repo.maven.apache.org/maven2" ).build() ) );
+     }
+ 
+-    @Nullable
++    @NotNull // Paper - always return a non-null library loader
+     public ClassLoader createLoader(@NotNull PluginDescriptionFile desc)
+     {
+         if ( desc.getLibraries().isEmpty() )
+         {
+-            return null;
++            // Paper start - always return a non-null library loader
++            URLClassLoader loader;
++            if (LIBRARY_LOADER_FACTORY == null) {
++                loader = new URLClassLoader( new URL[ 0 ], getClass().getClassLoader() );
++            } else {
++                loader = LIBRARY_LOADER_FACTORY.apply( new URL[ 0 ], getClass().getClassLoader() );
++            }
++            return loader;
++            // Paper end - always return a non-null library loader
+         }
+         logger.log( Level.INFO, "[{0}] Loading {1} libraries... please wait", new Object[]
+         {


patches/api/0480-Always-return-a-non-null-library-loader.patch
```diff
@@ -0,0 +1,34 @@
...
++    @NotNull // Paper - always return a non-null library loader
+     public ClassLoader createLoader(@NotNull PluginDescriptionFile desc)
+     {
+         if ( desc.getLibraries().isEmpty() )
+         {
+-            return null;
++            // Paper start - always return a non-null library loader
++            URLClassLoader loader;
++            if (LIBRARY_LOADER_FACTORY == null) {
++                loader = new URLClassLoader( new URL[ 0 ], getClass().getClassLoader() );
```
powercasgamer, line 26:
```suggestion
+                loader = new URLClassLoader(new URL[0], getClass().getClassLoader());
```

patches/api/0480-Always-return-a-non-null-library-loader.patch
```diff
@@ -0,0 +1,34 @@
...
++            } else {
++                loader = LIBRARY_LOADER_FACTORY.apply( new URL[ 0 ], getClass().getClassLoader() );
```
powercasgamer, line 28:
```suggestion
+                loader = LIBRARY_LOADER_FACTORY.apply(new URL[0], getClass().getClassLoader());
```

powercasgamer (changes requested):

patches/api/0480-Always-return-a-non-null-library-loader.patch
frengor: I was keeping the convention used in the rest of the file, see https://github.com/PaperMC/Paper/blob/7f2d5315fd85fda43b5a5e7040c3a47a68a052b5/patches/api/0473-Allow-modifying-library-loader-jars-bytecode.patch

Owen1212055:
Are we wanting to encourage such a flawed api? (Cough cough, maven central)

frengor:
This PR doesn't add any public API, it just makes it possible for libraries like [Libby](https://github.com/AlessioDP/libby) (of which I am a maintainer) to keep working.

The issue arises when loading spigot mapped libraries using Libby, since they don't get automatically remapped with the method we're using (see https://github.com/PaperMC/Paper/issues/10713). As suggested in https://github.com/PaperMC/Paper/issues/10713#issuecomment-2107380684, applying the remapping manually and then using the `libraryLoader` to load the additional libraries fixes the issues. It's just that the `libraryLoader` is usually set to `null`, and since it's marked as `final` we can't do anything about that from Libby.

patches/api/0480-Always-return-a-non-null-library-loader.patch
frengor: I was keeping the convention used in the rest of the file, see https://github.com/PaperMC/Paper/blob/7f2d5315fd85fda43b5a5e7040c3a47a68a052b5/patches/api/0473-Allow-modifying-library-loader-jars-bytecode.patch

ItsRainingHP:
Can't update my plugin without this fix merged.

lynxplay:
I am unsure what the point of this is?
You will already need internals beyond this as you need to run jars loaded by libby through the `io.papermc.paper.pluginremap.PluginRemapper` to remap them.

At that point, what prevents you from using the `io.papermc.paper.plugin.entrypoint.classloader.BytecodeModifyingURLClassLoader` ?
Libby isn't currently injecting its libraries into the library classloader from what I am gathering (given it can be null), so what is the point in hijacking this instead of just creating a BytecodeModifiyingUCL ?

frengor:
> At that point, what prevents you from using the `io.papermc.paper.plugin.entrypoint.classloader.BytecodeModifyingURLClassLoader` ?

The plugins using Libby need to access libraries loaded by Libby. This is possible only if the class loader of the plugin can resolve them, which is why Libby now calls `addURL` on it. Creating a new *separate* class loader isn't gonna work! It requires support from `org.bukkit.plugin.java.PluginClassLoader` (which already has special support for the `libraryLoader`).

The plan is to modify Libby to call `addURL` on the libraryLoader so that, if the plugin is spigot-mapped, library reflections will be remapped (since the libraryLoader will be an instance of `BytecodeModifyingURLClassLoader`). I know this isn't a perfect solution, but it is a minimal one.

A better solution would be to always have a `BytecodeModifyingURLClassLoader` *in addition* to the usual libraryLoader, even for mojang-mapped plugins, but I don't know if that would be preferred. (Actually, maybe a better compromise would be to have two separate libraryLoaders, one for mojang-mapped libraries and one for non-mojang-mapped ones)

lynxplay:
Ah fair, I thought it was completely injecting itself deeper, but yea if its just adding to #addUrl I guess I see the point.
I personally don't mind "hacking" the PluginClassLoader on that level, the spigot classloading system is already a hot mess.

I am unsure if we'd want to always initialise it or potentially leave it mutable, tho always init seems better imo so I would incline to approve the PR.

frengor:
> I am unsure if we'd want to always initialise it or potentially leave it mutable

Leaving it mutable opens to more things from the Libby side, like implementing what I was talking before *outside* the server implementation (citing myself):
>Actually, maybe a better compromise would be to have two separate libraryLoaders, one for mojang-mapped libraries and one for non-mojang-mapped ones

Always initializing the field is already enough though, so I don't really have a strong preference either.

jpenilla:
We've decided not to merge this, with Bukkit plugins having access to the PluginLoader API for programmatically adding runtime dependencies there is no reason for us to maintain patches enabling unsupported behavior like arbitrary addUrl calls.

ItsRainingHP:
Cool another reason to hate this update.

PR #10727, opened 2024-05-14
Add InteractionResult API

Dueris:
Exposes the InteractionResult API to Paper. Allows certain events to create "overrides" for the returned InteractionResult where the event is called. Allows plugin devs to use a more specific approach to handling the return value of the event without directly canceling it.

The enum InteractionResult in ``io.papermc.paper.world.interaction.InteractionResult`` has the following Enum Values:
- SUCCESS
- SUCCESS_NO_ITEM_USED
- CONSUME
- CONSUME_PARTIAL
- PASS
- FAIL

The override of the InteractionResult will only occur when the event isn't canceled. Getters inside the events modified return null for when no override is present.

### DIFF
--- a/patches/api/0480-InteractionResult-API.patch
+++ b/patches/api/0480-InteractionResult-API.patch
@@ -1,0 +1,249 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dueris <122416109+Dueris@users.noreply.github.com>
+Date: Tue, 14 May 2024 17:33:35 +0000
+Subject: [PATCH] InteractionResult API
+
+
+diff --git a/src/main/java/io/papermc/paper/world/interaction/InteractionResult.java b/src/main/java/io/papermc/paper/world/interaction/InteractionResult.java
+new file mode 100644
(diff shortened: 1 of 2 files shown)


Dueris:
I'll be fixing the paper-api test later today. Also planning on one last sweep to make sure that I got all the events I can

Dueris:
Modifying the result/return value of those methods allows to define what happens to the item and/or entity, and the client. For example:

Overriding the InteractionResult with the new API to SUCCESS would make it so the client swings its arm, instead of SUCCESS_NO_ITEM_USED which would still return a success but without swinging the arm. PASS would make it do nothing, FAIL would make it cancel whatever it's doing, etc. Its basically to add more detail to the interaction-based events. CONSUME iirc consumes the entire stack, and CONSUME_PARTIAL will only consume 1 of the item(if the stack has a higher amt value of 1).

onebeastchris:
Doesn't the Java client run these checks on its own too? Specifically, when deciding whether to swing the arm or not.. that's how the offhand functionality works, by checking the interactionresult of both hands, and seeing which of them would result in what

electronicboy:
This generally looks inherently dangerous in the long run, and fairly bad in the short term

Dueris:
I can add getters for the InteractionResults, however I've been quite busy lately so it will be a while

PR #10731, opened 2024-05-16
Fix CraftMetaBlockState for data components

### LINKED ISSUE #10697: Some items with empty attribute list receive additional data after interacting with them in the inventory
### Expected behavior

Moving an item in an inventory shouldn't change the item data

### Observed/Actual behavior

Moving a shield with attribute_modifiers:{show_in_tooltip: 0b, modifiers:[]} adds banner_patterns:[] to the item components

### Steps/models to reproduce

As requested in this bug report (https://github.com/PaperMC/Paper/issues/10686), here is the new bug report for the shield issue.

1. Give yourself a custom shield with `/give @p minecraft:shield[minecraft:attribute_modifiers={show_in_tooltip:0b,modifiers:[]}]`
2. Use `/data get entity @s SelectedItem` to confirm that the shield just has this component
3. Pick this item up with your cursor and place it in the inventory again
4. Use `/data get entity @s SelectedItem` again and notice how the shield now also has the component banner_patterns:[]

### Plugin and Datapack List

No plugins or data packs

### Paper version

This server is running Paper version git-Paper-64 (MC: 1.20.6) (Implementing API version 1.20.6-R0.1-SNAPSHOT) (Git: 6c02dbc)
Previous version: git-Paper-62 (MC: 1.20.6)

### Other

An (incomplete) list of other affected items:
shulker_box, chest: receive empty container:[] component
potion: receives empty potion_contents:{} component
bee_nest, beehive: receive block_entity_data:{id:beehive,Bukkit.MaxEntities:3},bees:[]
decorated_pot: receives pot_decorations:[brick,brick,brick,brick],container:[]

Machine-Maker:
This will go on forever...

----

Upstream's "restorePatch" function doesn't actually apply the patch which will compare the patch values to the itemtype's implicit values and not include the component if they equal. This fixes https://github.com/PaperMC/Paper/issues/10697.

---
Download the paperclip jar for this pull request: [paper-10731.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1537909977.zip)

### DIFF
--- a/patches/server/0638-Fix-upstreams-block-state-factories.patch
+++ b/patches/server/0638-Fix-upstreams-block-state-factories.patch
@@ -13,7 +13,7 @@ the material type of the block at that location.
 public net.minecraft.world.level.block.entity.BlockEntityType validBlocks
 
 diff --git a/src/main/java/net/minecraft/world/level/block/entity/BlockEntity.java b/src/main/java/net/minecraft/world/level/block/entity/BlockEntity.java
-index 6e24673a793017ee857cf75bf9a74105ce76b773..e3c5f99b3ad91a9bb454f9ab95b1ccff0bb7b34c 100644
+index 5e4cc5f54e8bf6afdab3afdf7f25c7b494e0d53b..7fa16b8a99509cc8f28b25513f0a1595219fe607 100644
 --- a/src/main/java/net/minecraft/world/level/block/entity/BlockEntity.java
 +++ b/src/main/java/net/minecraft/world/level/block/entity/BlockEntity.java
 @@ -371,7 +371,7 @@ public abstract class BlockEntity {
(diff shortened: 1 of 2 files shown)


electronicboy (approved):

Machine-Maker:
I found some more issues related to the beehive thing. the fix for that in this PR currently isn't correct and only fixes one instance of it. Fixing all of them will require more diff so am going back to draft until I get that done.

Machine-Maker:
Ok, I think this is now the most correct fix for block entities and itemmeta. Instead of storing the data in a CraftBlockEntityState which will add a bunch of extra stuff, this changes it to 2 fields for the custom data and the rest of the components that apply to the specific block entity type.

lynxplay (approved):

PR #7265, opened 2022-01-03
Fix crafting in custom inventories

Machine-Maker:
This uses some pretty absurd solutions I think to fix some existing issues with bukkit's inventory system and inventories created vis Bukkit.createInventory(). I recommend people use the HumanEntity#openSomething methods cause those are going to be pretty good, but they don't afford as much customization as the other ones.

Creating this as a draft because I haven't tested all the inventories and I'm sure others suffer from the same problem anvil, smithing tables, and enchant tables did. It's mainly non-tile-entity crafting block inventories, so I think looms, stonecutters, and grindstones are probably affected.

Fixes https://github.com/PaperMC/Paper/issues/10621

### Fixed

- [x] Enchanting
- [x] Anvil
- [x] Smithing
- [ ] Beacon
- [x] Loom
- [ ] Cartography
- [ ] Grindstone
- [x] Stonecutter

PR #10732, opened 2024-05-16
Properly handle 'result' type inventories

Machine-Maker:
Partially replaces https://github.com/PaperMC/Paper/pull/7265. That PR still has some stuff that this PR won't address.

----

While this might seem invasive, the pattern is very straightforward and easy to implement across all "result" type inventories. The worst part about it is the "mirroring" of the NonNullList that is initially created to hold the stacks. We have to use the exact same list for every Menu created so that the container is shared across multiple menus. We don't want to replace the `Container` instances on each Menu, because each `Container` instance created there has special logic specific to that menu, so instead we pass in the list of itemstacks from the created inventory. 

```[tasklist]
- [ ] Smithing
- [ ] Cartography
- [x] Stonecutter
- [ ] Anvil
- [x] Loom
- [ ] Grindstone
- [ ] Crafter
```

---
Download the paperclip jar for this pull request: [paper-10732.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2351995390.zip)

### DIFF
--- a/patches/server/1046-Properly-handle-result-type-inventories.patch
+++ b/patches/server/1046-Properly-handle-result-type-inventories.patch
@@ -1,0 +1,552 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Wed, 15 May 2024 20:15:39 -0700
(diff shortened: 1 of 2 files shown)


Machine-Maker:
I think this is now obsolete. With MenuType, and creating views correctly, this shouldn't be an issue anymore.

PR #10734, opened 2024-05-16
Restore vanilla teleportation logic for non-players

Machine-Maker:
Also adds more EntityTeleportEvent calls in the correct places cause there were some missing.

----

This can probably be squashed into some other patches.

---
Download the paperclip jar for this pull request: [paper-10734.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2351976654.zip)


ShaneBeee:
I did a PR once on CraftBukkit that did a big overhaul to the teleport methods, using Minecraft's logic.
I did this because Minecraft now supports TP with passengers via the TP command.
Unfortunately a few people over there felt that my PR took away the "old behaviours" of the Bukkit that people were used to.
OH NO... how dare I 😆 

So my goal for this PR was just to fix the bug.
I'm no pro when it comes to the NMS stuff like some of you guys are, so I didnt want to take on such a huge task, I just wanted to do a simple fix.

If another PR has a greater chance of making this so much better, I'm ok with this being closed and let someone else handle it.

That said I would be nice to fix the bug for the time being.

PR #10735, opened 2024-05-16
Fixes issues in the suppress dismount cancellation patch

Machine-Maker:
The `suppressCancellation` functionality here isn't used almost anywhere, especially in teleportation despite the patch title. However it will be once the re-work of internal teleportation logic is done. This fixes an issue in that patch where the original OBFHELPER name selected for the overloads turned out to conflict with a mojang named method so the overrides where *almost* linked together in such a way to cause a StackOverflow. I don't think there was a possibility of one, but this removes all chance.

### DIFF
--- a/patches/server/0271-force-entity-dismount-during-teleportation.patch
+++ b/patches/server/0271-force-entity-dismount-during-teleportation.patch
@@ -19,18 +19,39 @@ this is going to be the best soultion all around.
 
 Improvements/suggestions welcome!
 
+diff --git a/src/main/java/net/minecraft/server/level/ServerPlayer.java b/src/main/java/net/minecraft/server/level/ServerPlayer.java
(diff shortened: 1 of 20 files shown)


patches/server/0271-force-entity-dismount-during-teleportation.patch
```diff
@@ -92,8 +126,25 @@ index 9bb2b2baecfd4cfb9dbbf7cf5c8be939f1ecc4d6..47b1a4ec278cf9762c9eb9c1448cf789
...
+ 
+     @Override
+     public void stopRiding() {
++        // Paper start - Force entity dismount during teleportation
++        this.stopRiding(false);
++    }
++    @Override
++    public void stopRiding(boolean suppressCancellation) {
++        // Paper end - Force entity dismount during teleportation
+         super.stopRiding();
```
kennytv: Shouldn't this pass the boolean?

Machine-Maker: yep

electronicboy (approved):

Issue #10677, opened 2024-05-09
Custom banners patterns cause several bugs and crashes. (MC 1.20.5/1.20.6)


david85258:
### Stack trace

Forcing a timings report (Waiting 5 minutes cause the same result): https://paste.gg/p/anonymous/8fe00967c3d24d70b17da9c80831912a

Modify a BannerMeta of a Banner with a custom pattern: https://paste.gg/p/anonymous/bca4815e992041d698dbfcf0e5173bfb

Changing the slot of the banner with a custom pattern: https://paste.gg/p/anonymous/c2bdb4fa3a984863ac2750af121287f7

### Plugin and Datapack List

Server Plugins (1):
Bukkit Plugins:
  - Test

There are 4 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)], [file/many-more-banners-v0.2.2.zip (world)]

### Actions to reproduce (if known)

# In All
Add a data pack that's add custom patterns ([like this](https://modrinth.com/datapack/many-more-banners))

# Timings
Give a banner with a custom pattern (from the data pack "/give @s minecraft:black_banner[minecraft:banner_patterns=[{pattern:'mmb:anchor',color:'red'}]]")
Place the banner and wait to timings collect the data. (Can be forced cause using "/timings report")
Video: https://youtu.be/DP30mUcqIzs?si=dt4w9vNn1aZT3Qxr

# BannerMeta
Create a plugin that's modify the BannerMeta of a banner and modify the data of a banner with a custom pattern.

```java
ItemMeta itemMeta = p.getInventory().getItemInMainHand().getItemMeta();
if (itemMeta instanceof BannerMeta bannerMeta){
    bannerMeta.addPattern(new Pattern(DyeColor.RED, PatternType.CREEPER));
    ItemStack item = p.getInventory().getItemInMainHand();
    item.setItemMeta(bannerMeta);
    p.getInventory().setItemInMainHand(item);
}
````
Video: https://youtu.be/k93NWKhxX8E?si=x2NJHBXrjSTJJe8x

# Handle packet
Give a banner with a custom pattern (from the data pack "/give @s minecraft:black_banner[minecraft:banner_patterns=[{pattern:'mmb:anchor',color:'red'}]]")
Move the item in the inventory.
Video: https://www.youtube.com/watch?v=JO2bKsEw5FA

### Paper version

> version
[13:49:34 INFO]: Checking version, please wait...
[...]

PR #10740, opened 2024-05-17
Fix a few issues with ItemMeta

### LINKED ISSUE #7427: FireworkMeta inconsistencies with vanilla
### Expected behavior

FireworkMeta#get/setPower methods respect Vanilla bounds on the power.

### Observed/Actual behavior

`FireworkMeta#setPower` restricts the to >= 0 and < 128 when a value of 

### Steps/models to reproduce

To produce the error linked below, just serialize and then deserialize the itemstack you get from `/give @p minecraft:fireworks 1 0 {Fireworks:{Flight:128,Explosions:[]}}`

### Plugin and Datapack List

just my test plugin

### Paper version

This server is running Paper version git-Paper-"41263d9" (MC: 1.18.1) (Implementing API version 1.18.1-R0.1-SNAPSHOT) (Git: 41263d9)
Previous version: git-Paper-"cccb5d8" (MC: 1.18.1)

### Other

Mentioned in #7409, but should have its own issue because there's a discussion to be had about the correct fix.

#### Issue
Minecraft uses the byte nbt tag to store the firework rocket's power (or flight) which produces values of between -128 and 127 (inclusive). In the ctor for CraftMetaFirework, that value is converted to an unsigned int with
```java
this.power = 0xff & fireworks.getByte(FLIGHT.NBT);
```
which will produce values between 0 and 255 (inclusive).
That value is then returned by `FireworkMeta#getPower`.

When the itemmeta is serialized with craftbukkit's api serialization for metas, that 0-255 value is stored, but when it's deserialized
[this happens](https://paste.gg/p/Pencil/d0c487237c824916b46509b082003e9b). (from @Phoenix616)

#### Possible solutions
Allow values between 0 and 255 to be passed into setPower and note that values > 127 will be turned into negative numbers.
OR  
Keep the validations and just in fix the problem in deserialization.

Either one I think requires some note in the javadocs that getPower can return between 0 and 255 not 0-127 that might be inferred from the javadocs on setPower

Lulu13022002:
NPEs:
- OminousBottleMeta#getAmplifier when ominous_bottle_amplifier data component is absent from the patch
- TropicalFishBucketMeta when variant is null for almost all the color getter
- AxolotlBucketMeta#getVariant when variant is null
- ItemMeta#getAttributeModifiers(Attribute) when attribute modifiers is null
- SpawnEggMeta#getSpawnedEntity when entityTag is null

Codec/item validation:
This often kick the player holding the bad item instantly without saving the last data.
- FoodComponent#addEffect with out of bound probability (outside 0-1)
- the default food component created by Spigot (ItemMeta#getFood) is invalid since 'eat_seconds' cannot be equals to zero
- ItemMeta#set(Max)Damage doesn't check the range
- BundleMeta#addItem/setItems with empty item (amount=0)
- FireworkMeta#setPower takes 0-255 (fixes https://github.com/PaperMC/Paper/issues/7427)

ItemMeta bad equality:
This breaks the contract of equals that say x.equals(y) == y.equals(x)
- CraftMetaEntityTag (equals)
- CraftMetaSpawnEgg (equals)
- CraftMetaArmorStand (equals, hashCode), Paper only: doesn't check entityTag and is not consistent with hashCode
- CraftItemMeta (hashCode) handles the max damage as a simple boolean and not as an integer

Misc:
- empty fireworks effect is cleared by the item meta conversion (firework_rocket[minecraft:fireworks={}])
- firework effect with no main color got white added to it
- charged projectile for CrossbowMeta should allow other items and not only arrow or fireworks especially since now the picked up item is custom
- AxolotlBucketMeta#setVariant is not null but yet allow null to be handled as LUCY variant
- ItemMeta#getAttributeModifiers(EquipmentSlot) create an empty attribute_modifiers component, but getter shouldn't affect the underlying data
- FireworkMeta is not applied to firework because of the empty default item
- damage component can change if out of bound
- invalid colors are cleared when only the top byte should be ignored (not cleared) (PotionMeta, FireworkMeta/FireworkEffectMeta, ColorableArmorMeta, LeatherArmorMeta, MapMeta)
- mitigate crashes of https://github.com/PaperMC/Paper/issues/10677, pattern is lost however
- book page limit is infinite however internal still limit to 100 pages for writable book

<details><summary>Known issues (unfixed):</summary>
- No way to set some components: base_color (shield), intangible_projectile (for crossbow, in charged_projectiles component), tool, debug_stick_state etc...<br>
- The LeatherArmorMeta fallback to the default leather color even for wolf armor which doesn't make sense<br>
- Shield with base_color component gets an empty banner_patterns components once moved into the inventory<br>
</details>

Opened this as a draft since there are more stuff to check

### DIFF
--- a/patches/server/0692-More-Projectile-API.patch
+++ b/patches/server/0692-More-Projectile-API.patch
@@ -16,6 +16,7 @@ public net.minecraft.world.entity.projectile.Projectile hasBeenShot
 public net.minecraft.world.entity.projectile.Projectile leftOwner
 public net.minecraft.world.entity.projectile.Projectile preOnHit(Lnet/minecraft/world/phys/HitResult;)V
 public net.minecraft.world.entity.projectile.Projectile canHitEntity(Lnet/minecraft/world/entity/Entity;)Z
+public net.minecraft.world.entity.projectile.FireworkRocketEntity getDefaultItem()Lnet/minecraft/world/item/ItemStack;
 
 Co-authored-by: Nassim Jahnke <nassim@njahnke.dev>
 
@@ -203,8 +204,21 @@ index 98f46fadd60ea688fefa8d83dbd6fe9b61b6a96f..d4e0170694409e674d488f913e61c205
 +    }
 +    // Paper end
  }
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java
+index 7ba6302ecb72fa6e523054e7e3223d79eedf6589..907904da7f89e8e5e5cfab80977f04af3fdf17c7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java
+@@ -388,7 +388,7 @@ public final class CraftEntityTypes {
+             BlockPos pos = BlockPos.containing(spawnData.x(), spawnData.y(), spawnData.z());
+             return new FallingBlockEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), spawnData.world().getBlockState(pos)); // Paper - create falling block entities correctly
+         }));
+-        register(new EntityTypeData<>(EntityType.FIREWORK_ROCKET, Firework.class, CraftFirework::new, spawnData -> new FireworkRocketEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), net.minecraft.world.item.ItemStack.EMPTY)));
++        register(new EntityTypeData<>(EntityType.FIREWORK_ROCKET, Firework.class, CraftFirework::new, spawnData -> new FireworkRocketEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), FireworkRocketEntity.getDefaultItem()))); // Paper - pass correct default to rocket for data storage
+         register(new EntityTypeData<>(EntityType.EVOKER_FANGS, EvokerFangs.class, CraftEvokerFangs::new, spawnData -> new net.minecraft.world.entity.projectile.EvokerFangs(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), (float) Math.toRadians(spawnData.yaw()), 0, null)));
+         register(new EntityTypeData<>(EntityType.COMMAND_BLOCK_MINECART, CommandMinecart.class, CraftMinecartCommand::new, spawnData -> new MinecartCommandBlock(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z())));
+         register(new EntityTypeData<>(EntityType.MINECART, RideableMinecart.class, CraftMinecartRideable::new, spawnData -> new Minecart(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z())));
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java
-index 3c22799d36f07e349df207ce8a39236bbf6c17bf..2124b31ca4d994ad159556d47a315004b7246265 100644
+index e4e23a7b6d308ee476b5b8c2ad80efe4608b8346..241914415f74e0559fef59aa3f87f3e303f6c2c4 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftFireball.java
 @@ -32,20 +32,7 @@ public class CraftFireball extends AbstractProjectile implements Fireball {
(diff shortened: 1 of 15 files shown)


patches/server/1039-General-ItemMeta-fixes.patch
```diff
@@ -194,8 +216,34 @@ index 3f78a0935d738854182254b345064e3c225dcd5f..218df87c596d47b431dbbf2aa42822ef
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaEntityTag.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaEntityTag.java
+index 3f6c5cbbf63631e4b72dc43558651ea94f31ca78..da474a5b963d8e6769d120e9091e60ed0a468a9f 100644
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaEntityTag.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaEntityTag.java
+@@ -98,7 +98,7 @@ public class CraftMetaEntityTag extends CraftMetaItem {
+         if (meta instanceof CraftMetaEntityTag) {
+             CraftMetaEntityTag that = (CraftMetaEntityTag) meta;
+ 
+-            return this.entityTag != null ? that.entityTag != null && this.entityTag.equals(that.entityTag) : this.entityTag == null;
++            return this.entityTag != null ? that.entityTag != null && this.entityTag.equals(that.entityTag) : that.entityTag == null; // Paper
```
Machine-Maker: What changed here? Was something supposed to change here or was the comment just not removed?

patches/server/1039-General-ItemMeta-fixes.patch
```diff
@@ -248,6 +296,15 @@ index 8e0dd4b7a7a25a8beb27b507047bc48d8227627c..cf5d27ccc2225bac3aa57912f444f95d
  
          itemTag.put(CraftMetaFirework.FIREWORKS, new Fireworks(this.power, effects));
      }
+@@ -218,7 +231,7 @@ class CraftMetaFirework extends CraftMetaItem implements FireworkMeta {
+     }
+ 
+     boolean isFireworkEmpty() {
+-        return !(this.hasEffects() || this.hasPower());
++        return !(this.effects != null || this.hasPower()); // Paper - empty effects list should stay on the item
```
Machine-Maker: I think by changing this, you also need to change `hasEffects` or at least change in the places where `hasEffects` is used, like equalsCommon, hashCode, serialize

patches/server/1039-General-ItemMeta-fixes.patch
Lulu13022002: this to that at the end

Machine-Maker: wow. I thought I looked carefully at each word, afraid I was going to miss something. And I did anyways.

electronicboy:
having a skim, so far outside of what MM has pointed to, this generally looks fine; I'd need to double check that API def of that method, however

Machine-Maker:
> having a skim, so far outside of what MM has pointed to, this generally looks fine; I'd need to double check that API def of that method, however

Yeah, we don't *have* to change the API, but then we have to change where that method is used in the implementation so that in hashCode and equals and serialization they aren't treated the same.

electronicboy:
Concern is more "what's empty" for me

Machine-Maker:
> - Shield with base_color component gets an empty banner_patterns components once moved into the inventory

I think this is fixed in https://github.com/PaperMC/Paper/pull/10731 but I would appreciate you testing this as well. That PR should fix a lot of issues that are on itemtypes that can hold BlockEntities.

lynxplay (approved):

lynxplay:
Moved to ready for review, further changes can be their own PR.

Machine-Maker (approved):

lynxplay (approved):

patches/server/0692-More-Projectile-API.patch
```diff
@@ -203,6 +204,19 @@ index 98f46fadd60ea688fefa8d83dbd6fe9b61b6a96f..d4e0170694409e674d488f913e61c205
...
+index 7ba6302ecb72fa6e523054e7e3223d79eedf6589..907904da7f89e8e5e5cfab80977f04af3fdf17c7 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftEntityTypes.java
+@@ -388,7 +388,7 @@ public final class CraftEntityTypes {
+             BlockPos pos = BlockPos.containing(spawnData.x(), spawnData.y(), spawnData.z());
+             return new FallingBlockEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), spawnData.world().getBlockState(pos)); // Paper - create falling block entities correctly
+         }));
+-        register(new EntityTypeData<>(EntityType.FIREWORK_ROCKET, Firework.class, CraftFirework::new, spawnData -> new FireworkRocketEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), net.minecraft.world.item.ItemStack.EMPTY)));
++        register(new EntityTypeData<>(EntityType.FIREWORK_ROCKET, Firework.class, CraftFirework::new, spawnData -> new FireworkRocketEntity(spawnData.minecraftWorld(), spawnData.x(), spawnData.y(), spawnData.z(), FireworkRocketEntity.getDefaultItem()))); // Paper - pass correct default to rocket for data storage
```
electronicboy, line 216: by any chance is this the fix to firework meta not working due to the lack of an item?

Machine-Maker: Yes it is. vanilla changed something, and it caused the "empty" stack to actually take effect here. Should always have been defaulting to the rocket.

electronicboy (approved):

patches/server/1038-General-ItemMeta-fixes.patch
```diff
@@ -752,10 +832,134 @@ index 12a193db7475870e5107c86c7611bb4b92feacb8..860bc5ec4baec5e09a456cc5559d0de2
...
+     }
+ 
++    @io.papermc.paper.annotation.DoNotUse // Paper
+     static void readColor(LeatherArmorMeta meta, CraftMetaItem other) {
+         if (!(other instanceof CraftMetaLeatherArmor armorMeta)) {
+             return;
+         }
+ 
+-        meta.setColor(armorMeta.color);
++        // meta.setColor(armorMeta.color); // Paper - NOOP
```
Machine-Maker: Probably don't need this diff if we are `DoNotUse`ing the method

patches/server/1038-General-ItemMeta-fixes.patch
```diff
@@ -269,6 +269,77 @@ index f8c02fe01fd95aa5de8523c9ad452d91f5d3c16f..4447c754458e9fdead003b4044bf2bee
...
+ 
+     @Override
+     public void setColor(Color color) {
+-        this.color = color == null ? DEFAULT_LEATHER_COLOR : color;
++        this.color = color == null ? null : color.asRGB(); // Paper
+     }
+ 
+     boolean hasColor() {
+-        return CraftMetaLeatherArmor.hasColor(this);
++        return this.color != null; // Paper
```
Machine-Maker, line 339: We expose this `hasColor` method as a different name (`isDyed`), perhaps we should put that in the javadocs for `getColor` like other has/get combos. Use one before the other.

PR #10758, opened 2024-05-21
Allow using PluginLoader classpath API from Bukkit plugins

jpenilla:
Alternative solution to #10726

Allows using the PluginLoader API without any of the other changes imposed by switching to a `paper-plugin.yml`.

Used by setting `paper-plugin-loader` in `plugin.yml` to the class name of your [PluginLoader](https://jd.papermc.io/paper/1.20.6/io/papermc/paper/plugin/loader/PluginLoader.html).

Also allows skipping the `libraries` field by setting `paper-skip-libraries` to `true` (by default both `libraries` and jars provided by the PluginLoader are added to the classpath).


electronicboy (approved):

electronicboy:
#10726 is not inherently an issue, from what I gather, paper and spigot both inherently do not create a library loader if there is no dependencies registered, I'm not sure what the argument here is for botching internals to do something we already have API that we can expand the use case for.

frengor:
I agree with you that https://github.com/PaperMC/Paper/pull/10726 is not inherently an issue *of Paper*, however it is an issue for every user of Libby (of which I am a maintainer).

Don't get me wrong, I am in favor of this PR, my argument here it's just that this isn't an "alternative solution to https://github.com/PaperMC/Paper/pull/10726".

> I'm not sure what the argument here is for botching internals to do something we already have API that we can expand the use case for.

I'll expand more on the use cases of Libby. It's possible to use Libby to manage dependencies in *libraries which provide a shadeable version*, not just plugins (or libraries distributed as plugins like ProtocolLib). For example, in [UltimateAdvancementAPI](https://github.com/frengor/UltimateAdvancementAPI) we use Libby to download HikariCP only when MySQL is used (94% of users uses SQLite, so for them there is no need to download it). Also, Libby is used to download the correct version of [CommandAPI](https://github.com/JorelAli/CommandAPI), since not every MC version is supported by their latest release and shading all of them considerably increases the plugin size.  
UltimateAdvancementAPI can be downloaded as a plugin OR shaded. I am not aware of any way to load libraries without modifying the `plugin.yml`, but clearly I wouldn't like to require users which shade it to modify their `plugin.yml`.

frengor:
@masmc05
That would be good, however that's already possible to do using `PluginInitializerManager.instance().pluginRemapper.remapLibraries(paths)` (a method in unsafe would be better since its API, but calling the remapper is technically already possible). The part that's not currently possible is to remap reflections in loaded libraries (see https://github.com/PaperMC/Paper/pull/10726#issuecomment-2123187021).

jpenilla:
> While I see why this change is good for plugins which _only_ support Paper, I also don't think this is going to be commonly used in Bukkit plugins, since usually Bukkit plugins wants to be compatible with Spigot and `paper-plugin-loader` works only on Paper.
> 
> Also, this _doesn't solve_ #10726, since [Libby](https://github.com/AlessioDP/libby) doesn't require users to add a field to their `plugin.yml` and is supposed to work "out of the box".

You can continue to use the same hacks as always on spigot. This PR adds a supported API for configuring a Bukkit plugin's classpath on Paper. It does solve #10726 for every case that we intend to support.

kennytv (approved):

PR #10761, opened 2024-05-22
only show now-in-commands.yml for legacy files

NonSwag:
I know this is pretty unnecessary but why create new configs with old values
keeping that outdated s**t just makes future updates even messier


lynxplay:
Can probably merge this into https://github.com/PaperMC/Paper/blob/master/patches/server/0844-Don-t-enforce-icanhasbukkit-default-if-alias-block-e.patch and rename the patch to something like "Reduce bukkit configuration defaults"

PR #10770, opened 2024-05-22
Add registry-related argument types

Machine-Maker:
Will squash into brig patches before merging. Also added example usages of the 2 new arg types into the test plugin.

You run into the "unsafe cast" issue here with TypedKey, but I think its mostly unavoidable here. No way are we creating separate concrete types for each possible generic.

### DIFF
--- a/patches/api/0003-Test-changes.patch
+++ b/patches/api/0003-Test-changes.patch
@@ -66,7 +66,7 @@ index 0000000000000000000000000000000000000000..77154095cfb8b259bdb318e8ff40cb6f
 +    }
 +}
 diff --git a/src/test/java/org/bukkit/AnnotationTest.java b/src/test/java/org/bukkit/AnnotationTest.java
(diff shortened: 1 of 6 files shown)


patches/server/1046-Add-registry-related-argument-types.patch
```diff
@@ -0,0 +1,70 @@
...
++            ResourceKeyArgument.key(PaperRegistries.toNms(registryKey)),
++            nmsRegistryKey -> TypedKey.create(registryKey, CraftNamespacedKey.fromMinecraft(nmsRegistryKey.location()))
++        );
++    }
++
++    @Override
++    public <T> ArgumentType<T> resource(final RegistryKey<T> registryKey) {
++        return this.resourceRaw(registryKey);
++    }
++    @SuppressWarnings({"unchecked", "rawtypes", "UnnecessaryLocalVariable"})
```
lynxplay: Missing empty line

lynxplay (approved):

jpenilla (approved):

PR #10782, opened 2024-05-24
Allow to define new map cursor types

Lulu13022002:
The method is deprecated however internally everything in the api still use the byte equivalent and non deprecated api redirect to this method.

### DIFF
--- a/patches/api/0171-Fix-Spigot-annotation-mistakes.patch
+++ b/patches/api/0171-Fix-Spigot-annotation-mistakes.patch
@@ -40,7 +40,7 @@ index ac420f0059fc50d3e1294f85df7515c9e17ff78f..24daba85ce4129fb0babe67570059ca8
      public static Art getById(int id) {
          return BY_ID.get(id);
 diff --git a/src/main/java/org/bukkit/Bukkit.java b/src/main/java/org/bukkit/Bukkit.java
-index de678fa681fbd94efa8fd85568ee092ae26fca85..cddf3250c4a7efe239248b5e9662f33f5c0b702f 100644
+index f1a4e9a4b54757b9b73dd3a66ef6083a7119378d..2f023dcab9fe1ea220ba04e575bb5efe78adbd45 100644
 --- a/src/main/java/org/bukkit/Bukkit.java
 +++ b/src/main/java/org/bukkit/Bukkit.java
 @@ -879,9 +879,8 @@ public final class Bukkit {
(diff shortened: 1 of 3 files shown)


patches/api/0478-Allow-new-map-cursor-type-in-map-cursor.patch
```diff
@@ -0,0 +1,21 @@
...
+index c99311aa31ee293f4e61bca4bc3f401bb20f42f9..a9c82c54f221286c0b1ab5fa174f9bc94c527283 100644
+--- a/src/main/java/org/bukkit/map/MapCursor.java
++++ b/src/main/java/org/bukkit/map/MapCursor.java
+@@ -221,8 +221,8 @@ public final class MapCursor {
+      */
+     @Deprecated(forRemoval = true, since = "1.20.2") // Paper
+     public void setRawType(byte type) {
+-        if (type < 0 || type > 26) {
+-            throw new IllegalArgumentException("Type must be in the range 0-26");
++        if (type < 0 || type > 34) { // Paper
```
lynxplay: Does it make sense to dynamically fetch this via e.g.
```patch
diff --git a/src/main/java/org/bukkit/map/MapCursor.java b/src/main/java/org/bukkit/map/MapCursor.java
index c99311aa..c3b44cbb 100644
--- a/src/main/java/org/bukkit/map/MapCursor.java
+++ b/src/main/java/org/bukkit/map/MapCursor.java
@@ -221,8 +221,9 @@ public final class MapCursor {
      */
     @Deprecated(forRemoval = true, since = "1.20.2") // Paper
     public void setRawType(byte type) {
-        if (type < 0 || type > 26) {
-            throw new IllegalArgumentException("Type must be in the range 0-26");
+        class Bounds { public static final int UPPER_MAP_CURSOR_TYPE_BOUND = Type.values().length - 1; } // Paper
+        if (type < 0 || type > Bounds.UPPER_MAP_CURSOR_TYPE_BOUND) { // Paper
+            throw new IllegalArgumentException("Type must be in the range 0-" + Bounds.UPPER_MAP_CURSOR_TYPE_BOUND); // Paper
         }
         this.type = type;
     }
````

Comment (Lulu13022002) [reply patches/api/0478-Allow-new-map-cursor-type-in-map-cursor.patch]: yeah i thought of that not sure if it's worth too i don't have a strong opinion on this one.

Comment (lynxplay) [reply patches/api/0478-Allow-new-map-cursor-type-in-map-cursor.patch]: Well I doubt we'll be rewriting this part of the API any time soon with how many other places need more care.
Might be worth it to just, not think about this for the near future, and its at 0 runtime cost so.

File: patches/api/0480-Allow-new-map-cursor-type-in-map-cursor.patch [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+index c99311aa31ee293f4e61bca4bc3f401bb20f42f9..a046d32fb34fa1ede6f6141f0ca8e2f7b7ec1eb4 100644
+--- a/src/main/java/org/bukkit/map/MapCursor.java
++++ b/src/main/java/org/bukkit/map/MapCursor.java
+@@ -221,8 +221,9 @@ public final class MapCursor {
+      */
+     @Deprecated(forRemoval = true, since = "1.20.2") // Paper
+     public void setRawType(byte type) {
+-        if (type < 0 || type > 26) {
+-            throw new IllegalArgumentException("Type must be in the range 0-26");
++        class Bounds { public static final int UPPER_MAP_CURSOR_TYPE = Type.values().length - 1; } // Paper
```
kennytv: Move this to a proper private field on mapcursor or type (or a local var, it's not like this is a heavy operation), this isn't an interface

kennytv (approved):

Machine-Maker (approved):

PR #10786, opened 2024-05-25
Make HumanEntity spawnable

PlayerSchark:


yannicklamprecht:
What's your use case?

PlayerSchark:
Mainly so that you can spawn a Human easily and use it for NPCs.
Also I think that in the abstraction way it should be spawnable now. In NMS you can't spawn a ServerPlayer wihout hack but you can spawn a Player with ClientboundAddEntityPacket. This wasn't possible in earlier versions because it required ClientboundAddPlayerPacket which differed from the usal spawning. But this packet doesn't exist anymore.

lynxplay:
Closing this as it is completely unfeasible to have an *actual* Player instance on the server that is not a ServerPlayer.
A lot of the server code expects these things to be a ServerPlayer, hacking in a player implementation like this will explode everything.

There are a lot of existing frameworks to create and manage NPCs by the community, I'd recommend using one of these instead of spawning a custom player instance like this. Not to mention how this would mess with plugins or the likes.

I hope you can understand why paper is not interested in such a change 👍 
Thank you anyway for the work.

PR #10768, opened 2024-05-22
Properly fix invalid block entities being created

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/10750
Replaces https://github.com/PaperMC/Paper/pull/10375 which fixed https://github.com/PaperMC/Paper/issues/10022

----

This fixes the invalid block entities, not by removing them which the original fix did. that causes data loss as evidenced by the new issue this fixes relating to loot tables disappearing. The correct fix is to not mutate the WorldGenRegion's ProtoChunks when calling `BlockGetter#getBlockEntity`. Vanilla, if calling getBlockEntity on a block which doesn't have one and has a BlockState that should have one, cretes a BlockEntity and inserts it into the map (the mutation). Then if the block is later changed, the block entity and blockstate might not match. This isn't an issue in vanilla, because getBlockEntity is only caused in very specific places, but the BLockTransformer system was making it be called for *every* block in order to create a BlockState for the API. The fix is to just get a block entity as well as a "saving callback" in order to write the block entity back to the correct map to save changes to it.

----

Some testing around on worlds being generated would be appreciated so as to ensure there isn't any new data loss issue that was created.

---
Download the paperclip jar for this pull request: [paper-10768.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1537292397.zip)

PR #10794, opened 2024-05-26
Only remove worldgen block entity on changed block

### LINKED ISSUE #10750: Underwater ruined portal chests don't spawn loot.
### Expected behavior

Ruined portal chests should have random loot inside.

### Observed/Actual behavior

When a ruined portal chest is underwater, it doesn't generate loot.

### Steps/models to reproduce

Install paper 1.20.4, generate new chunks with ruined nether portal that has an underwater chest, open chest to see it is empty.

### Plugin and Datapack List

There are 2 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)]
Server Plugins (0):

### Paper version

This server is running Paper version git-Paper-496 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: 7ac24a1 on ver/1.20.4)

### Other

I checked to see if this behavior was present in spigot, or because of any plugins I had. The issue was only able to be reproduced in paper and purpur. Not sure if 1.20.4 is supported anymore. I tried my best to find a similar issue thread, but I am not adept at searching issues in github.

lynxplay:
The previous PR #10375 removes block entities on setBlock calls to the WorldGen region to prevent the existence of a block entity on a block that does not support it as such.

The logic however assumed that such calls would always follow the order of:
  1. set the base blockstate
  2. create the block entity
  3. finish world gen

This is however not the case for all of world gen, specifically not for structure templates being placed under water, as these first set the block state, then configure its block entity (include the chests lootable) and finally waterlogs the chest by calling another setBlock.

To prevent this logic from deleting the original block entity, this commit now only removes the block entity found if the blockstate to be set differs from the previous one, matching LevelChunk#setBlockState.

Supersedes: #10768
Fixes: #10750


jpenilla (approved):

kennytv (commented):
obfhelpers might be nice

lynxplay:
obfhelper added.

kennytv (approved):

PR #10798, opened 2024-05-27
Add EntityScareByEntityEvent for Armadillos

joshuaprince:
Add a new event for when an entity "scares" another entity. As of 1.20.6, only armadillos can be scared.

---

As far as I could tell, there is no way* to prevent a player from scaring 1.20.5's armadillos and making them roll up into their shells. This adds a simple way to do so.

*tried: EntityPoseChangeEvent (not called for armadillos, also no scared-by entity reference), EntityTargetEvent (not called for armadillos)

## Choices and Alternatives

1. This event is called pre-cancelled when *any* living entity is within 7 blocks of the armadillo.
    - Pro: Plugins have full control over the conditions that scare an armadillo, such as adding more entities that scare armadillos.
    - Con: Lots of pre-cancelled events. Might be easy to misuse or log spam with `ignoreCancelled = false`.
    - Alternative: Only call the event when the scare actually happens, so that the event can be cancelled but not forced.

2. The event is named generically.
    - Pro: Anything "scareable" in the future can be rolled into this event.
    - Con: This could be misinterpreted as being called when wolves scare skeletons, cats scare creepers, etc.
    - Alternative 1: Include skeleton and creeper scares by pets in this event.
    - Alternative 2: Rename the event `ArmadilloScareByEntityEvent`.

## Testing

- Event is called as expected while an entity is within 7 blocks of an armadillo, ~3 times per second.
- Event continues to be called while the entity is nearby, even while the armadillo is rolled up.
- `getEntity()` and `getScaredBy()` return the correct entities in a test plugin.
- `setCancelled(true)` makes armadillos unscareable.
- `setCancelled(false)` makes armadillos immediately scared as soon as any living entity is nearby.

### DIFF
--- a/patches/server/1051-Add-EntityScareByEntityEvent.patch
+++ b/patches/server/1051-Add-EntityScareByEntityEvent.patch
@@ -1,0 +1,40 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Joshua Prince <joshua@jtprince.com>
+Date: Sun, 26 May 2024 19:23:43 -0700
+Subject: [PATCH] Add EntityScareByEntityEvent
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/animal/armadillo/Armadillo.java b/src/main/java/net/minecraft/world/entity/animal/armadillo/Armadillo.java
+index b38281f963377cc82b360e8457da7cad033b8c36..c9dbf3444c73f2c983d511db33dbf74807e3b82c 100644
+--- a/src/main/java/net/minecraft/world/entity/animal/armadillo/Armadillo.java
++++ b/src/main/java/net/minecraft/world/entity/animal/armadillo/Armadillo.java
+@@ -229,17 +229,24 @@ public class Armadillo extends Animal {
+     public boolean isScaredBy(LivingEntity entity) {
+         if (!this.getBoundingBox().inflate(7.0D, 2.0D, 7.0D).intersects(entity.getBoundingBox())) {
+             return false;
+-        } else if (entity.getType().is(EntityTypeTags.UNDEAD)) {
+-            return true;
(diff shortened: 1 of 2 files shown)


patches/server/1051-Add-EntityScareByEntityEvent.patch
```diff
@@ -0,0 +1,40 @@
...
+-            return true;
++            isScared = true;
+         } else if (entity instanceof Player) {
+             Player entityhuman = (Player) entity;
+ 
+-            return entityhuman.isSpectator() ? false : entityhuman.isSprinting() || entityhuman.isPassenger();
++            isScared = entityhuman.isSpectator() ? false : entityhuman.isSprinting() || entityhuman.isPassenger();
+         } else {
+-            return false;
++            isScared = false;
```
kennytv, line 32: Shouldn't be called here

kennytv (commented):
Having it generic is good I think, but then you should already add a call in `AvoidEntityGoal` for the other uses

patches/server/1051-Add-EntityScareByEntityEvent.patch
joshuaprince: It's also called at L29, even if the Player didn't actually scare the armadillo. Is it worth calling the event pre-cancelled in these cases? I put my reasoning in the PR body (choice 1).

joshuaprince:
Replaced by https://github.com/PaperMC/Paper/pull/10835

patches/server/1051-Add-EntityScareByEntityEvent.patch
kennytv: Separate API to expand what it is scared by might be more appropriate, I think this would already be doable by adding a custom goal. In any case, we generally don't call events for such cases

PR #10804, opened 2024-05-28
Fix ProjectileHitEvent not called for multiple HitResults

### LINKED ISSUE #10801: ProjectileHitEvent not called multiple times for entites close to each other
### Expected behavior

When having 2 entities stand right behind each other. cancelling the ProjectileHitEvent would result in the event being called once for the first entity and once for the second entity.

### Observed/Actual behavior

the event only gets called once for the closest entity to the shooter however not for the second entity

### Steps/models to reproduce

cancel ProjectileHitEvent and print out when the event gets called with the playername. line up 2 players/entities (i would just use 2 players since this issue also relies on how big the hitbox of the entity is) as close as possible behind each other. shoot straight through both entites/players. see that the event only got called once

### Plugin and Datapack List

just a test plugin to cancel the event

### Paper version

[10:12:21 INFO]: Checking version, please wait...
[10:12:21 INFO]: This server is running Paper version 1.20.6-113-master@23fe116 (2024-05-27T17:40:35Z) (Implementing API version 1.20.6-R0.1-SNAPSHOT)

### Other

This issue happens because in the logic where the arrow checks if there is an entity to hit in the hit box it only returns the closest entity instead of all entities in that area, resulting in it not trying to hit the next entity if the event is cancelled.

### LINKED ISSUE #11478: Cancelling ProjectileHitEvent for entity collisions allows arrows to phase through blocks
### Expected behavior

Cancelling `ProjectileHitEvent` for entity collisions should not prevent block collisions.

### Observed/Actual behavior

Cancelling a `ProjectileHitEvent` fired for hitting an entity can cause arrows to phase through blocks.

### Steps/models to reproduce

Shoot an arrow at an entity on the edge of its bounding box, towards the ground. If positioned correctly, the arrow will phase into a block. If the ground is thin enough, it will continue travelling after phasing into the block. Otherwise, it will be lodged within the ground.
Example image of lining up a shot:
Image

### Plugin and Datapack List

Only a test plugin, with the following listener to cancel `ProjectileHitEvent`:
```java

    @EventHandler
    public void onHit(ProjectileHitEvent event) {
        if (event.getHitEntity() != null) event.setCancelled(true);
    }

}
```

### Paper version

This server is running Paper version 1.21.1-119-master@7cd4f2c (2024-10-03T15:19:34Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21.1-85-c5a1066 (MC: 1.21.1)

### Other

This seems to be a result of the logic in `AbstractArrow#tick` and `Projectile#preHitTargetOrDeflectSelf`. Arrows first ray trace to determine any block collisions, and then does another ray trace to find hit entities. If an entity that can be hit is found, the hit result is passed to `Projectile#preHitTargetOrDeflectSelf`. As such, the block collision is never considered, including when the `ProjectileHitEvent` is cancelled.

notTamion:
fixes #10801

---
Download the paperclip jar for this pull request: [paper-10804.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2351968475.zip)


notTamion:
just one thing when you guys decide to test the pr. arrows just disappearing client side when the event is cancelled is also an issue without the patch but i am not sure wether or not that is fixable

notTamion:
So right now this probably has issues with the piercing enchantment (haven't tested but is very likely). before i however resolve that i would like to know what design we will be going forwards with. Lynx suggested instead of adding new methods to ProjectileUtil which return all entities in the range, we should do something similar to piercingIngoreEntityIds. I have tried to implement this for 1 or 2 of the Projectile Classes however i noticed that the diff is often even bigger/ just as big as before, we also have to clip entites more often with his implementation.

patches/server/1065-Fix-ProjectileHitEvent-not-calling-for-multiple-enti.patch
```diff
@@ -0,0 +1,307 @@
...
++    ) {
++        Vec3 vec3 = pos.add(velocity);
++        HitResult blockHitResult = world.clip(new ClipContext(pos, vec3, raycastShapeType, ClipContext.Fluid.NONE, entity));
++        if (blockHitResult.getType() != HitResult.Type.MISS) {
++            vec3 = blockHitResult.getLocation();
++        }
++
++        java.util.List<HitResult> hitResults = (java.util.List<HitResult>)(java.util.List<?>) getEntityHitResults(world, entity, pos, vec3, entity.getBoundingBox().expandTowards(velocity).inflate(1.0), predicate, margin);
++        if (!hitResults.isEmpty()) {
++            if (blockHitResult.getType() != HitResult.Type.MISS) hitResults.add(blockHitResult);
```
TonytheMacaroni: The collection returned from `getEntityHitResults` is immutable, causing this to throw an exception.

notTamion: Must have not had the right angle when testing. thanks

notTamion:
Well get around to fully testing it later, for now it compiles and starts.

patches/server/1066-Fix-ProjectileHitEvent-not-called-for-multiple-HitRe.patch
```diff
@@ -0,0 +1,343 @@
...
++        for (Entity entity3 : world.getEntities(entity, box, predicate)) {
++            AABB aABB = entity3.getBoundingBox().inflate(margin);
++            Optional<Vec3> optional = aABB.clip(min, max);
++            if (optional.isPresent()) {
++                double e = min.distanceToSqr(optional.get());
++                entities.put(e, entity3);
++            }
++        }
++
++        return entities.values().stream().map(EntityHitResult::new).collect(java.util.stream.Collectors.toList());
```
TonytheMacaroni: Small nitpick: [`Collectors#toList` has undefined mutability](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Collectors.html#toList()). I believe it's almost always implemented to return a mutable list, but maybe use `Colllectors#toColllection` instead?

PR #10808, opened 2024-05-28
Handle alpha channel from potion_contents component

Lulu13022002:
The previous code assumed that PotionContents#getColor always returned a RGB color without alpha channel. However that's not the case it returns an ARGB color even for opaque color, the alpha channel is set to 0xFF. Only AreaEffectCloud always have an opaque RGB color and will force the first byte to 0xFF.

The following things happen:
- AreaEffectCloud#getColor throw an argument exception with extraneous data.
- Arrow#getColor would always return null no matter the color of the effect since the alpha channel is determined by the first byte of the integer. Which means for opaque color (0xFF) the first bit will be set to 1 and so negative color are fine. Only -1 is a special case here.
- Arrow#setColor doesn't work since the alpha channel is not defined.

### DIFF
--- a/patches/api/0056-Fix-upstream-javadocs.patch
+++ b/patches/api/0056-Fix-upstream-javadocs.patch
@@ -1571,7 +1571,7 @@ index a23d030d2204098be17d8b18021fd0bb79b4431b..f427334c6e875a13aa53052ac1b5a746
       * AttributeModifiers without a slot are active in any slot.<br>
       * If there are no attributes set for the given slot, an empty map
 diff --git a/src/main/java/org/bukkit/inventory/meta/LeatherArmorMeta.java b/src/main/java/org/bukkit/inventory/meta/LeatherArmorMeta.java
-index c1676991c3cc5f8d6e3f97d8cb356d6e2aa52809..1d61cc4ab36413fe3c1ecdf9824a5e18cb4bc148 100644
+index c1676991c3cc5f8d6e3f97d8cb356d6e2aa52809..c701d5fbc5fef503f18a3a46fa54c983bf96e895 100644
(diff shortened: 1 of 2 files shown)


patches/server/0692-More-Projectile-API.patch
```diff
@@ -204,6 +205,80 @@ index 98f46fadd60ea688fefa8d83dbd6fe9b61b6a96f..0cc1cdf91deb07ebb437ef5e61d149b2
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
+index 81f5e1d866128af8fb2acc13aca715580fdf9886..38d180eab104259a359e858bab1eca0d5ec0a8d1 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
+@@ -125,7 +125,7 @@ public class CraftAreaEffectCloud extends CraftEntity implements AreaEffectCloud
+ 
+     @Override
+     public Color getColor() {
+-        return Color.fromRGB(this.getHandle().potionContents.getColor());
++        return Color.fromRGB(this.getHandle().potionContents.getColor() & 0xFFFFFF); // Paper - skip alpha channel
```
jpenilla: Pretty sure you want `color & 0x00FFFFFF` here

Machine-Maker (approved):

kennytv (approved):

Owen1212055 (approved):

PR #10823, opened 2024-05-29
Fix max temper API for llama

Lulu13022002:
The custom field has been overwritten and cannot be changed using the api method for the llama.

### DIFF
--- a/patches/server/0570-Missing-Entity-API.patch
+++ b/patches/server/0570-Missing-Entity-API.patch
@@ -237,10 +237,10 @@ index d53ef42e64d87790ea86d806153f047005accb9f..57952ad65c905265799ba2d8f99cc4cf
          if (angry) {
              this.setEating(false);
 diff --git a/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java b/src/main/java/net/minecraft/world/entity/animal/horse/Llama.java
-index 37da58385ecf40baeb4665a0d9c262da03c01763..4fa1a12e7e5c23e1d902b59f2a315eac64134aa8 100644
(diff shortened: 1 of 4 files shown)


patches/server/0570-Missing-Entity-API.patch
```diff
@@ -249,6 +249,25 @@ index 37da58385ecf40baeb4665a0d9c262da03c01763..4fa1a12e7e5c23e1d902b59f2a315eac
...
+     public boolean isTraderLlama() {
+@@ -316,10 +317,14 @@ public class Llama extends AbstractChestedHorse implements VariantHolder<Llama.V
+         return Llama.getDyeColor(this.getItemBySlot(EquipmentSlot.BODY));
+     }
+ 
++    // Paper start - allow api to override
++    /*
+     @Override
+     public int getMaxTemper() {
+         return 30;
```
lynxplay: could instead just return super.getMaxTemper() here for
a) less diff
b) documenting that the super method is supposed to be used.

patches/server/0570-Missing-Entity-API.patch
```diff
@@ -249,6 +249,25 @@ index 37da58385ecf40baeb4665a0d9c262da03c01763..4fa1a12e7e5c23e1d902b59f2a315eac
  
      public Llama(EntityType<? extends Llama> type, Level world) {
          super(type, world);
++        this.maxDomestication = 30; // Paper
```
lynxplay: needs identifying comment as to what is happening here.

// Paper - Missing entity api - configure max temper instead of a hardcoded value

lynxplay (changes requested):

Machine-Maker (approved):

lynxplay (approved):

PR #10825, opened 2024-05-29
Add back CommandRegisteredEvent calls

### LINKED ISSUE #10805: CommandRegisteredEvent no longer called
### Expected behavior

CommandRegisteredEvent when adding commands to the command map. Although the event is deprecated it was not intentional to break it yet.

### Observed/Actual behavior

No event call.

### Steps/models to reproduce

Register a listener with a printout and see nothing in logs when adding commands.

### Plugin and Datapack List

N/A

### Paper version

1.20.6-115

### Other

_No response_

jpenilla:
We need to delay the event call to emulate the old behavior

fixes #10805


Owen1212055:
Can this be squashed into the brig patch when merged?

jpenilla:
really it should be squashed in the mojangapi patch but it's currently before the brig api patch

PR #10826, opened 2024-05-29
Use CommandSourceStack in AsyncPlayerSendCommandsEvent

jpenilla:


Machine-Maker (approved):
I'm pretty sure this doesn't require any bytecode rewriting. The use of generics will insert casts into the compiled code, and you can cast any CommandSourceStack to the legacy type as the implementation implements both interfaces.

PR #10836, opened 2024-05-31
Prevent internal NPE on ItemStack#damage

lynxplay:
ItemStack#damage internally uses ItemStack#hurtAndBreak, which previously would call a Consumer in case the item broke. Since 1.20.5 the break game event logic however resides in said method and was using the equipment slot passed, which is null in the case of the API ItemStack#damage method.

This commit prevents the NPE by first null checking the slot. Addittionally, hurtAndBreak also now checks if the player has infinite materials, e.g. is in creative mode, to prevent damaging the item.

As such as filter is undesirable for API calls, this commit also skips this logic in case of an API invocation.

### DIFF
--- a/patches/server/0788-ItemStack-damage-API.patch
+++ b/patches/server/0788-ItemStack-damage-API.patch
@@ -11,14 +11,15 @@ the logic associated with damaging them
 public net.minecraft.world.entity.LivingEntity entityEventForEquipmentBreak(Lnet/minecraft/world/entity/EquipmentSlot;)B
 
 diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
-index acce1b72185d4687ef5a72dd4014d87149f4b092..ee0e08be375234e003675b19c5b77b9249f5c355 100644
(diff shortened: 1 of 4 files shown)


lynxplay:
Sounds fair yea 👍

patches/server/0788-ItemStack-damage-API.patch
```diff
@@ -10,8 +10,35 @@ the logic associated with damaging them
...
 
+diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
+index acce1b72185d4687ef5a72dd4014d87149f4b092..ee0e08be375234e003675b19c5b77b9249f5c355 100644
+--- a/src/main/java/net/minecraft/world/item/ItemStack.java
++++ b/src/main/java/net/minecraft/world/item/ItemStack.java
+@@ -713,7 +713,13 @@ public final class ItemStack implements DataComponentHolder {
+                     return;
+                 }
+             }
++    // Paper start - ItemStack damage API - split hurtAndBreak to skip pre-checks like creative mode
```
Machine-Maker, line 21: You have to call hurtAndBreakWithoutChecks here still

Machine-Maker (changes requested):

Machine-Maker (approved):

PR #10841, opened 2024-06-02
Only assign blockstate data if super ctor did not

lynxplay:
The CraftBlockStateMeta constructor CraftBlockStateMeta(Map) invokes its parent constructor, which itself invokes deserializeInternal, which is implemented on CraftBlockStateMeta to read the components and block entity tag from the passed map.

Field initialization happens after the call to the super constructor, meaning the current code overwrites the parsed internal data with the EMPTY defaults.

This is prevented by moving the initialization into its own code block that can null check the fields prior to defaulting their value to EMPTY.

---
Download the paperclip jar for this pull request: [paper-10841.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1563688682.zip)


Machine-Maker (approved):

TwoPointInfinity:
All my staff are having an issue related to this bug, they've asked me to bump this thread.

electronicboy:
the PR was only created this morning, have you tested the jar?

PR #10846, opened 2024-06-02
Make callback command server-only

Machine-Maker:
Closes https://github.com/PaperMC/Paper/discussions/10840

----

Prevents sending our `/callback` command in the clientbound commands packet. The callbacks should still work, there just won't be any tab completion or suggestion for the callback command.

### DIFF
--- a/patches/server/1053-Make-callback-command-server-only.patch
+++ b/patches/server/1053-Make-callback-command-server-only.patch
@@ -1,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Sun, 2 Jun 2024 15:43:52 -0700
+Subject: [PATCH] Make callback command server-only
+
+
+diff --git a/src/main/java/io/papermc/paper/command/CallbackCommand.java b/src/main/java/io/papermc/paper/command/CallbackCommand.java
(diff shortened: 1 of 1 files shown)


patches/server/1053-Make-callback-command-server-only.patch
```diff
@@ -0,0 +1,47 @@
...
+index 0126906e2afc8dd525f27a0c5e82116075c9d352..9750bac6faf45d15a3ab96a3083afc6899e4fb78 100644
+--- a/src/main/java/net/minecraft/commands/Commands.java
++++ b/src/main/java/net/minecraft/commands/Commands.java
+@@ -547,7 +547,8 @@ public class Commands {
+             // Paper end - Brigadier API
+             if ( !org.spigotmc.SpigotConfig.sendNamespaced && commandnode2.getName().contains( ":" ) ) continue; // Spigot
+ 
+-            if (commandnode2.canUse(source)) {
++            boolean serverOnly = commandnode2.wrappedCached instanceof io.papermc.paper.command.brigadier.bukkit.BukkitCommandNode bukkitNode && bukkitNode.getBukkitCommand() instanceof io.papermc.paper.command.ServerOnlyCommand; // Paper
++            if (commandnode2.canUse(source) && !serverOnly) { // Paper
```
electronicboy: cheaper check first?

Owen1212055:
Future plans, but if this ever should be added what instead should be done is that the callback command is made into a brigadier node. Then, simply set the client side node to an empty node that returns false on canUse. This will cause it not to be sent to clients.

Machine-Maker:
Doesn't need to happen anymore, no more callback command, no rebase required.

PR #10847, opened 2024-06-02
1.20.1 Nur Spigot Started Here

NekoMonci12:
The Beginning of Nur Spigot


Malfrador:
I know you just PR'ed to the wrong repo.
But if you are planning on making a Paper fork, its highly recommend to use Paperweight for this instead of just forking the Github repository. It will make development and future maintenance of your fork much much easier.

See [here ](https://github.com/PaperMC/paperweight-examples) for an example on how to create a fork with Paperweight. You can also look at other Paper forks, like Purpur or Pufferfish - they use paperweight too.

PR #10852, opened 2024-06-03
delegate ItemStack

Machine-Maker:
This essentially removes API-only itemstacks without changing ItemStack to an interface, that will come later.

Just adds a field in api-ItemStack that should always be a CraftItemStack which the api-ItemStack will wrap around. CraftItemStack **has** to override any methods on api-ItemStack that use this `craftDelegate` to avoid recursion loops.

---
Download the paperclip jar for this pull request: [paper-10852.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1609690518.zip)

### DIFF
--- a/patches/api/0478-Proxy-ItemStack-to-CraftItemStack.patch
+++ b/patches/api/0478-Proxy-ItemStack-to-CraftItemStack.patch
@@ -1,0 +1,472 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jake Potrebic <jake.m.potrebic@gmail.com>
+Date: Tue, 14 May 2024 11:57:51 -0700
+Subject: [PATCH] Proxy ItemStack to CraftItemStack
+
+
+diff --git a/src/main/java/org/bukkit/UnsafeValues.java b/src/main/java/org/bukkit/UnsafeValues.java
+index 7332034bb1753f48f7904dafab1ef4b3ee117ea3..11bccf6c27b8b2a2ed558a1375a3b3adc7952076 100644
+--- a/src/main/java/org/bukkit/UnsafeValues.java
++++ b/src/main/java/org/bukkit/UnsafeValues.java
+@@ -277,4 +277,6 @@ public interface UnsafeValues {
+     @NotNull java.util.List<net.kyori.adventure.text.Component> computeTooltipLines(@NotNull ItemStack itemStack, @NotNull io.papermc.paper.inventory.tooltip.TooltipContext tooltipContext, @Nullable org.bukkit.entity.Player player); // Paper - expose itemstack tooltip lines
+ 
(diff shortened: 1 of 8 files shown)


patches/api/0484-delegate-ItemStack.patch
```diff
@@ -0,0 +1,382 @@
...
+ 
+     @Utility
+     protected ItemStack() {}
+@@ -86,8 +84,7 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
+     @Deprecated
+     public ItemStack(@NotNull final Material type, final int amount, final short damage, @Nullable final Byte data) {
+         Preconditions.checkArgument(type != null, "Material cannot be null");
+-        this.type = type;
+-        this.amount = amount;
++        this.craftDelegate = Bukkit.getUnsafe().createStack(type, amount); // Paper - delegate
```
masmc05: Why not use <https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemType.html#createItemStack(int)> and make CraftItemType always make CraftItemStack?

Machine-Maker: yep, that is better. I still haven't fully accepted that ItemType and BlockType actually exist. its amazing.

Machine-Maker:
It's possible that ItemStack#equals and ItemStack#hashCode should also directly delegate to a craftitemstack. I think that's probably most correct.

Lulu13022002 (commented):
This doesn't handle empty itemstack (ItemStack#empty) which are still full api stack without server side equivalent and will throw a NPE for most methods.

Machine-Maker:
> This doesn't handle empty itemstack (ItemStack#empty) which are still full api stack without server side equivalent and will throw a NPE for most methods.

I think I should just be able to change empty to return `ItemStack.create(Material.AIR)`. It'd have a stack size of 1 however, should I instead add a bridge method to create a truly empty stack? just by returning a copy of nms-ItemStack.EMPTY?

Lulu13022002:
Another issue is that asNMSCopy doesn't accept null itemstack anymore so doing /clear for example throw a NPE too via the PrepareItemCraftEvent.

patches/api/0478-Proxy-ItemStack-to-CraftItemStack.patch
```diff
@@ -0,0 +1,437 @@
...
++    // Paper start - add static factory methods
++    /**
++     * Creates an itemstack with the specified item type and a count of 1.
++     *
++     * @param type the item type to use
++     * @return a new itemstack
++     * @throws IllegalArgumentException if the Material provided is not an item ({@link Material#isItem()})
++     */
++    @org.jetbrains.annotations.Contract(value = "_ -> new", pure = true)
++    public static @NotNull ItemStack create(final @NotNull Material type) {
```
kennytv: of, also below

lynxplay (approved):

Owen1212055 (approved):

kennytv (approved):

PR #10877, opened 2024-06-13
Configurable damage tick when blocking with shield

lynxplay:
A long standing bug in spigot and its derivatives was the fact that players taking damage while blocking with a shield would not receive invulnerability, while they do in vanilla.

This enabled the pvp technique of disabling a shield and immediately attacking again to knock a player into the air.
While upstream fixed this and properly aligned itself with vanilla damage logic (in this specific case) changing such long standing behaviour has some downsides.

To allow players used to this specific bug to still use it, this patch introduces a configuration option to re-introduce said bug. As there is no easy way to *only* re-add this bug, the option is found in the unsupported section as it may introduce other damage related disparity from vanilla.

---
Download the paperclip jar for this pull request: [paper-10877.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1605148982.zip)


Machine-Maker (approved):

BurkenDev:
Please release 1.21 instead of doing these patches on old versions.

Issue #10873, opened 2024-06-12
ItemStack#removeEnchantments makes the item unenchantable


TothAdam0012:
### Expected behavior

Calling ItemStack#removeEnchantments should remove all current enchantments, making it possible to re-enchant the item with the enchanting table or commands.

### Observed/Actual behavior

After ItemStack#removeEnchantments is called on an item it can't be re-enchanted again. Putting it in the enchanting table doesn't show options and the /enchant command says the item has been enchanted but there are no enchantments present.
What I can see is that a diamond chestplate has 8 components by default and calling this method reduces that to 7.

### Steps/models to reproduce

I created an empty plugin with the following onEnable (and registered the ench command in the plugin.yml)

```java
    @Override
    public void onEnable() {
        PluginCommand ench = getCommand("ench");
        if(ench != null) {
            ench.setExecutor((sender, command, label, arg) -> {
                if(sender instanceof Player player) {
                    player.getInventory().getItemInMainHand().removeEnchantments();
                }
                return true;
            });
        }
    }
```

Holding an item and using the /ench command makes the item unenchantable, even if it didn't have enchantments.

### Plugin and Datapack List

Only the small plugin i wrote to test this.

### Paper version

```
This server is running Paper version 1.20.6-138-master@906df69 (2024-06-10T06:32:15Z) (Implementing API version 1.20.6-R0.1-SNAPSHOT)
```

### Other

There is a workaround: if I get the enchantments first and remove the 1 by 1 after removing the last one the item remains enchantable. This issue is only present if ItemStack#removeEnchantments is called.

Machine-Maker:
The removeEnchantments method needs to set the map on CraftMetaItem to null instead of just clearing. Should be a 1 line fix

PR #10881, opened 2024-06-14
Fix to issue 10873: Corrected removing all enchantments from ItemStack

GhastCraftHD:
Fixed issue [10873](https://github.com/PaperMC/Paper/issues/10873)
Set the enchantment map to null rather just clearing it


lynxplay:
Master is frozen right now, we will get to this issue once the 1.21 update is done.
We can backport it too :+1:

PR #10901, opened 2024-06-18
Correct duplicate attribute check

lynxplay:
Brackets got lost during the update, leading to an incorrect precondition call.


lynxplay:
See https://github.com/PaperMC/Paper/blob/ver/1.20.6/patches/server/0931-Fix-CraftMetaItem-getAttributeModifier-duplication-c.patch#L16 for reference

kennytv (approved):

PR #10911, opened 2024-06-18
Call PlayerLaunchProjectileEvent for wind charge

Lulu13022002:


patches/server/0215-PlayerLaunchProjectileEvent.patch
```diff
@@ -317,3 +317,47 @@ index 2d34f206b4b02d5f2ee35101233afc6b4f58a579..f1b2d388a1a40a1d909a2e726f32d6c1
...
                                  // CraftBukkit start - SPIGOT-5458 also need in this branch :(
+diff --git a/src/main/java/net/minecraft/world/item/WindChargeItem.java b/src/main/java/net/minecraft/world/item/WindChargeItem.java
+index 6222e867bb959fab05bcd8f2114ab00b26847c3e..bdf0f9177a65e70147311e71784c796f7edaf2af 100644
+--- a/src/main/java/net/minecraft/world/item/WindChargeItem.java
++++ b/src/main/java/net/minecraft/world/item/WindChargeItem.java
+@@ -24,10 +24,26 @@ public class WindChargeItem extends Item implements ProjectileItem {
+ 
+     @Override
+     public InteractionResultHolder<ItemStack> use(Level world, Player user, InteractionHand hand) {
++        ItemStack itemStack = user.getItemInHand(hand); // Paper - moved from below
```
lynxplay: // Paper - PlayerLaunchProjectileEvent - moved from below.

patches/server/0215-PlayerLaunchProjectileEvent.patch
```diff
@@ -317,3 +317,47 @@ index 2d34f206b4b02d5f2ee35101233afc6b4f58a579..f1b2d388a1a40a1d909a2e726f32d6c1
...
+         if (!world.isClientSide()) {
+             WindCharge windCharge = new WindCharge(user, world, user.position().x(), user.getEyePosition().y(), user.position().z());
+             windCharge.shootFromRotation(user, user.getXRot(), user.getYRot(), 0.0F, 1.5F, 1.0F);
+-            world.addFreshEntity(windCharge);
++            // Paper start - PlayerLaunchProjectileEvent
++            com.destroystokyo.paper.event.player.PlayerLaunchProjectileEvent event = new com.destroystokyo.paper.event.player.PlayerLaunchProjectileEvent((org.bukkit.entity.Player) user.getBukkitEntity(), org.bukkit.craftbukkit.inventory.CraftItemStack.asCraftMirror(itemStack), (org.bukkit.entity.Projectile) windCharge.getBukkitEntity());
++            if (!event.callEvent() || !world.addFreshEntity(windCharge)) {
++                user.containerMenu.sendAllDataToRemote();
++                if (user instanceof net.minecraft.server.level.ServerPlayer player) {
++                    player.connection.send(new net.minecraft.network.protocol.game.ClientboundCooldownPacket(this, 0));
```
lynxplay: Maybe worth some comment as to why this is needed?
given that iirc this is the first time we manually sent it in the codebase.

lynxplay (commented):
not tested but diff looks good 👍 will test this weekend.

patches/server/0215-PlayerLaunchProjectileEvent.patch
Lulu13022002: Yes similar thing should be done for the enderpearl too

lynxplay (approved):

electronicboy (approved):

PR #10918, opened 2024-06-19
Add Event to control who hears the dragon death sound

Lurkrul:
Allows plugins to add/remove players from who will hear the dragon death sound.

While this has a niche use case, it will be helpful in those cases.

Servers that have their own customized enderdragon can use the event to change it to where all damage dealers will hear the noise.
This can also be used in the case where a server wants to allow players to ignore the dragon death sound.

### DIFF
--- a/patches/api/0480-Add-event-so-you-can-modify-what-players-will-hear-t.patch
+++ b/patches/api/0480-Add-event-so-you-can-modify-what-players-will-hear-t.patch
@@ -27,65 +27,11 @@ index 0000000000000000000000000000000000000000..c9f821dd291c486c56597cc32135f481
 +public class DragonPrepareDeathSoundTargetsEvent extends EntityEvent {
 +
 +    private static final HandlerList HANDLER_LIST = new HandlerList();
-+    private Set<Player> playerList;
++    public Set<Player> playerList;
 +
 +    public DragonPrepareDeathSoundTargetsEvent(final @NotNull EnderDragon enderDragon, Collection<? extends Player> players) {
 +        super(enderDragon);
 +        playerList = new HashSet<>(players);
-+    }
-+
-+    /**
-+     * Gets the list of {@link Player}s that are going to hear the sound.
-+     *
-+     * @return Collection of Players
-+     */
-+    public Set<Player> getPlayerList() {
-+        return playerList;
-+    }
-+
-+    /**
-+     * Sets the player list to a Collection of defined {@link Player}s
-+     *
-+     * @param newPlayerList The new collection of players to hear the sound.
-+     */
-+    public void setPlayerList(Set<Player> newPlayerList) {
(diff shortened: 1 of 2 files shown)


patches/server/1028-add-event-so-you-can-modify-what-players-hear-the-dr.patch
```diff
@@ -0,0 +1,53 @@
...
+index bd6fee3e3ad9116802ff8bb57bfa741b881c4057..f225942db664b8ce6f97dd907f39daca5b85fc9b 100644
+--- a/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
++++ b/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+@@ -2204,4 +2200,21 @@ public class CraftEventFactory {
+         return event;
+     }
+     // Paper end - add EntityFertilizeEggEvent
++
++    // Paper start - add DragonPrepareDeathSoundTargetsEvent
++
```
yannicklamprecht:
```suggestion
```

patches/api/0480-Add-event-so-you-can-modify-what-players-will-hear-t.patch
```diff
@@ -0,0 +1,111 @@
...
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++    private Set<Player> playerList;
++
++    public DragonPrepareDeathSoundTargetsEvent(final @NotNull EnderDragon enderDragon, Collection<? extends Player> players) {
++        super(enderDragon);
++        playerList = new HashSet<>(players);
++    }
++
++    /**
```
yannicklamprecht, line 37: Either expose the collection directly and remove the other modifying accessors or return a copy of that Set and drop the setPlayerList(Set<....).  Having both will just unnecessarily bloat the event with methods.

yannicklamprecht (changes requested):

patches/api/0480-Add-event-so-you-can-modify-what-players-will-hear-t.patch
Lurkrul: removed all of the functions and changed access of the set to publci

patches/server/1028-add-event-so-you-can-modify-what-players-hear-the-dr.patch
Lurkrul: removed whitespace

Machine-Maker (changes requested):
This event should probably be "generalized" a bit. I think a more general event is probably better. You won't get the "dragon" context, but smth that fires for each player that receives a level event packet of any type might be useful. We have a type that represents level events (Effect). I'm not sure if that event would fire *too* often though.

PR #10930, opened 2024-06-20
Expose LivingEntity#canUseSlot

SoSeDiK:
- Adds `LivingEntity#canUseEquipmentSlot(EquipmentSlot)` to check whether an entity can use that slot.
- Adds missing `@throws` annotations to PlayerInventory.

Follow-up/related to https://github.com/PaperMC/Paper/pull/10822.

### DIFF
--- a/patches/api/0471-Fix-equipment-slot-and-group-API.patch
+++ b/patches/api/0471-Fix-equipment-slot-and-group-API.patch
@@ -3,7 +3,11 @@ From: Jake Potrebic <jake.m.potrebic@gmail.com>
 Date: Wed, 22 May 2024 10:00:19 -0700
 Subject: [PATCH] Fix equipment slot and group API
 
-was missing the 'body' slot group
+Adds the following:
+- Add missing 'body' slot group
+- Expose LivingEntity#canUseSlot
+
+Co-authored-by: SoSeDiK <mrsosedik@gmail.com>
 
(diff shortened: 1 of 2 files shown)


patches/api/0471-Fix-equipment-slot-and-group-API.patch
```diff
@@ -17,33 +21,55 @@ index 90f18f355a6a236a7e4273cc1258e7c8034b8276..a0750d457a9cd7cf3ab9c45bb28ce5a2
...
+--- a/src/main/java/org/bukkit/entity/LivingEntity.java
++++ b/src/main/java/org/bukkit/entity/LivingEntity.java
+@@ -1447,4 +1447,14 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
+      */
+     void setBodyYaw(float bodyYaw);
+     // Paper end - body yaw API
++
++    // Paper start - Expose canUseSlot
++    /**
++     * Checks whether this entity can use the equipment slot
```
kennytv: An example or two would be good, like specifically calling out the body slot

kennytv (approved):

patches/api/0471-Fix-equipment-slot-and-group-API.patch
SoSeDiK: Added example linking to body slot, and body slot has javadoc explaining that only certain entities can have it

PR #10935, opened 2024-06-20
Fixes exception log messages getting dropped during bootstrap

Machine-Maker:


emilyy-dev:
wouldn't adding a shutdown hook be a more appropriate solution than playing hide and seek with existing exit points?

electronicboy:
shutdown hooks run in order of registration, so, we could register one, but, then we'd need to remove it, otherwise upstream shutdown thread won't be able to log the shutdown process

electronicboy (approved):

PR #10961, opened 2024-06-22
Leashable API

Lulu13022002:
Expand a bit the api to support leashable boat.


patches/api/0480-Leashable-API.patch
```diff
@@ -0,0 +1,143 @@
...
++    boolean isLeashed();
++
++    /**
++     * Gets the entity that is currently leading this entity.
++     *
++     * @return the entity holding the leash
++     * @throws IllegalStateException if not currently leashed
++     */
++    @NotNull
++    Entity getLeashHolder() throws IllegalStateException;
```
lynxplay: What is the gain throwing ISE over null?

Lulu13022002: This is to keep the compatibility with the old method.

lynxplay:
Do you think it is worth merging with `Expand EntityUnleashEvent` into a general "improved leash API" patch?
Don't have strong feelings either way, just found it while reviewing.

Lulu13022002:
I would rather not. There's already a Shearable and Bucketable patch that exists which seems cleaner.

lynxplay (approved):

Machine-Maker (approved):

PR #10966, opened 2024-06-23
fix SoundEffects only to players diff 

### LINKED ISSUE #10940: Fire is cleared if hurt player is false
### Expected behavior

Fire not being cleared if the target.hurt(damagesource, f3) returns false

### Observed/Actual behavior

Fire is cleared if target.hurt(damagesource, f3) returns false

### Steps/models to reproduce

1.20.6
https://github.com/PaperMC/Paper/blob/ver/1.20.6/patches/server/0157-Send-attack-SoundEffects-only-to-players-who-can-see.patch#L54

1.21
https://github.com/PaperMC/Paper/blob/master/patches/server/0156-Send-attack-SoundEffects-only-to-players-who-can-see.patch#L73-L83

flag4 was a different variable in 1.20.6, and that clearFire method call shouldn't even exist now. 

### Plugin and Datapack List

None

### Paper version

Latest 1.21

### Other

_No response_

scarfacered:
closes https://github.com/PaperMC/Paper/issues/10940


kennytv (approved):
Thanks!

PR #10986, opened 2024-06-25
Change item/xp merge and entity tracking range defaults

kennytv:
Restore vanilla defaults for the merge radius, increase the tracking range values as the current ones are absolute ass. I also increased `raiderActivationRange`, as raiders like to get stuck just out of reach and become unfindable.

Waiting on feedback on possibly changing other config values. No Paper comments because it's just a Spigot file and Paper go zoom

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -5091,7 +5091,7 @@ index 192b6fbd34b9b90112f869ae6e367ab9ba5a5906..08b0ca7b68bf238366f4d6904478852e
          for ( Method method : clazz.getDeclaredMethods() )
          {
 diff --git a/src/main/java/org/spigotmc/SpigotWorldConfig.java b/src/main/java/org/spigotmc/SpigotWorldConfig.java
-index a6d2ce801b236b046b94913bccf7eccfc561f35a..b244b65799d9c082b8b1639bad15727442e63168 100644
+index a6d2ce801b236b046b94913bccf7eccfc561f35a..0149d9014d0c6eab772297e7caeb9e7272574306 100644
 --- a/src/main/java/org/spigotmc/SpigotWorldConfig.java
 +++ b/src/main/java/org/spigotmc/SpigotWorldConfig.java
 @@ -58,8 +58,14 @@ public class SpigotWorldConfig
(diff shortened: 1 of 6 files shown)


patches/server/0005-Paper-config-files.patch
```diff
@@ -5111,6 +5111,47 @@ index a6d2ce801b236b046b94913bccf7eccfc561f35a..b244b65799d9c082b8b1639bad157274
...
+     }
+ 
+-    public int playerTrackingRange = 48;
+-    public int animalTrackingRange = 48;
+-    public int monsterTrackingRange = 48;
+-    public int miscTrackingRange = 32;
++    public int playerTrackingRange = 128;
++    public int animalTrackingRange = 96;
++    public int monsterTrackingRange = 96;
++    public int miscTrackingRange = 96;
```
masmc05, line 5151: Why can't tracking ranges be -1 by default, which will result in using vanilla values?

Vanilla has various values for each mob, setting a universal value per category already will result in behaviour differences without any possibility to restore it

EterNityCH:
Overall a step in the right direction as most of the config values are outdated for modern Minecraft. Considering majority of the server are small, this will restore some Vanilla experience and addresses some common question people asked (like cannot see players too far out).

patches/server/0005-Paper-config-files.patch
kennytv: The purpose of this pr is to improve defaults, not to necessarily make them equal to vanilla, and those are visual only anyways

Leguan16:
Does this need changes in the vanilla breaking changes docs pr?

kennytv:
Yess, you can remove the entries for the two merge radius values

patches/server/0005-Paper-config-files.patch
```diff
@@ -5111,6 +5111,47 @@ index a6d2ce801b236b046b94913bccf7eccfc561f35a..b244b65799d9c082b8b1639bad157274
...
+-        this.itemMerge = this.getDouble("merge-radius.item", 2.5 );
++        this.itemMerge = this.getDouble("merge-radius.item", 0.5 );
+         this.log( "Item Merge Radius: " + this.itemMerge );
+     }
+ 
+     public double expMerge;
+     private void expMerge()
+     {
+-        this.expMerge = this.getDouble("merge-radius.exp", 3.0 );
++        this.expMerge = this.getDouble("merge-radius.exp", 0.5 );
```
Machine-Maker: It was pointed out that the vanilla value for this is actually to disable it completely with `-1`. There are 2 exp merge systems in the server, this only controls the custom one spigot added, and doesn't mess with vanilla's own merge system. So disabling this completely will result in vanilla xp behavior.

EterNityCH: Oh I was not aware of this and yeah that would certainly be a more favorable result.

EterNityCH:
The value is perfectly ok for small SMP which is what majority of server owners are according to bstat, the value does not affect existing servers, not to mention, it can also be changed at owners discretion.

PR #10990, opened 2024-06-26
Skip AI during inactive ticks for non-aware mobs

froobynooby:
Adds an awareness check to the inactiveTick of mobs (non-aware mobs aren't meant to have their AI ticked)


Owen1212055:
Does this check need to be expanded to for example Villager?

PR #11000, opened 2024-06-29
Optimise custom map rendering

barnabwhy:
- Remove floating point math from `getDistance` in `org.bukkit.map.MapPallete`
- ~~Use bitwise & for bounds checks in `obc.map.CraftMapCanvas` to reduce comparisons~~ Removed
- Reduce size of image being drawn in `drawImage` if it is larger than the available size.
- Use `System.arraycopy` to avoid iterating over the colors in the image.
- Swap x and y loops in `CraftMapRenderer.render` to be more cache-friendly

Saw a ~30% performance improvement in `obc.map.CraftMapView.render()`

Tested with a world containing many custom maps created using the `ImageOnMap` plugin, modified to remove [this optimisation](https://github.com/zDevelopers/ImageOnMap/blob/indev/src/main/java/fr/moribus/imageonmap/image/Renderer.java#L100) so that it renders every tick.
Before: https://spark.lucko.me/NVdWKmSlFr
After: https://spark.lucko.me/j6AJiFG9pD


patches/server/1029-Reduce-work-done-in-CraftMapCanvas.drawImage-by-limi.patch
```diff
@@ -0,0 +1,103 @@
...
+diff --git a/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java b/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
+index ff59f759669620795ef355c988b664bdcda39f52..59f61beb0f179d543fa6adab0989e676a06ab802 100644
+--- a/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
++++ b/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
+@@ -1,8 +1,10 @@
+ package org.bukkit.craftbukkit.map;
+ 
+ import com.google.common.base.Preconditions;
+ import java.awt.Color;
++import java.awt.Graphics2D;
```
Leguan16: Usually imports should generally be avoided in non paper classes to reduce diff ;)

patches/server/1029-Reduce-work-done-in-CraftMapCanvas.drawImage-by-limi.patch
```diff
@@ -0,0 +1,92 @@
...
+index ff59f759669620795ef355c988b664bdcda39f52..54a3c1cea840f45385f1102988f8d2c8c415127b 100644
+--- a/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
++++ b/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
+@@ -59,8 +59,10 @@ public class CraftMapCanvas implements MapCanvas {
+ 
+     @Override
+     public void setPixel(int x, int y, byte color) {
+-        if (x < 0 || y < 0 || x >= 128 || y >= 128)
++        // Paper start - Optimise bounds check using bitwise operations
++        if ((x & -128) != 0 || (y & -128) != 0)
```
kennytv: Are you sure this is needed/did you benchmark this change? I can't imagine this will scrape off much even beyond a nanosecond, such simple integer comparisons are generally free, but this does make readability a lot worse

barnabwhy: I don't recall it made much difference to be honest, I can remove it if you feel that makes more sense.

kennytv:
Thanks, I'll get to it some time in the following week(s)

barnabwhy:
rebased

kennytv (approved):
Both distance and drawing improvements look great - thank you! Sorry it took so long to get a review out

PR #11004, opened 2024-07-01
Fix summon_entity effect attempting to add incorrect entity

granny:
Fixes an error that appears when a registered enchantment uses the summon_entity effect. Original issue was reported here: https://github.com/PurpurMC/Purpur/issues/1545

Should probably be upstreamed into CraftBukkit by someone - personally cba'd to deal with their contribution requirements.


kennytv (approved):
Thanks

PR #11033, opened 2024-07-07
Item stack name tooltip API

radstevee:
This adds an API for getting whether the current item stack that the player is holding is showing its name as a tooltip. It's important to note that this behavior can be configured via the accessbility settings, under notification time. This is still very useful for server owners, e.g. when you have a UI Overlay in your action bar, which is slightly shifted downwards using fonts, this can collide with the item's name tooltip. But instead, with this API, you can detect if it's shown and then raise the overlay a bit. Here is an example of this in action (not made by me):

https://github.com/PaperMC/Paper/assets/96977790/eaf40378-9e7f-494c-b96a-48596840c462

### DIFF
--- a/patches/api/0480-Item-stack-name-tooltip-API.patch
+++ b/patches/api/0480-Item-stack-name-tooltip-API.patch
@@ -5,21 +5,30 @@ Subject: [PATCH] Item stack name tooltip API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 7c56182acaf827f4b1a986a61cea8e9960604c98..f7d0d7937e86a4a1b51fc2105f072f40ae623396 100644
+index 7c56182acaf827f4b1a986a61cea8e9960604c98..9da64bc368b1eac55fad336f58619509542dd5e1 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
-@@ -3855,6 +3855,14 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
+@@ -3855,6 +3855,23 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      boolean isChunkSent(long chunkKey);
      // Paper end
  
++    // Paper start - Item stack name tooltip API
 +    /**
-+     * Checks whether the currently held item is displaying the item name tooltip on the client.
++     * Checks whether the currently held item is assumed to be displaying the item name tooltip on the client.
 +     * However, this behavior can be changed via the Accessibility Settings, which can lead to inaccuracies.
 +     * Keep in mind that this assumes default settings and is prone to break.
 +     * @return true if it is showing with default settings, false if not.
 +     */
-+    boolean isItemNameTooltipShown(); // Paper - Item stack name tooltip API
++    boolean isItemNameTooltipEstimatedToBeVisible();
 +
++    /**
++     * Gets the last timestamp when the player has changed their selected item stack in their hotbar.
++     * This time is measured using {@link net.minecraft.Util#getMillis}.
++     * @return The last timestamp, or zero, if the player hasn't changed selected item stacks.
++     */
++    long getLastItemStackChangeTime();
++
++    // Paper end - Item stack name tooltip API
      @NotNull
      @Override
      Spigot spigot();
--- a/patches/api/0480-Item-stack-name-tooltip-API.patch
+++ b/patches/api/0480-Item-stack-name-tooltip-API.patch
@@ -5,27 +5,20 @@ Subject: [PATCH] Item stack name tooltip API
 
 
 diff --git a/src/main/java/org/bukkit/entity/Player.java b/src/main/java/org/bukkit/entity/Player.java
-index 7c56182acaf827f4b1a986a61cea8e9960604c98..5899dcdcbb95c2d3da3869ca202270ce7b106336 100644
+index 7c56182acaf827f4b1a986a61cea8e9960604c98..f7d0d7937e86a4a1b51fc2105f072f40ae623396 100644
 --- a/src/main/java/org/bukkit/entity/Player.java
 +++ b/src/main/java/org/bukkit/entity/Player.java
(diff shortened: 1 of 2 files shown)


patches/server/1036-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,68 @@
...
+index d01b45a48d412e3cb591acee101730704574448a..846c8d8577515e7641b264c2c0790d6d4b770971 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -3554,4 +3554,25 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
+         ((ca.spottedleaf.moonrise.patches.chunk_system.player.ChunkSystemServerPlayer)this.getHandle())
+             .moonrise$getViewDistanceHolder().setSendViewDistance(viewDistance);
+     }
++
++    // Paper start - Item stack name tooltip API
++    private long lastItemStackChange = 0;
```
lynxplay: I'd move this logic to ServerPlayer, given that is  where this state is maintained.

CraftPlayer can then access/set on API calls instead of this weird bi-directional dependency.

patches/server/1036-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,68 @@
...
++    private ItemStack currentlySelectedItemStack = ItemStack.EMPTY; // Paper - Item stack name tooltip API
+ 
+     // Paper start - rewrite chunk system
+     private ca.spottedleaf.moonrise.patches.chunk_system.player.RegionizedPlayerChunkLoader.PlayerChunkLoaderData chunkLoader;
+@@ -805,6 +806,19 @@ public class ServerPlayer extends net.minecraft.world.entity.player.Player imple
+         this.trackEnteredOrExitedLavaOnVehicle();
+         this.updatePlayerAttributes();
+         this.advancements.flushDirty(this);
++
++        // Paper start - Item stack name tooltip API
```
lynxplay, line 16: Please note that this code is taken from `net.minecraft.client.gui.Gui#tick` and adopted to a non-tick behaviour
so updates can easily find where this logic is if values change.

patches/api/0480-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,32 @@
...
+     boolean isChunkSent(long chunkKey);
+     // Paper end
+ 
++    // Paper start - Item stack name tooltip API
++    /**
++     * Checks whether the currently held item is displaying the item name tooltip on the client.
++     * However this behavior can be changed via the Accessibility Settings, which can lead to inaccuracies.
++     * @return true if it is showing with default settings, false if not
++     */
++    boolean isItemNameTooltipShown();
```
lynxplay: I am a bit iffy on the method name.
As you correctly highlighted, this method may be completely inaccurate if the client changes their notificationDisplayTime.

Idk if we can get the fact that this method is more a guess/approximation into the method name.

patches/api/0480-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,32 @@
...
++
++    /**
++     * Gets the last time where the selected item stack in the hotbar has been changed.
```
lynxplay: the last time ~where~ when

patches/api/0480-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,32 @@
...
++     * @return The last time, in milliseconds, if it exists.
```
lynxplay: "if if exists" is misleading, it will yield 0 if it doesn't exist.

lynxplay (changes requested):

lynxplay:
Welcome to paper 🎉 Thank you for your first PR!

patches/server/1036-Item-stack-name-tooltip-API.patch
```diff
@@ -0,0 +1,94 @@
...
+ }
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+index d01b45a48d412e3cb591acee101730704574448a..156f64452a766f235c8a3d7c75e1129fbc260aef 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+@@ -161,7 +161,6 @@ import org.bukkit.craftbukkit.map.CraftMapView;
+ import org.bukkit.craftbukkit.map.RenderData;
+ import org.bukkit.craftbukkit.potion.CraftPotionEffectType;
+ import org.bukkit.craftbukkit.potion.CraftPotionUtil;
+-import org.bukkit.craftbukkit.profile.CraftPlayerProfile;
```
lynxplay: Get rid of the removed imports here please.

lynxplay (changes requested):

patches/server/1036-Item-stack-name-tooltip-API.patch
radstevee: intellij moment

lynxplay (commented):
Needs general re-evaluation on its usefulness/api usage.

Given the server has 0 authority over this, the PR adds a nice utility (certainly) but said utility cannot be consistent (as evident by the javadocs).

Needs more input of how worth this utility is.

Malfrador:
As someone who does the things mentioned in the PR with the action bar, this is certainly very useful. 

And it's clearly mentioned in the Javadocs that the server is limited in predicting it. 
Maintenance cost should be very low too. 

However maybe just exposing itemStackChangeTime directly would make more sense? Then that value could also be used for other things (though I am not sure what those could be)

NonSwag:
I think having `isItemNameTooltipEstimatedToBeVisible` is too unpredictable because of the already mentioned fact that it is entirely controller by the client and the server doesn't know the tooltip visibility at all
I would suggest only exposing the time in **ticks** like vanilla already has and plugins can use it like they want

PR #11066, opened 2024-07-13
Fix HelpCommand searching

notTamion:
The Help command used to throw an StringIndexOutOfBoundsException when the command is:
`/help /`
Also fixes the Help command sending an empty help list to the player incase only commands the player doesn't have permissions for are found


Dominilk01:
issu: https://discord.com/channels/289587909051416579/289587909051416579/1261702464986550464

patches/api/0480-Fix-HelpCommand-searching.patch
```diff
@@ -0,0 +1,41 @@
...
+--- a/src/main/java/org/bukkit/help/IndexHelpTopic.java
++++ b/src/main/java/org/bukkit/help/IndexHelpTopic.java
+@@ -49,10 +49,13 @@ public class IndexHelpTopic extends HelpTopic {
+         if (sender instanceof ConsoleCommandSender) {
+             return true;
+         }
+-        if (permission == null) {
+-            return true;
+-        }
+-        return sender.hasPermission(permission);
```
lynxplay: I am unsure how clever it is to not check the permission here at all.
While nothing currently sets a permission, it is technically still a valid thing to do unless I am missing something?

The HelpMap is public API, so are the help topics.
Plugins making use of such a permission flag should be respected.

notTamion: I find it weird Bukkit has this IndexHelpTopic in the first place. What is the advantage against just using a List?

lynxplay: Yea sadly we cannot crystal ball into the minds of bukkit developers xD

lynxplay (approved):

kennytv (approved):

PR #11068, opened 2024-07-13
Don't require author and title to be set on signed book meta

### LINKED ISSUE #11067: Written books with pages require author/title
### Expected behavior

Before https://github.com/PaperMC/paper/commit/4d20922b79f5b3976e240932d087357a3808fde4 creating a book with pages but without an author or title worked without issues

### Observed/Actual behavior

After that commit when creating a book that has pages but without an author or title it can no longer encode the component correctly
https://paste.enginehub.org/uQZTHfDcR

### Steps/models to reproduce

```java
ItemStack book = new ItemStack(Material.WRITTEN_BOOK);
BookMeta meta = (BookMeta) book.getItemMeta();
meta.addPages(Component.text("awawawa"));
book.setItemMeta(meta);
e.getPlayer().getInventory().addItem(book);
```

### Plugin and Datapack List

n/a

### Paper version

Paper version 1.21-62-master@5a503d7

### Other

It's a bit weird, because both `setTitle()` and `setAuthor()` are Nullable, and when the book doesn't have any pages it's fine to not set them, but as soon as you add pages to the book those become required

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/11067

Fix a regression in https://github.com/PaperMC/Paper/commit/4d20922b79f5b3976e240932d087357a3808fde4 and make pages optional too while still keeping the author and title.

### DIFF
--- a/patches/server/0970-General-ItemMeta-fixes.patch
+++ b/patches/server/0970-General-ItemMeta-fixes.patch
@@ -12,7 +12,7 @@ public net/minecraft/world/level/block/entity/BlockEntity saveId(Lnet/minecraft/
 Co-authored-by: GhastCraftHD <julius.gruenberg@leghast.de>
 
 diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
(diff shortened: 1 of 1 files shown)


lynxplay (approved):
Cannot test, but diff looks good.

lynxplay (approved):
Still looks good after testing

kennytv (approved):

PR #11082, opened 2024-07-15
Add CrafterCraftItemEvent

### LINKED ISSUE #11029: Crafters lead to duplication with custom crafting plugins
### Expected behavior

I would expect a CrafterCraftEvent to exist in order to hook into crafters. Failing this, I would assume there is a way to disable certain CraftingRecipes from working in crafters.

### Observed/Actual behavior

There is no such API and it is not possible to blacklist recipes from working in crafters.

### Steps/models to reproduce

Use a custom recipe plugin like EcoItems, CustomCrafting, or any plugin with custom item crafting / support for crafting recipes with stacked items (e.g. EcoItems, EcoArmor, etc.) and you will be able able to craft items without sufficient items.

### Plugin and Datapack List

eco and EcoItems. Both are open-source, the crux of the crafting logic for eco can be found [here](https://github.com/Auxilor/eco/tree/master/eco-core/core-plugin/src/main/kotlin/com/willfp/eco/internal/spigot/recipes)

### Paper version

1.21-38-master@1f5db50 (2024-06-27T20:15:58Z) (Implementing API version 1.21-R0.1-SNAPSHOT)

### Other

There doesn't seem to be any API in the works or any open PRs for this. If this is not planned or if it's not considered essential before 1.21 is considered stable, please tell me what would be considered as a PR so I can make it.

lynxplay:
Adds an event fired when the crafter crafts an item. The event only exposes the recipe and the potential result for event modifications.

The PR does not contain a preparation related event along the lines of PrepareItemCraftEvent, as exposing the relative players in such an event is out of scope of the PR. It is also not necessary for plugins to prevent crafting specific items.

Future improvements to this event may include boolean flags to prevent items from being consumed, but this is also out of scope for the PR.


kennytv (approved):

patches/api/0481-Add-CrafterCraftItemEvent.patch
```diff
@@ -0,0 +1,119 @@
...
++        final @NotNull Recipe recipe,
++        final @NotNull ItemStack result
++    ) {
++        super(theBlock);
++        this.recipe = recipe;
++        this.result = result;
++    }
++
++    /**
++     * Yields the recipe crafted by the crafted.
```
DerEchtePilz: This should probably be `crafted by the crafter` instead of `crafted by the crafted`.

lynxplay: indeed xD Thanks

Owen1212055 (approved):

lynxplay:
Unfortunately, in all of their wisdom, upstream has begun work on this event to.
Will monitor their progress, but presumably their PR will supersede this one.

I'll try my best to ensure all methods present on this PR will be in the upstream version too.

kennytv (approved):

PR #11086, opened 2024-07-15
Fix fetching biomes during world generation

### LINKED ISSUE #11084: Lake feature crashes server
### Stack trace

If i make a dimension with custom features all of the features work except for lakes, for some reason they attempt to get biome during world gen (no other features do this) which causes server to crash: https://0bin.net/paste/GAqifVkm#KtLtjtDme8Jax0s4ld6TlCYsK7gX1WF3blXSScKNps4

It shouldn't be crashing when its trying to get the biome.

LavaFeature.java in paper code:

```java
if (blockState.getFluidState().is(FluidTags.WATER)) {
                for (int ab = 0; ab < 16; ab++) {
                    for (int ac = 0; ac < 16; ac++) {
                        int ad = 4;
                        BlockPos blockPos4 = blockPos.offset(ab, 4, ac);
                        if (worldGenLevel.getBiome(blockPos4).value().shouldFreeze(worldGenLevel, blockPos4, false) // here is where it blows up
                            && this.canReplaceBlock(worldGenLevel.getBlockState(blockPos4))) {
                            worldGenLevel.setBlock(blockPos4, Blocks.ICE.defaultBlockState(), 2);
                        }
                    }
                }
            }
```
            
```
            java.lang.IllegalStateException: Requested chunk unavailable during world generation
	at net.minecraft.server.level.WorldGenRegion.getChunk(WorldGenRegion.java:168)
	at net.minecraft.world.level.LevelReader.getNoiseBiome(LevelReader.java:73)
	at net.minecraft.world.level.biome.BiomeManager.getBiome(BiomeManager.java:64)
	at net.minecraft.world.level.LevelReader.getBiome(LevelReader.java:53)
	at net.minecraft.world.level.levelgen.feature.LakeFeature.place(LakeFeature.java:137)
```

### Plugin and Datapack List

I am adding the features via code, not a datapack, but im 99% sure it would be the same situation using a datapack
```
  Holder<ConfiguredFeature<LakeFeature.Configuration, ?>> hadradaLake = CustomFeatures.registerFeatureConfig(
                "hadrada_lake", Feature.LAKE, new LakeFeature.Configuration(BlockStateProvider.simple(WATER), BlockStateProvider.simple(STONE)));

        CustomFeatures.registerPlacedFeature("hadrada_lake", hadradaLake, RarityFilter.onAverageOnceEvery(3), InSquarePlacement.spread(), hadradaHeightRange);
```

_No response_

### Paper version

This server is running Paper version 1.21-DEV-master@34a2917 (2024-07-14T13:04:22Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
You are 6 version(s) behind
Previous version: 1.21-63-2288e13 (MC: 1.21)

### Other

_No response_

NewwindServer:
Fixes #11084

getBiome() on LevelReader calls getNoiseBiome(), which itself calls getChunk() in order to get the biome cached inside the chunk.
However, calling getChunk() on WorldGenRegion will cause a crash if the chunk isn't found: "Requested chunk unavailable during world generation"
This means that features which attempt to fetch biomes during generation (LakeFeature/SnowAndFreezeFeature)
can cause the server to crash.
This patch ensures that these features are fetching biomes in a way that doesn't call getChunk()


fucksophie:
Currently running my server with this exact patch, works perfectly and server no longer crashes on LakeFeature. Would love to see this merged!

jpenilla:
LakeFeature is deprecated and vanilla has moved away from using it for anything except for lava lakes.
The error for unsafe chunk gets you are removing here is an intentional safeguard against bad code, such as that in the dead path in LakeFeature when creating water lakes. While it may appear to 'fix' the problem at hand it is not a proper solution and will cause other more serious problems to go undetected. This should be reported to Mojang if it's reproducible with a datapack. Given the circumstances I don't really think this is something we should care to fix, but if it is it would need to be a targeted change to LakeFeature for the aforementioned reasons.

fucksophie:
This patch works, and stops my datapack heavy server (Terralith, Nullscape, etc...) from crashing. But, at the same time it makes new chunk generation performance x3-x4 worse!

codebycam:
Hi @NewwindServer, can you please create a new issue with proper reproduction steps for 1.21? The issue you're linking that this PR fixes (https://github.com/PaperMC/Paper/issues/11084) was closed due to being incomplete and a deleted user.

PR #11088, opened 2024-07-16
Fix HIDE_ATTRIBUTES flag not working for items without attributes set

### LINKED ISSUE #10744: Hide attributes not working in 1.20.6
### Expected behavior

Hide attributes flag should hide attributes

### Observed/Actual behavior

Adding item flag for hide attributes still shows "when in main hand blah"

### Steps/models to reproduce

itemMeta.addItemFlags(ItemFlag.HIDE_ATTRIBUTES)

Worked perfectly in 1.20.4
weirdly enough, works perfectly with viaversion on 1.20.6 client, so not a client problem

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.20.6-DEV-master@5a9afbe (2024-05-18T13:33:02Z) (Implementing API version 1.20.6-R0.1-SNAPSHOT)
Previous version: git-Paper-"4fd58a1" (MC: 1.20.6)

### Other

_No response_

NewwindServer:
Fixes #10744

Currently, on paper, when you add HIDE_ATTRIBUTES flag to an item, e.g. a diamond sword, it incorrectly still shows the "When in main hand 7 attack damage..." default attributes, which are supposed to be hidden.

On spigot, however, these default attributes are correctly hidden, upon investigation, the general-itemmeta-fixes patch [breaks this functionality](https://github.com/PaperMC/Paper/commit/9d8d38d13714433271fa9ed17485bc1578629b9e#diff-22b14b0a6e7f7cb65eacc9eb15cb1664dbc8d48dc20813dabf7cbce9b7b70573R268), This PR fixes the functionality and brings it back in line with spigot/pre 1.20.5 behaviour.

Note, on #10744 there is a comment stating that having empty attributes defined on the item overrides the default attributes, but from my testing this does not appear to be the case.

### DIFF
--- a/patches/server/0970-General-ItemMeta-fixes.patch
+++ b/patches/server/0970-General-ItemMeta-fixes.patch
@@ -859,19 +859,6 @@ index 58da8cb19a8444c634cbc1f39e93503ca8e2ecab..776f4dcc0b61a5b8ee4020a283cfcfac
              itemTag.put(CraftMetaItem.DAMAGE, this.damage);
          }
  
-@@ -970,10 +976,8 @@ class CraftMetaItem implements ItemMeta, Damageable, Repairable, BlockDataMeta {
-     }
- 
-     void applyModifiers(Multimap<Attribute, AttributeModifier> modifiers, CraftMetaItem.Applicator tag) {
--        if (modifiers == null || modifiers.isEmpty()) {
--            if (this.hasItemFlag(ItemFlag.HIDE_ATTRIBUTES)) {
--                tag.put(CraftMetaItem.ATTRIBUTES, new ItemAttributeModifiers(Collections.emptyList(), false));
--            }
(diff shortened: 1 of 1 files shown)


lynxplay:
This was an intentional change. Have you tested this with non-armor items? (Those have special handling on the client to fallback to their default when empty attribute modifiers are defined)

electronicboy:
This just relies on highly inconsistent behavior which seems to confuse everybody who ends up touching it.

Adding a component will override the default component, mojang has some logic that they scatter around which is smart enough in some cases to get the default values, but, there are many cases where they just don't, and so you end up with an item which will behave unpredictably in many cases. I think that the best solution here would be to allow people to explicitly opt into having an empty component, however, ItemMeta doesn't exactly handle that well and just further perpetuates the potential headaches of promising sane behavior around such a clugy API.

PR #11090, opened 2024-07-16
Fix inventory desync when placing blocks with spawn protection

### LINKED ISSUE #11012: Spawn protection does not update the client inventory.
### Expected behavior

When a BlockPlaceEvent is canceled it should display the same amount of item in the player inventory as if no block where placed.

### Observed/Actual behavior

Currently in paper-1.21-40 item isn't visible in the player inventory after the BlockPlaceEvent have been canceled.
It's only a display bug, moving the item inside the inventory will show the right number of block back.

### Steps/models to reproduce

1. Create a simple plugin that cancel every BlockPlaceEvent.
2. Try to place a block on an 1.21 server.

### Plugin and Datapack List

I have tried with towny and flagsh, both have the same behavior.

### Paper version

[21:32:38 INFO]: This server is running Paper version 1.21-40-master@b45d9b6 (2024-07-01T08:23:14Z) (Implementing API version 1.21-R0.1-SNAPSHOT)

### Other

_No response_

notTamion:
fixes #11012 

Should we maybe call BlockPlaceEvent in a cancelled state when a player tries placing blocks with spawn protection on? This could allow plugins to pick which blocks should be placeable at spawn.

### DIFF
--- a/patches/server/0748-Fix-a-bunch-of-vanilla-bugs.patch
+++ b/patches/server/0748-Fix-a-bunch-of-vanilla-bugs.patch
@@ -52,6 +52,9 @@ https://bugs.mojang.com/browse/MC-200092
 https://bugs.mojang.com/browse/MC-158900
   Fix error when joining after tempban expired
 
+https://bugs.mojang.com/browse/MC-99075
+  Fix inventory desync within spawn protected area
+
 == AT ==
 public net/minecraft/world/entity/Mob leashInfoTag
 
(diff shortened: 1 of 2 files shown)


TJIMOSHIMA:
what the fck

notTamion:
do you have anything of value to add to this pr or are you just here to troll?

Machine-Maker (approved):
Throw `MC-99075` into the comment, it looks like this is a reported vanilla issue.

lynxplay:
Changed with the following line of thought:

Cancelling the inventory open event is *less* restrictive than cancelling the entire interaction all together.
As such, *some* logic should still be run, hence a lot of places will be returning consuming interact results, however *immediate* logic, like stats related to *inspecting* such blocks are not incremented.

lynxplay (approved):

PR #11093, opened 2024-07-16
Bundle spark

kashike:
discussion: https://github.com/PaperMC/Paper/discussions/10565
docs: https://github.com/PaperMC/docs/pull/447

New configuration options (global configuration):
```yaml
spark:
  # If the bundled spark profiler should be enabled at all.
  enabled: true
  # If the bundled spark profiler should be enabled as soon as possible.
  # If true, spark enables as early as possible. This can be useful for profiling the server during startup.
  # If false, spark enables once the server has finished starting up (when the `Done (4.720s)! For help, type "help"` message is sent in console).
  enable-immediately: false
  # Prefers the spark plugin over the bundled spark profiler.
  # If enabled and the spark plugin is found in the plugins directory, it will be loaded and the bundled spark profiler will not be loaded.
  # If enabled and the spark plugin is not found in the plugins directory, the bundled profiler will be enabled.
  prefer-spark-plugin: false
```

### DIFF
--- a/patches/api/0011-Timings-v2.patch
+++ b/patches/api/0011-Timings-v2.patch
@@ -1261,10 +1261,10 @@ index 0000000000000000000000000000000000000000..df142a89b8c43acb81eb383eac0ef048
 +}
 diff --git a/src/main/java/co/aikar/timings/Timings.java b/src/main/java/co/aikar/timings/Timings.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..9812d668ad945aba486fbf6d5bf83c4292cb5d03
+index 0000000000000000000000000000000000000000..445b5d3b61f31cf263475d0155a7b0252c54a3f8
 --- /dev/null
 +++ b/src/main/java/co/aikar/timings/Timings.java
-@@ -0,0 +1,337 @@
+@@ -0,0 +1,331 @@
 +/*
 + * This file is licensed under the MIT License (MIT).
 + *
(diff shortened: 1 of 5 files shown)


MiniDigger:
see https://github.com/PaperMC/docs/issues/446 for docs changes

patches/server/1043-spark.patch
```diff
@@ -0,0 +1,187 @@
...
+     public static void registerCommands(final MinecraftServer server) {
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index aeae4f8d4ead24db315631c3d2c0b930d0d51e02..adc789f7241b347ca239601f04f9494c33563e94 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -1228,6 +1228,8 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+             LOGGER.info("Done ({})! For help, type \"help\"", doneTime);
+             // Paper end
+ 
++            this.server.spark.enableBackgroundProfiler(); // Paper - spark
```
SirYwell: Does this mean the OpenJ9 warning is printed twice?

Outfit8TSB (changes requested):
remove timings patch as it's going to replaced with spark.
i heard that it's now deprecated and marked for removal in paper.

zlataovce:
> remove timings patch as it's going to replaced with spark. i heard that it's now deprecated and marked for removal in paper.

it is deprecated and turned off by default now, but it shouldn't be removed entirely yet, since there are people which may still want to use it/need to use it

Outfit8TSB:
> remove timings patch as it's going to replaced with spark. i heard that it's now deprecated and marked for removal in paper.

disregard this comment, this is not gonna happen as this will turned off as default

Leguan16:
Should also add a config option to disable spark.

patches/server/1041-spark.patch
```diff
@@ -0,0 +1,336 @@
...
++        this.logger.log(Level.INFO, "This server bundles the spark profiler. For more information please visit https://docs.papermc.io/paper/profiling");
++        this.spark = PaperSparkModule.create(Compatibility.VERSION_1_0, server, this.logger, new PaperScheduler() {
++            @Override
++            public void executeAsync(final Runnable runnable) {
++                MCUtil.scheduleAsyncTask(runnable);
++            }
++
++            @Override
++            public void executeSync(final Runnable runnable) {
++                MCUtil.ensureMain(runnable);
```
kennytv: Should try catch the runnable with logging instead of breaking on main, just in case

patches/server/1041-spark.patch
```diff
@@ -0,0 +1,336 @@
...
+index 1a734293c9416f13324bb0edf8f950c9029f8bc4..901575015df18e3f07c271b1384ca72e7c363810 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -61,6 +61,10 @@ dependencies {
+     implementation("io.papermc:reflection-rewriter-runtime:$reflectionRewriterVersion")
+     implementation("io.papermc:reflection-rewriter-proxy-generator:$reflectionRewriterVersion")
+     // Paper end - Remap reflection
++    // Paper start - spark
++    implementation("me.lucko:spark-api:0.1-SNAPSHOT")
++    implementation("me.lucko:spark-paper:1.10.79-SNAPSHOT")
```
kennytv: Don't forget to update spark before merging 👀

kennytv (changes requested):
The `We recommend installing the spark profiler as a replacement:` warning in the timings patch needs updating

It also currently throws exceptions on stop (if you have any other plugin installed) due to a bukkit service being registered with a null plugin, presumably within spark

kennytv (approved):

Outfit8TSB:
I swear that this will lead to confusion as lucko/spark#424 didn't merge. If Spark keeps disabling after the update, it will get confused, This Pull was necessary to merge. As Spark will get disabled without warning.

PR #11094, opened 2024-07-16
Add argument for FinePosition to brig API

Machine-Maker:


patches/api/0468-Brigadier-based-command-API.patch
```diff
@@ -1065,13 +1066,23 @@ index 0000000000000000000000000000000000000000..1d5c599d1b9c8bf07720e651bdbe9dad
...
 +        return provider().blockPosition();
 +    }
 +
 +    /**
++     * A fine position argument.
++     *
++     * @param centerIntegers if whole numbers should be centered (+0.5)
++     * @return fine position argument
++     */
++    public static @NotNull ArgumentType<FinePositionResolver> finePosition(final boolean centerIntegers) {
```
lynxplay, line 1091: Is it worth exposing a method without this boolean flag?
Centering seems like an edge case usage of this, always passing false sounds annoying.

lynxplay (approved):
LGTM otherwise, overload is up to subjective opinion of the merger.

patches/api/0468-Brigadier-based-command-API.patch
Machine-Maker: Ok, added one that defaults to false.

PR #11099, opened 2024-07-17
Add option to throttle spawn attempts after several failures

froobynooby:
This patch adds an option to use a longer ticks-per-spawn for given mob types in chunks where previous spawn attempts have consecutively failed (e.g. spawnproofed chunks). The ticks-per-spawn and the number of failed attempts before they kick in are both configurable (the setting is disabled by default). The number of failed spawn attempts is stored with the chunk so it persists across unloads. As soon as a mob is able to spawn in a chunk, the failed spawn attempt counter resets and the chunk starts using the default ticks-per-spawn again.

This behaviour is useful on servers where players use mob farms. Mob farms typically require spawnproofing the surrounding chunks to improve rates, which causes the server to waste CPU cycles trying to find suitable spawn locations. By identifying suspected spawnproof chunks and throttling attempts at spawning mobs in them, it is possible to improve performance without noticeably affecting overall mob spawning behaviour.

I've been using a [similar patch](https://github.com/FroobWorld/Nabulus/commit/edd916abed9fbf921a023fccbbedd77bc7475858) on my reasonably busy survival server for nearly a year without any complaints from players.

For illustrative purposes I ran a simple test spending 5 minutes AFK at standard spawning-tower mob farm which had been properly spawnproofed.

Without patch (default ticks-per-spawn):
image
With patch (multiplying default ticks-per-spawn by 10 after 1200 failed attempts):
Screenshot from 2024-07-17 22-08-00

### DIFF
--- a/patches/server/1042-Throttle-failed-spawn-attempts.patch
+++ b/patches/server/1042-Throttle-failed-spawn-attempts.patch
@@ -1,0 +1,117 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: froobynooby <froobynooby@froobworld.com>
+Date: Wed, 17 Jul 2024 18:46:11 +0930
+Subject: [PATCH] Throttle failed spawn attempts
+
+This patch adds the option to use longer ticks-per-spawn for a given
+mob type in chunks where spawn attempts are consecutively failing.
+
+This behaviour is particularly useful on servers where players build
+mob farms. Mob farm designs often require making surrounding chunks
+spawnproof, which causes the server to waste CPU cycles trying to spawn mobs in
+vain. Throttling spawn attempts in suspected spawnproof chunks improves
+performance without noticeably advantaging or disadvantaging the mob farm.
+
+diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+index 7509b17414f836e8b53fc62b02b386ba4e8c5ca9..fe202580ce199417bab4cf458a6445a23c3fd4cb 100644
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+@@ -186,6 +186,15 @@ public class WorldConfiguration extends ConfigurationPart {
+             @MergeMap
+             public Reference2IntMap<MobCategory> ticksPerSpawn = Util.make(new Reference2IntOpenHashMap<>(NaturalSpawner.SPAWNING_CATEGORIES.length), map -> Arrays.stream(NaturalSpawner.SPAWNING_CATEGORIES).forEach(mobCategory -> map.put(mobCategory, -1)));
+ 
++            public SpawningThrottle spawningThrottle;
++
++            public class SpawningThrottle extends ConfigurationPart {
++                public int failedAttemptsThreshold = 1200;
++
++                @MergeMap
++                public Reference2IntMap<MobCategory> throttledTicksPerSpawn = Util.make(new Reference2IntOpenHashMap<>(NaturalSpawner.SPAWNING_CATEGORIES.length), map -> Arrays.stream(NaturalSpawner.SPAWNING_CATEGORIES).forEach(mobCategory -> map.put(mobCategory, -1)));
++            }
++
+             @ConfigSerializable
+             public record DespawnRange(@Required int soft, @Required int hard) {
+             }
+diff --git a/src/main/java/net/minecraft/world/level/NaturalSpawner.java b/src/main/java/net/minecraft/world/level/NaturalSpawner.java
+index ed8032495af9ce9c23419224814b8d27e4a97c17..7eece3f7d7b2ea732502b03697ac3d843ee5e2b4 100644
+--- a/src/main/java/net/minecraft/world/level/NaturalSpawner.java
++++ b/src/main/java/net/minecraft/world/level/NaturalSpawner.java
+@@ -140,8 +140,14 @@ public final class NaturalSpawner {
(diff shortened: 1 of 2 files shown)


molor:
It looks a bit like https://github.com/PaperMC/Paper/pull/9400

froobynooby:
Similar concept, but different implementations targeted at different problems. The other PR relates to spawn attempts that are failing due to plugins cancelling or aborting the PreCreatureSpawnEvent, while this one primarily relates to chunks where mobs have repeatedly failed to spawn for other reasons (e.g. no suitable spawning locations)

patches/server/1042-Throttle-failed-spawn-attempts.patch
```diff
@@ -0,0 +1,117 @@
...
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+@@ -186,6 +186,15 @@ public class WorldConfiguration extends ConfigurationPart {
+             @MergeMap
+             public Reference2IntMap<MobCategory> ticksPerSpawn = Util.make(new Reference2IntOpenHashMap<>(NaturalSpawner.SPAWNING_CATEGORIES.length), map -> Arrays.stream(NaturalSpawner.SPAWNING_CATEGORIES).forEach(mobCategory -> map.put(mobCategory, -1)));
+ 
++            public SpawningThrottle spawningThrottle;
++
++            public class SpawningThrottle extends ConfigurationPart {
++                public int failedAttemptsThreshold = 1200;
```
lynxplay: Am I missing something here or is this value arbitary?

In which case, imo, it should be an IntOr.Disabled.

patches/server/1042-Throttle-failed-spawn-attempts.patch
```diff
@@ -0,0 +1,117 @@
...
+ 
+diff --git a/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java b/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
+index 75c8125e20b70433fe9d143a3193d821043327c3..03463c0aa67773d1a8800662d1e72fe6c428a633 100644
+--- a/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
++++ b/src/main/java/net/minecraft/world/level/chunk/ChunkAccess.java
+@@ -84,6 +84,7 @@ public abstract class ChunkAccess implements BlockGetter, BiomeManager.NoiseBiom
+     public final Map<BlockPos, BlockEntity> blockEntities = new Object2ObjectOpenHashMap();
+     protected final LevelHeightAccessor levelHeightAccessor;
+     protected final LevelChunkSection[] sections;
++    public final Map<net.minecraft.world.entity.MobCategory, Long> failedSpawnAttempts = new java.util.EnumMap<>(net.minecraft.world.entity.MobCategory.class); // Paper - throttle failed spawn attempts
```
lynxplay: Is there a point in e.g. Int2Long/LongArrayList here?
May also just be premature tho. Sad fastutil hasn't added Enum2LongMaps yet.

patches/server/1042-Throttle-failed-spawn-attempts.patch
```diff
@@ -0,0 +1,117 @@
...
+This behaviour is particularly useful on servers where players build
+mob farms. Mob farm designs often require making surrounding chunks
+spawnproof, which causes the server to waste CPU cycles trying to spawn mobs in
+vain. Throttling spawn attempts in suspected spawnproof chunks improves
+performance without noticeably advantaging or disadvantaging the mob farm.
+
+diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
+index 7509b17414f836e8b53fc62b02b386ba4e8c5ca9..fe202580ce199417bab4cf458a6445a23c3fd4cb 100644
+--- a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
++++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
```
lynxplay: Configuration diff goes into the Paper Configurations patch.

lynxplay (changes requested):
I like the general concept of the patch 👍 
Keeping this on the chunk presumably also makes sense, I think there is some discussion to be had regarding EnumMap and object Long usage, but beyond that, just some minor nitpicks.

patches/server/1042-Throttle-failed-spawn-attempts.patch
froobynooby: Not too sure on what the best map is to be honest. There seems to be some use of Object2LongOpenHashMap for enum keys in other parts of the code - maybe that?

froobynooby: I've just changed it to use an array which I think is probably the simplest approach, let me know if you think there's a better way

lynxplay: Idk, didn't have to change it just yet, I was hoping for some feedback from other team members 😅 
But yea, this certainly works and avoids unboxing nicely. Thank you :)

froobynooby:
> This patch makes creeper farms not work. Can you fix this? https://www.youtube.com/watch?v=qyUZ0iXkdBA

Do you have a world download/schematic and configuration you could share?

The patch slows down spawn attempts in chunks that have repeatedly failed to spawn mobs - it shouldn't be stopping spawn attempts altogether. Make sure you're not configuring `failed-attempts-threshold` too low (I recommend setting it to 1200) or the `throttled-ticks-per-spawn` settings too high (I recommend setting them to 10x the normal `ticks-per-spawn` settings)

electronicboy:
I would imagine that anything relying on spawn proofing is generally going to suck hard with this sort of feature, the concepts are pretty much incompatible, IMHO.

PR #11107, opened 2024-07-18
Migrate ArmorStand meta to using entity tag

### LINKED ISSUE #11105: ArmorStand with meta deserialized by bukkit not saveable.
### Expected behavior

Usually, you would be able to get and set items to/from a config and they should work as normal items and be saved by minecraft

### Observed/Actual behavior

ArmorStands with any meta (name only, lore only) cannot be saved by minecraft IF the steps below are done.
The issue seems to lie in the deserialized item having a nbt compound called "entity-tag". This tag does not seem to be saveable by minecraft nbt. /paper dumpitem also errors for this deserialized item.

### Steps/models to reproduce

Steps to replicate

1) Get an armorstand with any meta (name changed or any lore)
2) Serialize using Bukkit(via config, BukkitObjectOutputStream etc)
3) Deserialized using Bukkit(the opposite of how you serialized)
4) Add this deserialized item to inv
5) This item is unsaveable, causing an error. Inventories with this item, chunks etc won't save

### Plugin and Datapack List

Not necessary.

### Paper version

version 1.21-50-master@20889d0 (2024-07-10T16:28:03Z) (Implementing API version 1.21-R0.1-SNAPSHOT)

### Other

https://pastebin.com/adS1NMnR

Stacktrace seen when leaving server with item in inventory.

electronicboy:
Our handling of this class causes some headaches due to ItemMeta being lossy, we can double patch this by properly storing the data we care about, there is still some stuff to be addressed, i.e. we really should ensure that the entity tag is stored to the tag properly

TODO: actually, we should just serialise the entity tag directly as b64

---
Download the paperclip jar for this pull request: [paper-11107.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1821847609.zip)

### DIFF
--- a/patches/server/0169-Add-ArmorStand-Item-Meta.patch
+++ b/patches/server/0169-Add-ArmorStand-Item-Meta.patch
@@ -33,10 +33,10 @@ index eef3517833ff5c0cf41b89973ebc972b8ed31e0f..c9fbc01be0b0e7fd1cafb091d06496f4
              (type, meta) -> meta instanceof CraftMetaArmorStand armorStand ? armorStand : new CraftMetaArmorStand(meta));
  
 diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
-index c4f12f96e39cb6189799a796b4cb2cb4f0b92392..59bdac414e8205ed608f79ef0d1502acd826d216 100644
+index c4f12f96e39cb6189799a796b4cb2cb4f0b92392..ecce5d0da946ca279c5608068442cc53437dd2a5 100644
 --- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
 +++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaArmorStand.java
-@@ -11,9 +11,22 @@ import org.bukkit.Material;
+@@ -11,9 +11,17 @@ import org.bukkit.Material;
  import org.bukkit.configuration.serialization.DelegateDeserialization;
  
  @DelegateDeserialization(SerializableMeta.class)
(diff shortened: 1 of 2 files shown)


patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
```diff
@@ -0,0 +1,294 @@
...
+-        // Paper end
+         if (this.entityTag != null) {
+             tag.put(CraftMetaArmorStand.ENTITY_TAG, CustomData.of(this.entityTag));
+         }
+@@ -148,7 +109,7 @@ public class CraftMetaArmorStand extends CraftMetaItem implements com.destroysto
+     }
+ 
+     boolean isArmorStandEmpty() {
+-        return !(this.invisible != null || this.noBasePlate != null || this.showArms != null || this.small != null || this.marker != null || this.entityTag != null); // Paper
++        return entityTag.isEmpty();
```
Lulu13022002: entityTag can be null here and i think a null check should be enough

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
```diff
@@ -0,0 +1,313 @@
...
+         // Paper end
+ 
+@@ -236,54 +211,66 @@ public class CraftMetaArmorStand extends CraftMetaItem implements com.destroysto
+     }
+ 
+     // Paper start
++    private void populateTagIfNull() {
++        if (this.entityTag == null) {
++            this.entityTag = new CompoundTag();
++            this.entityTag.putString(ENTITY_ID.NBT, "minecraft:armorstand");
```
TehBrian: Should the entity ID be `minecraft:armor_stand`?

The ID doesn't even seem to matter for armor stands, from what I can tell.

Sikatsu:
Hey, is there any update on this pull request?
We're running into an issue where the player is being kicked, which seems related to this.
https://pastebin.com/NpU9tkEL

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
```diff
@@ -0,0 +1,313 @@
...
+     }
+@@ -200,24 +159,40 @@ public class CraftMetaArmorStand extends CraftMetaItem implements com.destroysto
+         super.serialize(builder);
+ 
+         // Paper start
+-        if (invisible != null) {
+-            builder.put(INVISIBLE.BUKKIT, invisible);
++        if (entityTag == null) {
++            return builder;
++        } else if (true) {
```
TehBrian: `else {`?

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
```diff
@@ -0,0 +1,313 @@
...
+ 
+-        if (marker != null) {
+-            builder.put(MARKER.BUKKIT, marker);
++        if (this.entityTag.get(CraftMetaArmorStand.SMALL.NBT) != null) {
++            builder.put(SMALL.BUKKIT, this.entityTag.getBoolean(CraftMetaArmorStand.SMALL.NBT));
++        }
++
++        if (this.entityTag.get(CraftMetaArmorStand.MARKER.NBT) != null) {
++            builder.put(MARKER.BUKKIT, this.entityTag.getBoolean(CraftMetaArmorStand.MARKER.NBT));
+         }
```
TehBrian: Unreachable code. Was there supposed to be an expression instead of `true` up above?

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
```diff
@@ -0,0 +1,313 @@
...
++        if (this.entityTag == null) {
++            this.entityTag = new CompoundTag();
++            this.entityTag.putString(ENTITY_ID.NBT, "minecraft:armorstand");
++        }
++    }
++
+     @Override
+     public boolean isInvisible() {
+-        return invisible != null && invisible;
++        return entityTag != null && entityTag.contains(INVISIBLE.NBT) && entityTag.getBoolean(INVISIBLE.NBT);
```
TehBrian: Could just be `return entityTag != null && entityTag.getBoolean(INVISIBLE.NBT);` since getBoolean defaults to 0. Same with all the other getters below.

TehBrian:
> We're running into an issue where the player is being kicked, which seems related to this. https://pastebin.com/NpU9tkEL
Yup, it is.

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
electronicboy: that's to silence the dead code check

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
electronicboy: No, this was just a general update of the code here, followed by me adding the entity tag serialisation and keeping it

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
electronicboy: Yes, I ended up pulling in the old string somehow

patches/server/1041-fixup-Add-ArmorStand-Item-Meta.patch
electronicboy: I just followed the pattern elsewhere, and I generally prefer to not rely on default behavior impls

TehBrian: Gotcha, plus including #contains is clearer, too. I just wanted to bring it up :P

Machine-Maker (approved):
Tested, deserialized stacks are valid

PR #11145, opened 2024-07-22
Early Access to Dispatcher

Dueris:
This adds a static variable to the Commands class to provide early access to the command dispatcher before bootstrapping the command system or before MinecraftServer has been loaded.

This is mainly needed for a plugin I'm working on, which needs access to the command dispatcher at plugin bootstrap to allow for mcfunctions to access the commands I add. Previously, I needed to wait for the Commands class to be created for bootstrapping, which occurs after plugin bootstrapping.

### DIFF
--- a/patches/server/1045-Early-Dispatcher-Access.patch
+++ b/patches/server/1045-Early-Dispatcher-Access.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: dueris <jedimastertoothless@hotmail.com>
+Date: Mon, 22 Jul 2024 00:56:21 -0700
+Subject: [PATCH] Early Dispatcher Access
+
+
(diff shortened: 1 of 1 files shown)


lynxplay:
What is preventing you from simply registering such commands in the plugin bootstrap via the bootstrap context?

BootstrapContext exposes a LifecycleEventManager in which you can also register commands during the initial commands class creation?

Dueris:
Im currently using NMS for command creation, as most of my plugin follows the NMS system with ServerPlayers, ResourceLocations, etc

kennytv:
There's no point in using internals if you can do the same thing via API

lynxplay:
Then reflect into the exposed Commands during the bootstrap lifecycle, it holds onto the dispatcher.

Given that internals are not API and we may change anything about it when we see fit, hence I don't think this PR is of much use for you. There is little point in promising a plugin like yours a static method for the dispatcher, when such a method may not make sense anymore in a couple of versions (or already barely does, given we have exposed brigadier completely to the API).

Thank you for the PR work, but I don't think Paper is interested in such a change due to the reasons laid out above.

PR #11147, opened 2024-07-22
Implement more methods for horse inventories

Lulu13022002:
This depends on my previous PR to avoid conflict.

Implement some missing methods (set/getContents and isEmpty) to take in account the armor slot. This fixes by extension other api using those methods such as addItem and more.

### DIFF
--- a/patches/server/0950-Improve-tag-parser-handling.patch
+++ b/patches/server/0950-Improve-tag-parser-handling.patch
@@ -109,8 +109,65 @@ index da101bca71f4710812621b98f0a0d8cab180346a..3cd112584accb8e8f050ac99738eed11
 +        }
 +    }
  }
+diff --git a/src/main/java/net/minecraft/network/chat/ComponentUtils.java b/src/main/java/net/minecraft/network/chat/ComponentUtils.java
+index 3365aed2b67ae0e4dd0410f5190ba474f146139b..8b776434a733b91129b089d702b2583b727bf3d2 100644
+--- a/src/main/java/net/minecraft/network/chat/ComponentUtils.java
++++ b/src/main/java/net/minecraft/network/chat/ComponentUtils.java
+@@ -33,10 +33,30 @@ public class ComponentUtils {
+         }
+     }
+ 
++    @io.papermc.paper.annotation.DoNotUse // Paper - validate separators - right now this method is only used for separator evaluation. Error on build if this changes to re-evaluate.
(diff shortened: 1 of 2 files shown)


Machine-Maker (approved):
I originally modeled the changes off of CraftResultInventory which *doesn't* override these various methods. Perhaps it should be changed there as well, but that can be in a different PR.

patches/server/1031-fix-horse-inventories.patch
```diff
@@ -74,6 +74,42 @@ index 4946da593713f4d11d88ac1bb68a089f2f6d5ae0..abef7f23361e6c5d18243dd18439ffd1
...
++        }
++
++        return items;
++    }
++
++    @Override
++    public void setContents(ItemStack[] items) {
++        com.google.common.base.Preconditions.checkArgument(items.length <= this.getSize(), "Invalid inventory size (%s); expected %s or less", items.length, this.getSize());
++
++        this.setSaddle(items[net.minecraft.world.entity.animal.horse.AbstractHorse.INV_SLOT_SADDLE]);
```
lynxplay: These two setters seems wrong. The passed `items` array may be empty. Or only include a single item.
These two lines would IOB.

Lulu13022002: yes i found a similar issue on the getter.

Lulu13022002:
Yes you already have a pr for other inventories (https://github.com/PaperMC/Paper/pull/10721/files) however you only override the getter for some reason.

patches/server/1031-fix-horse-inventories.patch
```diff
@@ -74,6 +74,42 @@ index 4946da593713f4d11d88ac1bb68a089f2f6d5ae0..abef7f23361e6c5d18243dd18439ffd1
...
++    }
++
++    @Override
++    public void setContents(ItemStack[] items) {
++        com.google.common.base.Preconditions.checkArgument(items.length <= this.getSize(), "Invalid inventory size (%s); expected %s or less", items.length, this.getSize());
++
++        ItemStack saddle = items.length > net.minecraft.world.entity.animal.horse.AbstractHorse.INV_SLOT_SADDLE ? items[net.minecraft.world.entity.animal.horse.AbstractHorse.INV_SLOT_SADDLE] : null;
++        ItemStack armor = items.length > net.minecraft.world.inventory.HorseInventoryMenu.SLOT_BODY_ARMOR ? items[net.minecraft.world.inventory.HorseInventoryMenu.SLOT_BODY_ARMOR] : null;
++
++        this.setSaddle(saddle);
```
lynxplay: org.apache.commons.lang3.ArrayUtils#get would make this a lot nicer to look at.

PR #11154, opened 2024-07-23
Fix PickupStatus getting reset

### LINKED ISSUE #10874: Arrow bouncing back on entity causes it to allow pickup again
### Expected behavior

An arrow projectile modified with the following code:
```
Arrow arrow;
arrow.setPickupStatus(AbstractArrow.PickupStatus.DISALLOWED);
```
should never allow pickup.

### Observed/Actual behavior

An arrow, modified with the above code, after it 'bounces back' from an entity, is suddenly allowed to be picked up.
'Bouncing back' occurs when the entity is on damage cooldown (The video demonstrates this by damaging the entity with an arrow prior to firing other arrows that will bounce back).

### Steps/models to reproduce

Code used (this video is for context, not the actual issue):
```java
arrow.setPickupStatus(AbstractArrow.PickupStatus.CREATIVE_ONLY);
```
https://github.com/PaperMC/Paper/assets/69715898/2b446294-5c43-4e41-b781-1bacfb5bdd45

Code used (the actual issue):
```java
arrow.setPickupStatus(AbstractArrow.PickupStatus.DISALLOWED);
```

https://github.com/PaperMC/Paper/assets/69715898/bb4de958-9230-4ce9-b64d-c62426f94a82

### Plugin and Datapack List

Plugins:
```
[22:21:17 INFO]: Server Plugins (1):
[22:21:17 INFO]: Bukkit Plugins:
[22:21:17 INFO]:  - CrossYourBows
```
Datapacks
```
[22:21:32 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[22:21:32 INFO]: There are no more data packs available
```

### Paper version

```
[22:20:24 INFO]: Checking version, please wait...
[22:20:24 INFO]: This server is running Paper version 1.20.6-138-master@906df69 (2024-06-10T06:32:15Z) (Implementing API version 1.20.6-R0.1-SNAPSHOT)
```

### Other

The plugin named 'CrossYourBows' is the one modifying the arrow.
This is done in the `EntityShootBowEvent`.
In addition, the 'knockbackStrenght', 'pierceLevel', 'damage' and 'critical' properties are adjusted as well.
The plugin does not modify the arrow entity at any other time than in that specific event.

Note that (as shown in the video's) the items get returned, even in creative mode.

notTamion:
fixes #10874


patches/server/1045-Fix-PickupRule-reset-when-deflected.patch
```diff
@@ -0,0 +1,55 @@
...
+         super.setOwner(entity);
++        if (!updatePickup) return; // Paper - Fix PickupRule reset when deflected
+         Entity entity1 = entity;
+         byte b0 = 0;
+ 
+@@ -702,6 +702,13 @@ public abstract class AbstractArrow extends Projectile {
+         this.pickup = entityarrow_pickupstatus;
+     }
+ 
++    // Paper start - Fix PickupRule reset when deflected
```
lynxplay: This diff can be moved into the other diff.
E.g.
```java
@Override
public void setOwner(Entity) {
// Paper start - ...
setOwner(entity, true);
}

public void setOwner(Entity, boolean) {
  super.setOwner(entity);
  if (!updatePickup) return;
// Paper end - ...
  ..
}

notTamion: so we don't have as many paper comments?

lynxplay: Just keeps the diff nicely together in the class.

TonytheMacaroni:
Would it be possible to use the overload for the API `Projectile#setShooter` as well? Currently, using the method changes the pickup status. Or at least, an override for the API method as well on API `AbstractArrow`.

lynxplay:
I think best solved in a separate PR.

notTamion:
Could also do it in this one though, then we can rename the patch and pr to "Fix PickupRule being reset" or something more generic like that (considering we already have too many very specific patches). Your decision though, just my opinion.

lynxplay:
If you are fine with it yea, just means we need some more eyes on the API decisions (I am going to suggest an overload in the API to not fuck with spigot compat, but I am not 100% convinced so, yea, just means the PR might need some more eyes)

notTamion:
I will just rename the patch to make it more generic and add another commit to add the changes to the api so if you guys decide its best handled in another pr you can just reset that commit

lynxplay (approved):
Generally, I think this is a valid enough solution.

Alternatively we could have override deflect in AbstractArrow
instead, store previous state there and restore. Kinda annoying tho
as it does unwanted computation.

The diff right now also uses an early return, so we are hoping setOwner isn't further abused
in regards to added logic, so that future versions do not have logic skipped there that would
have been needed.
Seems unlikely tho, so I am fine with this under the pretext of minimal diff.

kennytv (approved):

PR #11155, opened 2024-07-23
Fix MC-99601: Async gameprofile fetching in GameProfileArgument

EpicPlayerA10:
Fixes lag when fetching gameprofile in commands, like `/whitelist add`, `/whitelist remove` `/ban`, `/op` etc. Adds async gameprofile fetching in GameProfileArgument.

See https://bugs.mojang.com/browse/MC-99601


lynxplay:
After some more time and feedback from the team, I'll be closing this as a wont-merge.
Still thank you for your time and the objectively great work on the PR, even if the concept is something we do not feel comfortable merging.

The main reason for closing this is the potential reliance of this from third-party tools or plugins.
They may invoke these commands and simply rely on the fact that it happens *instantly*.

Given that third party plugins can 100% implement this PR in custom commands that would allow server owners to enjoy a lag-free whitelist/ban call without interfering with the vanilla behaviour here adds to the above reason.

I hope this makes sense to you, and again, thank you for the work, it is greatly appreciated even if we decided to not go forward with it. If you do end up developing this as a plugin or the likes, feel 100% free to link it here for others in need to find.

PR #11159, opened 2024-07-24
Add Vault Block API

Boy0000:
Adds bunch of basic API to get/set properties of vault blocks

Not finished but opening to get some feedback on structure and whatnot

The API Break with renaming of Vault-api methods get/setTrialSpawnerState -> get/setVaultState needs bytecode modification
If a contributor could sort that before merging this PR

All accessing of VaultServerData is done through TileEntity instead of a snapshot, due to snapshot always having a ServerLevel null
Therefore VaultServerData would always return null
Unsure if its a bug in my implementation or another issue
image


dawon:
I would also add methods accepting player UUIDs, so you can manage rewarded players when they are offline/you don't have a Player instance. It is done similarly for example for Wolf owners etc...

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,278 @@
...
+     );
+     private static final int MAX_REWARD_PLAYERS = 128;
+     private final Set<UUID> rewardedPlayers = new ObjectLinkedOpenHashSet<>();
+@@ -52,13 +52,15 @@ public class VaultServerData {
+         return this.lastInsertFailTimestamp;
+     }
+ 
+-    Set<UUID> getRewardedPlayers() {
++    // Paper start - Vault API
++    public Set<UUID> getRewardedPlayers() {
```
dawon: You should use Access Transformers for this and also other occurences below. It will reduce the diff quite a bit...: https://github.com/PaperMC/Paper/blob/master/CONTRIBUTING.md#access-transformers

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,278 @@
...
+-                    ItemStack.CODEC.listOf().lenientOptionalFieldOf("items_to_eject", List.of()).forGetter(data -> data.itemsToEject),
+-                    Codec.INT.lenientOptionalFieldOf("total_ejections_needed", Integer.valueOf(0)).forGetter(data -> data.totalEjectionsNeeded)
+-                )
+-                .apply(instance, VaultServerData::new)
++                UUIDUtil.CODEC_LINKED_SET.lenientOptionalFieldOf("rewarded_players", Set.of()).forGetter(data -> data.rewardedPlayers),
++                Codec.LONG.lenientOptionalFieldOf("state_updating_resumes_at", Long.valueOf(0L)).forGetter(data -> data.stateUpdatingResumesAt),
++                ItemStack.CODEC.listOf().lenientOptionalFieldOf("items_to_eject", List.of()).forGetter(data -> data.itemsToEject),
++                Codec.INT.lenientOptionalFieldOf("total_ejections_needed", Integer.valueOf(0)).forGetter(data -> data.totalEjectionsNeeded)
++            )
++            .apply(instance, VaultServerData::new)
```
notTamion: I don't think these indentations should be changed here as this just adds more potential conflicts for future updates

notTamion (changes requested):

patches/server/1045-Vault-API.patch
Boy0000: Yeah not to sure why this is being weird...
Saw it originally though didnt indent it myself

Recent commit here is directly copied from the original file and should therefore not touch the indentation, though seems it does

patches/server/1045-Vault-API.patch
Boy0000: Think this is done now 👍

patches/api/0483-Vault-API.patch
```diff
@@ -0,0 +1,97 @@
...
+ }
+diff --git a/src/main/java/org/bukkit/block/data/type/Vault.java b/src/main/java/org/bukkit/block/data/type/Vault.java
+index 87e46dfb951b6e4b753af7c31efd444e549052b9..9a8ee2bcd1b3bd89d11aab87bb5153f0d8e88458 100644
+--- a/src/main/java/org/bukkit/block/data/type/Vault.java
++++ b/src/main/java/org/bukkit/block/data/type/Vault.java
+@@ -16,14 +16,14 @@ public interface Vault extends Directional {
+      * @return the 'vault_state' value
+      */
+     @NotNull
+-    State getTrialSpawnerState();
```
Brokkonaut: This might break plugins. Maybe keep those old methods but deprecate them

notTamion: In my opinion this is a fine api break, considering this is clearly a wrong name and the methods have not been around for long and since the naming of these methods might cause confusion.

lynxplay: While I don't hate the API break here, we cannot break ABI for spigot plugins as those will need to use said methods.
As such, this PR would have to address the conversion in commodore byte code rewriting.

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
+     );
+     private static final int MAX_REWARD_PLAYERS = 128;
+     private final Set<UUID> rewardedPlayers = new ObjectLinkedOpenHashSet<>();
+@@ -52,7 +52,7 @@ public class VaultServerData {
+         return this.lastInsertFailTimestamp;
+     }
+ 
+-    public Set<UUID> getRewardedPlayers() {
++    Set<UUID> getRewardedPlayers() {
```
Lulu13022002: This diff shouldn't appear if handled by AT, same below

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
+         return this.rewardedPlayers;
+     }
+ 
+@@ -60,6 +60,27 @@ public class VaultServerData {
+         return this.rewardedPlayers.contains(player.getUUID());
+     }
+ 
++    // Paper start - Vault API
++
```
Lulu13022002: No need to add those extra newline between the comments

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
++    public boolean hasRewardedPlayer(UUID uniqueId) {
++        return this.rewardedPlayers.contains(uniqueId);
++    }
++
++    public void addToRewardedPlayers(UUID uniqueId) {
```
Lulu13022002: A lot of AT can be removed if you use your new methods instead and just get the uuid from the player and to reduce the diff a bit you could just change the original method and make a new overload with player, same applies for the api.

patches/api/0483-Vault-API.patch
```diff
@@ -0,0 +1,97 @@
...
+ /**
+- * Represents a captured state of a trial spawner.
++ * Represents a captured state of a vault.
+  */
+ public interface Vault extends TileState {
++
++    // Paper start - Vault API
++
++    // Vault Config
++    public LootTable getLootTable();
```
Lulu13022002: public modifier is not needed in interface

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
++            if (player != null) {
++                players.add(player);
++            }
++        }
++        return players.build();
++    }
++
++    @Override
++    public Collection<UUID> getRewardedPlayerUUIDs() {
++        return this.getTileEntity().getServerData().getRewardedPlayers();
```
Lulu13022002: This should be immutable i think

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
++        this.getTileEntity().getServerData().markChanged();
++    }
++
++    @Override
++    public List<ItemStack> getItemsToEject() {
++        return this.getTileEntity().getServerData().getItemsToEject().stream().map(i -> i.asBukkitCopy()).toList();
++    }
++
++    @Override
++    public void setItemsToEject(final List<ItemStack> itemsToEject) {
```
Lulu13022002: This doesn't check empty item, so i can easily create a corrupted block entity

patches/server/1045-Vault-API.patch
```diff
@@ -0,0 +1,305 @@
...
++    // Paper end - Vault API
++
+     @Override
+     public CraftVault copy() {
+         return new CraftVault(this, null);
+@@ -24,4 +183,5 @@ public class CraftVault extends CraftBlockEntityState<VaultBlockEntity> implemen
+     public CraftVault copy(Location location) {
+         return new CraftVault(this, location);
+     }
++
```
Lulu13022002: this change is not needed

patches/server/1045-Vault-API.patch
electronicboy: You have to be weary of where the "original file" comes from, we use a different decompiler to spigot which tends to produce a more stable (but different) output vs spigots decompiler. if spigot touched this class, it will be handled by their decompiler

Boy0000: should be fixed now, it and the few other weird unintended diffs 👍

patches/server/1045-Vault-API.patch
Boy0000: Empty items seems to be checked internally and handled so null and AIR items should work fine ithink
Unless you want me to explicitly mark ItemStack as notnull here
image

Lulu13022002: The codec doesn't support those empty item.
For reference here is the stacktrace when you save the world without interacting with the vault (or using the data command): https://pastes.dev/faeBdNZWcj

Lulu13022002:
You should also check with requirePlaced/isPlaced to take in account unplaced blockstate for all methods that works on the block entity directly (and require a level to be set) and not the snapshot

patches/api/0483-Vault-API.patch
yannicklamprecht: That part can be remove when pr is rebased. see https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/8a9ecf29405887b9e58244cd5d87e8ca0a5970d4#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Fblock%2Fdata%2Ftype%2FVault.java

yannicklamprecht (changes requested):

Boy0000:
merged onto master for 1.21.4 🙏

PR #11164, opened 2024-07-25
Make max interaction range configurable

### LINKED ISSUE #11163: Max interaction range is too low
### Expected behavior

Max interaction range server-side was 6 blocks, this could result in hits not registering for players with very high ping, but it was uncommon. 
In 1.20.5 and onwards, mojang lowered it to 4 blocks, this makes hits not registering a lot more common.

### Observed/Actual behavior

in 1.20.5+, mojang lowered the max interaction range from 6 to 4 blocks.

1.20.4 (6 blocks):
`
public static final double MAX_INTERACTION_DISTANCE = Mth.square(6.0D);
if (axisalignedbb.distanceToSqr(this.player.getEyePosition()) < ServerGamePacketListenerImpl.MAX_INTERACTION_DISTANCE) {
`

1.20.5 (+1 block reach expansion = 4 blocks):
`if (this.player.canInteractWithEntity(axisalignedbb, 1.0D)) {`

This is causing hit registration issues on my server and other PVP servers.
The max interaction range should be raised back to 6 or perhaps be configurable.

### Steps/models to reproduce

If a player with 250ms ping is attacking another player who is sprinting away with speed 2, it creates enough desync for almost none of the hits in the video to register. 

https://github.com/user-attachments/assets/988e2e8b-fc67-4f74-bd09-f06e1e205114

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.21-DEV-master@5a5035b (2024-07-23T09:31:35Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
Previous version: git-Paper-"f4c7d37" (MC: 1.20.4)

### Other

I will open a PR to fix this.

NewwindServer:
Fixes #11163 

In 1.20.5, mojang, failing to consider network latency, lowered the max interaction range from 6 blocks to 4.
This causes hits from high latency players in specific scenarios such as the target sprinting away to not register.

This patch allows configuration of the expansion amount, allowing changing back of the range to 6 blocks, as it was in 1.20.4


MachineBreaker:
That should be configurable imo

lynxplay:
Would also agree to make this configurable via a DoubleOr.Default, with default falling back to vanillas default value.

lynxplay:
Great 👍 Can you move the diff regarding the config changes into the Paper-Configurations patch?
We generally keep all configuration diff there.

lynxplay:
The hash is not absolute. Run `git log` in your Paper-Server directory and locate it 👍

electronicboy:
you can use git blame, be that in the terminal or in your IDE (IJ right click the gutter, show annotations), to get the ID, or search up git log as said

lynxplay (approved):
Split the diff for ya, lgtm otherwise.

kennytv (approved):

patches/server/1045-Increase-max-interaction-range.patch
```diff
@@ -0,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Newwind <support@newwindserver.com>
+Date: Thu, 25 Jul 2024 13:00:37 +0200
+Subject: [PATCH] Increase max interaction range
+
+In 1.20.5, mojang, failing to consider network latency, lowered the max interaction range from 6 blocks to 4.
+This causes hits from high latency players in specific scenarios such as the target sprinting away to not register.
+this patch increases the expansion to 3 blocks, meaning the max interaction range is 6 blocks, as it was in 1.20.4
```
masmc05: I think this is outdated, this patch doesn't increase anything but makes it configurable

lynxplay: Yea, can be handled by the merger.
Needs a docs PR.

NewwindServer: Should I increase the default interaction range expansion value to 3.0 to match the pre 1.20.5 6 blocks behaviour? Feels like a more sane default especially since I know of other server owners reporting the same issue with the default 4 block limit.

lynxplay: I'd rather have the server stick to the more restrictive value and allow server owners to a) deviate from vanilla and b) make the server more lenient by changing the config, not the other way around.

lynxplay:
Anything else from your side for this PR?
Otherwise I'll be merging it and https://github.com/PaperMC/docs/pull/456 in ~30mins.

Issue #11167, opened 2024-07-25
World#createExplosion does not damage or deal knockback to the explosion cause


sh-dima:
### Expected behavior

Explosions created using [World#createExplosion](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#createExplosion(double,double,double,float,boolean,boolean,org.bukkit.entity.Entity)) should deal damage and knockback to all entities within the range of the explosion, including the entity specified as the explosion cause.

### Observed/Actual behavior

Explosions created using [World#createExplosion](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#createExplosion(double,double,double,float,boolean,boolean,org.bukkit.entity.Entity)) deal the proper damage and knockback to all entities other than the entity specified as the `source` argument.

### Steps/models to reproduce

I encountered this issue while developing [my plugin](https://github.com/Slqmy/bad-piggies-plugin). The instant TNT feature does not work correctly because of this, and functions as described above.

Steps to reproduce:
- Create a new server with the latest version of Paper.
- Install [my plugin](https://github.com/Slqmy/bad-piggies-plugin) (or any other plugin that uses the [World#createExplosion](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#createExplosion(double,double,double,float,boolean,boolean,org.bukkit.entity.Entity)) method).
- Do whatever action triggers the [World#createExplosion](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#createExplosion(double,double,double,float,boolean,boolean,org.bukkit.entity.Entity)) method.

For example, here is the code used in [my plugin](https://github.com/Slqmy/bad-piggies-plugin):
```java
    private void detonateInstantTnt(@NotNull Block instantTnt, @Nullable Entity cause) {
        instantTnt.setType(Material.AIR);

        float power = (float) plugin.getConfig().getDouble("features.instant-tnt.explosion.power");
        boolean shouldSetFire = plugin.getConfig().getBoolean("features.instant-tnt.explosion.sets-fire");
        boolean shouldBreakBlocks = plugin.getConfig().getBoolean("features.instant-tnt.explosion.breaks-blocks");

        instantTnt.getWorld().createExplosion(BlockUtil.getBlockCenterLocation(instantTnt), power, shouldSetFire, shouldBreakBlocks, cause);

        removeInstantTnt(instantTnt);
    }
```

My `config.yml`:
```yaml
features:
[...]

PR #11180, opened 2024-07-27
Add API for explosions to damage the explosion cause

sh-dima:
This PR is to add an API feature which expands the functionality of `World#createExplosion` by adding the option to deal damage to the source entity (this is not the default behaviour). This deals with [this issue](https://github.com/PaperMC/Paper/issues/11167) (not technically a bug, but still something that I think should be addressed).

The API is expanded by overloading existing `World#createExplosion` methods with a new `boolean shouldDamageSource` parameter. When `true`, the entity specified as the `source` will be damaged by the explosion. Existing behaviour is not changed.


lynxplay (changes requested):
Sorry for the long delay on the initial review, this PR seems to have fallen completely under my radar.

In general, when expanding a heavily overloaded API like this, we'd only want to add one more method.

In this case `public boolean createExplosion(@Nullable Entity source, @NotNull Location loc, float power, boolean setFire, boolean breakBlocks, boolean damageSource);`.
Callers of as niche of an API as the `damageSource` flag will be, can be expected to fully define all other parameters. The amount of methods added by this patch is way too many.

Beyond that, generally LGTM, but yea, please reduce this to a single added method on World, preferably just the method signature mentioned above. Your changes will also end up in the Expand-Explosions-API patch, so if you want to merge your changes into that patch directly, that would be great. Our contribution doc has a tutorial explaining how to edit patches.

Slqmy:
Thanks for reviewing. I'll make sure to change what you requested ASAP.

I'm currently on holidays, but I'll get to it as soon as I'm back.

Slqmy:
Sorry, didn't realise that would close the PR.

So far, I've reduced the PR to add a single method, as well as merge the changes into the Expand-Explosions-API patch, as requested.

However, the build is currently failing at the javadoc task, not sure why.

Slqmy:
I believe I have fixed the issue just now.

Slqmy:
I just tested this with the test plugin, and it does damage the source entity as intended, but there's an issue with the explosion sounds and particles, which I'll look into tomorrow.

Slqmy:
Just fixed the issue with particles and sound.

Slqmy:
Just tested this using the test plugin and everything works perfectly.

lynxplay:
The more I look at the diff the less happy I am with it.
The explosion invocation code is already ugly in vanilla with the large amounts of overload.
Having to add a boolean param to *that* many method seems annoying.

Not that we can really get around overloading internal methods here, but maybe a Consumer<Explosion> would be more suited to avoid further diff if we expand the API even more?

EsotericEnderman:
> The more I look at the diff the less happy I am with it. The explosion invocation code is already ugly in vanilla with the large amounts of overload. Having to add a boolean param to _that_ many method seems annoying.
> 
> Not that we can really get around overloading internal methods here, but maybe a Consumer would be more suited to avoid further diff if we expand the API even more?
> 
> E.g. something like these two https://pastebin.com/4c1ZhtL0 https://pastebin.com/J6QwPZ4Q

I implemented these changes and everything seems to work.

lynxplay (approved):

Machine-Maker (approved):

EsotericEnderman:
Thanks for merging!

I noticed you used my old username (Slqmy) instead of my new username (EsotericEnderman) when you added `Co-authored-by`, I don't mind too much, but at the same time why?

lynxplay:
Ah sorry, that was the one uses in the patch files you originally submitted. I'll change it up tonight if you can quickly give me your preferred name and email.

EsotericEnderman:
Sure thing, my preferred name is `Esoteric Enderman` and my preferred email is `90862990+EsotericEnderman@users.noreply.github.com`.

If you could change it to these in both of the patch files, that would be great! Thanks in advance.

EsotericEnderman:
> Ah sorry, that was the one uses in the patch files you originally submitted. I'll change it up tonight if you can quickly give me your preferred name and email.

@lynxplay any update on this...?

lynxplay:
https://github.com/PaperMC/Paper/commit/e3c8a8eef16eb2051b541cabc7419549c3e1e1f6

PR #11188, opened 2024-07-29
Fix potion effect overriding

TonytheMacaroni:
Removed arbitrary restrictions in `CraftAreaEffectCloud`, `CraftArrow` and `CraftMetaPotion` that prevented adding multiple potion effects of the same type. There seems to be no such restriction in vanilla, and multiple effects of the same type do apply properly.

During testing, I noticed an issue introduced by the handling of `EntityPotionEffectEvent` that prevented some potion effects from properly applying, when compared to vanilla. I wasn't entirely sure of the best approach to fix this - as such, I tried sticking to the Javadoc description of [`EntityPotionEffectEvent#isOverride`](https://jd.papermc.io/paper/1.21/org/bukkit/event/entity/EntityPotionEffectEvent.html#isOverride()), where if `override` is true, the previous effect is entirely replaced. As such, a new effect hiding a previous effect is currently not considered "overriding", and `EntityPotionEffectEvent#setOverride(true)` force applies the new effect, removing the old effect.

Also includes a fix for [MC-259832](https://bugs.mojang.com/browse/MC-259832). There's a remaining issue with the `/effect` command improperly considering applying an effect with a lower amplifier, but longer duration as compared to the current effect, as failing to apply, but I'm unsure of how best to fix that. It only impacts messaging, so I did not attempt to fix it.

### DIFF
--- a/patches/server/0678-More-Projectile-API.patch
+++ b/patches/server/0678-More-Projectile-API.patch
@@ -248,7 +248,7 @@ index 329ca9c743a7f2feeabbfb769ff9a71f60165006..faa08ad912fa43e7a6c5a2359e23c04c
 +    // Paper end
  }
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
-index 81f5e1d866128af8fb2acc13aca715580fdf9886..a988b8465a2c115414fb7ee540fa41bb8dcd6d9d 100644
+index 81f5e1d866128af8fb2acc13aca715580fdf9886..d6ee4870e562a7d53e34eb87d515499558553066 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftAreaEffectCloud.java
 @@ -125,7 +125,7 @@ public class CraftAreaEffectCloud extends CraftEntity implements AreaEffectCloud
@@ -268,7 +268,7 @@ index 81f5e1d866128af8fb2acc13aca715580fdf9886..a988b8465a2c115414fb7ee540fa41bb
 -            if (!override) {
 -                return false;
 -            }
-+        if (override && this.hasCustomEffect(effect.getType())) { // Paper - support multiple effects of the same type
++        if (override) { // Paper - support multiple effects of the same type
              this.removeCustomEffect(effect.getType());
          }
          this.getHandle().addEffect(CraftPotionUtil.fromBukkit(effect));
@@ -287,7 +287,7 @@ index 81f5e1d866128af8fb2acc13aca715580fdf9886..a988b8465a2c115414fb7ee540fa41bb
  
      @Override
 diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftArrow.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftArrow.java
-index 5232fbef0d014edd32a5d18d4a1500ab215313f5..0d5564d21e8c16d4d6fe6e4c23f672e635cf07ae 100644
+index 5232fbef0d014edd32a5d18d4a1500ab215313f5..2474c78c52e21e29a37555a19170e076271dbd54 100644
 --- a/src/main/java/org/bukkit/craftbukkit/entity/CraftArrow.java
 +++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftArrow.java
 @@ -36,14 +36,11 @@ public class CraftArrow extends CraftAbstractArrow implements Arrow {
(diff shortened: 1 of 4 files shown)


patches/server/0678-More-Projectile-API.patch
```diff
@@ -260,7 +260,15 @@ index 81f5e1d866128af8fb2acc13aca715580fdf9886..88f2a9f310f30a08893f3fa68af13a54
...
-@@ -143,7 +143,7 @@ public class CraftAreaEffectCloud extends CraftEntity implements AreaEffectCloud
+@@ -136,14 +136,11 @@ public class CraftAreaEffectCloud extends CraftEntity implements AreaEffectCloud
+ 
+     @Override
+     public boolean addCustomEffect(PotionEffect effect, boolean override) {
+-        if (this.hasCustomEffect(effect.getType())) {
+-            if (!override) {
+-                return false;
+-            }
++        if (override && this.hasCustomEffect(effect.getType())) { // Paper - support multiple effects of the same type
```
Lulu13022002: I don't think hasCustomEffect is needed here since removeCustomEffect already check that, same for the other method in CraftArrow.

patches/server/0750-Fix-a-bunch-of-vanilla-bugs.patch
```diff
@@ -149,6 +152,19 @@ index d430e6c08bae0a630e71c11c9eae26bef85bde36..a929b8f1c452db08a64e4be53b488cf2
...
+diff --git a/src/main/java/net/minecraft/world/effect/MobEffectInstance.java b/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
+index 959c0f82a3a7dae801914be5a126c87ac3b99a63..31a4f5716e646ea7cb4eead94b20694d97480709 100644
+--- a/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
++++ b/src/main/java/net/minecraft/world/effect/MobEffectInstance.java
+@@ -224,7 +224,7 @@ public class MobEffectInstance implements Comparable<MobEffectInstance> {
+             }
+ 
+             this.tickDownDuration();
+-            if (this.duration == 0 && this.hiddenEffect != null) {
++            while (this.duration == 0 && this.hiddenEffect != null) { // Paper - Fix MC-259832
```
Lulu13022002, line 164: Does the overwrite callback need to be run each time in the loop? Seems it will trigger the effect changed criteria for already expired effect too.

patches/server/0967-General-ItemMeta-fixes.patch
```diff
@@ -1375,7 +1384,39 @@ index 7f9182809f6e67ff571db0f365bc7e05f600775a..01c49df291f721bab3acb788ff2f2787
...
+-        int index = this.indexOfEffect(effect.getType());
+-        if (index != -1) {
+-            if (overwrite) {
++        // Paper start - support multiple effects of the same type
++        if (overwrite) {
++            int index = this.indexOfEffect(effect.getType());
++            if (index != -1) {
++        // Paper end - support multiple effects of the same type
+                 PotionEffect old = this.customEffects.get(index);
+                 if (old.getAmplifier() == effect.getAmplifier() && old.getDuration() == effect.getDuration() && old.isAmbient() == effect.isAmbient()) {
```
Lulu13022002: This restriction is weird, it doesn't check hasIcon and hasParticles

patches/server/1047-Fix-EntityPotionEffectEvent-effect-overriding.patch
```diff
@@ -0,0 +1,37 @@
...
++++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+@@ -1159,10 +1159,7 @@ public abstract class LivingEntity extends Entity implements Attackable {
+             boolean flag = false;
+ 
+             // CraftBukkit start
+-            boolean override = false;
+-            if (mobeffect1 != null) {
+-                override = new MobEffectInstance(mobeffect1).update(mobeffect);
+-            }
++            boolean override = false; // Paper start - properly override effects
```
Lulu13022002: This shouldn't be a // Paper start comment, you can move it down on the next line on the removed diff

patches/server/0967-General-ItemMeta-fixes.patch
Lulu13022002: Actually now that multiple effects of the same type is possible it might be better to do like the other methods and remove all the previous effects with the same type when overwrite is true and set the new effect at the first index found. I think the current behavior (pr) is confusing since overwrite will let other effects of the same type and will only override the first effect found.

patches/server/0678-More-Projectile-API.patch
TonytheMacaroni: Removed the unnecessary check.

patches/server/0750-Fix-a-bunch-of-vanilla-bugs.patch
TonytheMacaroni: Initially ignored it, but on second thought it is weird, so added a duration check before running the callback.

patches/server/0967-General-ItemMeta-fixes.patch
TonytheMacaroni: Implemented what you suggested.

patches/server/1047-Fix-EntityPotionEffectEvent-effect-overriding.patch
TonytheMacaroni: Moved the comment down one line.

Machine-Maker:
This changes some of the "add effect" methods that have a boolean return value to never return false. It feels like too much of an API break to make this change in methods where the expectation is something else. Perhaps new methods to take advantage of this functionality is better? smth like addCustomEffectWithMerge to clarify that it will "combine" effects utilizing the hidden effect functionality.

PR #11206, opened 2024-08-01
Make 'Velocity required kick message' configurable

Twarug:
This PR makes it possible to change a kick message send when connecting directly to a Velocity configured server.


kashike:
If your server is configured correctly this message should never be seen by the public. For this reason, the message is not configurable.

Twarug:
No in every scenario it is possible to hide backend server, making it inaccessible to outside world.
By adding this configuration, owners in such situation, could redirect players to right IP address/domain.

electronicboy:
exposing proxied servers to the public network is an unsupported configuration, and we are not inclined to support or provide configuration options which support or condone such setups.

PR #11219, opened 2024-08-07
Configuration for horizontal-only item merging

NewwindServer:
Most of the visual artifacts that result from having item merge radius above vanilla levels is from items merging vertically,
which realistically, only happens when a player is dropping items, or items are dropping from breaking a block.

Most of the scenarios where item merging makes sense involves the two item entities being on the same Y level. i.e on the ground next to each other.
This is even more apparent since paper fixed items being able to merge through blocks.

This patch allows us to configure items to only merge horizontally, which is what vanilla does.
This allows us to have both the reduced number of item entities a high item-merge radius provides,
without most of the visual artifacts caused by items merging vertically.


kennytv (approved):
The config change needs to go into the config patch

Machine-Maker (approved):

Issue #11214, opened 2024-08-05
Enchantments are ignored when damaging an ItemStack via API


JLyne:
### Expected behavior

Damaging ItemStacks via `ItemStack.damage` or `LivingEntity.damageItemStack` should either respect enchantments or have an option to do so.

### Observed/Actual behavior

Since #11063, enchantments are now always ignored when damaging an ItemStack via the API.

As far as I can see there is now no existing API available for damaging items while respecting enchantments. In addition, the documentation for `LivingEntity.damageItemStack` states that enchantments will be respected:

> This runs all logic associated with damaging an itemstack like gamemode and enchantment checks, events, stat changes, advancement triggers, and notifying clients to play break animations.

### Steps/models to reproduce

1. Add a plugin which damages an item via `ItemStack.damage` or `LivingEntity.damageItemStack`
2. Use the plugin to damage an item with an unbreaking enchantment
3. Trigger the plugin damage on the enchanted item

### Plugin and Datapack List

> pl
[18:30:16 INFO]: Server Plugins (1):
[18:30:16 INFO]: Paper Plugins:
[18:30:16 INFO]:  - Test

### Paper version

> ver
[18:31:05 INFO]: Checking version, please wait...
[18:31:06 INFO]: This server is running Paper version 1.21-124-master@df3b654 (2024-08-02T00:02:38Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
You are running the latest version

lynxplay:
Yea this is intended. Please open a new feature request for an overload that takes a toggle to additionally respect enchantments or alternatively effects.

Machine-Maker:
I think this actually is an issue. In 1.20.4 it took enchants into consideration, and the javadocs specifically say it does. And it's not a pain to fix, all the needed context for the new enchant system is available. Reopening...

PR #11233, opened 2024-08-10
Fix LivingEntity#damageItemStack javadocs

masmc05:
https://github.com/PaperMC/Paper/pull/11063 introduced behaviour changes that contradicts the javadocs, confirmed in https://github.com/PaperMC/Paper/issues/11214#issuecomment-2269574514 as intended changes. This pr fixes the javadocs to reflect the current behaviour.


Machine-Maker:
I re-opened https://github.com/PaperMC/Paper/issues/11214 cause I think that enchants actually should be taken into account. So probably don't need this anymore.

PR #11234, opened 2024-08-10
Fix damaging itemstacks without damage-component set

Boy0000:
Currently an item with MAX_DAMAGE-component but without DAMAGE-component is not considered damageable.
Thus trying to use Player#damageItemstack will have no effect, when logically an item with max-damage should be considered damageable.
This issue doesnt apply to vanilla tools as they by default have damage-component
Whilst adding damage-component with 0 damage does fix this, this feels like a bug as I cant imagine a scenario where this behaviour would be wanted

This PR fixes it by ignoring the damage-component, which will be set later on down after damage-calculations
image

Before fix trying to call Player#damageItemstack (with only maxDamage component):

https://github.com/user-attachments/assets/a9cb7a59-6a5a-47a4-aca6-8d3e5053ecf8

After fix calling Player#damageItemstack on item with max-damage but wihtout damage:

https://github.com/user-attachments/assets/03eb7515-65df-4fd2-8a4e-6490fbc76093

### DIFF
--- a/patches/server/0773-ItemStack-damage-API.patch
+++ b/patches/server/0773-ItemStack-damage-API.patch
@@ -11,9 +11,18 @@ the logic associated with damaging them
 public net.minecraft.world.entity.LivingEntity entityEventForEquipmentBreak(Lnet/minecraft/world/entity/EquipmentSlot;)B
 
 diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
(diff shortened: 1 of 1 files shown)


Machine-Maker:
This fix, as is, breaks the vanilla functionality of creating a custom item specifically removing the "damage" component and having that be an "undamageable" item.

Also, the correct fix is for API uses to also set a damage value when setting the max damage value. Calling Damageable#setDamage with a param value of 0 will add the damage component. This is assuming its an item type that doesn't implicitly have a damage component. Closing as a works as intended with the fix being to set both max damage and damage in the item's meta.

PR #11235, opened 2024-08-11
Add Item serialization as json api

masmc05:
Example output: https://pastes.dev/nXEnpQ8K3Q

Since 1.21, vanilla item stack codecs now fully support such a json format for deserialization without any possible data loss, the support was added for data packs.

This pr adds a possibility to deserialize and serialize to such format which will not have any possible data loss due to loss of precise types (like ints becoming bytes) by allowing serialization of custom data as snbt. Also, those api methods will process the items trough data converters.

Unlike ItemStack#serializeAsBytes, this will result in a much more human readable format, which also may be transformed to many other formats like yaml or used with libraries like Jackson for human interaction with the final result.

An alternative to this pr would be adding an api to serialize the ItemStack to snbt, but if the snbt is put in a yaml or json file, it will not receive any kind of support from any IDE while editing the snbt, and because of that will become hard to work with. From just only api sadly it will not be possible to fix this.


Machine-Maker:
Without putting too much thought into this... I wonder if it's better to expose a gson TypeAdapter or serialize/deserialize pair for people to use within their own Gson setups.

FireInstall:
In the ItemStack class, there should be a javadoc note at `#serialize()` and `#deserialize()` to stop using it and instead point to this in case human readability is needed or `#serializeAsBytes() `and `#deserializeBytes()` if not. (Can it even be deprecated, as there is a better alternative now?)
Also, a `@see` to link both byte and json methods together. Just to reduce confusion.

FireInstall:
Well, there is no harm to remove it, if the serialisation get reworked to snbt. But for now there is no argument to be made why anyone should still use it.

patches/server/1048-Item-serialization-as-json.patch
```diff
@@ -0,0 +1,88 @@
...
++    }
++
++    public static com.google.gson.JsonObject serializeAsJson(net.minecraft.world.item.ItemStack handle) {
++        Preconditions.checkNotNull(handle, "Cannot serialize empty ItemStack");
++        Preconditions.checkArgument(!handle.isEmpty(), "Cannot serialize empty ItemStack");
++
++        net.minecraft.core.RegistryAccess.Frozen reg = net.minecraft.server.MinecraftServer.getServer().registryAccess();
++        com.mojang.serialization.DynamicOps<com.google.gson.JsonElement> ops = reg.createSerializationContext(com.mojang.serialization.JsonOps.INSTANCE);
++        com.google.gson.JsonObject item;
++        net.minecraft.world.item.component.CustomData.SERIALIZE_CUSTOM_AS_SNBT.set(true);
```
kennytv: I assume this is done to properly retain types? if so, you should add a comment explaining that here

kennytv (approved):
Generally looks good

patches/server/1048-Item-serialization-as-json.patch
masmc05: Added the comment

Machine-Maker:
@masmc05 I think that the "human-readable"-ness of this is really not that useful, especially when this would be a data-versioned format. Say you serialize an ItemStack on this version and it's got whatever version, say 1000. Then, in 3 years you go to deserialize it, but you want to change it first. 3 years from now, the format for itemstacks may have changed drastically, but you have to go back and know **exactly** what the format is for your specific data version. That information isn't readily available. I think it being human readable is really not worth much. Do you have a specific reason why human readable would be good?

electronicboy:
I do wish that there was a "human readable" standard agreed upon, but, json tangengially datapack related stuff is IMHO, not it. I do understand the arguments for having it in favour of people who want to bite that bullet to deal with, but, as said, it's a raw low level type of deal which can/will change often without warning, datapacks themselves already break once or twice a year as-is.

snbt based serialisation in configs is one of those areas of, yea, sure, you can tweak SNBT, but it's still a fairly low-level representation of the data stored.

There is never going to be a real good, long term solution for dealing with representing consistently evolving data. you could agree on the forefront, but, that would be library material, not built-in API type of stuff.

Malfrador:
I don't have a strong opinion on JSON or SNBT, both seem fine to me. I generally believe such a system would be very useful. 

But I don't really think the format changing between versions is as much of an issue. One of the main use cases of such a PR, at least for me, would be purely _reading_ items from configs. Being able to write an item definition inside a config is preferable over long and unwieldy commands to create such items in-game in my opinion, those get really confusing and too long to enter in the chat box very fast. It also allows for easier version control than base64 encoded strings from serialized in-game items. 
The Bukkit format sucks, and doesn't nicely represent the game anymore. So having some replacement, be it SNBT or JSON, for this usecase would be nice. 

And versioning really doesn't matter here that much - it is kinda expected that if Minecraft changes something my items use, I will have to update my config file with those items. That's the same thing that would happen with command blocks, or data packs and is fine imo.

patches/api/0487-Item-serialization-as-json.patch
```diff
@@ -0,0 +1,52 @@
...
+     // Paper start - spawn egg color visibility
+diff --git a/src/main/java/org/bukkit/inventory/ItemFactory.java b/src/main/java/org/bukkit/inventory/ItemFactory.java
+index bd0e55562f1cabef3078573182e0cf9fbc844585..fcbfc97eb5eeeccd45e25b61cd05127ab940b424 100644
+--- a/src/main/java/org/bukkit/inventory/ItemFactory.java
++++ b/src/main/java/org/bukkit/inventory/ItemFactory.java
+@@ -149,6 +149,7 @@ public interface ItemFactory {
+     @NotNull
+     ItemStack createItemStack(@NotNull String input) throws IllegalArgumentException;
+ 
++
```
Leguan16: Very important diff 😅

masmc05: idk how i've missed it, thanks!

Machine-Maker (approved):

ryantheleach:
My sincere thanks for merging this.

Having item's serialized out as a standardized human readable and editable format is great, and I don't think it matters all that much about the versioning worries people had above.

The **main reason** why backwards compatibility is critical with Minecraft Plugins, is that they often go unmaintained, and having them be valid for as long as possible helps the ecosystem immensely.

By outputting the items in the 'raw' format, you pass on the responsibility for keeping it up to date to both the dataconverters if it's got versioning information and the user.

Informed users can readily find the latest format to update configs, but rarely is it convenient to fork and upgrade plugins.

Having this, means that a plugin can exist that just loads items into minecraft from text, players can use whatever means they like to edit them, then serialize them back out for configs.

It means that a shop plugin using this, won't need to create it's own format, or reach into NMS, and that admin's won't need to learn the archaic bukkit serialization format, as the json is evergreen.

PR #11244, opened 2024-08-12
Change Art to an interface

### LINKED ISSUE #11237: IllegalArgumentException when using Painting#getArt() on custom painting variant
### Expected behavior

1.21 introduced support for data driven custom paintings. I didn't expect them to throw an exception.

### Observed/Actual behavior

You get the following stacktrace when trying to run the method.
[stacktrace.txt](https://github.com/user-attachments/files/16573260/stacktrace.txt)

### Steps/models to reproduce

1. Install the TestPainting datapack on the server 
[TestPainting.zip](https://github.com/user-attachments/files/16573266/TestPainting.zip)
2. Give yourself the custom painting with `/give @p painting[entity_data={id:"minecraft:painting",variant:"test:new_painting"}]`
3. Run Painting#getArt(), e.g., with this event listener:
```
    @EventHandler
    public void onHangingPlace(HangingPlaceEvent event) {
        Entity entity = event.getEntity();
        if (!(entity instanceof Painting painting)) {
            return;
        }

        painting.getArt();
    }
}
```

### Plugin and Datapack List

Only the TestPainting datapack and a plugin for the event listener.

### Paper version

This server is running Paper version 1.21.1-5-master@7d7f123 (2024-08-10T22:35:17Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21-49-b0ef3ca (MC: 1.21)

### Other

_No response_

Machine-Maker:
Fixes https://github.com/PaperMC/Paper/issues/11237


patches/server/1046-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,209 @@
...
++    private final org.bukkit.NamespacedKey key;
++    private final PaintingVariant paintingVariant;
++    private final String name;
++    private final int ordinal;
+ 
+-        Preconditions.checkArgument(bukkit != null);
++    public CraftArt(final org.bukkit.NamespacedKey key, final PaintingVariant paintingVariant) {
++        this.key = key;
++        this.paintingVariant = paintingVariant;
++        // For backwards compatibility, minecraft values will stile return the uppercase name without the namespace,
```
kennytv: still

patches/server/1046-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,209 @@
...
+diff --git a/src/test/java/org/bukkit/ArtTest.java b/src/test/java/org/bukkit/ArtTest.java
+index ea288c3ff30656569abd60a2839c470cb5d953f2..f6c5a03d4f41fc16aaa830ef96c66522fee0cf4a 100644
+--- a/src/test/java/org/bukkit/ArtTest.java
++++ b/src/test/java/org/bukkit/ArtTest.java
+@@ -57,7 +57,7 @@ public class ArtTest extends AbstractTestingBase {
+ 
+     @Test
+     public void testCraftArtToBukkit() {
+-        Map<Art, Holder<PaintingVariant>> cache = new EnumMap(Art.class);
++        Map<Art, Holder<PaintingVariant>> cache = new java.util.HashMap<>(); // Paper
```
kennytv: Is the duplicate assertion even needed anymore?

kennytv (approved):

paper-api-generator/src/main/java/io/papermc/generator/Generators.java
```diff
@@ -36,6 +37,7 @@ public interface Generators {
         simpleKey("DamageTypeKeys", DamageType.class, Registries.DAMAGE_TYPE, RegistryKey.DAMAGE_TYPE, true),
         simpleKey("WolfVariantKeys", Wolf.Variant.class, Registries.WOLF_VARIANT, RegistryKey.WOLF_VARIANT, true),
         simpleKey("ItemTypeKeys", ItemType.class, Registries.ITEM, RegistryKey.ITEM, false),
+        simpleKey("PaintingVariantKeys", Art.class, Registries.PAINTING_VARIANT, RegistryKey.PAINTING_VARIANT, true),
```
Lulu13022002, line 55: Can you add missing api like banner patterns, jukebox song etc?

patches/server/1046-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,209 @@
...
++    private final String name;
++    private final int ordinal;
+ 
+-        Preconditions.checkArgument(bukkit != null);
++    public CraftArt(final org.bukkit.NamespacedKey key, final PaintingVariant paintingVariant) {
++        this.key = key;
++        this.paintingVariant = paintingVariant;
++        // For backwards compatibility, minecraft values will stile return the uppercase name without the namespace,
++        // in case plugins use for example the name as key in a config file to receive pattern type specific values.
++        // Custom pattern types will return the key with namespace. For a plugin this should look than like a new pattern type
```
Lulu13022002: painting variant not pattern type

patches/api/0483-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,287 @@
...
+-    }
++    @Deprecated(forRemoval = true, since = "1.21.1") // Paper - deprecate
++    public int getId(); // Paper - make interface
+ 
+     @NotNull
+     @Override
+-    public NamespacedKey getKey() {
+-        return key;
+-    }
++    public NamespacedKey getKey(); // Paper - make interface
```
Lulu13022002: Should this be deprecated here or in the other patch for inlined painting variant?

patches/api/0483-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,287 @@
...
++     */
++    @Deprecated(since = "1.21.1", forRemoval = true)
++    static @NotNull Art valueOf(final @NotNull String name) {
++        final Art art = Registry.ART.get(NamespacedKey.fromString(name.toLowerCase(Locale.ROOT)));
++        Preconditions.checkArgument(art != null, "No art found with the name %s", name);
++        return art;
++    }
++
++    /**
++     * @deprecated Use {@link Registry#ART}
```
Lulu13022002: First letter case is not consistent with above

patches/api/0483-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,287 @@
...
+-        this.id = id;
+-        this.width = width;
+-        this.height = height;
+-        this.key = NamespacedKey.minecraft(name().toLowerCase(Locale.ROOT));
++    /**
++     * @deprecated use {@link Registry#ART}
++     */
++    @Deprecated(since = "1.21.1", forRemoval = true)
++    static @NotNull Art valueOf(final @NotNull String name) {
++        final Art art = Registry.ART.get(NamespacedKey.fromString(name.toLowerCase(Locale.ROOT)));
```
Lulu13022002: NamespacedKey.fromString is nullable this should be checked first before blindly getting the value from the registry, same for other similar impl otherwise a NPE is thrown.

patches/api/0483-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,287 @@
...
++        Preconditions.checkArgument(art != null, "No art found with the name %s", name);
++        return art;
++    }
++
++    /**
++     * @deprecated Use {@link Registry#ART}
++     */
++    @Deprecated(since = "1.21.1", forRemoval = true)
++    static Art[] values() {
++        return com.google.common.collect.Lists.newArrayList(Registry.ART).toArray(Art[]::new);
```
Lulu13022002: This is weird why not directly use the stream to avoid the double transition?

Machine-Maker: turns out I should've looked more into how upstream was doing these transitions. I naively just copied what they had done to other enums.

paper-api-generator/src/main/java/io/papermc/generator/Generators.java
Machine-Maker: just gonna hold off on jukebox songs. We have to update the "naming" because there are 3 pure-number songs, like 11, 13, 5. Need to add logic to convert those to the spelled out versions for field names.

patches/api/0487-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,278 @@
...
+      * @return The painting
++     * @deprecated use {@link io.papermc.paper.registry.RegistryAccess#getRegistry(io.papermc.paper.registry.RegistryKey)} and {@link io.papermc.paper.registry.RegistryKey#PAINTING_VARIANT}
+      */
+     @Nullable
++    @Deprecated(since = "1.21.1") // Paper - deprecate
+     public static Art getByName(@NotNull String name) {
+         Preconditions.checkArgument(name != null, "Name cannot be null");
+ 
+-        return BY_NAME.get(name.toLowerCase(Locale.ROOT));
++        return registry().get(NamespacedKey.minecraft(name)); // Paper - replace impl
```
Lulu13022002: I think the lowercase on the name is still needed like before to respect the javadoc and to allow Art.getByName(Art.X.name()).

patches/api/0487-Change-Art-to-an-interface.patch
```diff
@@ -0,0 +1,278 @@
...
++        Preconditions.checkState(art != null, "Missing art: %s", id);
++        return art;
++    }
++
++    /**
++     * @deprecated use {@link io.papermc.paper.registry.RegistryAccess#getRegistry(io.papermc.paper.registry.RegistryKey)} and {@link io.papermc.paper.registry.RegistryKey#PAINTING_VARIANT}
++     */
++    @Deprecated(since = "1.21.1", forRemoval = true)
++    static @NotNull Art valueOf(final @NotNull String name) {
++        final NamespacedKey key = NamespacedKey.fromString(name.toLowerCase(Locale.ROOT));
```
Lulu13022002: Technically this should only allow previous name of the enum in upper case, but this works on lower/mixed case too here.

Lulu13022002 (commented):
Maybe add a test in RegistryConstantsTest to make sure some constant are not missed

patches/server/1046-Change-Art-to-an-interface.patch
Lulu13022002: This is not resolved yet

paper-api-generator/src/main/java/io/papermc/generator/Generators.java
```diff
@@ -8,35 +8,54 @@
...
+
+        // data-driven
+        simpleKey("BiomeKeys", Biome.class, Registries.BIOME, RegistryKey.BIOME, true),
+        simpleKey("StructureKeys", Structure.class, Registries.STRUCTURE, RegistryKey.STRUCTURE, true),
+        simpleKey("TrimMaterialKeys", TrimMaterial.class, Registries.TRIM_MATERIAL, RegistryKey.TRIM_MATERIAL, true),
+        simpleKey("TrimPatternKeys", TrimPattern.class, Registries.TRIM_PATTERN, RegistryKey.TRIM_PATTERN, true),
         simpleKey("DamageTypeKeys", DamageType.class, Registries.DAMAGE_TYPE, RegistryKey.DAMAGE_TYPE, true),
         simpleKey("WolfVariantKeys", Wolf.Variant.class, Registries.WOLF_VARIANT, RegistryKey.WOLF_VARIANT, true),
-        simpleKey("ItemTypeKeys", ItemType.class, Registries.ITEM, RegistryKey.ITEM, false),
+        simpleKey("EnchantmentKeys", Enchantment.class, Registries.ENCHANTMENT, RegistryKey.ENCHANTMENT, false),
```
Lulu13022002, line 53: Any reason this one data driven registry doesn't allow creation of custom key like others? the doc website could then be updated

paper-api-generator/src/main/java/io/papermc/generator/Generators.java
```diff
@@ -8,35 +8,54 @@
...
 import org.bukkit.inventory.meta.trim.TrimMaterial;
 import org.bukkit.inventory.meta.trim.TrimPattern;
+import org.bukkit.map.MapCursor;
 import org.bukkit.potion.PotionEffectType;
 
 public interface Generators {
 
     SourceGenerator[] API = {
+        // built-in
         simpleKey("GameEventKeys", GameEvent.class, Registries.GAME_EVENT, RegistryKey.GAME_EVENT, false),
```
Lulu13022002, line 34: GameEvent has a registry event so the create method could be public i think

Machine-Maker:
Superceeded by https://github.com/PaperMC/Paper/pull/11561

PR #11254, opened 2024-08-13
Fix `TooltipContext.create` being wrong (#11253)

### LINKED ISSUE #11253: TooltipContext - Advanced/Creative are backwards
### Expected behavior

I would expect these to be correctly done.

### Observed/Actual behavior

The advanced/creative options are backwards in TooltipContext
Screenshot 2024-08-12 at 8 43 49 PM

### Steps/models to reproduce

n/a - see image above

### Plugin and Datapack List

n/a

### Paper version

n/a

### Other

_No response_

hyperdefined:
Swaps to the correct order, closes #11253.


lynxplay:
I think it would be beneficial to instead swap around the argument order on the impl record.
That way we don't have this backwards order in our code.

kennytv (approved):

lynxplay (approved):

PR #11255, opened 2024-08-13
[ci skip] Specify rebase location in CONTRIBUTING

hyperdefined:
As someone who doesn't understand how git works super well, I was confused when trying to rebase. Turns out I was in the wrong folder, and thought this could be added to help others understand where they need to rebase.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,6 @@ split into different directories which target certain parts of the code. These
 directories are:
 
 - `Paper-API` - Modifications to `Spigot-API`/`Bukkit`;
-- `Paper-MojangAPI` - An API for [Mojang's Brigadier](https://github.com/Mojang/brigadier);
 - `Paper-Server` - Modifications to `Spigot`/`CraftBukkit`.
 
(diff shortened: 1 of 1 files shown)


lynxplay:
All of the commands in these sections are to be run in the respective -server or -api directories.
At best we could add something to the 
```
Modifying Patches
Modifying previous patches is a bit more complex:

```
block saying that the below commands, unless specified otherwise, are ran in the respective -api or -server folders.

Leguan16:
> All of the commands in these sections are to be run in the respective -server or -api directories.
> 
> At best we could add something to the 
> 
> ```
> 
> Modifying Patches
> 
> Modifying previous patches is a bit more complex:
> 
> 
> 
> ```
> 
> block saying that the below commands, unless specified otherwise, are ran in the respective -api or -server folders.

That what I wanted to propose. But then I thought, it's only confusing until you understand patches. And that each patch is a commit to the -API, -Server module, and when you understand that, I think it's clear that you need to run these commands in those directories.

lynxplay (approved):

Machine-Maker (approved):

PR #11270, opened 2024-08-15
add ItemFrameMeta API

TehBrian:
Draft PR; CraftMetaItemFrame is (probably) currently non-functional, but I wanted to share progress so far.

The second commit is a rewrite based on #11107, and I'm waiting on that PR to finalize changes here.

Concerns:
- Do we have to care about entity ID being different for glow item frames? Does that warrant a separate GlowItemFrameMeta? Entity IDs only even seem to matter on spawn eggs. 
- Should CraftMetaItemFrame be wrapped in `// Paper`? It's in `org.bukkit` (because there were a nightmarish amount of access transformers until I moved it from `io.papermc`), but it's also a new file.

Please let me know your thoughts :D


TehBrian:
oopsie :3

PR #11271, opened 2024-08-15
add ItemFrameMeta API

TehBrian:
~~Draft PR; CraftMetaItemFrame is (probably) currently non-functional, but I wanted to share progress so far.~~

The second commit is a rewrite based on https://github.com/PaperMC/Paper/pull/11107, ~~and I'm waiting on that PR to finalize changes here.~~

Concerns:
- Do we have to care about entity ID being different for glow item frames? Does that warrant a separate GlowItemFrameMeta? Entity IDs only even seem to matter on spawn eggs.
- Should CraftMetaItemFrame be wrapped in // Paper? It's in org.bukkit (because there were a nightmarish amount of access transformers until I moved it from io.papermc), but it's also a new file.

Please let me know your thoughts :D

### DIFF
--- a/patches/api/0484-add-ItemFrameMeta-API.patch
+++ b/patches/api/0484-add-ItemFrameMeta-API.patch
@@ -1,0 +1,88 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TehBrian <tehbrian@proton.me>
+Date: Thu, 15 Aug 2024 17:06:29 -0400
+Subject: [PATCH] add ItemFrameMeta API
+
+
+diff --git a/src/main/java/io/papermc/paper/inventory/meta/ItemFrameMeta.java b/src/main/java/io/papermc/paper/inventory/meta/ItemFrameMeta.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..65b388e0a48a5927448a79b0f2e7742895038034
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/inventory/meta/ItemFrameMeta.java
+@@ -0,0 +1,53 @@
++package io.papermc.paper.inventory.meta;
++
++import org.bukkit.inventory.meta.ItemMeta;
++
++public interface ItemFrameMeta extends ItemMeta {
++    /**
++     * Gets whether the item frame is fixed.
++     *
++     * <p>Fixed item frames:</p>
++     * <ul>
++     *     <li>Cannot have its contents modified or rotated.</li>
++     *     <li>Will not break in absence of a supporting block.</li>
++     *     <li>Will not take damage (except by players in creative).</li>
++     *     <li>Cannot be moved by pistons.</li>
++     * </ul>
++     *
++     * @return whether the item frame is fixed
++     */
++    boolean isFixed();
++
++    /**
++     * Sets whether the item frame is fixed.
++     *
++     * <p>Fixed item frames:</p>
++     * <ul>
++     *     <li>Cannot have its contents modified or rotated.</li>
++     *     <li>Will not break in absence of a supporting block.</li>
++     *     <li>Will not take damage (except by players in creative).</li>
++     *     <li>Cannot be moved by pistons.</li>
++     * </ul>
++     *
++     * @param fixed true if the item frame is fixed
++     */
++    void setFixed(boolean fixed);
++
++    /**
++     * Gets whether the item frame's <b>frame</b> (background) is invisible.
++     *
++     * <p>Contents (items, maps) within invisible item frames are still visible.</p>
++     *
++     * @return whether the frame is invisible
++     */
++    boolean isInvisible();
++
++    /**
++     * Sets whether the item frame's <b>frame</b> (background) is invisible.
++     *
++     * <p>Contents (items, maps) within invisible item frames are still visible.</p>
++     *
++     * @param invisible true if the frame is invisible
++     */
++    void setInvisible(boolean invisible);
++}
+diff --git a/src/main/java/org/bukkit/inventory/ItemType.java b/src/main/java/org/bukkit/inventory/ItemType.java
+index 0168f0a14a3e899e84c5e36963ff79950ab580fb..b52c871297fb351e3781fe6163790def4f95f17f 100644
+--- a/src/main/java/org/bukkit/inventory/ItemType.java
++++ b/src/main/java/org/bukkit/inventory/ItemType.java
+@@ -1733,8 +1733,16 @@ public interface ItemType extends Keyed, Translatable, net.kyori.adventure.trans
+      */
+     ItemType.Typed<BookMeta> WRITTEN_BOOK = getItemType("written_book");
+     ItemType.Typed<ItemMeta> MACE = getItemType("mace");
+-    ItemType.Typed<ItemMeta> ITEM_FRAME = getItemType("item_frame");
+-    ItemType.Typed<ItemMeta> GLOW_ITEM_FRAME = getItemType("glow_item_frame");
++    // Paper start - add ItemFrameMeta API
++    /**
++     * ItemMeta: {@link io.papermc.paper.inventory.meta.ItemFrameMeta}
++     */
++    ItemType.Typed<io.papermc.paper.inventory.meta.ItemFrameMeta> ITEM_FRAME = getItemType("item_frame");
++    /**
++     * ItemMeta: {@link io.papermc.paper.inventory.meta.ItemFrameMeta}
++     */
++    ItemType.Typed<io.papermc.paper.inventory.meta.ItemFrameMeta> GLOW_ITEM_FRAME = getItemType("glow_item_frame");
++    // Paper end - add ItemFrameMeta API
+     ItemType.Typed<ItemMeta> FLOWER_POT = getItemType("flower_pot");
+     ItemType.Typed<ItemMeta> CARROT = getItemType("carrot");
+     ItemType.Typed<ItemMeta> POTATO = getItemType("potato");
--- a/patches/server/1046-add-ItemFrameMeta-API.patch
+++ b/patches/server/1046-add-ItemFrameMeta-API.patch
@@ -1,0 +1,314 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TehBrian <tehbrian@proton.me>
+Date: Thu, 15 Aug 2024 17:06:38 -0400
+Subject: [PATCH] add ItemFrameMeta API
+
+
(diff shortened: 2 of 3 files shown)


TehBrian:
#11107 has been merged. I'll clean this up, test it, and then mark it ready for review.

patches/server/1046-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,320 @@
...
++    static final ItemMetaKey ENTITY_ID = new ItemMetaKey("id", "entity-id");
++    static final ItemMetaKey FIXED = new ItemMetaKey("Fixed", "fixed");
++    static final ItemMetaKey INVISIBLE = new ItemMetaKey("Invisible", "invisible");
++
++    CraftMetaItemFrame(CraftMetaItem meta) {
++        super(meta);
++
++        if (meta instanceof CraftMetaItemFrame itemFrame) {
++            this.entityTag = itemFrame.entityTag;
++        }
```
TehBrian: Other CraftMeta classes still use a guard and manual casting. This is more concise and (imo) easier to read, but should we stick to what the other classes are doing for consistency?

patches/server/1046-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,320 @@
...
++            // Paper start - add ItemFrameMeta API
++            new StackProvider(Material.ITEM_FRAME) {
++                @Override ItemStack operate(ItemStack cleanStack) {
++                    final CraftMetaItemFrame meta = ((CraftMetaItemFrame) cleanStack.getItemMeta());
++                    meta.setFixed(true);
++                    meta.setInvisible(true);
+                     cleanStack.setItemMeta(meta);
+                     return cleanStack;
+                 }
+             },
```
TehBrian: Messy patch. Going to recommit this such that the test between `Paper start` and `Paper end` is a single addition.

patches/server/1046-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,320 @@
...
++
++@DelegateDeserialization(SerializableMeta.class)
++public class CraftMetaItemFrame extends CraftMetaItem implements ItemFrameMeta {
++
++    private static final Set<Material> ENTITY_TAGGABLE_MATERIALS = Sets.newHashSet(
++        Material.ITEM_FRAME,
++        Material.GLOW_ITEM_FRAME
++    );
++
++    static final ItemMetaKeyType<CustomData> ENTITY_TAG = new ItemMetaKeyType<>(DataComponents.ENTITY_DATA, "EntityTag", "entity-tag");
```
TehBrian: Is NBT supposed to be `entity-tag` instead of `EntityTag`?

patches/server/1046-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,320 @@
...
++    @Override
++    int applyHash() {
++        final int original;
++        int hash = original = super.applyHash();
++
++        if (this.entityTag != null) {
++            hash = 73 * hash + this.entityTag.hashCode();
++        }
++
++        return original != hash ? CraftMetaArmorStand.class.hashCode() ^ hash : hash;
```
TehBrian: NO

patches/server/1046-add-ItemFrameMeta-API.patch
TehBrian: mmmmm this is annoying to do and I don't think anyone cares

patches/server/1046-add-ItemFrameMeta-API.patch
kennytv: Nah no need to stay consistent with what's just old language code

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
+                 map.put(CraftMetaFirework.class, Set.of(CraftMetaFirework.FIREWORKS.TYPE));
++                map.put(CraftMetaItemFrame.class, Set.of(CraftMetaItemFrame.ENTITY_TAG.TYPE)); // Paper - add ItemFrameMeta API
+                 map.put(CraftMetaKnowledgeBook.class, Set.of(CraftMetaKnowledgeBook.BOOK_RECIPES.TYPE));
+                 map.put(CraftMetaLeatherArmor.class, Set.of(CraftMetaLeatherArmor.COLOR.TYPE));
+                 map.put(CraftMetaMap.class, Set.of(CraftMetaMap.MAP_COLOR.TYPE, CraftMetaMap.MAP_POST_PROCESSING.TYPE, CraftMetaMap.MAP_ID.TYPE));
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItemFrame.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItemFrame.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..b13444c63c37e5c9f4ff969ad473f9280a961332
+--- /dev/null
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaItemFrame.java
```
DerEchtePilz, line 75: Isn't it better to put this into some paper package since this is not a CraftBukkit addition?

TehBrian: I mentioned that up above in my original PR. The issue is that putting it in a different package requires stupid amounts of access transformers on CraftMetaItem and making nearly everything in CraftMetaItemFrame itself public. I'm still open to opinions, though.

DerEchtePilz: Well, I personally would probably put it into a paper package but I would also say that input by developers is needed here.

lynxplay: Its fine in the CB package.

patches/server/1046-add-ItemFrameMeta-API.patch
kennytv: entity-tag for both, so you can remove that param

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
++import net.minecraft.nbt.NbtIo;
++import net.minecraft.nbt.Tag;
++import net.minecraft.world.item.component.CustomData;
++import org.bukkit.Material;
++import org.bukkit.configuration.serialization.DelegateDeserialization;
++
++@DelegateDeserialization(SerializableMeta.class)
++public class CraftMetaItemFrame extends CraftMetaItem implements ItemFrameMeta {
++
++    private static final Set<Material> ENTITY_TAGGABLE_MATERIALS = Sets.newHashSet(
```
kennytv, line 106: Set.of

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
++            if (!this.entityTag.contains(ENTITY_ID.NBT)) {
++                this.entityTag.putString(ENTITY_ID.NBT, "minecraft:item_frame");
++            }
++        }
++    }
++
++    @Override
++    void serializeInternal(Map<String, Tag> internalTags) {
++        // correctly serialised as entity tag.
++    }
```
Machine-Maker, line 172: Just don't override this at all.

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
++            return;
++        }
++        SerializableMeta.getObjectOptionally(Boolean.class, map, FIXED.BUKKIT, true).ifPresent((value) -> {
++            populateTagIfNull();
++            this.entityTag.putBoolean(FIXED.NBT, value);
++        });
++        SerializableMeta.getObjectOptionally(Boolean.class, map, INVISIBLE.BUKKIT, true).ifPresent((value) -> {
++            populateTagIfNull();
++            this.entityTag.putBoolean(INVISIBLE.NBT, value);
++        });
```
Machine-Maker, line 152: None of this is needed. We are always serializing it as a full nbt tag instead of splitting it up. The only reason ArmorStand has similar logic is were weren't always doing that. But since this is brand new, we can do this.

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
++    CraftMetaItemFrame(Map<String, Object> map) {
++        super(map);
++
++        String entityTag = SerializableMeta.getString(map, ENTITY_TAG.BUKKIT, true);
++        if (entityTag != null) {
++            ByteArrayInputStream buf = new ByteArrayInputStream(Base64.getDecoder().decode(entityTag));
++            try {
++                this.entityTag = NbtIo.readCompressed(buf, NbtAccounter.unlimitedHeap());
++            } catch (IOException ex) {
++                Logger.getLogger(CraftMetaItem.class.getName()).log(Level.SEVERE, null, ex);
```
Machine-Maker, line 141: Lets make this an slf4j logger in a private static field like `LogUtils.getClassLogger()`

patches/api/0484-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,88 @@
...
+new file mode 100644
+index 0000000000000000000000000000000000000000..65b388e0a48a5927448a79b0f2e7742895038034
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/inventory/meta/ItemFrameMeta.java
+@@ -0,0 +1,53 @@
++package io.papermc.paper.inventory.meta;
++
++import org.bukkit.inventory.meta.ItemMeta;
++
++public interface ItemFrameMeta extends ItemMeta {
```
Machine-Maker, line 17: Are we making the same mistake here lots of other ItemMetas have done? technically there are 3 states for every boolean. It could be true, false, or not present in the custom data.

patches/server/1048-add-ItemFrameMeta-API.patch
```diff
@@ -0,0 +1,323 @@
...
++        // correctly serialised as entity tag.
++    }
++
++    @Override
++    protected void applyToItem(CraftMetaItem.Applicator tag) {
++        super.applyToItem(tag);
++
++        if (this.entityTag != null) {
++            tag.put(CraftMetaItemFrame.ENTITY_TAG, CustomData.of(this.entityTag));
++        }
```
Machine-Maker, line 180: We probably need to do a sanity check here to make sure the custom data compound, if its not empty, that it has the "id" key/value in it. If it doesn't and its not empty, it'll throw errors all over the place as the codec requires it.

Machine-Maker (changes requested):

TehBrian:
Is this PR still needed with the new item component API?

lynxplay:
Yea, the component API does not expose entity data.

PR #11279, opened 2024-08-16
Add Configuration for vertical Despawn Ranges

### LINKED ISSUE #10177: Add a separate Y-value check for "despawn-ranges"

notTamion:
resolves #10177 
Docs-pr at: PaperMC/docs#465

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1411,7 +1411,7 @@ index 0000000000000000000000000000000000000000..990d1bb46e0f9719f4e9af928d80ac6f
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..b5f1c369f727161fba8ae6a820cf1f901953d6f8
+index 0000000000000000000000000000000000000000..6d97ea32829f1d22923d6c38f75446de59701b18
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 @@ -0,0 +1,556 @@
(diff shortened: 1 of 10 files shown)


kennytv (approved):

lynxplay (approved):

19MisterX98:
This should fall back to the vanilla despawn sphere if no values are provided. Instead it takes the vanilla default range and checks horizontal and vertical separately. Creating a cylinder that is much larger than the vanilla sphere.

Machine-Maker:
> This should fall back to the vanilla despawn sphere if no values are provided. Instead it takes the vanilla default range and checks horizontal and vertical separately. Creating a cylinder that is much larger than the vanilla sphere.

Yeah, that makes sense. Will change it to default to checking the ellipsoid shape, and add a configuration to make it a cylinder as well.

PR #11282, opened 2024-08-17
Add enchantment seed update API

okx-code:
It's already possible to control the player's enchantment seed from 7cd3244d, but the enchantment seed in the actual enchanting menu cannot be controlled. This change would allow more fine-grained control over what enchanting seed in the menu is - it could be rerolled every time PreparedItemEnchantEvent is called to implement pre-1.8 enchanting, or the seed could be associated with a group of players instead of an individual player.


kennytv (approved):

okx-code:
@kennytv I've rebased this PR so it should be ready

Machine-Maker (approved):

PR #11299, opened 2024-08-19
verify returned suggestion type

456dev:
draft: need to figure out where else in the api is/isnt typesafe for lists.

since generic types get erased at runtime, java doesn't actually know what is in the returned. 
here i convert it to a String array, 
which only accepts the one type, which throws the ArrayStoreException when failed. 

additionally, since this occurs in BukkitCommandNode, the plugin/commands class is not in the stacktrace, so i add the command class to the error log.

made to help diagnose what plugin was causing [this](https://discord.com/channels/289587909051416579/289587909051416579/1275042054480597015)

---
Download the paperclip jar for this pull request: [paper-11299.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1827170120.zip)

### DIFF
--- a/patches/server/1051-verify-returned-suggestion-type.patch
+++ b/patches/server/1051-verify-returned-suggestion-type.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Isaac - The456 <the456@the456gamer.dev>
+Date: Mon, 19 Aug 2024 12:14:38 +0100
+Subject: [PATCH] verify returned suggestion type
+
+since generic types get erased at runtime, java doesn't actually know what is in the returned. List
+convert it to a String array, which only accepts the one type, which throws the ArrayStoreException when failed. additionally, since this occurs in BukkitCommandNode, the plugin/commands class is not in the stacktrace.
+
+diff --git a/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitCommandNode.java b/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitCommandNode.java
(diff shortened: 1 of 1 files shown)


patches/server/1051-verify-returned-suggestion-type.patch
```diff
@@ -0,0 +1,25 @@
...
+@@ -116,9 +116,10 @@ public class BukkitCommandNode extends LiteralCommandNode<CommandSourceStack> {
+             Location pos = context.getSource().getLocation();
+             try {
+                 results = this.command.tabComplete(sender, this.literal, args, pos.clone());
+-            } catch (CommandException ex) {
++                results.toArray(new String[0]);
++            } catch (CommandException | ArrayStoreException ex) {
+                 sender.sendMessage(ChatColor.RED + "An internal error occurred while attempting to tab-complete this command");
+-                Bukkit.getServer().getLogger().log(Level.SEVERE, "Exception when " + sender.getName() + " attempted to tab complete " + builder.getRemaining(), ex);
++                Bukkit.getServer().getLogger().log(Level.SEVERE, "Exception when " + sender.getName() + " attempted to tab complete " + builder.getRemaining() + "from command "+this.command.getClass(), ex);
```
dawon:
```suggestion
+                Bukkit.getServer().getLogger().log(Level.SEVERE, "Exception when " + sender.getName() + " attempted to tab complete " + builder.getRemaining() + " from command "+this.command.getClass(), ex);
```

Machine-Maker:
I think this falls into a similar category as the other PR you opened relating to generic type checks at runtime.

456dev:
Yeah understandable. This was mainly so I could give extra debug logging build without sending a random jar. 
Closing

PR #11300, opened 2024-08-19
Update launchProjectile API

Lulu13022002:
From the dispenser:
Update CraftBlockProjectileSource#launchProjectile to use right power/uncertainty value from the dispenser config.
Fix wrong initial velocity for small fireball and some NPE for Enderpearl/SmallFireball.
Drop support for things that can't be dispensed: Enderpearl/LargeFireball/WitherSkull and add support for new entities: Fireworks and WindCharge.

From living entities:
Use the right power/uncertainty value for wind charge and make firework projectile being launched using the same logic as with a crossbow and not attached to the entity.


patches/server/0677-More-Projectile-API.patch
```diff
@@ -791,3 +791,122 @@ index e8a455eb5e17bcfcae3f03664f2b47773fbdf37e..08178a88ba7d0881a6c2843eef24a846
...
+         net.minecraft.world.entity.Entity launch = null;
++        net.minecraft.world.item.Item item = null; // Paper
+ 
+         if (Snowball.class.isAssignableFrom(projectile)) {
+             launch = new net.minecraft.world.entity.projectile.Snowball(world, iposition.x(), iposition.y(), iposition.z());
+         } else if (Egg.class.isAssignableFrom(projectile)) {
+             launch = new ThrownEgg(world, iposition.x(), iposition.y(), iposition.z());
+-        } else if (EnderPearl.class.isAssignableFrom(projectile)) {
++        } else if (false && EnderPearl.class.isAssignableFrom(projectile)) { // Paper - disable enderpearl, previously would NPE and this can't be approximate
+             launch = new ThrownEnderpearl(world, null);
```
Machine-Maker: WHy does this *have* to NPE? Can't we just use the other constructor that ends up not setting the owner?

Machine-Maker (changes requested):
I think we should just copy almost everything from ProjectileDispenseBehavior. There should be an if/elseif/else chain to select the itemtype and then all the methods on ProjectileItem should be used to create the entity as well as launch it.

Machine-Maker (approved):

PR #11303, opened 2024-08-19
Make most Collections in events Checked

456dev:
prevents plugins from adding  unexpected types of objects and causing hard-to-trace errors.

as written, its currently build on top of #11302

### DIFF
--- a/patches/api/0073-AsyncTabCompleteEvent.patch
+++ b/patches/api/0073-AsyncTabCompleteEvent.patch
@@ -11,6 +11,8 @@ and avoid going to main for tab completions.
 Especially useful if you need to query a database in order to obtain the results for tab
 completion, such as offline players.
 
+Also Enforces mutability of the existing TabCompleteEvent.
+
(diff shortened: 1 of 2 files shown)


kennytv:
While the idea definitely makes sense, this is a pretty sizable patch for something that's kind of a bottom of the barrell case that only happens if someone is grossly mishandling generics. So not sure we really want to maintain such a patch

456dev:
thats definitely understandable, and most stuff would still break with an error, just slightly later, as a time of use rather than a time of insert.
just makes it harder to attribute to a specific plugin, but as it should be rare anyway im happy to close this

Machine-Maker:
I think at this point, when this diff touches so many files and we are still using patches, such a change isn't super needed. I'd be for re-visiting this when we have the API as source-in-repo. The benefits of this don't outweigh the increased maintenance costs of messing with all these events.

456dev:
No worries

PR #11304, opened 2024-08-19
Fix InventoryOpenEvent cancellation

### LINKED ISSUE #9734: Cancelling inventory open doesn't fully cancel
### Expected behavior

The chest lid closes, the stat for opening opening a container is not incremented, nearby piglins are not angered

### Observed/Actual behavior

The lid stays open (for ender chests only), the stat is incremented and piglins are angered

### Steps/models to reproduce

Listen to `InventoryOpenEvent` and cancel if the inv type is `ENDER_CHEST`, `BARREL`, `CHEST`. Then try and open it in game.

### Plugin and Datapack List

none

### Paper version

581b683931db32618aa676160b731a276a3a6730

### Other
Affects:
* Barrel: piglins, stats
* Ender Chest: lid, piglins, stats
* Shulker Box: stats, piglins
* Chest: piglins, stats

notTamion:
fixes #9734 
stats are also incremented in most other container blocks but importing files and potentially having to deal with decompile errors doesn't seem worth it for statistics. this could be expanded upon after hardfork where we don't have to worry about that.


Machine-Maker:
I think its fine to fix all the cases here.

notTamion:
Alright. I will work on this when i get back from vacation on thursday

Machine-Maker (approved):

lynxplay (approved):

PR #11307, opened 2024-08-19
Fix structure check API

roro1506HD:
Fixes structure check API to look for structure in the Structures registry, instead of trying to look for structure types in the Structures registry (which throws an exception when looking for villages and other structures that doesn't have the same structure id and structure type id)

### DIFF
--- a/patches/server/0913-Add-Structure-check-API.patch
+++ b/patches/server/0913-Add-Structure-check-API.patch
@@ -5,24 +5,19 @@ Subject: [PATCH] Add Structure check API
 
 
 diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
-index 612dc787880e58e8325a658a63f9fe7536d0860c..cb6df90fd65f0286f4ff74e6d1e203293e6e0429 100644
+index 612dc787880e58e8325a658a63f9fe7536d0860c..c664e8ca115c0a0f64e6b54478a0a2e88e8a58b7 100644
 --- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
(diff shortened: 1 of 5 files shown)


patches/server/0913-Add-Structure-check-API.patch
```diff
@@ -17,7 +17,7 @@ index 612dc787880e58e8325a658a63f9fe7536d0860c..ebb119ab9f5a8ae580e54cb3c102cd86
 +    public boolean hasStructureAt(final io.papermc.paper.math.Position position, final Structure structure) {
 +        net.minecraft.world.level.levelgen.structure.Structure vanillaStructure = this.world.registryAccess()
 +            .registryOrThrow(net.minecraft.core.registries.Registries.STRUCTURE)
-+            .getHolder(CraftNamespacedKey.toMinecraft(structure.getStructureType().getKey()))
++            .getHolder(CraftNamespacedKey.toMinecraft(structure.getKey()))
 +            .orElseThrow()
 +            .value();
```
Machine-Maker: This should really just be replaced with `CraftStructure.bukkitToMinecraft()` now I think. Upstream added a bunch of utility methods in the "Craft" classes to convert between API and nms types.

Machine-Maker (changes requested):

patches/server/0913-Add-Structure-check-API.patch
roro1506HD: Should I put it back to a one liner?

Machine-Maker: I mean it can be. Or just assign the `vanillaStructure` local var to `CraftStructure.bukkitToMinecraft(structure);`.

Machine-Maker (approved):

PR #11308, opened 2024-08-20
Add a method to reset Lodestone compasses back to normal ones

### LINKED ISSUE #11051: Unable turn a lodestone compass back into compass
### Expected behavior

When removing the lodestone properties of a lodestone compass it should become a compass.

### Observed/Actual behavior

The compass remains a lodestone compass when trying to set it back to a compass.

### Steps/models to reproduce

```java
ItemStack compass = new ItemStack(Material.COMPASS, 1);
CompassMeta compassMeta = (CompassMeta)compass.getItemMeta();
compassMeta.setLodestoneTracked(false);
compass.setItemMeta(compassMeta);
player.getInventory().addItem(compass);

ItemStack compassFromInv = getPlayerCompass(player);
CompassMeta compassFromInvMeta = (CompassMeta)compassFromInv.getItemMeta();
compassFromInvMeta.setLodestone(null);
compassFromInvMeta.setLodestoneTracked(true);
compassFromInv.setItemMeta(compassFromInvMeta);

public ItemStack getPlayerCompass(Player player) {
    PlayerInventory inv = player.getInventory();

    for (int slot = 0; slot < inv.getSize(); slot++) {
        ItemStack stack = inv.getItem(slot);
        if (stack != null && stack.getType() == Material.COMPASS) {
            return stack;
        }
    }

    return null;
}
```

### Plugin and Datapack List

```
> plugins
[12:46:52 INFO]: Server Plugins (1):
[12:46:52 INFO]: Bukkit Plugins:
[12:46:52 INFO]:  - PlayerTracker
```

### Paper version

```
> version
[12:44:14 INFO]: Checking version, please wait...
[12:44:15 INFO]: This server is running Paper version 1.21-49-master@b0ef3ca (2024-07-09T15:22:17Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
```

### Other

The issue is not present in the latest Spigot as I figured the hard way only after having opened the issue.
https://hub.spigotmc.org/jira/browse/SPIGOT-7826

DerEchtePilz:
Fixes #11051


patches/server/1052-Add-method-to-reset-Lodestone-compasses-back-to-norm.patch
```diff
@@ -0,0 +1,55 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
+index 6362df65424e53098701b8d54c74b5905648b78a..95f41c1315d6cca8f35eb915fa4d6a9f510a4633 100644
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
+@@ -36,6 +36,7 @@ public class CraftMetaCompass extends CraftMetaItem implements CompassMeta {
+     private int lodestoneY;
+     private int lodestoneZ;
+     private Boolean tracked = null; // Paper - tri-state
++    private boolean shouldResetLodestone = true; // Paper - Add method to reset Lodestone compasses back to normal ones
```
Machine-Maker: How is adding this field different from just setting the tracked Boolean to null. That's why we made the `tracked` field a "tri-state" with a nullable boolean.

DerEchtePilz: I actually did not think about that, but it definitely makes sense.
I did change that now.

patches/server/1052-Add-method-to-reset-Lodestone-compasses-back-to-norm.patch
```diff
@@ -0,0 +1,45 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
+@@ -94,6 +94,12 @@ public class CraftMetaCompass extends CraftMetaItem implements CompassMeta {
+         if (target.isPresent() || this.hasLodestoneTracked()) {
+             tag.put(CraftMetaCompass.LODESTONE_TARGET, new LodestoneTracker(target, this.tracked));
+         }
++
++        // Paper start  - Add a method to reset Lodestone compasses back to normal ones
++        if (tracked == null) {
++            tag.builder.remove(LODESTONE_TARGET.TYPE);
++        }
```
Lulu13022002: This shouldn't be needed, the patch is recreated from scratch each time the meta is applied (non empty ones).

Lulu13022002 (commented):
While testing this, I noticed a NPE when having an undefined tracked value with a target location set:
https://pastes.dev/iWWixkNCro
```Java
ItemStack compass = ItemStack.of(Material.COMPASS);
compass.editMeta(CompassMeta.class, meta -> {
    meta.setLodestone(new Location(world, 0, 5, 0));
    meta.resetLodestoneTracked();
});
```

Machine-Maker:
Yes, I too saw that possible NPE when reviewing the code... I think, instead, this should be a reset method that resets both things. Technically it's actually incorrect for "tracked" to be a TriState, as the value is a primitive boolean. But there does need to be a 3rd state for "no lodestone component at all". I think a reset location and a clear lodestone (that clears both the tracked and position) is the correct solution.

DerEchtePilz:
> I think a reset location and a clear lodestone (that clears both the tracked and position) is the correct solution.

I mean, according the the Javadocs, the location can be cleared by passing `null` into the `CompassMeta#setLodestone(Location)` method.
As for the tri state, the `tracked` boolean is responsible for deciding whether the location is stored in the component. If we say we convert the `tracked` Boolean into a primitive, in the method I am adding here we could simply set the `tracked` to false and the location to `null` which is what you're saying I guess.

Machine-Maker:
Yes, that's exactly what I'm saying. tracked probably still has to be a **B**oolean, because the compassmeta could have no lodestone at all, so both fields would be null, but essentially yeah.

patches/api/0486-Add-method-to-reset-Lodestone-compasses-back-to-norm.patch
```diff
@@ -0,0 +1,26 @@
...
+      */
+     void setLodestoneTracked(boolean tracked);
+ 
++    // Paper start - Add method to reset Lodestone compasses back to normal ones
++    /**
++     * If this compass is currently set to track a specific lodestone, this method
++     * will reset this lodestone compass back to a normal compass so that
++     * {@link #isLodestoneTracked()} returns {@code false}.
++     */
++    void resetLodestoneTracked();
```
Machine-Maker: I think the name is now a little wrong. What about `clearLodestone` or something like that? I don't think we need the "tracked" reference anymore.

patches/server/1052-Add-method-to-reset-Lodestone-compasses-back-to-norm.patch
```diff
@@ -0,0 +1,33 @@
...
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
+index 6362df65424e53098701b8d54c74b5905648b78a..6652d8647524c3049bc7dca4c08c62c595d37e01 100644
+--- a/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaCompass.java
+@@ -137,6 +137,7 @@ public class CraftMetaCompass extends CraftMetaItem implements CompassMeta {
+             this.lodestoneY = lodestone.getBlockY();
+             this.lodestoneZ = lodestone.getBlockZ();
+         }
++        this.tracked = false; // Paper - Add a method to reset Lodestone compasses back to normal ones
```
Machine-Maker: I think the default value is `true`.

Machine-Maker (changes requested):
Really, I think all the logic would get simpler if all the fields on CraftMetaItem were just replaced with a single field of LodestoneTracker and then that field was just changed or queried in all the method implementations.

Machine-Maker (approved):

PR #11309, opened 2024-08-20
Deprecate API relating to menu title changes

Machine-Maker:
In addition to just having super undefined functionality in lots of cases, I just realized this API is broken in another respect. Any call to it will open an inventory, even when interacting with an InventoryView that has not been opened to the player yet (such as in an InventoryOpenEvent). This is not well defined API and should never have been added.

Closes https://github.com/PaperMC/Paper/discussions/10575


Malfrador:
Already commented on Discord, but I am very much against deprecating this, especially for removal.

Yes, this API can cause issues in some cases with cursor items, and yes it does not have well defined behaviour. Frankly it sucks. But its also the only option we got for this feature. Just adding a note should be enough, no reason to deprecate it.

Being able to change an inventory title is a very important feature for making GUIs. And inventories are currently the only way of making clickable GUIs in this game.
This is especially an important feature when creating GUIs using resource pack fonts, which use the inventory title to display a background. Removing this method (which a deprecation for removal would imply happens at some point) would make creating GUIs needlessly complicated, error-prone and annoying as you would need to re-create the entire view just to change the title, which makes managing GUIs in a plugin a lot more complicated than being able to hold on the the same View instance. 

Of course a plugin would still have the ability to send the ClientboundOpenScreenPacket itself, however Paper really should not require using userdev/reflection for something Spigot simply offers in the API.

kennytv (approved):
After more internal discussion we decided it's alright as long as it's not deprecated for removal - show that the method has a pretty bad implementation, but acknowledge it's useful/doesn't have a replacement

dawon:
> After more internal discussion we decided it's alright as long as it's not deprecated for removal - show that the method has a pretty bad implementation, but acknowledge it's useful/doesn't have a replacement

A good "replacement" IMO is just to open a new inventory (with the new name) without closing the previous one. I guess it has the same or similar caveats as the methods, but even if those will be removed, this might be used.

lynxplay (approved):

PR #11312, opened 2024-08-20
fix despawn ranges by defaulting to an ellipsoid shape

Machine-Maker:
Adds a configuration setting to choose the despawn range's shape, defaulting to an ellipsoid, but allowing for a cylinder. Testing shows it works, and the different shapes have the correct results. A double check on the math would be super appreciated.


patches/server/1052-fix-despawn-ranges.patch
```diff
@@ -0,0 +1,167 @@
...
++        final double dySqr = Math.pow(dy, 2);
++        if (!this.cached) {
++            this.cacheValues(defaultDistance);
++        }
++        if (shape == Shape.ELLIPSOID) {
++            if (!this.horizontalLimit.isDefined() && !this.verticalLimit.isDefined()) {
++                return distanceSquared > defaultDistance * defaultDistance;
++            }
++            final double dxSqr = Math.pow(Math.abs(nearestPlayer.getX() - origin.x()), 2);
++            final double dzSqr = Math.pow(Math.abs(nearestPlayer.getZ() - origin.z()), 2);
```
19MisterX98: Taking the absolute value before squaring is redundant

patches/server/1052-fix-despawn-ranges.patch
```diff
@@ -0,0 +1,167 @@
...
+-    @Required boolean wasDefinedViaLongSyntax
+-) {
++/*
++(x/a)^2 + (y/b)^2 + (z/c)^2 < 1
++a == c
++ac = horizontal limit
++b = vertical limit
++x^2/ac^2 + y^2/b^2 + z^2/ac^2 < 1
++(x^2 + z^2)/ac^2 + y^2/b^2 < 1
++x^2 + z^2 + (y^2 * (ac^2/b^2)) < ac^2
```
19MisterX98: Looks good to me

lynxplay (approved):

PR #11313, opened 2024-08-21
Fix InventoryAction wrong for Bundles

notTamion:
- Follow-up pr to #10703. Fixes an issue i discovered while discussing the pr. Since craftbukkit manually calculates the InventoryAction for the click events it doesn't take bundles into account. I am not sure whether PICKUP_ONE_INTO_BUNDLE and PLACE_ONE_INTO_BUNDLE are needed since there isn't a way to explicitly get 1 item out of a bundle like with right clicking a normal stack (or whether the new enums are needed at all)

post hardfork #11902


Leguan16:
Is this worth updating given that bundles are still experimental?

notTamion:
Considering mojang is now also adding bundles to bedrock and have an experiment for further updates to it, i would definitely say that it is worth fixing this. As long as there aren't any changes in the way bundles are used in terms of how you pickup items and etc. i don't think this fix will pose an issue when updating mc versions.

Machine-Maker (changes requested):
Have you looked ahead to the 1.21.2 changes to the bundle? I haven't looked in depth at some of the functionality changes to see if any of these "actions" would no longer exist or be named something else. 

Either way, these need to be annotated with the `ApiStatus.Experimental` annotation so we can make any future changes until the bundle isn't experimental anymore.

notTamion:
there is a small change in the experiment where right and left click were exchanged which here could just be fixed by making a change to an if statement. With my comment above i meant bigger changes for which we would need to delete an enum or move big portions of code around. We can ofc also wait for what mojang actually decides to do with bundles in 1.21.2 and decide what we do from there

notTamion:
updated for bundle changes in 1.21.2

PR #11317, opened 2024-08-22
Fix setDamage being ran on compasses

NewwindServer:
In VanillaChestLoot, compasses accidentally have a setdamage loot function on them, but compasses don't take durability, resulting in a warning 
"Couldn't set damage of loot item 1 minecraft:compass"
This patch simply removes attempting to add damage to compass


Owen1212055:
Is there a vanilla bug report on this?

Machine-Maker (changes requested):
Changing it here doesn't do anything. You are just changing it for the data generator. You have to actually change the vanilla loot table.

Vanilla issue: https://bugs.mojang.com/browse/MC-266489

lynxplay (approved):

Machine-Maker (approved):

MiniDigger:
do we add the vanilla issue to the patch description and the paper comment?

PR #11318, opened 2024-08-22
Fix incorrect timestamp in region file header recalculation

### LINKED ISSUE #11311: Logic error in region file header recalculation
### Expected behavior

Timestamps in a region file's header are unix seconds, stored as ints:
```java
private static int getTimestamp() {
        return (int) (Util.getEpochMillis() / 1000L);
}
```

In the header recalculation patch, timestamps can be accidentally set to `(int)System.currentTimeMillis()` without dividing by 1000 first
https://github.com/PaperMC/Paper/blob/32711191cdad920b4818915b21699f9433800bf2/patches/server/1038-Attempt-to-recalculate-regionfile-header-if-it-is-co.patch#L414

Very ironic comment :>
`// write a valid timestamp for valid chunks, I do not want to find out whatever dumb program actually checks this`

### Observed/Actual behavior

N/A

### Steps/models to reproduce

N/A

### Plugin and Datapack List

N/A

### Paper version

Paper 1.21.1 32711191cdad920b4818915b21699f9433800bf2

### Other

_No response_

NewwindServer:
Timestamps in a region file's header are meant to be unix seconds, stored as ints.
In the header recalculation patch, timestamps can be accidentally set to (int)System.currentTimeMillis() without dividing by 1000 first,
causing an invalid timestamp to be saved. This patch fixes the issue by saving the correct timestamp (unix seconds, rounded to int)

Fixes #11311


kennytv (approved):
Needs to go into the correct patch

Machine-Maker (approved):

PR #11321, opened 2024-08-24
Fix allowSpiderWorldBorderClimbing world config

Lulu13022002:
The implementation has been dropped some version ago and now only toggle the climbing ability of spiders without checking the border collision. Additionally fixed a flaw in the previous implementation that disallowed spider outside the border to climb any block.

### DIFF
--- a/patches/server/0786-Add-config-option-for-spider-worldborder-climbing.patch
+++ b/patches/server/0786-Add-config-option-for-spider-worldborder-climbing.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add config option for spider worldborder climbing
 
 
 diff --git a/src/main/java/net/minecraft/world/entity/monster/Spider.java b/src/main/java/net/minecraft/world/entity/monster/Spider.java
-index c80019f0c9f814c5259b4d3ec2d8a85669dc728f..6006480d9f6d60bb7b5628eabe6740013066cde4 100644
+index 02171952fad7a17d93e81d95498b8244b807ffe2..a30fb47559eb74b7fe634678e63a85e7e2cad9a4 100644
 --- a/src/main/java/net/minecraft/world/entity/monster/Spider.java
 +++ b/src/main/java/net/minecraft/world/entity/monster/Spider.java
 @@ -82,7 +82,7 @@ public class Spider extends Monster {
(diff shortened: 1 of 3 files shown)


patches/server/1026-Add-config-option-for-spider-worldborder-climbing.patch
```diff
@@ -5,15 +5,15 @@ Subject: [PATCH] Add config option for spider worldborder climbing
...
 --- a/src/main/java/net/minecraft/world/entity/monster/Spider.java
 +++ b/src/main/java/net/minecraft/world/entity/monster/Spider.java
 @@ -82,7 +82,7 @@ public class Spider extends Monster {
      public void tick() {
          super.tick();
          if (!this.level().isClientSide) {
 -            this.setClimbing(this.horizontalCollision);
-+            this.setClimbing(this.horizontalCollision && (this.level().paperConfig().entities.behavior.allowSpiderWorldBorderClimbing)); // Paper - Add config option for spider worldborder climbing
++            this.setClimbing(this.horizontalCollision && (this.level().paperConfig().entities.behavior.allowSpiderWorldBorderClimbing || !(ca.spottedleaf.moonrise.patches.collisions.CollisionUtil.isCollidingWithBorder(this.level().getWorldBorder(), this.getBoundingBox().inflate(ca.spottedleaf.moonrise.patches.collisions.CollisionUtil.COLLISION_EPSILON)) && this.level().getWorldBorder().isInsideCloseToBorder(this, this.getBoundingBox())))); // Paper - Add config option for spider worldborder climbing (Inflate by +EPSILON as collision will just barely place us outside border)
```
lynxplay: What do we check `this.level().getWorldBorder().isInsideCloseToBorder(this, this.getBoundingBox())` for here?
A collision is already checked by the prior statement. Preventing spiders from doing this from the outside?

Lulu13022002: This is a distance check relative to the edge of border.
Without that spider cannot climb any block outside the border when the config is disabled. I think the config should only apply to the edge. isCollidingWithBorder check if the mob is colliding with the border voxel shape so not only the edge. You can see similar things in CollisionUtil#getCollisionsForBlocksOrWorldBorder L1586.

Machine-Maker (approved):

lynxplay (approved):

lynxplay (approved):

PR #11332, opened 2024-08-26
Fix: synchronise sending chat to client with updating message signature cache

Gegy:
In the case where multiple messages from different players are being processed in parallel, there was a potential race condition where the messages would be sent to the client in a different order than the message signature cache was updated. However, the cache relies on the fact that the client and server get the exact same updates in the same order. This race condition would cause the caches to become corrupted, and any future message received by the client would fail to validate.

This also applies to the last seen state of the server, which becomes inconsistent in the same way as the message signature cache and would cause any messages sent to be rejected by the server too.

This can be easily reproduced by having two players very rapidly send chat messages to the server at the same time.


kennytv:
I'd assume this is from Spigot's chat offloading changes and not adventure (?), so this can go into a separate patch

kennytv (approved):
Thank you!

Issue #9404, opened 2023-06-27
Some neutral mobs do not retaliate with universalAnger enabled


Machine-Maker:
### Expected behavior

Neutral mobs, when attacked with the `universalAnger` gamerule set to true, should attack any nearby valid target.

### Observed/Actual behavior

Several neutral mobs do not attack any targets, effectively making them passive mobs.

This issue occurs with (tested from [list of neutral mobs here](https://minecraft.fandom.com/wiki/Category:Neutral_mobs)):
* Spider
* Cave Spider
* Panda
* Dolphin
* Goat *(maybe, it uses the new AI system and it doesn't attack very often)*
* Llama

### Steps/models to reproduce

1. `/gamerule universalAnger false` (the default)
2. Summon one of the entities listed above
3. Attack it while in survival mode and note its response (attack in lit up area for spiders)
4. `/gamerule universalAnger true`
5. Summon another entity
6. Attack again, and notice its lack of response

### Plugin and Datapack List

none

### Paper version

cf0f013f53ef6c51528dca6d8873b61dc9546b1e

### Other

There exists a similar [ticket](https://github.com/PaperMC/Paper/issues/5155) and [fix](https://github.com/PaperMC/Paper/pull/5156), as well as a [mojira issue](https://bugs.mojang.com/browse/MC-195278).

Machine-Maker:
I'm going to close this issue, because the fix for this is not clear. It's not clear what exactly the `universalAnger` gamerule is controlling or what is a neutral mob. Should spiders that are hit during the day be angry at all players, not just the one that hit them? They don't implement the NeutralMob interface. I don't think we can fix this without changing behavior in a way mojang might not.

PR #11334, opened 2024-08-27
Fix-MC-195278-universal_anger_gamerule

SoulxluoS:
Hi,
try fix this https://github.com/PaperMC/Paper/issues/9404

I understand that the changes may be questionable, but I tested it and it seems to work.


Machine-Maker:
Thanks for your first contribution, however when going over this an the linked issue, it's clear that we cannot really fix this without changing some other behavior (which then departs from vanilla functionality). You can see a fuller explanation of it in the message that closed the linked issue. Sorry for not realizing this earlier and closing that issue before you spent time trying to fix it.

PR #11336, opened 2024-08-28
 Don't save entities when a chunk's save is cancelled

NewwindServer:
From searching code across GitHub, the primary use case for ChunkUnloadEvent.setSaveChunk(false) seems to be preventing certain worlds, like minigame worlds, from being saved.

In most cases, if you're not saving a chunk, it also makes sense not to save the entities within that chunk.
Before version 1.17, this was the default behavior, as entities were saved within chunks.

However, in the current system, even when chunk saving is canceled, entities like shot arrows or unexploded TNT are still saved, which can cause leftover items in minigame worlds.


lynxplay (changes requested):
Mojang disconnected chunk and entity saving a while ago.
Hence why spigot introduced the EntitiesUnloadEvent.

The fact that they save in the same logic again is entire an implementation detail of the chunk rewrite. Having the chunk event be responsible for disabling the entity saving hence becomes rough if this impl ever changes.

I think it would be better to have such a setting on the respective EntitiesUnloadEvent.

PR #11339, opened 2024-08-29
 Folia Scheduler : Do not throw an error if the delay is set to 0, just like for entities.

Euphillya:
On the entity scheduler, we see that a check using the Math.max method exists, but not on the others. If you set it to 0, you just get an error, so I modified it to default to the next tick instead.

### DIFF
--- a/patches/server/0848-Folia-scheduler-and-owned-region-API.patch
+++ b/patches/server/0848-Folia-scheduler-and-owned-region-API.patch
@@ -268,10 +268,10 @@ index 0000000000000000000000000000000000000000..94056d61a304ee012ae1828a33412516
 +}
 diff --git a/src/main/java/io/papermc/paper/threadedregions/scheduler/FoliaAsyncScheduler.java b/src/main/java/io/papermc/paper/threadedregions/scheduler/FoliaAsyncScheduler.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..374abffb9f1ce1a308822aed13038e77fe9ca08b
+index 0000000000000000000000000000000000000000..66f65192c06e4549cdd1e31db72644f215344cd0
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/threadedregions/scheduler/FoliaAsyncScheduler.java
-@@ -0,0 +1,328 @@
+@@ -0,0 +1,319 @@
 +package io.papermc.paper.threadedregions.scheduler;
 +
 +import ca.spottedleaf.concurrentutil.util.Validate;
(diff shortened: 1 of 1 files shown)


patches/server/0848-Folia-scheduler-and-owned-region-API.patch
```diff
@@ -363,15 +363,12 @@ index 0000000000000000000000000000000000000000..374abffb9f1ce1a308822aed13038e77
...
-+        if (delay < 0L) {
-+            throw new IllegalArgumentException("Delay may not be < 0");
-+        }
 +
 +        if (!plugin.isEnabled()) {
 +            throw new IllegalPluginAccessException("Plugin attempted to register task while disabled");
 +        }
 +
-+        return this.scheduleTimerTask(plugin, task, delay, -1L, unit);
++        return this.scheduleTimerTask(plugin, task, Math.max(1L, delay), -1L, unit);
```
Owen1212055, line 371: This changes behavior now? Since I cant set delay to 0 anymore, it'll be set to 1.

Machine-Maker:
If setting the delay to 0 leads to an exception somewhere (idk if this happens for sure, just assuming), then either that exception should be fixed, or if that isn't feasible, 0 just shouldn't be allowed as an argument value.

patches/server/0848-Folia-scheduler-and-owned-region-API.patch
Euphillya: This requires that if we set it to 0, it must be executed on the next tick, because it is obligatory.

Warriorrrr:
The exception is from inside the method itself, clearly stating that the delay/period must be greater than 0, which makes sense to me. The only way that 0 as an input is feasible is to Math.max it like done here

Euphillya:
> If setting the delay to 0 leads to an exception somewhere (idk if this happens for sure, just assuming), then either that exception should be fixed, or if that isn't feasible, 0 just shouldn't be allowed as an argument value.

It is not allowed precisely. But on the entity scheduler, we can set it to 0 BUT it is corrected by a Math.max so that the lowest value is 1 (see here: https://github.com/PaperMC/Paper/blob/25621248d3f30f35c3458b4b13a2474ffd0bee0e/patches/server/0848-Folia-scheduler-and-owned-region-API.patch#L178 )

Machine-Maker:
After further consideration, we thinking throwing is the more correct behavior. It's odd to clamp values to 1 instead of specifying what values are valid in the first place and disallowing other values. A PR adding a thrown exception for the EntityScheduler#execute method which I think is the only method that doesn't throw out of all the scheduler methods.

PR #11353, opened 2024-09-02
Fix NPE while trying to respawn an already disconnected player

### LINKED ISSUE #11352: Calling "Player#spigot#respawn" in EntityDeathEvent after Logging Out Throws NPE
### Expected behavior

I expected using the `player.spigot().respawn()` method to not throw a NPE regardless, and respawn the player regardless of their state. It seems that the `DimensionTransition` stuff is relatively new and probably not handled correctly somehow.

### Observed/Actual behavior

Instead, the server threw a NPE (https://paste.helpch.at/iritoqudur.css), complaining the following:
`java.lang.NullPointerException: Cannot invoke "net.minecraft.world.level.portal.DimensionTransition.pos()" because "dimensiontransition" is null`

### Steps/models to reproduce

1) Register an event listener for `PlayerQuitEvent`. In that event, kill the player by setting their health to 0f or `player.setHealth(0f)`
2) Register an event listener for `EntityDeathEvent`. In that event, check that the `Entity` is a `Player` and then use `player.spigot().respawn()` to attempt to respawn the player.
3) In-game, log out and check the error in console.

### Plugin and Datapack List

WorldEdit, Citizens, LuckPerms, [MurderRun](https://github.com/PulseBeat02/MurderRun) (my plugin)

### Paper version

```
[11:06:27 INFO]: Checking version, please wait...
[11:06:27 INFO]: This server is running Paper version 1.21.1-57-master@b483da4 (2024-09-01T18:09:05Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
```
(I am using the runServer task)

### Other

_No response_

Doc94:
Closes https://github.com/PaperMC/Paper/issues/11352 by adding a new check for not run the method if the connection is in disconnect state, the method depends in many other methods related to connections then can cause NPE in many places of the respawn logic call the method for things like respawn a player in the disconnection state.

Example related to the issue reported for replicate and test.
```java
@EventHandler
public void onQuit(PlayerQuitEvent event) {
    this.getLogger().info("Try to kill " + event.getPlayer().getName());
    event.getPlayer().setHealth(0);
}

@EventHandler
public void onPlayerDeath(EntityDeathEvent event) {
    if (event.getEntity() instanceof Player player) {
        this.getLogger().info("Try to Respawn " + player.getName());
        player.spigot().respawn();
    }
}
```

### DIFF
--- a/patches/server/0282-Add-PlayerPostRespawnEvent.patch
+++ b/patches/server/0282-Add-PlayerPostRespawnEvent.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Add PlayerPostRespawnEvent
 
 
 diff --git a/src/main/java/net/minecraft/server/players/PlayerList.java b/src/main/java/net/minecraft/server/players/PlayerList.java
-index 79203d0e5cdb86d9e2fb22cdaeb8cf3a93e43dcc..f9952da3e7463f9ef1d2444c5ac4fd2b22bcb02e 100644
+index 79203d0e5cdb86d9e2fb22cdaeb8cf3a93e43dcc..b85111d34388bc0e49891b10e28d19f7c643e5c0 100644
 --- a/src/main/java/net/minecraft/server/players/PlayerList.java
 +++ b/src/main/java/net/minecraft/server/players/PlayerList.java
 @@ -738,6 +738,10 @@ public abstract class PlayerList {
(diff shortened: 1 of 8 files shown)


electronicboy (approved):

Machine-Maker:
I feel like the "isOnline" method should already be checking this. What about changing the implementation of `isOnline` to also check this?

Doc94:
> I feel like the "isOnline" method should already be checking this. What about changing the implementation of `isOnline` to also check this?

i mean isOnline docs already mention the player can have instances where is online but is disconnected and recommed check other method.. make this change require change the behaviour of that method and can cause "side" effects...

and yeah exists isDisconnected but that check another place and not the method used in respawn then for this case i think is better just check the same thing...

Lulu13022002:
This only fix one specific NPE and not the other ones that might happens on the same var.

Doc94:
> This only fix one specific NPE and not the other ones that might happens on the same var.

You mean for the connection null? Or i missing another?

lynxplay (approved):

lynxplay:
@Lulu13022002 Moved the diff around, would like your review, I agree that the spigot logic here is rather terrible, but I don't think such a massive diff is nice to maintain for such a scuffed impl.

Machine-Maker (approved):

Lulu13022002 (approved):
LGTM, i have renamed the pr now that the original pr has been rewritten completely.
I think the old check can be removed instead of the diff on change. Otherwise there's a typo too.

PR #11358, opened 2024-09-04
Fix `AttributeModifier#getUniqueId`

qixils:
Returning null is a subjective fix (throwing an exception in a getter is odd but I can revert if desired), but the key changes are an objective fix to a completely broken method (`minecraft:305cec29-c12a-49ba-aebc-38b2b61dcd40` is not a valid UUID!)


lynxplay:
Upstream went another route in a recent [commit](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/cac68bfb9491dede699391865416c93f477fc2d0#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Fattribute%2FAttributeModifier.java) , we just need to pull that.

PR #11359, opened 2024-09-05
Fix Monster natural spawning problem due to light

### LINKED ISSUE #10265: Max Light Level monster spawn
### Expected behavior

Mobs should spawn if light level is below or equals to configured light level.

### Observed/Actual behavior

Entities only spawn at night with a vanilla (0) light level 

### Steps/models to reproduce

Time set day
Set server max-light setting to 15 or greater

### Plugin and Datapack List

No plugin

### Paper version

Latest

### Other

Here was the pull request that created that feature.
Regarding our recent discord discussion, it's appear to be bugged?

https://github.com/PaperMC/Paper/pull/7129

DavidTs93:
This fixes the natural spawning problem of monsters due to light, which causes any custom natural spawning light levels (from a datapack altering the dimension type or by changing `paper-world-defaults#monster-spawn-max-light-level`) to be completely ignored.

The natural spawning code (in `NaturalSpawner`) checks the mob's custom spawn rules (`Mob#checkSpawnRules`), which, for sentient creatures always checks if `PathfinderMob#getWalkTargetValue` is 0 or positive. This logic, specifically for monsters, is flawed (except for: Giant, Guardian, Pillager, Silverfish, and Warden), because the value of this function for monsters depends on the light level - in the overworld this means that light levels 7 or below are ok and above 7 are not (which is the default behavior).

This PR changes `checkSpawnRules` (`Monster#checkSpawnRules`) to ignore natural spawning (including chunk population) because all monsters already have a custom check for spawning rules in `SpawnPlacements`. Also fixes #10265.

### DIFF
--- a/patches/server/1056-Fix-Monster-natural-spawning-problem-due-to-light.patch
+++ b/patches/server/1056-Fix-Monster-natural-spawning-problem-due-to-light.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DavidTs93 <david.ts93@gmail.com>
+Date: Thu, 5 Sep 2024 12:26:18 +0300
+Subject: [PATCH] Fix Monster natural spawning problem due to light
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/monster/Monster.java b/src/main/java/net/minecraft/world/entity/monster/Monster.java
+index e7bfce0534c7ef3a1480a1082ae8514caf78778b..15c65d11be5430e45e8da7e3e4c0da759a949979 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/Monster.java
++++ b/src/main/java/net/minecraft/world/entity/monster/Monster.java
+@@ -85,6 +85,14 @@ public abstract class Monster extends PathfinderMob implements Enemy {
+     public float getWalkTargetValue(BlockPos pos, LevelReader world) {
+         return -world.getPathfindingCostFromLightLevels(pos);
+     }
++    
++    // Paper start - Natural spawning conditions already checked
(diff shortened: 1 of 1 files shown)


lynxplay:
While this PR fixes the issue, I think it is best for us to wait with merging this until we have some mojang input on the linked ticket.
Messing with spawn logic like this could break contraptions and until we know *why* mojang is even checking this in the first place when they are already checking brighness values in SpawnPlacements, this may cause more harm to normal players than good.

DavidTs93:
> While this PR fixes the issue, I think it is best for us to wait with merging this until we have some mojang input on the linked ticket.
> Messing with spawn logic like this could break contraptions and until we know *why* mojang is even checking this in the first place when they are already checking brighness values in SpawnPlacements, this may cause more harm to normal players than good.

I opened that ticket today, so doubt it'll be soon...
As I see it, they did the check to add some extra checks for each entity type, which I believe is just a leftover (why not just add that check to the default checker in `SpawnPlacements`?...).
Anyway, my solution is just one of the many possible solutions - maybe Mojang will say it's an intended behavior (even though that makes no sense - why add custom light levels for monster spawning if you then ignore it?...), or will fix it in another way.

I, for one, will just compile this patch into my own server.

lynxplay:
Yea, I agree that it *seems* to be a leftover.
But on the other hand, we *really* have 0 idea why that entire method even exists *and* diff around there was moved back in 22w12a, so not that long ago.

We'll try to get some info on the ticket, your patch certainly works yea, and for *most* users this change is probably 100% fine, but messing with spawning like this without a way to disable the early out on the Mob instance spawn check might break contraptions we would not want to break.

Thank you anyway for the PR, we'll poke around with the linked mojira :+1:

PR #11365, opened 2024-09-06
Add Decorated Pot Cracked API

Fabrimat:
Added `DecoratedPot#isCracked()` and `DecoratedPot#setCracked(boolean)`, not sure why they weren't there already.


Machine-Maker (approved):
I feel like we have a patch that touches some missing block data API, this can be merged into that patch by the author or whoever merges this PR.

Fabrimat:
I can merge it, but I'm not sure which patch you are talking about

Issue #11364, opened 2024-09-06
Lidded containers closing issue


Jakush:
### Expected behavior

I should be able to close any container implementing Lidded interface using method `close()`.

### Observed/Actual behavior

Every lidded container remained opened when there were viewers (even though it should be closed because the method for closing was called).

### Steps/models to reproduce

Call `Lidded#close();` on PlayerInteractEvent

### Plugin and Datapack List

Mine and luckperms.

### Paper version

This server is running Paper version 1.21-130-master@b1b5d4c (2024-08-10T10:02:42Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
You are running the latest version
Previous version: 1.21-129-4d7cef3 (MC: 1.21)

Machine-Maker:
I'm not sure this has ever worked in the way you are imagining... It looks like every 5 ticks, the game re-checks the open status and will re-send the block event for opening/closing to clients in range. Additionally, the `close` method will only have an effect if the open method was called first. You can see this if you just call open right before close, you can see the chest *start* to close, but then open back again (cause 5 ticks have passed). 

I'm not sure what the correct fix is here. It seems this API is fundamentally broken. We *could* add something like a force-closed boolean that prevents any opening from being sent to any clients. Plugins could toggle that on/off (or a force open). But tieing that into the existing Lidded API doesn't seem feasible.

456dev:
I've also had a look at this, and I think I've got a patch exposing the ability to force open/force close  containers lids in the nms container count manager (so chests, barrels, and enderchests, but not shulker boxes yet).

I also got a new feature working, a flag to automatically remove the force flag (once the first real viewer opens it for for forced open, or once the last real viewer has closed it for forced closed)

The enum based API sounds sensible though, since all the states are mutually exclusive

PR #11370, opened 2024-09-07
Fixed issues with lidded not closing patch

Jakush:
Hey, currently on latest version of paper, I'm having issues with actually using interface Lidded#close to actually close the container (barrel, shulkerbox or etc...), so I tried fixing it and well it works, I still have to fix some things there as it does not look open but still sends sound (which it shouldn't if it already closed), but I also wanted to hear a feedback before continuing.

Issue (maybe more detailed why): https://github.com/PaperMC/Paper/issues/11364

(sorry for previous invalid PR)


lynxplay (changes requested):
Unless I am reading this wrong, the PlayerInteractEvent is called before the menu (and lidded state) is ever opened/changed.

Even with this (hardcoded as fuck) patch, you'd be too early to call close on the container anyway?

Machine-Maker:
I explained over in the other thread why I think the Lidded open/close methods have a more fundamental problem with how lids work in the game.

PR #11371, opened 2024-09-07
Deprecate for removal all OldEnum-related methods

Machine-Maker:
These methods will be removed from the API and only available via asm-utils bytecode rewriting for older plugins.


yannicklamprecht (approved):
LGTM

lynxplay (approved):

electronicboy (approved):

PR #11379, opened 2024-09-09
Add a New and Improved Lidded API

456dev:
Featuring
- 5 LidModes, 3 of which were not possible before
- The ability to get if the lid actually should be open, if a player is in the container
- The ability to get if the is open, from api or the player itself

This also deprecates Bukkits api because of the confusing and outright incorrect javadocs (so those now accurately represent the behaviour)

Testable with the included testplugin commands.

Im not 100% sure if the PaperLidded interface is the best way of doing this, but it is working
Also, let me know if the naming needs to change.

```[tasklist]
### Tasks
- [ ] see if anything important is skipped in the check method scheduled every 5 ticks on chests 
```

### DIFF
--- a/patches/api/0490-New-and-Improved-Lidded-API.patch
+++ b/patches/api/0490-New-and-Improved-Lidded-API.patch
@@ -77,15 +77,16 @@ index 0000000000000000000000000000000000000000..ea775e7cd28b68e520d1381eee8c732f
 +}
 diff --git a/src/main/java/io/papermc/paper/block/Lidded.java b/src/main/java/io/papermc/paper/block/Lidded.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..b432f070e1689b0a0c65ca8015760c8b24c4769f
+index 0000000000000000000000000000000000000000..0f73d78a95cde2326aed4885d01282d177c2f245
 --- /dev/null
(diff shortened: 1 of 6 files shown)


Machine-Maker:
I'm wondering if an event is better suited for more control over what happens. I feel like some event in the correct spot would enable even more than just these 5 options. I'm not sure how that would look just now, I just had this idea. Maybe some event for when the open count changes? You could then query if it was open or not before, and decide whether to allow it to open, or to close or something.

456dev:
just an event wouldn't allow forcing the lid open I believe. 
Additionally, once the event completes, I can't see an easy way to undo whatever changes that were made.
for example, this could be used to flip the lid up and down, no matter if a player is viewing it or not.

dawon:
I could imagine an event here (PlayerContainerOpenEvent - maybe someone would come up with a better name ideally not confusable with InventoryOpenEvent) that would allow to prevent opening the lid for a specific player, container and/or situation. I would add that along with the api you already implemented. I think that would handle most possible use cases...

456dev:
@dawon @Machine-Maker I've now added PlayerLiddedOpenEvent, which triggers on open, and allows players to be removed from consideration at that stage. I've done it as a separate patch, as its not fully linked to the new Lidded api.

(its easier to merge those patches than re-separate them).

PR #11385, opened 2024-09-09
Cache Bukkit Command when wrapping CommandNodes

### LINKED ISSUE #11378: Brigadier commands registered through `Commands#getDispatcher` require incorrect permission when executed by `Bukkit#dispatchCommand`
### Expected behavior

When I execute my `/test` command as a player with no permissions, it runs successfully.

### Observed/Actual behavior

When I execute my `/test` command using `Bukkit#dispatchCommand`, the command fails with the message `I'm sorry, but you do not have permission to ...`

image

### Steps/models to reproduce

I created this plugin that registers two commands:

```java
public final class BukkitTestPlugin extends JavaPlugin {
    @Override
    public void onEnable() {
        getLifecycleManager().registerEventHandler(LifecycleEvents.COMMANDS, event -> {
            Commands commands = event.registrar();
            CommandDispatcher<CommandSourceStack> dispatcher = commands.getDispatcher();

            dispatcher.register(Commands.literal("test")
                .executes(context -> {
                    context.getSource().getSender().sendMessage("You ran the test command");

                    return 1;
                })
            );

            dispatcher.register(Commands.literal("run")
                .then(Commands.argument("command", StringArgumentType.greedyString())
                    .executes(context -> {
                        String command = context.getArgument("command", String.class);

                        Bukkit.dispatchCommand(context.getSource().getSender(), command);

                        return 1;
                    })
                )
            );
        });
    }
}
```

The `/test` command can be executed without any permissions. The `/run` command executes the given command string using `Bukkit#dispatchCommand`.

### Plugin and Datapack List

```
> plugins
[13:44:11 INFO]: Server Plugins (1):
[13:44:11 INFO]: Bukkit Plugins:
[13:44:11 INFO]:  - BukkitTestPlugin
> datapack list
[13:44:20 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[13:44:20 INFO]: There are no more data packs available
```

### Paper version

```
> version
[13:44:39 INFO]: Checking version, please wait...
[13:44:40 INFO]: This server is running Paper version 1.21.1-69-master@925c3b9 (2024-09-07T20:43:21Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
```

### Other

Server log: https://paste.gg/p/anonymous/19898849d44b4a5e884d147aeb2882ad

I'm looking into this problem to try to resolve https://github.com/JorelAli/CommandAPI/issues/583. My analysis of the Paper code that causes that issue and the one here can be found in [this comment](https://github.com/JorelAli/CommandAPI/issues/583#issuecomment-2272079761) on that issue. In summary, when `Bukkit#dispatchCommand` is run, the `BukkitBrigForwardingMap` added by #8235 converts the `CommandNode`s in the Brigadier dispatcher into Bukkit `Command`s. For vanilla commands, CommandAPI commands, and my `/test` and `/run` commands mentioned above, this constructor in `VanillaCommandWrapper` is used:

```java
public class VanillaCommandWrapper extends BukkitCommand {
    public VanillaCommandWrapper(Commands dispatcher, CommandNode<CommandSourceStack> vanillaCommand) {
        super(vanillaCommand.getName(), "A Mojang provided command.", vanillaCommand.getUsageText(), Collections.EMPTY_LIST);
        this.vanillaCommand = vanillaCommand;
        // Incorrectly sets the permission to, for example, `minecraft.commands.test`
        this.setPermission(VanillaCommandWrapper.getPermission(vanillaCommand));
    }
}
```

This constructor calls `setPermission`. In the case of my `/test` command, this means the permission `minecraft.commands.test` is required to run it through `Bukkit#dispatchCommand`, rather than being `null`, or no permission required.

On Spigot, the CommandAPI [fixes this](https://github.com/JorelAli/CommandAPI/blob/aacd95712386f2342e051a0331306298645eec56/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/SpigotCommandRegistration.java#L154-L184) by looping through the `CommandMap` and calling `setPermission` again to give CommandAPI commands the correct permission String. This works because Spigot only creates the `VanillaCommandWrappers` once, just before the server finishes start up. The same does not work on Paper, because a new `VanillaCommandWrapper` is created each time a vanilla node is retrieved from the `BukkitBrigForwardingMap`, so calls to `setPermission` do not persist to the next time the `Command` is retrieved and used.

willkroboth:
Resolves https://github.com/PaperMC/Paper/issues/11378 by "restoring" the Spigot behavior where VanillaCommandNodes are only created once. This allows command frameworks that insert CommandNodes directly into the Brigadier dispatcher to change the permission String of the VanillaCommandNodes created for their commands, rather than it always being the default `"minecraft.commands.<name>"`.

For example, the test plugin described in https://github.com/PaperMC/Paper/issues/11378 can do something like this now:

```java
public final class BukkitTestPlugin extends JavaPlugin {
    @Override
    public void onEnable() {
        getLifecycleManager().registerEventHandler(LifecycleEvents.COMMANDS, event -> {
            Commands commands = event.registrar();
            CommandDispatcher<CommandSourceStack> dispatcher = commands.getDispatcher();

            dispatcher.register(Commands.literal("test")...);

            dispatcher.register(Commands.literal("run")...);

            CommandMap commandMap = Bukkit.getCommandMap();
            commandMap.getCommand("test").setPermission(null);
            commandMap.getCommand("run").setPermission(null);
        });
    }
}
```

When the `/test` command is ran using `Bukkit#dispatchCommand`, Bukkit will now see that its permission is `null` and allow a Player without any permission to run it, which is the case when the same Player runs the command directly.

Without this PR, BukkitBrigForwardingMap creates a new VanillaCommandWrapper each time a CommandNode is requested via the Bukkit CommandMap. This meant that calls to `Command#setPermission` did not persist between retrievals from the map.

I can squash this into the `Brigadier-based-command-API` patch file if desired, but I figured it would be easier initially to review the changes when they are in a separate patch.

### DIFF
--- a/patches/server/1060-Cache-Bukkit-Command-when-wrapping-CommandNodes.patch
+++ b/patches/server/1060-Cache-Bukkit-Command-when-wrapping-CommandNodes.patch
@@ -1,0 +1,51 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: willkroboth <46540330+willkroboth@users.noreply.github.com>
+Date: Mon, 9 Sep 2024 19:11:58 -0400
+Subject: [PATCH] Cache Bukkit Command when wrapping CommandNodes
+
+
+diff --git a/src/main/java/com/mojang/brigadier/tree/CommandNode.java b/src/main/java/com/mojang/brigadier/tree/CommandNode.java
+index dc76fcf4c6cc6cd65ce117b1855c15ede60f30ab..5dd43e561c0dcd9346bd381ad1574535f9caef48 100644
+--- a/src/main/java/com/mojang/brigadier/tree/CommandNode.java
++++ b/src/main/java/com/mojang/brigadier/tree/CommandNode.java
+@@ -37,6 +37,7 @@ public abstract class CommandNode<S> implements Comparable<CommandNode<S>> {
+     public CommandNode<CommandSourceStack> clientNode; // Paper - Brigadier API
+     public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> unwrappedCached = null; // Paper - Brigadier Command API
+     public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> wrappedCached = null; // Paper - Brigadier Command API
++    public org.bukkit.command.Command wrappedBukkitCommandCached = null; // Paper - Brigadier Command API
+     // CraftBukkit start
+     public void removeCommand(String name) {
+         this.children.remove(name);
+diff --git a/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java b/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
+index 5eef7ae5197bd395fbd6800530ffe34d147651ff..4512985b2e5e37b3fda33621fe1b2681796e1b0b 100644
+--- a/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
++++ b/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
+@@ -83,11 +83,13 @@ public class BukkitBrigForwardingMap extends HashMap<String, Command> {
+             return null;
+         }
+ 
+-        if (node instanceof BukkitCommandNode bukkitCommandNode) {
+-            return bukkitCommandNode.getBukkitCommand();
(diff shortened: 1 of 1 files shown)


jpenilla:
It's not possible to properly represent a tree of requirements as a permission string. To solve the linked issue properly we need to reimplement dispatchCommand without using the legacy command map. We could also set the permission to null for plugin commands and override hasPermission(Silent) to improve the legacy representation.

PR #11397, opened 2024-09-13
PlayerInventory#setItem check for slot outside of inventory

### LINKED ISSUE #11391: `PlayerInventory#setItem` doesn't check for slot outside of inventory
### Expected behavior

`playerInventory.setItem(46, someItemStack);` should throw error as described in [the javadoc](https://jd.papermc.io/paper/1.21.1/org/bukkit/inventory/PlayerInventory.html#setItem(int,org.bukkit.inventory.ItemStack)) as the slot is outside of this inventory. But the server accept it.

### Observed/Actual behavior

Actually, the server side allow to do this, and just send it through the packet. But, we are disconnected for "Network protocol error" as the client cannot handle it. Also, it means the javadoc has different content that the real effect

### Steps/models to reproduce

- Login on server
- Use a plugin that will do:
```java
player.getInventory().setItem(46, new ItemStack(Material.STICK));
```

### Plugin and Datapack List

The test plugin

### Paper version

This server is running Paper version 1.21.1-76-master@e945cfe (2024-09-10T17:56:53Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)

### Other

_No response_

strainxx:
fixes #11391 
Now if setItem is < 0 or > 40 it will throw ArrayIndexOutOfBoundsException

### DIFF
--- a/patches/server/1037-Bundle-spark.patch
+++ b/patches/server/1037-Bundle-spark.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Bundle spark
 
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 4f6136ae3ac4890b21a5fb3f69f9c1474a0773d1..32832deabc39be99ae0d69fe81439b86789f09e4 100644
+index 4f6136ae3ac4890b21a5fb3f69f9c1474a0773d1..648281575eb8d45a5c06549eb3d0f517c086fe64 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -61,6 +61,10 @@ dependencies {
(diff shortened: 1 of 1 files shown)


patches/server/1060-check-for-slot-outside-of-inventory.patch
```diff
@@ -0,0 +1,22 @@
...
++++ b/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventory.java
+@@ -97,6 +97,11 @@ public class CraftInventory implements Inventory {
+ 
+     @Override
+     public void setItem(int index, ItemStack item) {
++        // Paper start - Validate setItem index
++        if (index < 0 || index > 40) {
++            throw new ArrayIndexOutOfBoundsException("Index must be between 0 and 40");
++        }
++        // Paper end - Validate setItem index
```
notTamion: This check should be done in CraftInventoryPlayer#setItem and not the super method. we could also modify the underlying nms method to set the item regardless of whether it is inbounds however that might confuse plugin developers as the player inventory is split into compartments and the index and size displayed in the exception will not match the slot input by the dev.

PR #11406, opened 2024-09-15
Add startingBrewTime

notTamion:
This fixes the issue that when setting a different Total brew time in BrewingStartEvent the arrow in the brewing stand menu will either start of already full to a certain amount or not filling up for some time depending on whether you set the brew time to under or over 400


lynxplay (commented):
Any reason this cannot live in the block entities data access?

But also I am not *too* keen on this patch in general given it "fixes" a visual bug by hiding it behind a single value (the brew time)
IMO it would make more sense to store a separate field in the brewing stand to configure the total visual brew time.
E.g. right now a plugin calling setBrewingTime to something above 400 will "lock" the stand to that "startBrewTime" value.

notTamion:
I didn't put it on the data access as we would be sending that extra data to the client (i believe there are some size checks in there so that might break something). 

Second point: Do you just want to make the value we calculate when sending the data to the client an extra variable? I don't see how that would be beneficial. Could you please elaborate on that?

lynxplay (commented):
Looking at this further, hijacking the slot setup is not gonna work for most implementations sadly, e.g. opening a custom inventory of that type currently just has hardcoded values for this. So I guess ctor overload it is or some fuckery with instanceOf in the ctor for ContainerData.

So to that degree I guess we need to force it being passed as an actual variable to a constructor/overload both.

Is `starting` really a good description for what this value does?
I mean, *from a technical* perspective, yea, but maybe something along the lines of "total" or "required"?

notTamion:
Not sure if this fixes it? Don't have time to test rn.

lynxplay:
Personal note keeping after talking it over with MM.
Create new BrewingSimpleContainerData subtype that extends SimpleContainerData but inits itself with value 2 set to 400.
Pass that in simple ctro and CraftContainer. Set both BrewingBLockEntityData and BrewingSimpleContainerData to count 3, keep the current "new containerData() { }` in setDataSlots to "translate" back to count 2 and do the conversion,

lynxplay:
PR Session log:

1. Changed to `setRecipeBrewingTicks` as our best attempt at naming whatever this concept is.
2. Experimental annotations for this naming. While it is our best attempt and will be merged, I am not comfortable in guaranteeing people won't have issues with it. If we receive large amounts of negative feedback and confusion, this allows us to revisit.
3. No longer bump `recipeBrewingTime` to set `brewingTime`. Spigot users calling this method and *trying* to get this weird behaviour would see a behaviour change. Additionally, if you want to configure this value (like you'd do in the event) we want you to also call the `setRecipeBrewingTicks` method if you end up overshooting the 400 ticks.

lynxplay (approved):

Machine-Maker (approved):

PR #11407, opened 2024-09-16
Prevent NPE when serializing unresolved profile

lynxplay:
In the recent upstream update, the paper player profile was updated to
correctly return null for both name and id if constructed as such. This
change however broke the serialisation logic, as it depended on the name
never being null.

The commit moves the checks over to the newly introduced emptyName/UUID
fields that track if the profile was constructed with a null name or
uuid to differentiate it against an empty string or the NIL_UUID.


Machine-Maker:
Does the deserialize method correctly set the fields somehow (via constructor or smth else)?

lynxplay:
Yea, 
```java
UUID uniqueId = ConfigSerializationUtil.getUuid(map, "uniqueId", true);
String name = ConfigSerializationUtil.getString(map, "name", true);

// This also validates the deserialized unique id and name (ensures that not both are null):
CraftPlayerProfile profile = new CraftPlayerProfile(uniqueId, name);
```

If not present (null), the ctor correctly handles those.

kennytv (approved):

Machine-Maker (approved):

PR #11412, opened 2024-09-17
Remove wall-time / unused skip tick protection

### LINKED ISSUE #11404: Items in lazy chunks continue to age until they despawn
### Expected behavior

Items in lazy chunks should not age

### Observed/Actual behavior

The item age stays consistent until the chunks are fully loaded, then increases by how many ticks passed since the item last ticked

### Steps/models to reproduce

- Go into a non-spawn chunk
- /forceload add ~ ~
- throw an item 2 chunks away
  - If you would like to track the item do /data modify entity @n[type=minecraft:item] Health set value 10s
  - And to see the age /data get entity @n[type=minecraft:item,nbt={Health:10s}] Age
- go in the nether to unload the chunks (or far away)
- wait (or tick warp) 2000 ticks
- return to see the age update to be 2000 higher than it was 1 tick ago (if over 6000 the item will despawn as expected)

### Plugin and Datapack List

No plugins

No datapacks

### Paper version

This server is running Paper version 1.21.1-77-master@4ff58c4 (2024-09-12T18:05:09Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)

### Other

_No response_

lynxplay:
Spigot still maintains some partial implementation of "tick skipping", a
practice in which the MinecraftServer.currentTick field is updated not
by an increment of one per actual tick, but instead set to
System.currentTimeMillis() / 50. This behaviour means that the tracked
tick may "skip" a tick value in case a previous tick took more than the
expected 50ms.

To compensate for this in important paths, spigot/craftbukkit
implements "wall-time". Instead of incrementing/decrementing ticks on
block entities/entities by one for each call to their tick() method,
they instead increment/decrement important values, like
an ItemEntity's age or pickupDelay, by the difference of
`currentTick - lastTick`, where `lastTick` is the value of
`currentTick` during the last tick() call.

These "fixes" however do not play nicely with minecraft's simulation
distance as entities/block entities implementing the above behaviour
would "catch up" their values when moving from a non-ticking chunk to a
ticking one as their `lastTick` value remains stuck on the last tick in
a ticking chunk and hence lead to a large "catch up" once ticked again.

Paper completely removes the "tick skipping" behaviour (See patch
"Further-improve-server-tick-loop"), making the above precautions
completely unnecessary, which also rids paper of the previous described
incompatibility with non-ticking chunks.


lynxplay:
Notice for reviewers, the first commit removes a "related" diff that was noticed during the removal of wall-time.
Technically not related, but still useless.

Machine-Maker (approved):

kennytv (approved):

PR #11413, opened 2024-09-17
Only mark decorations dirty if a removal actually occurs

electronicboy:
Vanilla calls this method blindly inside of a loop which erroniously marks
map data files as being dirty even if nothing has actually changed.

I'm not sure if this will have a big impact as it looks like something
may have changed around here? But, this is something we identified occuring
fairly often which would cause issues on servers with lots of item frame maps.


lynxplay (approved):

lynxplay:
Should be merged into `Improve Maps (in item frames) performance and bug fixes` by the merger.

kennytv (approved):

NoahvdAa (approved):

Issue #9877, opened 2023-10-28
Using adventure components in conversion APIs


lijinhong11:
### Is your feature request related to a problem?

Prompt#getPromptText(ConversationContext) requires a return type of string, which is too old for me. I only create message handlers that only convert message to net.kyori.adventure.text.Component. I don't want to make another string converter.

### Describe the solution you'd like.

1. Change Prompt#getPromptText(ConversationContext)'s return type to net.kyori.adventure.text.Component.
2. Or create another basic prompt class to use net.kyori.adventure.text.Component

### Describe alternatives you've considered.

Create another class, but I don't know how the conversation system works. So, I have to give up.

Owen1212055:
There was a similar issue to this opened in the past. In general, we are more set on deprecating this api.

lijinhong11:
so when can it be true

kashike:
The conversation API will not be expanded to support `Component`s - it will be deprecated entirely at some point.

PR #11414, opened 2024-09-17
Switch to Component API for Conversations API

njsharpe:
Switch over to using Adventure's Component API for the Conversations API to bring things in line with generic Paper standards

### DIFF
--- a/patches/api/0490-Switch-to-Component-API-for-Conversations-API.patch
+++ b/patches/api/0490-Switch-to-Component-API-for-Conversations-API.patch
@@ -1,0 +1,267 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: njsharpe <nelsonj.sharpe@gmail.com>
+Date: Tue, 17 Sep 2024 17:51:52 -0400
+Subject: [PATCH] Switch to Component API for Conversations API
+
+
+diff --git a/src/main/java/org/bukkit/conversations/Conversation.java b/src/main/java/org/bukkit/conversations/Conversation.java
+index bf2407c838bc20197802687c150d513f4e86ed2b..6957d18c8f3f5ecd55ca6bf87b0a7bd29c0c9c2c 100644
+--- a/src/main/java/org/bukkit/conversations/Conversation.java
++++ b/src/main/java/org/bukkit/conversations/Conversation.java
+@@ -219,7 +219,11 @@ public class Conversation {
+ 
+             // Echo the user's input
+             if (localEchoEnabled) {
+-                context.getForWhom().sendRawMessage(prefix.getPrefix(context) + input);
++                // Paper Start - Switch to Component API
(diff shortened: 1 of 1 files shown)


Leguan16:
Correct me if I am wrong but i think you should deprecate the old methods and create entirely new ones that take Components :)

njsharpe:
> Correct me if I am wrong but i think you should deprecate the old methods and create entirely new ones that take Components :)

I cannot deprecate the old methods due to the method signatures. The signature of the old methods and the new methods are the same but the return type is different and because Java doesn't consider the return type in the method signature, I simply need to replace them entirely.

Leguan16:
> > Correct me if I am wrong but i think you should deprecate the old methods and create entirely new ones that take Components :)
> 
> I cannot deprecate the old methods due to the method signatures. The signature of the old methods and the new methods are the same but the return type is different and because Java doesn't consider the return type in the method signature, I simply need to replace them entirely.

Usually you just use the record style getter/setter methods then 
- getInputNotNumericText -> inputNotNumericText
- getPrefix -> prefix
- getPromptText -> promptText

But how you do it right now it breaks backwards compatibility with bukkit and all plugins that used those methods before.

njsharpe:
> > > Correct me if I am wrong but i think you should deprecate the old methods and create entirely new ones that take Components :)
> > 
> > 
> > I cannot deprecate the old methods due to the method signatures. The signature of the old methods and the new methods are the same but the return type is different and because Java doesn't consider the return type in the method signature, I simply need to replace them entirely.
> 
> Usually you just use the record style getter/setter methods then
> 
>     * getInputNotNumericText -> inputNotNumericText
> 
>     * getPrefix -> prefix
> 
>     * getPromptText -> promptText
> 
> 
> But how you do it right now it breaks backwards compatibility with bukkit and all plugins that used those methods before.

You make an excellent point about the record-style formatting for Component-based API changes, this is an oversight on my part and will be resolved ASAP. Thanks!

Owen1212055 (changes requested):
I think that we didn't want to continue this API, and instead just deprecate it.... Although that has not been done yet! :)

njsharpe:
> I think that we didn't want to continue this API, and instead just deprecate it.... Although that has not been done yet! :)

Why though? I rather like this API, as to why I'm making changes for it. Will there be a replacement planned? Capturing user input via the chat box with this API is much easier than book or sign captures. I would hate to see it go...

lynxplay:
Thank you for the PR but sadly Owen is right.
We already declined further maintenance in an earlier [issue](https://github.com/PaperMC/Paper/issues/9877#issuecomment-2081338086).

In general the API is rather old and has become rather undesirable to maintain given its low adoption in the community.
Upstream also does not seem interested in this API being further developed.
Between the small userbase, a lack of interest in our and apparently upstream's team and its *interesting* interactions with chat signatures, the conversation API is, at least for this project, sunset. While it "works", we do not want to invest effort into maintaining patches for the conversation API.

I am sorry you put in all this work into a PR towards a sunset system, but I hope my above points at least make our reasoning clear as to why we are considering it sunset.

Thanks again for the effort!

PR #11425, opened 2024-09-21
remove redundant fillUsableCommands call

Machine-Maker:
Since the brig API PR, we should be able to remove one of the 2 calls to fillUsableCommands. Before brig, it had to populate the map with the vanilla command nodes, but we are more closely matching vanilla now relating to commands and the dispatcher so sticking to vanilla logic here is better.

---
Download the paperclip jar for this pull request: [paper-11425.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/1988419681.zip)


kennytv (approved):

lynxplay (approved):
Tested with paper brig API commands via test plugin and legacy commands via WE:
LGTM.

PR #11430, opened 2024-09-22
Ver/1.20.6

DersWasTaken:
Remove kicking the player for Chat Verification Exceptions


DersWasTaken:
I'm unsure why this is even a thing, but its a pretty simple fix.

I'm not great at git commands, so my commits history is less than optimal.

Warriorrrr:
1\) The 1.20.6 branch is no longer being actively developed, and 2) not kicking the player when the chat chain breaks is not feasible, and will only cause more problems down the line (such as other clients disconnecting themselves when receiving bad chat state).

DersWasTaken:
@Warriorrrr The first one is fair, though I'm unsure if I get the full picture on your second point.

On my end, clients don't kick themselves because the message goes unsent, and even if it was sent, it would appear that clients don't disconnect themselves when receiving a CVE

PR #11436, opened 2024-09-25
Configurable Per-World Void Damage Height Offset & Amount

Axionize:


CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-Contributing to Paper
+gradContributing to Paper
```
lynxplay: What happened here?

patches/api/0490-voidDamageThreshold-Height-at-which-entiies-take-voi.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/src/main/java/org/bukkit/World.java b/src/main/java/org/bukkit/World.java
+index d8a23aa0d898ca3360757721e38ddb97387f7d21..ec1ba8e4164da278f55441e1f3bb4edf724a45d3 100644
+--- a/src/main/java/org/bukkit/World.java
++++ b/src/main/java/org/bukkit/World.java
+@@ -53,6 +53,26 @@ import org.jetbrains.annotations.Nullable;
+ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient, Metadatable, PersistentDataHolder, Keyed, net.kyori.adventure.audience.ForwardingAudience { // Paper
+ 
+     // Paper start
++    /**
++     * @return Sets the damage applied to the player per-tick when they are in the void in this world
```
lynxplay: The method isn't returning anything, `@return` does not make sense here.
It is also missing a `@param voidDamageAmount` description.

Same for the threshold.

patches/server/1059-Implement-configurable-per-world-void-damage-thresho.patch
```diff
@@ -0,0 +1,103 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index bc45c74987974b4828201e06fc8b1f3fbc0af8b4..305b1d80559077333cc5a31a7b217cf2f764cf5d 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -310,7 +310,7 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+     protected int boardingCooldown;
+     @Nullable
+     private Entity vehicle;
+-    private Level level;
++    Level level;
```
lynxplay: Why was this changed?
Entity#level() exists already as a getter.

CONTRIBUTING.md
RealTriassic: Looks like you accidentally edited this?

patches/server/1059-Implement-configurable-per-world-void-damage-thresho.patch
```diff
@@ -0,0 +1,103 @@
...
+     }
+ 
+     public void checkBelowWorld() {
++        if (!this.level.paperConfig().environment.voidDamage.enabled) {
++            return;
++        }
++
+         // Paper start - Configurable nether ceiling damage
+-        if (this.getY() < (double) (this.level.getMinBuildHeight() - 64) || (this.level.getWorld().getEnvironment() == org.bukkit.World.Environment.NETHER
++        if (this.getY() < this.level.paperConfig().environment.voidDamage.damage.heightThreshold  || (this.level.getWorld().getEnvironment() == org.bukkit.World.Environment.NETHER
```
lynxplay: In general, this breaks compatibility with worlds that do not conform to the expected "size" used when you calculated the default of -128.

This could be fixed by simply using DoubleOr.Default, allowing the default to evaluate to the specific min - 64 value.

patches/server/1059-Implement-configurable-per-world-void-damage-thresho.patch
```diff
@@ -0,0 +1,103 @@
...
+             // Paper end - Configurable nether ceiling damage
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+index cd3381d8225e0322b3e285b3c4aa0a1265a91cc0..64b4ee8bb1adf735efda00dfc87f71bc2c60f8bb 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -163,8 +163,30 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+     private final Object2IntOpenHashMap<SpawnCategory> spawnCategoryLimit = new Object2IntOpenHashMap<>();
+     private final CraftPersistentDataContainer persistentDataContainer = new CraftPersistentDataContainer(CraftWorld.DATA_TYPE_REGISTRY);
+     private net.kyori.adventure.pointer.Pointers adventure$pointers; // Paper - implement pointers
++    private float voidDamageAmount;
```
lynxplay: These values seem completely unused, the logic in entity uses the world config, no the CB worlds values.

patches/server/1059-Implement-configurable-per-world-void-damage-thresho.patch
```diff
@@ -0,0 +1,103 @@
...
+         public FrostedIce frostedIce;
++        public VoidDamage voidDamage;
++
++        public class VoidDamage extends ConfigurationPart {
++            public boolean enabled = true;
++            public Damage damage;
++
++            public class Damage extends ConfigurationPart {
++                public double heightThreshold = -128.0;
++                public float amount = 4.0F;
```
lynxplay: This value seems completely unused.

lynxplay (changes requested):
Welcome to paper :tada: 
Thank you for your first PR.

Left a couple of comments that need addressing before something like this can be merged.

patches/server/1059-Implement-configurable-per-world-void-damage-thresho.patch
Machine-Maker: I might prefer just allowing the *offset* from the min build height to be configured. So the default would then be -64. Changing it to 0 would then make damage happen right at the min build height.

Axionize:
I've fixed all the issues you've shown me and changed the PR to be an offset from minBuildHeight() as you said you wanted.

lynxplay:
Do you have a specific use-case for adding API getters/setters?
We'd rather hold off on adding more methods until we have a strong API specific usecase, given this seems rather admin specific with the config values.

The API should already be able to deal void damage via the damage cause API if needed and these methods add to the general clutter that is the World interface already.

lynxplay:
Beyond that, we can combine enabled and damage into a FloatOr.Disabled (needs to be created)

Axionize:
Yes, dynamically loading/building worlds with WorldCreator from region files on disk, mainly for minigames. 

Currently you can generate a custom world at runtime with a custom generator, biomeProvider, etc... but obviously void kill height can not be set when creating the world, much less at runtime.

An alternative may be to modify WorldCreator.java to support passing in a custom world.yaml config settings (and potentially reloading those settings at runtime in World.java), but this is obviously much simpler and I would very much enjoy not having to take on such a task.

lynxplay:
Eh sounds fair, I guess.
I'll chat it over with the rest of the team.

You can deal void damage perfectly fine via https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Damageable.html#damage(double,org.bukkit.damage.DamageSource) and https://jd.papermc.io/paper/1.21.1/org/bukkit/damage/DamageType.html#OUT_OF_WORLD respectively tho.

Maybe it is worth consolidating the void damage configuration into a single interface type, e.g.

World#getVoidDamageConfiguration()
World#setVoidDamageConfiguration(WorldVoidDamageConfiguration)

interface WorldVoidDamageConfiguration {
  
  boolean isVoidDamageEnabled();
  double getVoidDamagePerTick();
  float getMinWorldHeightOffset();

  static WorldVoidDamageConfiguration vanilla() { }
  static WorldVoidDamageConfiguration enabled(double damage, float minWorldHeightOffset) { }
  static WorldVoidDamageConfiguration disabled() { }
}

but I'll get back to concrete next steps once we have more team input on this.
For now, could you please address machines comment regarding the "threshold".
It cannot be a constant y-value, it should be expressed as an offset from the min build height, as the default world configuration otherwise breaks custom height worlds.

Axionize:
I've made it an offset from the build height as requested, defaulting to -64 blocks from the minBuildHeight.

lynxplay (approved):

Machine-Maker (approved):

PR #11438, opened 2024-09-25
Add enchantWithLevels with tag specification

masmc05:
Adds a possibility to enchant an itemstack enchantments from a specific tag

Should I merge this with `0340-Add-enchantWithLevels-API`? This would mean that I will need to move it up after registry modification api

### DIFF
--- a/patches/api/0490-Add-enchantWithLevels-with-tag-specification.patch
+++ b/patches/api/0490-Add-enchantWithLevels-with-tag-specification.patch
@@ -1,0 +1,62 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: masmc05 <masmc05@gmail.com>
+Date: Wed, 25 Sep 2024 16:26:04 +0300
+Subject: [PATCH] Add enchantWithLevels with tag specification
+
+
+diff --git a/src/main/java/org/bukkit/inventory/ItemFactory.java b/src/main/java/org/bukkit/inventory/ItemFactory.java
+index bd0e55562f1cabef3078573182e0cf9fbc844585..b7a5a9d294333541e36de89c8556ddb9665e8555 100644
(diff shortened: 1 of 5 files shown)


patches/api/0490-Add-enchantWithLevels-with-tag-specification.patch
```diff
@@ -0,0 +1,32 @@
...
++     * <p>Levels must be in range {@code [1, 30]}.</p>
++     *
++     * @param itemStack ItemStack to enchant
++     * @param levels levels to use for enchanting
++     * @param tagKey tag key which will be used to determine the enchantments to apply (e.g. {@link io.papermc.paper.registry.keys.tags.EnchantmentTagKeys#IN_ENCHANTING_TABLE})
++     * @param random {@link java.util.Random} instance to use for enchanting
++     * @return enchanted copy of the provided ItemStack
++     * @throws IllegalArgumentException on bad arguments
++     */
++    @NotNull ItemStack enchantWithLevels(@NotNull ItemStack itemStack, @org.jetbrains.annotations.Range(from = 1, to = 30) int levels, @NotNull io.papermc.paper.registry.tag.TagKey<Enchantment> tagKey, @NotNull java.util.Random random);
```
Machine-Maker: It would probably be better to make the parameter type, `RegistryKeySet<Enchantment>`, that way it could be a tag **or** just a specific set enchants a plugin might want to have available.

Machine-Maker (changes requested):
I think its fine to have it as a separate patch at the end now. Or move the other enchantWithLevels patch to the end. Either way, don't do either until right before merging cause its way easier to review as is.

Machine-Maker:
Not sure how I feel about having the method on ItemStack directly. Do we have any other methods on ItemStack which operate on copies of itself?

lynxplay (approved):

Machine-Maker (approved):

PR #11448, opened 2024-09-29
Begin switching to JSpecify annotations

Machine-Maker:
Adjusts the annotation test to check if the class as the NullMarked annotation, and if it does, to skip the whole class.

This will probably be several PRs so it's not just one HUGE pr.

### DIFF
--- a/patches/api/0002-Build-system-changes.patch
+++ b/patches/api/0002-Build-system-changes.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] Build system changes
 
 
 diff --git a/build.gradle.kts b/build.gradle.kts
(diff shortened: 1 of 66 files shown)


electronicboy:
Because JSpecify is the new annotation library developed by a commit, including JetBrains, google, etc., in order to create one true framework. it being new means that it's not going to have as many users yet.

electronicboy:
Once again, it's new; it's not going to show up high on the lists, they don't blindly show preference to their own library

powercasgamer:
>  It doesn't even show on the list.

It obviously won't show up in the list if you haven't added it to your project. https://i.imgur.com/wixYnGI.png

kennytv:
There is no "default" included nullness annotations beyond javax', jetbrains annotations are their own library you have to explicitly depend on like any other. If you need help updating your dependencies or understanding the purpose of these libraries, you can ask in our Discord

PR #11453, opened 2024-09-30
Fix/scoreboard delegate

### LINKED ISSUE #11296: Entity#getScoreboardTags directly exposes the Set<String>
An entity will only load 1024 scoreboard tags from it's NBT, and the addTag methods on nms.Entity automatically check this and return false if it's over 1024. These methods are directly called by the API addScoreboardTag method. The problem is, the API also has `getScoreboardTags` which directly exposes the underlying set, no copy, no unmodifiable wrapper, nothing. So a plugin can just add strings to that set to add over 1024 tags which are then saved, and lost on a reload of the entity.

Either we copy the set to make it a mutable copy or wrap it in an unmodifiable wrapper or we do an immutable copy.

Intybyte:
Fixes #11296


Leguan16 (commented):
Usually, if nms/bukkit files are modified FQN should be used instead of imports to reduce the diff. Also, every line that is added needs to be marked with a comment. For more information refer to the CONTRIBUTING.md :)

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
+From: Intybyte <vaan.testwork@gmail.com>
+Date: Mon, 30 Sep 2024 17:16:36 +0200
+Subject: [PATCH] Create Limited set for Entity
+
+
+diff --git a/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java b/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..456c8cdb38d29a73a96dc44f7e8e4ce799cbab38
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java
```
Machine-Maker: Lets put this in some `util` package or smth. Instead of `entity`. Also, just call it like `SizeLimitedSet`, Paper doesn't really apply to it.

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
+@@ -0,0 +1,40 @@
++package io.papermc.paper.entity;
++
++import org.jetbrains.annotations.NotNull;
++import java.util.Collection;
++import java.util.HashSet;
++
++public class PaperLimitedSet extends HashSet<String> {
```
Machine-Maker: Use `@NullMarked`

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
+
+diff --git a/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java b/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..456c8cdb38d29a73a96dc44f7e8e4ce799cbab38
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/entity/PaperLimitedSet.java
+@@ -0,0 +1,40 @@
++package io.papermc.paper.entity;
++
++import org.jetbrains.annotations.NotNull;
```
Machine-Maker: Use `@NullMarked` on the type and remove any `NotNull/NonNull` annotations.

Machine-Maker (dismissed):

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
++        this.maxTags = maxTags;
++    }
++
++    @Override
++    public boolean add(String tag) {
++        // Enforce the max tag limit
++        if (this.size() >= maxTags) {
++            return false;
++        }
++        return this.add(tag);
```
imDaniX: Did you mean `super.add(tag)`? Otherwise, seems like a recursive loop to me

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
++    }
++
++    @Override
++    public boolean addAll(@NotNull Collection<? extends String> tags) {
++        int currentSize = this.size();
```
imDaniX: The `currentSize` field seems redundant? We already have the `size()` method, which is automatically increased anyway

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,74 @@
...
++
++        for (String tag : tags) {
++            if (currentSize >= maxTags) {
++                break;
++            }
++
++            boolean added = this.add(tag);
++            if (added) currentSize++;
++
++            edited = added || edited;
```
imDaniX: Without `currentSize`, this can be simplified to `edited |= this.add(tag)`.

imDaniX (changes requested):

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,71 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index 4b54d0ea31062972e68ee8fafe3cfaf68f65a5cd..94e6637045e4a4f0a4d79d3aba245b13465dfa27 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -566,7 +566,7 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+         this.packetPositionCodec = new VecDeltaCodec();
+         this.uuid = Mth.createInsecureUUID(this.random);
+         this.stringUUID = this.uuid.toString();
+-        this.tags = Sets.newHashSet();
++        this.tags = new io.papermc.paper.util.SizeLimitedSet(1024); // Paper - Create Limited set for Entity
```
Lulu13022002: you can use the constant MAX_ENTITY_TAG_COUNT

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,71 @@
...
++    private final int maxTags;
++
++    public SizeLimitedSet(int maxTags) {
++        this.maxTags = maxTags;
++    }
++
++    @Override
++    public boolean add(String tag) {
++        // Enforce the max tag limit
++        if (this.size() >= maxTags) {
```
Lulu13022002: Add this. before field access

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,71 @@
...
++++ b/src/main/java/io/papermc/paper/util/SizeLimitedSet.java
+@@ -0,0 +1,37 @@
++package io.papermc.paper.util;
++
++import org.jspecify.annotations.NullMarked;
++import java.util.Collection;
++import java.util.HashSet;
++
++@NullMarked
++public class SizeLimitedSet extends HashSet<String> {
```
Lulu13022002: I feel the name is too generic for what this does unless you add a generic to it and remove tag reference.

patches/server/1065-Create-Limited-set-for-Entity.patch
```diff
@@ -0,0 +1,71 @@
...
+@@ -0,0 +1,37 @@
++package io.papermc.paper.util;
++
++import org.jspecify.annotations.NullMarked;
++import java.util.Collection;
++import java.util.HashSet;
++
++@NullMarked
++public class SizeLimitedSet<T> extends HashSet<T> {
++    private final int maxTags;
```
Amejonah1200: It has nothing to do with tags, consider renaming to `maxSize` instead.

electronicboy (approved):
I do ponder if we should move the set over to mc-util, but, otherwise, this lgtm

lynxplay (approved):

Intybyte:
what are the advantages of using a delegate in this case?

electronicboy:
it's a reusable generic container wrapper class

lynxplay:
The fact that we can reuse this type for different set implementations. The previous implementation would lock the type into the speed/contracts of hashset.

PR #11454, opened 2024-09-30
Configurable Entity Despawn Time

Machine-Maker:
Despawn time limit configurable by each entity type. Overrides any other configurations for despawning, such as the arrow or item despawn rate configurations which give more granularity.


patches/server/1065-Configurable-Entity-Despawn-Time.patch
```diff
@@ -0,0 +1,39 @@
...
+         this.passengers = ImmutableList.of();
+         this.deltaMovement = Vec3.ZERO;
+         this.bb = Entity.INITIAL_AABB;
+@@ -845,6 +847,12 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+     }
+ 
+     public void tick() {
++        // Paper start - entity despawn time limit
++        if (this.despawnTime >= 0 && this.tickCount >= this.despawnTime) {
++            this.discard();
```
lynxplay: Needs a despawn reason, I'd argue for `org.bukkit.event.entity.EntityRemoveEvent.Cause.DESPAWN` but it may be worth introducing a new cause for this instead too.

kennytv (approved):

lynxplay:
https://github.com/PaperMC/docs/pull/497

lynxplay (approved):

PR #11455, opened 2024-09-30
Fix and enhance openInventory(InventoryView)

masmc05:
The new method allows overriding the title on the menu which is being opened without the necessity to listen to InventoryOpenEvent and setting [InventoryOpenEvent#titleOverride(Component)](<https://jd.papermc.io/paper/1.21.1/org/bukkit/event/inventory/InventoryOpenEvent.html#titleOverride(net.kyori.adventure.text.Component)>)

When creating this new api discovered an issue with the current method which has a very small fix, the issue being that currently when running
```java
    @EventHandler
    public void onInventoryOpen(org.bukkit.event.inventory.InventoryOpenEvent event) {
        if (event.getInventory() instanceof HorseInventory) {
            event.getPlayer().getScheduler().runDelayed(this, task -> {
                event.getPlayer().openInventory(event.getPlayer().getOpenInventory());
            }, () -> {}, 40L);
        }
    }
```
You will get this error https://pastes.dev/yAFvj1yF6I

Should I instead of current behavior, pass the title override as the default value of InventoryOpenEvent#titleOverride?


Machine-Maker:
You can already create views with any title via the MenuType API. Is there a point to adding this with the MenuType API already in?

PR #11460, opened 2024-10-01
Fix annotation testing could be interrupted

LemonCaramel:
[comment](https://github.com/PaperMC/Paper/commit/f9c7f2a5c125e92a986f146ba859de189e26814b#diff-824941af49178f5596d8df15aca56c8b02934fae22f80e87e758dfff5410f227R112)

Also, the issue of failing tests in nested classes has been resolved.

### DIFF
--- a/patches/api/0003-Test-changes.patch
+++ b/patches/api/0003-Test-changes.patch
@@ -66,7 +66,7 @@ index 0000000000000000000000000000000000000000..77154095cfb8b259bdb318e8ff40cb6f
 +    }
 +}
 diff --git a/src/test/java/org/bukkit/AnnotationTest.java b/src/test/java/org/bukkit/AnnotationTest.java
-index 64e7aef6220097edefdff3b98a771b988365930d..7ff939ea41417bad3a436a87c89d5efa7ecefe86 100644
+index 64e7aef6220097edefdff3b98a771b988365930d..f8b8969ee7a0b6f7b3224ff081e35c14a398c9d0 100644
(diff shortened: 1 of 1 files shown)


kennytv:
Not sure if nullmarked actually applies to inner classes, those might be valid fails

patches/api/0003-Test-changes.patch
```diff
@@ -187,20 +187,31 @@ index 64e7aef6220097edefdff3b98a771b988365930d..7ff939ea41417bad3a436a87c89d5efa
...
      private static void collectClasses(@NotNull File from, @NotNull Map<String, ClassNode> to) throws IOException {
-@@ -125,6 +186,12 @@ public class AnnotationTest {
+@@ -125,6 +186,23 @@ public class AnnotationTest {
          }
      }
  
 +    // Paper start - skip class if it's @NullMarked
-+    private static boolean isClassNullMarked(@NotNull ClassNode clazz) {
++    private static boolean isClassNullMarked(@NotNull ClassNode clazz, @NotNull Map<String, ClassNode> allClasses) {
++        if (clazz.nestHostClass != null) {
```
lynxplay, line 196: ~~While technically useless nitpick, this would fail if the annotation was present on the inner class but not on the outer one? E.g. this may instead have to be something like. if we truely care about supporting inner types correctly in this. Not that that is really important lol~~ never mind

Machine-Maker (approved):

lynxplay (approved):

PR #11464, opened 2024-10-02
Don't run player loot table for spectators

### LINKED ISSUE #11463: Spectators drop player loot table
### Expected behavior

Players in spectator mode should NOT drop a loot table on death, just like in vanilla.

### Observed/Actual behavior

Players in spectator mode drop the `minecraft:entities/player` loot table on death.

### Steps/models to reproduce

1. Install this [minimal reproduction data pack](https://github.com/user-attachments/files/17235653/player-loot-table.zip) by copying it into the server's `world > datapacks` folder. This adds a player head to the player's loot table unconditionally.
2. Op yourself in the server console.
3. Join the server, and enter `/gamemode spectator`.
4. Enter `/kill @s`.
5. Observe that a player head dropped. In vanilla, nothing drops.

### Plugin and Datapack List

Just the minimal reproduction data pack.

### Paper version

```
> version
[16:38:13 INFO]: This server is running Paper version 1.21.1-118-master@fb76840 (2024-10-02T12:49:56Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21.1-2315-2a50638 (MC: 1.21.1)
```

### Other

I have not tested if this is also a bug upstream in Spigot, as the Spigot download link currently times out for me.

Machine-Maker:
I think there's also an issue with spectators and xp dropping, but I can't test that out right now. Until I check that, this can be a draft.

Fixes https://github.com/PaperMC/Paper/issues/11463

---
Download the paperclip jar for this pull request: [paper-11464.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2008118646.zip)


Doc94:
I test and experience its not dropped in spectator mode but this is because the "keepLevel" behaviour use the same logic for keep is the keep inventory is true or the player is in spectator. also the "logic" for get how many exp need to be dropped is 0 for spectator but not sure from where comes that.. i just found a call for `EnchantmentHelper.processMobExperience` but not sure.

PR #11469, opened 2024-10-05
Add ItemStack#effectiveName

masmc05:
Adds a new method which returns the name of an itemstack which the player sees in their inventory. Takes into account any data components exactly as vanilla.

Unlike [ItemStack#displayName](https://jd.papermc.io/paper/1.21.1/org/bukkit/inventory/ItemStack.html#displayName()), does not add additional brackets etc


electronicboy (dismissed):

lynxplay:
Duplicate of #9050

kennytv:
Something else that was brought up internally is the whole "display name" naming being outdated/confusing in general now. effectiveName is good for the new method on ItemStack, but we'd like this to go alongside improving ItemStack#displayName docs and replacing ItemMeta displayName methods with customName (since that's what it returns, but other data components can change or be the effective display name). Replace in this case means marking the old as obsolete and linking to the new

kennytv:
Yeah sure

jpenilla:
I'm confused with what's going on in this diff, what kenny asked for was:
- Improve docs for ItemStack#displayName()
- Deprecate/obsolete ItemMeta#displayName(Component) and ItemMeta#displayName() in favor of new customName methods
- Add ItemStack#effectiveName() method

Malfrador:
Not a fan of deprecating the ItemMeta method either.
Yes its objectively wrong since two versions, but that thing has been called "display name" in vanilla too for ten years now and just spamming more deprecation will make more people simply ignore them.

PR #11472, opened 2024-10-06
Configurable bypass for anvil enchantment over the max level

Doc94:
Currently if you try to use an anvil with a enchantment book using levels over the max defined then the anvil just ignore level in book and override with the max level defined in registry, this PR allow by config disable this for allow using that enchantment books.

PS: Talking about why this... in my case is just wanna give books with things like sharpness 7 and dont wanna listen the PrepareAnvilEvent and make a whole thing for check compatibility and manually set all in code... and also dont wanna override the max level in bootstrap because this make players can get this levels in vanilla ways.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -487,10 +487,10 @@ index 0000000000000000000000000000000000000000..d9502ba028a96f9cc846f9ed428bd806
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java b/src/main/java/io/papermc/paper/configuration/GlobalConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..b92bfd89e32becde2e7630c6116c16f8a4f6614a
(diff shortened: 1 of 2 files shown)


Machine-Maker:
I feel like an event for this is better. Some event for books combining would provide more control.

Doc94:
> I feel like an event for this is better. Some event for books combining would provide more control.

Like AnvilEnchantItemResult? Im not sure how handle that or what can this event do in that place

Doc94:
> I think a PreEnchantCombinationEvent or PreAnvilCombinationEvent where you can override different things like what the base item will be used (so for example you could do pumpkin + enchantment book = enchanted diamond), ignore some limits etc

but the thing with items is this already being handle in the PrepareAnvil (but called in multiple things and for that i dont wanna mess with this)

or just add a value in the AnvilView for that and make the AnvilMenu handle from this for make like in the OpenInventory event override the behaviour...

Doc94:
Based in comments i create this https://github.com/PaperMC/Paper/pull/11475 where allow change the bevahiour per Anvil View.

PR #11475, opened 2024-10-06
Allow change max level enchanting behaviour in anvil using AnvilView

Doc94:
This PR allow change the behaviour in anvils where enchantments applied are limited to the max level registered.

Code to test a case where only books can override the max level limit.
```java
    @EventHandler
    public void onAnvilPrepare(PrepareAnvilEvent event) {
        final AnvilView anvilView = event.getView();
        final ItemStack itemStackIngredient = anvilView.getItem(1);
        if (itemStackIngredient == null || itemStackIngredient.isEmpty()) {
            return;
        }
        if (itemStackIngredient.getType() == Material.ENCHANTED_BOOK) {
            event.getView().bypassEnchantmentLevelRestriction(true);
        } else {
            event.getView().bypassEnchantmentLevelRestriction(false);
        }
    }
```

PS: This PR is an alternative from https://github.com/PaperMC/Paper/pull/11472


Doc94:
> > setCanApplyEnchantmentsOverMaxLevel
> 
> the name is too long....
> 
> Maybe `ignoresEnchantmentMaxLevel()` and `ignoresEnchantmentMaxLevel(value)`?

i wanna the method being a little explain about what can do (or maybe just a thing for docs) because ignoresEnchantmentMaxLevel sounds like a ignore the level not the "limit" or clamp... but still not sure about the names...

lynxplay:
`bypassEnchantmentLevelRestriction(boolean)` ?
I 100% agree with masjson tho, that name is not something I'd be interesting in merging as it is giving some strong spigot vibes xD

Doc94:
> `bypassEnchantmentLevelRestriction(boolean)` ? I 100% agree with masjson tho, that name is not something I'd be interesting in merging as it is giving some strong spigot vibes xD

im not good for names :c

but `bypassEnchantmentLevelRestriction` sounds a good method name and the docs talks more about how this means to the max level then i think can work (and can allow users again use that custom books)

lynxplay (approved):

PR #11479, opened 2024-10-08
Add new event TickRateChangedEvent

Axionize:
This event fires when the tick rate is changed, either by a command or by a plugin. 

Useful if you have code dependent on the tickRate (my use case is simulating and predicting possible player movements) and you don't want to have to poll the tickRate all the time.

### DIFF
--- a/patches/api/0495-Add-new-TickRateChangedEvent.patch
+++ b/patches/api/0495-Add-new-TickRateChangedEvent.patch
@@ -1,0 +1,84 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Axionize <154778082+Axionize@users.noreply.github.com>
+Date: Tue, 8 Oct 2024 20:38:01 -0400
+Subject: [PATCH] Add new TickRateChangedEvent
+
+
+diff --git a/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java b/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..c68ef49ea9fa61b869fce1ba3f73a9272bde2ce7
+--- /dev/null
++++ b/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java
+@@ -0,0 +1,72 @@
++package org.bukkit.event.server;
++
++import com.google.common.base.Preconditions;
++import io.papermc.paper.command.brigadier.CommandSourceStack;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++import org.jetbrains.annotations.Nullable;
++import org.jspecify.annotations.NullMarked;
++
++/**
++ * Fired when the server's target tickrate changes
++ */
++@NullMarked
++public class TickRateChangeEvent extends ServerEvent implements Cancellable {
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    private final CommandSourceStack commandSourceStack;
++    private float tickRate;
++    private boolean cancelled;
++
++    @ApiStatus.Internal
++    public TickRateChangeEvent(CommandSourceStack commandSourceStack, float tickRate) {
++        this.commandSourceStack = commandSourceStack;
++        this.tickRate = tickRate;
++    }
++
++    /**
++     * @return {@link io.papermc.paper.command.brigadier.CommandSourceStack} representing source of the /tick rate command
++     * Will return null if the tick rate was set by a plugin or setTickRate() was otherwise directly called
++     */
++    @Nullable
++    public CommandSourceStack getCommandSourceStack() {
++        return commandSourceStack;
++    }
++
++    /**
++     * @return float representing new tick rate of the server between 1.0F and 10,000F
++     */
++    public float getTickRate() {
++        return tickRate;
++    }
++
++    /**
++     * @param tickRate float overriding the new tickrate being applied to the server
++     */
++    public void setTickRate(float tickRate) {
++        Preconditions.checkArgument(tickRate >= 1.0F && tickRate <= 10_000.0F, "The given tick rate must not be less than 1.0 or greater than 10,000.0");
++        this.tickRate = tickRate;
++    }
++
++    @Override
++    public @NotNull HandlerList getHandlers() {
++        return HANDLER_LIST;
++    }
++
++    public static HandlerList getHandlerList() {
++        return HANDLER_LIST;
++    }
++
++    @Override
++    public boolean isCancelled() {
++        return cancelled;
++    }
++
++    @Override
++    public void setCancelled(boolean cancel) {
++        cancelled = cancel;
++    }
++}
--- a/patches/server/1065-Fire-and-handle-new-cancellable-TickRateChangedEvent.patch
+++ b/patches/server/1065-Fire-and-handle-new-cancellable-TickRateChangedEvent.patch
@@ -1,0 +1,67 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Axionize <154778082+Axionize@users.noreply.github.com>
+Date: Tue, 8 Oct 2024 20:32:34 -0400
+Subject: [PATCH] Fire and handle new cancellable TickRateChangedEvent
+
+
+diff --git a/src/main/java/net/minecraft/server/ServerTickRateManager.java b/src/main/java/net/minecraft/server/ServerTickRateManager.java
+index 37dcf3dc3e50afd85912a7496c828576a38a4e9c..e9198a83de38e5f0fc564f6c6c87c9cdbf1abd67 100644
+--- a/src/main/java/net/minecraft/server/ServerTickRateManager.java
++++ b/src/main/java/net/minecraft/server/ServerTickRateManager.java
+@@ -121,13 +121,19 @@ public class ServerTickRateManager extends TickRateManager {
+         this.sprintTimeSpend += System.nanoTime() - this.sprintTickStartTime;
+     }
+ 
+-    @Override
+-    public void setTickRate(float tickRate) {
++    // Paper start - Fire and handle cancellable TickRateChangeEvent
++    public void setTickRate(io.papermc.paper.command.brigadier.CommandSourceStack commandSourceStack, float tickRate) {
+         super.setTickRate(tickRate);
+         this.server.onTickRateChanged();
+         this.updateStateToClients();
+     }
+ 
++    @Override
++    public void setTickRate(float tickRate) {
++        setTickRate(null, tickRate);
++    }
++    // Paper end - Fire and handle cancellable TickRateChangeEvent
++
+     public void updateJoiningPlayer(ServerPlayer player) {
+         player.connection.send(ClientboundTickingStatePacket.from(this));
+         player.connection.send(ClientboundTickingStepPacket.from(this));
+diff --git a/src/main/java/net/minecraft/server/commands/TickCommand.java b/src/main/java/net/minecraft/server/commands/TickCommand.java
+index 5ce845a9cd84c355e2716dfcb0b62686c783c9f9..4de4b27ec563a9b37de4355374a1ff45fa8bd551 100644
+--- a/src/main/java/net/minecraft/server/commands/TickCommand.java
++++ b/src/main/java/net/minecraft/server/commands/TickCommand.java
+@@ -61,7 +61,12 @@ public class TickCommand {
+ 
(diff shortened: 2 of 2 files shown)


dawon:
Isn't the api/event itself missing?

Axionize:
Thank you for noticing that. Not used to the paper build system. Let me fix that real quick.

Axionize:
@masmc05 I addressed all the changes you requested. Added comments starting where my patch modifies code, switched to using event.callEvent() and the patch now uses the FQDN instead of imports for changes in Paper-Server.

patches/server/1065-Fire-and-handle-canellable-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,51 @@
...
+--- a/src/main/java/net/minecraft/server/ServerTickRateManager.java
++++ b/src/main/java/net/minecraft/server/ServerTickRateManager.java
+@@ -121,12 +121,22 @@ public class ServerTickRateManager extends TickRateManager {
+         this.sprintTimeSpend += System.nanoTime() - this.sprintTickStartTime;
+     }
+ 
++    // Paper start - Fire and handle cancellable TickRateChangeEvent
++    public void setTickRate(org.bukkit.command.CommandSender commandSender, float tickRate) {
++        org.bukkit.event.server.TickRateChangeEvent event = new org.bukkit.event.server.TickRateChangeEvent(commandSender, tickRate);
++        event.callEvent();
```
lynxplay: callEvent yields back a boolean indicating cancellation state.

patches/server/1065-Fire-and-handle-canellable-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,51 @@
...
++            this.updateStateToClients();
++        }
++    }
++
+     @Override
+     public void setTickRate(float tickRate) {
+-        super.setTickRate(tickRate);
+-        this.server.onTickRateChanged();
+-        this.updateStateToClients();
++        setTickRate(null, tickRate);
```
lynxplay: This diff can be minimized by moving the override down and doing an early return instead of an if(notCancelled).

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,72 @@
...
++
++import org.bukkit.command.CommandSender;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.jetbrains.annotations.NotNull;
++
++/**
++ * Fired when the server's target tickrate changes
++ */
++public class TickRateChangeEvent extends ServerEvent implements Cancellable {
```
lynxplay, line 28: `@NullMarked`

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,72 @@
...
++  public TickRateChangeEvent(CommandSender commandSender, float tickRate) {
++    this.commandSender = commandSender;
++    this.tickRate = tickRate;
++  }
++
++  /**
++   * @return {@link org.bukkit.command.CommandSender} representing source of the /tick rate command
++   * Will return null if the tick rate was set by a plugin or setTickRate() was otherwise directly called
++   */
++  public CommandSender getCommandSender() {
```
lynxplay: `@Nullable`

lynxplay (changes requested):

patches/api/0495-Add-new-TickRateChangedEvent.patch
Lulu13022002, line 1: Indentation should be 4 spaces wide, javadoc description and nullable annotations are missing (and should be jspecify ones)

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,72 @@
...
++ * Fired when the server's target tickrate changes
++ */
++public class TickRateChangeEvent extends ServerEvent implements Cancellable {
++
++  private float tickRate;
++  private final CommandSender commandSender;
++  private boolean cancelled;
++  private static HandlerList handlers = new HandlerList();
++
++  public TickRateChangeEvent(CommandSender commandSender, float tickRate) {
```
Lulu13022002: Missing ApiStatus$Internal annotation on the constructor

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,72 @@
...
++
++/**
++ * Fired when the server's target tickrate changes
++ */
++public class TickRateChangeEvent extends ServerEvent implements Cancellable {
++
++  private float tickRate;
++  private final CommandSender commandSender;
++  private boolean cancelled;
++  private static HandlerList handlers = new HandlerList();
```
Lulu13022002: Constant should be uppercased and before the fields

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,72 @@
...
++   */
++  public float getTickRate() {
++    return tickRate;
++  }
++
++  /**
++   *
++   * @param tickRate float overriding the new tickrate being applied to the server
++   */
++  public void setTickRate(float tickRate) {
```
Lulu13022002: The setter doesn't seems to be implemented

patches/api/0495-Add-new-TickRateChangedEvent.patch
Lulu13022002, line 28: This event doesn't implement the static method getHandlerList so it just doesn't work at all and plugins that use it fail to load

patches/server/1065-Fire-and-handle-canellable-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,51 @@
...
+index 5ce845a9cd84c355e2716dfcb0b62686c783c9f9..727e7df4231bc5797ab3e2046111b9483dab40fa 100644
+--- a/src/main/java/net/minecraft/server/commands/TickCommand.java
++++ b/src/main/java/net/minecraft/server/commands/TickCommand.java
+@@ -61,7 +61,9 @@ public class TickCommand {
+ 
+     private static int setTickingRate(CommandSourceStack source, float rate) {
+         ServerTickRateManager serverTickRateManager = source.getServer().tickRateManager();
+-        serverTickRateManager.setTickRate(rate);
++        // Paper start - Fire and handle cancellable TickRateChangeEvent
++        serverTickRateManager.setTickRate(source.getBukkitSender(), rate);
```
Lulu13022002: Single line comment should be enough here

Axionize: it is a single line what are you talking about

Axionize:
I've addressed all issues that have been brought up. Is there anything else left to do before this is merged?

patches/server/1065-Fire-and-handle-canellable-TickRateChangedEvent.patch
kokiriglade: > it is a single line what are you talking about

the line is wrapped in `// Paper start` and `// Paper end` - just put `// Paper - Fire and handle cancellable TickRateChangeEvent` at the end of the line 🙂

Axionize: Understood, done

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,82 @@
...
++        this.commandSender = commandSender;
++        this.tickRate = tickRate;
++    }
++
++    /**
++     * @return {@link org.bukkit.command.CommandSender} representing source of the /tick rate command
++     * Will return null if the tick rate was set by a plugin or setTickRate() was otherwise directly called
++     */
++    @Nullable
++    public CommandSender getCommandSender() {
```
lynxplay: Brought up during some chatting regarding this PR, but it would probably be worth exposing the entire io.papermc.paper.command.brigadier.CommandSourceStack. Exposes more context over what is essentially a "Someone ran the tick command" event.

Axionize: Done

Was the chatting about this PR in a public channel I can see?

Lulu13022002:
The event mix jetbrains and jspecify nullable annotations. And it's a bit iffy when a plugin change the tick rate cause the command executor doesn't receive the right tick rate in the feedback message (unlike the WorldGameRuleChangeEvent for example)

patches/api/0495-Add-new-TickRateChangedEvent.patch
lynxplay: Thanks for the quick fix, no that was just a fleeting discussion in internal channels.
I'll try to make time for merging it this weekend, which usually happens in one of our discord VCs.

Axionize:
I'll fix the notations but I don't understand your second concern, what exactly is the problem? That the message returned to the player won't reflect the new tick rate?

patches/api/0495-Add-new-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,82 @@
...
+From: Axionize <154778082+Axionize@users.noreply.github.com>
+Date: Tue, 8 Oct 2024 20:38:01 -0400
+Subject: [PATCH] Add new TickRateChangedEvent
+
+
+diff --git a/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java b/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..279e471f00495cc08ed2297efb21fb801cd43851
+--- /dev/null
++++ b/src/main/java/org/bukkit/event/server/TickRateChangeEvent.java
```
electronicboy, line 11: Paper events go in the paper package

electronicboy:
> That the message returned to the player won't reflect the new tick rate?

Yes, the server prints the output of the tick rate, and yet, if a plugin mutates it, this value will be wrong; if we want to support mutation, the logic should be adjusted to get the new rate so that the message is displayed properly, rather than having invisible side-effects to end users

patches/api/0495-Add-new-TickRateChangedEvent.patch
Axionize: I was thinking about upstreaming this into Spigot later

electronicboy: if that's the case, then this PR is inherently blocked as it makes little sense for us to finalise the review and such of this into a state we're happy to merge it if we're going to potentially have to deal with breaking ABI changes from upstream

lynxplay: Which will inherently be the case as spigot does not expose the CommandSourceStack this event now provides.
Would be great if you can come to a decision whether or not you want to have this event in spigot-api soonish then. As cat mentioned, if this is going upstream, there is 0 point in this PR receiving reviews on paper and we can close it instead.

lynxplay: Any update on this? I'd really like to either merge or close this depending on your decision regarding moving this to upstream.

Axionize:
I've restructured this PR so that mutations will be reflected in return values.

patches/api/0495-Add-new-TickRateChangedEvent.patch
Axionize: I've decided to leave it in paper. Are there any other changes you'd like me to make?

patches/server/1065-Fire-and-handle-new-cancellable-TickRateChangedEvent.patch
```diff
@@ -0,0 +1,80 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/LivingEntity.java b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+index 2aa6374cd4a96efd85899be8cd3172a8257bfe6b..3a5382ec47584815ec4a5074c3334a5984721701 100644
+--- a/src/main/java/net/minecraft/world/entity/LivingEntity.java
++++ b/src/main/java/net/minecraft/world/entity/LivingEntity.java
+@@ -2343,7 +2343,7 @@ public abstract class LivingEntity extends Entity implements Attackable {
+                     }
+                 } else {
+                     ((net.minecraft.world.entity.player.Player) damagesource.getEntity()).resetAttackStrengthTicker();
+-                }
++               a }
```
masmc05: a

patches/api/0495-Add-new-TickRateChangedEvent.patch
lynxplay: Well as cat stated, the event should be in the paper event package

Axionize:
Hi, is there anything left for me to do before this can be merged?

PR #11495, opened 2024-10-14
Add run configurations to intellij

brickmonster:
Add run configurations to intellij that run the gradle server tasks. More can be added.

It's just convenient to not have to add them yourself when debugging or building repeatedly.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,6 +3,8 @@ import io.papermc.paperweight.tasks.BaseTask
 import io.papermc.paperweight.util.*
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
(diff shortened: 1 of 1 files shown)


lynxplay:
I don't know how useful I find such a change, at least not for my workflow.

It removes the need for three to four clicks but instead adds "random" run configurations for people that might have a different naming schema. I'll leave it open for some other team members to chime in, I don't have a strong opinion on this either way.

jpenilla:
This is just adding certain Gradle tasks as runs to the IDE, which can be done manually for the tasks someone cares about by running them once from the IJ Gradle menu (after which they will stay in the run menu). I don't see any utility here. Actual IntelliJ run configs that don't invoke Gradle could be interesting but we would likely need to generate the XML directly as the intellij-ext plugin is quite limiting.

PR #11498, opened 2024-10-18

xymb-endcrystalme:


electronicboy:
This would likely want to be PR'd to moonrise and then ported over, too

xymb-endcrystalme:
I didn't even know moonrise was a thing. 😁 All I really did was to follow the convention.

Since I'm definitely not gonna learn how to Mixin and stuff, I guess I'll just remove all incorrect associations with moonrise. After all, it's not their code.

PR #11500, opened 2024-10-18

xymb-endcrystalme:

### DIFF
--- a/patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
+++ b/patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
@@ -1,0 +1,95 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Xymb <xymb@endcrystal.me>
+Date: Fri, 18 Oct 2024 20:44:05 +0200
+Subject: [PATCH] Memory optimization: Lazy initialization of IBlockDataList
+ for ticking blocks.
+
+In my tests (1 player, large render distance) it lowered heap dump size from 928MB to 872M.
+
+diff --git a/src/main/java/ca/spottedleaf/moonrise/patches/block_counting/BlockCountingChunkSection.java b/src/main/java/ca/spottedleaf/moonrise/patches/block_counting/BlockCountingChunkSection.java
+index a08ddb0598d44368af5b6bace971ee31edf9919e..277c7d8bd04eb83dcba7fbe0b87660ea338ed653 100644
+--- a/src/main/java/ca/spottedleaf/moonrise/patches/block_counting/BlockCountingChunkSection.java
++++ b/src/main/java/ca/spottedleaf/moonrise/patches/block_counting/BlockCountingChunkSection.java
+@@ -4,6 +4,8 @@ import ca.spottedleaf.moonrise.common.list.IBlockDataList;
+ 
+ public interface BlockCountingChunkSection {
+ 
++    public boolean anyTickingBlocks();
++
+     public int moonrise$getSpecialCollidingBlocks();
+ 
+     public IBlockDataList moonrise$getTickingBlockList();
+diff --git a/src/main/java/net/minecraft/server/level/ServerLevel.java b/src/main/java/net/minecraft/server/level/ServerLevel.java
+index f9abf63e12ea930275121b470e4e4906cff0fc12..010c2969bf5fa5bb7281acb3cdb44f3193e086e7 100644
+--- a/src/main/java/net/minecraft/server/level/ServerLevel.java
++++ b/src/main/java/net/minecraft/server/level/ServerLevel.java
+@@ -817,6 +817,10 @@ public class ServerLevel extends Level implements WorldGenLevel, ca.spottedleaf.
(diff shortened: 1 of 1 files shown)


patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
```diff
@@ -0,0 +1,95 @@
...
+     public final ca.spottedleaf.moonrise.common.list.IBlockDataList moonrise$getTickingBlockList() {
++        this.createTickingBlocks();
+         return this.tickingBlocks;
+     }
+     // Paper end - block counting
+@@ -126,9 +136,11 @@ public class LevelChunkSection implements ca.spottedleaf.moonrise.patches.block_
+         }
+ 
+         if (iblockdata1.isRandomlyTicking()) {
++            this.createTickingBlocks();
```
Brokkonaut, line 71: why create list when just trying to remove an entry?

patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
```diff
@@ -0,0 +1,95 @@
...
++        if (tickingBlocks == null) tickingBlocks = new ca.spottedleaf.moonrise.common.list.IBlockDataList();
++    }
+ 
+     @Override
+     public final int moonrise$getSpecialCollidingBlocks() {
+@@ -44,6 +53,7 @@ public class LevelChunkSection implements ca.spottedleaf.moonrise.patches.block_
+ 
+     @Override
+     public final ca.spottedleaf.moonrise.common.list.IBlockDataList moonrise$getTickingBlockList() {
++        this.createTickingBlocks();
```
Brokkonaut, line 63: why create that list here? why not check for null at the caller?

patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
xymb-endcrystalme: Because if `this.tickingBlocks` is null, a call to `this.tickingBlocks.remove()` would crash immediately.

In theory there shouldn't really be any calls to `remove` when there are no tickingBlocks... But I decided to create the structure, just to not modify the original behavior.

It could be changed to:

```
if (this.tickingBlocks != null) {
    this.tickingBlocks.add(x, y, z, state);
}
```
but I decided just creating it is simpler (1 loc vs 2 loc).

patches/server/1068-Memory-optimization-Lazy-initialization-of-IBlockDat.patch
xymb-endcrystalme: Because I wanted to leave the functionality intact.

The function expects to return a `IBlockDataList`, so I made it return `IBlockDataList`.
"Sometimes" returning `null` here would be bad practice and could introduce unknown bugs if someone ever uses it from some unexpected place.

In practice the only call to `getTickingBlockList()` never executes because `if(anyTickingBlocks() == false) continue;` is executed before it.

lynxplay:
After some chatter with the team, I are not going to pull a PR that modifies moonrise-only logic.
Patching over moonrise in this matter means additional upkeep during moonrise updates, which are already a pain to begin with.

We'll expand the contribution documentation to make this clear, but this diff belongs into moonrise.
If you do not feel comfortable with contributing it to moonrise yourself due to the mixins usage over patches, I can offer to pick up that work and open the respective PR on the project some time next week, with you as the author :+1: 

Obviously does not mean this change is something 100% accepted into moonrise, but yea.
Hope you understand my reasons for closing this and thank you for the generally smart contribution, even if it was the wrong project :sweat_smile:

xymb-endcrystalme:
> If you do not feel comfortable with contributing it to moonrise yourself due to the mixins usage over patches, I can offer to pick up that work and open the respective PR on the project some time next week, with you as the author 👍

It makes perfect sense to go `Moonrise` -> `Paper` in this case.

Unfortunately learning Forge and Mixins and yet another software is something I don't really have time for. Perfectly happy for someone else to implement it - it's a very simple patch anyway. ❤️

xymb-endcrystalme:
I see that nothing moved in the last half a year, every `LevelChunkSection` is still creating empty `ShortList` on every section  in newest Paper. So Moonrise decided to not implement it on their own end?

Maybe it would be ok to apply a `1.21.4` version of this patch? :D
NGL, I don't think that users care if it's a "Moonrise" patch or "Paper" patch.

masmc05:
> So Moonrise decided to not implement it on their own end?

I don't see anyone making a pr with this change to moonrise for them to implement this

electronicboy:
Its not a user concern, its a “this change will get lost on updates due to how moonrise is updated” concern

xymb-endcrystalme:
Paper is not a patchset on top of Moonrise? I always assumed it is.

Well, I don't use Paper/Folia directly, but a fork, so it doesn't affect me if it ever gets upstream.

electronicboy:
Paper is a patch set over vanilla, and moonrise is one is one of the
feature patches which is basically copied over more or less manually, we
generally do not want patches over feature patches because they become a
headache to track

PR #11503, opened 2024-10-20
Update 1032-Bundle-spark.patch

NoltoxGit:
Update spark-paper to **v1.10.110**.


lynxplay:
Thanks for the PR but we reserve dependency updates (especially ones as non-impactful as this one) to the team.
Additionally, the version does not seem to be deployed rn.

Issue #11026, opened 2024-07-06
TrialSpawnerConfiguration does not play nicely with live blockstates


lynxplay:
Live blockstates do not need to be applied to have their changes reflected in game.
The current TrialSpawnerConfiguration however is a fully copy and does only apply when updating.

To fix this, a new implemented of the interface is needed that re-creates the internal record on every setter call and is used if the block state is not a snapshot.

Strokkur424:
I'd love to work on this if it hasn't been started by somebody else yet!

lynxplay:
Go for it 👍

Strokkur424:
Update: I am still having technical issues with my Windows + WSL setup. If anybody is faster than me, feel free to take this one instead of me

NithinSravan:
Hi @lynxplay , I am new around here and would like to have a look into this issue if it is still open. It would be great if some more detail is provided around this issue. Thanks!

Machine-Maker:
@NithinSravan Paper adds non-snapshot BlockState (or really TileState)s. Changes made can then avoid unneeded copies improving performance. But the TileSpawnerConfiguration implementation doesn’t respect this.

tlm9201:
Does there need to be an entire new implementation of `TrialSpawnerConfiguration`? Or should the existing `CraftTrialSpawnerConfiguration` be modified and check `isSnapshot` on every setter call instead?

jha9262:
Hi @lynxplay,
I’d like to work on this issue if it’s still available.
I’ve gone through the description and would be happy to start working on a fix/implementation once you confirm.
Thanks!

PR #11515, opened 2024-10-25
Live trial spawner configurations

tlm9201:
#11026


tlm9201:
I hope I understood the issue properly.

PR #11524, opened 2024-10-27
Allow using old ender pearl behavior & apply ender pearl exploit patch

jpenilla:
When enabled, ender pearls will not load chunks and will save to the world instead of the player.

Also changes the exploit config to be default false, as it only makes sense when legacy behavior is enabled.
I didn't bother with a config migration since it won't do anything without legacy behavior enabled.

### DIFF
--- a/patches/server/0005-Paper-config-files.patch
+++ b/patches/server/0005-Paper-config-files.patch
@@ -1423,10 +1423,10 @@ index 0000000000000000000000000000000000000000..279b24c689b9979884b65df7eb1f0590
 +}
 diff --git a/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..a81a332ffb80e67d7f886295099b5cd2ae8994c5
+index 0000000000000000000000000000000000000000..dad3fcc689ec806f985122a7cbd501a7d0fd0d36
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/configuration/WorldConfiguration.java
-@@ -0,0 +1,580 @@
+@@ -0,0 +1,581 @@
 +package io.papermc.paper.configuration;
 +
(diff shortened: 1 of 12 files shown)


kennytv (approved):
Someone else should test it:tm:, but the changes look fine. Bonus points for a telling config name, but I also can't think of one, so that's not a blocker

Leguan16:
> Bonus points for a telling config name

Config name for what? `legacyEnderPearlBehaviour`? `disableEnderpearlLoadChunks` or something like that (I think that's the new behaviour or did I miss something?).

jpenilla:
I chose `legacyEnderPearlBehaviour` rather than `enderPearlsTickChunks` as it encompasses more than that, it also affects how pearls are saved and whether they get removed on logout, for example. I think the current name is fine and it should be addressed by docs.

Owen1212055:
Maybe we could have seperate config options for controlling if they are serialized with the player and tick chunks then?

jpenilla:
doing that properly would require much more extensive changes, the behaviors are tightly coupled

Leguan16:
> I chose `legacyEnderPearlBehaviour` rather than `enderPearlsTickChunks` as it encompasses more than that, it also affects how pearls are saved and whether they get removed on logout, for example. I think the current name is fine and it should be addressed by docs.

Ok yeah then i don't have another name either.

ryantheleach:
Naming disablePlayerTickedEnderpearls ? encompasses the fact that enderpearls are loaded and ticked, and player owned by default.

codebycam:
`legacyEnderPearlBehaviour` is fine IMO. It indicates there has been a change, and people can find more info about it on the docs.

ryantheleach:
> `legacyEnderPearlBehavior` is fine IMO. It indicates there has been a change, and people can find more info about it on the docs.

As long as enderpearls are never changed again, otherwise you might need a legacyLegacyEnderPearlBehavior.

Personally, anything named 'legacy' or 'old' is a code smell in any coding project, unless it clarifies *which* legacy behaviour it is almost immediately.

Issue #11511, opened 2024-10-22
1.21.2 / 1.21.3 Tracking issue


lynxplay:
Priority tasks (new, the other list will also have some of those):
- [x] Verify per player mob spawning behavior, especially this change (either it was broken for multiple versions or it is broken now): https://github.com/PaperMC/Paper/commit/3cb16c9d91eb133a9d95605bdb8afc6ed40d6ee9#diff-950deb69c365ce9e698f6f455f174056fc72d2958391fa9bd1ae49e7e029124fL155
- [x] Fix TeleportFlags mapping
- [x] https://github.com/PaperMC/Paper/pull/10675 / https://github.com/PaperMC/Paper/pull/11631
- [x] spark async-profiler update
- [x] https://github.com/PaperMC/Paper/pull/11606
- [x] https://github.com/PaperMC/Paper/pull/11599
---

- [x] ~Check on "Fix entity teleportation and cancel velocity if teleported" since new tp param has delta movement.~ (patch got removed)
- [x] PlayerItemCooldownEvent adoption to cooldown groups (it exposes Material) https://github.com/PaperMC/Paper/pull/11625
- [x] Upstreams (and ours) ItemDamageEvent called for 0 damage https://github.com/PaperMC/Paper/pull/11900
- [x] Check in on PlayerTeleportEndGatewayEvent (might need dimension == targetDimension)
- [x] Fix tripwire disarming not working as intended - do we want this to be configurable still? also this diff is different than ours
- [x] https://github.com/PaperMC/Paper/pull/11605
- [x] https://github.com/PaperMC/Paper/pull/11532
- [x] MCUTILS: look at chunkmap changes around line 555. I removed our changes because I wasn't sure what changed, needs to be looked at!
- [x] timings v2: random line somewhere in serverchunkcache near tick. doesn't matter mutch, entirety of timings needs to be removed
- [x] Check in on EntityZapEvent implementation, heavy overload usage in Mob#convertTo
- [x] Check in on EndermanAttackPlayerEvent - the event may be called in the freeze goal check, sounds unsmart.
- [ ] The entirety of PlayerLaunchProjectileEvent needs review (specifically trident pickup status) see 7c04a31c9. Should work, but check
- [ ] Check on net.minecraft.world.level.material.FlowingFluid#getSpread no longer using a spread cache for initial block state lookup
- [x] Check removed config options.
- [x] w.i.ench.effects.DamageItem > ChangeItemDamage
[...]

PR #11536, opened 2024-10-30
Add test for org.bukkit.craftbukkit.entity.CraftMinecart#minecartEntityTypeToMaterial

Abelkrijgtalles:
Hi, I hope you're having a wonderful day.
I've added a test for org.bukkit.craftbukkit.entity.CraftMinecart#minecartEntityTypeToMaterial, as requested in https://github.com/PaperMC/Paper/issues/11511.
In the current approach I've hardcoded the Minecart types, as is done in the function.


lynxplay:
The whole point of the test is to ensure that the hardcoded values in the function are still correct.
Testing this with *another* set of hardcoded function is useless.
Iterate the item registry, find MinecartItems, reflectively access their entity type, ensure the static method correctly matches that.

Abelkrijgtalles:
Alright, I will try to fix it to use the Item registry

lynxplay (changes requested):
Also merge this into the respective patch that introduces the method under test.
No need for a new patch here.

Abelkrijgtalles:
I've fixed it to not be hard-coded, except for `EntityType.SPAWNER_MINECART`, as that doesn't have a corresponding item, and the function just returns a normal minecart. I've used a string based system, as I couldn't find any other way to do it. The current approach gets the short string of the entitytype (like `chest_minecart`), and gets the item with the corresponding  `ResourceLocation`

lynxplay:
Thanks!
You'll need to annotate the class with `@AllFeatures` in order for the test suite to pick it up and run correctly.

Abelkrijgtalles:
I'll probably also make a PR for changing it to a switch statement, as that's a better choice then 7 if/else statements.

Abelkrijgtalles:
> Thanks! You'll need to annotate the class with `@AllFeatures` in order for the test suite to pick it up and run correctly.

Will do

Abelkrijgtalles:
I've added it, so it should be good to go.

patches/server/0448-API-to-get-Material-from-Boats-and-Minecarts.patch
```diff
@@ -61,3 +61,53 @@ index ee010d53f8c671d17d68f3f43dca9978e23ac8ab..d35c1a10e58932b19c8053c5dacdc25f
...
++
++        return arguments.stream();
++
++    }
++
++    @AllFeatures
++    @ParameterizedTest
++    @MethodSource("minecartTypesAsArguments")
++    void minecartEntityTypeToMaterial(EntityType type, Item material) {
++
```
Leguan16, line 108: I know it's a massive nit but couldn't the diff be reduced by a bit by removing the empty lines? 
(Edit: I meant inside the method body not all lines, I thought I wouldn't have to specify)

masmc05: Is the diff size important in new files? I thought it was important purely to prevent conflicts in future merges, which is unrelated in context of new files

Edit: yeah, removing the whitespaces indeed worsened the readability of the code, and I'm sure making it a nightmare to work with in future. Not worth at all the decreased patch size

Abelkrijgtalles: I'll remove it just in case

Abelkrijgtalles:
I've removed the empty lines. If it's to aggressive, just let me know.

Abelkrijgtalles:
Alright.

Abelkrijgtalles:
Hi, I was just wondering, are there any specific things holding the merge back?

lynxplay:
The fact that no developer had time to review this yet.
Give it time, someone will get it merged soon enough.

Abelkrijgtalles:
Alright, thanks for the update

lynxplay:
Can you reopen the PR to master please :)
I'll try to review it tonight and get that merged.

Abelkrijgtalles:
Alright, will do

PR #11537, opened 2024-10-31
Improve Random Block Selection RNG

ScriptLineStudios:
The RNG used in selecting random blocks for updates, is fundamentally flawed, plotting some values sampled from this RNG clearly indicates the problem:

weak

Modifying the algorithm to include a couple more terms yields a much nicer and random looking result:

strong

Additionally this PR removes the deprecated threadSafeRandom object used for providing the client with seeds for sound variant selection in favor of using the improved randValue :D

### DIFF
--- a/patches/server/1069-improve-level-rng.patch
+++ b/patches/server/1069-improve-level-rng.patch
@@ -1,0 +1,66 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: scriptlinestudios <scriptlinestudios@protonmail.com>
+Date: Thu, 31 Oct 2024 13:37:55 +0200
+Subject: [PATCH] improve level rng
+
+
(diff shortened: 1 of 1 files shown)


lynxplay:
Thank you for the PR.

We are not going to deviate from vanilla's logic at the whim that it "looks more random".

vadcx:
_I remember some fork a while ago attempted to optimize block update RNG at some point._

@ScriptLineStudios  If you can prove that this RNG pattern causes problems, you would need to raise issues with Mojang directly? Otherwise it must have been a deliberate choice, the same magic number is used here as an increment:

- https://en.wikipedia.org/wiki/Linear_congruential_generator#Parameters_in_common_use 1013904223
- and here's fun a reading on LFSR https://jacopretorius.net/2017/09/wolfeinstein-3d-fizzlefade-algorithm.html

PR #11555, opened 2024-11-01
Fix ItemDamageEvents called for 0 damage

notTamion:
#11511 
post hardfork #11900


Lulu13022002:
This break negative damage amount that is possible with enchantment component. I'm not sure if the event/api should run in this case too.

notTamion:
For later:
1. do we want the event to be called in case of negative damage?
2. should we allow setDamage and other methods which damage the itemstack to deal negative damage? as this might break some plugins that are relying on those methods not being able to deal negative damage

lynxplay:
Changed this for the following logic:

Neither damage event will be called for non-positive damage.
This 100% preserves old behaviour.
Negative or zero damage is still processed, however no events will be emitted, as this isn't a damage, this would be "repairing" the item.

How and if such a usecase should be represented via the API is up for later discussion, but should not be a blocked for ensuring the existing API maintains backwards compatibility.

notTamion:
Yeah that sounds good

PR #11556, opened 2024-11-01
Make CommandSourceStack respect hidden players

### LINKED ISSUE #11545: Players aren't hidden in the /team command
### Expected behavior

If a player is hidden using the https://jd.papermc.io/paper/1.21.3/org/bukkit/entity/Player.html#hidePlayer(org.bukkit.plugin.Plugin,org.bukkit.entity.Player) method does hide players in all commands

### Observed/Actual behavior

If a player is hidden using the https://jd.papermc.io/paper/1.21.3/org/bukkit/entity/Player.html#hidePlayer(org.bukkit.plugin.Plugin,org.bukkit.entity.Player) method does not hide players in all commands like in the /team join or /team leave command

### Steps/models to reproduce

https://github.com/user-attachments/assets/0a158cc5-84a3-499c-8653-34246f598e14

### Plugin and Datapack List

Only this plugin coded by myself https://github.com/NikeyV1/NikeySystem.git

### Paper version

This server is running Paper version 1.21.1-99-master@1bc02e6 (2024-09-25T02:41:02Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are 32 version(s) behind

### Other

Im trying to make a system plugin and am trying to hide a player completely and you can still see player using this command

notTamion:
fixes #11545 
post hardfork #11898


electronicboy (approved):

lynxplay:
We have this diff already https://github.com/PaperMC/Paper/blob/85c870e9c7c97ff9f698dc1137586ee4cd54dda7/patches/server/0988-Fix-entity-type-tags-suggestions-in-selectors.patch#L80-L92

Either we remove this or duplicate it.
Idk what else calls this method and if that is of interest to maybe not solve as deep.

notTamion:
its only called by argument types so i think fixing this that deep is the correct solution.
image

lynxplay:
Okay, thanks for the quick check :+1:

Machine-Maker (approved):

lynxplay (approved):

lynxplay:
While this works, we should probably at least discuss if we *want* certain commands to obey this visibility.
Banning and deop both sounds like commands where I'd never like this command to *not* show me the members?

notTamion:
Perhaps we can add an exclusion for operators?

Leguan16:
I think excluding operators makes sense.

notTamion:
just an idea, if anyone has a solution they find better please leave a comment

Machine-Maker:
I don't like making more functionality dependant on ops. The concept of "op" is not a good thing to really encourage more. People should be using proper permissions. I thought about maybe a new permission that could be granted to bypass all visibility hiding, but I'm not sure about that either.

Leguan16:
I don't know. The paper command for example also works for ops and mixing op and permission requirements sucks.

notTamion:
i agree with machine, considering servers will most likely not give staff operator but instead give them the permissions they need i think it would be smarter to add a permission for this. i don't think comparing this to the paper command is fair as the paper command is more of a debug command rather than something a staff member would actually have to use frequently

Machine-Maker:
Ok, I think we want a permission, but *just* for command suggestions, `paper.bypass-visibility.tab-completion`

Machine-Maker (approved):

PR #11560, opened 2024-11-01
configure mockito agent

yannicklamprecht:
Addresses issues with newer Java versions and Mockito.

- https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
- https://openjdk.org/jeps/451

### DIFF
--- a/patches/api/0003-Test-changes.patch
+++ b/patches/api/0003-Test-changes.patch
@@ -7,16 +7,52 @@ Subject: [PATCH] Test changes
 - Ignore package-private methods for nullability annotations
 - Add excludes for classes which don't pass
 - Disable stupid BukkitMirrorTest
+- configure mockito agent to address changes in newer java versions see https://openjdk.org/jeps/451
 
 Co-authored-by: Riley Park <rileysebastianpark@gmail.com>
 Co-authored-by: Jake Potrebic <jake.m.potrebic@gmail.com>
+Co-authored-by: Yannick Lamprecht <yannicklamprecht@live.de>
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 0b837b485bec96fa37ed65c18df97e55cecd0e9d..c8a8903d1b0c9822743549ecb8e4fdc7d0fd07c1 100644
+index 0b837b485bec96fa37ed65c18df97e55cecd0e9d..c7f660d52e1b5f085919a8aebf8476918705a391 100644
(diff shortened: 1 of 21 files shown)


electronicboy (approved):
maybe somebody more in-tune with gradle will flag this, but, this LGTM

jpenilla (changes requested):
This should use a CommandLineArgumentProvider to avoid resolving dependencies during configuration time - I think there are some other spots that may need to change in this direction too, for example stuff used by Javadoc, but it would be better to not increase the tech debt here.

yannicklamprecht:
> This should use a CommandLineArgumentProvider to avoid resolving dependencies during configuration time - I think there are some other spots that may need to change in this direction too, for example stuff used by Javadoc, but it would be better to not increase the tech debt here.

Hopefully addressed that to your liking.

patches/api/0003-Test-changes.patch
```diff
@@ -7,27 +7,69 @@ Subject: [PATCH] Test changes
...
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 0b837b485bec96fa37ed65c18df97e55cecd0e9d..c8a8903d1b0c9822743549ecb8e4fdc7d0fd07c1 100644
+index 0b837b485bec96fa37ed65c18df97e55cecd0e9d..ffa4851c98d30a2ac7d3c5bfa52a1033c682db0d 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
-@@ -107,6 +107,12 @@ tasks.test {
-     useJUnitPlatform()
+@@ -1,3 +1,5 @@
++import org.gradle.kotlin.dsl.newInstance // Paper - configure mockito agent that is needed in newer java versions
```
jpenilla: I don't think this should be needed from a kts build script

jpenilla (changes requested):

patches/api/0003-Test-changes.patch
jpenilla: Also, you don't need to annotate with `InputFiles` on properties annotated with `CompileClasspath` - our build script won't run with Gradle versions old enough to require both.

jpenilla (approved):
Looks fine, but one thing to consider is if Spigot configures the equivalent in Maven then this should move to the setup gradle build patch.

Machine-Maker (approved):

PR #11563, opened 2024-11-02
Fix incorrect CommandSourceStack creation for players

### LINKED ISSUE #11562: [1.21.3] Player Perform Command API Failing
### Expected behavior

I expected my command to be executed, but instead, there is no output in the console, and it seems that the command has not been executed. I used this snippet of code inside my plugin class to test the command execution when a player joins. It worked on Paper 1.21.1

```java
    Bukkit.getPluginManager().registerEvents(new Listener() {
      @EventHandler
      public void onPlayerJoin(final PlayerJoinEvent event) {
        player.performCommand("say Hello World!");
      }
    }, this);
```

(As for a different but similar issue, I used Adventure components with text that runs a command when the player presses on it, and there is nothing that happens either. I'm not sure if this is the same issue or not)

### Observed/Actual behavior

There is nothing sent out into console, with no output sent out to the player. No error is displayed into the console either, and the implementation fails silently.

### Steps/models to reproduce

1) Register a quick `Listener` for Player join for executing a `say` command message
2) Join the server as any player
3) There is nothing displayed inside of chat

### Plugin and Datapack List

None besides the current plugin

### Paper version

```
[19:47:47 INFO]: Checking version, please wait...
[19:47:47 INFO]: This server is running Paper version 1.21.3-10-master@92131ad (2024-11-01T21:30:00Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
Previous version: 1.21.3-8-85c870e (MC: 1.21.3)
```

### Other

_No response_

notTamion:
fixes #11562

### DIFF
--- a/patches/server/1060-Fix-incorrect-CommandSourceStack-creation-for-player.patch
+++ b/patches/server/1060-Fix-incorrect-CommandSourceStack-creation-for-player.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tamion <70228790+notTamion@users.noreply.github.com>
+Date: Sat, 2 Nov 2024 10:34:58 +0100
+Subject: [PATCH] Fix incorrect CommandSourceStack creation for players
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/command/VanillaCommandWrapper.java b/src/main/java/org/bukkit/craftbukkit/command/VanillaCommandWrapper.java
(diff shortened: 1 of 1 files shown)


Machine-Maker:
On a slightly related note... The API still allows dispatching Commands from an Entity. These commands will be executed with the NULL CommandSource, this is kind of a change? Do we care?

Machine-Maker (approved):

kennytv (approved):

electronicboy (approved):

notTamion:
Fixed by upstream

PR #11570, opened 2024-11-03
Fix below-zero-generation-in-existing-chunks

### LINKED ISSUE #11531: Disabling chunk generation below Y=0 does not work
### Expected behavior

Chunk generation below Y=0 to be disabled in existing chunks when a 1.12.2 world is migrated to 1.21.1 when setting below-zero-generation-in-existing-chunks: false in spigot.yml.

### Observed/Actual behavior

World migrates successfully, but generates chunks below Y=0 when players move around

### Steps/models to reproduce

1. Start new paper-1.21.1-128 server in empty directory
2. Accept EULA
3. Restart server to generate default configs
4. Stop server
5. Change below-zero-generation-in-existing-chunks to false in spigot.yml
6. Delete world/
7. Copy 1.12.2 world into directory to replace world/
8. Start server
9. Join server and fly below Y=0 in spectator mode

### Plugin and Datapack List

/plugins: None
/datapack list: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]

### Paper version

This server is running Paper version 1.21.1-128-master@d348cb8 (2024-10-21T16:23:24Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: git-Paper-550 (MC: 1.19.4)

### Other

_No response_

notTamion:
fixes #11531 
i am not entirely sure what exactly is happening here. my best guess is that the chunk status "full" didn't yet exist in older versions and fully generated chunks were just left on "mobs_spawned".

### DIFF
--- a/patches/server/1060-Fix-belowZeroGenerationInExistingChunks.patch
+++ b/patches/server/1060-Fix-belowZeroGenerationInExistingChunks.patch
@@ -1,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tamion <70228790+notTamion@users.noreply.github.com>
+Date: Sun, 3 Nov 2024 11:26:10 +0100
+Subject: [PATCH] Fix belowZeroGenerationInExistingChunks
+
+
+diff --git a/src/main/java/net/minecraft/world/level/chunk/storage/ChunkStorage.java b/src/main/java/net/minecraft/world/level/chunk/storage/ChunkStorage.java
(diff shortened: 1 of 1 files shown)


kennytv:
How exactly did you get to mobs_spawned? The equivalent for full in pre 1.14-ish versions would be `fullchunk`

notTamion:
i generated a world in 1.12.2 made sure to wait a bit and also flew around in spectator mode in the ground  to make sure everything was generated. (should note that i used single player to generate this) i also tried the same in 1.17 and when i checked there the chunks were showing up as full

kennytv:
Pretty sure if you if you did the same with a 1.14 world it would break again, it's kind of awkward to have the check before datafixing

PR #11573, opened 2024-11-03
Expose particle status client option

Lulu13022002:


patches/api/0183-Add-Player-Client-Options-API.patch
```diff
@@ -60,6 +61,24 @@ index 0000000000000000000000000000000000000000..ed08f823e0620289392f7fc2ff0ac721
 +            return "options.chat.visibility." + this.name;
 +        }
 +    }
++
++    public enum ParticleStatus implements Translatable {
```
MiniDigger: is ParticleStatus a good name for this? maybe ParticleVisibility?

Lulu13022002: Yes ParticleVisibility is more explicit, maybe ParticleAmount too idk.
Do you remember when ChatVisibility is unknown? the field doesn't seems to be nullable and is init in the constructor

MiniDigger: cant look at code rn, but theres a time period where the client options arent known to the server yet, right?

Lulu13022002: Yes but in that case the player just have the default settings value. So we can't really say when the option is unknown. For example in the PlayerLoginEvent chat visibility is always full.

Machine-Maker (approved):

lynxplay (approved):

PR #11574, opened 2024-11-04
Add player weather support to Entity#isInRain

bridgelol:
Fixes trident behaviour when using player weather. Currently riptide does not register serverside when using player-weather from plugins like Essentials (https://github.com/EssentialsX/Essentials/blob/f7a8f86850c9c4f9dee9b6120e6da151cbe81ed0/Essentials/src/main/java/com/earth2me/essentials/commands/Commandpweather.java#L19).


electronicboy:
This does not consider if the biome that a player is standing in actually rains, etc

bridgelol:
> This does not consider if the biome that a player is standing in actually rains, etc
Ah, you're right. Will push something tonight

lynxplay:
This is a lot more involved of a PR I think.
From the current looks, the entire per-player weather seems to be "visual only".
The weather value is used for nothing but computing the packets.

This change now makes it a complete behaviour change.
The player would now be cleared of fire and well, can jump tridents.
I am unsure if this is what this method was intended for.

PR #11580, opened 2024-11-04
Manually reset attribute map on restoring player instance

electronicboy:
Spigots reuse of the entity instances causes the server to try to copy a players existing attributes back onto itself, which causes an error.

We'll manually clear the applicable attribute modifiers, but, in the long run, we should just revert out of the broken behavior of trying to reuse ServerPlayer instances.

---
Download the paperclip jar for this pull request: [paper-11580.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2175618506.zip)

### DIFF
--- a/patches/server/1060-Manually-reset-attribute-map-on-restoring-player-ins.patch
+++ b/patches/server/1060-Manually-reset-attribute-map-on-restoring-player-ins.patch
@@ -1,0 +1,68 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Shane Freeder <theboyetronic@gmail.com>
+Date: Mon, 4 Nov 2024 20:23:02 +0000
+Subject: [PATCH] Manually reset attribute map on restoring player instance
+
+Spigots reuse of the entity instances causes the server to try to copy
+a players existing attributes back onto itself, which causes an error.
+
+We'll manually clear the applicable attribute modifiers, but, in the long run,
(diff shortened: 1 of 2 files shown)


patches/server/1060-Manually-reset-attribute-map-on-restoring-player-ins.patch
```diff
@@ -0,0 +1,68 @@
...
+@@ -156,5 +156,21 @@ public class AttributeMap {
+         attributes.put(attributeBase, attributeModifiable);
+     }
+     // Paper - end - living entity allow attribute registration
++    // Paper start - deal with upstream stupidity around attribute modifiers and recycling entity instances.
++    public void removeAllModifiers() {
++        this.attributes.values().forEach(AttributeInstance::removeModifiers);
++    }
++
++    public void removeAllTransientModifiers() {
```
kennytv: This method scared me for a second, should add a note that we expect the getters to return copies

kennytv (approved):

lynxplay:
Removing transient modifiers means armor needs to reapply.
That does not happen because we are not actually a new entity.

lynxplay (changes requested):
Removing transient modifiers means armor needs to reapply.
That does not happen because we are not actually a new entity.

lynxplay:
Closing this as the immediate issue has been solved by upstream.
It was solved however by breaking vanilla logic and no longer removing attribute modifiers on death.

The API needs to be fleshed out to support plugins to set attribute modifiers that do/do not persist through death.

PR #11587, opened 2024-11-05
allow players to be ridden

Gerrygames:
this was possible before 1.21.3

### DIFF
--- a/patches/server/1060-allow-players-to-be-ridden.patch
+++ b/patches/server/1060-allow-players-to-be-ridden.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gerrygames <gecam59@gmail.com>
+Date: Tue, 5 Nov 2024 15:36:34 +0100
+Subject: [PATCH] allow players to be ridden
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index a0876d3f88620bb24ef69101fc67b0dcd5dca0d2..7040a0217ba862924fee15a5d4c2cf3375868baa 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -3226,7 +3226,7 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+             return false;
+         } else if (!entity.couldAcceptPassenger()) {
+             return false;
+-        } else if (!this.level().isClientSide() && !entity.type.canSerialize()) {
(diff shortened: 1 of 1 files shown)


Floppy012:
While this is somewhat a breaking change introduced by Mojang. Riding a player has the side-effect of the entity riding no longer being persisted (and thus, being lost).

I'd prefer to have a `force` parameter (the `startRiding` already has one iirc) that then can be documented to tell devs that there may be side-effects when setting this to true. However, I'm not sure atm what the current force parameter does and if it could be used. And a second force parameter would be confusing I guess.

Nevertheless, simply allowing the player to be ridden reintroduces a "bug" they fixed. So another way could be a config flag? Probably the most backwards-compatible and server-admin friendly way to allow this.

patches/server/1060-allow-players-to-be-ridden.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index a0876d3f88620bb24ef69101fc67b0dcd5dca0d2..7040a0217ba862924fee15a5d4c2cf3375868baa 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -3226,7 +3226,7 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+             return false;
+         } else if (!entity.couldAcceptPassenger()) {
+             return false;
+-        } else if (!this.level().isClientSide() && !entity.type.canSerialize()) {
++        } else if (!this.level().isClientSide() && entity.type != EntityType.PLAYER && !entity.type.canSerialize()) { // Paper - Allow players to be ridden
```
yannicklamprecht, line 16:
```suggestion
+        } else if (!this.level().isClientSide() && !entity.type.canSerialize() && !force) { // Paper - Allow players to be ridden
```

yannicklamprecht:
> While this is somewhat a breaking change introduced by Mojang. Riding a player (or any other entity that won't be serialized) has the side-effect of the entity riding no longer being persisted (and thus, being lost).
> 
> I'd prefer to have a `force` parameter (the `startRiding` already has one iirc) that then can be documented to tell devs that there may be side-effects when setting this to true. However, I'm not sure atm what the current force parameter does and if it could be used. And a second force parameter would be confusing I guess.
> 
> Nevertheless, simply allowing the player to be ridden reintroduces a "bug" they fixed. So another way could be a config flag? Probably the most backwards-compatible and server-admin friendly way to allow this.


The current force parameter allows overriding the decision if
1. the entity cannot ride the other one
2. e.g. there are limitation how many passenger the "vehicle" can have. e.g. Camel => 2

Gerrygames:
Fixed by upstream

PR #11590, opened 2024-11-06
Fix Entity#isTicking and update Paper entity command

### LINKED ISSUE #10299: Entity#isTicking not acting as it should
### Expected behavior

Entity#isTicking should return whether or not an Entity is ticking

### Observed/Actual behavior

It appears to return whether or not the chunk the entity is in is ticking.

### Steps/models to reproduce

I don't have proper steps to show this, because I've been testing via Skript and an addon for Skript I have created (written in Java using this API)

I want to show some visuals:
Ok, in this one the villager is standing in the Yellow Chunk.
Green = Chunk.LoadLevel#ENTITY_TICKING = All game logic is processed.
Yellow = Chunk.LoadLevel#TICKING = All game logic except entities is processed.
Red = Chunk.LoadLevel#BORDER = Most game logic is not processed, including entities and redstone.

The entity is not ticking, yet Entity#isTicking is returning true
(this can be seen in the scoreboard, where I'm using (ticking/total) for each type of entity.
You can see here there are 20 ticking villagers out of 20 total villagers, when in fact they're not actually ticking)
Screenshot 2024-03-04 at 6 37 55 PM

If I move back a chunk, the villagers are in the RED zone, which they're definitely not ticking.
(This part is working as expected)
Screenshot 2024-03-04 at 6 40 38 PM

### Plugin and Datapack List

[18:41:47 INFO]: Server Plugins (9):
[18:41:47 INFO]: Bukkit Plugins:
[18:41:47 INFO]:  - BeeConomy, PermissionsEx, RandomTeleport, SkBee, SkBriggy, Skript, skript-reflect, spark, Vault
> 

### Paper version

This server is running Paper version git-Paper-446 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: 6ad63fb)

### Other

I noticed while peeking in Paper code, in the NMS Entity Class:
```java
public boolean isTicking() {
    return ((net.minecraft.server.level.ServerChunkCache) level.getChunkSource()).isPositionTicking(this);
}
```
It appears this method is checking if the chunk is ticking, not specifically the entity

I think a better option would be to check if the chunk load level at said entity is actually ENTITY_TICKING instead of TICKING (where entities are in fact not ticked)

jpenilla:
fixes #10299

### DIFF
--- a/patches/server/0009-MC-Utils.patch
+++ b/patches/server/0009-MC-Utils.patch
@@ -5344,7 +5344,7 @@ index 019228b1809e3816b0b4dfb9f19b8d42876cc240..6c2339d6a93172e25040c4868a3a4747
  
          while (objectiterator.hasNext()) {
 diff --git a/src/main/java/net/minecraft/server/level/ServerChunkCache.java b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
-index 9cdcab885a915990a679f3fc9ae6885f7d125bfd..c615510f3f59292715bcff1bd9e4e896c9733436 100644
+index 9cdcab885a915990a679f3fc9ae6885f7d125bfd..a970ceab0176d97f9d274ed257c2f86f3b63e430 100644
 --- a/src/main/java/net/minecraft/server/level/ServerChunkCache.java
 +++ b/src/main/java/net/minecraft/server/level/ServerChunkCache.java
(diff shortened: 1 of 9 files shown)


kennytv (approved):

electronicboy (approved):
Forgot I'd look at this this morning

PR #11595, opened 2024-11-08
Optimize Player#canSee

MrKinau:
Replaces the `invertedVisibilityEntities` HashMap with an `OpenHashMap` for better performance in `Player#canSee`.

### DIFF
--- a/patches/server/1060-Optimize-Player-canSee.patch
+++ b/patches/server/1060-Optimize-Player-canSee.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: David Luedtke <david@ldtke.de>
+Date: Fri, 8 Nov 2024 20:23:41 +0100
+Subject: [PATCH] Optimize Player#canSee
+
+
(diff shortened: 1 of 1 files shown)


electronicboy:
These collections are not designed for performance, they're designed for reducing memory overheads; I think that in these high read low write situations they tend to perform better, however

Malfrador:
canSee performance can be a limitation if you have a lot of otherwise cheap entities (e.g. displays), so generally making improvements here seems sensible to me.

However some benchmarks would be interesting. Did you do any? 
The OpenHashMap won't be automatically faster in all cases and just blindly replacing it without proper benchmarking isn't great.

MrKinau:
> However some benchmarks would be interesting. Did you do any?

I did see some non-optimal spark reports with `canSee` using quite some time of tick and tried to improve it. There may be more room for improvement, but just this small change seems to be a good addition to reduce the time it needs for the `containsKey`.
I do not feel comfortable sharing the spark reports publicly, but I'll attach some images. Those spark reports may not be 100% comparable, but they are taken on the same server and ig only the player count / player activity should make a difference.

<details>
  <summary>HashMap</summary>
  
</details>

<details>
  <summary>Fastutil OpenHashMap</summary>
  
</details>

kennytv:
In local benchmarks, misses are about 20% faster, but hits only half as fast once you go past a few dozen entries of hidden entities. Lynx suggested reducing the number of calls to the method from ChunkMap instead

Issue #11480, opened 2024-10-08
Reducing damage in EntityDamageByEntityEvent causes invulnerability timer to be bypassed


okx-code:
### Expected behavior

Reducing the damage in EntityDamageByEntityEvent has no side effects and just reduces the damage.

### Observed/Actual behavior

If the damage is reduced in EntityDamageByEntityEvent, the invulnerability timer for an entity is bypassed for subsequent attacks, due to a mismatch in how Paper calculates the most recent attack damage (lastHurt)

### Steps/models to reproduce

Add a listener to EntityDamageByEntityEvent to multiply the damage by 0.9.

Place a breakpoint at LivingEntity.java#1470 (at `return false`):

```java
                if (amount <= this.lastHurt) {
                    return false;
                }
```

If you spam click an entity, verify that this the breakpoint at 1470 is never hit with the listener, but is hit without.

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.21.1-119-master@7cd4f2c (2024-10-03T15:19:34Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are running the latest version
Previous version: 1.21.1-40-2fdb2e9 (MC: 1.21.1)

### Other

I'm diagnosing a combat bug caused by this if statement `if (amount <= this.lastHurt)`, what I think it does is it allows the invulnerability timer to be bypassed if a player receives an attack with more damage(?). Due to <https://github.com/PaperMC/Paper/pull/11045> this has caused a bug on our server where if you reduce the damage (`amount`) in the EntityDamageEvent, this if statement will always be false, and players will always receive damage, bypassing the invulnerability timer.

My workaround is to make sure that the amount stored in this.lastHurt is always the original amount (see screenshot), but I'm not sure if this is the best solution. Possibly a PreDamageEvent could make sense here? Any thoughts?

![Image](https://github.com/user-attachments/assets/d59fc67c-0f42-4add-8e28-6af1f6c1bcd9)

okx-code:
Related #11476

PR #11599, opened 2024-11-09
Fix incorrect invulnerability damage reduction

### LINKED ISSUE #11476: Damage ticks producing non vanilla behaviour, vanilla overlapping removed/canceled
### Expected behavior

intended behaviour:

When a player is red they can only recive damage from one source, however if a greater source of damage hits, the damage from the greater source is used in priority.

### Observed/Actual behavior

expected behaviour:

When a player is red they are fully inveunerable instead of taking damage from the greater source

### Steps/models to reproduce

https://cdn.discordapp.com/attachments/289587909051416579/1292576996920131594/Base_Profile_2024.10.06_-_20.58.11.17.mp4?ex=67043da2&is=6702ec22&hm=a83d3a7ef61d7cc62107b3c6f04c40092134b0d911fdf944a5ba67097fa597b4&

(Vanilla behaviour)

### Plugin and Datapack List

no plugins / datapacks

### Paper version

This server is running Paper version 1.21.1-119-master@7cd4f2c (2024-10-03T15:19:34Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21.1-18-63ce67d (MC: 1.21.1)

### Other

this bug has been detrimental to the vanilla PvP community and has caused a lot of servers to be unable to update further, thanks for looking at the report and i hope this can be fixed soon.

lynxplay:
Fixes incorrect spigot handling of the invulnerability damage
reduction applied when an already invulnerable entity is damaged with a
larger damage amount than the initial damage.
Vanilla still damages entities even if invulnerable if the damage to be
applied is larger than the previous damage taken. In that case, vanilla
applies the difference between the previous damage taken and the
proposed damage.

Spigot's damage modifier API takes over the computation of damage
reducing effects, however spigot invokes this handling with the initial
damage before computing the difference to the previous damage amount.
This leads to the reduction values to generally be larger than expected,
as they are computed on the not-yet-reduced value.
Spigot applies these reductions after calling the EntityDamageEvent and
*then* subtracts the previous damage point, leading to the final damage
amount being smaller than expected.

This patch cannot simply call the EntityDamageEvent with the reduced
damage, as that would lead to EntityDamageEvent#getDamage() returning
the already reduced damage, which breaks its method contract.
Instead, this patch makes use of the DamageModifier API, implementing
the last-damage-reduction as a DamageModifier.

---
Download the paperclip jar for this pull request: [paper-11599.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2206676892.zip)


patches/server/1062-Fix-invulnerability-reduction-in-EntityDamageEvent.patch
```diff
@@ -0,0 +1,93 @@
...
+                 // CraftBukkit start
+@@ -2322,8 +2322,19 @@ public abstract class LivingEntity extends Entity implements Attackable {
+     }
+ 
+     // CraftBukkit start
+-    private EntityDamageEvent handleEntityDamage(final DamageSource damagesource, float f) {
++    private EntityDamageEvent handleEntityDamage(final DamageSource damagesource, float f, final float invulnerabilityRelatedLastDamage) { // Paper - fix invulnerability reduction in EntityDamageEvent
+         float originalDamage = f;
++        // Paper start - fix invulnerability reduction in EntityDamageEvent
++        final com.google.common.base.Function<Double, Double> invulnerabilityReductionFormular = d -> {
```
mbax: formular?

Malfrador:
Just so it doesn't get lost:
I've tested this PR and it seems to fix the issue. Damage is behaving correctly again, as it does in Vanila. In my tests, changing damage in the EntityDamageEvent now also works correctly, which also fixes https://github.com/PaperMC/Paper/issues/11480.

Further testing with the build attached to the PR would be welcome though, especially if you have a plugin that changes damage in the EntityDamageEvent/EntityDamageByEntityEvent, to make sure this change doesn't cause unexpected behaviour there.

patches/server/1062-Fix-invulnerability-reduction-in-EntityDamageEvent.patch
lynxplay: Done xD

kennytv (approved):

PR #11603, opened 2024-11-10
Optimize CraftPlayer#invertedVisibilityEntities

Gerrygames:
Tested on a lobby server with 20-40 players.
Before: https://i.imgur.com/hmUYRWX.png
After: https://i.imgur.com/OJkDvpN.png
containsKey is much faster

### DIFF
--- a/patches/server/1064-Optimize-CraftPlayer-invertedVisibilityEntities.patch
+++ b/patches/server/1064-Optimize-CraftPlayer-invertedVisibilityEntities.patch
@@ -1,0 +1,91 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gero <gecam59@gmail.com>
+Date: Sun, 10 Nov 2024 13:57:36 +0100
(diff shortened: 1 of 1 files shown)


Gerrygames:
> I'm sure this breaks the moment an entity changes its entity id. UUID is persistent, entity id can change, for example when a player respawns

The entity id of a player does not change (in one server session). It is only transmitted to the client in the ClientboundLoginPacket, not in the ClientboundRespawnPacket. The entity id of entities in the world will only change if they are unloaded and then loaded again, but this calls CraftPlayer#onEntityRemove which resets the visibility and is thus irrelevant.

Gerrygames:
It looks like the entity id will change when changing dimensions. This does not apply to players but would still be a problem. Will do further analysis/testing.

Spottedleaf:
You may have better luck optimising the function by using !isEmpty && containsKey instead of just containsKey.

lynxplay:
Closing this as any larger improvements probably need some deep cuts into how the tracker handles "vanilla would track but CB doesn't" players.

PR #11605, opened 2024-11-10
Added API to check if the server is sleeping

Abelkrijgtalles:
Hi,
How are you doing?

I added the API to check if the server is sleeping/paused as requested in #11511.
This pr adds the API function `CraftServer#isPaused()`, which is easily accessible using `Bukkit#getServer()`.
Under the hood this adds a `paused`-variable (`false` by default) to `MinecraftServer`, which `isPaused()` now returns (instead of previously just `false`).
The `paused`-variable gets set to `true` when the `Server empty for x seconds, pausing`-message gets logged, and set back to `false` when the player count isn't 0 (I couldn't find a better way to detect when the server was paused).

That's all, have a nice day!


jpenilla:
See https://github.com/PaperMC/Paper/tree/is-tick-sleeping-api

The method names could change to align with the spigot apis, but the impl there is simpler

Abelkrijgtalles:
So should I just copy them? Or what should I do?

Abelkrijgtalles:
As it's not included anymore, I'll close the repo.

jpenilla:
I don't understand why this was closed, it should be ok if you copy my impl, it would just need adjusting again after your other pr is merged

Abelkrijgtalles:
> I don't understand why this was closed, it should be ok if you copy my impl, it would just need adjusting again after your other pr is merged

Alright, I'm going to do that once I get home.

Abelkrijgtalles:
I have copied the implementation and rebased.

patches/server/1065-API-to-check-if-the-server-is-sleeping.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index 64b56abf8900d0424100da460fc68ac964394793..f0f2111b5610e493baf718a8a2257f9fb8c679e7 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -1614,9 +1614,11 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+ 
+     public void onServerExit() {}
+ 
++    // Paper start - API to check if the server is sleeping
+     public boolean isPaused() {
```
jpenilla: This method is not related to the server tick sleeping, it's implemented on the integrated server for when singleplayer is paused. We should add a new method instead of reusing this one.

Abelkrijgtalles:
I have changed the method name, but the API method will stay the same.

kennytv (approved):

PR #11611, opened 2024-11-12
Add API to allow/disallow tick sleeping

Abelkrijgtalles:
Hi,
I added a simple API to `Server` which allows or disallows the server sleeping, as requested in #11511.
This works with a simple boolean in `MinecraftServer` which is now one of the requirements to sleep.
`CraftServer` sets this boolean whenever `Server#allowSleeping` gets called.
That's all, and I wish you a wonderful rest of your day!


Abelkrijgtalles:
I could also add a check which puts the server out of sleep when this gets called.

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,25 @@
...
+     boolean isOwnedByCurrentRegion(@NotNull Entity entity);
+     // Paper end - Folia region threading API
++
++    // Paper start - API to allow/disallow tick sleeping
++    /**
++     * Allow or disallow the server to sleep/pause. This only effects the next time the server is paused. If the server is currently paused, it will only go out of sleep when someone joins.
++     * If there isn't any pausing configured, the server will never pause.
++     * @param value Whether to allow sleeping of the server (defaults to true).
++     */
++    void allowSleeping(boolean value);
```
kashike:
```suggestion
+    void allowPausing(boolean value);
```

patches/server/1065-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,56 @@
...
+                     MinecraftServer.LOGGER.info("Server empty for {} seconds, pausing", this.pauseWhileEmptySeconds());
+                     this.autoSave();
+                 }
+@@ -3186,4 +3187,11 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         }
+     }
+     // Paper end - Add tick times API and /mspt command
++
++    // Paper start - API to allow/disallow tick sleeping
++    public static void allowPausing(boolean value) {
```
kashike: Does not need to be `static`.

kashike (dismissed):

Abelkrijgtalles:
I have implemented the changes. The original reason to make it static was a bit dumb, because I didn't know that `MinecraftServer#getServer` existed. The good thing is that you learn from your mistakes :)

jpenilla:
plugins could easily step on each other like this, it might be best to have this be a list of plugins blocking sleep or something to that effect instead

Abelkrijgtalles:
> plugins could easily step on each other like this, it might be best to have this be a list of plugins blocking sleep or something to that effect instead

Something like that all plugins have to allow it, but when only one plugin disallows it (and the others allow it) that pausing still gets disabled?

Abelkrijgtalles:
The new system adds a `Map<JavaPlugin, Boolean>` to `CraftServer`. When the hashmap contains the value `false` (a plugin not allowing pausing), the server will not pause.

Abelkrijgtalles:
I don't know how to fix this test error. Could I maybe get a bit of help?

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,35 @@
...
++
++    // Paper start - API to allow/disallow tick sleeping
++    /**
++     * Allow or disallow the server to sleep/pause.This only effects the next time the server is paused. If the server is currently paused, it will only go out of sleep when someone joins.
++     * When only one plugin disallows pausing, the server will never pause.
++     * If there isn't any pausing configured, the server will never pause.
++     * @param plugin The {@link JavaPlugin} that's allowing or disallowing pausing.
++     * @param value Whether to allow sleeping of the server (defaults to true).
++     */
++    void allowPausing(JavaPlugin plugin, boolean value);
```
Leguan16:
```suggestion
+    void allowPausing(@NotNull JavaPlugin plugin, boolean value);
```

Leguan16 (commented):
That should fix the test.

Abelkrijgtalles:
The tests are fixed.

jpenilla:
I don't think a map is needed - a list or set should be enough. If the collection is not empty then sleep would be blocked.

Abelkrijgtalles:
> I don't think a map is needed - a list or set should be enough. If the collection is not empty then sleep would be blocked.

I have made these changes, the only thing is that Intellij randomly imported a bunch of things, but hopefully this isn't a problem.

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,35 @@
...
++
++    // Paper start - API to allow/disallow tick sleeping
++    /**
++     * Allow or disallow the server to sleep/pause.This only effects the next time the server is paused. If the server is currently paused, it will only go out of sleep when someone joins.
++     * When only one plugin disallows pausing, the server will never pause.
++     * If there isn't any pausing configured, the server will never pause.
++     * @param plugin The {@link JavaPlugin} that's allowing or disallowing pausing.
++     * @param value Whether to allow sleeping of the server (defaults to true).
++     */
++    void allowPausing(@NotNull JavaPlugin plugin, boolean value);
```
Leguan16: This should be a FQN import

patches/server/1065-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,74 @@
...
+     // Paper start - Folia region threading API
+     private final io.papermc.paper.threadedregions.scheduler.FallbackRegionScheduler regionizedScheduler = new io.papermc.paper.threadedregions.scheduler.FallbackRegionScheduler();
+@@ -3246,4 +3248,14 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+     // Paper end
++
++    // Paper start - API to allow/disallow tick sleeping
++    @Override
++    public void allowPausing(final org.bukkit.plugin.java.@NotNull JavaPlugin plugin, final boolean value) {
```
Leguan16: The annotation as well.

Abelkrijgtalles:
I am now using FQN imports

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/src/main/java/org/bukkit/Server.java
++++ b/src/main/java/org/bukkit/Server.java
+@@ -2572,4 +2572,16 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
+      */
+     boolean isOwnedByCurrentRegion(@NotNull Entity entity);
+     // Paper end - Folia region threading API
++
++    // Paper start - API to allow/disallow tick sleeping
++    /**
++     * Allow or disallow the server to sleep/pause.This only effects the next time the server is paused. If the server is currently paused, it will only go out of sleep when someone joins.
```
jpenilla: >If the server is currently paused, it will only go out of sleep when someone joins

The diff  in tickServer should be moved up so that it stops sleep on the next tick

jpenilla (dismissed):

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
Abelkrijgtalles: What do you mean by this?

jpenilla: There are earlier if statements you can edit that will be checked during tick sleep

Abelkrijgtalles: So the `if (j > 0)` statement?

Abelkrijgtalles:
I have moved the check up to `if (j>0)`

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
jpenilla: Probably the one with the getPlayerCount check to ensure the log message prints when tick sleeping starts again

jpenilla: This is not resolved, see my last comment, also not sure why you have deleted comments

Abelkrijgtalles: > This is not resolved, see my last comment, also not sure why you have deleted comments

I had also posted the first message I deleted in the main thread, and the other one was just me messing around

Abelkrijgtalles: I have moved it down to the getPlayerCount() if-statement

patches/server/1065-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,74 @@
...
+ import org.bukkit.util.permissions.DefaultPermissions;
++import org.jetbrains.annotations.NotNull;
+ import org.yaml.snakeyaml.LoaderOptions;
+ import org.yaml.snakeyaml.Yaml;
+ import org.yaml.snakeyaml.constructor.SafeConstructor;
+@@ -313,6 +314,7 @@ public final class CraftServer implements Server {
+     private final io.papermc.paper.logging.SysoutCatcher sysoutCatcher = new io.papermc.paper.logging.SysoutCatcher(); // Paper
+     private final io.papermc.paper.potion.PaperPotionBrewer potionBrewer; // Paper - Custom Potion Mixes
+     public final io.papermc.paper.SparksFly spark; // Paper - spark
++    private List<org.bukkit.plugin.java.JavaPlugin> pluginsBlockingSleep = new ArrayList<>(); // Paper - API to allow/disallow tick sleeping
```
kennytv: Storing plugin instances makes memory leaks easy, it might be better to store their names instead and remove entries during plugin disables/unloads. You can also just move the list to MinecraftServer and check if it's empty there instead of adding the extra boolean to MinecraftServer

Abelkrijgtalles: Will do

Abelkrijgtalles: But should I leave the instance part in the API? I think this is more up to standard with the API as a whole and simpler.

kennytv: Yup, only talking about the implementation here

Abelkrijgtalles: I haven't made very many commits yet, so I don't really know exactly how every part works. Do you maybe know a good place to add a call, as there are around 20 onDisable related functions and events

jpenilla: I think it should be fine to just iterate the list if it's not empty in tickServer, removing any entries that point to missing or disabled plugins

Abelkrijgtalles: Wouldn't that be less efficient? Checking every tick if every plugin in the list is enabled sounds not so efficient to me

jpenilla: not enough to make it relevant

1) it's just a map lookup and field read
2) we only iterate when it's non-empty, so during tick sleep and rarely when an invalid entry gets removed

jpenilla: Can probably just use removeIf instead of iterating manually, should keep the isEmpty check though

Abelkrijgtalles: I have made the change with the removeIf() implementation

Abelkrijgtalles: Will now also rebase

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,27 @@
...
++     * When only one plugin disallows pausing, the server will never pause.
++     * If there isn't any pausing configured, the server will never pause.
++     * @param plugin The {@link org.bukkit.plugin.java.JavaPlugin} that's allowing or disallowing pausing.
++     * @param value Whether to allow sleeping of the server (defaults to true).
++     */
++    void allowPausing(@NotNull org.bukkit.plugin.java.JavaPlugin plugin, boolean value);
```
kennytv: This should take Plugin instead of JavaPlugin

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
kennytv:
```suggestion
+     * Allow or disallow the server to sleep/pause. This only affects the next time the server is paused. If the server is currently paused, it will only go out of sleep when someone joins.
```
Is that last sentence still true?

patches/server/1066-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,82 @@
...
+index 64b56abf8900d0424100da460fc68ac964394793..5540994697d987898557d49e28dc1dc9f4a5e8db 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -3,9 +3,6 @@ package net.minecraft.server;
+ import com.google.common.base.Preconditions;
+ import com.google.common.base.Splitter;
+ import com.google.common.collect.ImmutableList;
+-import co.aikar.timings.Timings;
+-import com.destroystokyo.paper.event.server.PaperServerListPingEvent;
+-import com.google.common.base.Stopwatch;
```
kennytv: Drop this

patches/server/1066-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,82 @@
...
+             } else {
+                 this.emptyTicks = 0;
+@@ -3186,4 +3185,18 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         }
+     }
+     // Paper end - Add tick times API and /mspt command
++
++    // Paper start - API to allow/disallow tick sleeping
++    public void addPluginAllowingSleep(String pluginName, boolean value) {
++        if (!value) pluginsBlockingSleep.add(pluginName);
```
kennytv: add braces for the if and else

kennytv (commented):
Nitpicks; for using fields or methods of the current class, it'd be nice if you could also add `this.` qualifiers

patches/server/1066-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,82 @@
...
++    private List<String> pluginsBlockingSleep = new ArrayList<>(); // Paper - API to allow/disallow tick sleeping
+ 
+     public static <S extends MinecraftServer> S spin(Function<Thread, S> serverFactory) {
+         AtomicReference<S> atomicreference = new AtomicReference();
+@@ -1624,7 +1622,8 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         int j = this.pauseWhileEmptySeconds() * 20;
+ 
+         if (j > 0) {
+-            if (this.playerList.getPlayerCount() == 0 && !this.tickRateManager.isSprinting()) {
++            removeDisabledPluginsBlockingSleep(); // Paper - API to allow/disallow tick sleeping
```
jpenilla: this line should be above the j>0 if statement

patches/api/0496-API-to-allow-disallow-tick-sleeping.patch
Abelkrijgtalles: Don't know, will test in a bit

patches/server/1068-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,73 @@
...
++
++    // Paper start - API to allow/disallow tick sleeping
++    public void addPluginAllowingSleep(String pluginName, boolean value) {
++        if (!value) { this.pluginsBlockingSleep.add(pluginName); }
++        else { this.pluginsBlockingSleep.remove(pluginName); }
++    }
++    private void removeDisabledPluginsBlockingSleep() {
++        if (this.pluginsBlockingSleep.isEmpty()) { return; }
++        this.pluginsBlockingSleep.removeIf(plugin -> (
++            io.papermc.paper.plugin.manager.PaperPluginManagerImpl.getInstance().isPluginEnabled(plugin)
```
kennytv: If not enabled

patches/server/1068-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,73 @@
...
+             } else {
+                 this.emptyTicks = 0;
+@@ -3186,4 +3188,18 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         }
+     }
+     // Paper end - Add tick times API and /mspt command
++
++    // Paper start - API to allow/disallow tick sleeping
++    public void addPluginAllowingSleep(String pluginName, boolean value) {
++        if (!value) { this.pluginsBlockingSleep.add(pluginName); }
```
kennytv: Missing line breaks

patches/server/1068-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,73 @@
...
+ import org.yaml.snakeyaml.constructor.SafeConstructor;
+@@ -3246,4 +3247,12 @@ public final class CraftServer implements Server {
+         return this.potionBrewer;
+     }
+     // Paper end
++
++    // Paper start - API to allow/disallow tick sleeping
++    @Override
++    public void allowPausing(final @NotNull org.bukkit.plugin.Plugin plugin, final boolean value) {
++        console.addPluginAllowingSleep(plugin.getName(), value);
```
kennytv: this.

patches/server/1068-API-to-allow-disallow-tick-sleeping.patch
```diff
@@ -0,0 +1,73 @@
...
+     public static final long SERVER_INIT = System.nanoTime(); // Paper - Lag compensation
++    private List<String> pluginsBlockingSleep = new ArrayList<>(); // Paper - API to allow/disallow tick sleeping
+ 
+     public static <S extends MinecraftServer> S spin(Function<Thread, S> serverFactory) {
+         AtomicReference<S> atomicreference = new AtomicReference();
+@@ -1623,8 +1624,9 @@ public abstract class MinecraftServer extends ReentrantBlockableEventLoop<TickTa
+         long i = Util.getNanos();
+         int j = this.pauseWhileEmptySeconds() * 20;
+ 
++        removeDisabledPluginsBlockingSleep(); // Paper - API to allow/disallow tick sleeping
```
kennytv: this

kennytv (changes requested):

kennytv (approved):

Issue #11623, opened 2024-11-17
Server crashes if enchantment is removed with datapacks


david85258:
### Stack trace

```
[03:06:39 ERROR]: Encountered an unexpected exception
java.lang.ExceptionInInitializerError: null
        at net.minecraft.world.inventory.EnchantmentMenu.lambda$slotsChanged$0(EnchantmentMenu.java:168) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.ContainerLevelAccess.lambda$execute$0(ContainerLevelAccess.java:85) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.ContainerLevelAccess$2.evaluate(ContainerLevelAccess.java:72) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.ContainerLevelAccess.execute(ContainerLevelAccess.java:84) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.EnchantmentMenu.slotsChanged(EnchantmentMenu.java:125) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.EnchantmentMenu$1.setChanged(EnchantmentMenu.java:70) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.SimpleContainer.setItem(SimpleContainer.java:226) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.Slot.set(Slot.java:66) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.Slot.setByPlayer(Slot.java:62) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.Slot.setByPlayer(Slot.java:58) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.EnchantmentMenu.quickMoveStack(EnchantmentMenu.java:371) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
        at net.minecraft.world.inventory.AbstractContainerMenu.doClick(AbstractContainerMenu.java:576) ~[paper-1.21.3.jar:1.21.3-29-be886cf]
[...]

PR #11624, opened 2024-11-17
Fix enchantment init

Doc94:
Fixs #11623 like Tags enchantments (and pretty sure many other Registry things) can be removed by datapacks and how CrafBukkit (for legacy) expects vanilla ever exists can cause issues...
This PR just make the Vanillla Enchantments can be null in the Bukkit Enchantment class also mention in docs about this (more "easy" than add Nullable to all the static instances of Enchantment with the same issue (or maybe this can be better)

### DIFF
--- a/patches/api/0496-fix-Enchantment-init.patch
+++ b/patches/api/0496-fix-Enchantment-init.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Doc <nachito94@msn.com>
+Date: Sun, 17 Nov 2024 11:28:34 -0300
+Subject: [PATCH] fix Enchantment init
+
+
+diff --git a/src/main/java/org/bukkit/enchantments/Enchantment.java b/src/main/java/org/bukkit/enchantments/Enchantment.java
+index 43e306584988e39a6daca78c621a937acdce48a1..a6c56e0c5b464a03692985d26336d7fa50d7b145 100644
+--- a/src/main/java/org/bukkit/enchantments/Enchantment.java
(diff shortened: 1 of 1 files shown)


electronicboy (approved):
This is potentially going to be fun if nothing is expecting those fields to be nullable, but, the reality here is that this is somewhat unavoidable

Doc94:
> There is one concern that removing enchantments will break even in vanilla somewhere down the line, as is the case with many other client synchronized registries making the client crash once it does try to reference individual ones somewhere in game logic

hmm exists a form to check this? i mean the datapack allow make this in vanilla if a issue happen is more a report for the MOJIRA? if they allow make this

kennytv:
Okay for enchantments specifically I don't see such use, so it might be okay to remove them as long as all loot tables etc. that reference the enchantments are removed by the datapack? but it makes for weird api here, it'd have to be denoted that they can be null

lynxplay:
At least my two cents on this.
Given we cannot expand this to all data driven types (see damage type), I am unsure if pulling such a move is smart for our API as the Enchantment value would be the *only* one that could be null.

Additionally, removing vanilla enchants is also a rather edge-case imo.
You still have to manually remove from them from *everything else* so I think the workaround of "remove them from everywhere *but* the registry" is a fine one to make to preserve the API here.

More opinions obviously needed tho.

kennytv:
After more internal discussion, this is okay as a temporary solution for those that really need it - can you put this behind an unsupported config option (like `allow-default-registry-removal`)? and otherwise still throw.

Not 100% sure if the config is actually loaded *before* the Enchantment class init, so that'd be something you should verify first plz

Doc94:
Based in the messages of hardfork a better solution require a little more of work currently dont know how to do correctly.

PR #11631, opened 2024-11-17
fix: move to jline-terminal-ffm on java 22+ and fall back to jni on 21

### LINKED ISSUE #10405: Java 22 breaks typing in console on windows.

### LINKED ISSUE #10674: Chore: Update Terminal dependencies

MiniDigger:
ffm requires 1) native access allowed (the jdk cracks down on undocumented native access in 22) and 2) reverting the default console back to java.base, so the internal jline doesnt take over

tested on windows 11, windows terminal cmd/ps java 21/22/23 and cmd/ps java 21/22/23 without windows terminal.

(Closes #10674)
(Closes #10405)
(Replaces #10675)
(Requires https://github.com/PaperMC/paperweight/pull/263)

### DIFF
--- a/patches/server/0011-Use-TerminalConsoleAppender-for-console-improvements.patch
+++ b/patches/server/0011-Use-TerminalConsoleAppender-for-console-improvements.patch
@@ -25,7 +25,7 @@ Other changes:
 Co-Authored-By: Emilia Kond <emilia@rymiel.space>
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index 220d2696bf40d1657c87d049f563ccfffed2e8ad..c808f36b30d0f8edb8365875039fd20b8c20007d 100644
+index 220d2696bf40d1657c87d049f563ccfffed2e8ad..99e44684a3f6340ed3c0f73c690a9d4d51872f0d 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -5,6 +5,12 @@ plugins {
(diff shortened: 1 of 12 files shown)


patches/server/0011-Use-TerminalConsoleAppender-for-console-improvements.patch
```diff
@@ -64,7 +65,20 @@ index 220d2696bf40d1657c87d049f563ccfffed2e8ad..c808f36b30d0f8edb8365875039fd20b
      implementation("org.apache.logging.log4j:log4j-iostreams:2.22.1") // Paper - remove exclusion
      implementation("org.ow2.asm:asm-commons:9.7.1")
      implementation("org.spongepowered:configurate-yaml:4.2.0-SNAPSHOT") // Paper - config files
-@@ -92,6 +112,19 @@ tasks.check {
+@@ -71,6 +92,12 @@ tasks.jar {
+     }
+ }
+ 
++// Paper start - allow native access for jline-ffm
++rootProject.tasks.named<io.papermc.paperweight.tasks.CreateBundlerJar>("createMojmapBundlerJar") {
```
jpenilla: Kind of annoying that we have to resort to cross-project configuration here, since putting it in the root build file would require forks to copy the diff and it wouldn't be associated with the patch. Maybe we should just add the attribute by default in paperweight (using `.convention(mapOf(...))`); it would still allow adding extras manually and removing this one if needed.

kennytv (approved):

PR #11648, opened 2024-11-23
Painting variant registry modification API

kokiriglade:


lynxplay:
Your tests are failing.
Also this diff should be merged into the existing patch for registry entries and builders.
Add yourself as a co-author obviously, but no point in a new patch.

Shared diff, like the one in  Conversions, should go into the appropriate Registry Modification patch too.

kokiriglade:
_should_ be fixed now, sorry about that. I had some trouble moving the patch around, but I think it's OK

lynxplay (approved):

Machine-Maker (approved):

PR #11659, opened 2024-11-24
RegistryValue API

Machine-Maker:


Machine-Maker:
No longer needed because we need the Either pattern anyways.

PR #11660, opened 2024-11-24
Fix painting variant builder

Lulu13022002:
Fixes a typo when a plugin try to change already existing variant where the author of all the other paintings would be set to the title. And generalize the test to avoid that in the future.


patches/server/1035-DataComponent-API.patch
```diff
@@ -1533,7 +1533,7 @@ index 0000000000000000000000000000000000000000..c7e40bd15b7063f155b2065927e8201f
 +    static final class BuilderImpl implements ItemAdventurePredicate.Builder {
 +
 +        private final List<net.minecraft.advancements.critereon.BlockPredicate> predicates = new ObjectArrayList<>();
-+        private boolean showInTooltip = net.minecraft.world.item.component.ItemAttributeModifiers.EMPTY.showInTooltip();
```
lynxplay, line 1536: What happened here?

lynxplay (approved):

PR #11667, opened 2024-11-25
Add EntityEquipmentChangeEvent

TonytheMacaroni:
Adds `EntityEquipmentChangeEvent`, a more generic version of `PlayerArmorChangeEvent` that applies to the equipment changes of living entities.

At the moment, the patch changes `PlayerArmorChangeEvent` to extend `EntityEquipmentChangeEvent`. I'm not sure on this, especially since that would make `PlayerArmorChangeEvent` no longer extend `PlayerEvent`.

Another alternative I considered was for the event to instead expose a map of slots -> equipment changes, so that the event is called only once when multiple slots are changed, and avoiding changing the inheritance of `PlayerArmorChangeEvent`.


lynxplay:
Alternatively we deprecate PlayerArmorChangeEvent in favour of a PlayerEquipmentChangeEvent which then can extend that.

Machine-Maker:
If we are gonna replace the playerarmorchangeevent, I would kind of like to chose a new name that more accurately reflects what its detecting. It's detecting on the next tick if the armor has changed from what it was previously to send to clients.

PR #11669, opened 2024-11-26
 Added PlayerUseBowWithoutProjectileEvent #11668 

Chaosdave34:
Added PlayerUseBowWithoutProjectileEvent when a player tries to load a crossbow or draw a bow without having a suitable projectile in their inventory. Makes it possible to set a projectile and enable the player to use the weapon nevertheless.

### DIFF
--- a/patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
+++ b/patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
@@ -6,10 +6,10 @@ Subject: [PATCH] Added PlayerUseBowWithoutProjectileEvent
 
 diff --git a/src/main/java/io/papermc/paper/event/player/PlayerUseBowWithoutProjectileEvent.java b/src/main/java/io/papermc/paper/event/player/PlayerUseBowWithoutProjectileEvent.java
 new file mode 100644
-index 0000000000000000000000000000000000000000..9ddb8900f33063b4e93cda38b2fb760f1ce50eaf
+index 0000000000000000000000000000000000000000..dbc2fd3a33d528789fdd39a1c98c656ae4c7be5d
 --- /dev/null
 +++ b/src/main/java/io/papermc/paper/event/player/PlayerUseBowWithoutProjectileEvent.java
-@@ -0,0 +1,53 @@
+@@ -0,0 +1,54 @@
 +package io.papermc.paper.event.player;
 +
 +import org.bukkit.entity.Player;
@@ -32,6 +32,7 @@ index 0000000000000000000000000000000000000000..9ddb8900f33063b4e93cda38b2fb760f
 +    @ApiStatus.Internal
 +    public PlayerUseBowWithoutProjectileEvent(final Player player) {
 +        super(player);
++        this.projectile = ItemStack.empty();
 +    }
 +
 +    /**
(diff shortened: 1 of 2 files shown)


patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
```diff
@@ -0,0 +1,47 @@
...
+@@ -0,0 +1,35 @@
++package io.papermc.paper.event.player;
++
++import org.bukkit.entity.Player;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.player.PlayerEvent;
++import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.NotNull;
++
++public class PlayerUseBowWithoutProjectileEvent extends PlayerEvent {
```
lynxplay, line 26: `@NullMarked`

patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++
++    private static final HandlerList HANDLER_LIST = new HandlerList();
++
++    private ItemStack projectile;
++
++    public PlayerUseBowWithoutProjectileEvent(@NotNull final Player player) {
```
lynxplay: Remove `@NotNull` as per contributing.md, we use jspecify annotations in new paper types.

patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++        super(player);
++    }
++
++    public ItemStack getProjectile() {
++        return projectile;
```
lynxplay: Requires defensive copy.

patches/api/0501-Added-PlayerUseBowWithoutProjectileEvent.patch
```diff
@@ -0,0 +1,47 @@
...
++    }
++
++    public void setProjectile(ItemStack projectile) {
++        this.projectile = projectile;
```
lynxplay: Requires defensive copy.

lynxplay (changes requested):
Welcome to paper and thank you for your first PR :partying_face: 

Left some basic comments regarding annotations.
The PR is also missing javadocs on the API types.

Beyond that, the event name feels a bit off, the player isn't using the bow there, it is searching for a projectile. We already have some other events there so *maybe* we can look into merging those concepts? Not sure tho.

Chaosdave34:
Maybe PlayerTryUseBowWithoutProjectileEvent would be better?

I already thought about using the PlayerReadyArrowEvent for this.
The simplest way to implement the functionality to let a player fire a bow without having an arrow in their inventory seems to be this for me, because when creating a new ItemStack (even with CraftItemStack#asNmsCopy(), to set the arrow from the CraftItemStack used in the event), it would be diffcult to consume the arrow (removing the arrow from the player inventory) when the arrow ist not manipulated by the event.
(get the arrow ->  create a craftbukkit copy -> call the event with the arrow -> the arrow is not manipulated -> get the arrow from the event -> create a nms copy -> created a new ItemStack although the original one was not changed -> Player#useAmmo() breaks; or am I missing something?)

lynxplay:
No you are not missing anything I don't think.
I don't see a nice way to have this interact with player ready arrow event either no.

Naming wise, well, realistically this should probably not be limited to bows only.
It is a general "PlayerFailedProjectileSearch" or something (shitty name but you get the gist :sweat_smile:) as this is going to be called for crossbows as well.

I'll try to get this PR into a pr party next weekend so I can poke some other maintainers on a better name for this, but I think the functionality this is adding is worth having :+1:

PR #11671, opened 2024-11-26
Fix incorrect command node serialization by creating new Command

### LINKED ISSUE #11649: Brigadier: Incorrect serialization of same-name command nodes causing client to suggest invalid literals
### Expected behavior

With the following snippet:
```java
            event.registrar().register(
                    literal("hide")
                            .executes(this::hide)
                            .then(literal("something-else")
                                    .executes(this::hideSomethingElse)).build());

            event.registrar().register(
                    literal("chat")
                            .then(literal("hide")
                                    .executes(this::chatHide)).build());
        });
```

One would expect the client to suggest the following commands, and it does:
`hide`
`hide something-else`
`chat hide`

Image
Image
Image

I expect no other hide/chat related commands to be suggested.

### Observed/Actual behavior

Interestingly, I found that the client **also** suggests

`chat hide something-else`

Image

which is of course not a valid command:

Image

So why does the client suggest it? (See analysis below)

### Steps/models to reproduce

```java
            event.registrar().register(
                    literal("hide")
                            .executes(this::hide)
                            .then(literal("something-else")
                                    .executes(this::hideSomethingElse)).build());

            event.registrar().register(
                    literal("chat")
                            .then(literal("hide")
                                    .executes(this::chatHide)).build());
        });
```

### Plugin and Datapack List

Only the example plugin to enable the above snippet.

Image

### Paper version

This server is running Paper version 1.21.3-48-master@817550c (2024-11-22T12:02:23Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
Previous version: 1.21.1-132-b48403b (MC: 1.21.1)

Also on 1.21 build 132 (previous version)

### Other

The direct cause of the problem is that `ClientboundCommandsPacket` maps the two `hide` nodes into one, because it thinks they are equal (`equals()` of both `hide` nodes returns true! and hashcode is probably equivalent as well). This is usually normal and correct behaviour, but these nodes really should not be considered equal.

Looking into the `equals` method of (Literal)CommandNode, we see that `children` and `command` are checked for equality. But from our example, `command` should be different for both. At serialization time, however, it _is_ equal because Commands#fillUsableCommands overwrites the to-be-serialized nodes with a local lambda.

Chat-hide node:
Image

Hide node:
Image

Note the `command` field being Commands$lambda@29095 for both nodes. Since our command structure does not add any children to the hides, the `equals` method returns true, because the command is equal.

My suggestion is altering the `Commands#fillUsableCommands` method from
```java
argumentbuilder.executes((commandcontext) -> {
  return 0;
});
```
to, for example, return a unique `Command` every time:
```java
argumentbuilder.executes(new Command<CommandSourceStack>() {
    @Override
    public int run(CommandContext<CommandSourceStack> commandContext) throws CommandSyntaxException {
        return 0;
    }
});
```
[Example_patch.patch](https://github.com/user-attachments/files/17880116/Example_patch.patch)

This could prevent reuse of nodes in the serialized version at all because equals() will now differ for all command nodes. So maybe a more proper solution is to track a mapping from original command -> new command (or just keep the original command in? Why would it be overwritten?)

There likely is also a related issue with `redirect` not being used in the check for equality nor the hashcode, causing two otherwise identical nodes that redirect to other nodes to be serialized as one and the same - thus potentially redirecting incorrectly on the client. At time of writing I'm unable to make a proof of concept though, because there seems to be a bit of magic going on around redirects.

If this is deemed a suitable fix, I can make a PR (or someone else can with the attached patch, I don't mind)

Rickyboy320:
Fixes #11649 - As noted in the issue, when CommandNodes are serialized they are used as the key in a Map. Their equals()/hashcode() should only match if they are equal nodes (name & command), but due to the erasure of the command field pre-serialization, nodes with different commands can be mapped onto the same value. This causes the client to interpret both nodes as the same, causing suggestions where they should not.

This is fixed by creating a different no-op Command for the erasure, instead of them holding the same lambda.

### DIFF
--- a/patches/server/1074-Fix-incorrect-command-serialization-by-creating-new-.patch
+++ b/patches/server/1074-Fix-incorrect-command-serialization-by-creating-new-.patch
@@ -1,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Rick <rickw320@hotmail.com>
+Date: Tue, 26 Nov 2024 20:45:52 +0100
+Subject: [PATCH] Fix incorrect command serialization by creating new Command
+
+Fixes #11649 - As noted in the issue, when CommandNodes are serialized
+they are used as the key in a Map. Their equals()/hashcode() should only
+match if they are equal nodes (name & command), but due to the erasure of the command field pre-serialization, nodes with different commands can be mapped onto the same value. This causes the client to interpret both nodes as the same, causing suggestions where they should not.
+
+This is fixed by creating a different no-op command for the
+erasure, instead of them holding the same lambda.
+
+diff --git a/src/main/java/net/minecraft/commands/Commands.java b/src/main/java/net/minecraft/commands/Commands.java
+index 64bf4444ffba25cb40743a32267aa790ad1738f9..517cb238ec280aadd1fc54bcb675ed386e798eaf 100644
+--- a/src/main/java/net/minecraft/commands/Commands.java
++++ b/src/main/java/net/minecraft/commands/Commands.java
+@@ -587,9 +587,14 @@ public class Commands {
+                     return true;
+                 });
+                 if (argumentbuilder.getCommand() != null) {
+-                    argumentbuilder.executes((commandcontext) -> {
+-                        return 0;
++                    // Paper start - fix suggestions due to falsely equal nodes
++                    argumentbuilder.executes(new com.mojang.brigadier.Command<io.papermc.paper.command.brigadier.CommandSourceStack>() {
++                        @Override
++                        public int run(com.mojang.brigadier.context.CommandContext<io.papermc.paper.command.brigadier.CommandSourceStack> commandContext) throws CommandSyntaxException {
++                            return 0;
++                        }
+                     });
++                    // Paper end
+                 }
+ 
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

lynxplay (approved):

Machine-Maker:
This fix has created an issue. It breaks the compactness serialized size of the commands. For any plugin that uses "flat" aliases (aliases that just shallow copy the root literal), it now doubles the serialized size of the command tree instead of re-using the references to the children command nodes. This is because the hashmap that tracks a serialization id for each CommandNode no longer finds existing nodes because the Command instances are not equal.

Rickyboy320:
I suggested another approach in the [related issue](https://github.com/PaperMC/Paper/issues/11649), where we keep the original command in the node, so that the equality of the fake node matches the original node (and other copies thereof). I'm not sure what side-effects this has, however, since it seems like the command field is intentionally wiped

lynxplay:
Yea we are presumably going to opt towards that approach down the line to avoid the tree reduction from failing as hard as it has been :+1: The approach here worked well enough, it was only ever exposed as an issue cause once a couple of other bugs were fixed as well.

PR #11675, opened 2024-11-27
Fix getMaxHeight to be exclusive instead of inclusive to match docs

### LINKED ISSUE #11674: `getMaxHeight()` has incorrect behavior since 1.21.3
### Expected behavior

That it works like what the docs says, that value should be exclusive, and not inclusive, which is how it worked before 1.21.3.

### Observed/Actual behavior

It returns inclusive, not exclusive. In a world that has 320 block height, the method returns 319. In 1.21.1 it worked correctly, returning 320.

### Steps/models to reproduce

1. Use `world.getMaxHeight()` on a normal, vanilla, 320 block height world
2. See that it returns 319 instead of 320

### Plugin and Datapack List

Not applicable

### Paper version

```
This server is running SparklyPaper version 1.21.3-DEV-ver/1.21.3@586de79 (2024-11-25T21:05:13Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
Unknown version
Previous version: 1.21.3-DEV-32b3c0c (MC: 1.21.3)
```

Yeah it is my own fork, the fork is based off `a63013cf48e46e2ec056c7ace60e018ec9be208e`... jmp did say that "mojang changed the internal method to be inclusive" so I'm 99,99999% sure that this also affects Paper, and upstream Spigot.

### Other

_No response_

MrPowerGamerBR:
The `getMaxHeight` docs says that the value is exclusive, not inclusive. This has changed in 1.21.3 and since 1.21.3 the value is now inclusive due to Mojang changes. This patch changes the value to be exclusive by +1 the returned value from NMS.

Worth noting that this needs to be tested! I haven't tested it yet, and I seen that there are some CraftBukkit code that does call `getMaxHeight` (`setBiome`)

Fixes #11674

### DIFF
--- a/patches/server/1073-Fix-getMaxHeight-to-be-exclusive-instead-of-inclusiv.patch
+++ b/patches/server/1073-Fix-getMaxHeight-to-be-exclusive-instead-of-inclusiv.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MrPowerGamerBR <git@mrpowergamerbr.com>
+Date: Wed, 27 Nov 2024 02:26:32 -0300
+Subject: [PATCH] Fix getMaxHeight to be exclusive instead of inclusive to
+ match docs
+
+
+diff --git a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+index 8f234b46c24a0ae08aa5f8190c5b27e1f62dfbab..372581a1a44870c53afa5ecd456b1b09ea77b2ae 100644
+--- a/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
++++ b/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
+@@ -1538,7 +1538,7 @@ public class CraftWorld extends CraftRegionAccessor implements World {
+ 
+     @Override
+     public int getMaxHeight() {
+-        return this.world.getMaxY();
++        return this.world.getMaxY() + 1; // Paper - Fix getMaxHeight to be exclusive instead of inclusive
+     }
+ 
+     @Override


Lulu13022002:
This is not the only method affected:
Chunk#getBlock/getChunkSnapshot are now inclusive
Chunk#getEmptyChunkSnapshot throw an ArrayIndexOutOfBoundsException with your changes (at y = getMaxHeight)

NonSwag:
Wouldn't it be smarter to adjust the docs instead of the method?

electronicboy:
No, the API docs describe the contract of the method, there is no sane justification for changing the API contract here

MrPowerGamerBR:
> This is not the only method affected: ~Chunk#getBlock/getChunkSnapshot are now inclusive~, nvm got confused by what you called exclusive compared to the doc Chunk#getEmptyChunkSnapshot throw an ArrayIndexOutOfBoundsException with your changes (at y = getMaxHeight) instead of the IllegalArgumentException

Looks like I ended up checking only `getMaxHeight` usages on the API module, not on the server...

Well now this is a bit trickier, I don't know if I should assume that all `getMaxHeight` on the server are assuming that the value will be inclusive (new docs) or exclusive (old docs).

> Wouldn't it be smarter to adjust the docs instead of the method?

Plugins already rely on the behavior of the current docs, I found this issue because my plugin stopped checking for the block at the top of the world due to this, so I dug deeper to find what was causing the issue. Other plugins also have issues due to things change, like mcMMO.

```kotlin
for (y in world.minHeight until world.maxHeight) {
   ...
}
```

jacob1:
I ran into this same issue while updating Dynmap. I reported it to Spigot and it's [fixed upstream](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/b75b792ec3df00b6a6b07f9f559e2d016a964486) now.

Looks like an identical patch to this one. Any potential issues included.

Brokkonaut:
It has been fixed upstream, so I think this patch is no longer needed, just an upstream pull would do it

Lulu13022002:
I just checked all the usages and the issue has been fixed in 1.21.4 (technically latest commit of 1.21.3 with paper fixes), thanks for the report regardless.

PR #11685, opened 2024-11-28
Add ItemMeta customName methods

kennytv:
This also needs to rename some PotionMeta methods to not conflict (which we really should do now-ish rather than later, since we can't ignore the customName name forever), but it still needs plugin rewriting. so this might benefit from waiting for 1.21.4


patches/api/0006-Adventure.patch
```diff
@@ -4713,35 +4713,74 @@ index 9bab73c3c2ca759b8e1c7d07d98cc593c961666a..f0c6943da3f783101ca647b75b3230fa
...
++     * Sets the custom name.
++     *
++     * @param customName the custom name to set
++     */
++    void customName(final net.kyori.adventure.text.@Nullable Component customName);
++
+     /**
+      * Checks for existence of a display name.
+      *
+      * @return true if this has a display name
```
Machine-Maker, line 4751: Lets add an apiNote/see to link to the other method

Machine-Maker (changes requested):
Paper comments in server patch

Machine-Maker (approved):

PR #11689, opened 2024-11-30
Extend HumanEntity#dropItem API

### LINKED ISSUE #11656: Player#dropItem(ItemStack)

Currently, there is only a `Player#dropItem(boolean)` method to drop the held item of the player, so when dropping an item from the inventory from a player you have to implement the entire drop logic yourself _or use nms_

A `Player#dropItem(ItemStack)` method
and maybe even a `Player#dropItem(EquipmentSlot)` method

Using NMS or implementing dropping yourself

### Other

My best attempt at implementing it myself (basically what vanilla does)
```java
player.getInventory().remove(item);
var location = player.getEyeLocation().clone();
location.setY(location.getY() - 0.30000001192092896D);
player.getWorld().dropItem(location, item, drop -> {

    var random = ThreadLocalRandom.current();

    var f1 = Mth.sin(player.getPitch() * 0.017453292F);
    var f2 = Mth.cos(player.getPitch() * 0.017453292F);
    var f3 = Mth.sin(player.getYaw() * 0.017453292F);
    var f4 = Mth.cos(player.getYaw() * 0.017453292F);
    var f5 = random.nextFloat() * 6.2831855F;
    var f6 = 0.02F * random.nextFloat();

    var x = (double) (f4 * f2 * 0.3F) + Math.sin(f5) * (double) f6;
    var y = (double) -f1 * 0.3F + 0.1F + (random.nextFloat() - random.nextFloat()) * 0.1F;
    var z = (double) (-f3 * f2 * 0.3F) + Math.cos(f5) * (double) f6;
    drop.setVelocity(new Vector(z, y, x));

    drop.setThrower(player.getUniqueId());
    drop.setPickupDelay(40);
});
```
or with NMS
```java
((CraftPlayer) player).getHandle().drop(((CraftItemStack) item).handle, false, true, true);
```
Additionally, to make it seem like the player dropped the item themselves, swing their main hand
```java
player.swingMainHand();
```

Strokkur424:
Closes #11656

# Motive
Developers currently have to either use NMS or copy-paste the already existing NMS implementation. This PR aims to fix that by exposing ways to call that NMS implementation more easily by extending onto the (Human)Player#dropItem method to provide more extensive ways of making the player drop items.

## What does it add?
The following overloads to the `HumanPlayer#dropItem(...)` method have been added:
* `Item dropItem(ItemStack itemStack)`
* `Item dropItem(ItemStack itemStack, boolean persistThrower)`
* `Item dropItem(ItemStack itemStack, boolean persistThrower, boolean throwRandomly)`
* `Item dropItem(int slot)`
* `Item dropItem(int slot, boolean persistThrower)`
* `Item dropItem(int slot, boolean persistThrower, boolean throwRandomly)`
* `Item dropItem(EquipmentSlot slot)`
* `Item dropItem(EquipmentSlot slot, boolean persistThrower)`
* `Item dropItem(EquipmentSlot slot, boolean persistThrower, boolean throwRandomly)`

## Implementation details
Internally, the implementation calls `net.minecraft.world.entity.player.Player#dropItem(...)`, making it a fairly maintainable API. But as that method does not remove the item, I have added that by calling `this.inventory.setItem(slot, null)`.

## JavaDocs status
I have added simple JavaDocs, but the methods should be fairly self explanatory.

# Usage example
```java
/**
 * Makes the player drop an item of the type they using. Does not cancel the event
 */
@EventHandler
public void onInteract(final PlayerInteractEvent event) {
    event.getPlayer().dropItem(event.getItem());
    event.setCancelled(true);
}

/**
 * Strips the player of all currently wearing armor
 */
private void stripOfArmor(final Player player) {
    player.dropItem(EquipmentSlot.HEAD);
    player.dropItem(EquipmentSlot.CHEST);
    player.dropItem(EquipmentSlot.LEGS);
    player.dropItem(EquipmentSlot.FEET);
}
```

> [!NOTE]
> I am not 100% certain that the methods work in every possible scenario, as I don't know how to test that. If somebody finds a bug, make sure to tell me and I will make sure to fix it.

### DIFF
--- a/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
+++ b/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
@@ -5,10 +5,10 @@ Subject: [PATCH] Extend HumanEntity#dropItem API
 
 
 diff --git a/src/main/java/org/bukkit/entity/HumanEntity.java b/src/main/java/org/bukkit/entity/HumanEntity.java
-index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48cc621b547 100644
+index 488604ba1a516b477693877c74712e4a45624a8b..75493dabe98490a74fc0ce20c47445fde4d0c171 100644
 --- a/src/main/java/org/bukkit/entity/HumanEntity.java
 +++ b/src/main/java/org/bukkit/entity/HumanEntity.java
-@@ -706,6 +707,97 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
+@@ -706,6 +706,97 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
       */
      public boolean dropItem(boolean dropAll);
  
@@ -27,27 +27,27 @@ index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48c
 +    /**
 +     * Makes the entity drop an item from their inventory.
 +     * <br>
-+     * This method calls {@link HumanEntity#dropItem(int slot, boolean retainOwnership)}
++     * This method calls {@link HumanEntity#dropItem(int slot, boolean persistThrower)}
 +     * with the first {@link ItemStack} occurrence in the inventory
 +     *
 +     * @param itemStack The ItemStack to drop from their inventory
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @return The dropped item, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final ItemStack itemStack, final boolean retainOwnership);
++    public @Nullable Item dropItem(final ItemStack itemStack, final boolean persistThrower);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the specified ItemStack.
 +     * <br>
-+     * This method calls {@link HumanEntity#dropItem(int slot, boolean retainOwnership, boolean throwRandomly)}
++     * This method calls {@link HumanEntity#dropItem(int slot, boolean persistThrower, boolean throwRandomly)}
 +     * with the first {@link ItemStack} occurrence in the inventory
 +     *
 +     * @param itemStack The ItemStack to drop
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @param throwRandomly Whether the item should disperse randomly
 +     * @return The dropped item, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final ItemStack itemStack, final boolean retainOwnership, final boolean throwRandomly);
++    public @Nullable Item dropItem(final ItemStack itemStack, final boolean persistThrower, final boolean throwRandomly);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the slot.
@@ -61,20 +61,20 @@ index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48c
 +     * Makes the entity drop an item from their inventory based on the slot.
 +     *
 +     * @param slot The slot to drop
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @return The dropped item, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final int slot, final boolean retainOwnership);
++    public @Nullable Item dropItem(final int slot, final boolean persistThrower);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the slot.
 +     *
 +     * @param slot The slot to drop
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @param throwRandomly Whether the item should disperse randomly
 +     * @return The dropped item entity, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final int slot, final boolean retainOwnership, final boolean throwRandomly);
++    public @Nullable Item dropItem(final int slot, final boolean persistThrower, final boolean throwRandomly);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the {@link org.bukkit.inventory.EquipmentSlot}
@@ -87,20 +87,20 @@ index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48c
 +     * Makes the entity drop an item from their inventory based on the {@link org.bukkit.inventory.EquipmentSlot}
 +     *
 +     * @param slot The equipment slot to drop
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @return The dropped item entity, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean retainOwnership);
++    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean persistThrower);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the equipment slot.
 +     *
 +     * @param slot The equipment slot to drop
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param persistThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
 +     * @param throwRandomly Whether the item should disperse randomly
 +     * @return The dropped item entity, or null if the action was unsuccessful
 +     */
-+    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean retainOwnership, final boolean throwRandomly);
++    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean persistThrower, final boolean throwRandomly);
 +    // Paper end
 +
      /**
--- a/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
+++ b/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Extend HumanEntity#dropItem API
 
 
 diff --git a/src/main/java/org/bukkit/entity/HumanEntity.java b/src/main/java/org/bukkit/entity/HumanEntity.java
-index 488604ba1a516b477693877c74712e4a45624a8b..1f3bb91105882b1d729bd32274405bee0d2d42be 100644
+index 488604ba1a516b477693877c74712e4a45624a8b..47be1ec902194127b6d424fe808cdb4029906f29 100644
 --- a/src/main/java/org/bukkit/entity/HumanEntity.java
 +++ b/src/main/java/org/bukkit/entity/HumanEntity.java
 @@ -6,6 +6,7 @@ import org.bukkit.GameMode;
@@ -64,7 +64,7 @@ index 488604ba1a516b477693877c74712e4a45624a8b..1f3bb91105882b1d729bd32274405bee
 +     * @return The dropped item, or null if the action was unsuccessful
 +     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
 +     */
-+    public Item dropItem(final int slot);
++    public @Nullable Item dropItem(final int slot);
 +
 +    /**
 +     * Makes the entity drop an item from their inventory based on the slot.
--- a/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
+++ b/patches/api/0501-Extend-HumanEntity-dropItem-API.patch
@@ -1,0 +1,108 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Strokkur24 <strokkur24@gmail.com>
+Date: Sat, 30 Nov 2024 11:54:38 +0100
+Subject: [PATCH] Extend HumanEntity#dropItem API
+
+
+diff --git a/src/main/java/org/bukkit/entity/HumanEntity.java b/src/main/java/org/bukkit/entity/HumanEntity.java
+index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48cc621b547 100644
+--- a/src/main/java/org/bukkit/entity/HumanEntity.java
++++ b/src/main/java/org/bukkit/entity/HumanEntity.java
+@@ -706,6 +707,97 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder
+      */
+     public boolean dropItem(boolean dropAll);
+ 
++    // Paper start - Extend HumanEntity#dropItem API
++    /**
++     * Makes the entity drop an item from their inventory.
++     * <br>
++     * This method calls {@link HumanEntity#dropItem(int slot)}
++     * with the first {@link ItemStack} occurrence in the inventory
++     *
++     * @param itemStack The ItemStack to drop
++     * @return The dropped item, or null if the action was unsuccessful
++     */
++    public Item dropItem(final ItemStack itemStack);
++
++    /**
++     * Makes the entity drop an item from their inventory.
++     * <br>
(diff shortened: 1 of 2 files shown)


NonSwag:
Thanks for that PR, I was too lazy to do it myself :smile: 
Internally, the item is already being removed from the player's inventory by setting its type to air and amount to 0 so you don't need to remove it yourself though

NonSwag:
I don't think slot related methods are a good idea
having methods for itemstack and equipment slots cover basically all needs

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,95 @@
...
++        if (is == null || is.getType() == Material.AIR) {
++            return null;
++        }
++
++        final net.minecraft.world.entity.item.ItemEntity droppedEntity = player.drop(CraftItemStack.asNMSCopy(is), retainOwnership, throwRandomly);
++        if (droppedEntity == null) {
++            return null;
++        }
++
++        this.swingMainHand();
```
NonSwag: I am not sure if it is a smart idea to swing the main hand internally
maybe in some cases, the user of this method wants the player to swing their offhand so it should be up to them to do such a thing

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,95 @@
...
++    public @org.jetbrains.annotations.Nullable org.bukkit.entity.Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean retainOwnership, final boolean throwRandomly) {
++        final ItemStack stack = this.inventory.getItem(slot);
++        final org.bukkit.entity.Item itemEntity = dropItemRaw(stack, retainOwnership, throwRandomly);
++
++        this.inventory.setItem(slot, null);
++        return itemEntity;
++    }
++
++    private org.bukkit.entity.Item dropItemRaw(final ItemStack is, final boolean retainOwnership, final boolean throwRandomly) {
++        if (!(this.getHandle() instanceof ServerPlayer player)) {
```
NonSwag: Why not on the server player themselves?

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,95 @@
...
++            return null;
++        }
++
++        if (is == null || is.getType() == Material.AIR) {
```
NonSwag, line 80: Probably an exception would be more appropriate

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,95 @@
...
++
++    @Override
++    public @org.jetbrains.annotations.Nullable org.bukkit.entity.Item dropItem(final int slot, final boolean retainOwnership) {
++        return this.dropItem(slot, retainOwnership, false);
++    }
++
++    @Override
++    public @org.jetbrains.annotations.Nullable org.bukkit.entity.Item dropItem(final int slot, final boolean retainOwnership, final boolean throwRandomly) {
++        // Make sure the slot is in bounds
++        if (slot < 0 || this.inventory.getSize() <= slot) {
```
NonSwag, line 49: If this method stays (see my previous comment) an exception would be more appropriate

Strokkur424:
> I don't think slot related methods are a good idea having methods for itemstack and equipment slots cover basically all needs

The issue with that is that dropping by ItemStack pretty much always forces the **first** ItemStack to be dropped. Dropping by slot allows to choose an explicit one, like the one current held, as example

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++    public Item dropItem(final ItemStack itemStack);
++
++    /**
++     * Makes the entity drop an item from their inventory.
++     * <br>
++     * This method calls {@link HumanEntity#dropItem(int slot, boolean retainOwnership)}
++     * with the first {@link ItemStack} occurrence in the inventory
++     *
++     * @param itemStack The ItemStack to drop from their inventory
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: retainOwnership is very misleading (from nms itself) it is not setting the owner but the thrower
the difference here is that if an item is owned it can only be picked up by this very owner
the thrower is just the entity's uuid that threw the item
potentially a less misleading name would be appropriate or clarification in the docs

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++    public @Nullable Item dropItem(final ItemStack itemStack, final boolean retainOwnership);
++
++    /**
++     * Makes the entity drop an item from their inventory based on the specified ItemStack.
++     * <br>
++     * This method calls {@link HumanEntity#dropItem(int slot, boolean retainOwnership, boolean throwRandomly)}
++     * with the first {@link ItemStack} occurrence in the inventory
++     *
++     * @param itemStack The ItemStack to drop
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: same here

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++     * @param slot The slot to drop
++     * @return The dropped item, or null if the action was unsuccessful
++     */
++    public Item dropItem(final int slot);
++
++    /**
++     * Makes the entity drop an item from their inventory based on the slot.
++     *
++     * @param slot The slot to drop
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: same here

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++     * @return The dropped item, or null if the action was unsuccessful
++     */
++    public @Nullable Item dropItem(final int slot, final boolean retainOwnership);
++
++    /**
++     * Makes the entity drop an item from their inventory based on the slot.
++     *
++     * @param slot The slot to drop
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: here

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++     *
++     * @param slot The equipment slot to drop
++     * @return The dropped item entity, or null if the action was unsuccessful
++     */
++    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot);
++    /**
++     * Makes the entity drop an item from their inventory based on the {@link org.bukkit.inventory.EquipmentSlot}
++     *
++     * @param slot The equipment slot to drop
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: here

Strokkur424:
> Internally, the item is already being removed from the player's inventory by setting its type to air and amount to 0 so you don't need to remove it yourself though

That sadly not the case for items equipped on armor. I am not quite sure why, but in my testing, armor items did not get cleared by themselves

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -0,0 +1,108 @@
...
++     * @return The dropped item entity, or null if the action was unsuccessful
++     */
++    public @Nullable Item dropItem(final org.bukkit.inventory.EquipmentSlot slot, final boolean retainOwnership);
++
++    /**
++     * Makes the entity drop an item from their inventory based on the equipment slot.
++     *
++     * @param slot The equipment slot to drop
++     * @param retainOwnership Whether this player should own the resulting item entity
```
NonSwag: here

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
Strokkur424: Ah okay, thanks for clarifying that. It mislead even me 😅

NonSwag:
> > Internally, the item is already being removed from the player's inventory by setting its type to air and amount to 0 so you don't need to remove it yourself though
> 
> That sadly not the case for items equipped on armor. I am not quite sure why, but in my testing, armor items did not get cleared by themselves

in my tests it worked (1.21.1 and 1.21.3)

NonSwag:
> > I don't think slot related methods are a good idea having methods for itemstack and equipment slots cover basically all needs
> 
> The issue with that is that dropping by ItemStack pretty much always forces the **first** ItemStack to be dropped. Dropping by slot allows to choose an explicit one, like the one current held, as example

I did not think of that
in that case, it totally makes sense

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
Strokkur424: I would rather make it return null, as in no item was found, so no item was dropped. Because otherwise you'd have to manually run a check for whether the item actually exists, and it just turns into a mess

NonSwag: valid point

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
Strokkur424: Ah, the object returned by `this.getHandle()` already has the method definition for `#drop(...)`. I kinda just looked at the one `#dropItem(boolean)` method and they had used a ServerPlayer as well, so I kind of assumed the nms player to not have that method

Strokkur424:
> in my tests it worked (1.21.1 and 1.21.3)

When I remove that line, it turns into an dupe glitch 😥

https://github.com/user-attachments/assets/86db1db4-a9ba-4c8a-9ac8-1acd7d08951e

```java
@EventHandler
public void onInventoryClick(final InventoryClickEvent event) {
    if (!event.getClickedInventory().equals(event.getWhoClicked().getInventory())) {
        return;
    }

    event.setCancelled(true);
    event.getWhoClicked().dropItem(event.getSlot());
}
```

NonSwag:
hmm i see

Strokkur424:
Oops, the patch has a problem. Give me a second to fix it

patches/api/0501-Extend-HumanEntity-dropItem-API.patch
```diff
@@ -27,27 +27,27 @@ index 488604ba1a516b477693877c74712e4a45624a8b..2aab711ece14f008f2c4bd1b46d4e48c
 +    /**
 +     * Makes the entity drop an item from their inventory.
 +     * <br>
-+     * This method calls {@link HumanEntity#dropItem(int slot, boolean retainOwnership)}
++     * This method calls {@link HumanEntity#dropItem(int slot, boolean makeThrower)}
 +     * with the first {@link ItemStack} occurrence in the inventory
 +     *
 +     * @param itemStack The ItemStack to drop from their inventory
-+     * @param retainOwnership Whether this player should own the resulting item entity
++     * @param makeThrower Whether the resulting {@link Item} should have its thrower set to this {@link HumanEntity}
```
NonSwag: maybe "persistThrower" would be a more fitting name

Strokkur424: Changed!

patches/server/1073-Extend-HumanEntity-dropItem-API.patch
Strokkur424: Also just to note: As I am using this method when dropping via ItemStack, this would also throw an exception each time the ItemStack is not found, which seems counterproductive. So same case as the other one

NonSwag: In that case of the slot is outside the inventory bounds we should definitely throw

Strokkur424: The issue is:
1. I completely agree with you on user input. If you put `137` as the slot, is should throw an `IndexOutOfBoundsException`
2. But if they put in an ItemStack that hasn't been found, an `IndexOutOfBoundsException` feels weird

Strokkur424: Maybe I can add a check inside the ItemStack overload one to check if the ItemStack exists, and only then pass the slot in?

Strokkur424: Yeah okay so I made it now that if you pass an incorrect slot in, it throws an `IndexOutOfBoundsException`, like this:
```
[20:14:39 ERROR]: Command exception: /drop -1                                                                                                                                                           
java.lang.IndexOutOfBoundsException: Slot -1 out of range for inventory of size 41                                                                                                                      
        at org.bukkit.craftbukkit.entity.CraftHumanEntity.dropItem(CraftHumanEntity.java:839) ~[main/:?]
        at org.bukkit.craftbukkit.entity.CraftHumanEntity.dropItem(CraftHumanEntity.java:832) ~[main/:?]
        at test-plugin-1.0.0-SNAPSHOT.jar/io.papermc.testplugin.TestPlugin.logic(TestPlugin.java:29) ~[test-plugin-1.0.0-SNAPSHOT.jar:?]
        at test-plugin-1.0.0-SNAPSHOT.jar/io.papermc.testplugin.TestPlugin$1.execute(TestPlugin.java:21) ~[test-plugin-1.0.0-SNAPSHOT.jar:?]
        at io.papermc.paper.command.brigadier.bukkit.BukkitCommandNode$BukkitBrigCommand.run(BukkitCommandNode.java:82) ~[main/:?]
        at com.mojang.brigadier.context.ContextChain.runExecutable(ContextChain.java:73) ~[brigadier-1.3.10.jar:?]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:31) ~[filteredMinecraft.jar:?]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:19) ~[filteredMinecraft.jar:?]
        at net.minecraft.commands.execution.UnboundEntryAction.lambda$bind$0(UnboundEntryAction.java:8) ~[filteredMinecraft.jar:?]
        at net.minecraft.commands.execution.CommandQueueEntry.execute(CommandQueueEntry.java:5) ~[filteredMinecraft.jar:?]
        at net.minecraft.commands.execution.ExecutionContext.runCommandQueue(ExecutionContext.java:103) ~[filteredMinecraft.jar:?]
        at net.minecraft.commands.Commands.executeCommandInContext(Commands.java:448) ~[main/:?]
        at net.minecraft.commands.Commands.performCommand(Commands.java:355) ~[main/:?]
        at net.minecraft.commands.Commands.performCommand(Commands.java:342) ~[main/:?]
        at net.minecraft.commands.Commands.performCommand(Commands.java:337) ~[main/:?]
        at net.minecraft.server.network.ServerGamePacketListenerImpl.performUnsignedChatCommand(ServerGamePacketListenerImpl.java:2241) ~[main/:?]
        at net.minecraft.server.network.ServerGamePacketListenerImpl.lambda$handleChatCommand$11(ServerGamePacketListenerImpl.java:2215) ~[main/:?]
        at net.minecraft.server.TickTask.run(TickTask.java:18) ~[filteredMinecraft.jar:?]
        at net.minecraft.util.thread.BlockableEventLoop.doRunTask(BlockableEventLoop.java:154) ~[main/:?]
        at net.minecraft.util.thread.ReentrantBlockableEventLoop.doRunTask(ReentrantBlockableEventLoop.java:24) ~[filteredMinecraft.jar:?]
        at net.minecraft.server.MinecraftServer.doRunTask(MinecraftServer.java:1577) ~[main/:?]
        at net.minecraft.server.MinecraftServer.doRunTask(MinecraftServer.java:207) ~[main/:?]
        at net.minecraft.util.thread.BlockableEventLoop.pollTask(BlockableEventLoop.java:128) ~[main/:?]
        at net.minecraft.server.MinecraftServer.pollTaskInternal(MinecraftServer.java:1554) ~[main/:?]
        at net.minecraft.server.MinecraftServer.pollTask(MinecraftServer.java:1547) ~[main/:?]
        at net.minecraft.util.thread.BlockableEventLoop.managedBlock(BlockableEventLoop.java:138) ~[main/:?]
        at net.minecraft.server.MinecraftServer.managedBlock(MinecraftServer.java:1498) ~[main/:?]
        at net.minecraft.server.MinecraftServer.waitUntilNextTick(MinecraftServer.java:1508) ~[main/:?]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1344) ~[main/:?]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:340) ~[main/:?]
        at java.base/java.lang.Thread.run(Thread.java:1583) ~[?:?]
```

NonSwag (approved):

NonSwag:
Now let's wait for a maintainer for review

Malfrador:
Instead of the `persistThrower` boolean, I would find an optional `UUID` argument allowing you to directly _set_ the thrower more useful. Especially given that method already exists on the resulting `Item` entity anyways. 
Also some Javadoc clarification on what `throwRandomly` does surely would be helpful, without looking at the NMS code I have no clue what that would do.

Strokkur424:
Closing pre softspoon PR in favour of new one

PR #11691, opened 2024-12-01
Add "cancel on error" option for event listeners

NewwindServer:
Adds the option to cancel an event if an error is thrown inside its handler
useful for mission critical listeners, for example, preventing players from breaking blocks in an area
or preventing players from modifying certain items inside an anvil.

Of course, ideally, your plugin shouldn't be throwing errors in listeners, but mistakes can happen, this feature adds a way for plugin authors to have a fail-safe for important listeners, for example:

Before:
```java
@EventHandler(ignoreCancelled = true)
    public void onBlockPlace(BlockPlaceEvent event) {
        try {
            if (isSpecialItem(event.getItemInHand())) {
                event.setCancelled(true);
            } else if (isProtectedArea(event.getBlock().getLocation())) {
                event.getPlayer().sendMessage(getLocalized("cant.use.here")); // Maybe at some point this will fail
                event.setCancelled(true);
            }
        } catch (Throwable throwable) { // fail safe
            event.setCancelled(true);
        }
    }
```
After:
```java
@EventHandler(ignoreCancelled = true, cancelOnError = true)
    public void onBlockPlace(BlockPlaceEvent event) {
        if (isSpecialItem(event.getItemInHand())) {
            event.setCancelled(true);
        } else if (isProtectedArea(event.getBlock().getLocation())) {
            event.getPlayer().sendMessage(getLocalized("cant.use.here")); // Maybe at some point this will fail
            event.setCancelled(true);
        }
    }
```

Also, I didn't bother adding it for TimedEventListener since those are all behind `if (false)` statements and timings is about to be removed.

### DIFF
--- a/patches/api/0501-Add-cancel-on-error-option-for-events.patch
+++ b/patches/api/0501-Add-cancel-on-error-option-for-events.patch
@@ -1,0 +1,183 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Newwind <newwindserver@gmail.com>
+Date: Sat, 30 Nov 2024 16:03:37 +0200
+Subject: [PATCH] Add cancel on error option for event listeners
+
+Add the option to cancel an event if an error is thrown inside its handler
+useful for mission critical listeners, for example, preventing players from breaking blocks in an area
+or preventing players from modifying certain items inside an anvil
+
+diff --git a/src/main/java/org/bukkit/event/EventHandler.java b/src/main/java/org/bukkit/event/EventHandler.java
+index cc06f480..dd0ac8a5 100644
+--- a/src/main/java/org/bukkit/event/EventHandler.java
++++ b/src/main/java/org/bukkit/event/EventHandler.java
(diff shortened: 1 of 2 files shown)


lynxplay:
Somewhat meh on the concept, not the biggest fan nor hater.
If people are coding such critical stuff, they should be able to just wrap in try-catch.
Like, I just don't see how often this would actually be used in the wild? Over just try-catch, which also gives you proper control of any potential exceptions.

Machine-Maker:
Yeah, I think people that want this should just be more explicit about it and just use try-catch. That's always going to be more flexible than a boolean on the annotation. Like what if I want it to cancel on exceptions **unless** some condition is met. Using code is more flexible

mbax:
If you're going to add an annotation that requires manually adding it, whose only purpose is to effectively be `try{/*code*/}catch(Exception ignored){event.setCancelled(true);}`, why aren't you just try-catching? The only way this becomes a more-useful feature would be if it were a **breaking** change to enable it by default. (And I also don't like that idea, to be clear.)

PR #11695, opened 2024-12-01
Only attempt to resync display entities' position/rotation when it has actually changed to fix teleport interpolation issues

### LINKED ISSUE #11694: Resyncing entity positions' every 60 ticks breaks entity display's `teleport_duration` interpolation if the position gets sync during a interpolation
### Expected behavior

When spawning a display entity, delaying the teleport a bit (enough that it triggers the periodic pos/rot packet every 60 ticks lived), and then it with a set `teleport_duration`, the teleportation should work fine with a specific interpolation set.

By the way, THIS IS NOT A PAPER BUG, THIS IS A VANILLA BUG!!! ...but it would be nice if it is possible to fix this on Paper. :3

### Observed/Actual behavior

The entity gets slow when it gets to the end of the interpolation.

https://github.com/user-attachments/assets/61b9d7b9-7074-444b-8c6f-afb5dd11b710

### Steps/models to reproduce

1. Spawn a display entity
2. Set the `teleportDuration` to 59 ticks
3. Wait 8 ticks (enough that it goes over the 60 tick position sync)
4. Teleport the display entity
5. Watch the entity

```kotlin
                    val cameraEntity = player.world.spawn(
                        originalCamera,
                        TextDisplay::class.java
                    ) {
                        it.teleportDuration = 59
                    }

                    Bukkit.broadcastMessage("Camera Entity: ${cameraEntity.entityId}")

                    cameraEntity.text(textComponent {
                        content("hello")
                    })

                    Bukkit.getScheduler().runTaskLater(
                        m,
                        Runnable {
                            cameraEntity.teleport(originalCamera)

                            Bukkit.getScheduler().runTaskLater(
                                m,
                                Runnable {
                                    cameraEntity.teleport(targetCamera)
                                },
                                4L
                            )
                        },
                        4L
                    )
```

### Plugin and Datapack List

Not applicable

### Paper version

```
This server is running SparklyPaper version 1.21.3-DEV-ver/1.21.3@6921e60 (2024-12-01T03:37:24Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
```

(something something it is my own fork and stuff like that)

### Other

This happens due to `boolean flag2 = flag1 || this.tickCount % 60 == 0;`, because it attempts to resync the position every 60 ticks, however I'm not sure how could that be fixed without tracking when the entity is interpolating on the client side. (which, due to circumstances, the entity may not be interpolating on the client when the `teleport_duration` is set, example: if you set the teleport duration and teleport the entity on the same tick)

Now, I must say that there is a workaround for this issue... and the workaround is "create a new entity display every 60 ticks".

This DOES NOT HAPPEN if you manually send the add entity + set data + teleport packets

```kotlin
                        val add = ClientboundAddEntityPacket(
                            newId,
                            UUID.randomUUID(),
                            -28.5,
                            108.0,
                            -58.5,
                            0f,
                            0f,
                            net.minecraft.world.entity.EntityType.TEXT_DISPLAY,
                            0,
                            Vec3.ZERO,
                            0.0
                        )

                        Bukkit.broadcastMessage(net.minecraft.world.entity.Display.DATA_POS_ROT_INTERPOLATION_DURATION_ID.id.toString())

                        val metadata = ClientboundSetEntityDataPacket(
                            newId,
                            listOf(
                                SynchedEntityData.DataValue(net.minecraft.world.entity.Display.DATA_POS_ROT_INTERPOLATION_DURATION_ID.id, EntityDataSerializers.INT, 59),
                                SynchedEntityData.DataValue(
                                    23,
                                    EntityDataSerializers.COMPONENT,
                                    io.papermc.paper.adventure.PaperAdventure.asVanilla(
                                        textComponent("test packet") {}
                                    )
                                )
                            )
                        )

                        val nmsPlayer = (player as CraftPlayer).handle
                        nmsPlayer.connection.sendPacket(add)
                        nmsPlayer.connection.sendPacket(metadata)

                        Bukkit.getScheduler().runTaskLater(
                            m,
                            Runnable {
                                Bukkit.getScheduler().runTaskLater(
                                    m,
                                    Runnable {
                                        nmsPlayer.connection.sendPacket(
                                            ClientboundTeleportEntityPacket(
                                                newId,
                                                PositionMoveRotation(
                                                    Vec3(
                                                        targetCamera.x,
                                                        targetCamera.y,
                                                        targetCamera.z,
                                                    ),
                                                    Vec3(
                                                        targetCamera.x,
                                                        targetCamera.y,
                                                        targetCamera.z,
                                                    ),
                                                    0.0f,
                                                    0.0f
                                                ),
                                                setOf(),
                                                false
                                            )
                                        )

                                        Bukkit.getScheduler().runTaskLater(
                                            m,
                                            Runnable {
                                                nmsPlayer.connection.sendPacket(
                                                    ClientboundTeleportEntityPacket(
                                                        newId,
                                                        PositionMoveRotation(
                                                            Vec3(
                                                                targetCamera.x + 10.0,
                                                                targetCamera.y,
                                                                targetCamera.z,
                                                            ),
                                                            Vec3(
                                                                targetCamera.x + 10.0,
                                                                targetCamera.y,
                                                                targetCamera.z,
                                                            ),
                                                            0.0f,
                                                            0.0f
                                                        ),
                                                        setOf(),
                                                        false
                                                    )
                                                )
                                            },
                                            59L
                                        )
                                    },
                                    4L
                                )
                            },
                            4L
                        )
```

MrPowerGamerBR:
On display entities, any new Pos/Rot packet causes the teleport duration interpolation to reset

While this does not cause issues when you spawn a display entity -> set the interpolation -> teleport after the data has been sent to the client, this DOES cause issues on entities that have lived for >60 ticks, where if you spawn a display with interpolation -> wait ~8 ticks -> teleport it, the end of the teleport will look slow due to the server resyncing the display entity position/rotation

HOWEVER I will be honest, I'm not sure what are the side effects of not periodically resyncing entity positions, this could be a Very Bad Idea:tm:

I must say that there is a workaround for this issue... and the workaround is "create a new entity display every 60 ticks". So this PR is mostly to get a "is this ACTUALLY a good idea, or is this ACTUALLY a bug that Paper must fix, or should plugin devs manually workaround this issue?" discussion going. If this is something that plugin devs should manually workaround, I think that it would be a good idea to document that behavior on Paper’s docs on the website ([which has a section dedicated to displays](https://docs.papermc.io/paper/dev/display-entities)) because wow, the only place that somewhat touches on this issue is in the [snapshot 23w31a changelog on Minecraft's website](https://www.minecraft.net/en-us/article/minecraft-snapshot-23w31a) and even then they do not go in depth, because they said that they clamp the `teleport_duration` value to 59 because "this value is clamped to avoid glitches due to periodic position updates"... but that doesn't fix the issue at all?!?!

I thought about an alternative implementation that tracks when was the last teleport + `teleport_duration`, but that would be way more complex than the current implementation, and, due to circumstances, the entity may not be interpolating on the client when the teleport_duration is set, example: if you set the teleport duration and teleport the entity on the same tick, the teleport_duration would be set but the client would not interpolate it, because you need to force a entity refresh to the client to cause it to interpolate on the same tick.

Fixes #11694

This also has the cool side effect that you can actually set teleport duration of display values above >59 ticks and the client interpolates it without any issues. (`(cameraEntity as CraftTextDisplay).handle.entityData.set(net.minecraft.world.entity.Display.DATA_POS_ROT_INTERPOLATION_DURATION_ID, 120)`)

### DIFF
--- a/patches/server/1073-Only-attempt-to-resync-display-entities-position-whe.patch
+++ b/patches/server/1073-Only-attempt-to-resync-display-entities-position-whe.patch
@@ -1,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MrPowerGamerBR <git@mrpowergamerbr.com>
+Date: Sun, 1 Dec 2024 01:23:17 -0300
+Subject: [PATCH] Only attempt to resync display entities' position when it has
+ actually changed to fix teleport interpolation issues
+
+On display entities, any new Pos/Rot packet causes the teleport duration interpolation to reset
+
+While this does not cause issues when you spawn a display entity -> set the interpolation -> teleport after the data has been sent to the client, this DOES cause issues on entities that have lived for >60 ticks, where if you spawn a display with interpolation -> wait ~8 ticks -> teleport it, the end of the teleport will look slow due to the server resyncing the display entity position/rotation
+
+diff --git a/src/main/java/net/minecraft/server/level/ServerEntity.java b/src/main/java/net/minecraft/server/level/ServerEntity.java
+index 90eb4927fa51ce3df86aa7b6c71f49150a03e337..cb8f9a0182000fbf671f9ddccfd6b4999abf184b 100644
+--- a/src/main/java/net/minecraft/server/level/ServerEntity.java
++++ b/src/main/java/net/minecraft/server/level/ServerEntity.java
+@@ -208,6 +208,7 @@ public class ServerEntity {
+ 
+                     if (!this.forceStateResync && !flag5 && this.teleportDelay <= 400 && !this.wasRiding && this.wasOnGround == this.entity.onGround()) { // Paper - fix desync when a player is added to the tracker
+                         if ((!flag2 || !flag) && !(this.entity instanceof AbstractArrow)) {
++                            if (!(this.entity instanceof net.minecraft.world.entity.Display) || (flag || flag1)) { // Paper start - Only attempt to resync display entities' position when it has actually changed to fix teleport interpolation issues
+                             if (flag2) {
+                                 packet1 = new ClientboundMoveEntityPacket.Pos(this.entity.getId(), (short) ((int) i), (short) ((int) j), (short) ((int) k), this.entity.onGround());
+                                 flag3 = true;
+@@ -215,6 +216,7 @@ public class ServerEntity {
+                                 packet1 = new ClientboundMoveEntityPacket.Rot(this.entity.getId(), b0, b1, this.entity.onGround());
+                                 flag4 = true;
+                             }
++                            } // Paper end
+                         } else {
+                             packet1 = new ClientboundMoveEntityPacket.PosRot(this.entity.getId(), (short) ((int) i), (short) ((int) j), (short) ((int) k), b0, b1, this.entity.onGround());
+                             flag3 = true;


Owen1212055:
Can we potentially do this ONLY when a display entity is currently interpolating?

MrPowerGamerBR:
> Can we potentially do this ONLY when a display entity is currently interpolating?

It would be possible, but it is a bit more complex...

> I thought about an alternative implementation that tracks when was the last teleport + teleport_duration, but that would be way more complex than the current implementation, and, due to circumstances, the entity may not be interpolating on the client when the teleport_duration is set, example: if you set the teleport duration and teleport the entity on the same tick, the teleport_duration would be set but the client would not interpolate it, because you need to force a entity refresh to the client to cause it to interpolate on the same tick.

What I'm doing right now is using packets to create the text display & stuff, so technically this PR is not need, *but* I do think that an warning should be added to the https://docs.papermc.io/paper/dev/display-entities page because I spent an entire day pulling my hair out, only to find out that this is a vanilla issue.

Owen1212055:
Is there an MC bug report that explains this?

MrPowerGamerBR:
Scratch that, I found a bug report: https://bugs.mojang.com/browse/MC-272599?jql=text%20~%20%22teleport_duration%22

Owen1212055:
@MrPowerGamerBR 
> It would be possible, but it is a bit more complex..
How is it more complex? Imo, what I would do is expose the `posRotInterpolationTarget` on Display. check if its != null (meaning they are interpolating), and then don't resync the position.

MrPowerGamerBR:
> How is it more complex? Imo, what I would do is expose the `posRotInterpolationTarget` on Display. check if its != null (meaning they are interpolating), and then don't resync the position.

On the server the `posRotInterpolationTarget` is always null. The only code that sets it is the Display `lerpTo` function (which is not called on the server), and the code that ticks the `posRotInterpolationTarget` is behind a `if (this.level().isClientSide) {` clause. 

However I think it wouldn't be hard to make the server call those functions tho... But here's another issue:

Another issue is that this:

```kotlin
display.setTeleportDuration(59)
display.teleport(...)
```

Does NOT work as how you would expect. What I expected is that the display entity will teleport to the destination over 59 ticks. In reality the entity is teleported before the entity is updated on the client, so the display entity will teleport directly to the new location without any easing.

And how would you handle that on the server? You would need to track...

1. Did the display entity teleport?
2. Does it have a teleport duration?
3. Did the client already receive the updated teleport duration BEFORE the teleport happened?
4. Is the display entity currently easing?

If everything is true, then the teleport position update should be ignored.

Owen1212055:
Ah yeah this is in general kinda scuffed, then your solution is prolly fine. Although I would say that this should have the mojira issue linked in the PR.

PR #11710, opened 2024-12-05
Add ServerChangeMaxPlayerSlotsEvent

ammodev:
This pull request introduces a new event to handle changes in the maximum number of player slots on the server. The changes include the addition of a new event class and its integration into the server code.

### Event Addition:

* [`patches/api/0502-Add-ServerChangeMaxPlayerSlotsEvent.patch`](diffhunk://#diff-b73ef3a414bef72d2cc05c05d8f005f103b7c4aa770fb672a141127b6136f9ceR1-R93): Added a new event class `ServerChangeMaxPlayerSlotsEvent` to handle changes in the maximum number of player slots. This class includes methods to get and set the old and new maximum player slots, as well as to handle event cancellation.

### Server Integration:

* [`patches/server/1073-Add-ServerChangeMaxPlayerSlotsEvent.patch`](diffhunk://#diff-1a861686eefd12089c93250b2626ea6714848b2de830ec9a2896d22a478ef4caR1-R25): Integrated the `ServerChangeMaxPlayerSlotsEvent` into the `CraftServer` class. The event is called when the maximum number of player slots is set, allowing for the new maximum to be adjusted or the event to be cancelled.

### DIFF
--- a/patches/api/0502-Add-ServerChangeMaxPlayerSlotsEvent.patch
+++ b/patches/api/0502-Add-ServerChangeMaxPlayerSlotsEvent.patch
@@ -1,0 +1,93 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Ammo <ammo@slne.dev>
+Date: Thu, 5 Dec 2024 19:46:42 +0100
+Subject: [PATCH] Add ServerChangeMaxPlayerSlotsEvent
+
+
+diff --git a/src/main/java/io/papermc/paper/event/server/ServerChangeMaxPlayerSlotsEvent.java b/src/main/java/io/papermc/paper/event/server/ServerChangeMaxPlayerSlotsEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..081842250078e9037016afcf0c4faab9db9ada54
+--- /dev/null
++++ b/src/main/java/io/papermc/paper/event/server/ServerChangeMaxPlayerSlotsEvent.java
+@@ -0,0 +1,81 @@
++package io.papermc.paper.event.server;
++
++import com.google.common.base.Preconditions;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.server.ServerEvent;
++import org.jetbrains.annotations.Range;
++import org.jspecify.annotations.NullMarked;
++
++/**
++ * Called when the max player slots of the server are changed.
++ */
++@NullMarked
++public class ServerChangeMaxPlayerSlotsEvent extends ServerEvent implements Cancellable {
++
++    public static final HandlerList HANDLER_LIST = new HandlerList();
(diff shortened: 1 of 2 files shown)


lynxplay:
Hi and welcome to paper 🥳 

Thank you for your first PR.
I am a bit sceptical on the usecase for this.

We generally do not provide API for plugin calls as it makes interoperability between plugins a bit iffy if a plugin cannot rely on its API calls being executed.
In what sense does an event like this help a plugin?

ammodev:
Hi! What are you referring to? Are you referring the cancelability / setMaxPlayers in the event or are you concerned about the event in general?

lynxplay:
Yea I am concerned about the event in general.

Regarding a player queue, the max players are only ever updated this way by other plugins.
Given whatever *other* service needs to be informed here, something would need to actually send that message.
That would mean we'd have a plugin A that does change this number but is not self developed, then plugin B that listens to that event and then notifies the external service about it?

I am unsure of how many plugins I know that change the max player count like this, especially in the context of server networks, where this kind of logic is probably better of at the queue level itself, rather than a server deciding its own max player limit.

ammodev:
So you are saying that we would have to recode every plugin or ask for a changeevent on their side to provide this functionallity? I mean really the other choice is to send a request from the velocity to the server every tick in order to fetch this state change?

lynxplay:
Well no, I am not saying anything, I was just *asking* what kind of third party plugins exist that change the max player numbers like this on their own without being prompted to do so. It sounds like a rather destructive way of managing a full server on the plugins part, so I am just curious.
Do you have a specific plugin for this that I can glance at?

I am mostly asking because, as stated prior, we generally are not a fan of events for this as plugins expect their API calls to *work*.
A third party plugin that for some reason manages the entire servers max player count like this, might not be to pleased when the value is doubled by another plugin instead.

ammodev:
https://www.spigotmc.org/resources/changeslots-change-max-players-slots-bukkit-bungee.49648/ - Easy Plugin - Could do this on our own with no issue

This plugin is the main issue we are currently having, resulting in us trying to have this event included. Other than that you could take any gamemode (BedWars, SkyWars, etc.). Some of them change the playercount when the plugin is loaded resulting in the velocity not knowing the max player count anymore.

electronicboy:
That plugin doesn't even modify the slots in a manner that this event would fire;

It's long been tradition that, outside of some very key explicit areas, events are not fired; if a plugin does something, that operation occurs. I can maybe understand notification-esque events for some of this stuff, but, not fond of aspects like cancellation

lynxplay:
Well the reason I am asking so much for a usecase is that, for API we are generally not happy to include due to the mentioned reasons), we'd need a really good usecase to justify that.
If the plugin you were struggling with does not even use this in the first place, it just isn't a very good argument to merge something that breaks the existing API concepts.

As cat stated, a notification-like event should solve your usecase? (e.g. no setters or cancellation)

PR #11714, opened 2024-12-06
Re-add shear methods from bukkit Shearable

### LINKED ISSUE #11638: Sheep/Bogged/Snowman "regrow wool/mushroom/pumpkin" methods

With the deprecation of the bukkit `Shearable` API, we lost the ability to force regrow wool or mushrooms for sheep and bogged or "un-derp" snow golems

Methods for those entities to regrow their "features"

Using the deprecated and marked for removal API from Bukkit

### Other

_No response_

HaHaWTH:
This pull request closes #11638.

### Changes made

Readd shear related methods to Sheep and Bogged from bukkit Shearable.

This is my first PR to Paper, sorry if i did something wrong.

### DIFF
--- a/patches/api/0389-Add-Shearable-API.patch
+++ b/patches/api/0389-Add-Shearable-API.patch
@@ -55,15 +55,30 @@ index 0000000000000000000000000000000000000000..3991f2aa530c588f52e1f596d3b03743
 +    boolean readyToBeSheared();
 +}
 diff --git a/src/main/java/org/bukkit/entity/Bogged.java b/src/main/java/org/bukkit/entity/Bogged.java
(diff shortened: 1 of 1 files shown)


electronicboy:
We should expose what "sheared" represents, rather then just expanding the usage of flawed API; (and that should be added to the existing entity API patch)

HaHaWTH:
> We should expose what "sheared" represents, rather then just expanding the usage of flawed API; (and that should be added to the existing entity API patch)

Thanks for explanation! Should I add this to Missing Entity API patch?

electronicboy:
Yes

electronicboy:
Oh, actually, both of them called the internal data shreared, so, this is fine; Snowmen already have their "derp"'d methods

PR #11716, opened 2024-12-06
Call PlayerItemDamageEvent for tridents

### LINKED ISSUE #11655: Riptiding does not call PlayerItemDamageEvent
### Expected behavior

When player riptides using Riptide enchanted trident, PlayerItemDamageEvent should be called (and its cancellation should prevent damaging the item)

### Observed/Actual behavior

The PlayerItemDamageEvent is not called when riptiding.

### Steps/models to reproduce

- get yourself Riptide enchanted Trident
- try to cancel PlayerItemDamageEvent using a plugin
- the event won't be called and trident will get damage

### Plugin and Datapack List

only the test plugin

### Paper version

[00:09:43 INFO]: This server is running Paper version 1.21.3-56-master@bb32b05 (2024-11-23T22:23:12Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
Previous version: 1.21.3-46-f8e2a67 (MC: 1.21.3)

### Other

_No response_

notTamion:
fixes #11655 
It also previously wasn't being called for the trident being thrown

post hardfork #11899

### DIFF
--- a/patches/server/0680-Expand-PlayerItemDamageEvent.patch
+++ b/patches/server/0680-Expand-PlayerItemDamageEvent.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Expand PlayerItemDamageEvent
 
 
 diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
-index 18f84d54ec72debec652adb22067e11aa058b238..679112d1622744680b7d3c7d296acafac019d00a 100644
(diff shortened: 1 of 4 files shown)


kennytv (approved):

Machine-Maker:
Shouldn't this then mimic the behavior of https://github.com/PaperMC/Paper/pull/11555 where it's not called for non-positive damage. Then smth else needs to be added for "durability change" or something.

notTamion:
We should probably allow the item to break if the damage was changed using the API.

PR #11739, opened 2024-12-13
Support lowercase in Material.getMaterial()

dertfin0:
It's very inconvenient that `Material.getMaterial()` requires capital letters. This can be confusing for newbies. 

This change will allow you to pass the name of Material as arguments in any case


FireInstall:
Even if the Material enum wasn't on it's way out, the problem you are tying to solve was solved long ago: Material#matchMaterial()

PR #11770, opened 2024-12-22
Add effective name to ItemStack

masmc05:
Original https://github.com/PaperMC/Paper/pull/11469

Adds a new method which returns the name of an itemstack which the player sees in their inventory. Takes into account any data components exactly as vanilla.

Unlike [ItemStack#displayName](https://jd.papermc.io/paper/1.21.1/org/bukkit/inventory/ItemStack.html#displayName()), does not add additional brackets etc


paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
```diff
@@ -467,6 +467,13 @@ void skullCallback(final net.minecraft.world.item.component.ResolvableProfile re
         return true;
     }
 
+    // Paper start - add ItemStack#effectiveName
```
kennytv: No comments needed

kennytv (approved):

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
```diff
@@ -467,6 +467,11 @@ void skullCallback(final net.minecraft.world.item.component.ResolvableProfile re
         return true;
     }
 
+    @Override
+    public net.kyori.adventure.text.Component effectiveName() {
+        return this.handle == null ? net.kyori.adventure.text.Component.empty() : io.papermc.paper.adventure.PaperAdventure.asAdventure(this.handle.getStyledHoverName());
```
masmc05: Now with hard fork, maybe can I do imports here? Since it's not a patch

Leguan16: Yes

Machine-Maker (approved):

PR #11771, opened 2024-12-22
Add sinces to javadocs

masmc05:
Adds since tags for api added since bukkit 1.0 to latest 1.21.4 api published to maven repo.

Noticed https://github.com/orgs/PaperMC/projects/6?pane=issue&itemId=57550283 and decided to put this, as now with hard fork this isn't a 300 000 lines patch anymore


Machine-Maker:
We also want to skip adding them for `Deprecated(forRemoval = true)` classes and methods. If a class is marked with that, skip it and everything in it. If a method/field is, just skip that.

We can also skip anything annotated with `@ApiStatus.Internal`, following the same rules as above.

kennytv (dismissed):

kennytv:
Could maybe also remove versions below 1.8 or 1.7... especially ones like 1.1.0 R4 are a bit weird given it's two projects down the line where Paper didn't exist yet lol

Machine-Maker (approved):
I don't think we have to strip out the really old versions, I think its pretty cool to have them.

kennytv:
Sorry for turning this around now, but internal discussion has shown that doing the scanning in a "safe" way we want here and then also enforcing it properly for all future PRs is pretty involved and intrusive to some party in every reasonable way that was suggested.

For the relatively small benefit that it provides vs. just picking your min target version and us keeping track of it separately if needed, I don't feel comfortable adding a bunch of extra complexity/jankness onto either ourselves or our contributors in enforcing this later or regularly running scripts

We might revisit this a lot later when we're at a point with a cleaner code base and tools around it, with time for "extras", but for now I'll have to close it. Until then, thanks for the effort

MiniDigger:
similarly, but maybe more useful, I created https://minidigger.github.io/api-diff/ which automatically diffs the different paper api versions (plus can also generate a report of sinces)

https://github.com/MiniDigger/api-diff
https://minidigger.github.io/api-diff/since.html
https://minidigger.github.io/api-diff/diff-1.21.3-1.21.4.html

PR #11772, opened 2024-12-22
[ci skip] remove patch updating related tasks

NoahvdAa:
Title. We already changed the branch to main so I can't directly push :P


kennytv:
They'll likely be useful for Minecraft updates still

NoahvdAa:
fair enough; should probably at least be commented out like the other updating tasks then

PR #11775, opened 2024-12-23
Port basic lookAt methods from Player to Entity

### LINKED ISSUE #11754: Port "lookAt" API from Player to Entity

Currently, it is not possible to make entities look directly at other entities or locations without a bunch of math nobody really understands

Move the `lookAt` methods from the `Player` class to the `Entity` class

_Well... Math_

### Other

_No response_

Doc94:
Exists methods for lookAt available in the entity NMS class like mention #11754 this PR just move that methods from player to entity.

Close #11754

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Entity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Entity.java
@@ -150,6 +150,26 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
      * @return <code>true</code> if the teleport was successful
      */
     boolean teleport(@NotNull Location location, @NotNull TeleportCause cause, @NotNull io.papermc.paper.entity.TeleportFlag @NotNull... teleportFlags);
+
+    /**
+     * Causes the entity to look towards the given position.
+     *
+     * @param x x coordinate
+     * @param y y coordinate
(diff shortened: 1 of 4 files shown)


paper-api/src/main/java/org/bukkit/entity/Entity.java
```diff
@@ -150,6 +150,26 @@ default boolean teleport(@NotNull Location location, @NotNull io.papermc.paper.e
...
+
+    /**
+     * Causes the entity to look towards the given position.
+     *
+     * @param x x coordinate
+     * @param y y coordinate
+     * @param z z coordinate
+     * @param entityAnchor What part of the entity should face the given position
+     */
+    void lookAt(double x, double y, double z, @NotNull io.papermc.paper.entity.LookAnchor entityAnchor);
```
Machine-Maker: IMPORT 🎆

Doc94: IMPORTED

Doc94:
note: the other method i think can be moved just need figure how the hell call for avoid repeat so many times the word "entity" and not confuse me in the.. try

kennytv (approved):

NoahvdAa (approved):

PR #11779, opened 2024-12-23
Add removal reason to EntityRemoveFromWorldEvent

TonytheMacaroni:
Supercedes #10149.

### DIFF
--- a/paper-api/src/main/java/com/destroystokyo/paper/event/entity/EntityRemoveFromWorldEvent.java
+++ b/paper-api/src/main/java/com/destroystokyo/paper/event/entity/EntityRemoveFromWorldEvent.java
@@ -4,6 +4,7 @@ import org.bukkit.World;
 import org.bukkit.entity.Entity;
 import org.bukkit.event.HandlerList;
 import org.bukkit.event.entity.EntityEvent;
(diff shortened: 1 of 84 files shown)


kennytv:
Building fails - you can also now remove the // Paper comments inside of source files, and for those lines that had a CB comment only from adding the event cause, you can replace their comment with ours

TonytheMacaroni:
Re-added the missing test file diff; seems I missed it when updating the PR. Modified the comments.

Machine-Maker (changes requested):
For the `Cause` enum, I think the 2 booleans should be removed and just be made booleans on the event itself. I don't like tying those to the removal reason. What if a removal reason can be used in 2 places where those values should be different. I don't see much benefit tying them together instead of just putting them on the event itself.

lynxplay:
After *even more* discussion on this, apparently the intention of our RemoveFromWorldEvent is not the same as EntityRemoveEvent, specifically around the time they are called.

The bukkit remove event has a cause but may be called at times (especially around chunk unloads) where edits to the entity are not longer respected.

I'll open a follow up PR undeprecating the bukkit event with your as co-author given all the work you did here (and it was a lot of work) so all credits to you and sorry for the confusion on this change.

PR #11783, opened 2024-12-23
Add DamageType RegistryEvent

Chaosdave34:
Add  RegistryEvent for DamageType

Changes:
- Add DamageTypeRegistryEntry 
- Add PaperDamageTypeRegistryEntry
- Make DamageType PaperRegistry writeable
- Add DamageType Event to RegistryEvents

But I ran into the problem, that DamageEffect uses Bukkit.getUnsafe() for initialization and that seems to be not available at bootstrap. I fixed that with the following changes, but I am not sure if that's the correct way to do is as it may break plugins: 
- Turn DamageType interface into an enum
- Add CraftDamageType damageEffectToBukkit() and  damageEffectToNMS() to CraftDamageType
- Add getSoundForDamageEffect() to UnsafeValues (and CraftMagicNumbers)
- Deprecated CraftDamageEffect, UnsafeValues#getDamageEffect() and DamageEffect#getDamageEffect()

The only breaking change I see is when someone tries to cast DamageType to CraftDamageType which is no longer possible with my approach.


paper-api/src/main/java/org/bukkit/damage/DamageEffect.java
```diff
@@ -6,39 +6,41 @@
...
 import org.jetbrains.annotations.NotNull;
 
+// Paper start
 /**
  * Represents a type of effect that occurs when damage is inflicted. Currently,
  * effects only determine the sound that plays.
  */
 @ApiStatus.Experimental
-public interface DamageEffect {
+public enum DamageEffect {
```
masmc05: Why downgrade a perfectly working interface to enum? Not only this brings a lot of issues that both spigot and paper are fighting right now, but it breaks plugins without any added rewriting in this pr.

paper-api/src/main/java/org/bukkit/damage/DamageEffect.java
```diff
@@ -6,39 +6,41 @@
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
+// Paper start
```
masmc05: No more paper comments needed outside of patches

paper-api/src/main/java/org/bukkit/damage/DamageEffect.java
neospark3: Using an enum instead of an interface was the first idea that came to my mind. I know that it comes with the mentioned problems.

Maybe use the id of the damage effect as String instead of the Bukkit Interface for DamageTypeRegistryEntry?

masmc05: Yeah, either other form of reference, or trying to resolve it not using CraftServer.

I think a pretty valid solution could be providing another access to unsafe values other than trough craft server. There was already a discussion of making some static access method in a pr, including to get the data and protocol versions in bootstrap, but that died down. Basically a provider class, implementation of that, retrieve it with https://www.baeldung.com/java-spi and hold it in a local class within the static getter method in UnsafeValues class

lynxplay: Agree with that sentiment ^
Making this an enum is not something I'd like to see.
We'll end up with *some* version of unsafe values at bootstrap time down the line anyway so, *yea*.

Alternatively, we can *attempt* to predict that damage effects *might* be a registry value at some point?
Which would mean we *could* just move this to keys instead of the actual resolved value.
That would need a generated key class tho.
I'd personally be semi in favour of that approach as it means we won't have to break API there if that ever changes and it does get rid of the bootstrap issue. Will poke MM about it tho for his ideas.

neospark3: Maybe [this](https://github.com/PaperMC/Paper/pull/11783/commits/5570b7bac8cece5062d35f79d2a9fab2c7716b9b) could be a solution.

Additionally instead of having an extra class (StaticUnsafeValues), Bukkit's UnsafeValues interface could have a static getter.
The point where the provider is initialized is probably not the best, but it's loaded before the plugin bootstrapper runs.

Machine-Maker: I want a separate type called something like “APIBridge” that uses the service loader API for stuff like this, so we can move actual API out of UnsafeValues. UnsafeValues should be for actual “unsafe” api, not just a place to store “bridge” methods. 

But either way, yeah, this shouldn’t be changed back to an enum.

paper-api/src/main/java/io/papermc/paper/registry/data/DamageTypeRegistryEntry.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return part of the translation key
+     */
+    String messageId();
+
+    /**
+     * Provides the amount of hunger exhaustion caused by this damage type.
+     *
+     * @return the exhaustion
+     */
+    Float exhaustion();
```
Machine-Maker: If this isn't nullable, shouldn't it be a primitive float?

paper-api/src/main/java/io/papermc/paper/registry/data/DamageTypeRegistryEntry.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    DamageScaling damageScaling();
+
+    /**
+     * Provides the {@link DamageEffect} for this damage type.
+     *
+     * @return the damage effect
+     */
+    @Nullable
+    DamageEffect damageEffect();
```
Machine-Maker: Nullable is annotating the type, not the method so it should be on the same line (`@Nullable DamageEffect damageEffect()`)

Machine-Maker (changes requested):

paper-server/src/main/java/io/papermc/paper/registry/data/PaperDamageTypeRegistryEntry.java
```diff
@@ -0,0 +1,121 @@
...
+import org.bukkit.craftbukkit.damage.CraftDamageEffect;
+import org.bukkit.craftbukkit.damage.CraftDamageType;
+import org.bukkit.damage.DamageEffect;
+import org.jspecify.annotations.Nullable;
+
+import static io.papermc.paper.registry.data.util.Checks.asConfigured;
+
+public class PaperDamageTypeRegistryEntry implements DamageTypeRegistryEntry {
+    protected @Nullable String messageId;
+    protected float exhaustion;
```
Machine-Maker: This should be nullable, and then the getter and `build` method should check that it's been set. It's not an optional field, so we want people to explicitly set it. Right now it defaults to 0.

paper-server/src/main/java/io/papermc/paper/registry/data/PaperDamageTypeRegistryEntry.java
```diff
@@ -0,0 +1,121 @@
...
+    protected @Nullable DamageScaling damageScaling;
+    protected @Nullable DamageEffects damageEffects;
+    protected @Nullable DeathMessageType deathMessageType;
```
Machine-Maker: I think these should be initialized to their defaults from the codec. That way you don't need to do the big complicated `build` method that chooses one of 3 constructors.

paper-api/src/main/java/io/papermc/paper/registry/data/DamageTypeRegistryEntry.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Provides the {@link DeathMessageType} for this damage type.
+     *
+     * @return the death message type
+     */
+    @Nullable DeathMessageType deathMessageType();
+
+    @ApiStatus.Experimental
+    @ApiStatus.NonExtendable
+    interface Builder extends DamageTypeRegistryEntry, RegistryBuilder<DamageType> {
```
Machine-Maker, line 68: The javadoc for the `Builder` interface should be formatted like the other registry entry builders. Take a look at enchantment or game event or painting variant to see the format.

Machine-Maker (changes requested):
As for the StaticUnsafeValues, I have a slightly different idea on how to solve it, but it can just be left in for now, not a big deal. 

There should be a standardized way to interact with the "pseudo-registries" types of which DamageEffect is part of. All the types on the server that implement StringRepresentable are a different sort of registry where there is a defined mapping between a string and an instance. There will be/already are other cases where a way to convert between them would be useful if we don't want to use an enum on the API side. 

But this isn't something you have to figure out, that'll be for us to decide on.

paper-server/src/main/java/io/papermc/paper/registry/data/PaperDamageTypeRegistryEntry.java
neospark3: Should the field, getter and setter for these objects no longer be nullable? Or should they be set to their defaults in the `build` method?

Machine-Maker: yeah, no longer be nullable since the field will be set in the constructor.

lynxplay (approved):

Machine-Maker (approved):

lynxplay:
Totally missed it on my first review, but welcome to paper :partying_face: 
Great first PR and thank you for your work!

Chaosdave34:
> Totally missed it on my first review, but welcome to paper 🥳 Great first PR and thank you for your work!

Tyvm.

Although you might missed this [one](https://github.com/PaperMC/Paper/pull/11669#pullrequestreview-2462354832).

lynxplay:
Well get congratulated twice then xD

PR #11786, opened 2024-12-23
Save players before disabling plugins

DerEchtePilz:
Original: https://github.com/PaperMC/Paper/pull/9679

While I still think that this PR is valid, I also thought about maybe changing the way it currently is in the `/restart` command since the goal is to sync the behaviour of shutting down the server and restarting the server when it comes to saving players/disabling plugins. I am open for input here.

### DIFF
--- a/paper-server/patches/features/0017-Moonrise-optimisation-patches.patch
+++ b/paper-server/patches/features/0017-Moonrise-optimisation-patches.patch
@@ -23552,7 +23552,7 @@ index 2b46ca9a2a046063cad422bec00d76107537b091..9aa664537cc37e44db46d5a2a64ae311
                  thread1 -> {
                      DedicatedServer dedicatedServer1 = new DedicatedServer(
 diff --git a/net/minecraft/server/MinecraftServer.java b/net/minecraft/server/MinecraftServer.java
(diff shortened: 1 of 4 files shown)


lynxplay:
To note that the same restrictions apply as before, even now that we are hardforked.
Will need to wait for the next release at the earliest.

lynxplay:
After some more thinking on this, I'll close this.
Generally post hardfork we'll try to use the fact that the PR queue was soft reset to keep it as low as possible.
This change (while I agree that it would be a nice thing *somewhere* down the line) is so unlikely to happen and if the team ever decides on actually implementing this would require a lot of community communication / announcements / planning etc to make sure plugin developers have time to actually adapt to this as we cannot slowly phase this behaviour out.

It would be a hard break in a single version.

As such, I'll close this PR with the idea of "if we ever feel like it is the right time for this to happen, we'll go look for this PR / implement it ourselfs and smack you in as a co-author, but for now, this PR adds no value to the PR queue as no one can merge it / review to make it mergable".

Hope that makes sense for you, thank you anyway for the PR, if we ever manage to have the communications in place to pull such a large breaking change, I'll make sure you are properly credited for your work here :+1:

DerEchtePilz:
Yeah, I understand. I honestly didn't even expect it to be merged even remotely soon despite hardfork being real now because of the nature of this PR.

PR #11788, opened 2024-12-23
Fix PotionMeta methods redirect

### LINKED ISSUE #11732: `PotionMeta#hasCustomName` compatability
### Expected behavior

For `PotionMeta#hasCustomName` to return whether the `PotionMeta` has a custom name.

### Observed/Actual behavior

`PotionMeta#hasCustomName` seems to call `ItemMeta#hasCustomName` instead of the `PotionMeta` deprecated redirect, I assume because the override in `CraftMetaItem` takes priority?

### Steps/models to reproduce

Wrote a test plugin:
```java
ItemStack item = player.getEquipment().getItemInMainHand();
PotionMeta meta = (PotionMeta) item.getItemMeta();
player.sendMessage(meta.hasCustomName() + "|||||" + meta.getCustomName());
try {
    player.sendMessage(meta.getClass().getMethod("hasCustomName").toString());
}
catch (NoSuchMethodException e) {
    throw new RuntimeException(e);
}
return true;
```
On latest Spigot:
Image
On latest Paper:
Image
Item used: `/give @s minecraft:potion[minecraft:potion_contents={custom_name:"shadow_walk"}]`

### Plugin and Datapack List

```
> plugins
[21:39:16 INFO]: Server Plugins (2):
[21:39:16 INFO]: Bukkit Plugins:
[21:39:16 INFO]:  - Debuggery, SpigotTestPlugin
```
```
> datapack list
[21:40:00 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[21:40:00 INFO]: There are no more data packs available
```

### Paper version

```
> version
[21:40:31 INFO]: Checking version, please wait...
[21:40:31 INFO]: This server is running Paper version 1.21.4-14-master@b746d9d (2024-12-11T21:22:39Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.3-36-d8b66dd (MC: 1.21.3)
```

### Other

_No response_

HaHaWTH:
This pull request fixes #11732.

### Api changes

Move deprecate methods redirect to CraftMetaPotion, make sure they have same behavior as Spigot. Java prior to call method implementation in classes first instead of default methods in interfaces

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaPotion.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaPotion.java
@@ -310,6 +310,23 @@ class CraftMetaPotion extends CraftMetaItem implements PotionMeta {
         return this.customName != null;
     }
 
+    // Paper start - Deprecated PotionMeta methods redirect
+    @Override
+    public boolean hasCustomName() {
+        return this.hasCustomPotionName();
+    }
+
+    @Override
+    public String getCustomName() {
+        return this.getCustomPotionName();
+    }
+
+    @Override
+    public void setCustomName(String name) {
+        this.setCustomPotionName(name);
+    }
+    // Paper end - Deprecated PotionMeta methods redirect
+
     @Override
     public String getCustomPotionName() {
         return this.customName;
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaPotion.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaPotion.java
@@ -310,7 +310,6 @@ class CraftMetaPotion extends CraftMetaItem implements PotionMeta {
         return this.customName != null;
     }
 
-    // Paper start - Deprecated PotionMeta methods redirect
     @Override
     public boolean hasCustomName() {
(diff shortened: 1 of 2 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftMetaPotion.java
```diff
@@ -310,6 +310,23 @@ public boolean hasCustomPotionName() {
         return this.customName != null;
     }
 
+    // Paper start - Deprecated PotionMeta methods redirect
```
lynxplay: No more paper comments.

HaHaWTH: Comments are removed.

kennytv:
The break here was semi intentional, the deprecated methods will be gone fairly soon

kennytv:
See above, I'll also backport the PotionMeta changes without the ItemMeta break to 1.21.2 and 1.21.3 to make adoption easier, but the updated hasCustomName behavior in 1.21.4 will remain as is, since that's also what it will have to be later anyways. the names from upstream were just really poorly chosen. the other two method changes have no effect, since they're not overwritten anywhere

PR #11789, opened 2024-12-23
Cache Bukkit Command when wrapping CommandNodes

### LINKED ISSUE #11378: Brigadier commands registered through `Commands#getDispatcher` require incorrect permission when executed by `Bukkit#dispatchCommand`

willkroboth:
Reopens #11385.

Resolves #11378 by "restoring" the Spigot behavior where VanillaCommandNodes are only created once. Before this commit, `BukkitBrigForwardingMap` would create a new `VanillaCommandWrapper` each time a `CommandNode` was requested via the Bukkit `CommandMap`. This meant that calls to `Command#setPermission` would not persist between retrievals from the map.

This allows command frameworks that insert `CommandNode`s directly into the Brigadier dispatcher to change the permission String of the `VanillaCommandNode`s created for their commands, rather than it always being the default `"minecraft.commands.<name>"`.

As discussed on the original PR, it's probably a good idea to eventually migrate the behavior of `Bukkit#dispatchCommand` to use Brigadier directly rather than working via the Bukkit `CommandMap`, which causes issues like this. Now that the hardfork is real, Paper may eventually fully remove `CommandMap` in favor of direct Brigadier API, which would make fixing `CommandMap` behavior like this PR does irrelevant. However, I feel like while Paper is still using the `CommandMap` fixes like this are still valuable.

If we'd rather fix #11378 by changing the implementation of `Bukkit#dispatchCommand` to use Brigadier directly, feel free to close this PR. Although, while not a requirement of the Map interface, it is also neat when `map.get(a) == map.get(a)` for all `a`, which isn't necessarily true for the `BukkitBrigForwardingMap`  before this PR.

### DIFF
--- a/paper-server/patches/sources/com/mojang/brigadier/tree/CommandNode.java.patch
+++ b/paper-server/patches/sources/com/mojang/brigadier/tree/CommandNode.java.patch
@@ -13,7 +13,7 @@
 +    public CommandNode<net.minecraft.commands.CommandSourceStack> clientNode; // Paper - Brigadier API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> unwrappedCached = null; // Paper - Brigadier Command API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> wrappedCached = null; // Paper - Brigadier Command API
-+    public org.bukkit.command.Command wrappedBukkitCommandCached = null; // Paper - Brigadier Command API
++    public org.bukkit.command.Command cachedWrappedBukkitCommand = null; // Paper - Brigadier Command API
 +    // CraftBukkit start
 +    public void removeCommand(String name) {
 +        this.children.remove(name);
--- a/paper-server/patches/sources/com/mojang/brigadier/tree/CommandNode.java.patch
+++ b/paper-server/patches/sources/com/mojang/brigadier/tree/CommandNode.java.patch
@@ -1,6 +1,6 @@
 --- a/com/mojang/brigadier/tree/CommandNode.java
 +++ b/com/mojang/brigadier/tree/CommandNode.java
-@@ -27,11 +_,21 @@
+@@ -27,11 +_,22 @@
      private final Map<String, CommandNode<S>> children = new LinkedHashMap<>();
      private final Map<String, LiteralCommandNode<S>> literals = new LinkedHashMap<>();
      private final Map<String, ArgumentCommandNode<S, ?>> arguments = new LinkedHashMap<>();
@@ -13,6 +13,7 @@
 +    public CommandNode<net.minecraft.commands.CommandSourceStack> clientNode; // Paper - Brigadier API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> unwrappedCached = null; // Paper - Brigadier Command API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> wrappedCached = null; // Paper - Brigadier Command API
++    public org.bukkit.command.Command wrappedBukkitCommandCached = null; // Paper - Brigadier Command API
 +    // CraftBukkit start
 +    public void removeCommand(String name) {
 +        this.children.remove(name);
--- a/paper-server/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
+++ b/paper-server/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
@@ -83,12 +83,12 @@ public class BukkitBrigForwardingMap extends HashMap<String, Command> {
             return null;
         }
 
-        if (node.wrappedBukkitCommandCached != null) {
-            return node.wrappedBukkitCommandCached;
+        if (node.cachedWrappedBukkitCommand != null) {
+            return node.cachedWrappedBukkitCommand;
         }
 
         Command bukkitCommand = PaperBrigadier.wrapNode(node);
-        node.wrappedBukkitCommandCached = bukkitCommand;
+        node.cachedWrappedBukkitCommand = bukkitCommand;
         return bukkitCommand;
     }
 
--- a/paper-server/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
+++ b/paper-server/src/main/java/io/papermc/paper/command/brigadier/bukkit/BukkitBrigForwardingMap.java
@@ -83,11 +83,13 @@ public class BukkitBrigForwardingMap extends HashMap<String, Command> {
             return null;
         }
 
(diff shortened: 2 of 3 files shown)


Owen1212055:
> This allows command frameworks that insert CommandNodes directly into the Brigadier dispatcher to change the permission String of the VanillaCommandNodes created for their commands, rather than it always being the default "minecraft.commands.<name>".

The issue that if at any point `Bukkit#dispatchCommand` is reimplemented, wouldnt the above change kinda be invalidated? I find it very likely that bukkit dispatcher will be ripped out at one point or another, so I don't think this will have *any* effect command execution wise?

willkroboth:
Indeed, if `CommandMap` is ever removed, then `BukkitBrigForwardingMap` would not need to exist, and nothing would use the new `CommandNode.wrappedBukkitCommandCached` field added by this PR.

This PR only fixes `CommandMap` behavior, which is currently used by `Bukkit#dispatchCommand`. The current API is a bit broken in this edge case, and this PR fixes that bug. I don't expect `CommandMap` is going to be removed immediately, so I think it is a good idea to merge this PR. Since the change is so small, it will be obvious how to undo the changes from here when the time comes, so I don't think there is any downside to merging this PR.

But yeah, again, if we just want to fix #11378 by changing the implementation of `Bukkit#dispatchCommand` and don't care about the map identity thing, feel free to close this PR.

paper-server/patches/sources/com/mojang/brigadier/tree/CommandNode.java.patch
```diff
@@ -13,6 +13,7 @@
 +    public CommandNode<net.minecraft.commands.CommandSourceStack> clientNode; // Paper - Brigadier API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> unwrappedCached = null; // Paper - Brigadier Command API
 +    public CommandNode<io.papermc.paper.command.brigadier.CommandSourceStack> wrappedCached = null; // Paper - Brigadier Command API
++    public org.bukkit.command.Command wrappedBukkitCommandCached = null; // Paper - Brigadier Command API
```
Warriorrrr: Semantics but, I think cachedWrappedBukkitCommand would read better than what's currently there

willkroboth: Yeah, makes sense to me. I think I originally went with `wrappedBukkitCommand` + `Cached` to match the naming of `unwrapped` + `Cached` and `wrapped` + `Cache` directly above.

Maybe Owen had a reason to go with the original naming, but in isolation I prefer `cached` + `WrappedBukkitCommand`.

Owen1212055:
I believe that it's the best solution to instead refactor bukkit dispatch command to use the brigadier system rather than going through the command map.

Thank you for your contribution, however. Hopefully I will be able to work on this refactor rather soon as to not hopefully keep this behavior for too much longer.

PR #11791, opened 2024-12-23
Fix BlockRedstoneEvent for tripwire hooks

### LINKED ISSUE #9714: Tripwire hook not registering properly on BlockRedstoneEvents

Lulu13022002:
Original: https://github.com/PaperMC/Paper/pull/9724

Closes https://github.com/PaperMC/Paper/issues/9714

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
@@ -1,21 +1,65 @@
 --- a/net/minecraft/world/level/block/TripWireHookBlock.java
 +++ b/net/minecraft/world/level/block/TripWireHookBlock.java
-@@ -173,9 +_,18 @@
+@@ -127,10 +_,10 @@
+         if (optionalValue.isPresent()) {
+             Direction direction = optionalValue.get();
+             boolean flag = hookState.getOptionalValue(ATTACHED).orElse(false);
+-            boolean flag1 = hookState.getOptionalValue(POWERED).orElse(false);
++            boolean flag1 = hookState.getOptionalValue(POWERED).orElse(false); // Paper - diff on change, for event below
+             Block block = hookState.getBlock();
+             boolean flag2 = !attaching;
+-            boolean flag3 = false;
++            boolean flag3 = false; // Paper - diff on change, for event below
+             int i = 0;
+             BlockState[] blockStates = new BlockState[42];
+ 
+@@ -166,21 +_,48 @@
+             flag2 &= i > 1;
+             flag3 &= flag2;
+             BlockState blockState1 = block.defaultBlockState().trySetValue(ATTACHED, Boolean.valueOf(flag2)).trySetValue(POWERED, Boolean.valueOf(flag3));
++            boolean cancelledEmitterHook = false, cancelledReceiverHook = false; // Paper - Call BlockRedstoneEvent
++            boolean wasPowered = flag1, willBePowered = flag3; // Paper - OBFHELPER
+             if (i > 0) {
+                 BlockPos blockPosx = pos.relative(direction, i);
++                // Paper start - Call BlockRedstoneEvent
++                if (wasPowered != willBePowered) {
++                    int newCurrent = willBePowered ? 15 : 0;
++                    org.bukkit.event.block.BlockRedstoneEvent event = new org.bukkit.event.block.BlockRedstoneEvent(
++                        org.bukkit.craftbukkit.block.CraftBlock.at(level, blockPosx), wasPowered ? 15 : 0, newCurrent
++                    );
++                    event.callEvent();
++                    cancelledReceiverHook = event.getNewCurrent() != newCurrent;
++                }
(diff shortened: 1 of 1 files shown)


paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
```diff
@@ -1,21 +1,52 @@
 --- a/net/minecraft/world/level/block/TripWireHookBlock.java
 +++ b/net/minecraft/world/level/block/TripWireHookBlock.java
-@@ -173,9 +_,18 @@
+@@ -166,21 +_,48 @@
+             flag2 &= i > 1;
+             flag3 &= flag2;
+             BlockState blockState1 = block.defaultBlockState().trySetValue(ATTACHED, Boolean.valueOf(flag2)).trySetValue(POWERED, Boolean.valueOf(flag3));
++            boolean cancelledEmitterHook = false, cancelledReceiverHook = false; // Paper - Call BlockRedstoneEvent properly
++            boolean wasPowered = flag1, willBePowered = flag3; // Paper - OBFHELPER
```
lynxplay, line 21: These should diff the actual final assignment of the values, otherwise this kind of diff does not *really* help during an update (e.g. if flag3 becomes flag4, the diff *might* still apply and the developer might not notice.

paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
```diff
@@ -1,21 +1,52 @@
...
+             if (i > 0) {
+                 BlockPos blockPosx = pos.relative(direction, i);
++                // Paper start - Call BlockRedstoneEvent properly
++                if (wasPowered != willBePowered) {
++                    int newCurrent = willBePowered ? 15 : 0;
++                    org.bukkit.event.block.BlockRedstoneEvent redstoneEvent = new org.bukkit.event.block.BlockRedstoneEvent(org.bukkit.craftbukkit.block.CraftBlock.at(level, blockPosx), wasPowered ? 15 : 0, newCurrent);
++                    level.getCraftServer().getPluginManager().callEvent(redstoneEvent);
```
lynxplay: callEvent();

paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
```diff
@@ -1,21 +1,52 @@
...
++                    if (redstoneEvent.getNewCurrent() != newCurrent) {
++                        cancelledReceiverHook = true;
```
lynxplay: `cancelledReceiverHook = redstoneEvent.getNewCurrent() != newCurrent;`
saves us a random if and imo is a fine change for such a readable expression.

paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
```diff
@@ -1,21 +1,52 @@
...
-+            if (eventRedstone.getNewCurrent() > 0) {
-+                return;
+-            }
++                } // Paper - Call BlockRedstoneEvent properly
++            }
++            // Paper start - Call BlockRedstoneEvent properly
++            if (wasPowered != willBePowered) {
++                int newCurrent = willBePowered ? 15 : 0;
++                org.bukkit.event.block.BlockRedstoneEvent eventRedstone = new org.bukkit.event.block.BlockRedstoneEvent(org.bukkit.craftbukkit.block.CraftBlock.at(level, pos), wasPowered ? 15 : 0, newCurrent);
++                level.getCraftServer().getPluginManager().callEvent(eventRedstone);
```
lynxplay: Same two comments as above. callEvent() and `if(bool) a = true` being a weird thing.

lynxplay (changes requested):
Tested and works, just nitpicks (especially around the obfhelpers, the other is whatever)

lynxplay (approved):

Machine-Maker (approved):

PR #8705, opened 2022-12-21
Parity for respawn events

Lulu13022002:
The PlayerPostRespawnEvent doesn't support the respawn anchor and doesn't have the respawn flag. This also fix a bug where isBedSpawn is true for respawn anchor.

PR #11792, opened 2024-12-23
Parity for respawn events

Lulu13022002:
Original: https://github.com/PaperMC/Paper/pull/8705
The RespawnReason should probably be moved into the super class later but require bytecode rewrite

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/player/AbstractRespawnEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/player/AbstractRespawnEvent.java
@@ -1,0 +1,97 @@
+package io.papermc.paper.event.player;
+
+import com.google.common.collect.ImmutableSet;
+import org.bukkit.Location;
+import org.bukkit.entity.Player;
+import org.bukkit.event.player.PlayerEvent;
+import org.bukkit.event.player.PlayerRespawnEvent;
+import org.jspecify.annotations.NullMarked;
+import java.util.Set;
+
+@NullMarked
+public abstract class AbstractRespawnEvent extends PlayerEvent {
+
+    protected Location respawnLocation;
+    private final boolean isBedSpawn;
+    private final boolean isAnchorSpawn;
+    private final boolean missingRespawnBlock;
+    private final PlayerRespawnEvent.RespawnReason respawnReason;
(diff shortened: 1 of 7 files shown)


paper-api/src/main/java/io/papermc/paper/event/player/AbstractRespawnEvent.java
```diff
@@ -0,0 +1,82 @@
...
+        this.respawnFlags = builder.build();
+    }
+
+    /**
+     * Gets the current respawn location
+     *
+     * @return the current respawn location
+     */
+    public Location getRespawnLocation() {
+        return this.respawnLocation;
```
Machine-Maker: This should be cloned if the server isn't using this value after the event is fired.

Machine-Maker (approved):

paper-api/src/main/java/org/bukkit/event/player/PlayerRespawnEvent.java
```diff
@@ -27,41 +24,16 @@ public PlayerRespawnEvent(@NotNull final Player respawnPlayer, @NotNull final Lo
...
-    }
-
-    /**
-     * Gets the current respawn location
-     *
-     * @return Location current respawn location
-     */
-    @NotNull
-    public Location getRespawnLocation() {
-        return this.respawnLocation;
```
Owen1212055, line 67: PlayerRespawnEvent supports mutability, this change removes that behavior as AbstractRespawnEvent clones it.

Owen1212055 (changes requested):

paper-api/src/main/java/org/bukkit/event/player/PlayerRespawnEvent.java
lynxplay: This was discussed, there is a setter for the respawn location, so this is intential bulletproofing.

Owen1212055: I guess, just not sure about doing changes like these silently.

Owen1212055 (approved):

lynxplay:
Rebased for 1.21.5

kennytv (approved):

PR #11795, opened 2024-12-23
Add Registry#getTags

### LINKED ISSUE #11758: Addition of Registry#getTags

Yes, I cannot find a way in the API to get all tags from a specific registry.

I would like a simple method, such as `Registry#getTags` to get a list/collection of all available tags for a specific registry, that are currently registered to the server.

At the moment I'm having to either use Bukkit's tag system, which leaves much to be desired.
I've also used NMS, which I'd rather stay away from.

### Other

_No response_

Machine-Maker:
Closes https://github.com/PaperMC/Paper/issues/11758

### DIFF
--- a/paper-api/src/main/java/org/bukkit/Registry.java
+++ b/paper-api/src/main/java/org/bukkit/Registry.java
@@ -208,7 +208,7 @@ public interface Registry<T extends Keyed> extends Iterable<T> {
      *
      * @see Particle
      */
-    Registry<Particle> PARTICLE_TYPE = RegistryAccess.registryAccess().getRegistry(RegistryKey.PARTICLE_TYPE); // Paper
+    Registry<Particle> PARTICLE_TYPE = registryFor(RegistryKey.PARTICLE_TYPE); // Paper
     /**
      * Server potions.
      *
@@ -234,7 +234,7 @@ public interface Registry<T extends Keyed> extends Iterable<T> {
      *
      * @see StructureType
      */
-    Registry<StructureType> STRUCTURE_TYPE = Objects.requireNonNull(RegistryAccess.registryAccess().getRegistry(RegistryKey.STRUCTURE_TYPE), "No registry present for StructureType. This is a bug."); // Paper
+    Registry<StructureType> STRUCTURE_TYPE = registryFor(RegistryKey.STRUCTURE_TYPE);
     /**
      * Sound events.
      *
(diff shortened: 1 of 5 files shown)


ShaneBeee:
First of all, thank you for doing this, I appreciate it.

Secondly, I didnt really properly think about this when I asked for getTags, but is it possible to also add `getTagKeys` to return a list of TagKeys?

Machine-Maker:
You can pretty easily just map the Stream from `Stream<Tag<T>> getTags` right?

ShaneBeee:
> You can pretty easily just map the Stream from `Stream<Tag<T>> getTags` right? That's one of the reasons a Stream here makes sense, lots of filtering or mapping or flatmapping people might want to do, so why turn it into a list just to go back into a stream again.

Sorry I may have miss worded that. Forget I said "list".
The point I was trying to make was something to return TagKeys instead of just Tags since Paper seems to be more keen to using TagKey and other things along those lines, felt like it fits more.

Machine-Maker:
Yeah, any such API method would just be doing the `getTags().map(Tag::key)` internally anyways. There isn't a nice way to get all the tag keys without doing that map. So probably wont add that, and I'll just suggest doing `getTags().map(Tag::key)` to get a `Stream<TagKey<T>>`.

ShaneBeee:
> Yeah, any such API method would just be doing the `getTags().map(Tag::key)` internally anyways. There isn't a nice way to get all the tag keys without doing that map. So probably wont add that, and I'll just suggest doing `getTags().map(Tag::key)` to get a `Stream<TagKey<T>>`.

Ok yeah thats fair. Not terribly difficult on the users end. 
Thanks :)

kennytv:
It feels weird to return a stream here, I would just toList it or skip the stream and directly iterate over the bound TagSet, performance shouldn't be of any concern there

kennytv (approved):

ShaneBeee:
Sorry to sound like a total pain in the rear end here...
Since you're now returning a Collection instead of a stream, getting keys would mean stream -> list -> stream again (and map) and then back to a list...
So with that said, maybe now is a good time to add a getTagKeys method?
(since internally you could do the map before toList return)

Machine-Maker:
Yeah, I asked about that, but there isn't really any API that consumes a collection of tag keys, so just make the stream yourself anyways. This shouldn't really be any sort of performance thing, tags don't change except on minecraft:reloads so checking constantly shouldn't be an issue.

PR #11796, opened 2024-12-23
For new registry values, allow copying from existing

Machine-Maker:
Replaces https://github.com/PaperMC/Paper/pull/11724

----

This creates a copy of the `power` enchant and just changes some things about it. This is useful because you don't have to redefine everything about the other enchant, and it will stay updated to the other enchant's values.
```java
@Override
    // io.papermc.testplugin.brigtests.Registration.registerViaBootstrap(context);

    final TypedKey<Enchantment> NEW = EnchantmentKeys.create(Key.key("mm:test"));
    context.getLifecycleManager().registerEventHandler(RegistryEvents.ENCHANTMENT.freeze(), event -> {
        event.registry().factoryRegistry(
			NEW,
			builderFactory -> builderFactory
				.copyFrom(EnchantmentKeys.POWER)
				.maxLevel(10);
            	.description(text("MOARR POWER")
        );
    });

    context.getLifecycleManager().registerEventHandler(LifecycleEvents.TAGS.preFlatten(RegistryKey.ENCHANTMENT), event -> {
        final PreFlattenTagRegistrar<Enchantment> registrar = event.registrar();
        registrar.addToTag(EnchantmentTagKeys.IN_ENCHANTING_TABLE, List.of(TagEntry.valueEntry(NEW)));
    });
}
```

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/registry/RegistryBuilderFactory.java
+++ b/paper-api/src/main/java/io/papermc/paper/registry/RegistryBuilderFactory.java
@@ -1,0 +1,39 @@
+package io.papermc.paper.registry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A factory to create a {@link RegistryBuilder} for a given {@link TypedKey}. For
+ * each instance of this class, once either {@link #empty()} or {@link #copyOf(TypedKey)}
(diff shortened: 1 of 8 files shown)


Machine-Maker:
Something I just thought about... perhaps we should "clone" the nms object before pre-filling the builder with it's values. Just so the newly built value doesn't use the instance of all the subtypes. I don't know if it would cause any problems, but to get out ahead of those.

lynxplay (approved):

PR #11799, opened 2024-12-24
add biome change per player

### LINKED ISSUE #9889: Change biomes per-player in chunk/biome packets

yannicklamprecht:
resolves #9889 

pre hard fork: https://github.com/PaperMC/Paper/pull/10284

# TODO:

- [ ] Implement and call event
- [ ] finish implementing BiomesSnapshot
- [ ] add a handle to Chunk
- [ ] test api with plugin
- [ ] test compatibility with plugins using PlayerChunkLoadEvent
- [ ] reduce diff
- [ ] add paper comments

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/packet/PlayerBiomesLoadEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/packet/PlayerBiomesLoadEvent.java
@@ -1,0 +1,78 @@
+package io.papermc.paper.event.packet;
+
+import org.bukkit.BiomesSnapshot;
+import org.bukkit.Chunk;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.world.ChunkEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Is called when a {@link Player} receives {@link org.bukkit.block.Biome Biomes}
+ * <p>
+ * Can for example be used for replacing Biomes when the player receives the Biome list.
+ * <p>
+ * Should only be used for packet/clientside related stuff.
+ * Not intended for modifying server side state.
+ */
+@NullMarked
+public class PlayerBiomesLoadEvent extends ChunkEvent {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Player player;
+    private @Nullable BiomesSnapshot biomesSnapshot;
+
+    @ApiStatus.Internal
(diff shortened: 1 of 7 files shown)


paper-api/src/main/java/io/papermc/paper/event/packet/PlayerBiomesLoadEvent.java
```diff
@@ -0,0 +1,78 @@
...
+import org.bukkit.Chunk;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.world.ChunkEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Is called when a {@link Player} receives {@link org.bukkit.block.Biome}s
```
ShaneBeee:
```suggestion
 * Is called when a {@link Player} receives {@link org.bukkit.block.Biome Biomes}
```
Looks nicer in the javadocs

Exhibit A:
Screenshot 2024-12-24 at 7 04 27 PM

yannicklamprecht: The plural doesn't seem to work then. At least what I can see in your screen shoot.

ShaneBeee: Screenshot 2024-12-25 at 11 25 40 AM
it puts the "s" as part of the word/link

yannicklamprecht: Thank you, changed it.

yannicklamprecht:
Nothing that I'll finish in near future.

PR #11802, opened 2024-12-24
Fix 0 parameter in Player#applyMending

### LINKED ISSUE #11800: Calling Player#applyMending(int) with zero exp causes ArithmeticException: / by zero
### Expected behavior

The applyMending method should not throw any exceptions when passing 0 as a value. At least that how it was in 1.20.4, since I just started running into that issue since updating to 1.21.4, and haven't changed anything to the plugin code.

I mean it would be a simple fix to not call the method when the value is 0, and I should implement that nonetheless, but throwing that exception seems wrong.

### Observed/Actual behavior

```
Caused by: java.lang.ArithmeticException: / by zero
	at org.bukkit.craftbukkit.entity.CraftPlayer.applyMending(CraftPlayer.java:1899) ~[paper-1.21.4.jar:1.21.4-31-d0d0efe]
	at erenos-3.6.0-SNAPSHOT.jar/net.aerulion.erenos.command.commands.DebugSubCommand.execute(DebugSubCommand.java:37) ~[erenos-3.6.0-SNAPSHOT.jar:?]
	at erenos-3.6.0-SNAPSHOT.jar/net.aerulion.erenos.command.ErenosSubCommand.validateAndExecute(ErenosSubCommand.java:63) ~[erenos-3.6.0-SNAPSHOT.jar:?]
	at erenos-3.6.0-SNAPSHOT.jar/net.aerulion.erenos.command.CommandService.onCommand(CommandService.java:249) ~[erenos-3.6.0-SNAPSHOT.jar:?]
	at org.bukkit.command.PluginCommand.execute(PluginCommand.java:45) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
	... 27 more
```

### Steps/models to reproduce

1. Get and hold a damaged pickaxe with mending: `/give @s minecraft:netherite_pickaxe[enchantments={levels: {"minecraft:mending": 1}},minecraft:damage=1]`
2. Call `Player#applyMending(int)` with 0 xp

### Plugin and Datapack List

Just a test plugin

### Paper version

[02:08:47 INFO]: This server is running Paper version 1.21.4-31-main@d0d0efe (2024-12-23T23:53:38Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-30-aa2c52b (MC: 1.21.4)

### Other

_No response_

lynxplay:
Prevent division by zero in Player#applyMending by simply using zero
instead if the divider itself is zero.
This matches vanilla behaviour in the sense that
ExperienceOrb#repairPlayerItems, which this logic is lifted from, does
pass zero to the enchantment helper and runs its first iteration on
repairing. Vanilla is not affected as the computation failing with zero
is used to compute the leftover exp value, which is not needed if zero
itself is passed in.
As the paper impl however exposes said value to the initial event, our
code needs to account for the zero division.

Resolves: #11800

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
@@ -1896,7 +1896,7 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
                 handle.serverLevel(), itemstack, amount
             );
             int i = Math.min(possibleDurabilityFromXp, itemstack.getDamageValue());
-            final int consumedExperience = possibleDurabilityFromXp > 0 ? i * amount / possibleDurabilityFromXp : possibleDurabilityFromXp; // Paper - taken from ExperienceOrb#repairPlayerItems + prevent division by 0
+            final int consumedExperience = i > 0 ? i * amount / possibleDurabilityFromXp : possibleDurabilityFromXp; // Paper - taken from ExperienceOrb#repairPlayerItems + prevent division by 0
             org.bukkit.event.player.PlayerItemMendEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callPlayerItemMendEvent(handle, orb, itemstack, stackEntry.get().inSlot(), i, consumedExperience);
             i = event.getRepairAmount();
             orb.discard(org.bukkit.event.entity.EntityRemoveEvent.Cause.DESPAWN);
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
@@ -1896,7 +1896,7 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
                 handle.serverLevel(), itemstack, amount
             );
             int i = Math.min(possibleDurabilityFromXp, itemstack.getDamageValue());
-            final int consumedExperience = i * amount / possibleDurabilityFromXp; // Paper - taken from ExperienceOrb#repairPlayerItems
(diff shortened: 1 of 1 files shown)


Doc94:
Then you now can get a event for "rapairs" with 0 exp? (based in how that not repair any)

lynxplay:
Yea, spawn in a `/summon minecraft:experience_orb ~ ~ ~ {Value:0}` and you'll get an event with 0 value.
Could also just happen with a funky enchantment reducing said amount so. 100% something possible.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -1896,7 +1896,7 @@ public int applyMending(int amount) {
                 handle.serverLevel(), itemstack, amount
             );
             int i = Math.min(possibleDurabilityFromXp, itemstack.getDamageValue());
-            final int consumedExperience = i * amount / possibleDurabilityFromXp; // Paper - taken from ExperienceOrb#repairPlayerItems
+            final int consumedExperience = possibleDurabilityFromXp > 0 ? i * amount / possibleDurabilityFromXp : possibleDurabilityFromXp; // Paper - taken from ExperienceOrb#repairPlayerItems + prevent division by 0
```
Machine-Maker: No, this should be i > 0. And rename i to a better name

Machine-Maker (changes requested):

Doc94:
Just for curious.. why this method now just create an ORB and call the method directly rather than copy the whole logic to CB?

Machine-Maker (approved):

kennytv:
Shouldn't the result just be 0?

kennytv (approved):

PR #11805, opened 2024-12-24
Add registry builders for SoundEvent and JukeboxSong

Machine-Maker:
~~Draft until we decided that `Either<L, R>` is the way to go for the fluent getter at `JukeboxSongRegistryEntry#soundEvent`~~ Either it is

```[tasklist]
- [ ] https://github.com/PaperMC/Paper/pull/11796
- [x] Decide to use Either<L, R>
```


paper-api/src/main/java/io/papermc/paper/registry/data/JukeboxSongRegistryEntry.java
```diff
@@ -0,0 +1,123 @@
+package io.papermc.paper.registry.data;
+
+
```
Lulu13022002: This newline is not needed

paper-api/src/main/java/io/papermc/paper/util/EitherRight.java
```diff
@@ -0,0 +1,9 @@
+package io.papermc.paper.util;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+@ApiStatus.Internal
+public record EitherRight<L, R>(R value) implements Either.Right<L, R> {
```
Lulu13022002: Why this is exposed to the api?

paper-server/src/main/java/io/papermc/paper/registry/data/util/Checks.java
```diff
@@ -40,6 +40,13 @@ public static int asArgumentMin(final int value, final String field, final int m
         return value;
     }
 
+    public static float asArgumentMinExclusive(final float value, final String field, final float min) {
+        if (value <= min) {
+            throw new IllegalArgumentException("argument " + field + " must be [" + min + ",+inf)");
```
Lulu13022002: This message should be different to show that the minimum must be exclusive ( or ]

paper-server/patches/sources/net/minecraft/core/HolderLookup.java.patch
```diff
@@ -0,0 +1,36 @@
...
++        Optional<T> getValueForCopying(ResourceKey<T> resourceKey); // Paper - add method to get the value for pre-filling builders in the reg mod API
++
+         ResourceKey<? extends Registry<? extends T>> key();
+ 
+         Lifecycle registryLifecycle();
+@@ -81,6 +_,11 @@
+         default HolderLookup.RegistryLookup<T> filterElements(final Predicate<T> predicate) {
+             return new HolderLookup.RegistryLookup.Delegate<T>() {
+                 @Override
++                public Optional<T> getValueForCopying(final ResourceKey<T> resourceKey) {
```
Lulu13022002, line 19: Missing paper comments for all the overrides

paper-server/src/main/java/io/papermc/paper/registry/data/PaperJukeboxSongRegistryEntry.java
```diff
@@ -0,0 +1,116 @@
...
+
+        @Override
+        public JukeboxSongRegistryEntry.Builder lengthInSeconds(final @Positive float lengthInSeconds) {
+            this.lengthInSeconds = asArgumentMinExclusive(lengthInSeconds, "lengthInSeconds", 0);
+            return this;
+        }
+
+        @Override
+        public JukeboxSongRegistryEntry.Builder comparatorOutput(final @Range(from = 0, to = 15) int comparatorOutput) {
+            this.comparatorOutput = OptionalInt.of(asArgumentRange(comparatorOutput, "comparatorOutput", 0, 15));
```
Lulu13022002: You can use the constants in the Redstone class

kennytv:
I think the Either should allow for non-optional use too, e.g. isLeft/Right and left/right that throw on incorrect use. right now I assume the idea is to use instanceof checks given the two extra interfaces are exposes, but I'd prefer methods on Either for that

lynxplay:
Decided to go with Either, rename however to specific usecase (e.g. Holder instance, so .key() and .directValue or *something*).

PR #11809, opened 2024-12-24
Updated Kyori-Adventure (4.17.0 -> 4.18.0)

96DarkCode96:
Updated dependencies of Adventure + added test for shadow_color.

### DIFF
--- a/paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
+++ b/paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
@@ -79,7 +79,7 @@ public final class AdventureCodecs {
     public static final Codec<Component> COMPONENT_CODEC = recursive("adventure Component",  AdventureCodecs::createCodec);
     public static final StreamCodec<RegistryFriendlyByteBuf, Component> STREAM_COMPONENT_CODEC = ByteBufCodecs.fromCodecWithRegistriesTrusted(COMPONENT_CODEC);
 
-    static final Codec<ShadowColor> SHADOW_COLOR_CODEC = Codec.INT.xmap(ShadowColor::shadowColor, ShadowColor::value);
+    static final Codec<ShadowColor> SHADOW_COLOR_CODEC = ExtraCodecs.ARGB_COLOR_CODEC.xmap(ShadowColor::shadowColor, ShadowColor::value);
 
     static final Codec<TextColor> TEXT_COLOR_CODEC = Codec.STRING.comapFlatMap(s -> {
         if (s.startsWith("#")) {
--- a/paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
+++ b/paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
@@ -79,10 +79,7 @@ public final class AdventureCodecs {
     public static final Codec<Component> COMPONENT_CODEC = recursive("adventure Component",  AdventureCodecs::createCodec);
     public static final StreamCodec<RegistryFriendlyByteBuf, Component> STREAM_COMPONENT_CODEC = ByteBufCodecs.fromCodecWithRegistriesTrusted(COMPONENT_CODEC);
 
-    static final Codec<ShadowColor> SHADOW_COLOR_CODEC = Codec.INT.comapFlatMap(s -> {
-        @Nullable ShadowColor value = ShadowColor.shadowColor(s);
-        return value != null ? DataResult.success(value) : DataResult.error(() -> "Cannot convert " + s + " to adventure ShadowColor");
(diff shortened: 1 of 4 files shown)


paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
```diff
@@ -79,6 +79,11 @@ public final class AdventureCodecs {
     public static final Codec<Component> COMPONENT_CODEC = recursive("adventure Component",  AdventureCodecs::createCodec);
     public static final StreamCodec<RegistryFriendlyByteBuf, Component> STREAM_COMPONENT_CODEC = ByteBufCodecs.fromCodecWithRegistriesTrusted(COMPONENT_CODEC);
 
+    static final Codec<ShadowColor> SHADOW_COLOR_CODEC = Codec.INT.comapFlatMap(s -> {
+        @Nullable ShadowColor value = ShadowColor.shadowColor(s);
```
lynxplay: Your IDE should already be screaming at you that shadow color is never null...
There is no point in a comap flat map for something that is just an xmap.
You are also using the wrong codec, see what vanilla supports for defining shadow numbers in their own codecs (`ExtraCodecs.ARGB_COLOR_CODEC.optionalFieldOf("shadow_color").forGetter(serializer -> Optional.ofNullable(serializer.shadowColor)),`)

paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
```diff
@@ -228,8 +233,9 @@ public String getSerializedName() {
             CLICK_EVENT_CODEC.optionalFieldOf("clickEvent").forGetter(nullableGetter(Style::clickEvent)),
             HOVER_EVENT_CODEC.optionalFieldOf("hoverEvent").forGetter(nullableGetter(Style::hoverEvent)),
             Codec.STRING.optionalFieldOf("insertion").forGetter(nullableGetter(Style::insertion)),
-            KEY_CODEC.optionalFieldOf("font").forGetter(nullableGetter(Style::font))
-        ).apply(instance, (textColor, bold, italic, underlined, strikethrough, obfuscated, clickEvent, hoverEvent, insertion, font) -> {
+            KEY_CODEC.optionalFieldOf("font").forGetter(nullableGetter(Style::font)),
+            SHADOW_COLOR_CODEC.optionalFieldOf("shadow_color").forGetter(nullableGetter(Style::shadowColor))
```
lynxplay: I'd like to mirror the order of net.minecraft.network.chat.Style.Serializer here. Move the shadow_color up under the color. Same below for application.

lynxplay (changes requested):

paper-server/src/main/java/io/papermc/paper/adventure/AdventureCodecs.java
96DarkCode96: Will fix that.

lynxplay: Thank you for the fast fix :pray: the update is super nice for us and you beat me to it by like, half an hour (sorry if my review sounded a bit salty lol)

96DarkCode96: Yea, no worries. I have tried to fix it. It should be okey? Its kinda hard to make it abstract for ARGBLike cause the only implementation is ShadowColorImpl. I'm okay with any criticism 👍

lynxplay: Looks better tho it is still the wrong codec. Just replace Codec.INT with `ExtraCodecs.ARGB_COLOR_CODEC`.
It does not have to be an ARGBLike, that is only for the adventure setter. The getter returns a shadow color.

The mojang codec allows people to specify the color via an int[4] so we need to mirror that (hence the other base codec)

96DarkCode96: Okay, will try that.

96DarkCode96: Now it should be fine. 👍

lynxplay:
Adoption of the shadow color looks good, we'll need to figure out what/how we need to support virtual components.
That will be done in the team tho and (if we need to do anything) we'll relay it here properly.

lynxplay:
Okay so, from my understanding, there is no need to adopt virtual components because we expect them to die on any interaction with the API and the server can/should just treat them as text component impl.
Thank you zml for the insights on that <3

lynxplay (approved):

Machine-Maker (approved):

PR #11810, opened 2024-12-24
Extend HumanEntity#dropItem API

### LINKED ISSUE #11656: Player#dropItem(ItemStack)

Strokkur424:
Closes #11656
Post-softspoon version of https://github.com/PaperMC/Paper/pull/11689

# Motive
Developers currently have to either use NMS or copy-paste the already existing NMS implementation. This PR aims to fix that by exposing ways to call that NMS implementation more easily by extending onto the (Human)Player#dropItem method to provide more extensive ways of making the player drop items from their inventory.

## What does it add?
The following method have been added:
* `Item dropItem(int slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItem(int slot, int amount)`
* `Item dropItem(@NotNull EquipmentSlot slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItem(@NotNull EquipmentSlot slot, int amount)`
* `Item dropItemRandomly(int slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(int slot, int amount)`
* `Item dropItemRandomly(@NotNull EquipmentSlot slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(@NotNull EquipmentSlot slot, int amount)`
* `Item dropItems(int slot, @Nullable Consumer<Item> entityOperation)`
* `Item dropItems(int slot)`
* `Item dropItems(@NotNull EquipmentSlot slot, @Nullable Consumer<Item> entityOperation)`
* `Item dropItems(@NotNull EquipmentSlot slot)`
* `Item dropItemsRandomly(int slot, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemsRandomly(int slot)`
* `Item dropItemsRandomly(@NotNull EquipmentSlot slot, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemsRandomly(@NotNull EquipmentSlot slot)`
* `Item dropItem(@Nullable ItemStack itemStack, @Nullable Consumer<Item> entityOperation)`
* `Item dropItem(@Nullable ItemStack itemStack)`
* `Item dropItemRandomly(@Nullable ItemStack itemStack, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(@Nullable ItemStack itemStack)`

This may look like a lot at first glance (and it kind of is), but in reality, most of these are just default methods on the `HumanEntity` interface. The only, actually implemented methods are:
* `Item dropItem(int slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItem(@NotNull EquipmentSlot slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(int slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(@NotNull EquipmentSlot slot, int amount, @Nullable Consumer<Item> entityOperation)`
* `Item dropItem(@Nullable ItemStack itemStack, @Nullable Consumer<Item> entityOperation)`
* `Item dropItemRandomly(@Nullable ItemStack itemStack, @Nullable Consumer<Item> entityOperation)`

And even among these neither have more than 5 lines in their implementation in `CraftHumanEntity`.

All logic goes through the added methods `private CraftHumanEntity#dropItemRaw(...)` and `private CraftHumandEntity#dropAnyItemRaw(...)`, where the logic is fairly simple due to the pre-implemented NMS method `Player.dropItem(...)`.

Furthermore, `booleam dropItem(boolean)` has been deprecated in favor of `Item dropItems(EquipmentSlot.HAND)` or `Item dropItem(EquipmentSlot.HAND, 1)`.

## Implementation details
Internally, the implementation calls `net.minecraft.world.entity.player.Player#dropItem(...)`, making it a fairly maintainable API.

I have added an additional overload to the the `Player#dropItem(...)` method, which takes in a nullable `Consumer<org.bukkit.entity.Item>` to allow for entity operations before the Entity has been added to the world.

## JavaDocs status
JavaDocs have been added.

# Usage example
```java
// Simple command to test some of the added API
    event.registrar().register(Commands.literal("drop")
        .then(Commands.literal("one")
            .executes(ctx -> {
                ((Player) ctx.getSource().getExecutor()).dropItemRandomly(EquipmentSlot.HAND, 1);
                return 1;
            }))

        .then(Commands.literal("all")
            .executes(ctx -> {
                ((Player) ctx.getSource().getExecutor()).dropItems(EquipmentSlot.HAND);
                return 1;
            }))

        .then(Commands.literal("any")
            .then(Commands.argument("item", ArgumentTypes.itemStack())
                .executes(ctx -> {
                    ((Player) ctx.getSource().getExecutor()).dropItem(ctx.getArgument("item", ItemStack.class), item -> {
                        item.setGlowing(true);
                    });
                    return 1;
                })))

        .build());
});
```

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -703,48 +703,54 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * This will force the entity to drop the item they are holding with
      * an option to drop the entire {@link ItemStack} or just 1 of the items.
      *
+     * @deprecated You should instead use {@link #dropItem(EquipmentSlot, int)} with a {@link EquipmentSlot#HAND} parameter.
      * @param dropAll True to drop entire stack, false to drop 1 of the stack
      * @return True if item was dropped successfully
      */
+    @Deprecated
     public boolean dropItem(boolean dropAll);
-
-    /**
-     * Makes the entity drop an item from their inventory based on the specified ItemStack.
-     * <br>
-     * This method calls {@link HumanEntity#dropItem(int slot, boolean throwRandomly)}
-     * with the first {@link ItemStack} occurrence in the inventory
-     *
-     * @param itemStack     The ItemStack to drop
-     * @param throwRandomly Whether the item should disperse randomly.
-     *                      This means that instead of the item being dropped where the player is currently looking,
-     *                      it instead throws it in any direction, similar to how items drop after a player's death.
-     * @return The dropped item, or null if the action was unsuccessful
-     */
-    public @Nullable Item dropItem(final @NotNull ItemStack itemStack, final boolean throwRandomly);
 
     /**
      * Makes the entity drop an item from their inventory based on the slot.
      *
-     * @param slot          The slot to drop
-     * @param throwRandomly Whether the item should disperse randomly.
-     *                      This means that instead of the item being dropped where the player is currently looking,
-     *                      it instead throws it in any direction, similar to how items drop after a player's death.
+     * @param slot   The slot to drop
+     * @param amount The number of items to drop from this slot. Values below 1 don't drop an Item
      * @return The dropped item entity, or null if the action was unsuccessful
      * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
      */
-    public @Nullable Item dropItem(final int slot, final boolean throwRandomly);
+    public @Nullable Item dropItem(int slot, int amount);
 
     /**
      * Makes the player drop an item from their inventory based on the equipment slot.
      *
-     * @param slot          The equipment slot to drop
-     * @param throwRandomly Whether the item should disperse randomly.
-     *                      This means that instead of the item being dropped where the player is currently looking,
-     *                      it instead throws it in any direction, similar to how items drop after a player's death.
+     * @param slot   The equipment slot to drop
+     * @param amount The amount of items to drop from this equipment slot. Values below 1 don't drop an Item
      * @return The dropped item entity, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(final @NotNull EquipmentSlot slot, final boolean throwRandomly);
+    public @Nullable Item dropItem(@NotNull EquipmentSlot slot, int amount);
 
+    /**
+     * Makes the entity drop an item from their inventory based on the slot.
+     * Instead of the item being dropped where the player is currently looking, this method makes it drop in
+     * a random direction, similar to how items are dropped after a player's death.
+     *
+     * @param slot   The slot to drop
+     * @param amount The number of items to drop from this slot. Values below 1 don't drop an Item
+     * @return The dropped item entity, or null if the action was unsuccessful
+     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
+     */
+    public @Nullable Item dropItemRandomly(int slot, int amount);
+
+    /**
+     * Makes the player drop an item from their inventory based on the equipment slot.
+     * Instead of the item being dropped where the player is currently looking, this method makes it drop in
+     * a random direction, similar to how items are dropped after a player's death.
+     *
+     * @param slot   The equipment slot to drop
+     * @param amount The amount of items to drop from this equipment slot. Values below 1 don't drop an Item
+     * @return The dropped item entity, or null if the action was unsuccessful
+     */
+    public @Nullable Item dropItemRandomly(@NotNull EquipmentSlot slot, int amount);
 
     /**
      * Gets the players current exhaustion level.
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -2,10 +2,12 @@ package org.bukkit.entity;
 
 import java.util.Collection;
 import java.util.Set;
+import java.util.UUID;
 import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
+import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
 import org.bukkit.inventory.InventoryView;
@@ -706,7 +708,6 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      */
     public boolean dropItem(boolean dropAll);
 
-    // Paper start - Extend HumanEntity#dropItem API
     /**
      * Makes the entity drop the first declared {@link ItemStack} occurrence in the inventory
      *
@@ -720,17 +721,17 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
     /**
      * Makes the entity drop an item from their inventory based on the specified ItemStack.
      * <br>
-     * This method calls {@link HumanEntity#dropItem(int slot, java.util.UUID thrower, boolean throwRandomly)}
+     * This method calls {@link HumanEntity#dropItem(int slot, UUID thrower, boolean throwRandomly)}
      * with the first {@link ItemStack} occurrence in the inventory
      *
      * @param itemStack     The ItemStack to drop
-     * @param thrower       The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
+     * @param thrower       The {@link UUID} to set the resulting {@link Item}'s thrower to
      * @param throwRandomly Whether the item should disperse randomly.
      *                      This means that instead of the item being dropped where the player is currently looking,
      *                      it instead throws it in any direction, similar to how items drop after a player's death.
      * @return The dropped item, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(final @NotNull ItemStack itemStack, final @Nullable java.util.UUID thrower, final boolean throwRandomly);
+    public @Nullable Item dropItem(final @NotNull ItemStack itemStack, final @Nullable UUID thrower, final boolean throwRandomly);
 
     /**
      * Makes the entity drop an item from their inventory based on the slot.
@@ -747,22 +748,22 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * Makes the entity drop an item from their inventory based on the slot.
      *
      * @param slot          The slot to drop
-     * @param thrower       The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
+     * @param thrower       The {@link UUID} to set the resulting {@link Item}'s thrower to
      * @param throwRandomly Whether the item should disperse randomly.
      *                      This means that instead of the item being dropped where the player is currently looking,
      *                      it instead throws it in any direction, similar to how items drop after a player's death.
      * @return The dropped item entity, or null if the action was unsuccessful
      * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
      */
-    public @Nullable Item dropItem(final int slot, final @Nullable java.util.UUID thrower, final boolean throwRandomly);
+    public @Nullable Item dropItem(final int slot, final @Nullable UUID thrower, final boolean throwRandomly);
 
     /**
-     * Makes the entity drop an item from their inventory based on the {@link org.bukkit.inventory.EquipmentSlot}
+     * Makes the entity drop an item from their inventory based on the {@link EquipmentSlot}
      *
      * @param slot The equipment slot to drop
      * @return The dropped item entity, or null if the action was unsuccessful
      */
-    public default @Nullable Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot) {
+    public default @Nullable Item dropItem(final @NotNull EquipmentSlot slot) {
         return this.dropItem(slot, null, false);
     }
 
@@ -770,14 +771,14 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * Makes the player drop an item from their inventory based on the equipment slot.
      *
      * @param slot          The equipment slot to drop
-     * @param thrower       The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
+     * @param thrower       The {@link UUID} to set the resulting {@link Item}'s thrower to
      * @param throwRandomly Whether the item should disperse randomly.
      *                      This means that instead of the item being dropped where the player is currently looking,
      *                      it instead throws it in any direction, similar to how items drop after a player's death.
      * @return The dropped item entity, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot, final @Nullable java.util.UUID thrower, final boolean throwRandomly);
-    // Paper end
+    public @Nullable Item dropItem(final @NotNull EquipmentSlot slot, final @Nullable UUID thrower, final boolean throwRandomly);
+
 
     /**
      * Gets the players current exhaustion level.
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -708,27 +708,14 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
 
     // Paper start - Extend HumanEntity#dropItem API
     /**
-     * Makes the entity drop an item from their inventory.
-     * <br>
-     * This method calls {@link HumanEntity#dropItem(int slot)}
-     * with the first {@link ItemStack} occurrence in the inventory
+     * Makes the entity drop the first declared {@link ItemStack} occurrence in the inventory
      *
      * @param itemStack The ItemStack to drop
      * @return The dropped item, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(final @NotNull ItemStack itemStack);
-
-    /**
-     * Makes the entity drop an item from their inventory.
-     * <br>
-     * This method calls {@link HumanEntity#dropItem(int slot, java.util.UUID thrower)}
-     * with the first {@link ItemStack} occurrence in the inventory
-     *
-     * @param itemStack The ItemStack to drop from their inventory
-     * @param thrower   The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
-     * @return The dropped item, or null if the action was unsuccessful
-     */
-    public @Nullable Item dropItem(final @NotNull ItemStack itemStack, final @Nullable java.util.UUID thrower);
+    public default @Nullable Item dropItem(final @NotNull ItemStack itemStack) {
+        return this.dropItem(itemStack, null, false);
+    }
 
     /**
      * Makes the entity drop an item from their inventory based on the specified ItemStack.
@@ -752,17 +739,9 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @return The dropped item, or null if the action was unsuccessful
      * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
      */
-    public @Nullable Item dropItem(final int slot);
-
-    /**
-     * Makes the entity drop an item from their inventory based on the slot.
-     *
-     * @param slot    The slot to drop
-     * @param thrower The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
-     * @return The dropped item, or null if the action was unsuccessful
-     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
-     */
-    public @Nullable Item dropItem(final int slot, final @Nullable java.util.UUID thrower);
+    public default @Nullable Item dropItem(final int slot) {
+        return this.dropItem(slot, null, false);
+    }
 
     /**
      * Makes the entity drop an item from their inventory based on the slot.
@@ -783,16 +762,9 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @param slot The equipment slot to drop
      * @return The dropped item entity, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot);
-
-    /**
-     * Makes the entity drop an item from their inventory based on the {@link org.bukkit.inventory.EquipmentSlot}
-     *
-     * @param slot    The equipment slot to drop
-     * @param thrower The {@link java.util.UUID} to set the resulting {@link Item}'s thrower to
-     * @return The dropped item entity, or null if the action was unsuccessful
-     */
-    public @Nullable Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot, final @Nullable java.util.UUID thrower);
+    public default @Nullable Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot) {
+        return this.dropItem(slot, null, false);
+    }
 
     /**
      * Makes the player drop an item from their inventory based on the equipment slot.
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -737,7 +737,7 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @param slot   The slot to drop
      * @param amount The number of items to drop from this slot. Values below 1 don't drop an Item
      * @return The dropped item entity, or null if the action was unsuccessful
-     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
+     * @throws IllegalArgumentException If the slot is negative or bigger than the player's inventory
      */
     public @Nullable Item dropItemRandomly(int slot, int amount);
 
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -3,6 +3,7 @@ package org.bukkit.entity;
 import java.util.Collection;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Consumer;
 import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Material;
@@ -703,9 +704,9 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * This will force the entity to drop the item they are holding with
      * an option to drop the entire {@link ItemStack} or just 1 of the items.
      *
-     * @deprecated You should instead use {@link #dropItem(EquipmentSlot, int)} with a {@link EquipmentSlot#HAND} parameter.
      * @param dropAll True to drop entire stack, false to drop 1 of the stack
      * @return True if item was dropped successfully
+     * @deprecated You should instead use {@link #dropItem(EquipmentSlot, int)} or {@link #dropItemAll(EquipmentSlot)} with a {@link EquipmentSlot#HAND} parameter.
      */
     @Deprecated(since = "1.21.4")
     public boolean dropItem(boolean dropAll);
@@ -718,7 +719,11 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @return The dropped item entity, or null if the action was unsuccessful
      * @throws IllegalArgumentException If the slot is negative or bigger than the player's inventory
      */
-    public @Nullable Item dropItem(int slot, int amount);
+    public @Nullable Item dropItem(int slot, int amount, @Nullable Consumer<Item> entityOperation);
+
+    public default @Nullable Item dropItem(int slot, int amount) {
+        return this.dropItem(slot, amount, null);
+    }
 
     /**
      * Makes the player drop an item from their inventory based on the equipment slot.
@@ -727,7 +732,11 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @param amount The amount of items to drop from this equipment slot. Values below 1 don't drop an Item
      * @return The dropped item entity, or null if the action was unsuccessful
      */
-    public @Nullable Item dropItem(@NotNull EquipmentSlot slot, int amount);
+    public @Nullable Item dropItem(@NotNull EquipmentSlot slot, int amount, @Nullable Consumer<Item> entityOperation);
+
+    public default @Nullable Item dropItem(@NotNull EquipmentSlot slot, int amount) {
+        return this.dropItem(slot, amount, null);
+    }
 
     /**
      * Makes the entity drop an item from their inventory based on the slot.
@@ -739,7 +748,11 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
      * @return The dropped item entity, or null if the action was unsuccessful
      * @throws IllegalArgumentException If the slot is negative or bigger than the player's inventory
      */
-    public @Nullable Item dropItemRandomly(int slot, int amount);
+    public @Nullable Item dropItemRandomly(int slot, int amount, @Nullable Consumer<Item> entityOperation);
+
+    public default @Nullable Item dropItemRandomly(int slot, int amount) {
+        return this.dropItemRandomly(slot, amount, null);
+    }
 
     /**
      * Makes the player drop an item from their inventory based on the equipment slot.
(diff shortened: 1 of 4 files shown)


kennytv:
All of the default methods that are single-line redirects probably make it harder to parse than they help, I'd reduce some or even all of them, not sure. If you do want to keep default helpers, they should be moved as proper `default` methods to the interface, so it's immediately clear what they do

Strokkur424:
Oh that's smart, I hadn't though about that. Thanks for that suggestion

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +803,86 @@ public boolean dropItem(boolean dropAll) {
...
+
+    @Override
+    public @Nullable org.bukkit.entity.Item dropItem(final int slot, final @Nullable java.util.UUID thrower) {
+        return this.dropItem(slot, thrower, false);
+    }
+
+    @Override
+    public @Nullable org.bukkit.entity.Item dropItem(final int slot, final @Nullable java.util.UUID thrower, final boolean throwRandomly) {
+        // Make sure the slot is in bounds
+        if (slot < 0 || this.inventory.getSize() <= slot) {
```
kennytv:
```suggestion
        if (slot < 0 || slot >= this.inventory.getSize()) {
```
Turning it around makes it clearer

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +803,86 @@ public boolean dropItem(boolean dropAll) {
...
+    public @Nullable org.bukkit.entity.Item dropItem(final @NotNull org.bukkit.inventory.EquipmentSlot slot, final @Nullable java.util.UUID thrower, final boolean throwRandomly) {
+        final ItemStack stack = this.inventory.getItem(slot);
+        final org.bukkit.entity.Item itemEntity = dropItemRaw(stack, thrower, throwRandomly);
+
+        this.inventory.setItem(slot, null);
+        return itemEntity;
+    }
+
+    private org.bukkit.entity.Item dropItemRaw(final ItemStack is, final @Nullable java.util.UUID thrower, final boolean throwRandomly) {
+        if (is == null || is.getType() == Material.AIR) {
```
kennytv: itemStack.isEmpty instead of the type check (and rename the parameter)

kennytv (commented):
And you can remove the Paper comments and add imports instead of fqn

SoSeDiK:
Are this many overloads really needed?
Personally, I'd go only with just one `Item dropItem(ItemStack itemStack)` plus `boolean throwRandomly` and `Consumer<? super Item> function` to apply data before adding to the world, and would make that `itemStack` independent of anything, so that you could drop any item in any context you want. You can always easily manually remove/decrease item's amount in any equipment/inventory slot if needed.

NonSwag:
I think removing the thrower parameter would already decrease the method amount by a fair bit
If you want to set the thrower just do it but by default it should use the player as the thrower

Strokkur424:
> Personally, I'd go only with just one `Item dropItem(ItemStack itemStack)` plus `boolean throwRandomly` and `Consumer<? super Item> function`

I feel like that is too complicated and would kinda defeat its purpose. 
* The reason for the `int slot` overload is the fact that when having just an `ItemStack itemStack` overload, it always just chooses the first one that applies. With a slot you can choose the precise one to drop.
* The reason for the EquipmentSlot one is due to the the standard `int slot` having no way to select those in armor or offhand slots.


> [I] would make that `itemStack` independent of anything

That feels weird, as dropping an item sounds like you actually don't have it in your inventory anymore. Giving the developer the task of removing it themselves feels like a fair bit of unneeded code in this context.

Strokkur424:
> I think removing the thrower parameter would already decrease the method amount by a fair bit If you want to set the thrower just do it but by default it should use the player as the thrower

I mean, it would not remove a single method, as there is also `boolean throwRandomly`, which definitely should be exposed as API, due to the complexity of that operation. Though we could get rid of the `thrower` parameter instead always have `<itemselector>, boolean throwRandomly`, which usually would just make the last param be set to false.

That sounds like a good middle-ground solution to me. If the developer wants to set the thrower they can do so.
I am not sure if it should flag the inventory owner of the Item as its thrower by default, but that can be put up to discussion.

SoSeDiK:
> I feel like that is too complicated and would kinda defeat its purpose.

It's similar to already existing methods (e.g. [World#dropItem](https://jd.papermc.io/paper/1.21.4/org/bukkit/World.html#dropItem(org.bukkit.Location,org.bukkit.inventory.ItemStack,java.util.function.Consumer))), entity spawning methods have the save overloads.

> The reason for the `int slot` overload is the fact that when having just an `ItemStack itemStack` overload, it always just chooses the first one that applies. With a slot you can choose the precise one to drop.

No, I mean to literally not link the dropped itemstack to anything, i.e. the drop would simply drop the cloned item no matter its source, not remove it from inventory or anything else. Then you just get the item in that slot and decrease its amount.

SoSeDiK:
You can add overload to `net.minecraft.world.entity.player.Player#drop`, it (\*well, ServerPlayer) creates ItemEntity right at the beginning of the method

Strokkur424:
It'd honestly rather not change NMS code without a valid reason. Could add the JavaDocs information though

kennytv (dismissed):

Strokkur424:
I suggest that it is not yet merged. After some discussion, I'd like to change it one last time

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +806,53 @@ public boolean dropItem(boolean dropAll) {
...
+        final int slot = this.inventory.first(itemStack);
+        if (slot == -1) {
+            return null;
+        }
+
+        return this.dropItem(slot, throwRandomly);
+    }
+
+    @Override
+    public @Nullable Item dropItem(final int slot, final boolean throwRandomly) {
```
Machine-Maker: Do we have to do some math to convert this "slot" to a bukkit slot number? I remember them being different in the inventory API

Strokkur424: From my testing it seemed to work fine, though if you provide me with more information, I could very quickly check it

Machine-Maker: I think its fine actaually, the slot is only interacting with the bukkit inventory system `this.inventory` is a craftbukkit inventory. So it should be fine.

Machine-Maker (approved):

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -706,6 +708,44 @@ default void openSign(@NotNull org.bukkit.block.Sign sign) {
...
+     *
+     * @param slot          The equipment slot to drop
+     * @param throwRandomly Whether the item should disperse randomly.
+     *                      This means that instead of the item being dropped where the player is currently looking,
+     *                      it instead throws it in any direction, similar to how items drop after a player's death.
+     * @return The dropped item entity, or null if the action was unsuccessful
+     */
+    public @Nullable Item dropItem(final @NotNull EquipmentSlot slot, final boolean throwRandomly);
+
+
```
powercasgamer, line 818:
```suggestion
```

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +806,53 @@ public boolean dropItem(boolean dropAll) {
...
+
+        final ItemEntity droppedEntity = this.getHandle().drop(CraftItemStack.asNMSCopy(itemStack), throwRandomly, true);
+        if (droppedEntity == null) {
+            return null;
+        }
+
+        return (Item) droppedEntity.getBukkitEntity();
+    }
+
+
```
powercasgamer:
```suggestion
```

powercasgamer (changes requested):

Strokkur424:
Alright, I've done the following changes now:
* Removed `throwRandomly` parameter in favor of new methods: `dropItemRandomly(...)`
* Deprecated `boolean dropItem(boolean)` in favor of `Item dropItem(EquipmentSlot.HAND, int)`
* Removed `dropItem(ItemStack, ...)` overloads. They just felt out-of-place and not really particularly. The operations done in that method can be easily simulated by a dev using that API.
* The methods now look like this:
    * `Item dropItem(int slot, int amount)`
    * `Item dropItem(@NotNull EquipmentSlot slot, int amount)`
    * `Item dropItemRandomly(int slot, int amount)`
    * `Item dropItemRandomly(@NotNull EquipmentSlot slot, int amount)`
* The amount can be any arbitrary number. If it is 0 or below, the method just returns null, if it is too big, it will only drop as much as there is amount in that ItemStack from the selected slot.
* (Hopefully) improved the JavaDocs

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -701,11 +703,55 @@ default void openSign(@NotNull org.bukkit.block.Sign sign) {
...
+    @Deprecated
     public boolean dropItem(boolean dropAll);
 
+    /**
+     * Makes the entity drop an item from their inventory based on the slot.
+     *
+     * @param slot   The slot to drop
+     * @param amount The number of items to drop from this slot. Values below 1 don't drop an Item
+     * @return The dropped item entity, or null if the action was unsuccessful
+     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
```
Machine-Maker: I think I'd prefer just a standard IllegalArgumentException for an "invalid slot". That is more flexible for the future if slot numbers are ever non-contiguous.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +806,54 @@ public boolean dropItem(boolean dropAll) {
...
+        final int droppedAmount = Math.min(clonedItemStack.getAmount(), amount);
+        clonedItemStack.setAmount(droppedAmount);
+
+        final ItemEntity droppedEntity = this.getHandle().drop(CraftItemStack.asNMSCopy(clonedItemStack), throwRandomly, true);
+        if (droppedEntity == null) {
+            return null;
+        }
+
+        originalItemStack.setAmount(originalItemStack.getAmount() - droppedAmount);
+        return new CraftItem(this.server, droppedEntity);
```
Machine-Maker: Should just be droppedEntity.getBukkitEntity() instead of constructing a new instance yourself

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +806,54 @@ public boolean dropItem(boolean dropAll) {
...
+        final ItemStack clonedItemStack = originalItemStack.clone();
+        final int droppedAmount = Math.min(clonedItemStack.getAmount(), amount);
+        clonedItemStack.setAmount(droppedAmount);
+
+        final ItemEntity droppedEntity = this.getHandle().drop(CraftItemStack.asNMSCopy(clonedItemStack), throwRandomly, true);
+        if (droppedEntity == null) {
+            return null;
+        }
+
+        originalItemStack.setAmount(originalItemStack.getAmount() - droppedAmount);
```
Machine-Maker: I think this logic could be made nicer by operating on the underlying nms.ItemStack. All the Inventory#getItem should be returning mirrors (lets double check to be sure), but if they are, you can just get the underlying stack with `CraftItemStack#unwrap`. Then you can use the nms `.split` method which nicely handles all of the splitting/copying for you.

Machine-Maker (changes requested):

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
Strokkur424: Yeah they are all returning mirrors.

Machine-Maker: Perfect, then just unwrap the mirror and use the methods on the nms ItemStack

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -701,11 +703,55 @@ default void openSign(@NotNull org.bukkit.block.Sign sign) {
...
+
+    /**
+     * Makes the entity drop an item from their inventory based on the slot.
+     * Instead of the item being dropped where the player is currently looking, this method makes it drop in
+     * a random direction, similar to how items are dropped after a player's death.
+     *
+     * @param slot   The slot to drop
+     * @param amount The number of items to drop from this slot. Values below 1 don't drop an Item
+     * @return The dropped item entity, or null if the action was unsuccessful
+     * @throws IndexOutOfBoundsException If the slot is negative or bigger than the player's inventory
```
Machine-Maker: forgot about this IllegalArgumentException

Strokkur424: Woops

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +805,50 @@ public boolean dropItem(boolean dropAll) {
...
+        return dropItemRaw(this.inventory.getItem(slot), amount, true);
+    }
+
+    private @Nullable Item dropItemRaw(final ItemStack originalItemStack, final int amount, final boolean throwRandomly) {
+        if (originalItemStack == null || originalItemStack.isEmpty() || amount <= 0) {
+            return null;
+        }
+
+        final net.minecraft.world.item.ItemStack nmsItemStack = CraftItemStack.unwrap(originalItemStack);
+        final net.minecraft.world.item.ItemStack dropContent = nmsItemStack.split(Math.min(originalItemStack.getAmount(), amount));
```
Machine-Maker: I don't think you need to do the Math.min. That should be part of the nmsItemStack.split now. Check the source of ItemStack#split

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +805,50 @@ public boolean dropItem(boolean dropAll) {
...
+        return dropItemRaw(this.inventory.getItem(slot), amount, true);
+    }
+
+    private @Nullable Item dropItemRaw(final ItemStack originalItemStack, final int amount, final boolean throwRandomly) {
+        if (originalItemStack == null || originalItemStack.isEmpty() || amount <= 0) {
+            return null;
+        }
+
+        final net.minecraft.world.item.ItemStack nmsItemStack = CraftItemStack.unwrap(originalItemStack);
+        final net.minecraft.world.item.ItemStack dropContent = nmsItemStack.split(originalItemStack.getAmount());
```
Machine-Maker: should be split(amount)? To copy out that `amount` from the stack?

Strokkur424: Yeah, my bad. I should definitely go to sleep 😅

Machine-Maker (approved):

SoSeDiK:
I genuinely do not see the benefit of doing this, as opposed to:

<details>
<summary>Code sample</summary>

```java
// Get any item you want
ItemStack item = player.getInventory().getItem(EquipmentSlot.HAND);
ItemStack item = ItemStack.of(Material.GOWSTONE);
ItemStack item = // some other source

// Drop it, provide any amount with `item.asQuantity(amount)`
// Additionally with optional overload to apply data before drop event & item spawn occurred
player.dropItem(item, item -> {
    item.setGlowing(true);
});

// Subtract its amount if needed (this will decrease item's amount in inventory)
item.subtract(item.getAmount());
```

</details>

The difference would literally be a one-liner (subtract is done behind the scenes instead of manually at the cost of additional clutter), but with more control and wouldn't need deprecating `dropItem`.

Strokkur424:
I disagree with that approach. This is meant to serve as an API to simulate dropping items from the inventory in basically the same way as a normal player would. 

I understand what you mean, and I see value in it, yes. But it doesn't fit what I have in mind for it. Dropping an arbituary amount of any arbituary item may be nice, but it is an entirely different scenario.

The deprecation of `boolean dropItem(boolean)` was done due to the fact that is has been superseded by a better api.

Your suggestion essentially is saying to abandon all of the useful implications of a one liner to drop 20 items in a player's inventory and never have to worry about reducing items again.
Your variant instead would nullify all of that, forcing the user to care about stuff that usually wouldn't be the case.

The consumer to modify the resulting Item entity is nice, but not really required due to the fact that it returns thay very same entity, meaning you are free to make it glow even now.

This PR is meant to expose vanilla behaviour. And I see this as my final decision.

If you want to see that API inside Paper, you are free to PR it yourself, but I am not willing to remove useful functionality.

SoSeDiK:
> I disagree with that approach. This is meant to serve as an API to simulate dropping items from the inventory in basically the same way as a normal player would.

I don't necessarily mind such API, I just don't see the benefit of having multiple methods like this for something that could be done with just one method and cover all proposed cases at once.

> I understand what you mean, and I see value in it, yes. But it doesn't fit what I have in mind for it. Dropping an arbituary amount of any arbituary item may be nice, but it is an entirely different scenario.

Would be repeating of myself, but I don't see both cases as different scenarios to a degree of requiring separate methods. The only difference between dropping an arbitrary item and an inventory item is decreasing the inventory item's amount, which is very trivial to do.

My suggestion is meant to serve the same purpose (drop an item as if the player did that), except the item not being linked to the inventory. It solves the linked issue and allows for more control and possibilities.

> The deprecation of `boolean dropItem(boolean)` was done due to the fact that is has been superseded by a better api.

Subjective. I don't see anything that wrong with `boolean dropItem(boolean)` to a degree of "yeeting" (/requiring to replace) it. It may not be as descriptive & useful in providing `Item` vs `boolean`, but it serves its purpose. With the new approach, the old code will have to be rewritten as such:

<details>
<summary>Code comparison</summary>

```java
// Drop a single active item; is fine and I'd even say somewhat better in being more descriptive/explicit
/// Before
player.dropItem();
/// After
player.dropItem(EquipmentSlot.HAND, 1);

// Drop the whole hand; that's were I'm not a fan
/// Before
player.dropItem(true);
/// After
/// You need to either provide a large enough number to cover the whole stack
player.dropItem(EquipmentSlot.HAND, Integer.MAX_AMOUNT);
// Or obtain the item beforehand
player.dropItem(EquipmentSlot.HAND, player.getInventory().getItem(EquipmentSlot.HAND).getAmount());
```

</details>

> Your suggestion essentially is saying to abandon all of the useful implications of a one liner to drop 20 items in a player's inventory and never have to worry about reducing items again. Your variant instead would nullify all of that, forcing the user to care about stuff that usually wouldn't be the case.

Yes, because I don't see the usefulness. I'd argue the opposite, that this approach adds unnecessary complexity, while the alternative would be easier (subjective?) and cover more cases. "have to worry about reducing items" is a short trivial one-liner.

> The consumer to modify the resulting Item entity is nice, but not really required due to the fact that it returns thay very same entity, meaning you are free to make it glow even now.

It would allow modifying an item before passing it to other plugins and actually spawning in the world. As an example, you can mark your item with some persistent tag and handle it in drop/spawn event without requiring some extra cache.

Other places like world drop events already have such overload, so it wouldn't be alien to be on pair with them.

> This PR is meant to expose vanilla behaviour. And I see this as my final decision.
> 
> If you want to see that API inside Paper, you are free to PR it yourself, but I am not willing to remove useful functionality.

The decision is up to the team, I don't mind doing a separate PR to cover my use-case (I already have it done in my fork). It's just that my addition would automatically cover this use-case as well, producing multiple methods for the same purpose.

Strokkur424:
> Code comparison
Just saying, but that's actually incorrect. You'd have to call `player.dropItem(false)` to drop the item currently in-hand.
That method is also limited to just the main hand. You can't drop the off-hand, which you can do with my API.

These changes have been approved by the paper team and thus seem to be fine.

SoSeDiK:
> Just saying, but that's actually incorrect. You'd have to call `player.dropItem(false)` to drop one of the item currently in-hand. 

Updated to include `false`.

> That method is also limited to just the main hand. You can't drop the off-hand, which you can do with my API.

Current method is supposed to drop the currently holding item, not imitate dropping any item from the inventory. The comparison was also about that, since it's the only use-case in which the deprecated method was being used.

> These changes have been approved by the paper team and thus seem to be fine.

Still can discussed and re-reviewed.

___

I actually found an edge case in my proposed alternative. The drop event would have to be handled separately (one extra overload to NMS method) to not try re-adding the item back to the inventory since the item's source might be whatever and we're dropping a cloned item.

This edge case is also present in your methods. If you drop e.g. head item and cancel the drop event, the item will be:
\- placed in main hand if it's empty
\- increase main item's amount if only one item was dropped and main hand is similar
\- tried to be added back to the inventory (this does not include equipment slots, i.e. the head item will be lost if this fails)
\- erased forever
But this might be expected? Not sure

Machine-Maker:
I think going with the dropItem(ItemStack) method would require doing an extra copy of the itemstack. You make a copy with asQuantity(), but then we would have to copy the stack again to drop it. I don't think anywhere in the API (or very rarely) we just "unwrap" the itemstack passed in as a parameter to an API method. If the methods accept a slot, then we only need the single copy to get the stack with the specific amount to drop, and just directly mutate the stack in the inventory.

I think there is a usecase for dropping an arbitrary stack, but what about it just being an additional method that takes an arbitrary stack?

I also like the Consumer<Item>, but that can apply to all the methods.

Strokkur424:
It's also not as if I don't like the Consumer, but removing the existing methods feels limiting. 
Perhaps, we add a new group of methods, `dropAnyItem` and `dropAnyItemRandomly`, which does no operations on the player's inventory. This group would not have that quantity param, as you can just set the amount on the ItemStack.

Also, I can add those Consumer params as nullable ones

SoSeDiK:
> I think going with the dropItem(ItemStack) method would require doing an extra copy of the itemstack. You make a copy with asQuantity(), but then we would have to copy the stack again to drop it. I don't think anywhere in the API (or very rarely) we just "unwrap" the itemstack passed in as a parameter to an API method. If the methods accept a slot, then we only need the single copy to get the stack with the specific amount to drop, and just directly mutate the stack in the inventory.

Hmm, fair. In dropping an arbitrary stack there will be 2 or potentially 3 copies: `asQuantity` (could do `int amount` param to prevent it), our internal to detach from source itemstack, and one when creating the item entity.

Manually providing the amount and working with the source directly allows avoiding the first two copies. Looks like there's demand for this, so I'm fine with these and making a separate PR for arbitrary items 👍️
___

Regarding the drop event, seems to be a known bug (https://github.com/PaperMC/Paper/issues/7726) and covers normal player interactions too (`Q` on helmet with full inventory and cancelled player drop event will erase the helmet forever), so it's fine to ignore within this PR.

I wonder if the https://github.com/PaperMC/Paper/pull/7727 could be extended to try preserving the original slot in any scenario, or at least cover equipment slots too?

Strokkur424:
Alright, here is my proposal for methods (all return Item):
`dropItem(int slot, int amount, @Nullable Consumer<Item>)`
`dropItem(EquipmentSlot slot, int amount, @Nullable Consumer<Item>)`
`dropItemRandomly(int slot, int amount, @Nullable Consumer<Item>)`
`dropItemRandomly(EquipmentSlot slot, int amount, @Nullable Consumer<Item>)`

`dropItemAll(int slot, @Nullable Consumer<Item>)`
`dropItemAll(EquipmentSlot slot, @Nullable Consumer<Item>)`
`dropItemAllRandomly(int slot, @Nullable Consumer<Item>)`
`dropItemAllRandomly(EquipmentSlot slot, @Nullable Consumer<Item>)`

`dropAnyItem(ItemStack itemStack, @Nullable Consumer<Item>)`
`dropAnyItemRandomly(ItemStack itemStack, @Nullable Consumer<Item>)`

This equals to a fair bit of methods, but each has a purpose. Any comments @SoSeDiK?

SoSeDiK:
I think the Consumer param should be an optional overload (defaulting to `null`). It'll be an extra method for each, but the maintenance cost is pretty much zero, and they all make sense.

I like the `dropItemAll`, seems sensible deprecating the old method now.

Otherwise, looks good, happy with this 👍️

Strokkur424:
Very well, will go to work doing the changes!

NonSwag:
> `dropAnyItem(ItemStack itemStack, @Nullable Consumer<Item>)`
> `dropAnyItemRandomly(ItemStack itemStack, @Nullable Consumer<Item>)`

I would call those methods `dropItem` and `dropItemRandomly`

> `dropItemAll(int slot, @Nullable Consumer<Item>)`
> `dropItemAll(EquipmentSlot slot, @Nullable Consumer<Item>)`
> `dropItemAllRandomly(int slot, @Nullable Consumer<Item>)`
> `dropItemAllRandomly(EquipmentSlot slot, @Nullable Consumer<Item>)`

I would call those methods `dropItems` instead of `dropItemAll`

Strokkur424:
Just for testing, I setup the following command: 
```java
event.registrar().register(Commands.literal("drop")
        .then(Commands.literal("one")
            .executes(ctx -> {
                ((Player) ctx.getSource().getExecutor()).dropItemRandomly(EquipmentSlot.HAND, 1);
                return 1;
            }))

        .then(Commands.literal("all")
            .executes(ctx -> {
                ((Player) ctx.getSource().getExecutor()).dropItemAll(EquipmentSlot.HAND);
                return 1;
            }))

        .then(Commands.literal("any")
            .then(Commands.argument("item", ArgumentTypes.itemStack())
                .executes(ctx -> {
                    ((Player) ctx.getSource().getExecutor()).dropAnyItem(ctx.getArgument("item", ItemStack.class), item -> {
                        item.setGlowing(true);
                    });
                    return 1;
                })))

    .build());
```

You can see the result here:

https://github.com/user-attachments/assets/7e4eefaf-1dcd-459f-9c05-dc641cda5bbd

NonSwag:
Also, I don't really like that all methods require an item consumer (or null)
I'd propose adding overload methods that pass null for you
But on the other hand, those are way too many methods already even without overloads

Strokkur424:
> I would call those methods `dropItem` and `dropItemRandomly`

I feel like new API users could get confused on what those methods do, even when reading the JavaDoc (which I am yet to add).
I want to make it fairly clear which methods drop actual ItemStack **from** the player's inventory and which methods could cause a dupe glitch 😂.

> I would call those methods `dropItems` instead of `dropItemAll`

Yeah, sounds like a good idea.

Strokkur424:
> Also, I don't really like that all methods require an item consumer (or null) I'd propose adding overload methods that pass null for you But on the other hand, those are way too many methods already even without overloads

I already have overload methods for that

NonSwag:
> > I would call those methods `dropItem` and `dropItemRandomly`
> 
> I feel like new API users could get confused on what those methods do, even when reading the JavaDoc (which I am yet to add). I want to make it fairly clear which methods drop actual ItemStack **from** the player's inventory and which methods could cause a dupe glitch 😂.

`dropAnyItem` and `dropAnyItemRandomly` sounds even more confusing in my opinion
also the `Any` does not add any information to what the method does

Strokkur424:
Do you have any idea on what I could rename `dropAnyItem` to so that it is clear it is not tied to the Player in any way?

NonSwag:
> Do you have any idea on what I could rename `dropAnyItem` to so that it is clear it is not tied to the Player's inventory in any way?

Potentially somthing like `dropItemAt`

Strokkur424:
> Potentially somthing like `dropItemAt`

Not sure I like it. `dropItemAt(ItemStack)` sounds weird, since I'd assume `at` to have some sort of Location parameter

NonSwag:
but i think just mentioning in the docs that the item is not tied to the players inventory at all and will spawn a new item is better
in any case dropItem would be a better fit

Strokkur424:
Yeah probably. I am not having any ideas either

SoSeDiK:
> I would call those methods `dropItem` and `dropItemRandomly`

> but i think just mentioning in the docs that the item is not tied to the players inventory at all and will spawn a new item is better
in any case dropItem would be a better fit

Clashing with other methods isn't ideal either, since the behavior is different enough. Even with javadocs it'd be quite easy to miss.

I don't like `Any` too, but couldn't think of anything better yet.

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -1376,6 +1383,11 @@
 +                    return null;
 +                }
 +            }
++            // Paper start - Extend HumanEntity#dropItem API
++            if (entityOperation != null) {
++                entityOperation.accept((org.bukkit.entity.Item) itemEntity.getBukkitEntity());
```
SoSeDiK: Might make sense to apply it before `PlayerDropItemEvent`?

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
```diff
@@ -801,6 +806,89 @@ public boolean dropItem(boolean dropAll) {
...
+        final ItemStack itemStack, final boolean throwRandomly,
+        final @Nullable Consumer<Item> entityOperation
+    ) {
+        if (itemStack == null || itemStack.isEmpty()) {
+            return null;
+        }
+
+        final net.minecraft.world.item.ItemStack nmsItemStack = CraftItemStack.asNMSCopy(itemStack);
+
+        final ItemEntity droppedEntity = this.getHandle().drop(nmsItemStack, throwRandomly, true, true, entityOperation);
```
SoSeDiK: `drop` should have a separate handling for cancelled `PlayerDropItemEvent` in case of arbitrary items. This is a special case, the item's source might not be player's inventory, so the item should not be neccesarily added to the player's inventory, but rather let the dev handle it.

Edit: also, as an example, if the item was from the inventory, we will be adding back its copy, essentially duplicating the item, which might not be the desired behavior. It is way easier to manually add the item to inventory if this is wanted than to prevent this if this would be the default behavior.

willkroboth:
> I don't like `Any` too, but couldn't think of anything better yet.

Would `createDroppedItem` and `createDroppedItemRandomly` work? I think `create` indicates that the item is new and doesn't already exist in the player's inventory.

SoSeDiK:
As another suggestion, `dropItemStack` might work?

Machine-Maker:
Maybe `dropItemFrom(int/slot)` and dropItemStack()?

Strokkur424:
> Maybe `dropItemFrom(int/slot)` and dropItemStack()?

Yeah I like that one.

Machine-Maker:
Ok, now its just soo many methods. What was the original issue with a `randomly` boolean parameter instead? I kinda think just adding that to the methods and reducing the number of methods by half is better.

SoSeDiK (changes requested):

SoSeDiK:
I think non-entityOperation ones could stay (@Machine-Maker?), given that's the only overload left; while this would be 3 methods * 2 for item consumption overloads, it would be on pair with other such APIs, and requiring users to pass `null` each time is annoying

Machine-Maker (changes requested):
So we decided that there should be 10 methods added here
- `dropItem(int slot, int amount)`
- `dropItem(int slot, int amount, bool, consumer)`

- `dropItem(EquipmentSlot slot, int amount)`
- `dropItem(EquipmentSlot slot, int amount, bool, consumer)`

- `dropItem(int slot)`
- `dropitem(int slot, bool, consumer)`
- `dropItem(EquipmentSlot slot, bool, consumer)`
- `dropItem(EquipmentSlot, slot)`

- `dropItem(ItemStack)`
- `dropItem(ItemStack, bool, consumer)`

SoSeDiK:
> So we decided that there should be 10 methods added here

The methods make sense. Did you decide to go with the same naming for every method instead of separating inventory & arbitrary as proposed earlier?

Machine-Maker:
Yeah, I think the parameters and the javadocs will make it clear what happens. I think also, the ItemStack methods will make defensive copies, so dropping a stack from the inv won’t actually drop it, just copy it.

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -732,121 +734,56 @@ default void openSign(@NotNull org.bukkit.block.Sign sign) {
...
-     * @param slot   The equipment slot to drop
-     * @param amount The amount of items to drop from this equipment slot. Values below one always return null
+     * @param slot The equipment slot to drop
      * @return The dropped item entity, or null if the action was unsuccessful
      */
     @Nullable
-    public default Item dropItem(@NotNull EquipmentSlot slot, int amount) {
-        return this.dropItem(slot, amount, null);
+    public default Item dropItem(int slot) {
+        return this.dropItem(slot, Integer.MAX_VALUE, false, null);
```
NonSwag: int max value feels a bit weird
this can probably just be moved to the impl and use the size of the stack

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -732,121 +734,56 @@ default void openSign(@NotNull org.bukkit.block.Sign sign) {
...
+     *                        a random direction, similar to how items are dropped after a player's death.
+     * @param entityOperation The function to be run before adding the entity into the world
      * @return The dropped item entity, or null if the action was unsuccessful
-     * @throws IllegalArgumentException If the slot is negative or bigger than the player's inventory
      */
-    @Nullable
-    public default Item dropItemRandomly(int slot, int amount) {
-        return this.dropItemRandomly(slot, amount, null);
+    public default Item dropItem(int slot, boolean throwRandomly, @Nullable Consumer<Item> entityOperation) {
+        return this.dropItem(slot, Integer.MAX_VALUE, throwRandomly, entityOperation);
```
NonSwag: same here

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -856,62 +793,21 @@ public default Item dropItems(@NotNull EquipmentSlot slot, @Nullable Consumer<It
      * @return The dropped item entity, or null if the action was unsuccessful
      */
     @Nullable
-    public default Item dropItems(@NotNull EquipmentSlot slot) {
-        return this.dropItems(slot, null);
+    public default Item dropItem(@NotNull EquipmentSlot slot) {
+        return this.dropItem(slot, Integer.MAX_VALUE, false, null);
```
NonSwag: and here

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -856,62 +793,21 @@ public default Item dropItems(@NotNull EquipmentSlot slot, @Nullable Consumer<It
...
-     * a random direction, similar to how items are dropped after a player's death.
-     *
-     * @param slot The equipment slot to drop
-     * @return The dropped item entity, or null if the action was unsuccessful
-     */
-    @Nullable
-    public default Item dropItemsRandomly(@NotNull EquipmentSlot slot) {
-        return this.dropItemsRandomly(slot, null);
+    public default Item dropItem(@NotNull EquipmentSlot slot, boolean throwRandomly, @Nullable Consumer<Item> entityOperation) {
+        return this.dropItem(slot, Integer.MAX_VALUE, throwRandomly, entityOperation);
```
NonSwag: and here

Strokkur424:
@Machine-Maker alright, I have changed the methods to reflect the ones you have listed above. 

Also, just to confirm, the `dropItem(ItemStack)` one **does** indeed make a copy. So running that logic on an ItemStack currently in the inventory (I tested it on the currently held one) drops it without reducing it

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
lynxplay: Yes

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftHumanEntity.java
lynxplay: Yes

Machine-Maker: Yeah, the event shouldn't be even called in this case, as the event is for when dropping from the inventory. That will solve this issue.

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
Machine-Maker: Yeah. @lynxplay is making that change now.

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
lynxplay: Its fine :+1:

lynxplay (approved):

Machine-Maker (approved):

PR #11814, opened 2024-12-25
Add new Lidded API, deprecating existing bukkit Lidded API. Adds PlayerLiddedOpenEvent

### LINKED ISSUE #11364: Lidded containers closing issue

456dev:
replaces #11379, pre hard-fork pr

minimal changes from previous pr, mostly some variable renames + //paper comments, to be in-line with the new guidelines.

- [ ] Additionally, I need to double check the implementation of PlayerLiddedOpenEvent for shulker boxes, particularly if players can leak.

previous description below:

Featuring
- 5 LidModes, 3 of which were not possible before
- The ability to get if the lid actually should be open, if a player is in the container
- The ability to get if the lid is open, from api or the player itself

This also deprecates Bukkits api because of the confusing and outright incorrect javadocs (so those now accurately represent the behaviour)

Testable with the included testplugin commands.

I'm not 100% sure if the PaperLidded interface is the best way of doing this, but it is working
Also, let me know if the naming needs to change.
- [ ] see if anything important is skipped in the check method scheduled every 5 ticks on chests

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/block/LidState.java
+++ b/paper-api/src/main/java/io/papermc/paper/block/LidState.java
@@ -1,0 +1,9 @@
+package io.papermc.paper.block;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public enum LidState {
+    OPEN,
+    CLOSED
+}
--- a/paper-api/src/main/java/io/papermc/paper/block/Lidded.java
+++ b/paper-api/src/main/java/io/papermc/paper/block/Lidded.java
@@ -1,0 +1,37 @@
+package io.papermc.paper.block;
+
+import org.bukkit.block.TileState;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface Lidded extends TileState {
+
+    /**
+     * Gets the current state of the block, respecting the lidded mode.
+     *
+     * @return the effective lid state
+     */
+    LidState getEffectiveLidState();
+
+    /**
+     * Gets how the lid would be without any lidded mode, based on players interacting with the block.
+     * @return the true lid state
+     */
+    LidState getTrueLidState();
+
+    /**
+     * Gets the current lid mode of the block.
+     *
+     * @return the lid mode
+     */
+    LidMode getLidMode();
+
+    /**
+     * Sets the lid mode of the block.
+     *
+     * @param mode the new lid mode
+     * @return the actually set lid mode
+     */
+    LidMode setLidMode(LidMode mode);
+
+}
--- a/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLiddedOpenEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLiddedOpenEvent.java
@@ -1,0 +1,89 @@
+package io.papermc.paper.event.player;
+
+import io.papermc.paper.block.LidMode;
+import io.papermc.paper.block.LidState;
+import io.papermc.paper.block.Lidded;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called when a player opens a {@link Lidded} block.
+ *
+ * <p>
+ * This is called every time a player opens a {@link Lidded} block
+ * regardless of if the lid is already open (e.g. multiple players).
+ * <p>
+ * Cancelling this event prevents the player from being considered in other {@link Lidded} methods:
+ * they will not contribute to the {@link Lidded#getTrueLidState()} and {@link Lidded#getEffectiveLidState()}.
+ * <p>
+ * This event is called twice for double chests, once for each half.
+ */
+@NullMarked
+public class PlayerLiddedOpenEvent extends PlayerEvent implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+    private final Lidded blockState;
+    private final Block block;
+    private boolean cancelled;
+
+    @ApiStatus.Internal
+    public PlayerLiddedOpenEvent(final @NotNull Player who, final @NotNull Lidded blockState, final @NotNull Block block) {
+        super(who);
+        this.cancelled = false;
+        this.blockState = blockState;
+        this.block = block;
+    }
+
+
+    @Override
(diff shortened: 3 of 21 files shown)


456dev:
Fixed #11364 with new api

paper-api/src/main/java/org/bukkit/block/Lidded.java
```diff
@@ -1,25 +1,34 @@
 package org.bukkit.block;
 
+/**
+ * @deprecated Incomplete api. Use {@link io.papermc.paper.block.Lidded} instead.
+ */
+@Deprecated // Paper - Deprecate Bukkit's Lidded API
```
456dev, line 8: TODO: remove this lone `// paper` comment
```suggestion
@Deprecated
```

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLiddedOpenEvent.java
```diff
@@ -0,0 +1,89 @@
...
+ * Called when a player opens a {@link Lidded} block.
+ *
+ * <p>
+ * This is called every time a player opens a {@link Lidded} block
+ * regardless of if the lid is already open (e.g. multiple players).
+ * <p>
+ * Cancelling this event prevents the player from being considered in other {@link Lidded} methods:
+ * they will not contribute to the {@link Lidded#getTrueLidState()} and {@link Lidded#getEffectiveLidState()}.
+ * <p>
+ * This event is called twice for double chests, once for each half.
```
456dev, line 25: todo: fix global implementation for double chests.

iirc as well as this double-fire behaviour, modifying the state of the non-dominant half doesnt play a sound, but still keeps the animation open seperately.

paper-api/src/main/java/io/papermc/paper/block/Lidded.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Gets the current state of the block, respecting the lidded mode.
+     *
+     * @return the effective lid state
+     */
+    LidState getEffectiveLidState();
+
+    /**
+     * Gets how the lid would be without any lidded mode, based on players interacting with the block.
```
456dev, line 17: TODO: replace with smth like

```suggestion
     * Gets how the lid would be with {@link LidMode.DEFAULT}
     * I.E. based on players interacting with the block.
```

paper-api/src/main/java/io/papermc/paper/block/LidState.java
```diff
@@ -0,0 +1,9 @@
+package io.papermc.paper.block;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
```
456dev, line 5: Is this actually needed, since its a very simple enum?

Warriorrrr: This enum and LidMode don't have anything that would be affected by it, so I would say it's not needed.

paper-api/src/main/java/org/bukkit/block/Lidded.java
```diff
@@ -1,25 +1,36 @@
...
     /**
-     * Checks if the block's animation state.
+     * Checks is the Lid is currently forced open.
      *
-     * @return true if the block's animation state is set to open.
+     * @return true if the block's animation state is force open.
+     * @deprecated Misleading name. Use {@link io.papermc.paper.block.Lidded#getLidMode()} for the direct replacement, or {@link io.papermc.paper.block.Lidded#getEffectiveLidState()} to tell if the lid is visibly open to the player instead.
      */
+    @Deprecated
     boolean isOpen();
```
Warriorrrr, line 34: If these names are misleading, I think it would be better to fix the implementation rather than changing the javadoc and deprecating the methods.

I think it would also be better to add onto this class rather than creating a whole new interface.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftBarrel.java
```diff
@@ -35,49 +39,63 @@ public Inventory getInventory() {
...
     @Override
-    public void open() {
+    public CraftBarrel copy() {
+        return new CraftBarrel(this, null);
+    }
+
+    @Override
+    public CraftBarrel copy(Location location) {
+        return new CraftBarrel(this, location);
+    }
```
Warriorrrr, line 49: The methods being moved around like this creates a lot of unnecessary diff.

Warriorrrr (changes requested):

paper-api/src/main/java/org/bukkit/block/Lidded.java
456dev: Just to confirm, it would be preferred to completely replace this API in 1 step?

I didn't want to create any breaking changes, and making it separate would allow for it to be cleanly removed separately at a later point.

Warriorrrr: What I'd prefer would be to add the methods to the existing API and leave the old methods the same, and changing the impl of the old methods to use the new api impl.

I'll wait for others to respond with their thoughts in case I'm not making sense.

lynxplay:
This is the third of fourth time I am reading through this PR and I think the surface area of this is just too large for me to be confortable with merging this.
The fact that we are duplicating seemingly large parts of the process between the shulker and containers screams for externalized handling.

The same can be said for the commented out craftbukkit diff. Just delete it, it makes the code feel even *more* duplicate.

`private final java.util.Set<Player> cancelledPlayer` feels dangerous.

But yea, I think the only sane way we will ever get a PR like this in is completely ripping out CB lidded diff in commit A and reintroducing a very much externalised lidded system into the affected types.
With that I mean `io.papermc.something.ContainerLiddedState` classes that do *all* the things that are currently copied around like maintaining the lidded state, openCounter etc.

As said, the giant diff in the CB types also does not help making this more reviewable but moving thins from patches into normal types to handle state would be a good start to this.

lynxplay:
Going to close this PR for now, feel free to reopen if / when you find time to heavily reduce / externalize the diff.
Might also be worth to first flesh out the entire API in an API only PR or whatnot given it'd be a shame if you put in a lot of internal work just to then have a lot of changes requested to the API surface.

PR #11816, opened 2024-12-25
MenuType API addition InventoryView Builders

Y2Kwastaken:
This Pull request ports InventoryView Builders from Spigot (Quite the Christmas Surprise!)
Relevant Spigot Pulls that this PR pulls from

Bukkit[
Add MenuType ViewBuilder](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/3d4295ccbf0bd9acb4ea93581f7b65cc9d2f018f#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Finventory%2Fview%2Fbuilder%2FInventoryViewBuilder.java)
[Improve ViewBuilder Chaining](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/ea0b2d1cf5f6b77b98414223a94fb0efa653cfa0#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Finventory%2Fview%2Fbuilder%2FLocationInventoryViewBuilder.java)
[Deprecate HumanEntity#open[Menu]; Add Server#createMerchant()
](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/ea073278d25761bee8961274d1278716ce49da99#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Fentity%2FHumanEntity.java)

CraftBukkit
[Add MenuType ViewBuilder](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/643b231025217ba902bad1b764f11e47a330fdc2#nms-patches%2Fnet%2Fminecraft%2Fworld%2Flevel%2Fblock%2FBlockChest.patch)
[Improve ViewBuilder Chaining](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/e28af087d58f90b8600ad24d2035ba18aed57d1f#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Fcraftbukkit%2Finventory%2Fview%2Fbuilder%2FCraftAbstractInventoryViewBuilder.java)
[Implement Server#createMerchant; Fix MerchantView opening inconsistency](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/76fd7c056d2bd0a311e62f9f676c0c468f3f920e#src%2Fmain%2Fjava%2Forg%2Fbukkit%2Fcraftbukkit%2Finventory%2Futil%2FCraftMenus.java)

**Rational**
The Server#createInventory and current HumanEntity#open[Container] APIs are currently extremely limited and are not comparably close. These methods built around previously very non consistent and broken behavior. They goal of this Pull Request is to take one step closer to deprecating the old Server#createInventory PR and InventoryType to allow for a much more well rounded and modern API. Previously implementing MenuType with a simple `#create(Component, HumanEntity)` method started a basis for this API.

**Why Builders? And not just more create methods?**
InventoryView builders for both me, and those I discussed this idea with, were the clear answer to building off the current MenuType API, for a couple of reasons. The reasons behind my rational to builders specifically is below

InventoryView builders are an easy API to use. It is much easier for an API consumer not to have to worry about different, possibly unknown casting types if something were to be delegated to the Server interface. Currently, this ambiguity is present with the Server#createInventory API, which leads some to believe you can cast to the "expected" type of inventory. Or, with the unrelated ItemMeta, it may be unclear if you can cast in a kind.

InventoryView builders are flexible with nonstandard InventoryView types. Builders allow nonstandard types, such as the current MerchantMenu builder, to be added with less stress about developing a lasting API. Even though many InventoryViews can be created with a location parameter, adding a MenuType#create(Location, Component, HumanEntity) makes a non-true assumption about all current MenuTypes and perpetuates this assumption into the future. ViewBuilders aims to be as long-lasting as possible, and most, if not all, other implementations of such an API would be exposed to the drawbacks of these unsafe assumptions.
 
Expanding the InventoryView builders API will lead to a better "Inventory" API. It is clear and has been for years, that Server#createInventory is simply dysfunctional. Future API could be added to view builders to allow for the integration of an Inventory parameter on select InventoryViews. For example, an API added to a view builder might look like this:
```java
MenuType.GENERIC_9X3.builder()
	.inventory(someFutureMethodToCreateInventories())
	.build(player)
```

Such an API is the way forward, as we should begin to move away from Server#createInventory methods. InventoryView builders allow for more explicit control over which InventoryViews are permitted to have an Inventory attached to them.

The last reason I see InventoryView builders as a good idea is their ability to "store" values. In the future, as people shift away from the old Server#createInventory API, some benefits will be a pain to lose. InventoryView titles aren't actually declared until they are opened. While a standard title could be saved as a constant, using a "saving" type method with some standard values could help API consumers have some standard opening protocol for specific menus they want to create. An example of such a concept is below:
```java
private static final LocationInventoryViewBuilder<InventoryView> STANDARD_CHEST = MenuType.GENERIC_9X3.builder().title(Component.text("This is some title that won't ever change or is defined somewhere")

public static InventoryView createStandardChest(HumanEntity player) {
	return STANDARD_CHEST.copy().location(player.getBlockExact(10)).build(player);
}
```

While maybe not a "super" realistic example, this portrays another rational behind InventoryView builders.

**Final Notes and Edits**

I'm not exactly sure what the "new" standard is on paper comments, especially when it comes to pulling from upstream, are they no longer needed? What's the deal there. Feedback in that facet would be appreciated

### DIFF
--- a/paper-api/src/main/java/org/bukkit/Server.java
+++ b/paper-api/src/main/java/org/bukkit/Server.java
@@ -43,6 +43,7 @@ import org.bukkit.inventory.InventoryHolder;
 import org.bukkit.inventory.ItemCraftResult;
 import org.bukkit.inventory.ItemFactory;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.MenuType;
 import org.bukkit.inventory.Merchant;
 import org.bukkit.inventory.Recipe;
 import org.bukkit.inventory.meta.ItemMeta;
@@ -1565,11 +1566,11 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * <br>
      * {@link InventoryType#WORKBENCH} will not process crafting recipes if
      * created with this method. Use
-     * {@link Player#openWorkbench(Location, boolean)} instead.
+     * {@link MenuType#CRAFTING} instead.
      * <br>
      * {@link InventoryType#ENCHANTING} will not process {@link ItemStack}s
      * for possible enchanting results. Use
-     * {@link Player#openEnchanting(Location, boolean)} instead.
+     * {@link MenuType#ENCHANTMENT} instead.
      *
      * @param owner the holder of the inventory, or null to indicate no holder
      * @param type the type of inventory to create
@@ -1592,11 +1593,11 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * <br>
      * {@link InventoryType#WORKBENCH} will not process crafting recipes if
      * created with this method. Use
-     * {@link Player#openWorkbench(Location, boolean)} instead.
+     * {@link MenuType#CRAFTING} instead.
      * <br>
      * {@link InventoryType#ENCHANTING} will not process {@link ItemStack}s
      * for possible enchanting results. Use
-     * {@link Player#openEnchanting(Location, boolean)} instead.
+     * {@link MenuType#ENCHANTMENT)} instead.
      *
      * @param owner The holder of the inventory; can be null if there's no holder.
      * @param type The type of inventory to create.
@@ -1620,11 +1621,11 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * <br>
      * {@link InventoryType#WORKBENCH} will not process crafting recipes if
      * created with this method. Use
-     * {@link Player#openWorkbench(Location, boolean)} instead.
+     * {@link MenuType#CRAFTING} instead.
      * <br>
      * {@link InventoryType#ENCHANTING} will not process {@link ItemStack}s
      * for possible enchanting results. Use
-     * {@link Player#openEnchanting(Location, boolean)} instead.
+     * {@link MenuType#ENCHANTMENT} instead.
      *
      * @param owner The holder of the inventory; can be null if there's no holder.
      * @param type The type of inventory to create.
@@ -1691,21 +1692,12 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * @param title the title of the corresponding merchant inventory, displayed
      * when the merchant inventory is viewed
      * @return a new merchant
+     * @deprecated The title parameter is no-longer needed when used with
+     * {@link MenuType#MERCHANT} and {@link MenuType.Typed#builder()}.
      */
+    @Deprecated(since = "1.21.4")
     @NotNull Merchant createMerchant(net.kyori.adventure.text.@Nullable Component title);
     // Paper start
-    /**
-     * Creates an empty merchant.
-     *
-     * @param title the title of the corresponding merchant inventory, displayed
-     * when the merchant inventory is viewed
-     * @return a new merchant
-     * @deprecated in favour of {@link #createMerchant(net.kyori.adventure.text.Component)}
-     */
-    @NotNull
-    @Deprecated // Paper
-    Merchant createMerchant(@Nullable String title);
-
     /**
      * Gets the amount of consecutive neighbor updates before skipping
      * additional ones.
@@ -1714,6 +1706,27 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * negative then the limit it's not used
      */
     int getMaxChainedNeighborUpdates();
+
+    /**
+     * Creates an empty merchant.
+     *
+     * @param title the title of the corresponding merchant inventory, displayed
+     * when the merchant inventory is viewed
+     * @return a new merchant
+     * @deprecated in favour of {@link #createMerchant(net.kyori.adventure.text.Component)}, The title parameter is
+     * no-longer needed when used with {@link MenuType#MERCHANT} and {@link MenuType.Typed#builder()}.
+     */
+    @NotNull
+    @Deprecated // Paper
+    Merchant createMerchant(@Nullable String title);
+
+    /**
+     * Creates an empty merchant.
+     *
+     * @return a new merchant
+     */
+    @NotNull
+    Merchant createMerchant();
 
     /**
      * Gets user-specified limit for number of monsters that can spawn in a
--- a/paper-api/src/main/java/org/bukkit/Server.java
+++ b/paper-api/src/main/java/org/bukkit/Server.java
@@ -1699,15 +1699,6 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
     @NotNull Merchant createMerchant(net.kyori.adventure.text.@Nullable Component title);
     // Paper start
     /**
-     * Gets the amount of consecutive neighbor updates before skipping
-     * additional ones.
-     *
-     * @return the amount of consecutive neighbor updates, if the value is
-     * negative then the limit it's not used
-     */
-    int getMaxChainedNeighborUpdates();
-
-    /**
      * Creates an empty merchant.
      *
      * @param title the title of the corresponding merchant inventory, displayed
@@ -1719,6 +1710,15 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
     @NotNull
     @Deprecated // Paper
     Merchant createMerchant(@Nullable String title);
+
+    /**
+     * Gets the amount of consecutive neighbor updates before skipping
+     * additional ones.
+     *
+     * @return the amount of consecutive neighbor updates, if the value is
+     * negative then the limit it's not used
+     */
+    int getMaxChainedNeighborUpdates();
 
     /**
      * Creates an empty merchant.
--- a/paper-api/src/main/java/org/bukkit/Server.java
+++ b/paper-api/src/main/java/org/bukkit/Server.java
@@ -1597,7 +1597,7 @@ public interface Server extends PluginMessageRecipient, net.kyori.adventure.audi
      * <br>
      * {@link InventoryType#ENCHANTING} will not process {@link ItemStack}s
      * for possible enchanting results. Use
-     * {@link MenuType#ENCHANTMENT)} instead.
+     * {@link MenuType#ENCHANTMENT} instead.
      *
      * @param owner The holder of the inventory; can be null if there's no holder.
      * @param type The type of inventory to create.
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
@@ -1,0 +1,53 @@
+package org.bukkit.inventory.view.builder;
+
+import net.kyori.adventure.text.Component;
+import org.bukkit.entity.HumanEntity;
+import org.bukkit.inventory.InventoryView;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Generic Builder for InventoryView's with no special attributes or parameters
+ *
+ * @param <V> the type of InventoryView created from this builder
+ */
+@ApiStatus.Experimental
+public interface InventoryViewBuilder<V extends InventoryView> {
+
+    /**
+     * Makes a copy of this builder
+     *
+     * @return a copy of this builder
+     */
+    @NotNull
+    InventoryViewBuilder<V> copy();
+
+    /**
+     * Sets the title of the builder
+     *
+     * @param title the title
+     * @return this builder
+     * @deprecated use {@link #title(Component)} instead.
+     */
+    @Deprecated(forRemoval = true)
+    @NotNull
+    InventoryViewBuilder<V> title(@NotNull final String title);
+
+    /**
+     * Sets the title of the builder
+     *
+     * @param title the title
+     * @return this builder
+     */
+    @NotNull
+    InventoryViewBuilder<V> title(@NotNull final Component title);
+
+    /**
+     * Builds this builder into a InventoryView
+     *
+     * @param player the player to assign to the view
+     * @return the created InventoryView
+     */
+    @NotNull
+    V build(@NotNull final HumanEntity player);
+}
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
@@ -4,7 +4,6 @@ import net.kyori.adventure.text.Component;
 import org.bukkit.entity.HumanEntity;
 import org.bukkit.inventory.InventoryView;
 import org.jetbrains.annotations.ApiStatus;
-import org.jetbrains.annotations.NotNull;
 
 /**
  * Generic Builder for InventoryView's with no special attributes or parameters
@@ -19,7 +18,6 @@ public interface InventoryViewBuilder<V extends InventoryView> {
      *
      * @return a copy of this builder
      */
-    @NotNull
     InventoryViewBuilder<V> copy();
 
     /**
@@ -27,20 +25,8 @@ public interface InventoryViewBuilder<V extends InventoryView> {
      *
      * @param title the title
      * @return this builder
-     * @deprecated use {@link #title(Component)} instead.
      */
-    @Deprecated(forRemoval = true)
-    @NotNull
-    InventoryViewBuilder<V> title(@NotNull final String title);
-
-    /**
-     * Sets the title of the builder
-     *
-     * @param title the title
-     * @return this builder
-     */
-    @NotNull
-    InventoryViewBuilder<V> title(@NotNull final Component title);
+    InventoryViewBuilder<V> title(final Component title);
 
     /**
      * Builds this builder into a InventoryView
@@ -48,6 +34,5 @@ public interface InventoryViewBuilder<V extends InventoryView> {
      * @param player the player to assign to the view
      * @return the created InventoryView
      */
-    @NotNull
-    V build(@NotNull final HumanEntity player);
+    V build(final HumanEntity player);
 }
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
@@ -1,0 +1,63 @@
+package org.bukkit.inventory.view.builder;
+
+import net.kyori.adventure.text.Component;
+import org.bukkit.Location;
+import org.bukkit.inventory.InventoryView;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * An InventoryViewBuilder that can be bound by location within the world
+ *
+ * @param <V> the type of InventoryView created from this builder
+ */
+@ApiStatus.Experimental
+public interface LocationInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> copy();
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> title(final @NotNull Component title);
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> title(final @NotNull String title);
+
+    /**
+     * Determines whether or not the server should check if the player can reach
+     * the location.
+     * <p>
+     * Not providing a location but setting checkReachable to true will
+     * automatically close the view when opened.
+     * <p>
+     * If checkReachable is set to false and a location is set on the builder if
+     * the target block exists and this builder is the correct menu for that
+     * block, e.g. MenuType.GENERIC_9X3 builder and target block set to chest,
+     * if that block is destroyed the view would persist.
+     *
+     * @param checkReachable whether or not to check if the view is "reachable"
+     * @return this builder
+     */
+    @NotNull
+    LocationInventoryViewBuilder<V> checkReachable(final boolean checkReachable);
+
+    /**
+     * Binds a location to this builder.
+     * <p>
+     * By binding a location in an unloaded chunk to this builder it is likely
+     * that the given chunk the location is will load. That means that when,
+     * building this view it may come with the costs associated with chunk
+     * loading.
+     * <p>
+     * Providing a location of a tile entity with a non matching menu comes with
+     * extra costs associated with ensuring that the correct view is created.
+     *
+     * @param location the location to bind to this view
+     * @return this builder
+     */
+    @NotNull
+    LocationInventoryViewBuilder<V> location(@NotNull final Location location);
+}
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
@@ -14,17 +14,11 @@ import org.jetbrains.annotations.NotNull;
 @ApiStatus.Experimental
 public interface LocationInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
 
-    @NotNull
     @Override
     LocationInventoryViewBuilder<V> copy();
 
-    @NotNull
     @Override
     LocationInventoryViewBuilder<V> title(final @NotNull Component title);
-
-    @NotNull
-    @Override
-    LocationInventoryViewBuilder<V> title(final @NotNull String title);
 
     /**
      * Determines whether or not the server should check if the player can reach
@@ -41,7 +35,6 @@ public interface LocationInventoryViewBuilder<V extends InventoryView> extends I
      * @param checkReachable whether or not to check if the view is "reachable"
      * @return this builder
      */
-    @NotNull
     LocationInventoryViewBuilder<V> checkReachable(final boolean checkReachable);
 
     /**
@@ -58,6 +51,5 @@ public interface LocationInventoryViewBuilder<V extends InventoryView> extends I
      * @param location the location to bind to this view
      * @return this builder
      */
-    @NotNull
-    LocationInventoryViewBuilder<V> location(@NotNull final Location location);
+    LocationInventoryViewBuilder<V> location(final Location location);
 }
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/MerchantInventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/MerchantInventoryViewBuilder.java
@@ -1,0 +1,53 @@
+package org.bukkit.inventory.view.builder;
+
+import net.kyori.adventure.text.Component;
+import org.bukkit.Server;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.Merchant;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * An InventoryViewBuilder for creating merchant views
+ *
+ * @param <V> the type of InventoryView created by this builder
+ */
+@ApiStatus.Experimental
+public interface MerchantInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
+
+    @NotNull
+    @Override
+    MerchantInventoryViewBuilder<V> copy();
+
+    @NotNull
+    @Override
+    MerchantInventoryViewBuilder<V> title(final @NotNull Component title);
+
+    @NotNull
+    @Override
+    MerchantInventoryViewBuilder<V> title(final @NotNull String title);
+
+
+    /**
+     * Adds a merchant to this builder
+     *
+     * @param merchant the merchant
+     * @return this builder
+     */
+    @NotNull
+    MerchantInventoryViewBuilder<V> merchant(@NotNull final Merchant merchant);
+
+    /**
+     * Determines whether or not the server should check if the player can reach
+     * the location.
+     * <p>
+     * Given checkReachable is provided and a virtual merchant is provided to
+     * the builder from {@link Server#createMerchant(net.kyori.adventure.text.Component)} this method will
+     * have no effect on the actual menu status.
+     *
+     * @param checkReachable whether or not to check if the view is "reachable"
+     * @return this builder
+     */
+    @NotNull
+    MerchantInventoryViewBuilder<V> checkReachable(final boolean checkReachable);
+}
--- a/paper-api/src/main/java/org/bukkit/inventory/view/builder/MerchantInventoryViewBuilder.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/view/builder/MerchantInventoryViewBuilder.java
@@ -15,18 +15,11 @@ import org.jetbrains.annotations.NotNull;
 @ApiStatus.Experimental
 public interface MerchantInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
 
-    @NotNull
     @Override
     MerchantInventoryViewBuilder<V> copy();
 
-    @NotNull
     @Override
     MerchantInventoryViewBuilder<V> title(final @NotNull Component title);
-
-    @NotNull
-    @Override
-    MerchantInventoryViewBuilder<V> title(final @NotNull String title);
-
 
     /**
      * Adds a merchant to this builder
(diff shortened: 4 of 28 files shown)


paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Sets the title of the builder
+     *
+     * @param title the title
+     * @return this builder
+     * @deprecated use {@link #title(Component)} instead.
+     */
+    @Deprecated(forRemoval = true)
+    @NotNull
+    InventoryViewBuilder<V> title(@NotNull final String title);
```
NonSwag: Why add deprecated methods?
You already have components for titles so no need for legacy methods anyway, or am I missing something?

paper-server/patches/sources/net/minecraft/world/inventory/ChestMenu.java.patch
```diff
@@ -26,14 +26,21 @@
...
      private ChestMenu(MenuType<?> type, int containerId, Inventory playerInventory, int rows) {
          this(type, containerId, playerInventory, new SimpleContainer(9 * rows), rows);
-@@ -52,6 +_,9 @@
+@@ -51,7 +_,10 @@
+         checkContainerSize(container, rows * 9);
          this.container = container;
          this.containerRows = rows;
-         container.startOpen(playerInventory.player);
+-        container.startOpen(playerInventory.player);
++        // container.startOpen(playerInventory.player); // CraftBukkit - don't startOpen until menu actually opens
```
NonSwag: why // CraftBukkit instead of // Paper ?

paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
```diff
@@ -2485,6 +2486,11 @@ public Merchant createMerchant(String title) {
         return new CraftMerchantCustom(title == null ? InventoryType.MERCHANT.getDefaultTitle() : title);
     }
 
+    @Override
+    public @NotNull Merchant createMerchant() {
+        return new CraftMerchantCustom(net.kyori.adventure.text.Component.text(""));
```
NonSwag: wouldn't Component.empty() be smarter?

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.MenuType;
+import org.bukkit.craftbukkit.entity.CraftHumanEntity;
+import org.bukkit.craftbukkit.util.CraftChatMessage;
+import org.bukkit.entity.HumanEntity;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.view.builder.InventoryViewBuilder;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class CraftAbstractInventoryViewBuilder<V extends InventoryView> implements InventoryViewBuilder<V> {
```
NonSwag, line 15: this class can be null marked instead of everything annotated as nonnull

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+    protected boolean checkReachable = false;
+    protected Component title = null;
+
+    public CraftAbstractInventoryViewBuilder(MenuType<?> handle) {
+        this.handle = handle;
+    }
+
+    @NotNull
+    @Override
+    public InventoryViewBuilder<V> title(@NotNull final String title) {
```
NonSwag: again, why legacy text?

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+        this.title = LegacyComponentSerializer.legacySection().deserialize(title);
+        return this;
+    }
+
+    @Override
+    public @NotNull InventoryViewBuilder<V> title(final @NotNull Component title) {
```
NonSwag: title should not be forced
vanilla has for every inventory type a title already which should be displayed if title is null

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+        this.title = title;
+        return this;
+    }
+
+    @Override
+    public V build(final HumanEntity player) {
+        Preconditions.checkArgument(player != null, "The given player must not be null");
+        Preconditions.checkArgument(this.title != null, "The given title must not be null");
```
NonSwag, line 36: dont force titles

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+        Preconditions.checkArgument(player instanceof CraftHumanEntity, "The given player must be a CraftHumanEntity");
+        final CraftHumanEntity craftHuman = (CraftHumanEntity) player;
+        Preconditions.checkArgument(craftHuman.getHandle() instanceof ServerPlayer, "The given player must be an EntityPlayer");
+        final ServerPlayer serverPlayer = (ServerPlayer) craftHuman.getHandle();
+        final AbstractContainerMenu container = buildContainer(serverPlayer);
+        container.checkReachable = this.checkReachable;
+        container.setTitle(PaperAdventure.asVanilla(this.title));
```
NonSwag, line 43: just dont set the title if null

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftMerchantInventoryViewBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+        super(handle);
+    }
+
+    @Override
+    public MerchantInventoryViewBuilder<V> title(final @NotNull String title) {
+        return (MerchantInventoryViewBuilder<V>) super.title(title);
+    }
+
+    @Override
+    public MerchantInventoryViewBuilder<V> title(final @NotNull Component title) {
```
NonSwag: dont force titles

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+    protected Level world;
+    protected BlockPos position;
+
+    public CraftAbstractLocationInventoryViewBuilder(final MenuType<?> handle) {
+        super(handle);
+    }
+
+    @Override
+    public @NotNull LocationInventoryViewBuilder<V> title(final @NotNull String title) {
```
NonSwag: legacy

paper-api/src/main/java/org/bukkit/inventory/view/builder/MerchantInventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+@ApiStatus.Experimental
+public interface MerchantInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
+
+    @NotNull
+    @Override
+    MerchantInventoryViewBuilder<V> copy();
+
+    @NotNull
+    @Override
+    MerchantInventoryViewBuilder<V> title(final @NotNull Component title);
```
NonSwag, line 22: dont force titles

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    LocationInventoryViewBuilder<V> copy();
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> title(final @NotNull Component title);
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> title(final @NotNull String title);
```
NonSwag: legacy

paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Sets the title of the builder
+     *
+     * @param title the title
+     * @return this builder
+     */
+    @NotNull
+    InventoryViewBuilder<V> title(@NotNull final Component title);
```
NonSwag: dont force titles

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+        return (LocationInventoryViewBuilder<V>) super.title(title);
+    }
+
+    @Override
+    public LocationInventoryViewBuilder<V> title(final @NotNull Component title) {
```
NonSwag: dont force titles

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+        return (LocationInventoryViewBuilder<V>) super.title(title);
+    }
+
+    @Override
+    public LocationInventoryViewBuilder<V> copy() {
+        throw new UnsupportedOperationException("copy is not implemented on CraftAbstractLocationInventoryViewBuilder");
```
NonSwag, line 31: why not implemented?
is it planned or not?

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftMerchantInventoryViewBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+public class CraftMerchantInventoryViewBuilder<V extends InventoryView> extends CraftAbstractInventoryViewBuilder<V> implements MerchantInventoryViewBuilder<V> {
+
+    private net.minecraft.world.item.trading.Merchant merchant;
+
+    public CraftMerchantInventoryViewBuilder(final MenuType<?> handle) {
+        super(handle);
+    }
+
+    @Override
+    public MerchantInventoryViewBuilder<V> title(final @NotNull String title) {
```
NonSwag: legacy

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftMerchantInventoryViewBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        container.checkReachable = super.checkReachable;
+        container.setTitle(PaperAdventure.asVanilla(this.title));
+        return (V) container.getBukkitView();
+    }
+
+    @Override
+    protected AbstractContainerMenu buildContainer(final ServerPlayer player) {
+        throw new UnsupportedOperationException("buildContainer is not supported for CraftMerchantInventoryViewBuilder");
```
NonSwag, line 67: again, not implemented, why?

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.world.inventory.MenuType;
+import net.minecraft.world.level.Level;
+import org.bukkit.Location;
+import org.bukkit.craftbukkit.CraftWorld;
+import org.bukkit.craftbukkit.util.CraftLocation;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.view.builder.LocationInventoryViewBuilder;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class CraftAbstractLocationInventoryViewBuilder<V extends InventoryView> extends CraftAbstractInventoryViewBuilder<V> implements LocationInventoryViewBuilder<V> {
```
NonSwag, line 15: having "Abstract" _or any implementation details_ in class names feels weird 
renaming this class to CraftLocationInventoryBuilder would be a better name

NonSwag, line 15: @nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAccessLocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.ContainerLevelAccess;
+import net.minecraft.world.inventory.MenuType;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.PlayerInventory;
+import org.bukkit.inventory.view.builder.LocationInventoryViewBuilder;
+
+public class CraftAccessLocationInventoryViewBuilder<V extends InventoryView> extends CraftAbstractLocationInventoryViewBuilder<V> {
```
NonSwag, line 11: nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftStandardInventoryViewBuilder.java
```diff
@@ -0,0 +1,26 @@
+package org.bukkit.craftbukkit.inventory.view.builder;
+
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.MenuType;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.view.builder.InventoryViewBuilder;
+
+public class CraftStandardInventoryViewBuilder<V extends InventoryView> extends CraftAbstractInventoryViewBuilder<V> {
```
NonSwag, line 9: nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftMerchantInventoryViewBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+import org.bukkit.craftbukkit.entity.CraftHumanEntity;
+import org.bukkit.craftbukkit.inventory.CraftMerchant;
+import org.bukkit.craftbukkit.inventory.CraftMerchantCustom;
+import org.bukkit.entity.HumanEntity;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.Merchant;
+import org.bukkit.inventory.view.builder.MerchantInventoryViewBuilder;
+import org.jetbrains.annotations.NotNull;
+
+public class CraftMerchantInventoryViewBuilder<V extends InventoryView> extends CraftAbstractInventoryViewBuilder<V> implements MerchantInventoryViewBuilder<V> {
```
NonSwag, line 19: nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftDoubleChestInventoryViewBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.MenuType;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.ChestBlock;
+import net.minecraft.world.level.block.DoubleBlockCombiner;
+import net.minecraft.world.level.block.entity.ChestBlockEntity;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.view.builder.LocationInventoryViewBuilder;
+
+public class CraftDoubleChestInventoryViewBuilder<V extends InventoryView> extends CraftAbstractLocationInventoryViewBuilder<V> {
```
NonSwag, line 14: nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.MenuConstructor;
+import net.minecraft.world.inventory.MenuType;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+import org.bukkit.inventory.InventoryView;
+import org.bukkit.inventory.view.builder.LocationInventoryViewBuilder;
+
+public class CraftBlockEntityInventoryViewBuilder<V extends InventoryView> extends CraftAbstractLocationInventoryViewBuilder<V> {
```
NonSwag, line 16: nullmarked

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
NonSwag, line 15: rename to CraftInventoryViewBuilder

paper-api/src/main/java/org/bukkit/inventory/view/builder/package-info.java
NonSwag, line 1: does paper even use package information at all?
if not, this seems redundant and'd be better off in each class

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+@ApiStatus.Experimental
+public interface LocationInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> copy();
+
+    @NotNull
+    @Override
+    LocationInventoryViewBuilder<V> title(final @NotNull Component title);
```
NonSwag, line 21: dont enforce titles

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * An InventoryViewBuilder that can be bound by location within the world
+ *
+ * @param <V> the type of InventoryView created from this builder
+ */
+@ApiStatus.Experimental
+public interface LocationInventoryViewBuilder<V extends InventoryView> extends InventoryViewBuilder<V> {
```
NonSwag, line 15: nullmarked

paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Generic Builder for InventoryView's with no special attributes or parameters
+ *
+ * @param <V> the type of InventoryView created from this builder
+ */
+@ApiStatus.Experimental
+public interface InventoryViewBuilder<V extends InventoryView> {
```
NonSwag, line 14: nullmarked

NonSwag (commented):
Just wanted to leave my thoughts on some things I find a bit iffy
other than that, I'd love this getting merged, great work
_side note: I didn't read everything in order, so the comments might seem random, but the explanation for it is somewhere else_
edit: after reading the description i feel a bit stupid now

paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
Y2Kwastaken: This is up for internal discussion, regarding spigot compatibility this PR is a 1 to 1 copy of an upstream feature I added thus I copied it one for one and instead added adventure components. Its very likely a core team discussion will dictate the future of these methods. I am not here to make these decisions.

paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
Y2Kwastaken: I'll take a look at testing what Component.empty() does when used in this scenario, but in all likelihood yes probably

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
Y2Kwastaken: No its an abstract class, Naming conventions for abstract classes should include that they are abstract. This creates more clarity that it is not a standalone and is dependent on some further implementation

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
Y2Kwastaken: Not sure what you mean by this but if you're suggesting I shouldn't enforce passing a title that's not possible. Maybe I could create a util method of default paths for vanilla Menu names but that's another maintenance burden. InventoryType makes this seem trivial when its actually kind of annoying

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
Y2Kwastaken: See the actual implementations of CraftAbstractLocationInventoryViewBuilder, no its not planned.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractLocationInventoryViewBuilder.java
Y2Kwastaken: No if you reuse some constant implementation its not weird at all, it makes sense

Y2Kwastaken: I'll leave this to the core team if we go with jspecify or not I'm not going to make that call I simply ported how things are. If they tell me to head over and use Jspecify I have no such issue with that but I won't be making changes until they tell me to

paper-server/patches/sources/net/minecraft/world/inventory/ChestMenu.java.patch
Y2Kwastaken: These changes are from upstream if there is a general consensus I should switch the comments to Paper I will do so.

lynxplay: Please do yes :+1:

paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
NonSwag: If the title will becomes nullable, there will be no need for empty component

lynxplay: Titles for these menus are *not* part of the menu themselves.
They are provided by the respective menu providers, which can be *anything*.
They are not innate to the menu type itself.

A nullable title (while nice for the consumer) is a *LOT* of // diff on change work across the entire code base + hoping that won't change. I am not too convinced a nullable title makes sense here.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
DerEchtePilz: > No its an abstract class, Naming conventions for abstract classes should include that they are abstract. This creates more clarity that it is not a standalone and is dependent on some further implementation

There is no such convention. This creates a redundancy since you use the abstract keyword anyway and it's not even API so it's even less important to follow some convention you set for yourself.
Also you make the point of "This creates more clarity that it's not a standalone". This makes no sense as the presence of the `abstract` keyword already says as much. You can't instantiate an abstract class. Therefore you cannot really use an abstract class. An abstract class never is a standalone.

Y2Kwastaken: Whether or not you like the naming stripping the word `Abstract` from this class name furthers itself from a standard of other internal classes, feel free to look yourself to confirm, but also doesn't actually improve the content of the PR. Not worth continuing a debate over such a minutely unimportant detail that follows an arguably useful common distinction already made within the internals.

NonSwag: Not wanna fuel the fire here but even if it is a "standard" in the current state of the project it doesn't have to be continued
Just saying, when/if this gets merged, it is worth adhering to proper standards.

paper-api/src/main/java/org/bukkit/inventory/view/builder/package-info.java
Machine-Maker: I do want to start doing this, for documentation, but also putting NullMarked inside the package-info.java makes it apply to all the classes in that package (not recursively) which is nice as well.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftAbstractInventoryViewBuilder.java
Machine-Maker: `Abstract` seems fine to me. I followed a similar convention all over the lifecycle event system implementation.

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
Machine-Maker: Yeah, the server doesn't really have a relationship between menutype and default titles. Figuring out how to expose those titles somehow is a different problem, plus all the translation keys are all public anyways, people can go look through them. So I'm fine with requiring a title.

paper-api/src/main/java/org/bukkit/inventory/view/builder/InventoryViewBuilder.java
lynxplay: Yea, we won't need legacy string methods :+1:

paper-api/src/main/java/org/bukkit/inventory/view/builder/LocationInventoryViewBuilder.java
NonSwag: If the title is not defined the client will use the default one
so I am in favor of nullable titles

paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
masmc05: Also, shouldn't this component be a proper import? It's not a patch anymore

lynxplay (changes requested):
Two points:

 - While maybe not the most correct - preserving event order for open event / lid event means we need to split the openMenu calls out of initMenu() and have them be called *right* before the 5 places inventory open events are fired / initMenu methods are called.
 
 - While there is 100% not a default title that randomly appears, the nature of the menu provider specific builder implementations (BE, LocationAccess etc) does allow us to technically construct the MenuProvider for each and fetch the title from there.
 This way, LocationAccess Builder impl should get a reference to the nms.Block the menu will be sourced from so it can call the `BlockBehaviour#getMenuProvider` and source a default title from there.

paper-api/src/main/java/org/bukkit/Server.java
```diff
@@ -1655,29 +1656,41 @@ default World getWorld(@NotNull NamespacedKey worldKey) {
...
     @NotNull Merchant createMerchant(net.kyori.adventure.text.@Nullable Component title);
     // Paper start
+    /**
+     * Gets the amount of consecutive neighbor updates before skipping
+     * additional ones.
+     *
+     * @return the amount of consecutive neighbor updates, if the value is
+     * negative then the limit it's not used
+     */
+    int getMaxChainedNeighborUpdates();
```
masmc05: Why did you move `getMaxChainedNeighborUpdates()` from bellow `createMerchant(String)` to above it? It's a really weird diff

Y2Kwastaken: probably fat fingered a keybind

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/util/CraftMenus.java
```diff
@@ -34,83 +42,116 @@
...
         if (menuType == MenuType.LOOM) {
-            return CraftMenus.asType(new MenuTypeData<>(LoomView.class, CraftMenus.STANDARD));
+            return asType(new MenuTypeData<>(LoomView.class, () -> new CraftStandardInventoryViewBuilder<>(handle)));
         }
         if (menuType == MenuType.MERCHANT) {
-            return CraftMenus.asType(new MenuTypeData<>(MerchantView.class, CraftMenus.STANDARD));
+            return asType(new MenuTypeData<>(MerchantView.class, () -> new CraftMerchantInventoryViewBuilder<>(handle)));
+        }
+        if (menuType == MenuType.SHULKER_BOX) {
+            return asType(new MenuTypeData<>(InventoryView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.SHULKER_BOX, null)));
```
Lulu13022002: Any reason to not have an underlying Shulkerbox block entity? this means getType on the view doesn't return the right type and putting a shulkerbox inside the container has a weird behavior.

Y2Kwastaken: It isn't needed in this case as the block entity Is only used for faking it's open. See Anvil impl for example where it's used. In the case of opening a real shulker box the block entity is obtained from the world. And to be completely honest I don't care about InventoryType. This may be controversial but I personally have no plans to build around API this is meant to replace. I'll look into it, but no promises and almost certainly not my priority.

lynxplay: We should 100% fix the fact that a shulker menu from an API that is 100% focused around menu types is opened as a generic 9x3 because of CB code relying on the inventory type to figure out the MenuType.

Y2Kwastaken: Is it opened as a 9x3? I was under the impression only the InventoryType was incorrect. I'll take a look when I get a chance.

lynxplay: Yea, CraftHumanEnt#openInventory(View) has the fun bit of logic that is
```java
// Now open the window
MenuType<?> windowType = CraftContainer.getNotchInventoryType(inventory.getTopInventory());
```
so the type relies on the inventory type

Y2Kwastaken: That method is a real sore spot. In this case I must pass the entity until we go back and replace that internally. Easy fix I guess when I implemented it I ignored it because I didn't care about the distinction with the fake menus. Either way yeah sloppy as hell impl and easy solution.

lynxplay (approved):

Machine-Maker (approved):

PR #11819, opened 2024-12-25
Fix framed maps looping through all players in world

Warriorrrr:
This diff used to be part of CB ([link](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/server/level/EntityTrackerEntry.patch#54)) but seems to have been dropped, this fixes tickCarriedBy being called & update packets being sent unnecessarily to players who are not in range of the framed map.


kennytv (approved):
+1 for lynx

lynxplay:
Catching up one lost hunk at a time 💪

PR #11820, opened 2024-12-25
Add a way to check if pathfinder allows the mob to fly

GliczDev:
There's no universal way to check if mob can fly and the only shared logic between flying mobs in NMS is `FlyingPathNavigation`.
Not sure about the method name, but should be ok

### DIFF
--- a/paper-api/src/main/java/com/destroystokyo/paper/entity/Pathfinder.java
+++ b/paper-api/src/main/java/com/destroystokyo/paper/entity/Pathfinder.java
@@ -181,6 +181,13 @@ public interface Pathfinder {
     void setCanFloat(boolean canFloat);
 
     /**
+     * Checks if this pathfinder allows the mob to fly
+     *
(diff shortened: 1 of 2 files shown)


kennytv (approved):

GliczDev:
ok, so I just found out that there is also `FlyingMoveControl`.
is the current implementation ok, or should I add `Mob#canFly` and check for `FlyingMoveControl`?

lynxplay:
Yea, Phantoms and Ghasts don't use it either.
This seems like a bit of an impossible task to properly do.
This is at best gonna be an EntityTag once we have the ability to properly define custom paper tags, but for now I don't think a new API method like this makes sense given the server has many definitions for flying entities (e.g. FlyingMob vs FlyingPathNavigation vs whatever they may come up next).

Thank you anyway for the PR, does highlight that we should work towards custom tags maybe somewhat soon :)

PR #11823, opened 2024-12-25
Deprecate #furnace_materials item tag

Lulu13022002:
The tag has been removed since 1.16.2


Machine-Maker (approved):
longer term I think we should just deprecate this entire thing and start creating separate files for the tag constants. Similar to that PR that upstream has to split up the tags (but we should use our Tag type instead of bukkits)

lynxplay (approved):

PR #11831, opened 2024-12-26
Remove itemstack mutation on player drop method

### LINKED ISSUE #695: 我无法获得"华丽一剪"成就
### 服务端版本 Leaves Version

1.21.8最新测试版

### 预期的行为 Expected behavior

在为驯服的狼用剪刀剥落狼铠时获得成就"华丽一剪"

### 观测到的实际行为 Observed/Actual behavior

在为驯服的狼用剪刀剥落狼铠时没有获得成就

### 复现步骤 Steps/models to reproduce

1准备一只狼

Image

2驯服它

Image

3为它装备狼铠

Image

4用剪刀剥离

Image

5没有获得成就

Image

### 更多详细信息。 The more information.

无数据包，插件为AuthMe, FastLogin, LuckPerms, ProtocolLib, voicechat,AuthMeVelocity
外面套了一层velocity

### LINKED ISSUE #11520: Regression: PlayerDeathEvent custom drops are mutated into AIR by the server after the event is handled

### LINKED ISSUE #11765: [1.21.3 & 1.21.4] Unable to bulk drop craft
### Expected behavior

The entire input is crafted and dropped on the ground.

### Observed/Actual behavior

Acts as if q was pushed on its own and crafts a single result and drops it.

### Steps/models to reproduce

- Open a crafting table
- Place 2 or more logs in the crafting table
- Push `Ctrl + Q` while hovering over the planks

### Plugin and Datapack List

```
/plugins
Server Plugins (0):

/datapack list
```

### Paper version

This server is running Paper version 1.21.3-81-master@da71382 (2024-12-03T14:58:13Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
This server is running Paper version 1.21.4-15-master@bb76819 (2024-12-15T09:48:21Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

### LINKED ISSUE #12769: Shearing Brilliance Advancement
### Expected behavior

Expected to get advancement upon shearing armored wolf

### Observed/Actual behavior

Advancement is not given when shearing armored wolf

### Steps/models to reproduce

1. Befriend wolf
2. Put wolf armor on wolf
3. Shear the wolf

### Plugin and Datapack List

Bolt, Luckperms

### Paper version

1.21.6 Build 46

### Other

_No response_

Owen1212055:
This appears to cause alot of issues, as the general assumption when calling this method is that the itemstack would not be mutated.

Common issues that this seems to cause is item related actions that later increment a statistic, which is no longer true because the item type changes since its now considered empty.
The dupe that this was meant to resolve was patched.

Fixes #11520 
Fixes #11765
Supercedes https://github.com/PaperMC/Paper/pull/11521

### DIFF
--- a/paper-server/patches/features/0001-Moonrise-optimisation-patches.patch
+++ b/paper-server/patches/features/0001-Moonrise-optimisation-patches.patch
@@ -27025,7 +27025,7 @@ index f25d50bcccf9b35ae70f8162c6b2b67fffcd319a..5494b92ab1c3c202a640e483e8a4bcb6
      }
  
 diff --git a/net/minecraft/server/level/ServerPlayer.java b/net/minecraft/server/level/ServerPlayer.java
-index b09e6a9fa5fbe9a0d3df615d02109992612b4416..8688d1eb7427b8d07757af84d615781867a34288 100644
+index 09470e7c4238a66be16b3cf8cf93444700cc2508..55d27aa4cff91ea61ff2ab9256acd2cd8694304a 100644
 --- a/net/minecraft/server/level/ServerPlayer.java
 +++ b/net/minecraft/server/level/ServerPlayer.java
 @@ -201,7 +201,7 @@ import net.minecraft.world.scores.criteria.ObjectiveCriteria;
@@ -28123,7 +28123,7 @@ index 8cc5c0716392ba06501542ff5cbe71ee43979e5d..09fd99c9cbd23b5f3c899bfb00c9b896
 +    // Paper end - block counting
  }
 diff --git a/net/minecraft/world/entity/Entity.java b/net/minecraft/world/entity/Entity.java
-index 7cc307666b7b4e4597673852ba456034fcef7590..69020a6e12254b935ccc0a06a73357c64b7e9d1c 100644
+index f88ac0961d2474ed0d3c18ff77e789c0ad1c3b50..c272211766d931a5ee3524567781dd41e9dbca20 100644
 --- a/net/minecraft/world/entity/Entity.java
 +++ b/net/minecraft/world/entity/Entity.java
 @@ -152,7 +152,7 @@ import org.jetbrains.annotations.Contract;
@@ -28476,7 +28476,7 @@ index 7cc307666b7b4e4597673852ba456034fcef7590..69020a6e12254b935ccc0a06a73357c6
      }
  
      private static float[] collectCandidateStepUpHeights(AABB box, List<VoxelShape> colliders, float deltaY, float maxUpStep) {
-@@ -2708,21 +2856,110 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
+@@ -2707,21 +2855,110 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
      }
  
      public boolean isInWall() {
@@ -28598,7 +28598,7 @@ index 7cc307666b7b4e4597673852ba456034fcef7590..69020a6e12254b935ccc0a06a73357c6
      }
  
      public InteractionResult interact(Player player, InteractionHand hand) {
-@@ -4313,15 +4550,17 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
+@@ -4312,15 +4549,17 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
      }
  
      public Iterable<Entity> getIndirectPassengers() {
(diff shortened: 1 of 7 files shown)


Owen1212055:
I have tried to look deeply into all of the changes that may be caused by this and I have seen the following, note that I would like a second opinion from some of these. Note that for analyizing these you must look at the parameter passed and ensure that the itemstack is either cloned or removed from the container properly before being called.

Some interesting calls:
- ResultSlot.onTake(Player, ItemStack) (net.minecraft.world.inventory) — This looks like where the above bug fix occurs?
- Inventory.placeItemBackInInventory(ItemStack, boolean) (net.minecraft.world.entity.player) — looked suspicious, altough is fine
- StonecutterMenu.quickMoveStack(Player, int) (net.minecraft.world.inventory) — This is sus, there is defo mutation in that slot occuring here. Do we cause another bug here?
- Bucketable.bucketMobPickup(Player, InteractionHand, T) (net.minecraft.world.entity.animal) — behavior change, bucketItemStack will now respect in advancement
- AbstractContainerMenu#614 -mutation affects outer loop

Owen1212055:
Also fixes the item drop statistic

Hurming:
I have same issue with it version server 1.21.7

electronicboy (approved):

Machine-Maker (approved):

PR #11834, opened 2024-12-27
Better Stats API

Machine-Maker:
Replaces https://github.com/PaperMC/Paper/pull/5691


paper-api/src/main/java/io/papermc/paper/statistic/Statistic.java
```diff
@@ -0,0 +1,27 @@
...
+import org.bukkit.scoreboard.Criteria;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Represents an individual statistic
+ *
+ * @param <S> stat (one of {@link org.bukkit.entity.EntityType}, {@link org.bukkit.Material} or {@link CustomStatistic}).
+ */
+@ApiStatus.NonExtendable
+public interface Statistic<S> extends Criteria {
```
Warriorrrr, line 13: Perhaps add an example here in the javadoc about how to obtain an instance of this via the StatisticType or CustomStatistic classes, since this is the first place that people who see the deprecations will be looking at.

paper-api/src/main/java/org/bukkit/event/player/PlayerStatisticIncrementEvent.java
```diff
@@ -108,8 +98,10 @@ public EntityType getEntityType() {
      * or item statistic otherwise returns null.
      *
      * @return the Material of the statistic
+     * @deprecated use {@link #getStat()}
      */
     @Nullable
+    @Deprecated(since = "1.21,4")
```
Warriorrrr: contains a comma instead of a period

Warriorrrr (approved):

kennytv (approved):

Issue #11633, opened 2024-11-18
Can't properly cancel entity movement with passengers or for vehicles


NonSwag:
### Expected behavior

Entity move event should be called even with passengers

### Observed/Actual behavior

Vehicles can't enter unless steered into the region

[](https://github.com/user-attachments/assets/f35ca02b-1c50-411d-9bcb-bd8255dd8d00)

Player movement is cancelled, but the mount can freely move

[](https://github.com/user-attachments/assets/ace6bf16-2dbc-42a6-9f0b-e9d9be8f2de0)

Mobs without passengers can't enter (intended)

[](https://github.com/user-attachments/assets/4433db33-bbe9-4b53-9f9c-bc55b64ab1dc)

### Steps/models to reproduce

1) Cancel player, vehicle and entity movement into a certain region
2) Try to walk into the region (should and does not work)
3) Try to push a mob or vehicle into the region (should and does not work)
4) Steer a vehicle (horse, boat, pig, strider) into the region (glitches out like crazy because player movement is still called _for some reason_, but the player stays mounted, and the vehicle keeps moving)
5) Steer a minecart into the region (should not, but does work. When dismounting, the player is teleported back to their supposed position)

### Plugin and Datapack List

> pl
[18:16:53 INFO]: Server Plugins (7):
[18:16:53 INFO]: Paper Plugins:
[18:16:53 INFO]:  - Protect, ServiceIO, Tweaks, Worlds
[18:16:53 INFO]: Bukkit Plugins:
[18:16:53 INFO]:  - FastAsyncWorldEdit, LuckPerms, WorldEditSUI

> datapack list
[18:17:05 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[18:17:05 INFO]: There are no more data packs available

### Paper version

> ver
[18:17:16 INFO]: Checking version, please wait...
[18:17:16 INFO]: This server is running Paper version 1.21.3-35-master@4e01ede (2024-11-18T12:51:42Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
You are 1 version(s) behind
Download the new version at: https://papermc.io/downloads/paper
Previous version: 1.21.3-29-be886cf (MC: 1.21.3)

### Other

My code for cancelling the movements

```java
@EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
public void onPlayerMove(PlayerMoveEvent event) {
[...]

Issue #7791, opened 2022-05-04
PlayerMoveEvent is not always called


Leguan16:
### Expected behavior

I recently discovered that the PlayerMoveEvent does not trigger when moving in a minecart or on a llama.
But due to the fact that it triggers when moving in a boat I also expect it to trigger when moving in a minecart.

### Observed/Actual behavior

It does not trigger the event when moving in a minecart or on a llama.

### Steps/models to reproduce

Just to show if a player is moving or not:
```
@EventHandler
public void onMove(PlayerMoveEvent event) {
    System.out.println("MOVE");
}
```
Hop in a minecart and you will see that it does not trigger the event. 
If you hop in a boat and move it will trigger the event. 

### Plugin and Datapack List

just the plugin with the handler.

### Paper version

This server is running Paper version git-Paper-317 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: e5da93a)
You are running the latest version
Previous version: git-Paper-302 (MC: 1.18.2)

electronicboy:
The fact that it triggers in a boat sounds like the actual bug here
The player isn't moving, the vehicle that they're riding is, is the notation here

HexedHero:
Does the boat move event get triggered on Spigot?

Leguan16:
Which boat move event?

kennytv:
Ideally we would call the move event, but only with rotational values, or create a new event for (relative) rotation changes inside of a vehicle ... if you're only looking for x,y,z coordinate changes, you need to listen to VehicleMoveEvent

BuzzardM:
Personally I would expect player move event to be fired as well as the player is changing positions. 
You can get the vehicle from a player with Player#getVehicle() as well. This method wouldn't be there if this event isn't supposed to be fired when moving in a vehicle?

--edit--
just checked the javadocs for Vehicle Entities. [Vehicle](https://papermc.io/javadocs/paper/1.17/org/bukkit/entity/Vehicle.html)
[...]

PR #11835, opened 2024-12-27
Add PlayerMoveVehicleEvent

Owen1212055:
**Reworked Vehicle PlayerMoveEvent Handling**

The `PlayerMoveEvent` is triggered whenever a player controls a vehicle. 
This event is generally broken, and does not represent what is actually happen. For example, the event used the player's location as the "from" position and the vehicle's location as the "to" position. This caused significant issues and confusion:

- Teleporting the player from within this event was undone later by a positionPassengers call, rendering `setTo` calls effectively useless.  
- Cancelling the event led to desyncs and odd behavior, such as the player being rapidly teleported between two locations.  
- The event was not giving a straightforward way to control the actual movement of the player on a vehicle.

Fixes:
PaperMC/Paper#8214
PaperMC/Paper#7021

In order to best clean up the event logic, the solution is to center the event around the **vehicle** rather than the player.

https://github.com/user-attachments/assets/c56cf9aa-2e05-4153-91a8-8cb85a974cf4

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -227,6 +227,7 @@ public net.minecraft.world.entity.LivingEntity detectEquipmentUpdates()V
 public net.minecraft.world.entity.LivingEntity effectsDirty
 public net.minecraft.world.entity.LivingEntity entityEventForEquipmentBreak(Lnet/minecraft/world/entity/EquipmentSlot;)B
 public net.minecraft.world.entity.LivingEntity getDeathSound()Lnet/minecraft/sounds/SoundEvent;
+public net.minecraft.world.entity.LivingEntity getRiddenSpeed(Lnet/minecraft/world/entity/player/Player;)F
 public net.minecraft.world.entity.LivingEntity getSoundVolume()F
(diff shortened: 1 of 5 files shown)


NonSwag:
This Potentially resolves #11633

NonSwag:
I would suggest improving the vehicle move event and/or adding a "mount steer event" and completely stop calling the player move event for passengers

Leguan16:
A `VehicleMovePlayerEvent` would be nice. (See #7791)

electronicboy:
The problem with vehicle movement is that there are a lot of overarching concepts which look similar but are entirely different, there were discussions over input events in the past, and there was no viable workable solution that I could see for providing events which are actually useful to devs within the context of the discussions they had (i.e. overriding the entities move controller)

Owen1212055:
I have no interest in upkeeping this.

In the end, we upkeep a breaking change for an  "exploit" that is most likely not relevant nowadays. So until that is removed, riding statistics will not work.

PR #11836, opened 2024-12-27
fix more incorrect array nullability annotations

Machine-Maker:
Replaces https://github.com/PaperMC/Paper/pull/10013

### DIFF
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -346,8 +346,7 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
      *
      * @return Chunk[] containing all loaded chunks
      */
-    @NotNull
-    public Chunk[] getLoadedChunks();
+    public @NotNull Chunk @NotNull[] getLoadedChunks();
 
     /**
      * Loads the specified {@link Chunk}.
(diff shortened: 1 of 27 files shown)


paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -346,8 +346,7 @@ default boolean isChunkGenerated(long chunkKey) {
      *
      * @return Chunk[] containing all loaded chunks
      */
-    @NotNull
-    public Chunk[] getLoadedChunks();
+    public @NotNull Chunk @NotNull[] getLoadedChunks();
```
lynxplay: Formatting

paper-api/src/main/java/org/bukkit/inventory/view/EnchantmentView.java
```diff
@@ -36,14 +37,13 @@ public interface EnchantmentView extends InventoryView {
      *
      * @return The enchantment offers that are provided
      */
-    @NotNull
-    EnchantmentOffer[] getOffers();
+    @NotNull EnchantmentOffer @Nullable [] getOffers();
```
lynxplay: Wrong way around.
The arrays are not null, its elements may be.
Same for below.

lynxplay (changes requested):

Lulu13022002:
Is that even relevant to do before the jspecify migration for not null types that are not wrong? there's also some formatting issues for World#getLoadedChunks, SimplexNoiseGenerator#dot

Machine-Maker:
Yeah, should do it. The jspecify migration is just gonna happen in chunks when you modify a file, I don’t think there’s gonna be one “huge” PR to do it all at once. Reviewing it would be a nightmare

lynxplay (approved):

PR #11838, opened 2024-12-27
Add EntityEffectTickEvent

LemonCaramel:
Adds the EntityPotionTickEvent.

This event is triggered on each tick of the potion effect (e.g. every 25 ticks for Poison level 1), and if the event is cancelled, the potion effect will not be applied to the entity.

I have been using it in my personal fork for the past year, mainly to add additional effects on each tick of a potion effect.


kennytv:
Should be named EntityEffectTick or EntityPotionEffect tick, I think I'd prefer the first (since it's technically not always associated with potions or potion specific effects)

kennytv (approved):
Thanks!

PR #11840, opened 2024-12-27
Add ExplodeEvent and Improvement in Spigot Explode events

Doc94:
~~This PR is a "copy" from a 2 years old PR in Spigot related to explosion, when this start the issue is how the gamerule for avoid mob griefing make the explosion event not being called, with recent changes now CraftBukkit pass the explosion behavior for make the user know what is the behaviour of this explosion.. with this in mind this PR takes that for well.. call the event for that also handle a issue where "explosions" created by Wind Burst call the block explosion with not data.~~

This PR implement a new "generic" event for any explosion generated by the server, this brings the same behaviour of the Spigot explode events (block,entity) this this event is intented for being called for all the explosions.. affects or not the environment.


Doc94:
PS: Very new with the new contribution system.. currently the ExplosionServer has unused CB imports but if remove that the fixupSourcePatches just fails

lynxplay (changes requested):
Alright, thank you for the PR!
We have reviewed it but are a bit (a bit too) hesitant to call these events when their javadocs do define them to *not* be called in these exact spots.

We could fire them pre-cancelled, but block break related events are obviously rather important for plugin data, and I'd rather avoid plugins dying/duping because they did not correctly check cancelled here.

Instead, Machine and I think it'd be best to use this API downside to finally create a better version of the "SomethingExplodeEvent" (better name pending) that covers both entities, blocks and *nothing* (like in the mace's case). Such an event can be called for these cases, specifying that no blocks will be broken etc.
The current events types could then be marked as Obsolete and the new event would be called *after* to allow people to take priority with the new API.

If you are up for implementing such a change yourself, feel free to just use this API, otherwise we can close it and open a feature request issue :+1:

Doc94:
Then you mean a new event "ExplodeEvent" for handle all the explosions, like a merge of the current entity and block not?
This require send a canceled state of this new event?
PD in teory just need call this new event in the sane place but later of the bukkit?

Doc94:
Ok based in the comments here and in discord i rework this PR for keep the changes but also the new calls just being called in the new event called...
i dont like "duplicate" the CB code for the paper event but dont wanna mess with strange behaviours for use the same method for call the paper event and make changes...

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
```diff
@@ -0,0 +1,158 @@
...
+@ApiStatus.Experimental
+@NullMarked
+public class ExplodeEvent extends Event implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Location location;
+    private final List<Block> blocks;
+    private final ExplosionResult result;
+    private @Nullable Block block;
```
Machine-Maker: No Block. It's always air, location is enough.

Doc94: Then remove the getBlock methods?
Just BlockState and Entity?

Machine-Maker: yeah

Doc94: removed from the event class.

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Returns the list of blocks that would have been caught in the explosion.
+     * <br>
+     * <b>Note:</b> the behaviours with these blocks depends on {@link #getExplosionResult()}
+     *
+     * @return All blocks caught in the explosion
+     */
+    public List<Block> blockList() {
```
Machine-Maker: If we are going to stick with the "mutability" of this list, lets make that clear in the javadocs that the expected behavior is to mutate this list.

Doc94: ready.

Doc94:
Exposed the DamageSource used in the explosion to the event... i feel this can help if consider the damage source can have many behaviours based in how datapacks can use the [explode effect](https://minecraft.wiki/w/Enchantment_definition#explode).

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Called when an explodes happen. this includes explosion who not affect blocks and are just visual.
+ */
+@ApiStatus.Experimental
+@NullMarked
+public class ExplodeEvent extends Event implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private DamageSource damageSource;
```
Lulu13022002: This can be final

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Returns a mutable list of blocks that would have been caught in the explosion.
+     * <br>
+     * <b>Note:</b> the behaviours with these blocks depends on {@link #getExplosionResult()}.
+     *
+     * @return All blocks caught in the explosion.
+     */
+    public List<Block> blockList() {
+        return this.blocks;
```
Lulu13022002, line 117: A bit weird to suddenly drop the "get" prefix there

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
```diff
@@ -149,45 +145,22 @@
...
-+        for (org.bukkit.block.Block bblock : bukkitBlocks) {
-+            BlockPos coords = new BlockPos(bblock.getX(), bblock.getY(), bblock.getZ());
-+            blocks.add(coords);
-+        }
++        // Paper start - call Explode Event
++        io.papermc.paper.event.world.ExplodeEvent explodeEvent = CraftEventFactory.callExplodeEvent(this, blocks);
++        this.wasCanceled = explodeEvent.isCancelled();
++        this.yield = explodeEvent.getYield();
++        blocks.clear();
++        blocks.addAll(explodeEvent.blockList().stream().map(bblock -> new BlockPos(bblock.getX(), bblock.getY(), bblock.getZ())).toList());
```
Lulu13022002: Would be better to cast the Block to CraftBlock and get the underlying the position, same for others

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -2122,6 +2127,70 @@ public static boolean callEntityInteractEvent(Entity nmsEntity, Block block) {
...
+        return blockListPostEvent.stream().map(bblock -> new BlockPos(bblock.getX(), bblock.getY(), bblock.getZ())).toList();
+    }
+
+    public static ExplodeEvent callExplodeEvent(ServerExplosion serverExplosion, List<BlockPos> blockPositions) {
+        // First convert the blockPositionList related to explosion to Bukkit
+        org.bukkit.World bworld = serverExplosion.level().getWorld();
+        Location location = CraftLocation.toBukkit(serverExplosion.center(), bworld);
+
+        List<org.bukkit.block.Block> blockList = new ObjectArrayList<>();
+        for (int i1 = blockPositions.size() - 1; i1 >= 0; i1--) {
```
Lulu13022002: Can you rename decompiled vars and you can import all the types you want

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -2122,6 +2127,70 @@ public static boolean callEntityInteractEvent(Entity nmsEntity, Block block) {
...
+            if (!bblock.isEmpty()) {
+                blockList.add(bblock);
+            }
+        }
+
+        ExplosionResult explosionResult = CraftExplosionResult.toBukkit(serverExplosion.getBlockInteraction());
+
+        ExplodeEvent event;
+        if (serverExplosion.getDirectSourceEntity() != null || serverExplosion.getDamageSource().getCustomEventDamager() != null) {
+            Entity entity = (serverExplosion.getDirectSourceEntity() != null) ? serverExplosion.getDirectSourceEntity() : serverExplosion.getDamageSource().getCustomEventDamager();
```
Lulu13022002: Can simplify a bit by moving this var above and just check if the entity is null or not

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -2122,6 +2127,70 @@ public static boolean callEntityInteractEvent(Entity nmsEntity, Block block) {
...
+
+    public static ExplodeEvent callExplodeEvent(ServerExplosion serverExplosion, List<BlockPos> blockPositions) {
+        // First convert the blockPositionList related to explosion to Bukkit
+        org.bukkit.World bworld = serverExplosion.level().getWorld();
+        Location location = CraftLocation.toBukkit(serverExplosion.center(), bworld);
+
+        List<org.bukkit.block.Block> blockList = new ObjectArrayList<>();
+        for (int i1 = blockPositions.size() - 1; i1 >= 0; i1--) {
+            BlockPos cpos = blockPositions.get(i1);
+            org.bukkit.block.Block bblock = bworld.getBlockAt(cpos.getX(), cpos.getY(), cpos.getZ());
```
Lulu13022002: Can just use CraftBlock.at(serverExplosion.level(), cpos) and move the above vars down

paper-api/src/main/java/org/bukkit/ExplosionResult.java
```diff
@@ -10,14 +10,14 @@ public enum ExplosionResult {
 
     /**
      * Represents an explosion where no change took place.
-     *
+     * <br>
      * This is the case when {@link org.bukkit.GameRule#MOB_GRIEFING} is
-     * disabled.
+     * disabled or the {@link org.bukkit.enchantments.Enchantment#WIND_BURST}.
```
Lulu13022002: This is wrong, on my end the enchantment use TRIGGER_BLOCK? like the wind charge

Lulu13022002 (commented):
This pr break fire explosion done via world.createExplosion(location, 3, true, false) by filtering empty blocks i guess. Ideally should only affect the event but it's tricky cause block list is mutable

paper-api/src/main/java/org/bukkit/ExplosionResult.java
Doc94: Unless im wrong the enchatment call the explode effect not the trigger thing

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
Doc94: I base in the originals bukkit events.
Not sure if add or not the get if this list its mutable

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Doc94: Im a little lost in this

paper-api/src/main/java/org/bukkit/ExplosionResult.java
Lulu13022002: Yes but each explode effect can define the type there, here is the wind burst definition:
image

Doc94: Then can change the docs for just mention trigger things?

Doc94:
> This pr break fire explosion done via world.createExplosion(location, 3, true, false) by filtering empty blocks i guess. Ideally should only affect the event but it's tricky cause block list is mutable

not sure how handle this.. i mean still being the "normal" behaviour with or not this PR i think...

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Lulu13022002: What i mean generally is that you can cleanup a bit the old code when you copy things from the old events. In this case i1 can just be named i and Block is already imported so no need to have a FQN there (but that point applies to all your types in this class since it doesn't end up in a patch)

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Lulu13022002: Still applies for the old events you moved

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Lulu13022002: Still applies for the new event

Lulu13022002:
With your PR that fire explosion doesn't spread fire on the ground which is an unintended side effect

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Doc94: ohhh okay i make the changes than the other place...

Doc94:
> With your PR that fire explosion doesn't spread fire on the ground which is an unintended side effect

i see the issue... not sure when i broke maybe the change from getType.isAir to isEmpty... i make a change for allow the empty things for the fire and looks is working. if its not a good "fix" can just revert the thing for the empty check

also i remove the shuffle in the KEEP behaviour because just make changes not related to the vanilla behaviour...

Lulu13022002:
That change is completely irrelevant, i have already explained the cause of that issue in the original message. It's not only affect your new event but regular ones as well. So i will open a new issue about that (the fix you commited is a workaround). And you broke spigot further with your equivalent commit.

Doc94:
> That change is completely irrelevant, i have already explained the cause of that issue in the original message. It's not only affect your new event but regular ones as well. So i will open a new issue about that (the fix you commited is a workaround). And you broke spigot further with your equivalent commit.

Yeah for the not explosion is handled by the keep maybe and how im now calm the event in that place the filter take all... Currently not sure why even need filter that... That block is in the list maybe just send all blocks and if user wanna they can filter that... Based in the first commit.
Or not sure if you have another suggest for this thing

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
```diff
@@ -234,6 +207,22 @@
          this.level.gameEvent(this.source, GameEvent.EXPLODE, this.center);
          List<BlockPos> list = this.calculateExplodedPositions();
          this.hurtEntities();
+@@ -254,6 +_,15 @@
+             this.interactWithBlocks(list);
+             profilerFiller.pop();
+         }
++        // Paper start - handle not interacts in explosion events
++        else {
++            io.papermc.paper.event.world.ExplodeEvent explodeEvent = CraftEventFactory.callExplodeEvent(this, list);
```
Lulu13022002: Cancelling the event should probably stop the fire too to be consistent with the old events

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
```diff
@@ -149,45 +145,22 @@
...
++        List<BlockPos> bukkitBlocks = CraftEventFactory.handleExplodeEvent(this, blocks);
 +        blocks.clear();
++        blocks.addAll(bukkitBlocks);
 +
-+        for (org.bukkit.block.Block bblock : bukkitBlocks) {
-+            BlockPos coords = new BlockPos(bblock.getX(), bblock.getY(), bblock.getZ());
-+            blocks.add(coords);
-+        }
++        // Paper start - call Explode Event
++        io.papermc.paper.event.world.ExplodeEvent explodeEvent = CraftEventFactory.callExplodeEvent(this, blocks);
```
Lulu13022002: You don't propagate the cancellation state of the old events to the new one

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
```diff
@@ -234,6 +207,22 @@
...
          this.hurtEntities();
+@@ -254,6 +_,15 @@
+             this.interactWithBlocks(list);
+             profilerFiller.pop();
+         }
++        // Paper start - handle not interacts in explosion events
++        else {
++            io.papermc.paper.event.world.ExplodeEvent explodeEvent = CraftEventFactory.callExplodeEvent(this, list);
++            this.wasCanceled = explodeEvent.isCancelled();
++            this.yield = explodeEvent.getYield();
```
Lulu13022002: The yield shouldn't be needed there no blocks explode anyway

Doc94: yeah but not make "damage" keep that.

Doc94:
about resolve https://github.com/PaperMC/Paper/issues/11999 now the filter is gone for allow the event show the same blocks (from positions) used in vanilla and avoid the mess with reduction of the fire.

Lulu13022002:
You can't remove the filter for the old events without updating the javadocs which is kinda a breaking change so not sure but the new one can hold empty blocks fine.
Cancelling the events still spread fire it seems mainly because you removed the filter.

Doc94:
> You can't remove the filter for the old events without updating the javadocs which is kinda a breaking change so not sure but the new one can hold empty blocks fine. Cancelling the events still spread fire it seems mainly because you removed the filter.

well already can consider a "breaking" change the thing about how now spigot fire the events for the keep behaviour and paper not (not in that events but yes in the new) but i make the rollback with the patch for the bukkit events.

the fire thing is maybe because just check the cancel for the keep behaviour and not in the main condition.. i move that for cover all calls.

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
```diff
@@ -234,6 +207,25 @@
...
++            list.clear();
++            list.addAll(explodeEvent.blockList().stream().map(block -> ((org.bukkit.craftbukkit.block.CraftBlock) block).getPosition()).toList());
++        }
++        if (this.wasCanceled) {
++            return;
```
Lulu13022002: You can't just return there, you need to clear the moonrise cache too, imo you can just move the check on the fire check below for now, otherwise would need to patch moonrise and add a hook which is annoying since it would only be relevant for paper

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -2122,6 +2128,69 @@ public static boolean callEntityInteractEvent(Entity nmsEntity, Block block) {
...
     }
 
+    public static List<BlockPos> handleExplodeEvent(ServerExplosion serverExplosion, List<BlockPos> blockPositions) {
+        // First convert the blockPositionList related to explosion to Bukkit
+        World bworld = serverExplosion.level().getWorld();
+
+        List<Block> blockList = new ObjectArrayList<>();
+        for (int posBlockPositionList = blockPositions.size() - 1; posBlockPositionList >= 0; posBlockPositionList--) {
+            org.bukkit.block.Block bblock = CraftBlock.at(serverExplosion.level(), blockPositions.get(posBlockPositionList));
+            if (!bblock.isEmpty() || serverExplosion.fire) {
```
Lulu13022002: I still don't like this workaround, you don't have to fix the issue in this PR, but having an inconsistent block list is even worse for plugins. Can you just revert to the empty check for now?

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -2122,6 +2128,69 @@ public static boolean callEntityInteractEvent(Entity nmsEntity, Block block) {
         return !event.isCancelled();
     }
 
+    public static List<BlockPos> handleExplodeEvent(ServerExplosion serverExplosion, List<BlockPos> blockPositions) {
+        // First convert the blockPositionList related to explosion to Bukkit
+        World bworld = serverExplosion.level().getWorld();
+
+        List<Block> blockList = new ObjectArrayList<>();
+        for (int posBlockPositionList = blockPositions.size() - 1; posBlockPositionList >= 0; posBlockPositionList--) {
```
Lulu13022002: Since you already reverse the order for the old events, redoing it will make the list inverted compared to the old events. The block list should be consistent across events and ideally the list should remains intact after the events (so not inverted)

Doc94: Then just convert in all cases and not mess with order?

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
Doc94: I forget about the moonrise thing (this cuase a mess if try to remove the CB import in the start for migrate to fq..)
Then just move this to fire && !wasCancel?

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Doc94: Hmmm but this make all the events cause that bug... I feel better keep this.

paper-server/patches/sources/net/minecraft/world/level/ServerExplosion.java.patch
Lulu13022002: Yes, the alternative is not worth imo and if mojang add more things patch should conflict anyway

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Lulu13022002: I just checked why they reverse the order like that, and it's a relic of the past back in 1.2 which doesn't make sense anymore in modern version.
However i'm tired of this PR, so i will take over.

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Lulu13022002: reverted since you are unable to fix it properly and i don't want to spoonfeed

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
Doc94: Okay... Sorry for the issues and thanks for the help...

paper-api/src/main/java/org/bukkit/event/block/BlockExplodeEvent.java
```diff
@@ -1,11 +1,14 @@
 package org.bukkit.event.block;
 
 import java.util.List;
+
```
Doc94: this is intended?

paper-server/patches/sources/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java.patch
```diff
@@ -5,17 +5,20 @@
...
          super(EntityType.ENDER_DRAGON, level);
-@@ -101,6 +_,7 @@
+@@ -101,6 +_,10 @@
          this.setHealth(this.getMaxHealth());
          this.noPhysics = true;
          this.phaseManager = new EnderDragonPhaseManager(this);
-+        this.explosionSource = new net.minecraft.world.level.ServerExplosion(level.getMinecraftWorld(), this, null, null, new Vec3(Double.NaN, Double.NaN, Double.NaN), Float.NaN, true, net.minecraft.world.level.Explosion.BlockInteraction.DESTROY); // Paper
++        // Paper start - initialize explosion source
++        this.explosionSource = new net.minecraft.world.level.ServerExplosion(level.getMinecraftWorld(), this, null, null, new Vec3(Double.NaN, Double.NaN, Double.NaN), Float.NaN, true, net.minecraft.world.level.Explosion.BlockInteraction.DESTROY);
++        this.explosionSource.yield = 0f;
```
Doc94, line 20: Based in how previously this value is directly send in the old method (that method not exists) then just set the yield to 0 for handle in events.

Doc94:
Rebase notification D:
This still need more improvements?

lynxplay:
Uh, good question. I'll try to get to it again soon but I'll throw it at other peeps too.

Doc94:
Any thing need to be make in this PR?

Lulu13022002:
My first review is still relevant iirc, generally it's a bit tricky but you need to play with lists a bit

Doc94:
> This pr break fire explosion done via world.createExplosion(location, 3, true, false) by filtering empty blocks i guess. Ideally should only affect the event but it's tricky cause block list is mutable

This one? If not wrong currently only affects the legacy ay not? Because the new way not filter any what ws the issue in first place when you let that comment.
Later im test that for see what can do with the list or just pass the empty blocks because i dont see a nice way to filter correctly and avoid mess that...

Doc94:
This also soft close https://github.com/PaperMC/Paper/issues/12441

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
```diff
@@ -0,0 +1,176 @@
...
+        this.yield = yield;
+    }
+
+    /**
+     * Checks if this explosion should spread fire
+     * on the ground.
+     *
+     * @return if this explosion should spread fire
+     */
+    public boolean shouldSpreadFire() {
```
Strokkur424, line 145: shouldSpreadFire sounds to me more like a setter than a getter. Perhaps renaming to `getSpreadsFire()` would be better for clarity reasons?

Strokkur424 (approved):

paper-api/src/main/java/io/papermc/paper/event/world/ExplodeEvent.java
Doc94: well in the cases where the "should" is used is for get, and for set its like setShould...

Strokkur424: There isn't any setter though, or did I miss something?

Doc94: Nope, the behaviour is cancel rather than force fire

TheBrunoRM: maybe `spreadsFire()` and `setSpreadsFire()` or something like that?

PR #11853, opened 2024-12-28
Add combat tracker API

SoSeDiK:
Adds `LivingEntity#getCombatTracker` to get entity's combat tracker that allows getting/setting combat history (list of damage entries during a combat session), recording new combat entries without causing actual damage, getting/resetting current combat state/session and getting a death message based on the present combat history.

`FallLocationType` represents a location from which the entity fell, and is used by vanilla to generate fall death messages.

___

API implementation is basically done, but definitely open to feedback. Javadocs & namings could use some love.

### DIFF
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -138,7 +138,11 @@ public net.minecraft.world.BossEvent overlay
 public net.minecraft.world.CompoundContainer container1
 public net.minecraft.world.CompoundContainer container2
 public net.minecraft.world.SimpleContainer items
+public net.minecraft.world.damagesource.CombatTracker entries
 public net.minecraft.world.damagesource.CombatTracker getMostSignificantFall()Lnet/minecraft/world/damagesource/CombatEntry;
+public net.minecraft.world.damagesource.CombatTracker inCombat
+public net.minecraft.world.damagesource.CombatTracker mob
+public net.minecraft.world.damagesource.CombatTracker takingDamage
 public net.minecraft.world.damagesource.DamageSource <init>(Lnet/minecraft/core/Holder;Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/Vec3;)V
 public net.minecraft.world.effect.MobEffect attributeModifiers
 public net.minecraft.world.effect.MobEffect$AttributeTemplate
@@ -767,10 +771,6 @@ public-f net.minecraft.server.dedicated.DedicatedServer serverLinks
 public-f net.minecraft.server.dedicated.DedicatedServer settings
 public-f net.minecraft.server.dedicated.DedicatedServerProperties pauseWhenEmptySeconds
 public-f net.minecraft.server.players.PlayerList maxPlayers
-public-f net.minecraft.world.damagesource.CombatTracker entries
-public-f net.minecraft.world.damagesource.CombatTracker inCombat
-public-f net.minecraft.world.damagesource.CombatTracker mob
-public-f net.minecraft.world.damagesource.CombatTracker takingDamage
 public-f net.minecraft.world.entity.LivingEntity combatTracker
 public-f net.minecraft.world.entity.LivingEntity invulnerableDuration
 public-f net.minecraft.world.entity.Mob goalSelector
--- a/build-data/paper.at
+++ b/build-data/paper.at
@@ -138,6 +138,7 @@ public net.minecraft.world.BossEvent overlay
 public net.minecraft.world.CompoundContainer container1
 public net.minecraft.world.CompoundContainer container2
 public net.minecraft.world.SimpleContainer items
+public net.minecraft.world.damagesource.CombatTracker getMostSignificantFall()Lnet/minecraft/world/damagesource/CombatEntry;
 public net.minecraft.world.damagesource.DamageSource <init>(Lnet/minecraft/core/Holder;Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/phys/Vec3;)V
 public net.minecraft.world.effect.MobEffect attributeModifiers
 public net.minecraft.world.effect.MobEffect$AttributeTemplate
@@ -766,6 +767,10 @@ public-f net.minecraft.server.dedicated.DedicatedServer serverLinks
 public-f net.minecraft.server.dedicated.DedicatedServer settings
 public-f net.minecraft.server.dedicated.DedicatedServerProperties pauseWhenEmptySeconds
 public-f net.minecraft.server.players.PlayerList maxPlayers
+public-f net.minecraft.world.damagesource.CombatTracker entries
+public-f net.minecraft.world.damagesource.CombatTracker inCombat
+public-f net.minecraft.world.damagesource.CombatTracker mob
+public-f net.minecraft.world.damagesource.CombatTracker takingDamage
 public-f net.minecraft.world.entity.LivingEntity combatTracker
 public-f net.minecraft.world.entity.LivingEntity invulnerableDuration
 public-f net.minecraft.world.entity.Mob goalSelector
--- a/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
+++ b/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
@@ -10,8 +10,6 @@ import org.jspecify.annotations.Nullable;
 
 /**
  * Represents a combat entry
- *
- * @since 1.21.4
  */
 @NullMarked
 @ApiStatus.Experimental
--- a/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
+++ b/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
@@ -67,7 +67,7 @@ public interface CombatEntry {
      * @param fallLocationType fall location type
      * @param fallDistance fall distance
      * @return a new combat entry
-     * @see LivingEntity#calculateFallLocationType()
+     * @see CombatTracker#calculateFallLocationType()
      * @see Entity#getFallDistance()
      */
     static CombatEntry combatEntry(
--- a/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
+++ b/paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
@@ -37,7 +37,7 @@ public interface CombatEntry {
      *
      * @return the fall location
      */
-    @Nullable FallLocation getFallLocation();
+    @Nullable FallLocationType getFallLocation();
 
     /**
      * Gets the fall distance at the time of this damage
@@ -55,7 +55,7 @@ public interface CombatEntry {
      * @param damageSource damage source
      * @param damage damage amount
      * @return combat entry
-     * @see #combatEntry(DamageSource, float, FallLocation, float)
+     * @see #combatEntry(DamageSource, float, FallLocationType, float)
      */
     static CombatEntry combatEntry(LivingEntity entity, DamageSource damageSource, float damage) {
         return InternalAPIBridge.get().createCombatEntry(entity, damageSource, damage);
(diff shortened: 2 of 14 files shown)


paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatTracker.java
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @return the entity behind this combat tracker
+     */
+    LivingEntity getEntity();
+
+    /**
+     * Gets the list of recorded combat entries.
+     * <br>
+     * Mutating the list is safe and won't have
+     * an effect on this entity's combat history.
```
kennytv: It should just be immutable/say it's a copy

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatTracker.java
```diff
@@ -0,0 +1,104 @@
...
+     * Note that overriding the entity's combat history won't
+     * have an effect on the entity's current or new combat state.
+     * Reset the current combat state and register new combat entries instead
+     * if you want the new history to have an effect on the combat state.
+     *
+     * @param combatEntries combat entries
+     * @see #resetCombatState()
+     * @see #addCombatEntry(CombatEntry)
+     */
+    void setCombatEntries(List<CombatEntry> combatEntries);
```
kennytv, line 50: What's the use of this?

paper-api/src/main/java/io/papermc/paper/world/damagesource/FallLocation.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Represents a type of location from which the entity fell
+ *
+ * @since 1.21.4
+ */
+@NullMarked
+@ApiStatus.Experimental
+public sealed interface FallLocation extends Translatable permits FallLocationImpl {
```
kennytv: FallLocationType maybe? Rn it sounds like this is actually holding a location, the vanilla name doesn't really work in our API

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatTracker.java
SoSeDiK: Updated the wording to state it's a copy.
Haven't made this immutable because it's returning a copy internally each time, so modifications are safe. Otherwise, users would need to make a copy of a copy.

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatTracker.java
SoSeDiK: "Sneakily" setting prior combat history without overriding/resetting active combat state.
It's a neat shortcut for providing fake pre-made combat history scenarios and producing specific death messages.

paper-api/src/main/java/io/papermc/paper/world/damagesource/FallLocation.java
SoSeDiK: Yes, the Type makes more sense

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
```diff
@@ -0,0 +1,79 @@
...
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.LivingEntity;
+import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Represents a combat entry
+ *
+ * @since 1.21.4
```
lynxplay: We have for now decided against since annotation in the project.
Adding *some* is just gonna make our lifes even more miserable when/if we solve them semi automatically.

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatTracker.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Calculates the most significant fall damage entry
+     *
+     * @return the most significant fall damage entry
+     */
+    @Nullable CombatEntry getMostSignificantFall();
```
lynxplay: computeMostSignificantFall

paper-server/patches/sources/net/minecraft/world/damagesource/CombatTracker.java.patch
```diff
@@ -0,0 +1,48 @@
...
+-
+-            this.entries.clear();
+-        }
++            // Paper start - Combat tracker API
++            resetCombatState();
++        }
++    }
++
++    public void resetCombatState() {
++        // Paper end - Combat tracker API
```
lynxplay: The intend ruined your diff here.

paper-server/src/main/java/io/papermc/paper/PaperServerInternalAPIBridge.java
```diff
@@ -12,4 +21,17 @@ public class PaperServerInternalAPIBridge implements InternalAPIBridge {
     public DamageEffect getDamageEffect(final String key) {
         return CraftDamageEffect.getById(key);
     }
+
+    @Override
+    public CombatEntry createCombatEntry(LivingEntity entity, DamageSource damageSource, float damage) {
+        net.minecraft.world.entity.LivingEntity mob = ((CraftLivingEntity) entity).getHandle();
+        net.minecraft.world.damagesource.FallLocation fallLocation = net.minecraft.world.damagesource.FallLocation.getCurrentFallLocation(mob);
+        float fallDistance = mob.fallDistance;
+        return new PaperCombatEntryWrapper(new net.minecraft.world.damagesource.CombatEntry(((CraftDamageSource) damageSource).getHandle(), damage, fallLocation, fallDistance));
```
lynxplay: can you just call the `createCombatEntry` overload here.
Makes life easier if that ctor ever changes.

paper-server/src/main/java/io/papermc/paper/world/damagesource/PaperCombatTrackerWrapper.java
```diff
@@ -0,0 +1,114 @@
...
+public record PaperCombatTrackerWrapper(net.minecraft.world.damagesource.CombatTracker handle) implements CombatTracker {
+
+    @Override
+    public LivingEntity getEntity() {
+        return this.handle.mob.getBukkitLivingEntity();
+    }
+
+    @Override
+    public List<CombatEntry> getCombatEntries() {
+        List<CombatEntry> combatEntries = new ArrayList<>();
```
lynxplay: Pre-allocate with size.

paper-api/src/main/java/io/papermc/paper/world/damagesource/FallLocationType.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * The entity is in twisting vines
+     */
+    FallLocationType TWISTING_VINES = new FallLocationTypeImpl("twisting_vines");
+    /**
+     * The entity is in scaffolding
+     */
+    FallLocationType SCAFFOLDING = new FallLocationTypeImpl("scaffolding");
+    /**
+     * The entity is within some other climable block
```
Lulu13022002: Small typo on climable

paper-api/src/main/java/io/papermc/paper/world/damagesource/CombatEntry.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return the amount of damage caused
+     */
+    float getDamage();
+
+    /**
+     * Gets the fall location at the time of this damage
+     *
+     * @return the fall location
+     */
+    @Nullable FallLocationType getFallLocation();
```
Lulu13022002: Did you forget to rename this one too? same in the javadocs here and below

build-data/paper.at
```diff
@@ -713,6 +714,10 @@ public-f net.minecraft.server.dedicated.DedicatedServer settings
 public-f net.minecraft.server.dedicated.DedicatedServerProperties pauseWhenEmptySeconds
 public-f net.minecraft.server.level.TicketType timeout
 public-f net.minecraft.server.players.PlayerList maxPlayers
+public-f net.minecraft.world.damagesource.CombatTracker entries
+public-f net.minecraft.world.damagesource.CombatTracker inCombat
+public-f net.minecraft.world.damagesource.CombatTracker mob
+public-f net.minecraft.world.damagesource.CombatTracker takingDamage
```
Lulu13022002: There's no need to remove the final modifier for those?

kennytv (approved):

Warriorrrr (approved):

paper-server/patches/sources/net/minecraft/world/damagesource/CombatTracker.java.patch
```diff
@@ -0,0 +1,37 @@
...
++    public void resetCombatState() {
++            // Paper end - Combat tracker API
+             boolean flag = this.inCombat;
+             this.takingDamage = false;
+             this.inCombat = false;
+@@ -155,6 +_,5 @@
+             }
+ 
+             this.entries.clear();
+-        }
```
lynxplay: Add a second opening bracket to the method start instead.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftLivingEntity.java
```diff
@@ -1167,4 +1174,15 @@ public void setBodyYaw(final float bodyYaw) {
...
         return this.getHandle().canUseSlot(org.bukkit.craftbukkit.CraftEquipmentSlot.getNMS(slot));
     }
+
+    @Override
+    public CombatTracker getCombatTracker() {
+        return this.combatTracker;
+    }
+
+    @Override
+    public FallLocationType calculateFallLocationType() {
```
lynxplay: I am a bit confused on these.
The entry yields this nullable, which mirrors internals, while this falls back to GENERIC.
What is the line of thought here?

SoSeDiK: I did what vanilla does when calculating the fall message:
```java
FallLocation fallLocation = Objects.requireNonNullElse(combatEntry.fallLocation(), FallLocation.GENERIC);
```
The reasoning is, since this calculates the fall location type, then if there's no special fall type, yield `GENERIC` (generic is only used internally as a fallback when the type is `null` iirc, so it was `null` vs `GENERIC`, and I liked the latter more)
But if I've missed something, nullable is also fine

lynxplay: Yea I know it is used internally for the death message computation, but that seems to be about it in terms of usage for that type?
The disconnect between "LivingEntity#calculateFallLocationType" and "create new combat entry from living entity -> ah it returns null here now eventho it just yielded GENERIC" is a bit unfortunate imo.
Tho I guess that is more on mojang than on this API given they do the "fallback" so late in their logic.
It is fine like this then, tho maybe a bit of javadocs if you have some time.

SoSeDiK: > it returns null here now eventho it just yielded GENERIC"

It's quite ambiguous (javadoc suggestions?), but is expected from my point of view
Having `GENERIC` in the entry represents a fall (and this method is meant for that), while `null` means that the entry is not a fall damage one

lynxplay: Yea but `calculateFallLocationType` yielding GENERIC then, when the entity isn't falling rn is kind of a counterpoint.
Having a non null falltype in the entry (at least according to internals) does not define if that was a fall.
That is tracked (correctly so) via the DamageSource.
You can be in the air and killed by an arrow, that kill won't be a registered as a fall related death even though the last entry does hold a falllocation type.

Javadocs on CombatEntry#combatEntry also currently link to the calc method, which then would not be correct right?

SoSeDiK: Good points, definitely need to at least clarify this in javadocs

lynxplay (approved):

PR #11857, opened 2024-12-29
feat: add new message for velocity in online mode

mja00:
Back in https://github.com/PaperMC/Paper/pull/8812 I added support for detecting if the end user is using Velocity over Bungeecord for the offline mode warning.

This expands upon that and will swap up the message whenever they're using Velocity in online mode. It's still enough of a message to let people know but without being a complete nag. 

It'll also make it, so people are less confused when they see the message in server logs in the paper-help channel. 

Definitely open to different wording on the message. I'm not really good at Englishing properly some days.

### DIFF
--- a/paper-server/patches/features/0017-Moonrise-optimisation-patches.patch
+++ b/paper-server/patches/features/0017-Moonrise-optimisation-patches.patch
@@ -23552,7 +23552,7 @@ index 2b46ca9a2a046063cad422bec00d76107537b091..9aa664537cc37e44db46d5a2a64ae311
                  thread1 -> {
                      DedicatedServer dedicatedServer1 = new DedicatedServer(
 diff --git a/net/minecraft/server/MinecraftServer.java b/net/minecraft/server/MinecraftServer.java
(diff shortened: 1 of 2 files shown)


mja00:
> Do we even need a message like that. Like if they are using velocity online mode, can't we just say "Using Velocity for Authentication"?

Technically _this_ server is still not authenticating usernames. It's just trust Velocity's checks.

MiniDigger:
I have no idea about velocity, but is there a risk here? you can't bypass velocity like you can with bungee, right? 
so the only risk is that you configured velocity incorrectly (i.e. not authenticating there) or you configured a velocity instance that you don't control, I don't think we need to tell the user to "beware" for that (especially without suggesting concrete actions)

NonSwag:
One or two years ago we had a discussion about a different message and the outcome was how it is now (just the firewall link)
I dont think there is any change necessary

electronicboy:
The logic here is flawed and a bungee mode server in which a user has said velocity online to true would trip this message improperly; This is also generally not stuff that I see the need to throw time on

mja00:
Yeah after some thinking there's no real good way to convey all the decent information. I'll just close this up lol

PR #11858, opened 2024-12-29
CraftEntity - Fix retain passengers teleport flag

### LINKED ISSUE #11003: Teleport with RetainPassengers flag doesn't appear to retain passengers.
### Expected behavior

I would expect the passengers of an entity to be able to teleport 

### Observed/Actual behavior

I noticed when a player is riding an entity, if you teleport it (with flags) the player doesn't get teleported.
The player also gets stuck in some weird limbo.
For example if you move a few chunks, the chunks no longer load, and if you attempt to break a block, nothing.
If I relog, the player is then matched up with their vehicle.

### Steps/models to reproduce

Test code to reproduce:
```java
Random random = new Random();
Entity vehicle = player.getVehicle();
World world = vehicle.getWorld();

int x = random.nextInt(10000);
int z = random.nextInt(10000);
Location highest = world.getHighestBlockAt(x, z).getLocation().add(0, 1, 0);

vehicle.teleport(highest, TeleportFlag.EntityState.RETAIN_PASSENGERS);
```

I simply make myself ride a sheep, and ran a command with that code.

### Plugin and Datapack List
```
[18:17:09 INFO]: Server Plugins (1):
[18:17:09 INFO]: Bukkit Plugins:
[18:17:09 INFO]:  - TestPlugin
```
### Paper version

Paper version 1.21-39-master@aa8d38d 

### Other

When looking at vanilla NMS Entity#teleportTo, I see they do the following:
(not sure if I can show Mojang code here, so ill just write short form)
```java
this.moveTo(...
this.teleportPassengers();
this.setYHeatRot(...
```

For testing sakes, I decided to incorporate this into CraftEntity's teleport method:
Screenshot 2024-06-30 at 6 04 53 PM
Did a quick build, and did a few simple test runs, the code I provided above now seems to work flawlessly.

ShaneBeee:
This PR aims to fix the RETAIN_PASSENGERS teleport flag not actually teleporting passengers.

I'm not positive if this is 100% the way to do it, but it's working.
Open to suggestions.

(ignore all my commits, had a lil branch issue)

Fixes #11003


Owen1212055 (commented):
The fix in principal looks good here, however, I am more curious why we are not using the vanilla teleportion logic in all cases.

Looking at the logic in `teleportSameDimension`, it seems pretty extensive behavior that we are not doing here. (Teleport transitions being sent to players, etc)

Not a blocking comment, but curious what others think here.

Machine-Maker:
I have an open PR (old paper) that changes the logic to use vanilla's

Machine-Maker:
You can take a look at what I was planning here https://github.com/PaperMC/Paper/pull/10734

ShaneBeee:
I did a PR once on CraftBukkit that did a big overhaul to the teleport methods, using Minecraft's logic.
I did this because Minecraft now supports TP with passengers via the TP command.
Unfortunately a few people over there felt that my PR took away the "old behaviours" of the Bukkit that people were used to.
OH NO... how dare I 😆 

So my goal for this PR was just to fix the bug.
I'm no pro when it comes to the NMS stuff like some of you guys are, so I didnt want to take on such a huge task, I just wanted to do a simple fix.

If another PR has a greater chance of making this so much better, I'm ok with this being closed and let someone else handle it.

That said I would be nice to fix the bug for the time being.

electronicboy:
There is a general concern over changing what plugins end up seeing, but, some level of "kicking the can for 10 years" doesn't help either; Moving back to vanilla handling of stuff is pretty much always better when viable (assuming we don't 'lose' out on something)

Owen1212055 (approved):
Understandable, in which case this is fine for now.
General cleanup can be done in another PR, but would rather have this fixed sooner than later.

electronicboy (approved):

Machine-Maker (approved):

lynxplay (approved):

Issue #11862, opened 2024-12-29
org.bukkit.attribute.Attribute.BURNING_TIME description spelling mistake


phant0m2290:
### Expected behavior

The description says ingition but it should be ignition
https://jd.papermc.io/paper/1.21.4/org/bukkit/attribute/Attribute.html#BURNING_TIME
![Image](https://github.com/user-attachments/assets/0b8b093e-faab-4b2b-8937-2a9b383413d3)

### Observed/Actual behavior

Spelling mistake. It says ingition

### Steps/models to reproduce

Open javadocs or sourcecode or whatever and look at the description of org.bukkit.attribute.Attribute.BURNING_TIME

### Plugin and Datapack List

N/A

### Paper version

paper-api:1.21.4-R0.1-SNAPSHOT

x9nico:
I done a PR to fix it : https://github.com/PaperMC/Paper/pull/11864

lynxplay:
Resolved by https://github.com/PaperMC/Paper/pull/11864

PR #11864, opened 2024-12-29
[ci skip] Fix spelling of ignition's word into javadoc (#11862)

x9nico:
Just a small fix into javadoc of Attribute's class for spelling the ignition's word caused by Spigot.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/attribute/Attribute.java
+++ b/paper-api/src/main/java/org/bukkit/attribute/Attribute.java
@@ -88,10 +88,14 @@ public interface Attribute extends OldEnum<Attribute>, Keyed, Translatable, net.
      * Strength with which an Entity will jump.
      */
     Attribute JUMP_STRENGTH = getAttribute("jump_strength");
+
+    // Paper start - Fix spelling of ignition's word
     /**
-     * How long an entity remains burning after ingition.
+     * How long an entity remains burning after ignition.
      */
     Attribute BURNING_TIME = getAttribute("burning_time");
+    // Paper end - Fix spelling of ignition's word
+
     /**
      * Resistance to knockback from explosions.
      */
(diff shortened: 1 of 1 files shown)


Doc94:
The Paper comments are not needed with hardfork 😌

x9nico:
> The Paper comments are not needed with hardfork 😌

Oh ok, I fixed it.

paper-api/src/main/java/org/bukkit/attribute/Attribute.java
```diff
@@ -88,10 +88,12 @@ public interface Attribute extends OldEnum<Attribute>, Keyed, Translatable, net.
      * Strength with which an Entity will jump.
      */
     Attribute JUMP_STRENGTH = getAttribute("jump_strength");
+
     /**
-     * How long an entity remains burning after ingition.
+     * How long an entity remains burning after ignition.
      */
     Attribute BURNING_TIME = getAttribute("burning_time");
+
```
masmc05: Those spaces are kinda out of place, either the whole file needs to receive them (most probably bad without a maintainer's approval), or remove here the diff that adds them

x9nico: These spaces was due to my old Paper comments (which was uselesss), I fixed the spaces. Thanks.

x9nico: > Those spaces are kinda out of place, either the whole file needs to receive them (most probably bad without a maintainer's approval), or remove here the diff that adds them

These spaces was due to my old Paper comments (which was uselesss), I fixed the spaces. Thanks.

Machine-Maker (approved):

lynxplay (approved):

PR #11865, opened 2024-12-29
Fix MapCanvas#drawImage

### LINKED ISSUE #11856: Map updates are ignored after the first call to render()
### Expected behavior

Map updates should not be ignored from updates after the initial call to render().

### Observed/Actual behavior

Maps do not render unless you rejoin

### Steps/models to reproduce

This GitHub issue is the whole conversation between me and the developer of ImageMaps. He even provides a way to reproduce the issue: https://github.com/SydMontague/ImageMaps/issues/70

### Plugin and Datapack List

Not really relevant, all you technically need is the ImageMaps plugin or the code they provided in the issue I linked above.

### Paper version

  [Sun 01:47:38 INFO ] Checking version, please wait...
  [Sun 01:47:38 INFO ] This server is running Paper version 1.21.4-65-main@287eb52 (2024-12-28T23:19:11Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

### LINKED ISSUE #11880: MapCanvas#drawImage incorrectly renders images
### Expected behavior

When rendering a small image using the MapCanvas#drawImage, it correctly draws the image, like shown in the image (all textures are drawn as seperate small images) (Paper 1.21.1):
Image

### Observed/Actual behavior

The images do not render correctly, instead they render the top row multiple times, and fail to render transparancy correctly (Paper 1.21.4 latest):
Image

### Steps/models to reproduce

Try to render a small (< 128px) image on a map canvas(renders weirdly), or render any image with an x offset to repeat the top pixel row.

### Plugin and Datapack List

Tested on a clean installed paper server with only my plugin.

### Paper version

This server is running Paper version 1.21.4-69-main@bdfad1c (2025-01-01T00:57:45Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are 1 version(s) behind
Previous version: 1.21.1-128-d348cb8 (MC: 1.21.1)

### Other

I found https://github.com/PaperMC/Paper/commit/318f7b201442328a9d8085d9cd18791b17796e32 to be the causing commit:
- The x == 0 offset failed to consider images with a width < 128 pixels, which furthermore is not reflected in the width variable.

- in the fallback loop, the 2nd arraycopy parameter is always 0, meaning it will always read the first row, resulting in the repeating pattern, and since the width variable does not represent the actual width for images smaller than 128 pixels, next rows will be rendered too.

- As for the transparancy issue, this is caused by the arraycopy not using the setPixel method, which would ignore any pixel color with a transparency value < 128

Warriorrrr:
Closes #11856, closes #11880

Fixes issues where changed pixels weren't being marked as changed (and thus not being updated), and fixes drawing images with < 128 width.

---
Download the paperclip jar for this pull request: [paper-11865.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2386406748.zip)

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
@@ -92,6 +92,9 @@ public class CraftMapCanvas implements MapCanvas {
     @Override
     public void drawImage(int x, int y, Image image) {
         // Paper start - Reduce work done by limiting size of image and using System.arraycopy
+        if (x < 0 || y < 0 || x >= 128 || y >= 128)
+            return;
+
         int width = 128 - x;
         int height = 128 - y;
         if (image.getHeight(null) < height)
@@ -119,12 +122,15 @@ public class CraftMapCanvas implements MapCanvas {
         // If x is 0, we can just copy the entire image as width is 128 and height is <=(128-y)
         if (x == 0) {
             System.arraycopy(bytes, 0, this.buffer, y * 128, width * height);
-            return;
+        } else {
+            for (int y2 = 0; y2 < height; ++y2) {
+                System.arraycopy(bytes, 0, this.buffer, (y + y2) * 128 + x, width);
+            }
         }
 
-        for (int y2 = 0; y2 < height; ++y2) {
-            System.arraycopy(bytes, 0, this.buffer, (y + y2) * 128 + x, width);
-        }
+        // Mark all colors within the image as dirty
+        this.mapView.worldMap.setColorsDirty(x, y);
+        this.mapView.worldMap.setColorsDirty(width - 1, height - 1);
         // Paper end
     }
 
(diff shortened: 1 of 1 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
```diff
@@ -92,6 +92,9 @@ protected byte[] getBuffer() {
     @Override
     public void drawImage(int x, int y, Image image) {
         // Paper start - Reduce work done by limiting size of image and using System.arraycopy
+        if (x < 0 || y < 0 || x >= 128 || y >= 128)
```
electronicboy: not sure about this, did this logic allow using negative values to support offsetting an oversized values?

Warriorrrr: I didn't think about that, I added that because the current impl will throw if a negative value is supplied. does seem like the old impl works with negative values, so I'll try seeing if I can fix that too

Warriorrrr: Tested this with a bunch of offsets & image sizes and it works the same as it did in CB now

Folas1337:
Looking forward to this being accepted :D

lynxplay (approved):

PR #11868, opened 2024-12-30
Feat: Add 'with' methods to CommandSourceStack

### LINKED ISSUE #11866: Expand the CommandSourceStack interface by exposing mutators

While brigadier is capable of redirecting or forking a node modifying the command source via (Single)RedirectModifier, it is impossible to make use of this feature using the current command API

Expand the CommandSourceStack interface by adding `with`er methods:
* `CommandSourceStack withExecutor(Entity executingEntity)`
* `CommandSourceStack withLocation(Location location)` (`atLocation`?)
* For the sake of completion and because Minecraft's CSS also has this method, `CommandSourceStack withSender(CommandSender sender)`, but I'm not really sure how to feel about this, it's one to discuss: do we want to retain the original sender or be able to change it? We can also do both by exposing an additional originalSender method as well, or simply not add such method today but can be revisited in the future

This would allow for plugins to use the forwarding mechanisms of brigadier and better integrate with it without needing to give players permission to run `/execute` or API users recoursing to use internals

Using internals or giving players permission to use /execute

### Other

_No response_

Strokkur424:
fixes #11866 

The following methods have been added:
- `CommandSourceStack#withExecutor(@Nullable Entity)`
- `CommandSourceStack#withLocation(Location location)`

JavaDocs hasn't been added yet, but that is something I will be doing very soon.

Testing the code can be done via the following:
```java
    final LiteralCommandNode<CommandSourceStack> node = Commands.literal("test")
        .then(Commands.literal("run")
            .executes(ctx -> {
                Bukkit.broadcast(MiniMessage.miniMessage().deserialize("<name>: I am at <x> <y> <z>",
                    Placeholder.component("name", ctx.getSource().getExecutor().name()),
                    Placeholder.unparsed("x", String.valueOf(ctx.getSource().getLocation().x())),
                    Placeholder.unparsed("y", String.valueOf(ctx.getSource().getLocation().y())),
                    Placeholder.unparsed("z", String.valueOf(ctx.getSource().getLocation().z()))
                ));

                return 1;
            })).build();

    final LiteralCommandNode<CommandSourceStack> withNode = Commands.literal("testwith")
        .then(Commands.argument("entities", ArgumentTypes.entities())
            .then(Commands.literal("run")
                .fork(node, ctx -> {
                    final EntitySelectorArgumentResolver entities = ctx.getArgument("entities", EntitySelectorArgumentResolver.class);
                    final List<Entity> resolved = entities.resolve(ctx.getSource());

                    return resolved.stream()
                        .map(entity -> ctx.getSource().withExecutor(entity).withLocation(entity.getLocation()))
                        .toList();
                }))).build();

    event.registrar().register(node);
    event.registrar().register(withNode);
});
```

Running `/test run` works as expected, printing out the executing player's location.
Running `/testwith @e[limit=2,type=!player] run run` also works, printing out two entities names and locations.

testwith-command

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
+++ b/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
@@ -1,5 +1,7 @@
 package io.papermc.paper.command.brigadier;
 
+import com.mojang.brigadier.RedirectModifier;
+import com.mojang.brigadier.tree.CommandNode;
 import org.bukkit.Location;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Entity;
@@ -48,8 +50,24 @@ public interface CommandSourceStack {
      * @return entity that executes this command
      */
     @Nullable Entity getExecutor();
-    
+
+    /**
+     * Creates a new CommandSourceStack object with a different location for redirecting commands to other nodes.
+     *
+     * @param location The location to create a new CommandSourceStack object with
+     * @return The newly created CommandSourceStack
+     * @see #getLocation()
+     * @see com.mojang.brigadier.builder.ArgumentBuilder#fork(CommandNode, RedirectModifier)
+     */
     CommandSourceStack withLocation(Location location);
-    
+
+    /**
+     * Creates a new CommandSourceStack object with a different executor for redirecting commands to other nodes.
+     *
+     * @param executor The executing entity to create a new CommandSourceStack object with
+     * @return The newly created CommandSourceStack
+     * @see #getExecutor()
+     * @see com.mojang.brigadier.builder.ArgumentBuilder#fork(CommandNode, RedirectModifier)
+     */
     CommandSourceStack withExecutor(Entity executor);
 }
--- a/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
+++ b/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
@@ -4,7 +4,6 @@ import org.bukkit.Location;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Entity;
 import org.jetbrains.annotations.ApiStatus;
-import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
@@ -50,7 +49,7 @@ public interface CommandSourceStack {
      */
     @Nullable Entity getExecutor();
     
-    CommandSourceStack withLocation(@NonNull Location location);
+    CommandSourceStack withLocation(Location location);
     
-    CommandSourceStack withExecutor(@Nullable Entity executor);
+    CommandSourceStack withExecutor(Entity executor);
 }
--- a/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
+++ b/paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
@@ -4,6 +4,7 @@ import org.bukkit.Location;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Entity;
 import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
@@ -48,4 +49,8 @@ public interface CommandSourceStack {
      * @return entity that executes this command
      */
     @Nullable Entity getExecutor();
+    
+    CommandSourceStack withLocation(@NonNull Location location);
+    
+    CommandSourceStack withExecutor(@Nullable Entity executor);
 }
--- a/paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
+++ b/paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
@@ -18,6 +18,32 @@
  
      public CommandSourceStack(
          CommandSource source,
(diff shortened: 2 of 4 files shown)


paper-api/src/main/java/io/papermc/paper/command/brigadier/CommandSourceStack.java
```diff
@@ -48,4 +49,8 @@ public interface CommandSourceStack {
      * @return entity that executes this command
      */
     @Nullable Entity getExecutor();
+    
+    CommandSourceStack withLocation(@NonNull Location location);
```
Lulu13022002: NonNull annotation is redundant in a NullMarked class

paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
```diff
@@ -18,6 +18,32 @@
...
++            this.source,
++            new Vec3(location.x(), location.y(), location.z()),
++            new Vec2(location.getYaw(), location.getPitch()),
++            ((org.bukkit.craftbukkit.CraftWorld) location.getWorld()).getHandle(),
++            this.permissionLevel,
++            this.textName,
++            this.displayName,
++            this.server,
++            this.entity
++        );
```
Lulu13022002: Trailing whitespace are not needed

paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperCommandSourceStack.java
```diff
@@ -6,9 +6,11 @@
 import net.minecraft.world.phys.Vec3;
 import org.bukkit.Location;
 import org.bukkit.command.CommandSender;
+import org.bukkit.craftbukkit.entity.CraftEntity;
 import org.bukkit.entity.Entity;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jspecify.annotations.NonNull;
```
Lulu13022002, line 12: Don't mix nullability annotations either convert the class to jspecify or keep using jetbrains one for now

paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
```diff
@@ -18,6 +18,32 @@
...
++    
++    // Paper start - Expose 'with' functions from the CommandSourceStack
++    @Override
++    public CommandSourceStack withLocation(org.bukkit.Location location) {
++        return this.getLocation().equals(location)
++            ? this
++            : new CommandSourceStack(
++            this.source,
++            new Vec3(location.x(), location.y(), location.z()),
++            new Vec2(location.getYaw(), location.getPitch()),
```
Lulu13022002: Yaw and pitch looks inverted here?

paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperCommandSourceStack.java
```diff
@@ -40,6 +42,12 @@ default Entity getExecutor() {
         return nmsEntity.getBukkitEntity();
     }
 
+    @Override
+    @NonNull
+    default CommandSourceStack withExecutor(@Nullable Entity executor) {
```
Lulu13022002: This doesn't seems to be safe to pass null here?

Strokkur424: It actually is safe to pass null, since executors are actually nullable (mainly when executed as the console). It seems that internally, the method is not annotated as such, but judging from the implementation, it is only setting the executor field of CommandSourceStack, which may be null

paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
Strokkur424: In the Bukkit location, the yaw comes before pitch. That's what I used as a reference here. I am not sure whether this is the same in NMS terms

paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperCommandSourceStack.java
emilyy-dev: > It actually is safe to pass null, since executors are actually nullable (mainly when executed as the console). It seems that internally, the method is not annotated as such, but judging from the implementation, it is only setting the executor field of CommandSourceStack, which may be null

passing null will cause a NPE, n.m.c.CSS#withEntity calls getName and getDisplayName on the passed entity

Strokkur424: In that case, how is it possible that the Bukkit CSS has their withEntity nullable?

emilyy-dev:
Because it isn't required to construct a CSS initially, think console
commands

paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
Strokkur424: It was, in fact, inverted. I corrected it in the next commit 👍

paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperCommandSourceStack.java
Strokkur424: Made the parameter not nullable

paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperCommandSourceStack.java
Strokkur424: Converted the class to jspecify

kennytv (approved):

Machine-Maker (approved):

PR #11871, opened 2024-12-31
Improve performance of PoiCompetitorScan by unrolling stream

ChromMob:
I was debugging villagers and I noticed that the reduce on the stream has massive overhead so I decided to unroll it.
Here are sparks for before and after:
[Before](https://spark.lucko.me/3fTr1E6BK8)
[After](https://spark.lucko.me/rc6SrM21zA)
Those numbers are with 229 active villagers on R7 5700X with 64GB of RAM.
image
Here is the test world:
[world.tar.gz](https://github.com/user-attachments/files/18280935/world.tar.gz)

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java.patch
@@ -14,26 +14,26 @@
 +                                // Paper start - Improve performance of PoiCompetitorScan by unrolling stream
 +                                // The previous logic used Stream#reduce to simulate a form of single-iteration bubble sort
 +                                // in which the "winning" villager would maintain MemoryModuleType.JOB_SITE while all others
-+                                // would loose said memory module type by passing each "current winner" and incoming next
++                                // would lose said memory module type by passing each "current winner" and incoming next
 +                                // villager to #selectWinner.
 +                                poi -> {
 +                                    final List<LivingEntity> livingEntities = instance.get(nearestLivingEntities);
 +
 +                                    Villager winner = villager;
-+                                    for (LivingEntity entity : livingEntities) {
-+                                        if (entity == villager) {
++                                    for (final LivingEntity other : livingEntities) {
++                                        if (other == villager) {
 +                                            continue;
 +                                        }
-+                                        if (!(entity instanceof final net.minecraft.world.entity.npc.Villager other)) {
++                                        if (!(other instanceof final net.minecraft.world.entity.npc.Villager otherVillager)) {
 +                                            continue;
 +                                        }
-+                                        if (!entity.isAlive()) {
++                                        if (!other.isAlive()) {
 +                                            continue;
 +                                        }
-+                                        if (!competesForSameJobsite(globalPos, poi, other)) {
++                                        if (!competesForSameJobsite(globalPos, poi, otherVillager)) {
 +                                            continue;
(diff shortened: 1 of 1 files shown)


paper-server/patches/sources/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java.patch
```diff
@@ -0,0 +1,33 @@
...
+-                                    .map(entity -> (Villager)entity)
+-                                    .filter(LivingEntity::isAlive)
+-                                    .filter(v -> competesForSameJobsite(globalPos, poi, v))
+-                                    .reduce(villager, PoiCompetitorScan::selectWinner)
++                                poi -> {
++                                    List<LivingEntity> livingEntities = instance.get(nearestLivingEntities);
++
++                                    Villager winner = villager;
++                                    for (LivingEntity entity : livingEntities) {
++                                        if (!(entity instanceof Villager && entity != villager)) {
```
Warriorrrr: This can use a pattern instanceof, plus the entity != villager can be taken out of the negated part to be easier to read as was shortly talked about in discord

Warriorrrr (commented):
Since it's changing minecraft code, this can still use some paper comments

paper-server/patches/sources/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java.patch
ChromMob: You are right, I forgot about that, will do.

vadcx:
The profiling results look very impressive. You may have forgotten to return the winner to ifPresent? [.reduce](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#reduce(java.util.function.BinaryOperator))

ChromMob:
> The profiling results look very impressive. You may have forgotten to return the winner to ifPresent? [.reduce](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#reduce(java.util.function.BinaryOperator))

The result wasn't used at all before so there is no need to return anything because the only thing it does is the sideffect.

paper-server/patches/sources/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java.patch
```diff
@@ -0,0 +1,44 @@
+--- a/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java
++++ b/net/minecraft/world/entity/ai/behavior/PoiCompetitorScan.java
+@@ -17,18 +_,33 @@
+             instance -> instance.group(instance.present(MemoryModuleType.JOB_SITE), instance.present(MemoryModuleType.NEAREST_LIVING_ENTITIES))
+                 .apply(
+                     instance,
+-                    (jobSite, nearestLivingEntities) -> (level, villager, gameTime) -> {
++                    (jobSite, nearestLivingEntities) -> (level, villagerA, gameTime) -> {
```
Machine-Maker: Throw a paper comment up here too

ChromMob: Fixed

NoahvdAa (approved):

lynxplay (approved):

Machine-Maker (approved):

PR #11884, opened 2025-01-02
Fix EndPortal teleport entity's delta movement

s-yh-china:
The some events is inconsistent with the vanilla and the description in javadoc.
This commit may have fixed these issues.


kennytv:
Can you explain each of the changes and why they're needed?

lynxplay:
Been a week, given this PR isn't linked to an actual issue and you failed to reply, I'll close this.

Issue #11887, opened 2025-01-02
IllegalStateException when trying to create a blockState from a Material


T14D3:
### Stack trace

```log
Error occurred while enabling Zones v1.0-SNAPSHOT (Is it up to date?)
java.lang.IllegalStateException: The blockState must be placed to call this method
	at com.google.common.base.Preconditions.checkState(Preconditions.java:513) ~[guava-33.3.1-jre.jar:?]
	at org.bukkit.craftbukkit.block.CraftBlockState.requirePlaced(CraftBlockState.java:332) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockState.getWorld(CraftBlockState.java:101) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockEntityState.<init>(CraftBlockEntityState.java:53) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftMovingPiston.<init>(CraftMovingPiston.java:10) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates$BlockEntityStateFactory.createBlockState(CraftBlockStates.java:115) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates$BlockEntityStateFactory.createBlockState(CraftBlockStates.java:107) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates$BlockEntityStateFactory.createBlockState(CraftBlockStates.java:89) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates.getBlockState(CraftBlockStates.java:308) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates.getBlockState(CraftBlockStates.java:294) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
	at org.bukkit.craftbukkit.block.CraftBlockStates.getBlockState(CraftBlockStates.java:283) ~[paper-1.21.4.jar:1.21.4-72-daddcf6]
[...]

PR #11888, opened 2025-01-02
[ci skip] Add @throws to BlockData#createBlockState JavaDoc

T14D3:
Properly indicate in the JavaDoc that the method can throw an IllegalStateException if a block requires to be placed to get a BlockState for it (Currently only occurs for MOVING_PISTON).

Addresses #11887


emilyy-dev:
This is not a proper fix, that error should not happen altogether so the underlying issue causing this should be fixed instead

Machine-Maker:
Agreed with @emilyy-dev . The underlying issue should be fixed so the exception is not thrown at all. It can be avoided in this case. First to fix the swallowed exception and to fix the issue with moving piston.

PR #11896, opened 2025-01-04
Fix vanilla parity on enchantable items with no targets

### LINKED ISSUE #11821: Enchantable Component
### Expected behavior

When applying the enchantable component tag to an item that's normally not enchantable it should show some enchantment options when put in an enchanting table.

### Observed/Actual behavior

The enchantment options are only shown in vanilla and not on a Paper server.

### Steps/models to reproduce

1. Execute this command as an operator. "/give @s dirt[enchantable={value:1000}]"
2. Put the Dirt item into an enchanting table.
3. Check if any enchantment options pop up while ignoring that there are no enchantment names shown when it works.

Do this on a paper server and vanilla world/server and compare the results.

### Plugin and Datapack List

I've tested it on a completely clean setup of paper and survival server in 1.21.3 for both servers without any data packs, plugins, mods, resource packs, or shaders just plain vanilla.

### Paper version

This server is running Paper version 1.21.3-82-ver/1.21.3@5a60ffb (2024-12-23T19:10:30Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)

### Other

Image
Image

Y2Kwastaken:
This pull request fixes #11821

Previously after the enchantment event took place bukkit would clear out all empty enchantments and send a completely empty offer list to the client. Minecraft handles this differently by default by still sending a cost to the client so the slots illuminate as seen in the attached issue.

This pull request attempts to address this issue by not setting back enchantment offer cost to 0 after PrepareItemEnchantEvent. This keeps the original price that was generated through the loop earlier within the slotsChanged method.

As far as drawbacks go for this fix after looking into this it would largely be thinking about how things are handled event side, because something would need to be done. If a developer wants to null out a spot how do we address this while maintaining vanilla behavior when an item may have no offers available? Or should vanilla behavior be kept to start with?

For reference when blanking out all 3 offers an enchanment table now looks like this, client side the hover shows no enchantment names, however displays all levels.
```kotlin
    @EventHandler
    fun onEnchant(event: PrepareItemEnchantEvent) {
        for (index in event.offers.indices) {
            event.offers[index] = null
        }
    }
```
image

within default vanilla behavior as far as I'm aware there is no scenario where three enchantments aren't sent.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/inventory/EnchantmentMenu.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/inventory/EnchantmentMenu.java.patch
@@ -125,7 +125,7 @@
 +                        org.bukkit.craftbukkit.inventory.CraftItemStack craftItemStack = org.bukkit.craftbukkit.inventory.CraftItemStack.asCraftMirror(itemStack);
 +                        Holder<Enchantment> holder = registry.byId(this.enchantClue[id]);
 +                        if (holder == null) return;
-+                        org.bukkit.enchantments.Enchantment hintedEnchantment = org.bukkit.craftbukkit.enchantments.CraftEnchantment.minecraftHolderToBukkit(registry.byId(this.enchantClue[id]));
++                        org.bukkit.enchantments.Enchantment hintedEnchantment = org.bukkit.craftbukkit.enchantments.CraftEnchantment.minecraftHolderToBukkit(holder);
 +                        int hintedEnchantmentLevel = this.levelClue[id];
 +                        org.bukkit.event.enchantment.EnchantItemEvent event = new org.bukkit.event.enchantment.EnchantItemEvent((org.bukkit.entity.Player) player.getBukkitEntity(), this.getBukkitView(), this.access.getLocation().getBlock(), craftItemStack, this.costs[id], enchants, hintedEnchantment, hintedEnchantmentLevel, id);
 +                        level.getCraftServer().getPluginManager().callEvent(event);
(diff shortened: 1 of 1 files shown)


paper-server/patches/sources/net/minecraft/world/inventory/EnchantmentMenu.java.patch
```diff
@@ -124,6 +123,8 @@
 +                            enchants.put(org.bukkit.craftbukkit.enchantments.CraftEnchantment.minecraftHolderToBukkit(instance.enchantment), instance.level);
 +                        }
 +                        org.bukkit.craftbukkit.inventory.CraftItemStack craftItemStack = org.bukkit.craftbukkit.inventory.CraftItemStack.asCraftMirror(itemStack);
++                        Holder<Enchantment> holder = registry.byId(this.enchantClue[id]);
++                        if (holder == null) return;
 +                        org.bukkit.enchantments.Enchantment hintedEnchantment = org.bukkit.craftbukkit.enchantments.CraftEnchantment.minecraftHolderToBukkit(registry.byId(this.enchantClue[id]));
```
lynxplay: Can reuse the now pre-fetched holder in here

Machine-Maker:
Yeah, this is just a fundamental problem with how EnchantmentOffer is structured. It should've never been nullable in the array and instead had a nullable enchantment so the cost could still be set. I think it's fine to just revert to handle it like vanilla.

kennytv (approved):

Machine-Maker (approved):

lynxplay (approved):

PR #11897, opened 2025-01-04
Fix TargetHitEvent#getSignalStrength() post-hard-fork

Doc94:
Reported in discord `TargetHitEvent#getSignalStrength()` return just 20 or 8 based in the projectile this behaviour is a side-effect caused by hard-fork where pre-hard-fork the value used is called "i" (ref: https://github.com/PaperMC/Paper/blob/master/patches/server/0441-Add-TargetHitEvent.patch and broken in https://github.com/PaperMC/Paper/commit/e9680a5afedcf05341b332870e0c542a17d78efa#diff-3a492050417244fc5b90189b40134f0124f010ffb92b94158329562b71014f59R16) but now is called `redstoneStrength` and `i` is just used for another thing.

---
Download the paperclip jar for this pull request: [paper-11897.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2385196968.zip)

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/TargetBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/TargetBlock.java.patch
@@ -20,12 +20,12 @@
 +        if (projectile instanceof Projectile) {
 +            final org.bukkit.craftbukkit.block.CraftBlock craftBlock = org.bukkit.craftbukkit.block.CraftBlock.at(level, hit.getBlockPos());
 +            final org.bukkit.block.BlockFace blockFace = org.bukkit.craftbukkit.block.CraftBlock.notchToBlockFace(hit.getDirection());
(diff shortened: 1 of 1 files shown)


Machine-Maker (changes requested):
You also need to update the usage of `i` inside the last hunk in that file, the call to `awardTargetHitCriteria`.

lynxplay (approved):

Machine-Maker (approved):

Machine-Maker:
It's not so much "typos" as it's just hunks that applied successfully but the names of local variables changed which changed some functionality. The diff between this and pre-hard fork is way too littered with all the other renaming/decompiler changes. Our tooling also changes the local variable names to much nicer names (in some cases) which renames all the other local variables as well. So it's just a matter of identifying those cases. We did several pass throughs of the whole diff right after, and found a bunch.

PR #11899, opened 2025-01-04
Call PlayerItemDamageEvent for tridents

### LINKED ISSUE #11655: Riptiding does not call PlayerItemDamageEvent

notTamion:
fixes #11655
pre hardfork #11716

### DIFF
--- a/paper-server/patches/features/0028-Optimize-Hoppers.patch
+++ b/paper-server/patches/features/0028-Optimize-Hoppers.patch
@@ -60,10 +60,10 @@ index cc2d442682496197d29ace79b22e6cf6fb7edf5e..ae220a732c78ab076261f20b5a54c71d
              /* Drop global time updates
              if (this.tickCount % 20 == 0) {
 diff --git a/net/minecraft/world/item/ItemStack.java b/net/minecraft/world/item/ItemStack.java
-index 0eb9811e573f0c38de3eeb17099872efcbb4c2dd..b511fe8295369ac4014beb351cd2e3f770c10170 100644
+index b393be76bead3c66ab0bd8a0e6fd9b9ef81d8e28..76f50437396f8f856381d0fbef52953ef7c263f6 100644
 --- a/net/minecraft/world/item/ItemStack.java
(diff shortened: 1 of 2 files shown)


lynxplay:
I am unsure if breaking actually works here given that LivingEntity#onEquippedItemBroken isn't called if the item breaks.
Might be better to just *not* allow breaking. Needs investigation.

notTamion:
so all it seems to do is send the breaking packet to the client and removing modifiers from the itemstack (which i am not sure is necessary? considering its breaking anyways). using setDamage(reallyhighnumber) when throwing the trident without riptide it just gets stuck in the ground and can't be picked up anymore which i guess is to be expected, when using with riptide it just breaks but still gives you the boost. not sure whether disallowing breaking the trident, changing the onBreak Consumer or just leaving it as is is the best option
not allowing it to break would be kind of an inconsistency tho (copy pasted from [discord](https://discord.com/channels/289587909051416579/925530366192779286/1325217364349812787) incase you missed it)

lynxplay:
For now, the event won't allow for breaking the itemstack, mostly because doing so is down the line upkeep to ensure no caller expects the item to remain.

If people end up requesting this specific feature, we can revisit this and potentially enable this, but util we have requests it isn't worth the upkeep.

lynxplay (approved):

Machine-Maker (approved):

PR #10248, opened 2024-02-13
Call HangingBreakByEntityEvent for collision with boats

notTamion:
resolves #10058 
post hardfork #11901
perhaps a getRemovers should be added since mulitple boats can be responsible for breaking an entity. let me know what you think

PR #11901, opened 2025-01-04
Call HangingBreakByEntityEvent for collision with boats

### LINKED ISSUE #10058: Boats breaking hanging entities
### Expected behavior

When a boat breaks a hanging entity, it should fire HangingBreakByEntityEvent

### Observed/Actual behavior

When a boat breaks hanging entity, it fires HangingBreakEvent with PHYSICS cause, in thus not firing HangingBreakByEntityEvent

### Steps/models to reproduce

Just ram into itemframe with a boat

### Plugin and Datapack List

[02:07:12 INFO]: Bukkit Plugins: (78):
[02:07:13 INFO]:  - AdvancedPets, *AdvancedReplay, AdventCalendar, *AwesomeSnow, BetterVote, ChestSort, ChristmasPresents, Clickthrough, CMILib, CoreProtect
[02:07:13 INFO]:  DecentHolograms, FastChunkPregenerator, Jobs, JobsReborn-PatchPlaceBreak, LPX, LuckPerms, mcMMO, MyCommand, NexEngine, PlaceholderAPI
[02:07:13 INFO]:  PlugManX, ProtocolLib, SCAssist, spark, TAB, Vault, ViaBackwards, ViaVersion, Votifier, Vulcan
[02:07:13 INFO]:  XCatch, ZCAdminSystemAdvanced, ZCAdvancedEventsEnd, ZCAfk, ZCAmbientSounds, ZCAnimalTeleportation, ZCAPI, ZCAutoMessage, ZCAutoRestart, ZCAutoSellChest
[02:07:13 INFO]:  ZCBans, ZCBluePrintsServer, ZCChatFormat, ZCClaims, ZCClanAdvanced, ZCClanWarsSurvival, ZCCommands, ZCCrates, ZCCreators, ZCDaily
[02:07:13 INFO]:  ZCDailyTasks, ZCDeop, ZCEconomy, ZCEvents, ZCGuessing, ZCHolograms, ZCHotDeals, ZCMainCore, ZCMarket, ZCMessages
[02:07:13 INFO]:  ZCMeta, ZCMobOptimizer, ZCOneVersusOne, ZCPackages, ZCParkour, ZCPlayerDeathInventory, ZCPlayerLimit, ZCPvpMode, ZCQueue, ZCRanksTime
[02:07:13 INFO]:  ZCRules, ZCServices, ZCShop, ZCSignBuy, ZCSit, ZCTops, ZCTrade, ZCVoteOff

### Paper version

image

### Other

_No response_

notTamion:
resolves #10058
pre hardfork #10248

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/EntityGetter.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/EntityGetter.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/level/EntityGetter.java
 +++ b/net/minecraft/world/level/EntityGetter.java
+@@ -56,7 +_,7 @@
+             return List.of();
+         } else {
+             Predicate<Entity> predicate = entity == null ? EntitySelector.CAN_BE_COLLIDED_WITH : EntitySelector.NO_SPECTATORS.and(entity::canCollideWith);
+-            List<Entity> entities = this.getEntities(entity, collisionBox.inflate(1.0E-7), predicate);
++            List<Entity> entities = this.getEntities(entity, collisionBox.inflate(1.0E-7), predicate); // Paper - Call HangingBreakByEntityEvent for collision with boats; diff on change
+             if (entities.isEmpty()) {
+                 return List.of();
+             } else {
 @@ -71,6 +_,12 @@
          }
      }
--- a/paper-server/patches/sources/net/minecraft/world/level/Level.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/Level.java.patch
@@ -533,7 +533,7 @@
              return chunk.getBlockState(pos);
          }
      }
-@@ -454,32 +_,49 @@
+@@ -454,32 +_,53 @@
              this.pendingBlockEntityTickers.clear();
          }
  
@@ -585,6 +585,10 @@
 +    public boolean noCollision(@Nullable Entity entity, AABB box) {
 +        if (entity instanceof net.minecraft.world.entity.decoration.ArmorStand && !entity.level().paperConfig().entities.armorStands.doCollisionEntityLookups)
 +            return false;
++        // Paper start - Call HangingBreakByEntityEvent for collision with boats
++        if (entity instanceof net.minecraft.world.entity.decoration.HangingEntity hangingEntity)
++            hangingEntity.collidingEntities = this.getEntities(entity, box.inflate(1.0E-7), net.minecraft.world.entity.EntitySelector.NO_SPECTATORS.and(entity::canCollideWith));
++        // Paper end - Call HangingBreakByEntityEvent for collision with boats
 +        return LevelAccessor.super.noCollision(entity, box);
 +    }
 +    // Paper end - Option to prevent armor stands from doing entity lookups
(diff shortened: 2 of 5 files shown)


paper-server/patches/sources/net/minecraft/world/level/EntityGetter.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/level/EntityGetter.java
 +++ b/net/minecraft/world/level/EntityGetter.java
+@@ -56,7 +_,7 @@
+             return List.of();
+         } else {
+             Predicate<Entity> predicate = entity == null ? EntitySelector.CAN_BE_COLLIDED_WITH : EntitySelector.NO_SPECTATORS.and(entity::canCollideWith);
+-            List<Entity> entities = this.getEntities(entity, collisionBox.inflate(1.0E-7), predicate);
++            List<Entity> entities = this.getEntities(entity, collisionBox.inflate(1.0E-7), predicate); // Paper - Call HangingBreakByEntityEvent for collision with boats; diff on change
```
Warriorrrr, line 8: You could use Util.COLLISION_EPSILON as a constant rather than adding a diff on change.

paper-server/patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -585,6 +585,10 @@
 +    public boolean noCollision(@Nullable Entity entity, AABB box) {
 +        if (entity instanceof net.minecraft.world.entity.decoration.ArmorStand && !entity.level().paperConfig().entities.armorStands.doCollisionEntityLookups)
 +            return false;
++        // Paper start - Call HangingBreakByEntityEvent for collision with boats
++        if (entity instanceof net.minecraft.world.entity.decoration.HangingEntity hangingEntity)
++            hangingEntity.collidingEntities = this.getEntities(entity, box.inflate(1.0E-7), net.minecraft.world.entity.EntitySelector.NO_SPECTATORS.and(entity::canCollideWith));
```
Warriorrrr, line 590: Couldn't this be moved into the block attached entity class itself?

notTamion: thats what i did initially though then i also have to additionally check for block collisions to ensure the order stays the same so i thought it might be easier to just move this to where the entities are still returned. but yeah this is not optimal either

Owen1212055 (changes requested):
I am in general not really sure about this change.

As collision stuff like this kinda gets yucky.. especially when I feel like there may be multiple hard colliding entities in a hanging entities hitbox that can be causing it to break. Passing this context as a field is also... not 100%. So the order is "undefined" here.

Additionally this is skipping all of the additional checks in the hard collision checks, such as the canCollideWith etc checks.

Quite frankly, I am just not sure if it is worth exposing this.

notTamion:
> I feel like there may be multiple hard colliding entities in a hanging entities hitbox that can be causing it to break. Passing this context as a field is also... not 100%. So the order is "undefined" here

yeah i mentioned that we might want to add a getRemovers method which returns a list in the original pr description.

though given this is pretty hacky and its not really a huge issue i will be closing this PR.

PR #11902, opened 2025-01-04
Fix InventoryAction wrong for Bundles

notTamion:
pre hardfork #11313

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/inventory/InventoryAction.java
+++ b/paper-api/src/main/java/org/bukkit/event/inventory/InventoryAction.java
@@ -93,5 +93,37 @@ public enum InventoryAction {
      * An unrecognized ClickType.
      */
     UNKNOWN,
+    /**
+     * The first stack of items in the clicked bundle is moved to the cursor.
+     */
+    PICKUP_FROM_BUNDLE,
+    /**
+     * All of the items on the clicked slot are moved into the bundle on the cursor.
+     */
+    PICKUP_ALL_INTO_BUNDLE,
+    /**
+     * Some of the items on the clicked slot are moved into the bundle on the cursor.
+     */
+    PICKUP_SOME_INTO_BUNDLE,
+    /**
+     * One of the items on the clicked slot is moved into the bundle on the cursor.
+     */
+    PICKUP_ONE_INTO_BUNDLE,
+    /**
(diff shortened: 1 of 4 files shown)


paper-api/src/main/java/org/bukkit/event/inventory/InventoryAction.java
```diff
@@ -93,5 +93,37 @@ public enum InventoryAction {
...
+     */
+    PICKUP_ALL_INTO_BUNDLE,
+    /**
+     * Some of the items on the clicked slot are moved into the bundle on the cursor.
+     */
+    PICKUP_SOME_INTO_BUNDLE,
+    /**
+     * One of the items on the clicked slot is moved into the bundle on the cursor.
+     */
+    PICKUP_ONE_INTO_BUNDLE,
```
lynxplay: Is there any difference in ONE vs ALL/SOME?
ALL means the entire stack is gone, SOME means some is left.
What is the benefit of ONE here? Unless I am missing a mechanic of only inserting ONE item.

notTamion: yeah i think i for some reason thought that such a mechanic existed. my mistake

EDIT: Or i for some reason did it to be consistent with the other ones even though it doesn't make any sense for bundles don't remember

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -2056,11 +2056,23 @@
 +                                        ItemStack cursor = this.player.containerMenu.getCarried();
 +                                        if (clickedItem.isEmpty()) {
 +                                            if (!cursor.isEmpty()) {
-+                                                action = packet.getButtonNum() == 0 ? InventoryAction.PLACE_ALL : InventoryAction.PLACE_ONE;
++                                                // Paper start - Fix InventoryAction wrong for Bundles
```
lynxplay: No need for paper comments, all of this is file patches xD

notTamion: oh right, forgot that method was basically turned into pure cb code

Machine-Maker (approved):

lynxplay (approved):

PR #11903, opened 2025-01-04
Add API for client-side signs

Sytm:
Allows using the editing of signs that are not placed on the server as text-input without meddling with packets.

Original: https://github.com/PaperMC/Paper/pull/9786

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
@@ -1,0 +1,94 @@
+package io.papermc.paper.event.packet;
+
+import io.papermc.paper.math.BlockPosition;
+import io.papermc.paper.math.Position;
+import java.util.Collections;
+import java.util.List;
+import net.kyori.adventure.text.Component;
+import org.bukkit.block.sign.Side;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called when a client attempts to modify a sign, but the location at which the sign should be edited
+ * has not yet been checked for the existence of a real sign.
+ * <p>
+ * Cancelling this event will prevent further processing of the sign change, but needs further handling
+ * by the plugin as the client's local world might be in an inconsistent state.
+ *
+ * @see Player#openVirtualSign(Position, Side)
+ */
+@NullMarked
+@ApiStatus.Experimental
+public class UncheckedSignChangeEvent extends PlayerEvent implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+    private boolean cancel = false;
+    private final BlockPosition editedBlockPosition;
+    private final Side side;
+    private final List<Component> lines;
+
+    @ApiStatus.Internal
+    public UncheckedSignChangeEvent(
+        final Player editor,
+        final BlockPosition editedBlockPosition,
+        final Side side,
+        final List<Component> lines
+    ) {
+        super(editor);
+        this.editedBlockPosition = editedBlockPosition;
+        this.side = side;
+        this.lines = lines;
+    }
+
+    /**
+     * Gets the location at which a potential sign was edited.
+     *
+     * @return location where the change happened
+     */
+    public BlockPosition getEditedBlockPosition() {
+        return editedBlockPosition;
+    }
+
+    /**
+     * Gets which side of the sign was edited.
+     *
+     * @return {@link Side} that was edited
+     */
+    public Side getSide() {
+        return side;
+    }
+
+    /**
+     * Gets the lines that the player has entered.
+     *
+     * @return the lines
+     */
+    public @Unmodifiable List<Component> lines() {
+        return Collections.unmodifiableList(lines);
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return cancel;
+    }
+
+    @Override
+    public void setCancelled(final boolean cancel) {
+        this.cancel = cancel;
+    }
+
+    @Override
+    public HandlerList getHandlers() {
+        return HANDLER_LIST;
+    }
+
+    public static HandlerList getHandlerList() {
+        return HANDLER_LIST;
+    }
+}
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -12,6 +12,7 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import io.papermc.paper.entity.LookAnchor;
 import io.papermc.paper.entity.PlayerGiveResult;
+import io.papermc.paper.math.Position;
 import org.bukkit.BanEntry;
 import org.bukkit.DyeColor;
 import org.bukkit.Effect;
@@ -52,7 +53,6 @@ import org.bukkit.plugin.Plugin;
 import org.bukkit.plugin.messaging.PluginMessageRecipient;
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.potion.PotionEffectType;
-import org.bukkit.profile.PlayerProfile;
 import org.bukkit.scoreboard.Scoreboard;
 import org.jetbrains.annotations.ApiStatus;
 import org.jspecify.annotations.NullMarked;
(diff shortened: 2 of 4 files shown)


paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Called when a sign edit packet has been received, but the location at which the sign should be edited
+ * has not yet been checked for the existence of a real sign.
+ * <p>
+ * Cancelling this event will prevent further processing of the sign change, but needs further handling
+ * by the plugin because the local world might be in an inconsistent state.
+ */
+@NullMarked
+public class UncheckedSignChangeEvent extends PlayerEvent implements Cancellable {
```
lynxplay, line 28: Idk if I like the prefix "unchecked". I get what this is meant to do, but couldn't we just call it `UnknownSignChangeEvent` similar to `PlayerUseUnknownEntityEvent` and only call this in the path of no sign actually being there?

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -2471,6 +2471,19 @@
          this.player.resetLastActionTime();
          ServerLevel serverLevel = this.player.serverLevel();
          BlockPos pos = packet.getPos();
+         if (serverLevel.hasChunkAt(pos)) {
++            // Paper start - Add API for client-side signs
++            if (!new io.papermc.paper.event.packet.UncheckedSignChangeEvent(this.player.getBukkitEntity(),
++                io.papermc.paper.util.MCUtil.toPosition(pos),
++                packet.isFrontText() ? org.bukkit.block.sign.Side.FRONT : org.bukkit.block.sign.Side.BACK,
++                filteredText.stream().map(line -> (net.kyori.adventure.text.Component) net.kyori.adventure.text.Component.text(line.raw())).toList())
```
lynxplay: Idk if raw is the right thing to use here when usually the filteredOrEmpty output is passed through StringUtil.filterText.

lynxplay (changes requested):

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3477,6 +3478,18 @@ default void setNoTickViewDistance(int viewDistance) {
...
+     * Open a sign for editing by the player.
+     * <p>
+     * The sign must only be placed locally for the player, which can be done with {@link #sendBlockChange(Location, BlockData)} and {@link #sendBlockUpdate(Location, TileState)}.
+     * A side-effect of this is that no events will be called for this action.
+     * </p>
+     *
+     * @param block The block where the client has a sign placed
+     * @param side The side to edit
+     */
+    void openLocalSign(@NotNull BlockPosition block, @NotNull Side side);
```
MiniDigger: is local the right name here? do we use that naming elsewhere? maybe virtual? 

and I think the event that is used for the change should be mentioned here

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -2471,6 +2471,19 @@
          this.player.resetLastActionTime();
          ServerLevel serverLevel = this.player.serverLevel();
          BlockPos pos = packet.getPos();
+         if (serverLevel.hasChunkAt(pos)) {
++            // Paper start - Add API for client-side signs
++            if (!new io.papermc.paper.event.packet.UncheckedSignChangeEvent(this.player.getBukkitEntity(),
```
MiniDigger: why is this event always called, shouldn't it be only called if the if below is false, as in, the server doesnt know a sign there? the method and the event are connected and should only work together the way I understand it, plus the the naming can reflect that

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -3061,6 +3064,11 @@ public void openSign(@NotNull Sign sign, @NotNull Side side) {
         CraftSign.openSign(sign, this, side);
     }
 
+    @Override
+    public void openLocalSign(@NotNull BlockPosition block, @NotNull Side side) {
```
Lulu13022002: It would be better for the user to not be that strict on the type and allow a regular Position and do the conversion later. And see below comment for the name.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -3061,6 +3064,11 @@ public void openSign(@NotNull Sign sign, @NotNull Side side) {
...
+        getHandle().connection.send(new ClientboundOpenSignEditorPacket(MCUtil.toBlockPos(block), side == Side.FRONT));
```
Lulu13022002: This should check if the connection is null like the method below

paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
```diff
@@ -0,0 +1,84 @@
...
+        this.side = side;
+        this.lines = lines;
+    }
+
+    /**
+     * Gets the location at which a potential sign was edited.
+     *
+     * @return location where the change happened
+     */
+    public BlockPosition getBlock() {
```
Lulu13022002: This name is a bit misleading without looking the return type / javadocs

paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
Sytm: I don't like to change it only when no sign is actually there, because when using [UnsafeValues#nextEntityId()](https://jd.papermc.io/paper/1.21.4/org/bukkit/UnsafeValues.html#nextEntityId()), you are guaranteed that your fake entity id is unique and therefore interacting with your fake entity will always call that event.

Contrary to that there might be a sign already placed somewhere, and searching for a free spot to "place" the sign seems kinda whack for something that is not persistent.

As for the name of the event, I am open for suggestions, because that is best name that I could come up taking my reasoning into account

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Sytm: When I opened the original PR I was honestly stumped how FilteredText is used on the server, so I guessed I'll use the raw text the player entered. Does anyone have proper insight how this system is actually used?

lynxplay: FilteredText isn't *really* used on paper at all, it's realms tech.
Raw imo is fine tho, this is "unchecked" afterall xD

Machine-Maker (approved):

paper-api/src/main/java/io/papermc/paper/event/packet/UncheckedSignChangeEvent.java
```diff
@@ -0,0 +1,88 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+import java.util.List;
+
+/**
+ * Called when a client attempts to modify a sign, but the location at which the sign should be edited
+ * has not yet been checked for the existence of a real sign.
+ * <p>
+ * Cancelling this event will prevent further processing of the sign change, but needs further handling
```
lynxplay, line 21: In regards to this, is there a point in us just resending the block?
Usage of this event would always be with "ephemeral" signs.

Leaving this up to plugin developers feels pretty meh.

Sytm: I also pondered about doing it that way, but at the end I thought the event should just report back that the player tried to edit a sign and minimize other side effects.

electronicboy: I mean, it's unchecked, and so I generally expect it to be a lot more 'dumb' in terms of how it interacts. I feel that such low level events are always going to be kinda meh

lynxplay: Well I would not mind it if this event was low level on purely client side signs.
But the fact that this is also called for actual in-world signs makes this pretty annoying/easy to fuck up potentially leaving normal signs out of sync.

Could potentially have some `cancel` overload that allows passing some RESEND flag?

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3477,6 +3477,20 @@ default void setNoTickViewDistance(int viewDistance) {
      */
     public void openSign(@NotNull Sign sign, @NotNull Side side);
 
+    /**
+     * Open a sign for editing by the player.
+     * <p>
+     * The sign must only be placed locally for the player, which can be done with {@link #sendBlockChange(Location, BlockData)} and {@link #sendBlockUpdate(Location, TileState)}.
+     * A side-effect of this is that normal events, like {@link org.bukkit.event.block.SignChangeEvent} will not be called (unless there is an actual sign in the world).
+     * </p>
```
ShaneBeee, line 3458: I think it should be noted that the fake sign needs to be within 9 blocks of the player.
(It's block interaction range + 4).

If it's too far away, the sign will immediately close.

lynxplay (approved):

Owen1212055 (approved):

PR #11906, opened 2025-01-04
Add EntityCollideWithEntityEvent

### LINKED ISSUE #8729: Entity with Entity collide event

notTamion:
resolves #8729
pre hardfork #9998

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
@@ -1,0 +1,54 @@
+package io.papermc.paper.event.entity;
+
+import org.bukkit.entity.Entity;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import java.util.List;
+
+/**
+ * Fired when two entities collide with each other.
+ * If cancelled, the entities won't get pushed away from each other.
+ */
+public class EntityCollideWithEntityEvent extends Event implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+    private boolean cancelled;
+    private final List<Entity> entities;
+
+    @ApiStatus.Internal
+    public EntityCollideWithEntityEvent(@NotNull Entity entity1, @NotNull Entity entity2) {
+        entities = List.of(entity1, entity2);
+    }
+
+    /**
+     * Returns the entities involved in this event
+     *
+     * @return entities that are involved in this event
+     */
+    public @NotNull List<Entity> getEntities() {
+        return entities;
+    }
+
+    @Override
+    public @NotNull HandlerList getHandlers() {
+        return HANDLER_LIST;
+    }
+
+    @NotNull
+    public static HandlerList getHandlerList() {
+        return HANDLER_LIST;
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return cancelled;
+    }
+
+    @Override
+    public void setCancelled(final boolean cancel) {
+        this.cancelled = cancel;
+    }
+}
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
@@ -11,6 +11,9 @@ import java.util.List;
 /**
  * Fired when two entities collide with each other.
  * If cancelled, the entities won't get pushed away from each other.
+ * <p>
+ * Note that even if cancelled, the client may still run its own collision unless
+ * disabled via player teams.
  */
 public class EntityCollideWithEntityEvent extends Event implements Cancellable {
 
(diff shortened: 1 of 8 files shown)


kennytv (commented):
I think I would prefer having getters for the two individually and to properly distinguish with who is pushing who, since there is an underlying order to it (?)

notTamion:
mentioned that in the description of the original pr. 
each entity is actually pushed twice, once during its own tick and once during the other entities tick. i don't think we should give the user the wrong idea by making it seem like one is pushing the other when that is not the case (that would also imply that cancelling the event would only cause one entity to not be pushed)
image

kennytv (approved):
In that case I'd just make it a list to make it easier to get them individually

notTamion:
Might have some weirdness with minecarts at the moment, need to take a closer look later (there seems to also be an inconsistency between when the VehicleEntityCollisionEvent is called between boats and minecarts).

Owen1212055 (dismissed):
Please check the handlerlist before executing this event, as this is hot code.

lulu2002:
hype

OfficialTheosis:
Would love for this to be added, allows protection/claiming plugins to prevent untrusted players from being able to push mobs around in protected regions.

iambibi:
> Would love for this to be added, allows protection/claiming plugins to prevent untrusted players from being able to push mobs around in protected regions.

Yes, I totally agree, especially with 26.2. Detecting collisions with the sulfur cube would be really practical.

paper-server/patches/sources/net/minecraft/world/entity/vehicle/minecart/AbstractMinecart.java.patch
```diff
@@ -130,13 +130,16 @@
              if (!entity.noPhysics && !this.noPhysics) {
 +                if (!this.level().paperConfig().collisions.allowVehicleCollisions && this.level().paperConfig().collisions.onlyPlayersCollide && !(entity instanceof Player)) return; // Paper - Collision option for requiring a player participant
                  if (!this.hasPassenger(entity)) {
-+                    // CraftBukkit start
++                    // Paper start - Collision Events
++                    if (io.papermc.paper.event.entity.EntityCollideWithEntityEvent.getHandlerList().getRegisteredListeners().length == 0 ||
```
kennytv:
```suggestion
+                    if (io.papermc.paper.event.entity.EntityCollideWithEntityEvent.getHandlerList().getRegisteredListeners().length != 0 &&
```

paper-server/patches/sources/net/minecraft/world/entity/vehicle/minecart/AbstractMinecart.java.patch
```diff
@@ -130,13 +130,16 @@
...
++                            !new io.papermc.paper.event.entity.EntityCollideWithEntityEvent(this.getBukkitEntity(), entity.getBukkitEntity()).callEvent()) return;
++
 +                    org.bukkit.event.vehicle.VehicleEntityCollisionEvent collisionEvent = new org.bukkit.event.vehicle.VehicleEntityCollisionEvent(
```
kennytv, line 136: Might as well add the same here

paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -614,11 +614,16 @@
...
  
      public final void setOldPosAndRot() {
-@@ -1618,6 +_,7 @@
+@@ -1618,6 +_,12 @@
      public void push(Entity entity) {
          if (!this.isPassengerOfSameVehicle(entity)) {
              if (!entity.noPhysics && !this.noPhysics) {
 +                if (this.level.paperConfig().collisions.onlyPlayersCollide && !(entity instanceof ServerPlayer || this instanceof ServerPlayer)) return; // Paper - Collision option for requiring a player participant
++                // Paper start - Add EntityCollideWithEntity Event
++                if (io.papermc.paper.event.entity.EntityCollideWithEntityEvent.getHandlerList().getRegisteredListeners().length == 0 ||
```
kennytv:
```suggestion
+                if (io.papermc.paper.event.entity.EntityCollideWithEntityEvent.getHandlerList().getRegisteredListeners().length != 0 &&
```

kennytv (changes requested):

kennytv (approved):

paper-api/src/main/java/io/papermc/paper/event/entity/EntityCollideWithEntityEvent.java
```diff
@@ -0,0 +1,54 @@
...
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import java.util.List;
+
+/**
+ * Fired when two entities collide with each other.
+ * If cancelled, the entities won't get pushed away from each other.
```
Warriorrrr, line 13: It's worth mentioning that when called with a player, under normal circumstances (without enable-player-collisions disabled or otherwise teams that disable collision) the client will still perform pushing logic and move itself even if the event is cancelled.

Clexus:
why not just have two methods getFirstEntity(), getSecondEntity() or getEntity1() getEntity2()? why a list?

PR #11907, opened 2025-01-04
Add RayTraceConfigurationBuilder

### LINKED ISSUE #9942: Allow changing the raySize of LivingEntity#rayTraceEntities

notTamion:
closes #9942
pre hardfork #10237

Make sure to look at my last comment on the old pre hardfork PR (i still think this is the case)


lynxplay:
Just to make sure I recall the previous design correctly, the idea was that the builder itself contained executing methods and allowed configuration of starting position and direction vector right?

I still do not believe it is a good design choice to have execution methods on the builder/executor type itself, specifically when it itself just calls back to the world. We could however move starting position and direction vector onto the builder too (e.g. in the factory method) and then also expose a wither method on the configuration itself so people can easily update it?

E.g.
```java
final var config = RayTraceConfiguration.builder()
  .direction(dir).origin(pos)
  .target(BLOCKS);
world.rayTrace(config);
world.rayTrace(config.relocate(pos, dir));
```

Just an idea tho, would love your input on that.

notTamion:
> We could however move starting position and direction vector onto the builder too (e.g. in the factory method) and then also expose a wither method on the configuration itself so people can easily update it?

Doesn't that just defeat the purpose of having an immutable configuration? 

Unrelated I just think that having a configuration for this just doesn't really work out. The adventure JoinConfiguration (which was used as an example in previous discussions) only defines a format. though here we would basically be defining the values in the configuration. (can't really find a better analogy)

Might be worth having some more people in the community give feedback on what they prefer as i am sure that personal preference is playing a big part here.

lynxplay:
Well that was your issue in the last comment on the pre-hardfork PR tho right?
The fact that consumers need to pass location and direction on each invocation.

The configuration would still be immutable, consumers that ray trace the *exact* same configuration would be fine and consumers that ray trace the exact "configuration" but from a different location/direction would only have the performance downsides of a single record copy.

This problem however isn't even solved with the previous approach discussed.
The only difference being that the previous approach was completely mutable and took over the responsibility of casting a ray. But yea I agree, more input is definitely needed.

paper-api/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilder.java
Machine-Maker, line 1: Use `@NullMarked` either on the interface or a package-info in the new package.

paper-api/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilder.java
```diff
@@ -0,0 +1,184 @@
+package io.papermc.paper.raytracing;
+
+import org.bukkit.FluidCollisionMode;
+import org.bukkit.Location;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Entity;
+import org.bukkit.util.Vector;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
Machine-Maker: After NullMarked, only need to use the jspecify Nullable annotation.

paper-api/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+    @Contract("_, _ -> this")
+    PositionedRayTraceConfigurationBuilder targets(@NotNull Targets first, @NotNull Targets... others);
+
+    /**
+     * List of Targets the builder can target.
+     */
+    enum Targets {
+        ENTITIES,
+        BLOCKS
+    }
```
Machine-Maker: Lets put this in a separate enum inside the package. If we ever add the buildable/reusable builder that doesn't have location/direction, we will regret having this enum inside this type. Make it just like `RayTraceTarget` with `ENTITY` and `BLOCK` or something.

Machine-Maker (changes requested):

paper-api/src/main/java/io/papermc/paper/raytracing/RayTraceTargets.java
Machine-Maker, line 1: Generally we don't like plurals in type names or enum constant names. So these should be like `RayTraceTarget` and `ENTITY` and `BLOCK`.

paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -1938,6 +1939,29 @@ default Iterable<? extends net.kyori.adventure.audience.Audience> audiences() {
     @Nullable RayTraceResult rayTrace(io.papermc.paper.math.@NotNull Position start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, double raySize, @Nullable Predicate<? super Entity> filter, @Nullable Predicate<? super Block> canCollide);
     // Paper end
 
+    /**
+     * Performs a ray trace that checks for both block and entity collisions.
```
lynxplay: Technically wrong, depends on what the builder targets right?

paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -1938,6 +1939,29 @@ default Iterable<? extends net.kyori.adventure.audience.Audience> audiences() {
...
+     * ignored as well regardless of the fluid collision mode.
+     * <p>
+     * Portal blocks are only considered passable if the ray starts within them.
+     * Apart from that collisions with portal blocks will be considered even if
+     * collisions with passable blocks are otherwise ignored.
+     * <p>
+     * This may cause loading of chunks! Some implementations may impose
+     * artificial restrictions on the maximum distance.
+     *
+     * @param builderConsumer consumer for the builder
```
lynxplay: a consumer to configure the ray trace configuration. The received builder is not valid for use outside of the consumer.

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    private double raySize = 0.0D;
+    private java.util.function.Predicate<? super org.bukkit.entity.Entity> entityFilter;
+    private java.util.function.Predicate<? super org.bukkit.block.Block> blockFilter;
+    private List<RayTraceTarget> targets;
+
+    public PositionedRayTraceConfigurationBuilderImpl(World world) {
+        this.world = world;
+    }
+
+
```
lynxplay: Formatting

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,141 @@
...
+import org.bukkit.World;
+import org.bukkit.util.RayTraceResult;
+import org.bukkit.util.Vector;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
+public class PositionedRayTraceConfigurationBuilderImpl implements PositionedRayTraceConfigurationBuilder {
```
lynxplay, line 16: `@NullMarked`

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+
+    @Override
+    public @NotNull Location start() {
+        return this.start;
+    }
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder start(final @NotNull Location start) {
```
lynxplay: Remove the jetbrains annotations here (and below)

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,141 @@
...
+        this.start = start;
+        return this;
+    }
+
+    @Override
+    public @NotNull Vector direction() {
+        return this.direction;
```
lynxplay: wrong annotation and needs to clone if not null.

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public @NotNull PositionedRayTraceConfigurationBuilder direction(final @NotNull Vector direction) {
+        this.direction = direction;
```
lynxplay: Defensive clone.

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    private boolean ignorePassableBlocks;
+    private double raySize = 0.0D;
+    private java.util.function.Predicate<? super org.bukkit.entity.Entity> entityFilter;
+    private java.util.function.Predicate<? super org.bukkit.block.Block> blockFilter;
+    private List<RayTraceTarget> targets;
+
+    @Override
+    public Location start() {
+        return this.start;
+    }
```
Machine-Maker: clone (cause mutable and we have a setter)

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder start(final Location start) {
+        this.start = start;
```
Machine-Maker: clone the parameter to it can't be changed later

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public Vector direction() {
+        return this.direction;
+    }
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder direction(final Vector direction) {
+        this.direction = direction;
+        return this;
```
Machine-Maker: a clone for you, and a clone for you, clones for everyone!

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder blockFilter(Predicate<? super org.bukkit.block.Block> blockFilter) {
+        this.blockFilter = blockFilter;
+        return this;
+    }
+
+    @Override
+    public List<RayTraceTarget> targets() {
+        return this.targets;
```
Machine-Maker: make it an unmodifiable view so you can't mutate the list, we have a setter.

paper-api/src/main/java/io/papermc/paper/raytracing/RayTraceTarget.java
```diff
@@ -0,0 +1,9 @@
+package io.papermc.paper.raytracing;
+
+/**
+ * List of Targets a builder can target.
+ */
+public enum RayTraceTarget {
+    ENTITIES,
+    BLOCKS
```
Machine-Maker: singular, not plural

Machine-Maker (changes requested):

Machine-Maker (approved):

lynxplay (changes requested):

paper-server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
```diff
@@ -1248,6 +1251,20 @@ public RayTraceResult rayTrace(io.papermc.paper.math.Position start, Vector dire
...
     }
 
+    @Override
+    public RayTraceResult rayTrace(Consumer<PositionedRayTraceConfigurationBuilder> builderConsumer) {
+        PositionedRayTraceConfigurationBuilderImpl builder = new PositionedRayTraceConfigurationBuilderImpl();
+
+        builderConsumer.accept(builder);
+
+        if (builder.targets().contains(RayTraceTarget.ENTITY)) {
+            if(builder.targets().contains(RayTraceTarget.BLOCK))
```
Lulu13022002: Missing space between if and (

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder blockFilter(Predicate<? super org.bukkit.block.Block> blockFilter) {
+        this.blockFilter = blockFilter;
+        return this;
+    }
+
+    @Override
+    public @UnmodifiableView List<RayTraceTarget> targets() {
+        return Collections.unmodifiableList(this.targets);
```
Lulu13022002: For builder without targets this will cause a NPE, might be better to make it a requirement

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public PositionedRayTraceConfigurationBuilder targets(final RayTraceTarget first, final RayTraceTarget ... others) {
+        List<RayTraceTarget> targets = new ArrayList<>(List.of(others));
```
Lulu13022002: An enum set would be better no need to have duplicate there

paper-server/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilderImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    @Nullable
+    private Location start;
+    @Nullable
+    private Vector direction;
+    private double maxDistance;
+    private org.bukkit.FluidCollisionMode fluidCollisionMode = org.bukkit.FluidCollisionMode.NEVER;
+    private boolean ignorePassableBlocks;
+    private double raySize = 0.0D;
+    private java.util.function.Predicate<? super org.bukkit.entity.Entity> entityFilter;
+    private java.util.function.Predicate<? super org.bukkit.block.Block> blockFilter;
```
Lulu13022002: All the types can be imported in this class

paper-api/src/main/java/io/papermc/paper/raytracing/PositionedRayTraceConfigurationBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Gets if the raytrace will ignore passable blocks when looking for block collisions.
+     *
+     * @return if the raytrace will ignore passable blocks
+     */
+    boolean ignorePassableBlocks();
+
+    /**
+     * Gets if the raytrace will ignore passable blocks when looking for block collisions.
```
Lulu13022002: This is not a getter

lynxplay (approved):

PR #11910, opened 2025-01-05
Fix parse of sound event in MusicInstrument

Doc94:
This PR is refer a issue mentioned in Discord where a item with a sound event in the instrument component throw a convert issue when call getItemMeta.
in a few tests i come to the CraftMusicInstrument and notice the use of a deprecate registry thing i change this and works but not really sure if this is the correct fix.. mostly the issue is ItemMeta and the expected things...

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/CraftMusicInstrument.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/CraftMusicInstrument.java
@@ -1,24 +1,24 @@
 package org.bukkit.craftbukkit;
 
 import com.google.common.base.Preconditions;
+import io.papermc.paper.registry.RegistryAccess;
+import io.papermc.paper.registry.RegistryKey;
 import io.papermc.paper.util.Holderable;
 import net.minecraft.core.Holder;
 import net.minecraft.core.registries.Registries;
 import net.minecraft.world.item.Instrument;
 import org.bukkit.MusicInstrument;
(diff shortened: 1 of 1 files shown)


Machine-Maker:
Lets instead add overloads to CraftRegistry.minecraftToBukkit and CraftRegistry.minecraftHolderToBukkit that take RegistryKey so others can use this and we don't have to do the full RegistryAccess.registryAccess()

Doc94:
> Lets instead add overloads to CraftRegistry.minecraftToBukkit and CraftRegistry.minecraftHolderToBukkit that take RegistryKey so others can use this and we don't have to do the full RegistryAccess.registryAccess()

you mean move the RegistryAccess.registryAccess() to CraftRegistry or the new method using the RegistryKey and handle in a own way?

Machine-Maker:
well actually, `minecraftToBukkit` and `minecraftHolderToBukkit` should just have the bukkit Registry parameter replaced with RegistryKey

Doc94:
> well actually, `minecraftToBukkit` and `minecraftHolderToBukkit` should just have the bukkit Registry parameter replaced with RegistryKey. And then all the usages should be updated. Shouldn't be passing a Registry instance anywhere, just the key, and then get the registry instance in CraftRegistry `minecraftToBukkit` and `minecraftHolderToBukkit`

hmm im lost in there xd
if you wanna handle because (maybe the time) dont fully understand the changes in the implementation

PR #11914, opened 2025-01-05
Disallow calling JavaPlugin#getCommand during onEnable for Paper plugins

NoahvdAa:
Every other day someone in #paper-dev tries to declare commands in paper-plugin.yml and use JavaPlugin#getCommand, which is not supported, but simply returns null instead of a clear error message.
This commit hopes to clarify that calling getCommand on self is not supported for paper plugins.

Note: check only applies during onEnable because we do not want to break the ability for other plugins to check if a plugin defines a command by calling JavaPlugin#getCommand on it

### DIFF
--- a/paper-api/src/main/java/org/bukkit/plugin/java/JavaPlugin.java
+++ b/paper-api/src/main/java/org/bukkit/plugin/java/JavaPlugin.java
@@ -26,7 +26,6 @@ import org.bukkit.generator.ChunkGenerator;
 import org.bukkit.plugin.PluginBase;
 import org.bukkit.plugin.PluginDescriptionFile;
 import org.bukkit.plugin.PluginLoader;
-import org.bukkit.plugin.PluginLogger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -53,6 +52,7 @@ public abstract class JavaPlugin extends PluginBase {
     private final io.papermc.paper.plugin.lifecycle.event.LifecycleEventManager<org.bukkit.plugin.Plugin> lifecycleEventManager = org.bukkit.Bukkit.getUnsafe().createPluginLifecycleEventManager(this, () -> this.allowsLifecycleRegistration);
     private boolean allowsLifecycleRegistration = true;
     // Paper end
+    private boolean isBeingEnabled = false;
 
     public JavaPlugin() {
         // Paper start
@@ -111,8 +111,7 @@ public abstract class JavaPlugin extends PluginBase {
     }
 
     /**
-     * Returns a value indicating whether or not this plugin is currently
-     * enabled
+     * Returns a value indicating whether this plugin is currently enabled
      *
      * @return true if this plugin is enabled, otherwise false
      */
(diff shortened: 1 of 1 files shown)


Leguan16:
related to #10588

paper-api/src/main/java/org/bukkit/plugin/java/JavaPlugin.java
```diff
@@ -339,9 +336,18 @@ public List<String> onTabComplete(@NotNull CommandSender sender, @NotNull Comman
...
+     * @throws UnsupportedOperationException if this plugin is a paper plugin and the method is called in {@link #onEnable()}
      */
     @Nullable
     public PluginCommand getCommand(@NotNull String name) {
+        if (this.isBeingEnabled && !(pluginMeta instanceof PluginDescriptionFile)) {
+            throw new UnsupportedOperationException(
+                "You are trying to call JavaPlugin#getCommand on a Paper plugin during startup" +
+                    " - you are probably trying to get a command you tried to define in paper-plugin.yml." +
+                    " Paper plugins do not support YAML-based command declarations!\n" +
+                    "Please check the documentation for more information on how to define commands in Paper plugins: https://docs.papermc.io/paper/dev/getting-started/paper-plugins#commands"
```
Machine-Maker: Use a nicer multiline java comment

paper-api/src/main/java/org/bukkit/plugin/java/JavaPlugin.java
```diff
@@ -339,9 +336,18 @@ public List<String> onTabComplete(@NotNull CommandSender sender, @NotNull Comman
...
      * @param name name or alias of the command
      * @return the plugin command if found, otherwise null
+     * @throws UnsupportedOperationException if this plugin is a paper plugin and the method is called in {@link #onEnable()}
      */
     @Nullable
     public PluginCommand getCommand(@NotNull String name) {
+        if (this.isBeingEnabled && !(pluginMeta instanceof PluginDescriptionFile)) {
+            throw new UnsupportedOperationException("""
+                You are trying to call JavaPlugin#getCommand on a Paper plugin during startup:
+                you are probably trying to get a command you tried to define in paper-plugin.yml.
```
Leguan16, line 346: Are these first lines really needed? I feel like the last two are enough or am I wrong?

NoahvdAa: I wanted to be extra clear, since the majority of people asking about this in paper-dev seem to be new to plugin development (or development in general), so I wanted to make sure they go to exactly the right place.
That being said, the stacktrace will also show the right location, so maybe it's indeed too verbose?

Leguan16: But the 2 lines basically say the same. 
Maybe something like this? 
```
Paper plugins do not support YAML-based command declarations!
You are probably trying to get a command via JavaPlugin#getCommand that you defined in paper-plugin.yml.
This method of command declaration is not supported when using Paper plugins.
Please check the documentation for more information on how to define commands in Paper plugins: https://docs.papermc.io/paper/dev/getting-started/paper-plugins#commands
```
Also do you guys rename Paper plugins to Lifecycle plugins? Maybe reflect that already?

RE:
> That being said, the stacktrace will also show the right location, so maybe it's indeed too verbose?

Without wanting to make assumptions i think that "new to plugin development (or development in general)" wont know how to read a stacktrace. So we should not consider that as a factor here. As you said you want to be extra clear and I think being a bit more verboste makes sense.

Machine-Maker (approved):

PR #11923, opened 2025-01-06
Update CustomTimingsHandler Nag Message

ghost:
No point telling them to switch to timings.of since timings will be removed soon


Doc94:
Maybe can be good mention spark here and maybe an alternative for what this class is used (?

lynxplay (dismissed):

lynxplay (changes requested):
Timings is already removed, only the API types are left as NO-OP.
Probably good to mention that in this message.

Regarding spark, I am not aware of anything similar spark offers, not that there is much of a point.

lynxplay (approved):

PR #11924, opened 2025-01-06
Prevent day cycle on the client when using fixed player time

SoSeDiK:
`player.setPlayerTime(time, false)` allows freezing time on the client, as per javadoc:
```
When using non relative time the player's time will stay fixed at the
specified time parameter.
```
Doing so makes the client move the skybox forward, and then jump back due to the server resyncing the time, causing a constant skybox stutter.
This can be prevented by accounting for a non-relative (fixed/frozen) player time alongside the real world's day cycle gamerule.

___
tldr; avoid skybox stutter by accounting for frozen player time / tell the client to not move skybox when using frozen player time

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/MinecraftServer.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/MinecraftServer.java.patch
@@ -984,7 +984,7 @@
              ObjectArrayList<GameProfile> list = new ObjectArrayList<>(min);
              int randomInt = Mth.nextInt(this.random, 0, players.size() - min);
  
-@@ -1046,17 +_,65 @@
+@@ -1046,17 +_,66 @@
      protected void tickChildren(BooleanSupplier hasTimeLeft) {
          ProfilerFiller profilerFiller = Profiler.get();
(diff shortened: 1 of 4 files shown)


paper-server/patches/sources/net/minecraft/server/MinecraftServer.java.patch
```diff
@@ -1027,8 +1027,9 @@
 +                }
 +                ServerPlayer entityplayer = (ServerPlayer) entityhuman;
 +                long playerTime = entityplayer.getPlayerTime();
-+                ClientboundSetTimePacket packet = (playerTime == dayTime) ? worldPacket :
-+                    new ClientboundSetTimePacket(worldTime, playerTime, doDaylight);
++                boolean relativeTime = entityplayer.relativeTime;
++                ClientboundSetTimePacket packet = ((relativeTime || !doDaylight) && playerTime == dayTime) ? worldPacket :
```
lynxplay, line 1031: What is the point in also checking the negated doDaylight here?

Just a performance improvement to potentially reuse the world packet?

SoSeDiK: Yea, if the world time is frozen already there's no need to spoof it, so we can reuse the packet

kennytv (approved):

PR #11926, opened 2025-01-06
Fixup luck and random implementation in CB loot-tables

florianreuth:
Looking at the git history this is broken since 1.14. This fixes the CB loot table API to properly support the luck and random setters / to actually do something with them.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
@@ -20,18 +20,16 @@
                      output.accept(itemStack);
                  } else {
                      int count = itemStack.getCount();
-@@ -140,10 +_,23 @@
-         }
+@@ -141,9 +_,22 @@
      }
  
--    public void fill(Container container, LootParams params, long seed) {
+     public void fill(Container container, LootParams params, long seed) {
 -        LootContext lootContext = new LootContext.Builder(params).withOptionalRandomSeed(seed).create(this.randomSequence);
-+    public void fillInventory(Container container, LootParams params, long seed) {
 +        // CraftBukkit start
-+        this.fillInventory(container, params, seed == 0L ? null : RandomSource.create(seed), false);
++        this.fill(container, params, seed == 0L ? null : RandomSource.create(seed), false);
 +    }
 +
-+    public void fillInventory(Container container, LootParams params, RandomSource randomSource, boolean plugin) {
++    public void fill(Container container, LootParams params, RandomSource randomSource, boolean plugin) {
 +        // CraftBukkit end
 +        LootContext lootContext = new LootContext.Builder(params).withOptionalRandomSource(randomSource).create(this.randomSequence);
          ObjectArrayList<ItemStack> randomItems = this.getRandomItems(lootContext);
--- a/paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
@@ -20,19 +20,20 @@
                      output.accept(itemStack);
                  } else {
                      int count = itemStack.getCount();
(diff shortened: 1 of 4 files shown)


paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
```diff
@@ -20,19 +20,20 @@
...
                      int count = itemStack.getCount();
-@@ -141,9 +_,22 @@
+@@ -141,9 +_,21 @@
      }
  
      public void fill(Container container, LootParams params, long seed) {
+-        LootContext lootContext = new LootContext.Builder(params).withOptionalRandomSeed(seed).create(this.randomSequence);
 +        // CraftBukkit start
-+        this.fillInventory(container, params, seed, false);
++        this.fillInventory(container, params, RandomSource.create(seed), false);
```
Lulu13022002: This is not completely equivalent, a seed of 0 still produce random loots.

florianreuth: Ah, thanks for pointing this out! :) I believe I should have fixed that now.

Lulu13022002: Thanks can you rename the method to restore the overload (before it made sense in 77fd87e4d053a82b1a733a773b235c25c2a6c6a7) if you don't mind?

florianreuth: Done, though I am not sure if that's a good idea since it generates more diff?

Lulu13022002: Yeah sorry i meant on the other side fillInventory -> fill

florianreuth: Oh sorry, that makes more sense, should be fixed now.

Should I mark the RandomSource field on the fill() with the plugin boolean as Nullable? Intellij shows me an warning that the field is NotNull but not sure if it matters or not since it's implementation code

Lulu13022002: That's up to you, but then you will just move the warning to withOptionalRandomSource which is not null too

paper-server/src/main/java/org/bukkit/craftbukkit/CraftLootTable.java
```diff
@@ -86,32 +87,29 @@ public Collection<ItemStack> populateLoot(Random random, LootContext context) {
...
         Preconditions.checkArgument(inventory != null, "Inventory cannot be null");
         Preconditions.checkArgument(context != null, "LootContext cannot be null");
-        LootParams nmsContext = this.convertContext(context, random);
+        LootParams nmsContext = this.convertContext(context);
         CraftInventory craftInventory = (CraftInventory) inventory;
         Container handle = craftInventory.getInventory();
 
         // TODO: When events are added, call event here w/ custom reason?
-        this.getHandle().fillInventory(handle, nmsContext, random.nextLong(), true);
+        this.getHandle().fill(handle, nmsContext, new RandomSourceWrapper(random), true);
```
Lulu13022002: random is Nullable in the API, so you get a NPE here and above

florianreuth: This was broken before, wasn't it? I pushed a fix lol

paper-server/patches/sources/net/minecraft/world/level/storage/loot/LootTable.java.patch
florianreuth: Hm yeah right, will just keep it as is

Warriorrrr (approved):

Lulu13022002 (approved):

Machine-Maker (approved):

PR #11940, opened 2025-01-09
Add PlayerLoadedWorldEvent

florianreuth:
Starting with Minecraft 1.21.4, the client sends a newly `player_loaded` packet as soon as the client closes the downloading terrain screen / starts rendering the world. This PR adds a new event so plugins can execute/delay code until a player is considered as loaded, which happens if they send the packet or if they haven't sent the packet for 60 ticks.

Same as https://github.com/PaperMC/Velocity/issues/1466

I would appreciate any feedback on the API.

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
@@ -1,0 +1,62 @@
+package io.papermc.paper.event.player;
+
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called when a player is set as loaded. This either happens when the player sends the player loaded packet after loading the world (closing the downloading terrain screen) or when the player
+ * has not sent the packet for 60 ticks after joining the server or respawning.
+ */
+@NullMarked
+public class PlayerLoadedEvent extends PlayerEvent {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Cause cause;
+
+    @ApiStatus.Internal
+    public PlayerLoadedEvent(final @NotNull Player who, final Cause cause) {
+        super(who);
+        this.cause = cause;
+    }
+
+    /**
+     * The cause of the player load.
+     *
+     * @return the cause
+     */
+    public Cause getCause() {
+        return this.cause;
+    }
+
+    @Override
+    public @NotNull HandlerList getHandlers() {
+        return HANDLER_LIST;
+    }
+
+    public static HandlerList getHandlerList() {
+        return HANDLER_LIST;
+    }
+
+    /**
+     * The cause of the {@link PlayerLoadedEvent}.
+     */
+    public enum Cause {
+
+        /**
+         * The event was triggered by the player sending the player loaded packet.
+         */
+        PACKET,
+
+        /**
+         * The event was triggered because the server has not received the player loaded packet for 60 ticks after the player joined the server or respawned.
+         */
+        TIMEOUT
+
+    }
+
+}
--- a/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
@@ -19,7 +19,7 @@ public class PlayerLoadedEvent extends PlayerEvent {
     private final Cause cause;
 
     @ApiStatus.Internal
-    public PlayerLoadedEvent(final @NotNull Player who, final Cause cause) {
+    public PlayerLoadedEvent(final Player who, final Cause cause) {
         super(who);
         this.cause = cause;
     }
--- a/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
@@ -373,12 +373,25 @@
                      this.awaitingPositionFromClient.x,
                      this.awaitingPositionFromClient.y,
                      this.awaitingPositionFromClient.z,
-@@ -495,6 +_,7 @@
+@@ -495,12 +_,20 @@
              this.lastGoodZ = this.awaitingPositionFromClient.z;
              this.player.hasChangedDimension();
              this.awaitingPositionFromClient = null;
 +            this.player.serverLevel().getChunkSource().move(this.player); // CraftBukkit
          }
+     }
+ 
+     @Override
+     public void handleAcceptPlayerLoad(ServerboundPlayerLoadedPacket packet) {
+         PacketUtils.ensureRunningOnSameThread(packet, this, this.player.serverLevel());
++        // Paper start - PlayerLoadedEvent
++        if (this.player.hasClientLoaded()) {
++            return;
++        }
++        final io.papermc.paper.event.player.PlayerLoadedEvent event = new io.papermc.paper.event.player.PlayerLoadedEvent(this.player.getBukkitEntity(), io.papermc.paper.event.player.PlayerLoadedEvent.Cause.PACKET);
++        event.callEvent();
++        // Paper end - PlayerLoadedEvent
+         this.player.setClientLoaded(true);
      }
  
 @@ -521,6 +_,7 @@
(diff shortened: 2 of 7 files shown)


paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@NullMarked
+public class PlayerLoadedEvent extends PlayerEvent {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Cause cause;
+
+    @ApiStatus.Internal
+    public PlayerLoadedEvent(final @NotNull Player who, final Cause cause) {
```
Doc94: Unnecesary NotNull if the class has the NullMarked annotation

paper-server/patches/sources/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -674,3 +674,25 @@
  
                  for (int i = 0; i < this.inventory.getContainerSize(); i++) {
                      ItemStack item = this.inventory.getItem(i);
+@@ -2089,12 +_,20 @@
+     }
+ 
+     public boolean hasClientLoaded() {
+-        return this.clientLoaded || this.clientLoadedTimeoutTimer <= 0;
++        return this.clientLoaded; // Paper - Add PlayerLoadedEvent
```
Doc94: This is used in another places.. this change not break any?

florianreuth: No, the boolean is set to true. Take a look at the changes below.

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
florianreuth: Yeah, the NotNull annotation was created by Intellij. Will remove

lynxplay (changes requested):
Generally a bit unsure on both the naming of the event and the exposed caused.
PlayerLoadedEvent can mean *a lot*, the packet name is fine-ish given it is a packet.
But in the context of the API, it could easily be confused for player data being loaded. As for alternative suggestions, *uh* nothing immediately comes to mind?
PlayerClientLoadedEvent?

In regards to the cause, I'd rather not introduce the concept of "packet" triggered.
Tho I am having a bit of a hard time to think of a usecase for the cause, so if you can elaborate on that, maybe that'll help my understanding :+1:

FlorianMichael:
Hm yeah I can try to find a better name for the class (my personal favourite would be `PlayerLoadedPacketEvent` or the name you provided).

I added the `Cause` so that people can make a difference between the client actually setting it's local state to loaded or the server doing it, I personally don't need it.

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
```diff
@@ -0,0 +1,62 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called when a player is set as loaded. This either happens when the player sends the player loaded packet after loading the world (closing the downloading terrain screen) or when the player
+ * has not sent the packet for 60 ticks after joining the server or respawning.
+ */
+@NullMarked
+public class PlayerLoadedEvent extends PlayerEvent {
```
masmc05: What do you think about renaming this to `PlayerWorldLoadEvent`? Its main cause is the client side world finishing being loaded, and the name would be consistent with <https://jd.papermc.io/paper/1.21.4/io/papermc/paper/event/packet/PlayerChunkLoadEvent.html>

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * The cause of the {@link PlayerLoadedEvent}.
+     */
+    public enum Cause {
+
+        /**
+         * The event was triggered by the player sending the player loaded packet.
+         */
+        PACKET,
```
masmc05: What do you think about renaming this to `PLAYER`? Since it's the player that triggers this, the api doesn't need to have details on how the player triggered this

masmc05 (commented):
Since lynxplay mentioned bad naming, I have some suggestions

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
masmc05: Also I think maybe just a boolean #timeout could work better than a full cause class, don't think it's gonna evolve into something more than either player triggers this, or server is tired of waiting and force mark as loaded

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
florianreuth: Thanks, that's actually a quite good name. I will rename the event!

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
florianreuth: Yeah, makes sense. Fixed

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -1,5 +1,21 @@
 --- a/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -57,7 +_,6 @@
+ import net.minecraft.network.protocol.Packet;
+ import net.minecraft.network.protocol.PacketUtils;
+ import net.minecraft.network.protocol.common.ServerboundClientInformationPacket;
+-import net.minecraft.network.protocol.common.ServerboundCustomPayloadPacket;
```
masmc05: useless diff

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -1,5 +1,21 @@
...
+ import net.minecraft.network.protocol.configuration.ConfigurationProtocols;
+ import net.minecraft.network.protocol.game.ClientboundBlockChangedAckPacket;
+ import net.minecraft.network.protocol.game.ClientboundBlockUpdatePacket;
+@@ -183,7 +_,6 @@
+ import net.minecraft.world.level.block.entity.JigsawBlockEntity;
+ import net.minecraft.world.level.block.entity.SignBlockEntity;
+ import net.minecraft.world.level.block.entity.StructureBlockEntity;
+-import net.minecraft.world.level.block.state.BlockBehaviour;
```
masmc05: useless diff

masmc05 (commented):
I think you accidentally removed unused imports in the newest commits

molor:
Why not `PlayerReadyEvent`? Because `World` already loaded at the time of event call, so current event name looks a little unclear IMO.

FlorianMichael:
I don't like the term `Ready` honestly, renamed it to `PlayerLoadedWorldEvent` which should be clear enough now.

Cryptite:
`PlayerLoadedWorldEvent` to me sounds too much still like `WorldLoadEvent` and doesn't strongly enough indicate this is a client-state/behavior event. Might I propose `PlayerFinishJoinEvent` or something of that ilk?

I just feel it's important not to name it something that sounds like Bukkit-level actions, but rather is a _later phase of joining a server_.

nostalfinals:
Maybe PlayerLoadedInWorldEvent?

lynxplay:
`PlayerReadyEvent` - player ready for what tho xD
`PlayerFinishJoinEvent' and the other one don't work because this is also called when the player respawns.

Realistically, it is an event that is called when the client finished loading into the world.
But yea, I'll throw some names around the PR parties if we can't find a good one xD

Malfrador:
`PlayerClientLoadedWorldEvent`? That would make it explicitly clear that it has nothing to do with Bukkit world loading. And the naming would align with another client-triggered event, the `PlayerClientOptionsChangeEvent`

lynxplay:
I am also good with `PlayerClientLoadedWorldEvent` but will discuss that with machine tomorrow prior to merge so that rename can be done by us if we do feel like it. Beyond that, needs testing but LGTM.

paper-api/src/main/java/io/papermc/paper/event/player/PlayerLoadedEvent.java
Lulu13022002: This is not completely resolved for getHandlers and can you move the handler related methods at the end of the class for consistency with other events. Javadocs could be spread on more lines.

florianreuth: @Lulu13022002 Thanks, pushed a fix

lynxplay (approved):

Machine-Maker (approved):

PR #11945, opened 2025-01-10
Fix hopper updates source slot twice

### LINKED ISSUE #11740: chiseled bookshelf does not get updated after hopper pulls book
### Expected behavior

If you place a book in a chiseled bookshelf with a hopper beneath it, the hopper, if it has free inv space, pulls the book out of the bookshelf and the book is visually removed from the bookshelf, the state (f3 slot_x_occupied) is set to false and the book is removed from its inv.

### Observed/Actual behavior

If there already are books inside the hopper and after pulling out the book from the bookshelf the book stacks with the books inside the hopper the bookshelf is not updated correctly. But if there are for example 32 books in the last slot of the hopper and the other slots are free everything works normally because the book gets transfered into the first hopper slot and does not stack with 32 books in the last slot. The bug only occurs if the book stacks with books in the hopper. If the bug occurs the book gets pulled into the hopper but is still displayed in the bookshelf but you cannot take it out since it was already pulled out by the hopper and you also cannot put a book in that slot. In f3 you can see the info slot_x_occupied is still true. Only after you insert a book into another slot or take another book out of another slot the bookshelf gets updated

### Steps/models to reproduce

1. place a hopper and on top a chiseled bookshelf
2. put a book into the first slot of the hopper
3. put a book into the bookshelf
4. the hopper pulled out the book and stacked onto the containing book but the booshelf still displays the book
5. empty the hopper and put one book into the last slot (or any other thats not the first)
6. put a book into the last slot of the hopper
7. put a book into the bookshelf
8. the hopper pulled out the book and put it into the first slot and the bookshelf is empty

### Plugin and Datapack List

Server Plugins (0):
There are 3 data pack(s) enabled: [vanilla (build-in)], [file/bukkit (world)], [paper (build-in)]
There are no more datapacks available

### Paper version

This server is running Paper version 1.21.1-132-ver/1.21.1@b48403b (2024-11-21-T10:14:27Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21.1-119-7cd4f2c (MC: 1.21.1)

### Other

_No response_

nostalfinals:
Fix #11740 

The `hopperPush`/`hopperPull` methods introduced by the [Optimize Hoppers](https://github.com/PaperMC/Paper/blob/main/paper-server/patches/features/0029-Optimize-Hoppers.patch) patch already update the source slot correctly after moving items:

- `hopperPush`:
  image

- `hopperPull`:
  image

However, in `HopperBlockEntity#tryMoveInItem`, if the target slot isn't empty (i.e., a merge happens), the source slot is updated once again by shrinking the source `ItemStack`. This causes issues with certain source containers, such as the chiseled bookshelf.

image

Since the `ChiseledBookShelfBlockEntity` checks whether the `removeItem` call is meaningful (i.e., ensuring it doesn't attempt to remove an already empty slot), and `tryMoveInItem` is called before `removeItem`, the block state of the current chiseled bookshelf isn't updated properly. As a result, the book still visually remains in the bookshelf:

image

The return value of `tryMoveInItem` is essential (it’s used by `hopperPush`/`hopperPull`), so I've added a copy before operating the source `ItemStack` to resolve this issue.

---
Download the paperclip jar for this pull request: [paper-11945.zip](https://nightly.link/PaperMC/Paper/actions/artifacts/2418542664.zip)

### DIFF
--- a/paper-server/patches/features/0029-Optimize-Hoppers.patch
+++ b/paper-server/patches/features/0029-Optimize-Hoppers.patch
@@ -60,7 +60,7 @@ index cc2d442682496197d29ace79b22e6cf6fb7edf5e..ae220a732c78ab076261f20b5a54c71d
              /* Drop global time updates
              if (this.tickCount % 20 == 0) {
 diff --git a/net/minecraft/world/item/ItemStack.java b/net/minecraft/world/item/ItemStack.java
-index 47bf8051da9a04ab333bb2bf6421d6253556b71c..ba31bbdac96e58d299dc260b47aaedf767735a42 100644
+index 0eb9811e573f0c38de3eeb17099872efcbb4c2dd..b511fe8295369ac4014beb351cd2e3f770c10170 100644
 --- a/net/minecraft/world/item/ItemStack.java
 +++ b/net/minecraft/world/item/ItemStack.java
 @@ -808,10 +808,16 @@ public final class ItemStack implements DataComponentHolder {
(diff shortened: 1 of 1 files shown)


twisti-dev:
Any updates on this?

lynxplay:
No, the gist being that our hopper patch may or may not be needed anymore in the first place given item copies are incredibly cheap post 1.21.5 and this PR copying the stack prior is kinda *that* but without dropping our patch. So yea, it's a bit iffy to review because really this is a review to determine if the patch itself is needed.

PR #11949, opened 2025-01-11
Fix wrong nullability

NonSwag:


NonSwag:
Also set the initial array list capacity

lynxplay (approved):

PR #11950, opened 2025-01-12
Rename getPotentialBedLocation to getPotentialRespawnLocation

masmc05:
Spigot renamed [getBedSpawnLocation](<https://jd.papermc.io/paper/1.21.4/org/bukkit/entity/Player.html#getBedSpawnLocation()>) to [getRespawnLocation](<https://jd.papermc.io/paper/1.21.4/org/bukkit/entity/Player.html#getRespawnLocation()>) in 1.20.4, this pr makes getPotentialBedLocation match that change, making it also more accessible to find by people who are currently using the new getRespawnLocation name.


lynxplay (changes requested):
default it in the API type. No point in keeping it around in internals too.

lynxplay (approved):

NoahvdAa (approved):

PR #11951, opened 2025-01-12
Add EntityHarvestBlockEvent

Doc94:
This starts in [discord](https://canary.discord.com/channels/289587909051416579/555462289851940864/1327837651876970569) where the Composter when drop a bone meal not has any event... the first thing was use the Drop or Dispenser event but that not handle correctly in this behaviour... then Machine Maker suggest a new event.

Then this PR add a new event for the case when a Entity harvest a block currently for Vines and Composter. this new event cannot be a base of the player side by the original design... then add in the docs the mention for this cases...

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
@@ -1,9 +1,10 @@
-package org.bukkit.event.entity;
+package io.papermc.paper.event.entity;
 
 import org.bukkit.block.Block;
 import org.bukkit.entity.Entity;
 import org.bukkit.event.Cancellable;
 import org.bukkit.event.HandlerList;
+import org.bukkit.event.entity.EntityEvent;
 import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.NotNull;
 import java.util.List;
@@ -22,7 +23,7 @@ import java.util.List;
  */
 public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable {
 
-    private static final HandlerList handlers = new HandlerList();
+    private static final HandlerList HANDLER_LIST = new HandlerList();
     private boolean cancel = false;
     private final Block harvestedBlock;
     private final List<ItemStack> itemsHarvested;
@@ -66,11 +67,11 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
     @NotNull
     @Override
     public HandlerList getHandlers() {
-        return handlers;
+        return HANDLER_LIST;
     }
 
     @NotNull
     public static HandlerList getHandlerList() {
-        return handlers;
+        return HANDLER_LIST;
     }
 }
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
@@ -51,6 +51,9 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
      * Gets a list of items that are being harvested from this block.
      *
      * @return A mutable list of items that are being harvested from this block
+     * @apiNote {@link org.bukkit.entity.Fox} has a behavior where, if it does not have an item
+     * in {@link org.bukkit.inventory.EquipmentSlot#HAND}, it will take one unit from the first
+     * item in its inventory stack and use it in the hand slot.
      */
     public List<ItemStack> getItemsHarvested() {
         return itemsHarvested;
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
@@ -6,7 +6,8 @@ import org.bukkit.event.Cancellable;
 import org.bukkit.event.HandlerList;
 import org.bukkit.event.entity.EntityEvent;
 import org.bukkit.inventory.ItemStack;
-import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NullMarked;
 import java.util.List;
 
 /**
@@ -21,14 +22,17 @@ import java.util.List;
  * for {@link org.bukkit.event.block.BlockBreakEvent} and
  * {@link org.bukkit.event.block.BlockDropItemEvent}.
  */
+@NullMarked
 public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable {
 
     private static final HandlerList HANDLER_LIST = new HandlerList();
+
     private boolean cancel = false;
     private final Block harvestedBlock;
     private final List<ItemStack> itemsHarvested;
 
-    public EntityHarvestBlockEvent(@NotNull Entity entity, @NotNull Block harvestedBlock, @NotNull List<ItemStack> itemsHarvested) {
+    @ApiStatus.Internal
+    public EntityHarvestBlockEvent(final Entity entity, final Block harvestedBlock, final List<ItemStack> itemsHarvested) {
         super(entity);
         this.harvestedBlock = harvestedBlock;
         this.itemsHarvested = itemsHarvested;
@@ -39,7 +43,6 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
      *
      * @return The block that is being harvested
      */
-    @NotNull
     public Block getHarvestedBlock() {
         return harvestedBlock;
     }
@@ -49,7 +52,6 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
      *
      * @return A list of items that are being harvested from this block
      */
-    @NotNull
     public List<ItemStack> getItemsHarvested() {
         return itemsHarvested;
     }
@@ -64,13 +66,11 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
         this.cancel = cancel;
     }
 
-    @NotNull
     @Override
     public HandlerList getHandlers() {
         return HANDLER_LIST;
     }
 
-    @NotNull
     public static HandlerList getHandlerList() {
         return HANDLER_LIST;
     }
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
@@ -50,7 +50,7 @@ public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable 
     /**
      * Gets a list of items that are being harvested from this block.
      *
-     * @return A list of items that are being harvested from this block
+     * @return A mutable list of items that are being harvested from this block
      */
     public List<ItemStack> getItemsHarvested() {
         return itemsHarvested;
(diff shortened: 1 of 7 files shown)


paper-api/src/main/java/org/bukkit/event/player/PlayerHarvestBlockEvent.java
```diff
@@ -5,13 +5,16 @@
...
 import org.bukkit.event.HandlerList;
+import org.bukkit.event.entity.EntityHarvestBlockEvent;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * This event is called whenever a player harvests a block.
  * <br>
+ * For the entity case please use {@link org.bukkit.event.entity.EntityHarvestBlockEvent}
```
Warriorrrr: I think this line here and in the new event's javadoc could be a bit better, since there's an import it's also not necessary to use the fully qualified name.
```suggestion
 * For cases involving entities, please use {@link EntityHarvestBlockEvent}.
```

paper-api/src/main/java/org/bukkit/event/entity/EntityHarvestBlockEvent.java
```diff
@@ -0,0 +1,76 @@
+package org.bukkit.event.entity;
```
Warriorrrr: New events should go into the paper package

paper-api/src/main/java/org/bukkit/event/entity/EntityHarvestBlockEvent.java
```diff
@@ -0,0 +1,76 @@
...
+ * A 'harvest' is when a block drops an item (usually some sort of crop) and
+ * changes state, but is not broken in order to drop the item.
+ * <br>
+ * This event is not called for when a block is broken, to handle that, listen
+ * for {@link org.bukkit.event.block.BlockBreakEvent} and
+ * {@link org.bukkit.event.block.BlockDropItemEvent}.
+ */
+public class EntityHarvestBlockEvent extends EntityEvent implements Cancellable {
+
+    private static final HandlerList handlers = new HandlerList();
```
Warriorrrr: Should be HANDLER_LIST

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -344,11 +345,22 @@ public static EntityEnterLoveModeEvent callEntityEnterLoveModeEvent(net.minecraf
...
+        Bukkit.getPluginManager().callEvent(entityHarvestBlockEvent);
+        return entityHarvestBlockEvent;
+    }
+
     /**
      * Player Harvest Block Event
      */
     public static PlayerHarvestBlockEvent callPlayerHarvestBlockEvent(Level world, BlockPos blockposition, net.minecraft.world.entity.player.Player who, InteractionHand enumhand, List<ItemStack> itemsToHarvest) {
-        List<org.bukkit.inventory.ItemStack> bukkitItemsToHarvest = new ArrayList<>(itemsToHarvest.stream().map(CraftItemStack::asBukkitCopy).collect(Collectors.toList()));
+        List<org.bukkit.inventory.ItemStack> bukkitItemsToHarvest = itemsToHarvest.stream().map(CraftItemStack::asBukkitCopy).collect(Collectors.toList());
```
Warriorrrr, line 335: While this does work I think it's important to note that Collectors#toList doesn't guarantee that it's mutable, but since that's unlikely to change it's probably fine.

paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    public Block getHarvestedBlock() {
+        return harvestedBlock;
+    }
+
+    /**
+     * Gets a list of items that are being harvested from this block.
+     *
+     * @return A list of items that are being harvested from this block
+     */
+    public List<ItemStack> getItemsHarvested() {
```
Warriorrrr, line 60: It could be noted here and in the player event that this list is mutable.

Warriorrrr (changes requested):
The case of foxes harvesting a sweet berry bush is currently missing, can be added in the FoxEatBerriesGoal#pickSweetBerries method

Doc94:
> The case of foxes harvesting a sweet berry bush is currently missing, can be added in the FoxEatBerriesGoal#pickSweetBerries method

ready i try to handle the whole logic where the "first" item its used for the mount of the fox and the rest for drop.

paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    public Block getHarvestedBlock() {
+        return harvestedBlock;
+    }
+
+    /**
+     * Gets a list of items that are being harvested from this block.
+     *
+     * @return A mutable list of items that are being harvested from this block
+     */
+    public List<ItemStack> getItemsHarvested() {
```
notTamion, line 60: i would maybe add that for foxes the first item will be used for its mouth

paper-server/patches/sources/net/minecraft/world/level/block/ComposterBlock.java.patch
```diff
@@ -48,6 +48,33 @@
          if (!level.isClientSide) {
              Vec3 vec3 = Vec3.atLowerCornerWithOffset(pos, 0.5, 1.01, 0.5).offsetRandom(level.random, 0.7F);
              ItemEntity itemEntity = new ItemEntity(level, vec3.x(), vec3.y(), vec3.z(), new ItemStack(Items.BONE_MEAL));
+             itemEntity.setDefaultPickUpDelay();
```
notTamion: can also remove this call as we call it later anyway

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -344,11 +345,22 @@ public static EntityEnterLoveModeEvent callEntityEnterLoveModeEvent(net.minecraf
...
     }
 
+    /**
+     * Entity Harvest Block Event
+     */
+    public static EntityHarvestBlockEvent callEntityHarvestBlockEvent(Level world, BlockPos blockposition, net.minecraft.world.entity.Entity who, List<ItemStack> itemsToHarvest) {
+        List<org.bukkit.inventory.ItemStack> bukkitItemsToHarvest = itemsToHarvest.stream().map(CraftItemStack::asBukkitCopy).collect(Collectors.toList());
+        org.bukkit.entity.Entity entity = who.getBukkitEntity();
+        EntityHarvestBlockEvent entityHarvestBlockEvent = new EntityHarvestBlockEvent(entity, CraftBlock.at(world, blockposition), bukkitItemsToHarvest);
+        Bukkit.getPluginManager().callEvent(entityHarvestBlockEvent);
```
notTamion:
```suggestion
        entityHarvestBlockEvent.callEvent();
```

paper-server/patches/sources/net/minecraft/world/entity/animal/Fox.java.patch
```diff
@@ -130,11 +130,41 @@
...
-@@ -934,6 +_,7 @@
+@@ -934,15 +_,34 @@
          private void pickSweetBerries(BlockState state) {
              int ageValue = state.getValue(SweetBerryBushBlock.AGE);
              state.setValue(SweetBerryBushBlock.AGE, Integer.valueOf(1));
 +            if (!org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(Fox.this, this.blockPos, state.setValue(SweetBerryBushBlock.AGE, 1))) return; // CraftBukkit - call EntityChangeBlockEvent
              int i = 1 + Fox.this.level().random.nextInt(2) + (ageValue == 3 ? 1 : 0);
              ItemStack itemBySlot = Fox.this.getItemBySlot(EquipmentSlot.MAINHAND);
++            // Paper start - call HarvestBlockEvent for the item dropped
++            ItemStack itemHarvest = new ItemStack(Items.SWEET_BERRIES, i);
```
notTamion, line 157: Haven't tested but just a thought to make sure. considering in the vanilla logic we would popresource i times that would produce multiple items on the ground that are not immediately bundled together right? though since we just creating one itemstack with i amount here those would all immediately be bundled together when they get spawned. not sure but just going to leave it here so we don't overlook it

paper-server/patches/sources/net/minecraft/world/level/block/ComposterBlock.java.patch
```diff
@@ -48,6 +48,33 @@
...
+-            level.addFreshEntity(itemEntity);
++            // Paper start - call HarvestBlockEvent for the item dropped
++            // level.addFreshEntity(itemEntity); // used later
++            java.util.ArrayList<org.bukkit.inventory.ItemStack> itemsHarvested = new java.util.ArrayList<>();
++            if (entity instanceof Player player) {
++                org.bukkit.event.player.PlayerHarvestBlockEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callPlayerHarvestBlockEvent(level, pos, player, InteractionHand.MAIN_HAND, java.util.Collections.singletonList(itemEntity.getItem()));
++                itemsHarvested.addAll(event.getItemsHarvested());
```
notTamion: no need to clone the list here since we aren't modifying it later.

notTamion (changes requested):

paper-api/src/main/java/io/papermc/paper/event/entity/EntityHarvestBlockEvent.java
Doc94: Like apiNote?

notTamion: yeah or just normally in the comment. both is fine with me

paper-server/patches/sources/net/minecraft/world/entity/animal/Fox.java.patch
Doc94: maybe i bad read this but the logic is the same than vanilla, the event is called with only one item and is used for pop, if you add more items then pop for every element.. the only thing is if you make a thing like "add potato", "add potato" and that are different items added in the list... but still respect the original thing.

notTamion: oh yeah nvm i read the original if statement as a while statement

notTamion (approved):
LGTM

Warriorrrr (approved):

Owen1212055:
I'm a little torn, because of the fact that we arent sharing any inheritance I really think we should make the EntityHarvestBlockEvent "more proper".

- clone itemstacks, add getters/setters
- make the list immutable, add a setter/getter

Issue is then people wanting to use the other event may get a bit confused.

Doc94:
> I'm a little torn, because of the fact that we arent sharing any inheritance I really think we should make the EntityHarvestBlockEvent "more proper".
> 
> * clone itemstacks, add getters/setters
> * make the list immutable, add a setter/getter
> 
> Issue is then people wanting to use the other event may get a bit confused.

yeah i prefer keep the two events like the same.. if not then modify the two with this new "behaviours" but set itemstack? i mean all are in the list or i bad read something?

Warriorrrr (changes requested):
Looking back at this now it would indeed seem better to make the new event more proper with defensive copying for the list and items and adding setters.

Doc94:
> Looking back at this now it would indeed seem better to make the new event more proper with defensive copying for the list and items and adding setters.

like Owen mention not?
clone items and make list can be replaced but not allow modify the items not?

PS: Another PR in the conflict list :(

Doc94:
Ok now the entity event handle clone of items an the immutable list.... not sure if need make this to the player event too (what is a breaking change) or this is ready....

PR #11953, opened 2025-01-12
Fix issue in fillUsableCommands where child node requirements were not checked during redirection command flattening

LemonCaramel:
I recently found out that admin subcommands are visible to all players.


Warriorrrr:
Thank you for your work on this PR, though we've instead decided to go with #11954 as that PR would fix the whole underlying problem behind the issue being fixed here.

PR #11956, opened 2025-01-12
Fix locate command dist overflow/underflow

HaHaWTH:
This pull request fixes [MC-177381](https://bugs.mojang.com/browse/MC-177381), which causes distance of structure overflow/underflow during calculation.

### Changes

Replace `Mth.sqrt` in LocateCommand with `Math.hypot` to prevent any overflow/underflow.

This can be reproduced in latest Paper build with seed `-1064495411872206035`, use /locate command to search for a stronghold at coordinate -2735 ~ 688484.


kennytv (approved):
Thanks!

PR #11958, opened 2025-01-12
Expose all possible block data states

masmc05:
Adds a possibility to get a set of all possible combinations of block proprieties in a block data.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/block/BlockType.java
+++ b/paper-api/src/main/java/org/bukkit/block/BlockType.java
@@ -121,6 +121,7 @@ import org.bukkit.inventory.ItemType;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.Unmodifiable;
 
 /**
  * While this API is in a public interface, it is not intended for use by
@@ -176,9 +177,8 @@ public interface BlockType extends Keyed, Translatable, net.kyori.adventure.tran
          *
          * @return new block data collection
          */
-        @NotNull
         @Override
-        Collection<B> createBlockDataStates();
+        @Unmodifiable @NotNull Collection<B> createBlockDataStates();
 
         /**
          * Creates a new {@link BlockData} instance for this block type, with all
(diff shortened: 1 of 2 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftBlockType.java
```diff
@@ -147,6 +150,16 @@ public B createBlockData() {
         return this.createBlockData((String) null);
     }
 
+    @Override
+    public @NotNull Set<B> createBlockDataStates() {
+        return this.block.getStateDefinition()
```
lynxplay: If immutable then why not cached?

masmc05: Block datas needs to be cloned on every get, since the internal nms block state is mutable

lynxplay: Right, I forget the API is dumb.

paper-api/src/main/java/org/bukkit/block/BlockType.java
```diff
@@ -169,6 +170,16 @@ interface Typed<B extends BlockData> extends BlockType {
...
 
+        /**
+         * Creates a set of {@link BlockData} instances for this block type, with all
+         * possible combinations of properties values.
+         *
+         * @return new block data set
+         */
+        @NotNull
+        @Override
+        Set<B> createBlockDataStates();
```
Machine-Maker: The return type shouldn't be `Set`. The contents are mutable objects which can break set invariants. The type shouldn't be `List` either because the order and random access shouldn't be used, so I think Collection is the best option.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftBlockType.java
```diff
@@ -147,6 +149,16 @@ public B createBlockData() {
...
+            .getPossibleStates()
+            .stream()
+            .map(BlockState::createCraftBlockData)
+            .map(this.blockDataClass::cast)
+            .collect(ImmutableSet.toImmutableSet());
```
Machine-Maker: The implementation type to the Collection returned can just be a list.

Machine-Maker (changes requested):

paper-api/src/main/java/org/bukkit/block/BlockType.java
masmc05: As a wise man said above
> Right, I forget the API is dumb.

Machine-Maker:
> Applied the requested changes

javadocs still say "set"

Machine-Maker (approved):

lynxplay (approved):

PR #11959, opened 2025-01-12
Deprecate Turtle#isDigging

Lulu13022002:
The method is now redundant with what spigot added some time ago


paper-api/src/main/java/org/bukkit/entity/Turtle.java
```diff
@@ -49,14 +48,17 @@ public interface Turtle extends Animals {
      * Get if turtle is digging to lay eggs
      *
      * @return True if digging
+     * @deprecated in favor of {@link #isLayingEgg()}
      */
-    boolean isDigging();
+    @Deprecated
```
Machine-Maker: `since=`

(reminds me, I need to add a checkstyle check to require that...)

Machine-Maker (approved):

lynxplay (approved):

PR #11960, opened 2025-01-12
Fix inconsistencies between offline/online spawn position getter

### LINKED ISSUE #7860: Player vs OfflinePlayer getBedSpawnLocation

### LINKED ISSUE #12338: Abnormal API behavior when obtaining the player's respawn position causes Anchor's charging to be consumed
### Expected behavior

When `player.getBedSpawnLocation()` or `player.getRespawnLocation()` is called, the API will be running as usual if the player's respawn block is Anchor

### Observed/Actual behavior

The API works normally, but Anchor's charging is consumed abnormally

### Steps/models to reproduce

1. Go to nether
2. Place an Anchor and charge it, then right-click to set the respawn point.
3. Call API: `player.getBedSpawnLocation()` or `player.getRespawnLocation()`
4. Unexpected behavior can be observed: Anchor charging decreases

### Plugin and Datapack List

[17:46:08 INFO]: ℹ Server Plugins (3):
[17:46:08 INFO]: Bukkit Plugins:
[17:46:08 INFO]: - LetsExpand, PlugManX, spark

### Paper version

version
[17:46:28 INFO]: Checking version, please wait...
[17:46:30 INFO]: This server is running Paper version 1.21.4-221-main@c467df9 (2025-03-24T00:18:41Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-DEV-bc157c7 (MC: 1.21.4)

### Other

I observed the javadoc and found no text describing this behavior, and I suspect it might be an unexpected behavior rather than a deliberate design.

Lulu13022002:
Original: https://github.com/PaperMC/Paper/pull/8243
Also deprecated the old HumanEntity#getPotentialRespawnLocation since it doesn't track the respawn angle.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
+++ b/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
@@ -295,8 +295,7 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
      * @return respawn location if exists, otherwise {@code null}.
      * @see #getRespawnLocation(boolean) for more fine-grained control over chunk loading and validation behaviour.
      */
-    @Nullable
-    default Location getRespawnLocation() {
+    default @Nullable Location getRespawnLocation() {
         return this.getRespawnLocation(false); // keep old behavior for offline players
     }
 
@@ -309,8 +308,7 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
      *                                will induce a sync chunk load and must hence be used with caution.
      * @return respawn location if exists, otherwise {@code null}.
      */
-    @Nullable
-    Location getRespawnLocation(boolean loadLocationAndValidate);
+    @Nullable Location getRespawnLocation(boolean loadLocationAndValidate);
 
     /**
      * Increments the given statistic for this player.
--- a/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
+++ b/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
@@ -287,13 +287,28 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
     // Paper end
 
     /**
-     * Gets the Location where the player will spawn at, null if they
+     * Gets the Location where the player will spawn at, {@code null} if they
      * don't have a valid respawn point.
+     * <br>
+     * Unlike online players, the location if found will not be loaded by default.
      *
-     * @return respawn location if exists, otherwise null.
+     * @return respawn location if exists, otherwise {@code null}.
      */
     @Nullable
-    public Location getRespawnLocation();
+    default Location getRespawnLocation() {
+        return this.getRespawnLocation(false); // keep old behavior for offline players
+    }
+
+    /**
+     * Gets the Location where the player will spawn at, {@code null} if they
+     * don't have a valid respawn point.
+     *
+     * @param load load the current location to retrieve the exact position of the spawn block
+     *             and check if this position is still valid or not
+     * @return respawn location if exists, otherwise {@code null}.
+     */
+    @Nullable
+    Location getRespawnLocation(boolean load);
 
     /**
      * Increments the given statistic for this player.
--- a/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
+++ b/paper-api/src/main/java/org/bukkit/OfflinePlayer.java
@@ -293,6 +293,7 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
      * Unlike online players, the location if found will not be loaded by default.
      *
      * @return respawn location if exists, otherwise {@code null}.
+     * @see #getRespawnLocation(boolean) for more fine-grained control over chunk loading and validation behaviour.
      */
     @Nullable
     default Location getRespawnLocation() {
@@ -303,12 +304,13 @@ public interface OfflinePlayer extends ServerOperator, AnimalTamer, Configuratio
      * Gets the Location where the player will spawn at, {@code null} if they
      * don't have a valid respawn point.
      *
-     * @param load load the current location to retrieve the exact position of the spawn block
-     *             and check if this position is still valid or not
+     * @param loadLocationAndValidate load the expected respawn location to retrieve the exact position of the spawn
+     *                                block and check if this position is still valid or not. Loading the location
+     *                                will induce a sync chunk load and must hence be used with caution.
      * @return respawn location if exists, otherwise {@code null}.
      */
     @Nullable
-    Location getRespawnLocation(boolean load);
+    Location getRespawnLocation(boolean loadLocationAndValidate);
 
     /**
      * Increments the given statistic for this player.
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -545,6 +545,12 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      */
     public boolean isSleepingIgnored();
 
+    @Nullable
+    @Override
+    default Location getRespawnLocation() {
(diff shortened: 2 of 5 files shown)


paper-api/src/main/java/org/bukkit/OfflinePlayer.java
```diff
@@ -290,13 +291,26 @@ public default BanEntry banPlayer(@Nullable String reason, @Nullable java.util.D
...
+     * Gets the Location where the player will spawn at, {@code null} if they
      * don't have a valid respawn point.
      *
-     * @return respawn location if exists, otherwise null.
+     * @param load load the current location to retrieve the exact position of the spawn block
+     *             and check if this position is still valid or not
+     * @return respawn location if exists, otherwise {@code null}.
      */
     @Nullable
-    public Location getRespawnLocation();
```
Strokkur424, line 296: I am wondering, what really is the rule for interfaces and explicit `public` modifiers? I have seen files/PRs with and I have seen files/PRs without.

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -560,15 +560,11 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
     @Deprecated(since = "1.20.4")
     public Location getBedSpawnLocation();
 
-    /**
-     * Gets the Location where the player will spawn at, null if they
-     * don't have a valid respawn point.
-     *
-     * @return respawn location if exists, otherwise null.
-     */
```
Strokkur424: The Javadocs have been removed on this method. Is this intentional, or are you going to be adding different Javadocs soon?

456dev: iirc it should now inherit from offline player?

paper-api/src/main/java/org/bukkit/OfflinePlayer.java
FireInstall: As far as I know, it was an unnecessary rule enforced by spigot.

paper-api/src/main/java/org/bukkit/entity/Player.java
electronicboy: I don't think JDs are inherited by default?

electronicboy: oh, no, nvm, they are, ineritDoc is for additive

Lulu13022002:
This also fix Player#getRespawnLocation that consume the respawn anchor in the nether if binded

paper-api/src/main/java/org/bukkit/OfflinePlayer.java
```diff
@@ -290,13 +291,26 @@ public default BanEntry banPlayer(@Nullable String reason, @Nullable java.util.D
...
-     * Gets the Location where the player will spawn at, null if they
+     * Gets the Location where the player will spawn at, {@code null} if they
+     * don't have a valid respawn point.
+     *
+     * @return respawn location if exists, otherwise {@code null}.
+     */
+    @Nullable
+    default Location getRespawnLocation() {
+        return this.getRespawnLocation(false); // keep old behavior for offline players
+    }
```
SoSeDiK: Thoughts on deprecating this method (without overload) altogether?
`Player` extends `OfflinePlayer`, and both have different default chunk loading behavior, which is quite confusing (also would be nice to specify the default behavior in respective javadocs).

Lulu13022002: I have updated the javadoc, not sure for the deprecation, i kinda expect offline player data to be less accurate than online by default. What i think is confusing is the fact that an online player extends an offline player in the api.

SoSeDiK: > What i think is confusing is the fact that an online player extends an offline player in the api.

Yes, that was my concern. Technically, you can have `OfflinePlayer` object representing an online `Player` instance (`#getOfflinePlayer` too returns online player if present), so you may call the method on an offline player expecting not to load chunks, but they will load due to the object being an online player.

New devs might stumble upon this quirk and will have to account for it. Deprecation would require to explicitly specify the wanted behavior, but I don't have a strong feeling against leaving this as is.

paper-api/src/main/java/org/bukkit/OfflinePlayer.java
```diff
@@ -290,13 +291,28 @@ public default BanEntry banPlayer(@Nullable String reason, @Nullable java.util.D
...
      * don't have a valid respawn point.
      *
-     * @return respawn location if exists, otherwise null.
+     * @param load load the current location to retrieve the exact position of the spawn block
+     *             and check if this position is still valid or not
+     * @return respawn location if exists, otherwise {@code null}.
      */
     @Nullable
-    public Location getRespawnLocation();
+    Location getRespawnLocation(boolean load);
```
lynxplay: Not a fan of the param naming here.
loadAndValidate?

lynxplay (approved):
LGTM beyond the param name.

JRoy:
This PR also fixes an issue where `getRespawnLocation` consumed a respawn anchor charge when it was called.

Machine-Maker (approved):

FellowNeighbour:
Waiting for this to be merged into upstream so respawn anchors work properly with EssX

NonSwag:
@notTamion just hinted me to this PR 
getPotentialRespawnPosition is in my opinion a misleading name since it returns the respawn point block location and not the location the player would _potentially_ respawn at
I opened a PR that introduces `getRespawnBlock` which is a more fitting name (probably going with `getRespawnBlockLocation` though)

masmc05:
`getRespawnBlock` is a misleading name as the respawn location may not be linked to a block and may contain angle data, as you could define with `spawnpoint` command

Leguan16:
Maybe `getDefinedRespawnLocation` and `getActualRespawnLocation` with the later doing the lookup.

paper-api/src/main/java/org/bukkit/OfflinePlayer.java
```diff
@@ -290,13 +291,28 @@ public default BanEntry banPlayer(@Nullable String reason, @Nullable java.util.D
     // Paper end
 
     /**
-     * Gets the Location where the player will spawn at, null if they
+     * Gets the Location where the player will spawn at, {@code null} if they
+     * don't have a valid respawn point.
+     * <br>
+     * Unlike online players, the location if found will not be loaded by default.
```
Warriorrrr, line 293: This note will be inherited by the player class which may seem confusing, we may want to have some javadocs in the player class override for this method still.

NoahvdAa (approved):

PR #11963, opened 2025-01-12
BlockProperty API

Machine-Maker:
Replaces https://github.com/PaperMC/Paper/pull/7223

----

My take on how to add a "get block data properties individually" on top of bukkit's existing method-based system.

### Structure
3 types of data properties: `Boolean`, `Integer`, and `Enum`. This pr creates a base `BlockProperty` class with 3 (+ special cases) implementations of it for each type.

#### Special Cases
So far, I've noticed 2 API types that are used to wrap integers, so this also adds handling for those, not exposing them as integers.
* `rotation` as BlockFace
* `note` as Note

The `BlockPropertyHolder` interface contains all the methods to interact with a set of properties. I left it as a standalone interface in case other uses in the future popup where a separate impl might be warranted. I added a `Mutable` subinterface to that to separate out the "set" methods, some implementations of it might not want to support mutating properties (such as future FluidState API https://github.com/PaperMC/Paper/pull/8609).

### DIFF
--- a/.checkstyle/checkstyle_base.xml
+++ b/.checkstyle/checkstyle_base.xml
@@ -55,7 +55,9 @@
         <module name="AvoidNoArgumentSuperConstructorCall"/>
         <module name="ConstructorsDeclarationGrouping"/>
         <module name="CovariantEquals"/>
-        <module name="DeclarationOrder"/>
+        <module name="DeclarationOrder">
+            <property name="ignoreConstructors" value="true"/>
+        </module>
         <module name="DefaultComesLast"/>
         <module name="EmptyStatement"/>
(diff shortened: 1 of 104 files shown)


Doc94:
related? https://github.com/PaperMC/Paper/blob/404167841dcf3ac78792ffbd45c683a2173b4b08/paper-server/src/main/java/io/papermc/paper/datacomponent/item/PaperBlockItemDataProperties.java#L40

Machine-Maker:
Rebased for 1.21.6

Machine-Maker:
@Lulu13022002 probably should figure out how to generate this stuff

PR #11970, opened 2025-01-14
Remove redundant public modifiers and remove redurtant static key word in Holder in paper-server/src/main/java/ca/sp…

dkks112313:


lynxplay:
The file comes from https://github.com/Tuinity/Moonrise.
Changes would be PRed there.

PR #11974, opened 2025-01-14
Fix zombie villager infection chance hunk

### LINKED ISSUE #11973: zombie-villager-infection-chance is broken
### Expected behavior

When `zombie-villager-infection-chance` is set to 100.0, it should be impossible for a zombie to kill a villager.

### Observed/Actual behavior

A villager was killed by a zombie.

### Steps/models to reproduce

Set `zombie-villager-infection-chance` is set to 100.0 on a difficulty other than hard (easy or normal), spawn a villager and a zombie a few times, and in some cases the villager will be killed instead of being infected.

### Plugin and Datapack List

```
> plugins
[15:49:17 INFO]: Server Plugins (2):
[15:49:17 INFO]: Bukkit Plugins:
[15:49:17 INFO]:  - floodgate, Geyser-Spigot

> datapack list
[15:49:14 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[15:49:14 INFO]: There are no more data packs available
```

### Paper version

```
> version
[15:49:07 INFO]: This server is running Paper version 1.21.4-106-main@ab1b312 (2025-01-12T23:38:01Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are 4 version(s) behind
Previous version: 1.21.4-78-49d15f6 (MC: 1.21.4)
```

(While this is not the latest version, the last few commits do not change anything related to villager zombification.)

### Other

This bug seems to have been introduced in commit ea5c599, after which the following lines are no longer removed:
```java
if (level.getDifficulty() != Difficulty.HARD && this.random.nextBoolean()) {
    return flag;
}
```

That means that on a difficulty other than hard for the successful zombification both checks must pass - the one introduced by Paper, as well as the vanilla 50% chance of the normal difficulty. That means the effective chance is actually divided by two, and cannot be higher than 50% (except for the hard difficulty).

lynxplay:
Dropped during hardfork, now moved to a false-ed out if statement to
prevent dropping in future updates.

Resolves: #11973

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/monster/Zombie.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/monster/Zombie.java.patch
@@ -191,7 +191,7 @@
      }
  
      @Override
-@@ -423,12 +_,18 @@
+@@ -423,13 +_,19 @@
          if (compound.contains("DrownedConversionTime", 99) && compound.getInt("DrownedConversionTime") > -1) {
              this.startUnderWaterConversion(compound.getInt("DrownedConversionTime"));
          }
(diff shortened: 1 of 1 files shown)


lynxplay:
https://github.com/PaperMC/Paper-archive/blob/bb4cc9bd03d959f97706da7fdf9738ba30a046ff/patches/server/0317-Configurable-chance-of-villager-zombie-infection.patch#L19-L21 for reference

electronicboy (approved):

PR #11990, opened 2025-01-19
Disable the /reload command by default.

Mickey42302:
(Reposting this to another branch as stated by the github-actions bot.)

I would like to suggest adding a patch which disables the "/reload" command by default. All credit is given to the DivineMC project for this idea.

As stated in the output for the "/reload" command, it is known to cause memory leaks and other issues. Disabling the command by default would encourage users to restart their server instead of reloading it.


electronicboy:
This has been discussed many times over the years, and generally, the command is already gated by default, and people are well aware of the implications or just don't care. The diff here is also pointlessly noisy for what it's doing, and would generally not be a feature patch for a thing we add.

PR #11991, opened 2025-01-19
Implement again save for position of last lava contact

### LINKED ISSUE #11984: EntityDamageByBlockEvent / EntityCombustByBlockEvent returning null block for lava
### Expected behavior

Up to version 1.21.1, EntityDamageByBlockEvent.getDamager() returns the block responsible for the damage dealt as a valid Block object for damage caused by lava. I therefore expected it to work that way in newer versions too, as it's most certainly the expected behavior of this method in this case.

### Observed/Actual behavior

On 1.21.3+, EntityDamageByBlockEvent.getDamager() just returns null for damage caused by lava.

### Steps/models to reproduce

Here's what I did:

- Create 2 localhost servers, one with PaperMC 1.21.1 and the other one with PaperMC 1.21.3 (newest builds).
*(no changes to those servers except accepting the eula)*

- Create a minimalistic plugin to log EntityDamageByBlockEvent.getDamager().

- Join both servers, grab a bucket of lava and take a bath.

- Compare console output:
**1.21.1:** `CraftBlock{...}`
**1.21.3:** `null`

### Plugin and Datapack List

Just the minimalistic logging plugin:

```
public final class GetDamagerTest extends JavaPlugin implements Listener {

    @Override
    public void onEnable() {
    }

    @Override
    public void onDisable() {}

    @EventHandler
    public void onDamageByBlock(EntityDamageByBlockEvent event) {
        Bukkit.getLogger().info(event.getDamager());
    }
}
```

### Paper version

**1.21.1#132** & **1.21.3#82**

### Other

_No response_

Doc94:
This PR resolve #11984 by adding again set the position of last lava contact, also add the Nullable annotation to that field.

### DIFF
--- a/paper-server/patches/features/0003-Entity-Activation-Range-2.0.patch
+++ b/paper-server/patches/features/0003-Entity-Activation-Range-2.0.patch
@@ -366,7 +366,7 @@ index d95413af04121fe91ca0f3b0c70025b9808acef9..ad665c7535c615d2b03a3e7864be435f
  import org.slf4j.Logger;
  
 diff --git a/net/minecraft/server/level/ServerLevel.java b/net/minecraft/server/level/ServerLevel.java
-index 3164f3784131babf9a6663335517a12df7e88a7b..da8848e2a3e3745949eb2356a049451d30f763a7 100644
+index 79d23c3403efc6dbef2381a3fa5946023f862452..3b19229427d83290bba1431bee5357e2ced34f94 100644
 --- a/net/minecraft/server/level/ServerLevel.java
 +++ b/net/minecraft/server/level/ServerLevel.java
 @@ -551,6 +551,7 @@ public class ServerLevel extends Level implements ServerEntityGetter, WorldGenLe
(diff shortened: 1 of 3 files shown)


Doc94:
now.. im not good with the large feature touchs... im not sure if this is the better way for handle the add again that CB patch.. or if is better fix that in the source file and the moonrise patch just need avoid deleting that...

electronicboy:
The issue here is that we don't really want to be implementing this stuff into moonrise directly, as it will just be removed on the next moonrise update

Doc94:
> The issue here is that we don't really want to be implementing this stuff into moonrise directly, as it will just be removed on the next moonrise update

then the fix its just modify the file and handle the merge conflict or exists a better way?

electronicboy:
cept there will be no merge conflict due to how this stuff is applied, we need MR to add a hook or something so that the changes don't get wiped out on every update

Doc94:
Ok i remake this... just copy the part of `updateFluidHeightAndDoFluidPushing` where get the last BlockPos for the fluid... and use that for update the BlockPos of lastLavaContact this require keep update that new method but avoid mess with the method touched by moonrise.

Owen1212055 (changes requested):
In general, I do not think this is the correct solution.

We will need to work with moonrise to add a hook into that method, which then through `PaperHooks` we could add the field setting there. I have notified them internally.

lynxplay:
Fixed in 1.21.5 via mojangs InsideBlockEffectApplier

PR #11994, opened 2025-01-20
Consider off-hand when adding items to player inventory

### LINKED ISSUE #11921: addItem problem with offhand and full inventory
### Expected behavior

Hello.

The following method when there is a space to add the item should add the item in the inventory:

 PlayerInventory inv = p.getInventory();
inv.addItem(new ItemStack(Material.DIAMOND));

Image

On my screenshot, I have space in my offhand, so it should increase to 37, 38, 39 up to 64.

### Observed/Actual behavior

But currently, it doesnt work when the space is in the off hand. 
The additem method return a non empty Map.

### Steps/models to reproduce

Create a small plugin with

 PlayerInventory inv = p.getInventory();
inv.addItem(new ItemStack(Material.DIAMOND));

when you do a specific action

put at least one of the item you add in your offhand and fullfill your inventory with another item

### Plugin and Datapack List

[21:25:21 INFO]:  - AceStore, CustomPiglinsTrades, ExecutableBlocks, ExecutableEvents, ExecutableItems, LuckPerms, MyFurniture, NBTAPI, packetevents, PlaceholderAPI
[21:25:21 INFO]:  SCore, Vault, WorldEdit

### Paper version

[21:25:29 INFO]: This server is running Paper version 1.21.4-76-main@486e521 (2025-01-05T18:15:54Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
>

### Other

_No response_

niho2:
Fixes #11921

## Problem
The `addItem` method in `CraftInventory` doesn't consider the off-hand when adding items to a full player inventory. When an item is partially in the off-hand and the rest of the inventory is full, it incorrectly indicates that there's no space available, even though there might be space in the off-hand.

## Changes
- Extended the `addItem` method to check if it's a player inventory
- Added additional check for the off-hand when no partial stack is found in the main inventory
- Attempts to stack items in the off-hand before looking for an empty slot
- Maintains original logic for non-player inventories

## Test Cases
1. Player has an item in off-hand with free stack space
2. Player attempts to add the same item
3. Item is correctly added to the off-hand
4. Remaining items (if any) are placed normally in the inventory

## Additional Notes
- Change respects the maximum stack size of the item
- No API surface changes
- Backwards compatible with existing functionality

### DIFF
--- a/paper-api/src/main/java/co/aikar/timings/TimingsManager.java
+++ b/paper-api/src/main/java/co/aikar/timings/TimingsManager.java
@@ -45,8 +45,7 @@ import org.jetbrains.annotations.Nullable;
 @Deprecated(forRemoval = true)
 public final class TimingsManager {
     static final Map<TimingIdentifier, TimingHandler> TIMING_MAP = LoadingMap.of(
-        new ConcurrentHashMap<>(4096, .5F), TimingHandler::new
-    );
+            new ConcurrentHashMap<>(4096, .5F), TimingHandler::new);
     public static final FullServerTickHandler FULL_SERVER_TICK = new FullServerTickHandler();
     public static final TimingHandler TIMINGS_TICK = Timings.ofSafe("Timings Tick", FULL_SERVER_TICK);
     public static final Timing PLUGIN_GROUP_HANDLER = Timings.ofSafe("Plugins");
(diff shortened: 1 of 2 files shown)


niho2:
~Fixes #11921~ (Closing as won't fix)

After review, it was determined that the proposed solution would break the API contract specified in the Inventory.getStorageContents() javadocs. The off-hand slot is intentionally excluded from storage contents, and addItem() should only work with storage contents as specified.

Instead of modifying addItem(), we should consider adding a new method addItemAnySlot() (similar to the existing removeItemAnySlot()).

PR #11995, opened 2025-01-21
Add Player#give

### LINKED ISSUE #11921: addItem problem with offhand and full inventory

masmc05:
Closes https://github.com/PaperMC/Paper/issues/11921

Adds a new method that follows full vanilla inventory logic, like first trying to put in main hand, then in off hand, then in the rest inventory. It's also a utility method that supports dropping items if they did not fit to ensure that items are not lost.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -6,10 +6,12 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Date;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import io.papermc.paper.entity.LookAnchor;
+import io.papermc.paper.entity.PlayerGiveResult;
 import org.bukkit.BanEntry;
 import org.bukkit.DyeColor;
 import org.bukkit.Effect;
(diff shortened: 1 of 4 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -3541,4 +3543,35 @@ public void sendEntityEffect(final org.bukkit.EntityEffect effect, final org.buk
...
+    }
+
+    @Override
+    public @NotNull List<@NotNull Item> give(@NotNull final Collection<@NotNull ItemStack> items, final boolean dropIfFull) {
+        Preconditions.checkNotNull(items, "items cannot be null");
+        ImmutableList.Builder<Item> droppedItems = ImmutableList.builder();
+        for (ItemStack item : items) {
+            Preconditions.checkNotNull(item, "ItemStack cannot be null");
+            Preconditions.checkArgument(!item.isEmpty(), "ItemStack cannot be empty");
+            Preconditions.checkArgument(item.getAmount() <= item.getMaxStackSize(), "ItemStack amount cannot be greater than its max stack size");
```
Doc94: Why? or this mess with the add logic?
if so then the docs can add the throws for this cases.

masmc05: Vanilla doesn't support overstacked items, since we're giving it to vanilla, we can't risk with bugs from possible side effects

masmc05: And don't think mentioning in the docs is necessary, as many usages (like the new dropItem method) break on overstacked items, they are more supported only when mentioned as supported

Doc94:
this not require the update container like the give command?

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3892,4 +3893,38 @@ default boolean isChunkSent(@NotNull org.bukkit.Chunk chunk) {
...
+    /**
+     * Gives the player the items following full vanilla logic,
+     * making the player drop the items that did not fit into
+     * the inventory. This method will update the count of items
+     * to the amount that did not fit in the inventory.
+     *
+     * @param items the items to give, their count will be mutated
+     * @return spawned item entities
+     */
+    @NotNull List<@NotNull Item> give(@NotNull ItemStack @NotNull... items);
```
lynxplay: Just a note, this might wanna be a ItemStack first, ItemStack... other, but we decided against that because itwould make interop with getContents() methods that yield ItemStack[] a lot worse so, may people call this without any items I guess xD

lynxplay:
To note for tomorrows review/merge, leftovers are scuffed in creative as the method obviously doesn't drop for creative mode players.
This, I'd argue, is perfectly fine tho given this is supposed to be exactly like vanilla.

Machine-Maker (approved):

PR #12002, opened 2025-01-22
Fix a rare crash with a concurrent modification of a set

AverageGithub:
This PR attempts to fix an issue introduced in CraftBukkit 5 years ago in https://github.com/PaperMC/Paper/commit/fcfcbd39757f71150e2ba3845cbaeb687bc01a48 (how did noone catch this??)

Relevant part of a crash report: https://pastes.dev/cKg6BswffG (1.21.3 server - but it does not matter here)

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -2858,9 +2858,11 @@ public void injectScaledMaxHealth(Collection<AttributeInstance> collection, bool
         if (!this.scaledHealth && !force) {
             return;
         }
-        for (AttributeInstance genericInstance : collection) {
+        java.util.Iterator<AttributeInstance> iterator = collection.iterator();
```
lynxplay: We can use an import here :+1:

lynxplay (approved):
Beyond the import, LGTM.

emilyy-dev (approved):

AverageGithub:
I did not add an import because an other fully qualified import already exists in the file and I didn't want to touch that. Should I remove that aswell?

lynxplay:
Just leave the other FQN alone, we will be doing a cleanup of these files soonish once we have checkstyle ready on our repos.

But new code can already follow that :+1:

electronicboy (approved):

PR #12007, opened 2025-01-25
Fix wrong piston world border check

### LINKED ISSUE #11983: Pushing a block outside the world border creates a glitchy moving block
### Expected behavior

The piston should be able to push the block through the world border, or the block should be stopped as if it was pushed against an immovable block.

### Observed/Actual behavior

After a block is pushed through the world border it turns into a glitchy moving block. I am unable to interact with the block and after reconnecting the block turns invisible but I can still collide with it.

If you increase the world border size the block will be able to move.

> https://youtu.be/aEOddMXy0OA

### Steps/models to reproduce

Push a block through the world border using a normal piston

### Plugin and Datapack List

n/a

### Paper version

This server is running Paper version 1.21.4-114-main@cd9d6d6 (2025-01-15T11:21:28Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are 2 version(s) behind

### Other

_No response_

HaHaWTH:
Closes #11983.

This pull request fixes vanilla wrong piston world border check([MC-82010](https://bugs.mojang.com/browse/MC-82010), thanks to FX-PR0CESS)

Before(Paper 1.21.4 build 121):

before

After:

after

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/piston/PistonBaseBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/piston/PistonBaseBlock.java.patch
@@ -66,7 +66,7 @@
  
      public static boolean isPushable(BlockState state, Level level, BlockPos pos, Direction movementDirection, boolean allowDestroy, Direction pistonFacing) {
 -        if (pos.getY() < level.getMinY() || pos.getY() > level.getMaxY() || !level.getWorldBorder().isWithinBounds(pos)) {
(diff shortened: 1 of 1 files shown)


lynxplay:
Given this isn't a full fix for the linked mojira issue, I'll update the comment.
We should also keep the old check as well as otherwise pistons outside the world border now start pushing things into, being stuck with the same problem.

Machine-Maker (approved):

lynxplay (approved):

lynxplay:
LGTM, thanks!
For any readers of this PR down the line, this is not a fix for the linked MC issue but a band aid fix specifically for the piston attempting to push blocks out of the world border :+1:

PR #12011, opened 2025-01-25
Add EntityEquipmentChangedEvent

TonytheMacaroni:
Replaces #11667.

Opted to implement the alternative mentioned in the previous PR, which instead exposes of changed slots -> equipment changes. Feedback on naming and the approach would be appreciated.

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
@@ -8,6 +8,7 @@ import org.bukkit.event.entity.EntityEvent;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.Unmodifiable;
 import org.jspecify.annotations.NullMarked;
 
@@ -37,12 +38,12 @@ public class EntityEquipmentChangedEvent extends EntityEvent {
     public EntityEquipmentChangedEvent(final LivingEntity entity, final Map<EquipmentSlot, EquipmentChange> equipmentChanges) {
         super(entity);
 
-        this.equipmentChanges = Collections.unmodifiableMap(equipmentChanges);
+        this.equipmentChanges = equipmentChanges;
     }
 
     @Override
     public LivingEntity getEntity() {
-        return (LivingEntity) entity;
+        return (LivingEntity) this.entity;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
```diff
@@ -0,0 +1,66 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called whenever a change to an entity's equipment has been detected. This event is called after effects from
+ * attribute modifiers and enchantments have been updated.
+ */
+@NullMarked
+public class EntityEquipmentChangedEvent extends EntityEvent {
```
Warriorrrr: It might be worth documenting here that this event is also called upon entities being added to worlds, since it isn't immediately obvious to users of the event that the game counts that as "changed".

Warriorrrr (commented):
The naming and approach of this PR seems good to me

paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
TonytheMacaroni: Added some examples of situations that can call the event.

Warriorrrr (approved):

paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
```diff
@@ -0,0 +1,77 @@
...
+        return HANDLER_LIST;
+    }
+
+    /**
+     * Represents a change in equipment for a single equipment slot.
+     *
+     * @param oldItem the existing item that is being replaced
+     * @param newItem the new item that is replacing the existing item
+     */
+    public record EquipmentChange(ItemStack oldItem, ItemStack newItem) {
```
lynxplay: This should probably be another interface + package private record impl combo.
We also should be defensively copying the stacks on access to prevent plugins from changing these accidentally.

paper-api/src/main/java/io/papermc/paper/event/entity/EntityEquipmentChangedEvent.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Map<EquipmentSlot, EquipmentChange> equipmentChanges;
+
+    @ApiStatus.Internal
+    public EntityEquipmentChangedEvent(final LivingEntity entity, final Map<EquipmentSlot, EquipmentChange> equipmentChanges) {
+        super(entity);
+
+        this.equipmentChanges = Collections.unmodifiableMap(equipmentChanges);
```
lynxplay: Just move the "unmodifiable" wrapping into the getter instead.
It is cheap enough to do there and just ensures we don't accidentally move it later.

lynxplay (changes requested):

lynxplay (approved):

PR #12013, opened 2025-01-26
Allow For Default Titles in InventoryView Builders

Y2Kwastaken:
This PR addresses a previous critique of no default titles for the InventoryView builders PR. The goal of this PR is to integrate default titles and minimize the explicit declaration of translatable keys to ease on maintenance burden of this API.

I have bundled the test plugin I used along with this PR to ease of testing code if needed.

Below examplifies the change from an API perspective
```java
// Previous Disallowed
MenuType.CHEST.builder()
 .title(null); // the title can be null now
 .build(player);
``` 
```java
// Now Permitted
MenuType.CHEST.builder()
 .build(player);
```

Within this PR I took a few different approaches to accessing titles. The first two of, which are off of the BlockEntity and Block menu provider respectively. However, there are two cases where I directly declare a Translatable component within this API, once in CraftStandardInventoryViewBuilder and once in CraftDoubleChestInventoryViewBuilder.

- CraftStandardInventoryViewBuilder: Within this class I directly referenced the "container.chest" translatable key because, first, it is not necessary to create a BlockEntity for these "standard" menus as well as the fact that the possibility that the "container.chest" translation key will change is slim to none.
- CraftDoubleChestInventoryViewBuilder: This key was explicitly declared simply because its not easily accessible and writing around this lack of accessibility leads to less than cohesive implementation. This translation key is not as accessible because it is declared inside of ChestBlock#MENU_PROVIDER_COMBINER, while this field itself is accessible, the MenuProvider DoubleChest uses is nested with the "acceptDouble(ChestBlockEntity, ChestBlockEntity" method, which then nests its MenuProvider in the then Bukkit wrapper called "DoubleInventory", doing this for both location and non location backed containers, was in my opinion not beneficial enough to warrant the implementation.

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/util/CraftMenus.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/inventory/util/CraftMenus.java
@@ -17,6 +17,7 @@ import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.level.block.entity.BeaconBlockEntity;
 import net.minecraft.world.level.block.entity.BlastFurnaceBlockEntity;
 import net.minecraft.world.level.block.entity.BrewingStandBlockEntity;
+import net.minecraft.world.level.block.entity.ChestBlockEntity;
 import net.minecraft.world.level.block.entity.CrafterBlockEntity;
 import net.minecraft.world.level.block.entity.DispenserBlockEntity;
 import net.minecraft.world.level.block.entity.FurnaceBlockEntity;
@@ -87,7 +88,7 @@ public final class CraftMenus {
             return asType(new MenuTypeData<>(InventoryView.class, () -> new CraftDoubleChestInventoryViewBuilder<>(handle)));
         }
         if (menuType == MenuType.GENERIC_9X3) {
-            return asType(new MenuTypeData<>(InventoryView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.CHEST, null)));
+            return asType(new MenuTypeData<>(InventoryView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.CHEST, ChestBlockEntity::new, false)));
         }
         // this isn't ideal as both dispenser and dropper are 3x3, InventoryType can't currently handle generic 3x3s with size 9
         // this needs to be removed when inventory creation is overhauled
@@ -98,7 +99,7 @@ public final class CraftMenus {
             return asType(new MenuTypeData<>(CrafterView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.CRAFTER, CrafterBlockEntity::new)));
         }
         if (menuType == MenuType.ANVIL) {
-            return asType(new MenuTypeData<>(AnvilView.class, () -> new CraftAccessLocationInventoryViewBuilder<>(handle, AnvilMenu::new)));
+            return asType(new MenuTypeData<>(AnvilView.class, () -> new CraftAccessLocationInventoryViewBuilder<>(handle, Blocks.ANVIL)));
         }
         if (menuType == MenuType.BEACON) {
             return asType(new MenuTypeData<>(BeaconView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.BEACON, BeaconBlockEntity::new)));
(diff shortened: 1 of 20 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftDoubleChestInventoryViewBuilder.java
```diff
@@ -15,6 +16,7 @@ public class CraftDoubleChestInventoryViewBuilder<V extends InventoryView> exten
 
     public CraftDoubleChestInventoryViewBuilder(final MenuType<?> handle) {
         super(handle);
+        super.defaultTitle = Component.translatable("container.chestDouble");
```
lynxplay, line 19: Both these should get a `// Paper - diff on change - CraftDoubleChestInventoryViewBuilder.defaultTitle` on the respective bits in NMS that "define" them / use them

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftStandardInventoryViewBuilder.java
```diff
@@ -10,6 +11,7 @@ public class CraftStandardInventoryViewBuilder<V extends InventoryView> extends
 
     public CraftStandardInventoryViewBuilder(final MenuType<?> handle) {
         super(handle);
+        super.defaultTitle = Component.translatable("container.chest");
```
lynxplay, line 14: See other, diff on change.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -44,23 +51,36 @@ protected AbstractContainerMenu buildContainer(final ServerPlayer player) {
...
+
+        if (!this.forceTile) { // gets around open noise for chest
+            return handle.create(player.nextContainerCounter(), player.getInventory());
         }
+
         return inventory.createMenu(player.nextContainerCounter(), player.getInventory(), player);
     }
 
+    private void trySetDefaultTitle(BlockEntity blockEntity) {
+        if (blockEntity instanceof Nameable nameable) {
```
lynxplay: Early return instead.
`!(obj instanceOf final Type type)` works.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -44,23 +51,36 @@ protected AbstractContainerMenu buildContainer(final ServerPlayer player) {
...
-        if (this.builder == null) {
-            return handle.create(player.nextContainerCounter(), player.getInventory());
-        }
         final MenuProvider inventory = this.builder.build(this.position, this.block.defaultBlockState());
         if (inventory instanceof final BlockEntity tile) {
             tile.setLevel(this.world);
+            trySetDefaultTitle(tile);
+        }
+
+        if (!this.forceTile) { // gets around open noise for chest
```
lynxplay: We can get a better name here I am sure.
Mostly because I guess all the other "tile" references in the code here were not caught in the initial review xD
`forceBlockEntityUsage`, `useBlockEntityMenuCreation`, whatever.
If you are at it, there are a couple of other `tile` names around, might be a vibe to update those to `blockEntity` too :)

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -44,23 +51,36 @@ protected AbstractContainerMenu buildContainer(final ServerPlayer player) {
             return buildFakeTile(player);
         }
 
+        trySetDefaultTitle(entity);
```
lynxplay: Less of a `try`, this will error if it doesn't succeed and that error isn't an expected state unless we fucked up.
So just, `setDefaultTitle`, or honestly `super.defaultTitle = extractDefaultTitle(entity)` is better here.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -44,23 +51,36 @@ protected AbstractContainerMenu buildContainer(final ServerPlayer player) {
             return buildFakeTile(player);
```
lynxplay: This is one of those "buildFakeBlockEntity" things that would make me happy xD

lynxplay (changes requested):

Lulu13022002 (commented):
Maybe MenuType#create could have a new method without title and a nullable title.
Also i get this error while testing the lectern: https://pastes.dev/KgmpW6mqCE

Lulu13022002 (commented):
The loom has the title of the chest set by default.

I also get some other errors while testing
for the enchantment menu: https://pastes.dev/ZgBP7TjW9Y
for the merchant menu: https://pastes.dev/whPBf9wNNc

Y2Kwastaken:
> The loom has the title of the chest set by default.
> 
> I also get some other errors while testing for the enchantment menu: [pastes.dev/ZgBP7TjW9Y](https://pastes.dev/ZgBP7TjW9Y) for the merchant menu: [pastes.dev/whPBf9wNNc](https://pastes.dev/whPBf9wNNc)

Hey can you sure the code you're using when you get the second exception, thanks.

Lulu13022002:
I get another error when i tried to open a merchant menu without villager but with a title set: https://pastes.dev/hLGHRQQSSt

And the loom is still disguised as a chest.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/view/builder/CraftBlockEntityInventoryViewBuilder.java
```diff
@@ -3,6 +3,7 @@
 import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.MenuProvider;
+import net.minecraft.world.Nameable;
```
Warriorrrr: This new import is not used in the file

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/util/CraftMenus.java
```diff
@@ -140,16 +142,16 @@ public static <V extends InventoryView, B extends InventoryViewBuilder<V>> MenuT
             return asType(new MenuTypeData<>(InventoryView.class, () -> new CraftBlockEntityInventoryViewBuilder<>(handle, Blocks.SHULKER_BOX, ShulkerBoxBlockEntity::new)));
```
Warriorrrr, line 142: If opening noises should be prevented (as is done with chest using useFakeBlockEntity), then perhaps that could be done here too with the shulker.

Warriorrrr (commented):
Two small comments but otherwise all looks good to me

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/util/CraftMenus.java
Y2Kwastaken: Yeah The whole shulker thing is a bit more nuanced due to the need for it to be distinctly different than a chest with GENERIC_9x3. In the interest of trying to keep this PR focused on adding default titles, this will need to be addressed in a PR where I rewrite the opening logic that Bukkit does. Long story short getNotchInventoryType isn't actually smart enough to identify the difference between a shulker box and GENERIC_9x3 as they'd be created with this API. The way around this is to define and force the use of the BlockEntity for now* The chest is okay to avoid this because it is a GENERIC_9x3.

Warriorrrr (approved):

lynxplay (approved):

Machine-Maker (approved):

PR #12016, opened 2025-01-26
Make Sittable interface extend Entity

### LINKED ISSUE #12014: Sittable does not extend any entity class
### Expected behavior

As documented, `Sittable` applies to `Animals` and since all implementing classes of Sittable share the `Animals` class it should extend Animals

### Observed/Actual behavior

Sittable does not extend any entity class

### Steps/models to reproduce

-

### Plugin and Datapack List

-

### Paper version

This server is running Paper version 1.21.4-126-main@fb5b173 (2025-01-25T20:51:32Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-121-88bbead (MC: 1.21.4)

### Other

this makes it not possible to use sittable in generic contexts where you expect an instance of entity
Image

NonSwag:
closes #12014

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Panda.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Panda.java
@@ -172,22 +172,6 @@ public interface Panda extends Animals, Sittable {
     }
 
     /**
-     * Sets if this panda is currently sitting.
-     *
-     * @param sitting is currently sitting
-     */
-    @Override
-    void setSitting(boolean sitting);
-
-    /**
-     * Gets if this panda is sitting.
-     *
-     * @return is sitting
-     */
-    @Override
-    boolean isSitting();
-
-    /**
      * Gets this Panda's combined gene.
      * <p>
      * The combined gene can be modified using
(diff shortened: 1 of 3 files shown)


Machine-Maker:
I'm not sure this is smart. What happens when a non-animal sittable mob is added? Not really out of the realm of possibility, in fact its probably likely. I would rather Sittable be made to extend Entity or LivingEntity since that's pretty much guaranteed to always be the case.

Owen1212055:
Agree, this should not be on Animals but Entity.

NonSwag:
In that case I'd change the JD as well
They state an animal that may sit

Machine-Maker:
Yeah, that's fine. I don't really take a casual mention of "animal" to be an API contract that this type also implements `Animals` (btw, why is it plural? strange bukkit choices).

Owen1212055:
It’s been animals for ages, just an ancient relic.

NonSwag:
https://github.com/PaperMC/Paper/issues/12014#issuecomment-2616563524

Lulu13022002:
I think the overrides in Panda can be removed too, the javadoc is not more relevant than the parent one

Machine-Maker (approved):

kennytv (approved):

PR #12017, opened 2025-01-26
Adding PredicateChoice to Paper API (updated version)

CPieter:
This PR is an updated version of derverdox's PredicateChoice PR #9996. Since the branch in the old PR was quite outdated I reïmplemented the changes in a new branch, but most of the changes remain the same. Credit of the code should go to derverdox and Machine-Maker.
The feature seems to work fine from my testing but I'm unfamiliar with the Paper API so I please let me know if I missed something.


paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -290,4 +291,50 @@ public String toString() {
         }
         // Paper end - check valid ingredients
     }
+
+    // Paper start - add PredicateChoice
```
lynxplay: No need for paper comments in the API anymore.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -290,4 +291,50 @@ public String toString() {
...
+
+        /**
+         * @param predicate The predicate to test the crafting inputs to. Mutating
+         *                  the ItemStack in the predicate is not supported.
+         * @param exampleStack An example ItemStack to be shown in the recipe book.
+         */
+        public PredicateChoice(Predicate<ItemStack> predicate, ItemStack exampleStack) {
+            Preconditions.checkArgument(predicate != null, "The item predicate cannot be null");
+            Preconditions.checkArgument(exampleStack != null, "The example stack cannot be null");
+            Preconditions.checkArgument(!exampleStack.getType().isAir(), "Cannot have empty/air example stack");
```
lynxplay: Why not just !exampleStack.isEmpty() ?

paper-server/patches/sources/net/minecraft/world/item/crafting/Ingredient.java.patch
```diff
@@ -26,10 +27,15 @@
...
-@@ -60,6 +_,17 @@
+@@ -60,6 +_,22 @@
  
      @Override
      public boolean test(ItemStack stack) {
++        // Paper start - add PredicateChoice
++        if (this.stackPredicate != null) {
++            return this.stackPredicate.test(stack);
++        }
++        // Paper end
```
lynxplay: Missing ` - Add PredicateChoice`

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftRecipe.java
```diff
@@ -32,7 +35,14 @@ static Ingredient toIngredient(RecipeChoice bukkit, boolean requireNotEmpty) {
 
         if (bukkit == null) {
             stack = Ingredient.of();
-        } else if (bukkit instanceof RecipeChoice.MaterialChoice) {
+        }
+        // Paper start - add PredicateChoice
```
lynxplay: No more comments, can be properly formatted with the above closing if bracket.

lynxplay (dismissed):
Welcome to paper :tada: 
Thank you for updating the PR. I don'T have time for a indepth review rn, but I left some initial easy nitpicks around comments :+1: 

Will get to a review next week probably

Doc94:
Hi and welcome to the Paper PR world (well this more a thing can say a Paper Team Member xd)

The paper comments currently just apply to code in patches, its not required in the api,craftbukkit,paper classes.

CPieter:
I just noticed there is already a PredicateRecipeChoice for potions, with the addition of the PredicateChoice this could be a bit confusing. Maybe it should be renamed to something more clear like a PotionRecipeChoice/PotionPredicateChoice?

Doc94:
> I just noticed there is already a PredicateRecipeChoice for potions, with the addition of the PredicateChoice this could be a bit confusing. Maybe it should be renamed to something more clear like a PotionRecipeChoice/PotionPredicateChoice?

well rename sounds a little breaking i think, maybe this new PredicateChoice can be named ItemPredicateChoice (? or add an extension of RecipeChoice for Potion and move things to that for make clear the diff... but not really sure.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -290,4 +291,48 @@ public String toString() {
         }
         // Paper end - check valid ingredients
     }
+
+    /**
+     * Represents a choice that will be valid only if an item matches the
+     * given predicate
+     */
+    @NullMarked
+    public static class PredicateChoice implements RecipeChoice {
```
Machine-Maker: We'd like to avoid having classes/records or really any implementation types exposed to the API, preferring to just have an interface.

So make this an interface, and add a package-private implementation, in the api, like `PredicateChoiceImpl` that is a record. Then inside `RecipeChoice`, add a static method to create an instance of it.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -290,4 +291,48 @@ public String toString() {
...
+         *                  the ItemStack in the predicate is not supported.
+         * @param exampleStack An example ItemStack to be shown in the recipe book.
+         */
+        public PredicateChoice(Predicate<ItemStack> stackPredicate, ItemStack exampleStack) {
+            Preconditions.checkArgument(stackPredicate != null, "The item predicate cannot be null");
+            Preconditions.checkArgument(exampleStack != null, "The example stack cannot be null");
+            Preconditions.checkArgument(!exampleStack.isEmpty(), "Cannot have empty/air example stack");
+
+            this.stackPredicate = stackPredicate;
+            this.exampleStack = exampleStack;
```
Machine-Maker: In the constructor of the above-mentioned record, clone the itemstack input.

Machine-Maker (changes requested):

Machine-Maker:
> I just noticed there is already a PredicateRecipeChoice for potions, with the addition of the PredicateChoice this could be a bit confusing. Maybe it should be renamed to something more clear like a PotionRecipeChoice/PotionPredicateChoice?

Since PredicateRecipeChoice isn't public, delete it, and make `PotionMix#createPredicateChoice` return an instance of the new type. But also deprecate that method, and suggest to use the new method you create.

Here's a great example of what I talked about in my review, suggesting to keep implementation types out of the "api" because we can now just delete PredicateRecipeChoice entirely.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
CPieter: What's your opinion on also adding static constructors inside `RecipeChoice` for the `MaterialChoice` and `ExactChoice`? (for concistency)

Machine-Maker: Yeah, let’s do that. Then mark the constructors for those as deprecated. We’ll move those to interfaces too when we have the bytecode tools to support rewriting that.

Machine-Maker: Call the method “itemTypeChoice”.

CPieter: Is there even a point to having seperate interfaces? Maybe these methods could all just return a RecipeChoice.

Machine-Maker: might be too big of a change. There are specific public methods for each.

Actually just ignore MaterialChoice for now. I'm gonna open a separate PR handling that one, just do ExactChoice

paper-api/src/main/java/org/bukkit/inventory/PredicateChoiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+record PredicateChoiceImpl(Predicate<ItemStack> stackPredicate, ItemStack exampleStack) implements RecipeChoice.PredicateChoice {
+
+    public PredicateChoiceImpl(Predicate<ItemStack> stackPredicate, ItemStack exampleStack) {
+        Preconditions.checkArgument(stackPredicate != null, "The item predicate cannot be null");
+        Preconditions.checkArgument(exampleStack != null, "The example stack cannot be null");
+        Preconditions.checkArgument(!exampleStack.isEmpty(), "Cannot have empty/air example stack");
+
+        this.stackPredicate = stackPredicate;
+        this.exampleStack = exampleStack.clone();
+    }
```
Machine-Maker:
```suggestion
    public PredicateChoiceImpl {
        Preconditions.checkArgument(stackPredicate != null, "The item predicate cannot be null");
        Preconditions.checkArgument(exampleStack != null, "The example stack cannot be null");
        Preconditions.checkArgument(!exampleStack.isEmpty(), "Cannot have empty/air example stack");
        exampleStack = exampleStack.clone()
    }
```
can use a fancy record compact constructor here since its just preconditions

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -290,4 +343,14 @@ public String toString() {
...
     }
+
+    /**
+     * Represents a choice that will be valid only if an item matches the
+     * given predicate.
+     */
+    @NullMarked
+    sealed interface PredicateChoice extends RecipeChoice permits PredicateChoiceImpl {
+
+        Predicate<ItemStack> getPredicate();
```
Machine-Maker: Not sure we have to expose this, since it is a predicate itself. An empty interface is fine. To your point about maybe not needing interfaces, I think it's good to have even an empty one so you can at least know what type of choice it is. otherwise it's a little annoying to figure out.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftRecipe.java
Machine-Maker, line 1: You also have to handle the conversion in the other direction. When it converts from an nms ingredient to a bukkit RecipeChoice. Check if the predicate is set and create one there. It's just below the method you changed.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -79,10 +121,12 @@ public MaterialChoice(@NotNull Material... choices) {
          *
          * @param choices the tag
          */
+        @Deprecated
         public MaterialChoice(@NotNull Tag<Material> choices) {
             this(new ArrayList<>(java.util.Objects.requireNonNull(choices, "Cannot create a material choice with null tag").getValues())); // Paper - delegate to list ctor to make sure all checks are called
         }
 
+        @Deprecated
```
Machine-Maker: Lets not deprecate since we don't have the alternative in, and Material isn't deprecated yet.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -192,14 +236,23 @@ public static class ExactChoice implements RecipeChoice {
...
+         * @deprecated Use {@link RecipeChoice#exactChoice(ItemStack...)} instead
+         */
         public ExactChoice(@NotNull ItemStack... stacks) {
             this(Arrays.asList(stacks));
         }
 
+        /**
+         * @deprecated Use {@link RecipeChoice#exactChoice(List)} instead
+         */
         public ExactChoice(@NotNull List<ItemStack> choices) {
```
Machine-Maker: add `@Deprecated(since = "1.21.4", forRemoval = true)` to all of these.

paper-api/src/main/java/org/bukkit/inventory/RecipeChoice.java
```diff
@@ -35,6 +37,46 @@ public interface RecipeChoice extends Predicate<ItemStack>, Cloneable {
...
 
+    /**
+     * Creates a choice that will be valid only if one of the stacks is
+     * exactly matched (aside from stack size).
+     *
+     * @param stacks the ItemStacks to match against.
+     *               Cannot be empty or contain empty/air stacks.
+     * @return a new ExactChoice.
+     */
+    static @NotNull ExactChoice exactChoice(@NotNull ItemStack... stacks) {
```
Machine-Maker: Lets avoid upstream's mistake. For situations where it's not valid to pass 0 values, and you want a varargs parameter, you can do smth like.
```suggestion
    static @NotNull ExactChoice exactChoice(@NotNull ItemStack first, ItemStack... others) {
```
That forces them to supply at least one value, and optionally anymore

Machine-Maker (changes requested):
Looks good, just a few more things. I also [opened the PR](https://github.com/PaperMC/Paper/pull/12049) adding the future alternative to MaterialChoice. Will merge mine in after yours

CPieter:
>Since PredicateRecipeChoice isn't public, delete it, and make PotionMix#createPredicateChoice return an instance of the new type. But also deprecate that method, and suggest to use the new method you create.
>
>Here's a great example of what I talked about in my review, suggesting to keep implementation types out of the "api" because we can now just delete PredicateRecipeChoice entirely.

Most issues with the PR should now be addressed, but I'm not quite sure how to properly replace the `PotionMix` predicate choice. The new `PredicateChoice` requires an example stack, but this might not really make sense for use in the `PotionMix`. I'm hesitant to make the example stack nullable since not having one can break recipes. Maybe the `PotionMix` update should be left to a seperate PR?

Machine-Maker:
Everything looks mostly good, some small style stuff, but we can fix that.
> The new PredicateChoice requires an example stack, but this might not really make sense for use in the PotionMix

I think what we can do is just create a dummy example stack there. The only place the old predicate choice was used was in potion mixes where the example stack didn't matter. So just creating a random itemstack isn't going to break anything unless people start using the now-deprecated method for actual recipes. Which is their problem.

derverdox:
Thanks for continuing to work on the idea! All the best, Verdox

Machine-Maker (approved):
Ok, I found the issue, (ty Claude) and did a bunch of testing, I think this is fixed now. Works for shaped/shapeless and the recipe book.

paper-api/src/main/java/org/bukkit/inventory/PredicateRecipeChoiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public boolean test(final ItemStack itemStack) {
+        return this.stackPredicate.test(itemStack);
+    }
+
+    @Override
+    public RecipeChoice validate(final boolean allowEmptyRecipes) {
+        if (this.exampleStack.getType().isAir()) {
```
Doc94:
```suggestion
        if (this.exampleStack.isEmpty()) {
```

paper-server/patches/features/0020-Improve-exact-choice-recipe-ingredients.patch
```diff
@@ -108,7 +112,9 @@ index 0000000000000000000000000000000000000000..f47c12e9dd6cfa857ca07a764edc22de
 +    public void initialize(final Recipe<?> recipe) {
 +        this.exactIngredients.clear();
```
kennytv, line 113: Doesn't the new list have to be cleared here as well?

paper-api/src/main/java/org/bukkit/inventory/PredicateRecipeChoiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public ItemStack getItemStack() {
+        return this.exampleStack.clone();
+    }
+
+    @SuppressWarnings({"MethodDoesntCallSuperMethod", "FunctionalExpressionCanBeFolded"})
+    @Override
+    public PredicateRecipeChoiceImpl clone() {
+        return new PredicateRecipeChoiceImpl(this.stackPredicate::test, this.exampleStack.clone());
```
kennytv: The constructor already clones

paper-api/src/main/java/org/bukkit/inventory/PredicateRecipeChoiceImpl.java
kennytv: The constructor already checks against this, so the check can be removed fully

Machine-Maker:
Ok, addressed reviews

kennytv (approved):

lynxplay (approved):
Not much of a fan of the test logic, maybe the models of today can emit something more logical. A full test matrix seems possibly better but requires a good bit of work. So for now they can stay in.

LGTM

PR #12019, opened 2025-01-26
Bring Entity#teleportAsync up to parity with Entity#teleport

zJanny:
The Entity.teleportAsync function did not have a overload to be able to teleport to another entity directly, and this has annoyed me quite a bit while coding, because the normal teleport has it.

So i made this small change to add this functionality

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Entity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Entity.java
@@ -237,6 +237,25 @@ public interface Entity extends Metadatable, CommandSender, Nameable, Persistent
 
     /**
      * Loads/Generates(in 1.13+) the Chunk asynchronously, and then teleports the entity when the chunk is ready.
+     * @param destination Entity to teleport to
+     * @return A future that will be completed with the result of the teleport
+     */
+    default java.util.concurrent.@NotNull CompletableFuture<Boolean> teleportAsync(final @NotNull Entity destination) {
+        return this.teleportAsync(destination.getLocation(), TeleportCause.PLUGIN);
+    }
+
+    /**
+     * Loads/Generates(in 1.13+) the Chunk asynchronously, and then teleports the entity when the chunk is ready.
+     * @param destination Entity to teleport to
+     * @param cause Reason for teleport
+     * @return A future that will be completed with the result of the teleport
+     */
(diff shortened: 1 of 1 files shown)


lynxplay:
Hi and welcome to paper :wave: 

I don't see the usage of this? Why would you teleportAsync to an entity?
The entire point of that method is that you might teleport into a chunk that isn't loaded yet.
An entity instance existing means the respective chunk said entity is in also exists.
Just on a type-basis I don't see a usecase for these, but maybe I am just missing something. Could you elaborate on that a bit?

zJanny:
> Hi and welcome to paper 👋
> 
> I don't see the usage of this? Why would you teleportAsync to an entity? The entire point of that method is that you might teleport into a chunk that isn't loaded yet. An entity instance existing means the respective chunk said entity is in also exists. Just on a type-basis I don't see a usecase for these, but maybe I am just missing something. Could you elaborate on that a bit?

As far as i know, folia needs the teleportAsync method, so this would allow someone to create a Paper plugin with Folia support that uses this functionallity. 

This would be one example i can think of, because i just encountered this issue with GrimAC

lynxplay:
If this is porting methods up from folia, then yea, this might be something to do.
Tho that is a lot different of a premise than your PRs description.

electronicboy:
These methods don't exist in folia, this is a desire to pull up some of the helper util flood for the teleport method to the teleportAsync methods; which, I'm generally on the edge around adding more helper functions for trivial operations

zJanny:
> If this is porting methods up from folia, then yea, this might be something to do. Tho that is a lot different of a premise than your PRs description.

I just wanted to word it as a more general thing.
Anyways if this is not needed feel free to close this PR :D

lynxplay:
Yea, I think I'll close this then.

Thank you anyway for the PR <3
For documentation purposes as to why the PR was closed:

Teleporting to an entity instance means the target entity exists.
Unless a plugin is doing *terrible* things, it is also loaded, as such the teleport target location is loaded.
As visible in the PR, this is then pretty much just syntax sugar for `CompletableFuture.completedFuture(entity.teleport(targetEntity))`, which is just not enough code saved to really be worth yet another method around teleportation, which is already one of the most overloaded bits of logic we have :sweat_smile: 

Thank you again! Hope my line of reasoning here was understandable.

PR #12020, opened 2025-01-26
fix PlayerChangedMainHandEvent javadoc

LemonCaramel:
This event seems to have been behaving differently than described since it was first introduced.
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/106ced01e0bd24a20633d4f3aa99c2452736be74#nms-patches%2FEntityPlayer.patch?t=508

TestPlugin
```java
@EventHandler
void event(final PlayerChangedMainHandEvent event) {
    player.sendRawMessage("Current: " + player.getMainHand());
    player.sendRawMessage("New Hand: " + event.getMainHand());
}
```
image
`PlayerChangedMainHandEvent#getMainHand()` is returning an old mainhand.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/player/PlayerChangedMainHandEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/player/PlayerChangedMainHandEvent.java
@@ -12,22 +12,31 @@ public class PlayerChangedMainHandEvent extends PlayerEvent {
 
     private static final HandlerList handlers = new HandlerList();
     //
-    private final MainHand mainHand;
+    private final MainHand newMainHand;
 
-    public PlayerChangedMainHandEvent(@NotNull Player who, @NotNull MainHand mainHand) {
+    public PlayerChangedMainHandEvent(@NotNull Player who, @NotNull MainHand newMainHand) {
         super(who);
-        this.mainHand = mainHand;
+        this.newMainHand = newMainHand;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Doc94:
i feel this a little breaking just change the PlayerChangedMainHandEvent::getMainHand now returns the old...
this can be more a fix by making the method return correctly the new hand.

LemonCaramel:
> i feel this a little breaking just change the PlayerChangedMainHandEvent::getMainHand now returns the old... this can be more a fix by making the method return correctly the new hand.

I thought about how to make the method return the new hand correctly, but it's been so long since this method has been behaving incorrectly that I chose to create a new method.

electronicboy:
This is fun; getMainHand() fells redundant and generally useless given that the info is on the player at this point; I'd ponder if the better option here is to just replace the event so that plugins can use the new event or mitigate the busted behavior from the older event

lynxplay:
We already have a new event, PlayerClientOptionsChangeEvent.
I think the way forward here is to just, deprecate getMainHand, leave it as PR suggests it (technically broken).
Add the new method for people that wanna stick to this and `@Obsolete` the event.

lynxplay:
Done as stated above and reverted the javadocs to not have this method appear "non broken".

electronicboy (approved):

NoahvdAa (approved):

PR #12022, opened 2025-01-27
add method on ItemStack to edit pdc

Machine-Maker:
Adds `ItemStack#editPersistentDataContainer(Consumer<PersistentDataContainer>)` which complements `ItemStack#getPersistentDataContainer` which returns a read-only view of the pdc on that stack.

The reason the consumer model is needed here is to allow multiple changes to the PDC without having to do lots and lots of copies. If we just had direct `set` methods, it would have to make a full copy on each set.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
@@ -1,16 +1,15 @@
 package org.bukkit.inventory;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableMap;
 import io.papermc.paper.registry.RegistryKey;
 import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.function.Consumer;
 import net.kyori.adventure.text.Component;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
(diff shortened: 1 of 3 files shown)


lynxplay (approved):

electronicboy (approved):

paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
```diff
@@ -64,10 +64,26 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
     // Paper end
 
     // Paper start - pdc
+    /**
+     * @see #editPersistentDataContainer(Consumer)
+     */
```
NoahvdAa, line 69: Should this mention that the view is readonly?

lynxplay: Yea I guess the method would benefit from proper javadocs. If you can push it, I am not around rn.

Warriorrrr: The javadocs in the PersistentDataViewHolder interface which this method is from already seems good to me, it states that it's only capable of viewing tags, not modifying them.

Warriorrrr (approved):

lynxplay:
We should open a docs PR to include the changes on the PDC page (thanks Lulu)

PR #12024, opened 2025-01-27
Deprecate Server#setSpawnRadius

Lulu13022002:
This method doesn't works for a while, previously bukkit had a
migration path that was removed at some point and i don't think it's worth
to restore it.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/Bukkit.java
+++ b/paper-api/src/main/java/org/bukkit/Bukkit.java
@@ -1373,8 +1373,9 @@ public final class Bukkit {
      * Sets the radius, in blocks, around each worlds spawn point to protect.
      *
      * @param value new spawn radius, or 0 if none
-     * @deprecated don't work anymore, set the {@code spawn-protection} in
-     * server.properties or use a plugin
+     * @deprecated has not functioned for a long time as the spawn radius is defined by the server.properties file.
+     * There is no API replacement for this method. It is generally recommended to implement "protection"-like behaviour
+     * via events or third-party plugin APIs.
      */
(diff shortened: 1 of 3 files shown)


Machine-Maker (approved):

paper-api/src/main/java/org/bukkit/Bukkit.java
```diff
@@ -1373,7 +1373,10 @@ public static int getSpawnRadius() {
      * Sets the radius, in blocks, around each worlds spawn point to protect.
      *
      * @param value new spawn radius, or 0 if none
+     * @deprecated don't work anymore, set the {@code spawn-protection} in
+     * server.properties or use a plugin
```
FireInstall: Could you please specify that you should use a protection handling plugin?
If you are looking into the javadoc it's most likely you are writing a plugin yourself and that would render this comment of "using a plugin" kinda redundant.

paper-api/src/main/java/org/bukkit/Bukkit.java
```diff
@@ -1373,7 +1373,10 @@ public static int getSpawnRadius() {
      * Sets the radius, in blocks, around each worlds spawn point to protect.
      *
      * @param value new spawn radius, or 0 if none
+     * @deprecated don't work anymore, set the {@code spawn-protection} in
```
mbax: `does not work` feels better than `don't work anymore` which reads as a command to stop working.

electronicboy: The deprecation message is too user focused, this is a dev thing

mbax (approved):

PR #12025, opened 2025-01-27
Add TrialSpawner methods for end of cooldown and next mob spawn time

### LINKED ISSUE #12023: TrialSpawner API is very lacking

The trial spawner tilestate lacks features - the most important one for me being getting and setting the `cooldown_ends_at` value.

Providing access to this value would be highly beneficial, as it would allow users to get the remaining time before the trial spawner becomes available again, as well as modify or remove the cooldown as needed.

I propose these 2 functions:

`getCooldownEndTime()` returns a long that is the gametime at which the cooldown ends (this is what the game stores as the value), if there is no value it would return 0 or null.

`setCooldownEndTime(Long)` sets this value for the tilestate.

.

### Other

Image

Implementing this feature shouldn't be too complex, considering there is already a function in setting and getting the `target_cooldown_length` value. Of course, that's just my perspective. It's not me acutally doing the coding!

Doc94:
This resolve #12023 by adding two new methods related to the when the cooldown ends in gametime and when the next mob spawn in gametime.

ref: https://minecraft.wiki/w/Trial_Spawner

### DIFF
--- a/paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
+++ b/paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
@@ -4,13 +4,40 @@ import java.util.Collection;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
 import org.bukkit.spawner.TrialSpawnerConfiguration;
-import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * Represents a captured state of a trial spawner.
  */
 public interface TrialSpawner extends TileState {
+
+    /**
+     * Gets the Gametime in ticks when the cooldown ends. 0 if not currently in cooldown.
+     *
+     * @return the Gametime in ticks
+     */
+    public long getCooldownEndsAt();
+
+    /**
+     * Sets the Gametime in ticks when the cooldown ends.
+     *
+     * @param ticks the GameTime in ticks for the new cooldown
+     */
+    public void setCooldownEndsAt(long ticks);
+
+    /**
+     * Gets the Gametime in ticks when the next spawn attempt happens. 0 if not currently active.
+     *
+     * @return the Gametime in ticks
+     */
+    public long getNextMobSpawnsAt();
+
+    /**
+     * Sets the Gametime in ticks when the next spawn attempt happens.
+     *
+     * @param ticks the Gametime in ticks for the next mob spawn
+     */
+    public void setNextMobSpawnsAt(long ticks);
 
     /**
      * Gets the length in ticks the spawner will stay in cooldown for.
--- a/paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
+++ b/paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
@@ -132,7 +132,7 @@ public interface TrialSpawner extends TileState {
      * @return a collection of entities this spawner is tracking or an empty
      *         collection if there aren't any
      */
-     Collection<Entity> getTrackedEntities();
+    Collection<Entity> getTrackedEntities();
 
     /**
      * Checks if this spawner is currently tracking the provided entity.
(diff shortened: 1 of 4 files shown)


paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
```diff
@@ -4,14 +4,41 @@
...
  * Represents a captured state of a trial spawner.
  */
 public interface TrialSpawner extends TileState {
 
+    /**
+     * Gets the Gametime in ticks when the cooldown ends. 0 if not currently in cooldown.
+     *
+     * @return the Gametime in ticks
+     */
+    public long getCooldownEndsAt();
```
lynxplay: Not a big fan of these method names, they feel *really* spigot.
`getCooldownEnd` is probably better.

paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
```diff
@@ -4,14 +4,41 @@
...
+     * @param ticks the GameTime in ticks for the new cooldown
+     */
+    public void setCooldownEndsAt(long ticks);
+
+    /**
+     * Gets the Gametime in ticks when the next spawn attempt happens. 0 if not currently active.
+     *
+     * @return the Gametime in ticks
+     */
+    public long getNextMobSpawnsAt();
```
lynxplay: `getNextSpawnAttemptTick` or something?

lynxplay (changes requested):

paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
Doc94: okay i just take the same name in the var/wiki but maybe this works better.

paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
Doc94: maybe just getNextSpawnAttempt

lynxplay: Was a bit iffy on that because that could also be the next attempted mob or whatever.

Doc94: getNextSpawnAttemptTime maybe?

kennytv (approved):
`Gametime` -> `game time` and you could maybe add a `@see` tag for a method to get the game time, else looks good

Doc94:
based in others PR is valid implement jspecify in this classes or can be done in another PR?

kennytv:
Since you're already changing all method lines you might as well update the annotations here yeah 👍 add nullmarked at the top and remove the explicit notnull annotations

paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
```diff
@@ -102,16 +132,15 @@ public interface TrialSpawner extends TileState {
      * @return a collection of entities this spawner is tracking or an empty
      *         collection if there aren't any
      */
-    @NotNull
-    public Collection<Entity> getTrackedEntities();
+     Collection<Entity> getTrackedEntities();
```
Lulu13022002: Why there is a space here

test-plugin/src/main/java/io/papermc/testplugin/TestPlugin.java
```diff
@@ -1,14 +1,43 @@
...
 public final class TestPlugin extends JavaPlugin implements Listener {
 
+    BlockData testBlockData;
+
     @Override
     public void onEnable() {
         this.getServer().getPluginManager().registerEvents(this, this);
 
         // io.papermc.testplugin.brigtests.Registration.registerViaOnEnable(this);
+        testBlockData = Material.PISTON.createBlockData();
```
Lulu13022002: Don't forget to revert that

paper-api/src/main/java/org/bukkit/block/TrialSpawner.java
Doc94: skill issue.

lynxplay (approved):

PR #12031, opened 2025-01-27
Add methods for Armadillo

### LINKED ISSUE #12026: Expand Armadillo API

Currently, it is not possible to define an armadillo's state (idle, rolling scared and unrolling)

A method to set the state

NMS

### Other

_No response_

Doc94:
Resolve #12026 by adding methods to the State of the Armadillo and expose the roll behaviours in the API.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -16,7 +16,7 @@ public interface Armadillo extends Animals {
     State getState();
 
     /**
-     * Attempt to roll up if it in {@link State#IDLE}
+     * Attempt to roll up if the armadillo is {@link State#IDLE}
      */
     void rollUp();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -35,7 +35,7 @@ public interface Armadillo extends Animals {
     void rollOut();
 
     /**
-     * Represents the current state of the Armadillo.
+     * Represents the current state of the armadillo.
      */
     enum State {
         IDLE,
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -21,7 +21,7 @@ public interface Armadillo extends Animals {
     void rollUp();
 
     /**
-     * Attempt to roll out if it's not in {@link State#IDLE}
+     * Attempt to roll out if the armadillo is not {@link State#IDLE}
      */
     void rollOut();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -1,8 +1,47 @@
 package org.bukkit.entity;
 
+import org.jspecify.annotations.NullMarked;
+
 /**
  * Represents an Armadillo.
  */
+@NullMarked
 public interface Armadillo extends Animals {
+
+    /**
+     * Get the current state of the armadillo.
+     *
+     * @return the state of the armadillo
+     */
+    public State getState();
+
+    /**
+     * Set a new state for the armadillo.
+     * <br>
+     * This will also make the armadillo make the transition to the new state.
+     *
+     * @param state the new state
+     */
+    public void setState(final State state);
+
+    /**
+     * Attempt to rollUp if it in {@link State#IDLE}
+     */
+    public void rollUp();
+
+    /**
+     * Attempt to rollUp if it's not in {@link State#IDLE}
+     */
+    public void rollOut();
+
+    /**
+     * Represents the current state of the Armadillo.
+     */
+    public enum State {
+        IDLE,
+        ROLLING,
+        SCARED,
+        UNROLLING;
+    }
 
 }
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -25,12 +25,12 @@ public interface Armadillo extends Animals {
     public void setState(final State state);
 
     /**
-     * Attempt to rollUp if it in {@link State#IDLE}
+     * Attempt to roll up if it in {@link State#IDLE}
      */
     public void rollUp();
 
     /**
-     * Attempt to rollUp if it's not in {@link State#IDLE}
+     * Attempt to roll out if it's not in {@link State#IDLE}
      */
     public void rollOut();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -16,13 +16,14 @@ public interface Armadillo extends Animals {
     State getState();
 
     /**
-     * Set a new state for the armadillo.
-     * <br>
-     * This will also make the armadillo make the transition to the new state.
-     *
-     * @param state the new state
+     * Attempt to roll up if it in {@link State#IDLE}
      */
-    void setState(final State state);
+    void rollUp();
+
+    /**
+     * Attempt to roll out if it's not in {@link State#IDLE}
+     */
+    void rollOut();
 
     /**
      * Represents the current state of the armadillo.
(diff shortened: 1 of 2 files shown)


Lulu13022002:
You don't need to specify implicit modifier (public) in interface. Same for your previous PRs.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
```diff
@@ -14,8 +15,47 @@ public net.minecraft.world.entity.animal.armadillo.Armadillo getHandle() {
...
+    }
+
+    private net.minecraft.world.entity.animal.armadillo.Armadillo.ArmadilloState stateToNMS(State state) {
+        return switch (state) {
+            case State.IDLE -> net.minecraft.world.entity.animal.armadillo.Armadillo.ArmadilloState.IDLE;
+            case State.ROLLING -> net.minecraft.world.entity.animal.armadillo.Armadillo.ArmadilloState.ROLLING;
+            case State.SCARED -> net.minecraft.world.entity.animal.armadillo.Armadillo.ArmadilloState.SCARED;
+            case State.UNROLLING -> net.minecraft.world.entity.animal.armadillo.Armadillo.ArmadilloState.UNROLLING;
+        };
+    }
```
Machine-Maker: Lets make these static and public. Maybe useful elsewhere in the future to convert between the two types.

Also you can import the nms type. No need to fully qualify the name.

Machine-Maker (dismissed):

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
Doc94: Okay.

About the the FQN its IntelliJ who add this when make import.. not sure if that is for a rule in the .editorconfig because in the past the IDE just add the import and not force the FQN...

NonSwag: That's because the armadillo Bukkit wrapper has no armadillo state enum and IJ falls back to FQN instead of static imports for sub classes

Doc94: image
ohh... yeah right... if i make fqn for the bukkit class and try again target the NMS class... then its a "IntelliJ" behaviour...

paper-api/src/main/java/org/bukkit/entity/Armadillo.java
```diff
@@ -1,8 +1,47 @@
...
+     */
+    void rollUp();
+
+    /**
+     * Attempt to roll out if it's not in {@link State#IDLE}
+     */
+    void rollOut();
+
+    /**
+     * Represents the current state of the Armadillo.
```
kennytv:
```suggestion
     * Represents the current state of the armadillo.
```

kennytv (approved):

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
```diff
@@ -14,8 +16,47 @@ public net.minecraft.world.entity.animal.armadillo.Armadillo getHandle() {
         return (net.minecraft.world.entity.animal.armadillo.Armadillo) super.getHandle();
     }
 
+    @Override
+    public State getState() {
+        return this.stateToBukkit(this.getHandle().getState());
```
Lulu13022002: stateToBukkit and stateToNMS are static methods so it's incorrect to have "this" here

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
```diff
@@ -14,8 +16,47 @@ public net.minecraft.world.entity.animal.armadillo.Armadillo getHandle() {
...
+    }
+
+    @Override
+    public void setState(final State state) {
+        Preconditions.checkArgument(state != null, "state cannot be null");
+        this.getHandle().switchToState(this.stateToNMS(state));
+    }
+
+    @Override
+    public void rollUp() {
```
Lulu13022002, line 25: I'm not sure if it's intended but those methods should probably set/reset the DANGER_DETECTED_RECENTLY memory module type otherwise the changes are immediately reverted. Alternatively the key could be exposed in the api ig.

NonSwag: I would argue re/setting the memory and and also exposing the memory key to API would be a good call

Doc94: yeah i forget about that checks.. i set the value (based in the damage call) and call erase for the roll methods for handle that, also expose the memory key for allow modify this.

Doc94: currently rollUp looks like working but the rollOut not.. when call rollOut and rollUp again not sure what behaviour im missing for that.
or just maybe document the conditions for roll

Lulu13022002: I think you need to tweaks a bit more its brain.
The issue is that rollOut is called at the very end of the animation, if you want to keep the animation you would need to adjust the ttl to the duration of the unrolling animation.

Doc94: yeah but the animation are not a client side thing?
about the brain im a little lost for what touch i try things but still have the issue with rollOut roll out but the Armadillo rollUp again

NonSwag: iirc armadillos have "state ticks" which tell the server to roll out again immediately if not defined properly

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
```diff
@@ -14,8 +17,49 @@ public net.minecraft.world.entity.animal.armadillo.Armadillo getHandle() {
...
+
+    @Override
+    public void setState(final State state) {
+        Preconditions.checkArgument(state != null, "state cannot be null");
+        this.getHandle().switchToState(CraftArmadillo.stateToNMS(state));
+    }
+
+    @Override
+    public void rollUp() {
+        this.getHandle().getBrain().setMemoryWithExpiry(MemoryModuleType.DANGER_DETECTED_RECENTLY, true, 80L);
```
lynxplay: Should be using Armadillo#SCARE_CHECK_INTERVAL

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
Doc94: its that? this thing its really hard to find because in a few tests the behaviour running is ArmadilloBallUp

kennytv (dismissed):

Doc94:
I think im just remove that methods xd
When finish the blackout

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftArmadillo.java
Lulu13022002: I will see what i can do for that poor armadillo. iirc the setter is also immediately reverted so there's no other way around

paper-api/src/main/java/org/bukkit/entity/Armadillo.java
```diff
@@ -1,8 +1,38 @@
...
 
+    /**
+     * Get the current state of the armadillo.
+     *
+     * @return the state of the armadillo
+     */
+    State getState();
+
+    /**
+     * Attempt to roll up if it in {@link State#IDLE}
```
kennytv:
```suggestion
     * Attempt to roll up if the armadillo is {@link State#IDLE}
```

paper-api/src/main/java/org/bukkit/entity/Armadillo.java
```diff
@@ -1,8 +1,38 @@
...
+     */
+    void rollUp();
+
+    /**
+     * Attempt to roll out if it's not in {@link State#IDLE}
```
kennytv:
```suggestion
     * Attempt to roll out if the armadillo is not {@link State#IDLE}
```

kennytv (approved):

Owen1212055 (approved):

lynxplay (approved):

PR #12045, opened 2025-01-30
Fix Water Animal spawn height option for Squids and Dolphins

raikasdev:
In 1.21.2 when squids and dolphins got baby variants they were moved from WaterAnimal to AgeableWaterCreature, but the Paper water animal spawn configuration code wasn't updated there.

We found this out when squids didn't spawn in our custom map with a custom sea level, even though it was configured.


electronicboy (approved):
I ponder if these should replace initialisation rather than overriding them almost immediately, but, otherwise, from a quick look, lgtm

raikasdev:
> I ponder if these should replace initialisation rather than overriding them almost immediately, but, otherwise, from a quick look, lgtm

I was thinking about that myself, but as WaterAnimal.java has it like that I decided not to change it :shrug:

PR #12061, opened 2025-02-03
Blazingly Fast multi block changes

Devlrxxh:
This is a hacky way I have been using for a long time now original repo: https://github.com/Devlrxxh/BlockChanger. I wanted to implement this into Paper directly as I myself had the issue with stuff like setBlockData or setBlock being really slow for Massive size changes. I have been using this method for a long time no issues and its really fast


Devlrxxh (commented):
Multi block changes at fast speed while not costing performance

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
...
+
+        net.minecraft.world.level.block.state.BlockState currentBlockState =
+            levelChunkSection.getStates().getAndSetUnchecked(x & 15, y & 15, z & 15, blockState);
+
+        if (currentBlockState.equals(blockState)) {
+            return;
+        }
+
+        for (Player player : chunk.getPlayersSeeingChunk()) {
+            player.sendBlockChange(location, blockData);
```
electronicboy: would it generally not make more sense to let the dirtying logic deal with this, as then it can deal with batching, etc

Devlrxxh: This here was mainly made because of speed and performance, no matter which method you use from bukkit (example: setBlock, setBlockData) and I myself had tha issue that's why I made this it's solves the issue I had

electronicboy: Yes, because the bukkit methods run a whole bunch of extra logic; What I'm asking here is does it not make sense to rely on the servers logic to send dirty blocks to viewing clients which uses logic which supports batching

Devlrxxh: I am unsure what you mean by dirty blocks but for my understanding you mean fake blocks the blocks are placed into the DataPalette Directly which basically doesn't have any extra needed logic and we call Player#sendBlockChange to send the blocks update to the player which exists on the server

electronicboy: also, actually, I'm not sure we can support going this deep for block modifications, at best we could probably go towards LevelChunkSection#setBlockState to run the small amount of logic needed to keep some counters in check;

The big concern we had around this kinda thing in the past was how to expose it in a manner suitable for plugins wanting to use this stuff, as well as, how do you refine how much work should be done to allow plugins to minise the side-effects induced; I don't think that a plugin wants to use this API to paste a schematic for their game only for it to not function as intended due to stuff like not updating the tickables state, etc

I'm also not really sure where the logic is, but, the game already has the means to send updates to clients without having to manually send them, and that logic supports dealing with batching, instead of flooding clients with individual packets

Devlrxxh: So if this was instead switched to use LevelChunkSection#setBlockState would it be fine?

Devlrxxh: And also the sendBlockChange is called since results won't be called right away instead of that we could also just refreshChunk

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
...
+        LevelChunk levelChunk = chunkCache.computeIfAbsent(chunk, c -> world.getChunk(c.getX(), c.getZ()));
+
+        LevelChunkSection levelChunkSection = levelChunk.getSection(world.getSectionIndex(y));
+
+        if (levelChunkSection.hasOnlyAir() && blockData.getMaterial().isAir()) {
+            return;
+        }
+
+        net.minecraft.world.level.block.state.BlockState currentBlockState =
+            levelChunkSection.getStates().getAndSetUnchecked(x & 15, y & 15, z & 15, blockState);
```
Malfrador: I am not sure if this is the best approach. getAndSetUnchecked will bypass any ticking blocks and fluid updates. I don't think the small overhead of LevelChunkSection#setBlockState really is worth not using it. That will properly update all those things. But I guess this could be benchmarked first.

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
...
+
+        if (currentBlockState.equals(blockState)) {
+            return;
+        }
+
+        for (Player player : chunk.getPlayersSeeingChunk()) {
```
Malfrador: Sending all changes indidually like this seems a bad idea and will flood the poor client with packets. 
You could instead call `ChunkHolder#blockChanged`and then use `ChunkHolder#broadcastChanges`, as that will take care of properly batching and only sending the sections that actually changed. There also might be a method somewhere that does both these things already, not sure.

Malfrador (changes requested):
I really like the idea of this API, IMO having something like this (maybe with some disclaimers regarding ticking tile entities) would be very nice to have. Settings blocks with the API is currently indeed really slow. 

Few things that probably aren't ideal. Might be worth looking at WE/FAWE code here too to see how they deal with this, especially in regard to making sure things behave as expected after edits.

paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -2387,9 +2387,18 @@ public default boolean createExplosion(@NotNull Entity source, float power) {
...
 
+
     /**
-     * Saves world to disk
+     * Sets multiple blocks' data at once.
+     * This method can be run asynchronously.
+     *
+     * @param blocks A map of locations and their corresponding block data to be set
      */
+    public void setBlocks(Map<Location, BlockData> blocks);
```
kennytv, line 2401: This should be a list/collection rather than a map

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
Devlrxxh: I tested and it's a really small difference

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
     }
     // Paper end
 
+    // Paper start
+    public void setBlockData(Location location, BlockData blockData, HashMap<Chunk, LevelChunk> chunkCache) {
```
kennytv: Long2ObjectMap instead of HashMap, caching by the coordinate key instead of the full Chunk object
... although in general, what's the point of the cache? Loaded chunks are already stored with fast access in a thread safe map via chunk keys

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
...
+        net.minecraft.world.level.block.state.BlockState blockState = ((CraftBlockData) blockData).getState();
+
+        Chunk chunk = location.getChunk();
+        ServerLevel world = (ServerLevel) getHandle();
```
kennytv: You can't just cast this here, else it should be moved to CraftWorld

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
```diff
@@ -565,6 +571,38 @@ public boolean hasCollisionsIn(@org.jetbrains.annotations.NotNull org.bukkit.uti
...
+
+        int x = (int) location.getX();
+        int y = location.getBlockY();
+        int z = (int) location.getZ();
```
kennytv: Why manually cast for x and z but use blockY? this should be consistent

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
Devlrxxh: I will this tomorrow

Devlrxxh: Try

kennytv (changes requested):

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
Devlrxxh: On large scale I have noticed Math.floor starts to take some usage and getBlockY is only needed for the Y axis it's a micro optimisation but I have had no issues

paper-server/src/main/java/org/bukkit/craftbukkit/CraftRegionAccessor.java
SirYwell: > So if this was instead switched to use LevelChunkSection#setBlockState would it be fine?

No, stuff like height maps would still need updating.

SirYwell:
I'd generally love to have a nice batch-editing API in Paper, but I'm not so sure about this design. Maybe it's worth to go a step back and look at use-cases to determine what's needed. 
With a method that just takes a list of locations + the block data to set, you might be able to place blocks fast, but how do you know where you want to place which block? As soon as you need to read existing blocks from the world, this suddenly isn't really useful anymore. So I'd recommend to figure out what you actually want to cover with this API, and whether that's really a common use-case.

Also thread-safety is a bit more tricky, pretty sure this could lead to errors if the underlying palette is resized while a chunk packet for the chunk to modify is built.

Devlrxxh:
You would know where to place each block since each BlockData has its Location?

SirYwell:
If you already have a location -> block data mapping, yes. But from my experience, this is rarely what people want to do. As soon as you need to access world state to determine whether to place a block at a position, this doesn't work anymore. But that's a pretty common use-case (i.e. replacing specific blocks).
One common use-case where this solution is sufficient would be placing schematics. However, in such scenario, an explicit mapping of location -> block data in any way is pretty inefficient.

Devlrxxh:
What else would you suggest? since I know some people who do this stuff like this for arena regneration, or arena copying from one place to another and the default bukkit methods aren't fast enough

Devlrxxh:
I pushed changes which doesn't use Player#sendBlockChanges anymore and uses LevelChunkSection#setBlockState instead of DataPaletteBlock#getAndSet

Devlrxxh (commented):
/

Devlrxxh:
Here is a showcase of it no issues, and sorry for laggy recording I was using a online recorder since I don't have OBS installed
https://youtu.be/0yycaj0aTv0

electronicboy:
I don't really get what a whole bunch of this PR is trying to do, it looks like a bunch of logic all slapped in random locations trying to address issues which aren't really addressed with the proposed fixes; I also really do not like that the method proposal here looks like it's the 1st class mechanism of setting multiple blocks into the world.

Providing a level of low level access to the world would be nice and is something we have discussed with some parties before, but, I would much rather see this proposed as a discussion somewhere where we can focus on creating a concrete API plan before we start worrying about the implementation details within

Devlrxxh:
Okay, I get what you mean this is pr is mainly since me and some people I work with on specfic have issues when we try to set a large amount of blocks at once and I had made a solution using reflection and was wondering if it could be implemented into paper directly so other people wouldn't face the same problem

electronicboy:
I mean code wise, I get the intent of the PR, but, the logic looks a bit misdirected in some areas or doesn't take advantage of the proper underling mechanisms in the server which already solve issues that this API faces

I think we should have a proper discussion over what such API should look like, as most of the gains in terms of setting blocks is going to be avoiding the extra overhead of the CB layer which isn't always needed, but also in terms of batching. Having a raw means to toss in a Map isn't all that I could see desired from such an API, and so, we should probably have a proper discussion over if such a method is 'enough' or if there are other concerns for consumers of this API

Devlrxxh:
How should we have this discussion should?

Warriorrrr:
Closing this for now, discussion about how an API like this can look like can take place in #12070

PR #12063, opened 2025-02-03
Increase outdated build delay to two weeks

Warriorrrr:
Our old upstream loved to always randomly shift this value to arbitrary amounts of days, but since going stable this hasn't been touched yet. Two weeks seems like a good middle ground to keep it at (though I wonder if this message is even still needed).

The "as per instructions" part also seems like a holdout from spigot, since our downloads page has no instructions.


electronicboy (approved):

NoahvdAa (approved):

Doc94:
nice moment for in future improvement this? or just remove?
also can be good remove the Paper comments in that place?

electronicboy:
In the future we can maybe just switch this towards actually querying the API to print if an update is available post server init, but, that's low priority

PR #12068, opened 2025-02-04
Add Vault block API

Warriorrrr:
Supersedes #11159

Adds new API for vault blocks

Changes made compared to the previous PR:
- Added some javadocs.
- Removed itemsToEject methods, these items are only populated when a player is opening a vault, but these can already be changed with the block dispense loot event, and also can't think of any use case for changing items in the middle of ejecting.
- Removed rewarded players methods with player instances.
- Added API for state update time.
- Made the VaultBlockEntity#serverData field public to be able to directly update the server data, without having to access the actual tile entity.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/block/Vault.java
+++ b/paper-api/src/main/java/org/bukkit/block/Vault.java
@@ -6,6 +6,7 @@ import org.bukkit.loot.LootTable;
 import org.jetbrains.annotations.Unmodifiable;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
+import java.util.Collection;
 import java.util.Set;
 import java.util.UUID;
 
@@ -136,4 +137,30 @@ public interface Vault extends TileState {
      * @apiNote Only the most recent 128 player UUIDs will be stored by vault blocks.
      */
     boolean removeRewardedPlayer(UUID playerUUID);
+
+    /**
+     * Gets an unmodifiable collection of "connected players"; players who are inside this vault's activation range and who have not received rewards yet.
+     *
+     * @apiNote Vaults will only periodically scan for nearby players, so this collection may take until the next {@link #getNextStateUpdateTime() update time} to update
+     * upon a player entering its range.
+     *
+     * @return An unmodifiable list of connected player uuids.
+     */
+    @Unmodifiable
+    Collection<UUID> getConnectedPlayers();
+
+    /**
+     * Gets the item currently being displayed inside this vault. Displayed items will automatically cycle between random items from the {@link #getDisplayedLootTable()}
+     * or {@link #getLootTable()} loot tables while this vault is active.
+     *
+     * @return The item currently being displayed inside this vault.
+     */
+    ItemStack getDisplayedItem();
+
+    /**
+     * Sets the item to display inside this vault until the next cycle.
+     *
+     * @param displayedItem The item to display
+     */
+    void setDisplayedItem(ItemStack displayedItem);
 }
--- a/paper-api/src/main/java/org/bukkit/block/Vault.java
+++ b/paper-api/src/main/java/org/bukkit/block/Vault.java
@@ -80,14 +80,14 @@ public interface Vault extends TileState {
      * @return The {@link LootTable} that will be used to display items.
      */
     @Nullable
-    LootTable getDisplayLootTable();
+    LootTable getDisplayedLootTable();
 
     /**
      * Sets the loot table that this vault will display items from.
      *
-     * @param lootTable The new display loot table, or {@code null} to clear this display override.
+     * @param lootTable The new loot table to display, or {@code null} to clear this display override.
      */
-    void setDisplayLootTable(@Nullable LootTable lootTable);
+    void setDisplayedLootTable(@Nullable LootTable lootTable);
 
     /**
      * Gets the next time (in {@link World#getGameTime() game time}) that this vault block will be updated/ticked at.
--- a/paper-api/src/main/java/org/bukkit/block/Vault.java
+++ b/paper-api/src/main/java/org/bukkit/block/Vault.java
@@ -7,6 +7,7 @@ import org.jetbrains.annotations.Unmodifiable;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 import java.util.Collection;
+import java.util.Set;
 import java.util.UUID;
 
 /**
@@ -146,15 +147,15 @@ public interface Vault extends TileState {
     boolean hasRewardedPlayer(UUID playerUUID);
 
     /**
-     * Gets an unmodifiable collection of "connected players"; players who are inside this vault's activation range and who have not received rewards yet.
+     * Gets an unmodifiable set of "connected players"; players who are inside this vault's activation range and who have not received rewards yet.
      *
      * @apiNote Vaults will only periodically scan for nearby players, so it may take until the next {@link #getNextStateUpdateTime() update time} for this
      * collection to be updated upon a player entering its range.
      *
-     * @return An unmodifiable collection of connected player uuids.
+     * @return An unmodifiable set of connected player uuids.
      */
     @Unmodifiable
-    Collection<UUID> getConnectedPlayers();
+    Set<UUID> getConnectedPlayers();
 
     /**
      * Returns whether a given player is currently connected to this vault.
--- a/paper-api/src/main/java/org/bukkit/block/Vault.java
+++ b/paper-api/src/main/java/org/bukkit/block/Vault.java
@@ -1,7 +1,139 @@
 package org.bukkit.block;
 
+import org.bukkit.World;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.loot.LootTable;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+import java.util.Set;
+import java.util.UUID;
+
 /**
- * Represents a captured state of a trial spawner.
+ * Represents a captured state of a vault.
  */
+@NullMarked
 public interface Vault extends TileState {
+    /**
+     * Gets the range in blocks at which this vault will become active when a player is near.
+     *
+     * @return This vault's activation range.
+     */
+    double getActivationRange();
+
+    /**
+     * Sets the range in blocks at which the vault will become active when a player is near.
+     *
+     * @param activationRange The new activation range.
+     * @throws IllegalArgumentException if the new range is not a number, or if the new range is more than {@link #getDeactivationRange()}.
+     */
+    void setActivationRange(double activationRange);
+
+    /**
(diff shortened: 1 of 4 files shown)


Grabsky:
May not be the scope of this PR, but perhaps these could also be addressed here?
- https://github.com/PaperMC/Paper/discussions/11687
- https://github.com/PaperMC/Paper/discussions/11679

Or would this rather be a separate PR?

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -1,17 +1,32 @@
...
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.loot.LootTable;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
 
+@NullMarked
```
Doc94, line 23: This is necesary now in implementations?

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,96 @@ public CraftVault copy() {
...
     }
+
+    @Override
+    public double getActivationRange() {
+        return this.getSnapshot().getConfig().activationRange();
+    }
+
+    @Override
+    public void setActivationRange(final double activationRange) {
+        Preconditions.checkArgument(Double.isFinite(activationRange), "deactivation range must be a number");
```
Doc94: this message maybe can be more specific about the infinite rather than just "number"?

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,96 @@ public CraftVault copy() {
...
+    }
+
+    @Override
+    public ItemStack getKeyItem() {
+        return this.getSnapshot().getConfig().keyItem().asBukkitCopy();
+    }
+
+    @Override
+    public void setKeyItem(final ItemStack key) {
+        Preconditions.checkNotNull(key, "key must not be null");
```
Doc94: This throw a NPE and not a IllegalArgumentException like another cases where this type of check is made.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,96 @@ public CraftVault copy() {
...
+        this.getSnapshot().serverData.pauseStateUpdatingUntil(nextStateUpdateTime);
+    }
+
+    @Override
+    public @Unmodifiable Set<UUID> getRewardedPlayers() {
+        return ImmutableSet.copyOf(this.getSnapshot().serverData.getRewardedPlayers());
+    }
+
+    @Override
+    public boolean addRewardedPlayer(final UUID playerUUID) {
```
Doc94, line 127: Missing non-null check here and in the remove method.

Warriorrrr:
> May not be the scope of this PR, but perhaps these could also be addressed here?
> 
>     * [VaultUnlockEvent #11687](https://github.com/PaperMC/Paper/discussions/11687)
> 
>     * [VaultStateChangeEvent #11679](https://github.com/PaperMC/Paper/discussions/11679)
> 
> 
> Or would this rather be a separate PR?

My current scope is just the vault block, I'll change the PR name

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
Warriorrrr: It's not necessary, but I added it so that I don't have to add NonNull everywhere in the implementation or get warnings about it everywhere.

Doc94: oh make sense, if is valid then i take note of that too for the warnings.

paper-api/src/main/java/org/bukkit/block/Vault.java
```diff
@@ -1,7 +1,139 @@
...
+
+    /**
+     * Gets the loot table that this vault will display items from.
+     * <p>
+     * Falls back to the regular {@link #getLootTable() loot table} if unset.
+     *
+     * @return The {@link LootTable} that will be used to display items.
+     */
+    @Nullable
+    LootTable getDisplayLootTable();
```
kennytv: displayedLootTable or lootTableDisplay

Warriorrrr: I chose the current method name so that it would be compatible with spigot, or is it fine to diverge here? (I do like displayedLootTable more)

kennytv: Ah... yeah I'd say it's fine to diverge, don't wanna add new bad names

paper-api/src/main/java/org/bukkit/block/Vault.java
```diff
@@ -1,7 +1,139 @@
...
+
+    /**
+     * Gets the players who have used a key on this vault and unlocked it.
+     *
+     * @return An unmodifiable set of player uuids.
+     *
+     * @apiNote Only the most recent 128 player UUIDs will be stored by vault blocks.
+     */
+    @Unmodifiable
+    Set<UUID> getRewardedPlayers();
```
Warriorrrr: Would a Collection or SequencedCollection return type make more sense for this method, instead of guaranteeing a Set?

FireInstall: Well, the question boils down if ever in the future it could make sense listing a player twice, doesn't it?
My take would be, that it doesn't make sense to use a general Collection, since the vault basically just keeps track of whom to not reward any time soon again, not a complete history.

paper-api/src/main/java/org/bukkit/block/Vault.java
```diff
@@ -144,10 +144,10 @@ public interface Vault extends TileState {
      * @apiNote Vaults will only periodically scan for nearby players, so this collection may take until the next {@link #getNextStateUpdateTime() update time} to update
      * upon a player entering its range.
      *
-     * @return An unmodifiable list of connected player uuids.
+     * @return An unmodifiable set of connected player uuids.
```
Doc94, line 155: not require change the mention in the docs of collection to set? (apinote and desc)

Warriorrrr: Updated those now, thanks

paper-api/src/main/java/org/bukkit/block/Vault.java
```diff
@@ -1,7 +1,166 @@
...
+     * @return {@code true} if this player was previously rewarded, and has been removed as a result of this operation.
+     *
+     * @apiNote Only the most recent 128 player UUIDs will be stored by vault blocks.
+     */
+    boolean removeRewardedPlayer(UUID playerUUID);
+
+    /**
+     * Gets an unmodifiable set of "connected players"; players who are inside this vault's activation range and who have not received rewards yet.
+     *
+     * @apiNote  Vaults will only periodically scan for nearby players, so it may take until the next {@link #getNextStateUpdateTime() update time} for this
```
Doc94:
```suggestion
     * @apiNote Vaults will only periodically scan for nearby players, so it may take until the next {@link #getNextStateUpdateTime() update time} for this
```

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,114 @@ public CraftVault copy() {
...
+
+    @Override
+    public double getActivationRange() {
+        return this.getSnapshot().getConfig().activationRange();
+    }
+
+    @Override
+    public void setActivationRange(final double activationRange) {
+        Preconditions.checkArgument(Double.isFinite(activationRange), "deactivation range must not be NaN or infinite");
+        Preconditions.checkArgument(activationRange <= getDeactivationRange(), "New activation range (%f) must be less or equal to deactivation range (%f)".formatted(activationRange, getDeactivationRange()));
```
Doc94:
```suggestion
        Preconditions.checkArgument(activationRange <= this.getDeactivationRange(), "New activation range (%s) must be less or equal to deactivation range (%s)", activationRange, this.getDeactivationRange());
```
Not sure if change but Preconditions support format then not sure why need use the String format.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,114 @@ public CraftVault copy() {
...
+        return ImmutableSet.copyOf(this.getSnapshot().getSharedData().getConnectedPlayers());
+    }
+
+    @Override
+    public ItemStack getDisplayedItem() {
+        return CraftItemStack.asBukkitCopy(this.getSnapshot().getSharedData().getDisplayItem());
+    }
+
+    @Override
+    public void setDisplayedItem(final ItemStack displayedItem) {
```
Doc94, line 159: you can put an empty item?

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
Warriorrrr: Didn't think to check if it had that, will update it for both

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
Warriorrrr: Yes, that's the default for it too

paper-api/src/main/java/org/bukkit/block/Vault.java
lynxplay: Actually pretty unsure on this statement.
A set would be useful for insertion schematics, but the returned thing is unmodifiable/immutable anyway.
Consumers cannot do anything about the contents in the first place but calling contains which, that works on a collection just as well as on a set.
I don't see much benefit from this being a set, tho I'll chat it over with others.

lynxplay (approved):

Warriorrrr:
Added has methods for the methods that used to return a set and changed them to return a collection instead

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftVault.java
```diff
@@ -24,4 +39,124 @@ public CraftVault copy() {
...
+        return this.getSnapshot().serverData.stateUpdatingResumesAt();
+    }
+
+    @Override
+    public void setNextStateUpdateTime(final long nextStateUpdateTime) {
+        this.getSnapshot().serverData.pauseStateUpdatingUntil(nextStateUpdateTime);
+    }
+
+    @Override
+    public @Unmodifiable Set<UUID> getRewardedPlayers() {
```
lynxplay: Impl methods are still returning Set here.

I think getConnectedPlayers is fine to be a Set, the order there is completely irrelevant.

PR #12069, opened 2025-02-05
Add vault change state event

Warriorrrr:
Closes #11679

Adds an event for when a vault block changes its state, allowing plugins to get and/or modify the resulting vault state (though doing something like forcefully setting the vault state to ACTIVE will cause the event to be thrown every second if there's no player nearby).

I'm not completely sure about having a getPlayer, because it requires us to pass the player instance through the unlock method and may not be completely accurate for the case where the vault becomes active. Without it, the diff would be less complex but  the event would also be less useful.


lynxplay (approved):

kennytv (approved):

notTamion:
in which case would setNewState actually work? most changes there will just cause it to try and switch back to the previous state as not all the criteria are met

PR #12077, opened 2025-02-07
Separate tick count to ensure vanilla parity

### LINKED ISSUE #12075: The Wither's mob AI behavior is inconsistent with the vanilla Minecraft.
### Expected behavior

In vanilla,the Wither does not lose its hostility towards Iron Golems when the chunk unloads.

### Observed/Actual behavior

In Paper,after unloading and reloading a chunk,the Wither momentarily loses its hostility,causing it to escape the Wither cage.

### Steps/models to reproduce

Construct the relevant mechanism within a chunk.
Move away from the chunk to ensure it unloads
Return to reload the chunk and observe the Wither's behavior.

Image

### Plugin and Datapack List

Issue persists in a clean Paper instance without plugins or datapacks.
This issue is unrelated to plugins or datapacks.

### Paper version

All Paper version.

### Other

_No response_

HaHaWTH:
Closes #12075.

This pull request introduces another entity tick counter `totalTickCount` for api usage, reset vanilla `tickCount` instead of restoring value from Spigot's `Spigot.ticksLived` tag to keep vanilla parity.

## Analysis

Spigot's restore tickCount behavior([See original patch](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/browse/CraftBukkit-Patches/0027-Save-ticks-lived-to-nbttag.patch)) causes potential issues with Wither's ai, see code below, and lead to wither loses its target when entities are reloaded:
image

## Api Changes

Now `Entity#getTicksLived` gets value from totalTickCount to avoid api behavior breakage, `Entity#setTicksLived` sets both `tickCount` and `totalTickCount`.

Currently the totalTickCount is only used for apis, if the breakage is fine, the extra counter can be removed.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -109,15 +109,17 @@
      private static final Logger LOGGER = LogUtils.getLogger();
      public static final String ID_TAG = "id";
      public static final String PASSENGERS_TAG = "Passengers";
-@@ -196,7 +_,7 @@
+@@ -196,8 +_,9 @@
      public double zOld;
      public boolean noPhysics;
      private boolean wasOnFire;
 -    public final RandomSource random = RandomSource.create();
 +    public final RandomSource random = SHARED_RANDOM; // Paper - Share random for entities to make them more random
      public int tickCount;
++    public int totalTickCount; // Paper
      private int remainingFireTicks = -this.getFireImmuneTicks();
      public boolean wasTouchingWater;
+     protected Object2DoubleMap<TagKey<Fluid>> fluidHeight = new Object2DoubleArrayMap<>(2);
 @@ -233,7 +_,7 @@
      protected UUID uuid = Mth.createInsecureUUID(this.random);
      protected String stringUUID = this.uuid.toString();
@@ -798,7 +800,7 @@
 +            if (this.maxAirTicks != this.getDefaultMaxAirSupply()) {
 +                compound.putInt("Bukkit.MaxAirSupply", this.getMaxAirSupply());
 +            }
-+            compound.putInt("Spigot.ticksLived", this.tickCount);
++            compound.putInt("Spigot.ticksLived", this.totalTickCount); // Paper
 +            // CraftBukkit end
              Component customName = this.getCustomName();
              if (customName != null) {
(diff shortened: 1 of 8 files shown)


lynxplay (approved):

paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -1,5 +1,13 @@
 --- a/net/minecraft/world/entity/Entity.java
 +++ b/net/minecraft/world/entity/Entity.java
+@@ -3,7 +_,6 @@
+ import com.google.common.collect.ImmutableList;
+ import com.google.common.collect.Iterables;
+ import com.google.common.collect.Lists;
+-import com.google.common.collect.Sets;
+ import com.google.common.collect.ImmutableList.Builder;
+ import com.mojang.logging.LogUtils;
+ import it.unimi.dsi.fastutil.floats.FloatArraySet;
```
NoahvdAa: (can be handled by merger) unnecessary diff noise?

NoahvdAa (approved):

paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -860,7 +861,7 @@
 +            // CraftBukkit start
 +            // Spigot start
 +            if (this instanceof net.minecraft.world.entity.LivingEntity) {
-+                this.tickCount = compound.getInt("Spigot.ticksLived");
++                this.totalEntityAge = compound.getInt("Spigot.ticksLived"); // Paper
```
Warriorrrr, line 864: Kinda unrelated to this PR but, is there still a reason to only read this if the entity is a living entity?

Warriorrrr (commented):
In the despawn time check inside Entity#tick, totalEntityAge should also be used instead if we don't want unloading to affect their configured despawn time.

paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
HaHaWTH: Entity despawn now uses the `totalEntityAge`

HaHaWTH: > Kinda unrelated to this PR but, is there still a reason to only read this if the entity is a living entity?

Spigot api resets tickCount for non-living entities, so leaving it here just for api behavior is same as spigot.

Owen1212055 (approved):

Dreeam-qwq (approved):

Machine-Maker (approved):

PR #12088, opened 2025-02-09
Enhance javadoc for World#setAutoSave method

### LINKED ISSUE #11287: Settings worlds to not auto-save still makes them save
### Expected behavior

Worlds should not auto-save after doing world#setAutoSave(false);

### Observed/Actual behavior

Worlds auto-save (very visible on timings)

Paper timings: https://timings.aikar.co/dev/?id=22031926125f4a879eeffb5636e58c76#timings
Spigot timings for the exact same server/setup: https://timings.spigotmc.org/?url=uqiqowular

I am not 100% sure about how the Spigot timings work, whether they would show individual world saving per world or would group them up, but if it is individual this would indicate that my problem is specific to Paper.

### Steps/models to reproduce

I am using this class to initialize my temporary worlds

```java
public class WorldLoader {
    /**
     * Generates a world which does not save, does not keep spawn loaded and is a void world
     *
     * @param worldName
     * @param environment
     * @return
     */
    public static World loadVoidTemporaryWorld(String worldName, World.Environment environment) {
        //Case where the world is already loaded
        if (Bukkit.getWorld(worldName) != null) return Bukkit.getWorld(worldName);
        File folder = new File(Bukkit.getWorldContainer().getAbsolutePath());
        if (!Files.exists(Paths.get(folder.getAbsolutePath() + File.separatorChar + worldName))) {
            Logger.warn("File  " + folder.getAbsolutePath() + File.separatorChar + worldName + " does not exist!");
            return null;
        }
        Logger.info("Loading world " + worldName + " !");

        Filter filter = newFilter -> false;
        Filter previousFilter = Bukkit.getLogger().getFilter();
        Bukkit.getLogger().setFilter(filter);

        try {
            WorldCreator worldCreator = new WorldCreator(worldName);
            worldCreator.environment(environment);
            worldCreator.keepSpawnInMemory(false);
            worldCreator.generator(new VoidGenerator());
            World world = Bukkit.createWorld(worldCreator);
            if (world != null) world.setKeepSpawnInMemory(false);
            world.setDifficulty(Difficulty.HARD);
            world.setAutoSave(false);
            Bukkit.getLogger().setFilter(previousFilter);
            return world;
        } catch (Exception exception) {
            Bukkit.getLogger().setFilter(previousFilter);
            Logger.warn("Failed to load world " + worldName + " !");
            exception.printStackTrace();
        }
        return null;
    }

    private static class VoidGenerator extends ChunkGenerator {

        @Override
        public void generateSurface(WorldInfo info, Random random, int x, int z, ChunkData data) {}

        @Override
        public boolean shouldGenerateNoise() {
            return false;
        }

        @Override
        public boolean shouldGenerateBedrock() {
            return false;
        }

        @Override
        public boolean shouldGenerateCaves() {
            return false;
        }
    }
}
```

As you can see, pretty simple. You can use it on any non-default world to load it from the worlds folder. I have verified that the loggers messages do, in fact, run, so this is indeed how the world is being loaded, and no other plugin on my server can/would load these. Plus, the void world generator code is working perfectly fine, a big indicator that this is otherwise functional.

### Plugin and Datapack List

Essentials, FreeMinecraftModels, ProtocolLib, packetevents, LibsDisguises, EliteMobs, ResourcePackManager

### Paper version

Version | 1.21-97-e71c1df (MC: 1.21) (as you can see in the timings report)

### Other

electroniccat told me to report it.

NonSwag:
The autosave option has been causing confusion for quite some time now
This change clarifies that disabling auto-save does not stop all saving operations.
This addition explicitly mentions that the world will still save on shutdown and explains the intended purpose of the method.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -2778,13 +2778,11 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
 
     /**
      * Sets whether or not the world will automatically save
-     * <p>
-     * <b>Note:</b>
-     * This does not disable saving entirely, the world will still be saved on shutdown.<br>
-     * The intended use of this method is to disable the periodical autosave by the game.
      *
      * @param value true if the world should automatically save, otherwise
      *     false
+     * @apiNote This does not disable saving entirely, the world will still be saved on shutdown.<br>
+     * The intended use of this method is to disable the periodical autosave by the game.
      */
     public void setAutoSave(boolean value);
 
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -2777,7 +2777,8 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
     public boolean isAutoSave();
 
     /**
-     * Sets whether or not the world will automatically save<br><br>
+     * Sets whether or not the world will automatically save
+     * <p>
      * <b>Note:</b>
      * This does not disable saving entirely, the world will still be saved on shutdown.<br>
      * The intended use of this method is to disable the periodical autosave by the game.
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -2777,7 +2777,10 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
     public boolean isAutoSave();
 
     /**
-     * Sets whether or not the world will automatically save
+     * Sets whether or not the world will automatically save<br><br>
+     * <b>Note:</b>
+     * This does not disable saving entirely, the world will still be saved on shutdown.<br>
+     * The intended use of this method is to disable the periodical autosave by the game.
      *
      * @param value true if the world should automatically save, otherwise
      *     false


lynxplay:
Can you check other br usages? Semi certain we put them in their own new like just for visual clarity in source.

NonSwag:
> Can you check other br usages? Semi certain we put them in their own new like just for visual clarity in source.

It seems to be very mixed
sometimes a paragraph tag is used instead, but I feel like this is weird

NoahvdAa:
Pretty sure paragraph tags are the standard for Java (and recommended in the docs), we do seem to use both though. Probably better to use <p>?

MiniDigger:
ok, hear me out, markdown https://openjdk.org/jeps/467

Warriorrrr:
Would using `@apiNote` make sense for this?

electronicboy:
Probably more `@implNote`? 🤷

emilyy-dev:
`@implNote` is more of a note for API implementors rather than API users, that's what `@apiNote` is for, (besides we don't handle `@implNote` tag)

NonSwag:
apiNote sounds reasonable

Other methods in the class are already using **Note** so would it be smart to update those docs in a separate PR as well or just leave them be?

Warriorrrr:
I don't think we wanna do general cleanup prs *yet*, so just leave them be for now.

NonSwag:
Alright, so apiNote it is?

Warriorrrr (approved):

lynxplay (approved):

lynxplay (approved):

PR #12090, opened 2025-02-09
Add support for rotation argument handling.

### LINKED ISSUE #11395: RotationArgument missing
### Expected behavior

In the brig api there is no wrapped argument for RotationArgument
`net.minecraft.commands.arguments.coordinates.RotationArgument`

### Observed/Actual behavior

missing api

### Steps/models to reproduce

well it is just missing

### Plugin and Datapack List

irrelevant

### Paper version

This server is running Paper version 1.21.1-76-master@e945cfe (2024-09-10T17:56:53Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are 1 version(s) behind
Previous version: 1.21.1-65-21f125f (MC: 1.21.1)

### Other

_No response_

NonSwag:
Resolves #11395 

Introduced a `RotationResolver` to handle rotation arguments and added related methods to `ArgumentTypes`, and argument provider classes. This enables resolving and processing rotation data using `CommandSourceStack`.

The `Vector2f` in `RotationArgument` was just a temp solution because I wasn't sure on whether to add a rotation object that clearly combines a yaw and pitch

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/command/brigadier/argument/resolvers/RotationResolver.java
+++ b/paper-api/src/main/java/io/papermc/paper/command/brigadier/argument/resolvers/RotationResolver.java
@@ -1,0 +1,16 @@
+package io.papermc.paper.command.brigadier.argument.resolvers;
+
+import io.papermc.paper.command.brigadier.CommandSourceStack;
+import org.jetbrains.annotations.ApiStatus;
+import org.joml.Vector2f;
+
+/**
+ * An {@link ArgumentResolver} that's capable of resolving
+ * a rotation argument value using a {@link CommandSourceStack}.
+ *
+ * @see io.papermc.paper.command.brigadier.argument.ArgumentTypes#rotation()
+ */
+@ApiStatus.Experimental
+@ApiStatus.NonExtendable
+public interface RotationResolver extends ArgumentResolver<Vector2f> {
+}
--- a/paper-api/src/main/java/io/papermc/paper/command/brigadier/argument/resolvers/RotationResolver.java
+++ b/paper-api/src/main/java/io/papermc/paper/command/brigadier/argument/resolvers/RotationResolver.java
@@ -3,7 +3,6 @@ package io.papermc.paper.command.brigadier.argument.resolvers;
 import io.papermc.paper.command.brigadier.CommandSourceStack;
 import io.papermc.paper.math.Rotation;
 import org.jetbrains.annotations.ApiStatus;
-import org.joml.Vector2f;
 
 /**
  * An {@link ArgumentResolver} that's capable of resolving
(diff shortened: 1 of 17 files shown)


NonSwag:
image

Warriorrrr (commented):
I think a new rotation wrapper class would be good for this, but will want see what others think first.

lynxplay:
I'd also be in favour of a rotation type, mostly because we can add some utils on there to e.g. convert to a direction vector.
x and y are also just terrible names so, another good reason for a quick interface + package private record impl.

NonSwag:
> ... we can add some utils on there to e.g. convert to a direction vector.

My math skills are too... _non-existent_... to do something like that
as far as I know a vector consists out of 3 components_?_

lynxplay:
Just ignore for now, we don't need to add utils yet. 
Was just an argument for using a proper type instead of reusing joml's Vector2F

paper-api/src/main/java/io/papermc/paper/math/Rotation.java
```diff
@@ -0,0 +1,23 @@
+package io.papermc.paper.math;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Represents a rotation with specified pitch and yaw values.
+ */
+@NullMarked
+public interface Rotation {
```
Warriorrrr, line 9: There should be some way to obtain an instance of this if we want it to be useful outside of commands, RotationImpl can be moved into the math package in api and a static method can be added in this class, the same is currently done in Position for example.

paper-server/src/main/java/io/papermc/paper/util/RotationImpl.java
```diff
@@ -0,0 +1,7 @@
+package io.papermc.paper.util;
+
+import io.papermc.paper.math.Rotation;
+import org.bukkit.util.Vector;
```
Warriorrrr: leftover unused import here

NonSwag:
I also made Location implement Rotation

NonSwag:
hmm... you are right
might have not been thought through very well 
as long as there are no APIs that utilize rotation, there is no real point in making location implement rotation
I will revert that commit

paper-server/src/main/java/io/papermc/paper/util/MCUtil.java
```diff
@@ -23,6 +23,7 @@
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.Level;
+import net.minecraft.world.phys.Vec2;
```
Warriorrrr: Unused import

paper-api/src/main/java/io/papermc/paper/command/brigadier/argument/resolvers/RotationResolver.java
```diff
@@ -0,0 +1,17 @@
+package io.papermc.paper.command.brigadier.argument.resolvers;
+
+import io.papermc.paper.command.brigadier.CommandSourceStack;
+import io.papermc.paper.math.Rotation;
+import org.jetbrains.annotations.ApiStatus;
+import org.joml.Vector2f;
```
Warriorrrr: Unused import

Warriorrrr (approved):
Has some leftover imports but rest looks good

lynxplay (approved):

PR #12092, opened 2025-02-09
Move EntityUtil to correct directory

Dreeam-qwq:
Move `EntityUtil` to paper-server's `main` source path, which prevents the IDE error about the package name having an inconsistent file directory.

Before
beofre

After
after


electronicboy (approved):

masmc05:
Shouldn't this be kept in the feature patch, but just fix the location?

electronicboy:
No, the only parts of moonrise in the feature patch are basically the hooks into mojang code, etc; the majority of it lives outside of there already

Spottedleaf (changes requested):
Move the change into the patch

Spottedleaf (approved):
nevermind keep it as is

Warriorrrr (approved):

PR #12095, opened 2025-02-10
Add API for CreakingHeart

Doc94:
Add more methods to the Creaking Heart


paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,61 @@
...
+     * @throws IllegalArgumentException if the distance is negative
+     */
+    void setMaxDistanceForCreaking(int distance);
+
+    /**
+     * Attempts to spawn a creaking for protect this creaking heart.
+     *
+     * @return the {@link Creaking} for protect the creaking heart or null if fails
+     */
+    @NullMarked
```
notTamion: Wrong annotation

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,61 @@
...
+    Creaking spawnCreaking();
+
+    /**
+     * Attempts to spread resin to adjacent blocks.
+     *
+     * @return the location of spread resin or null if it cannot spread
+     */
+    @Nullable
+    Location spreadResin();
+
```
notTamion: Extra new line

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
```diff
@@ -0,0 +1,56 @@
...
+                         }
+@@ -187,9 +_,10 @@
+ 
+     @Nullable
+     public static Creaking spawnProtector(ServerLevel level, CreakingHeartBlockEntity creakingHeart) {
++        if (creakingHeart.getDistanceCreakingTooFar() <= 0) return null; // Paper - avoid spawn if the max distance is 0
+         BlockPos blockPos = creakingHeart.getBlockPos();
+         Optional<Creaking> optional = SpawnUtil.trySpawnMob(
+-            EntityType.CREAKING, EntitySpawnReason.SPAWNER, level, blockPos, 5, 16, 8, SpawnUtil.Strategy.ON_TOP_OF_COLLIDER_NO_LEAVES, true
++            EntityType.CREAKING, EntitySpawnReason.SPAWNER, level, blockPos, 5, Math.min(SPAWN_RANGE_XZ, creakingHeart.getDistanceCreakingTooFar()), Math.min(SPAWN_RANGE_Y, creakingHeart.getDistanceCreakingTooFar()), SpawnUtil.Strategy.ON_TOP_OF_COLLIDER_NO_LEAVES, true // Paper - handle too far distance for avoid spawn out of limits
```
notTamion: Not sure why we need to min this? i think we should also increase the spawn range when we increase the max Distance.

EDIT: or add extra api for that

notTamion (changes requested):
some grammar in the jd needs to be fixed

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: yeah not sure how handle this... the idea its just avoid if i set the too far distance to 2, the block not turns into a "spawn creaking, its far, break, try again, its far, try again" any suggest?

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
Doc94: IJ try to make me use this like 4 times.. and forget that last xd fixed...

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
notTamion: Well that isn‘t going to be a problem is we scale the spawn range according to the maxDistance. So i think the best thing to do is either add extra api for the spawn range (though i think thats a bit excessive) or change the spawn range by the same factor as maxDistance was changed. but the current behaivour for this is sort of weird

Doc94: yeah still a little lost in this... unless just restrict the distance for not being low than the range... but dont like that... then not sure how best can modify here.

notTamion: I was thinking about something along the lines of this:

```suggestion
+            EntityType.CREAKING, EntitySpawnReason.SPAWNER, level, blockPos, 5, creakingHeart.getDistanceCreakingTooFar() / creakingHeart.DISTANCE_CREAKING_TOO_FAR * SPAWN_RANGE_XZ, creakingHeart.getDistanceCreakingTooFar() / creakingHeart.DISTANCE_CREAKING_TOO_FARcreakingHeart.DISTANCE_CREAKING_TOO_FAR *SPAWN_RANGE_Y, SpawnUtil.Strategy.ON_TOP_OF_COLLIDER_NO_LEAVES, true // Paper - handle too far distance for avoid spawn out of limits
```

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,60 @@
...
+     *
+     * @param distance the max distance
+     * @throws IllegalArgumentException if the distance is negative
+     */
+    void setMaxDistanceForCreaking(int distance);
+
+    /**
+     * Attempts to spawn a creaking for protect this creaking heart.
+     *
+     * @return the {@link Creaking} for protect the creaking heart or null if fails
```
lynxplay: I am a bit confused regarding the usage of this method.

Simulating the spawning behaviour of an entity type is something that might be useful API, but I don't think I'd want it to live in random block state types like this.
Developers can very well just spawn a creeking and set it via the setter.
The spawnProtector method really doesn't do more than spawning it + game events.

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,60 @@
...
+     */
+    @Nullable
+    Creaking spawnCreaking();
+
+    /**
+     * Attempts to spread resin to adjacent blocks.
+     *
+     * @return the location of spread resin or null if it cannot spread
```
lynxplay: Should note that this might call events.

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
lynxplay: I'd be in favour of *potentially* adding extra API to increase the spawn radius down the line, for now I am unsure if we need this? The most important usecase I see for this is reducing the range.

I am fine with the min here to avoid the look of spawning one outside the rage and have it die immediately. 
Scaling it is a nice idea, but I think I'd prefer extra API.

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
```diff
@@ -0,0 +1,56 @@
...
+         );
+         if (optional.isEmpty()) {
+             return null;
+@@ -335,6 +_,7 @@
+         } else {
+             this.clearCreakingInfo();
+         }
++        this.distanceCreakingTooFar = tag.contains(PAPER_CREAKING_MAX_DISTANCE_TAG, 99) ? tag.getInt(PAPER_CREAKING_MAX_DISTANCE_TAG) : DISTANCE_CREAKING_TOO_FAR; // Paper - Custom max distance for Creaking
```
Lulu13022002: Use the constants in Tag

lynxplay: Use `Tag.TAG_ANY_NUMERIC` instead of 99.

Lulu13022002 (commented):
You should also handle unplaced block state properly

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
```diff
@@ -0,0 +1,56 @@
...
+     }
+ 
+     @Override
+@@ -343,5 +_,23 @@
+         if (this.creakingInfo != null) {
+             tag.putUUID("creaking", this.creakingInfo.map(Entity::getUUID, uuid -> (UUID)uuid));
+         }
+-    }
++        tag.putInt(PAPER_CREAKING_MAX_DISTANCE_TAG, this.distanceCreakingTooFar); // Paper - Custom max distance for Creaking
```
lynxplay: A bit rough here if we wanna write.
We should keep the int but we'd ideal also want a way for the developer to fall back to vanillas value *even if it changes across versions*.

The current implementation would lock every creaking heart at the current vanilla value.

lynxplay (dismissed):

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,60 @@
...
+@NullMarked
 public interface CreakingHeart extends TileState {
+
+    /**
+     * Gets the creaking protecting this creaking heart.
+     *
+     * @return the creaking or null if this creaking heart don't have protector.
+     */
+    @Nullable
+    Creaking getCreaking();
```
lynxplay: Anything against sticking to the vanilla naming here? E.g. `getProtector`?
Other peoples input here before actually changing this tho.

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
```diff
@@ -1,7 +1,60 @@
...
+     * @throws IllegalArgumentException if the creaking passed it's in another world
+     */
+    void setCreaking(@Nullable Creaking creaking);
+
+    /**
+     * Gets the max distance between the Creaking Heart and the Creaking before to remove.
+     *
+     * @return the max distance
+     */
+    int getMaxDistanceForCreaking();
```
lynxplay: I do not like this method name.
Especially when we also might add API for the creaking spawn distance.

`getProtectorSurvivalRange`
`setProtectorSurvivalRange`
`getProtectorSpawnRange`
`setProtectorSpawnRange`
*might* work, more peoples input tho pls.

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: oh.. then what is better?
- the last change for the scaling (broken need fix that)
- the min thing and in another PR implement a proper api for that

lynxplay: Ah, I personally am not much of a fan of *scaling* it, I think I'd just add the API for it.
Preconditions to verify that spawn <= survival range and survival range >= spawn range etc.
Scaling seems nice for auto vanilla-like feeling to the API but I think API just offers more specific control and doesn't "magically" change values.

Doc94: ok that logic currently just return 0 if try to set the distance in 2...

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
Doc94: hmmm yeah.. i fell useful just call the same NMS logic but maybe can just remove that and let users use the normal ways.

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
Doc94: the lasts ones not like just becauuse that value its not for spawn.. its for limit the max distance where a creaking can "live"

maybe the first can work.. but not sure..

lynxplay: The last ones were for the new API regarding the replacement for scaling, sorry xD
But yea, more people time chime in.

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: Maybe just a 
`if (this.distanceCreakingTooFar != DISTANCE_CREAKING_TOO_FAR) tag.putInt(PAPER_CREAKING_MAX_DISTANCE_TAG, this.distanceCreakingTooFar); // Paper - Custom max distance for Creaking` ?

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: okay i base in the beacon class (then that class need use tha Tag class too) but fixed here.

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
Owen1212055: Get creaking makes more sense than protector, because people wouldn't assume that the creaking is "protecting" the heart but instead is spawned by it.

paper-api/src/main/java/org/bukkit/block/CreakingHeart.java
Owen1212055: getCreakingRemovalDistance?

Doc94: > getCreakingRemovalDistance?

not sounds bad... maybe can use that then (based in not another comments xd)

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftCreakingHeart.java
```diff
@@ -24,4 +33,51 @@ public CraftCreakingHeart copy() {
...
+    }
+
+    @Override
+    public int getCreakingRemovalDistance() {
+        return this.getSnapshot().getDistanceCreakingTooFar();
+    }
+
+    @Override
+    public void setCreakingRemovalDistance(final int distance) {
+        Preconditions.checkArgument(distance >= 0, "the distance must be >= 0");
```
Lulu13022002: You are allowing 0, yet the codec it serialize into only allow positive integer.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftCreakingHeart.java
```diff
@@ -24,4 +33,51 @@ public CraftCreakingHeart copy() {
...
+        this.getSnapshot().setDistanceCreakingTooFar(distance);
+    }
+
+    @Nullable
+    @Override
+    public Creaking spawnCreaking() {
+        if (!this.getBlockEntity().hasLevel()) {
```
Lulu13022002: Like i said previously, you should handle unplaced blockstate (calling requiredPlaced) for methods that need it. (This is not the only method that needs it)

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
```diff
@@ -0,0 +1,56 @@
...
+-                            || creakingHeart.distanceToCreaking() > 34.0
++                            || creakingHeart.distanceToCreaking() > creakingHeart.getDistanceCreakingTooFar() // Paper - custom distance for creaking
+                             || creaking.playerIsStuckInYou()) {
+                             creakingHeart.removeProtector(null);
+                         }
+@@ -199,9 +_,10 @@
+ 
+     @Nullable
+     public static Creaking spawnProtector(ServerLevel level, CreakingHeartBlockEntity creakingHeart) {
++        if (creakingHeart.getDistanceCreakingTooFar() <= 0) return null; // Paper - avoid spawn if the max distance is 0
```
Lulu13022002: Is this even possible now with the codec?

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
```diff
@@ -0,0 +1,56 @@
...
+-            EntityType.CREAKING, EntitySpawnReason.SPAWNER, level, blockPos, 5, 16, 8, SpawnUtil.Strategy.ON_TOP_OF_COLLIDER_NO_LEAVES, true
++            EntityType.CREAKING, EntitySpawnReason.SPAWNER, level, blockPos, 5, Math.min(SPAWN_RANGE_XZ, creakingHeart.getDistanceCreakingTooFar()), Math.min(SPAWN_RANGE_Y, creakingHeart.getDistanceCreakingTooFar()), SpawnUtil.Strategy.ON_TOP_OF_COLLIDER_NO_LEAVES, true // Paper - handle too far distance for avoid spawn out of limits
+         );
+         if (optional.isEmpty()) {
+             return null;
+@@ -350,6 +_,7 @@
+     protected void loadAdditional(CompoundTag tag, HolderLookup.Provider registries) {
+         super.loadAdditional(tag, registries);
+         tag.read("creaking", UUIDUtil.CODEC).ifPresentOrElse(this::setCreakingInfo, this::clearCreakingInfo);
++        tag.read(PAPER_CREAKING_MAX_DISTANCE_TAG, net.minecraft.util.ExtraCodecs.POSITIVE_INT).ifPresentOrElse(this::setDistanceCreakingTooFar, this::resetDistanceCreakingTooFar); // Paper - Custom max distance for Creaking
```
Lulu13022002: I don't think the methods are needed if you only had a public field then you could do smth like:
`distanceCreakingTooFar = tag.read(PAPER_CREAKING_MAX_DISTANCE_TAG, ExtraCodecs.POSITIVE_INT).orElse(DISTANCE_CREAKING_TOO_FAR)`

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: if allow the 0 then yes.

paper-server/patches/sources/net/minecraft/world/level/block/entity/CreakingHeartBlockEntity.java.patch
Doc94: okay i remove the methods and handle in the read with only the ifPresent and keep the field with the default value.

Lulu13022002: I think you still need to fallback to the default (when the tag is not there) to fully support the data command and reloads.

notTamion:
There is a lot of stuff depending on the despawn distance being 34. e.g. Creaking roaming distance and follow distance, redstone output, player spawning range. When setting the despawn distance to anything lower it will cause semi weird behaivour as the creaking will willingly move outside its despawn distance and etc. So i think we should either scale that stuff, Math.min it or add extra api for it. though leaving it half broken like this feels bad

Doc94:
> There is a lot of stuff depending on the despawn distance being 34. e.g. Creaking roaming distance and follow distance, redstone output, player spawning range. When setting the despawn distance to anything lower it will cause semi weird behaivour as the creaking will willingly move outside its despawn distance and etc. So i think we should either scale that stuff, Math.min it or add extra api for it. though leaving it half broken like this feels bad

ok for that then im wait for some more feedback about what do...
- Add a min like `Math.min(Math.abs(DISTANCE_CREAKING_TOO_FAR - 2), PLAYER_DETECTION_RANGE)` where DISTANCE_CREAKING_TOO_FAR its the custom field
- Add tags for the others limits and make the user handle that if its going to modify any for not restrict that behaviours

Doc94:
A little bump of this... maybe can take the second option for add tags to the rest and make user handle all if wanna touch that (or remove the set and forget the problem) the only issue i has is where are the static fields being used.... many "32" and not sure what of all are for what :/

Doc94:
I remove the whole distance API based in the implications for support that, also update for the last changes in that block from NMS

Warriorrrr (approved):

Owen1212055 (approved):
The only thing I am unsure of is manually playing the sounds, I would think its fine without. Other than that, lgtm

Doc94:
> The only thing I am unsure of is manually playing the sounds, I would think its fine without. Other than that, lgtm

You prefer user call sounds if wanna then?

Warriorrrr (approved):

Warriorrrr (approved):
Re-approving to trigger workflow again

Warriorrrr (approved):
I approve of this

kennytv (approved):

kennytv (approved):
It's so good that I need to approve again

PR #12099, opened 2025-02-11
Add proper attached blocks API to AbstractArrow

### LINKED ISSUE #12096: AbstractArrow#getAttachedBlock() returns wrong block
### Expected behavior

getAttachedBlock() method should return the block the arrow is attached to

### Observed/Actual behavior

the getAttachedBlock() method always returns air if the arrow is attached to a solid block

### Steps/models to reproduce

1. shoot an arrow into a solid wall
2. run arrow.getAttachedBlock() after it hits the wall
3. wrong block

but if you shoot a non-solid block it is fine
Image
 

### Plugin and Datapack List

Server Plugins (5):
Bukkit Plugins: packetevents, Skript, skript-reflect, ViaBackwards, ViaVersion

### Paper version

This server is running Paper version 1.21.4-144-main@edacfdf (2025-02-09T10:56:49Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

notTamion:
AbstractArrows don't just have one block they are attached to. i definitely think we should expose all the blocks to the api. whether or not the existing method should be deprecated is up for debate though the current jd for it is misleading as it makes the user think there might only be one block keeping the arrow from falling

fixes #12096

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/AbstractArrow.java
+++ b/paper-api/src/main/java/org/bukkit/entity/AbstractArrow.java
@@ -5,6 +5,7 @@ import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.Unmodifiable;
 import java.util.List;
 
 /**
@@ -114,6 +115,7 @@ public interface AbstractArrow extends Projectile {
      * @return the attached block(s) or an empty list if not attached
      */
     @NotNull
+    @Unmodifiable
     List<Block> getAttachedBlocks();
 
     /**
--- a/paper-api/src/main/java/org/bukkit/entity/AbstractArrow.java
+++ b/paper-api/src/main/java/org/bukkit/entity/AbstractArrow.java
@@ -5,6 +5,7 @@ import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import java.util.List;
 
 /**
  * Represents an arrow.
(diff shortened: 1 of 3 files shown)


electronicboy (approved):
The only real concern I see would be getAttachedBlock not returning the block plugins are expecting in relation to the hit event. I would ponder deprecating the old method just for the "you should probably consider them all, or default to the first if that fits your needs"

Owen1212055 (approved):
Future api, but may be nice to have a method to get all the blocks in a provided bounding box in a world.

notTamion:
> The only real concern I see would be getAttachedBlock not returning the block plugins are expecting in relation to the hit event.

The Block that was hit in the hitevent is already Stored in a Variable (i think). could return that in the old method if it is still there

paper-api/src/main/java/org/bukkit/entity/AbstractArrow.java
```diff
@@ -99,10 +100,22 @@ public interface AbstractArrow extends Projectile {
...
     public Block getAttachedBlock();
 
+    /**
+     * Gets the block(s) which this arrow is attached to.
+     * All the returned blocks are responsible for preventing
+     * the arrow from falling.
+     *
+     * @return the attached block(s) or an empty list if not attached
+     */
+    @NotNull
```
lynxplay, line 117: Missing unmodifiable annotation.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftAbstractArrow.java
```diff
@@ -68,12 +72,16 @@ public boolean isInBlock() {
 
     @Override
     public Block getAttachedBlock() {
+        return getAttachedBlocks().stream().findFirst().orElse(null);
```
lynxplay: Complete nitpick, but if you wanna one-line this, just use Iterables.getFirst(getAttachedBlocks(), null).

lynxplay (approved):
rest lgtm

PR #12103, opened 2025-02-12
[ci skip] fix incomplete example in javadocs for PostFlattenTagRegistrar

Doc94:


lynxplay (approved):

Lulu13022002:
It's also the case for DatapackRegistrar, it doesn't compile currently (cause BoostrapContext is not valid)

Doc94:
> It's also the case for DatapackRegistrar, it doesn't compile currently (cause BoostrapContext is not valid)

dont compile? [this](https://github.com/PaperMC/Paper/blob/404167841dcf3ac78792ffbd45c683a2173b4b08/paper-api/src/main/java/io/papermc/paper/datapack/DatapackRegistrar.java#L24)?

PR #12106, opened 2025-02-12
Cleanup damage source a bit

### LINKED ISSUE #11025: Sweeping Edge knockback vanilla inconsistency
A recent [spigot commit](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/f672c351bb5559bf2adc687cdb3a3178e75befe6#nms-patches%2Fnet%2Fminecraft%2Fworld%2Fentity%2Fplayer%2FEntityHuman.patch) no longer applies knockback if Entity#hurt returned false.

This however is incorrect as vanilla would apply the knockback even if hurt returned false.

Another victim of spigots abuse of the boolean return value to indicate damage event cancel.
Cannot be solved nicely right now without heavy work on #hurt. Needs revisiting after softspoon.

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/11025


lynxplay (changes requested):
Generally I like the cleanup, passing the cause like this is a lot nicer.
The thing I am a bit confused on is the withers/setters in DamageSource now.
*Some* are withers, e.g. perform a copy, others actually mutate the instance.
It *seems* semi arbitrary, presumably this is fine like this as the setters are only called on instances copied prior, however that feels pretty much like we are asking for an explosion down the line given the inconsistent behaviour here.

The  two setters, critial and customBlockState, are I think not in a hot enough code path where the one extra copy of a damage source (which involves 0 deep copying anyway) would be a problem.
Are there other reasons I missed as to why you decided to make them setters?

lynxplay (approved):

PR #12109, opened 2025-02-13
Mention missing World#regenerateChunk implementation in jd

masmc05:
In 1.14 spigot removed the implementation of this api, but this information is still not present in the javadocs. This also deprecates for removal the api, as with newer versions it's harder and harder to imagine this api coming back.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -476,7 +476,7 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
      * @param z Z-coordinate of the chunk
      * @return Whether the chunk was actually regenerated
      *
-     * @throws UnsupportedOperationException api is not implemented since 1.14.
+     * @throws UnsupportedOperationException not implemented
      * @deprecated regenerating a single chunk is not likely to produce the same
      * chunk as before as terrain decoration may be spread across chunks. It may
      * or may not change blocks in the adjacent chunks as well.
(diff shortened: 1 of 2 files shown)


Doc94:
maybe change the message in the implementation too (and cleanup that comment code)? currently its like "if can fix..."

Warriorrrr:
It was re-implemented for a while in 1.18.1 in [this](https://github.com/PaperMC/Paper-Archive/commit/0cc2503b88343c4d10d9e6ecf7592d56762b4cae) commit, don't remember in what version it was removed again though

lynxplay:
Pretty recently iirc? We can deprecate for removal, the comment seems overkill.

masmc05:
> It was re-implemented for a while in 1.18.1 in [this](https://github.com/PaperMC/Paper-Archive/commit/0cc2503b88343c4d10d9e6ecf7592d56762b4cae) commit, don't remember in what version it was removed again though

Yeah, seems like around 1.21/1.21.1, as 1.20.6 still had the api implemented

paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -476,11 +476,12 @@ default boolean isChunkGenerated(long chunkKey) {
...
      * @return Whether the chunk was actually regenerated
      *
+     * @throws UnsupportedOperationException not implemented
      * @deprecated regenerating a single chunk is not likely to produce the same
      * chunk as before as terrain decoration may be spread across chunks. It may
      * or may not change blocks in the adjacent chunks as well.
      */
-    @Deprecated(since = "1.13")
+    @Deprecated(since = "1.13", forRemoval = true)
     public boolean regenerateChunk(int x, int z);
```
notTamion: make this a default method than we can yoink the entire method implementation

notTamion (changes requested):

lynxplay (approved):

notTamion (approved):

notTamion (approved):

Machine-Maker (approved):

Warriorrrr (approved):

PR #12112, opened 2025-02-14
Add missing nullability annotation to sendEquipmentChange method

NonSwag:
Marked ItemStack parameter as `@Nullable` to explicitly indicate it can be null. This improves clarity, avoids nullability problems, and aligns with the method's documented behavior for handling null values.


lynxplay (approved):
Was semi wanted because I wanted to move people to ItemStack.empty(), nullable stacks are something that'll *slowly* be yanked, but yea. Fair enough, I should do that in a proper follow up PR down the line.

PR #12113, opened 2025-02-14
Add EntityAttemptSmashAttackEvent

Chaosdave34:
Allows controlling of smash attacks. Makes it possible to bypass vanilla requirements (high enough fall distance and not flying) or set your own.

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/EntityAttemptSmashAttackEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/EntityAttemptSmashAttackEvent.java
@@ -1,7 +1,6 @@
 package io.papermc.paper.event.entity;
 
 import org.bukkit.entity.LivingEntity;
-import org.bukkit.event.Cancellable;
 import org.bukkit.event.HandlerList;
 import org.bukkit.event.entity.EntityEvent;
 import org.bukkit.inventory.ItemStack;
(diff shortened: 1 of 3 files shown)


lynxplay:
I think we'd benefit from potential passing the "result" as well? Also this is very much like, PlayerAttemptSmashAttackEvent (what a name). But just attack event might make it sound like this is the actual attack instead of just an event to control the check.

Chaosdave34:
PlayerAttemptSmashAttackEvent does describe the event better. What do you mean with "passing the "result" as well"?

lynxplay:
Well, it would be nice for plugins to know if the player *can* actually perform such an attack via the event.
So, instead of computing the value after and using the Result type, we could instead just fire the event pre-cancelled in case players do not meet the requirements.

Chaosdave34:
Oh I see, yeah that makes sense. I would still run the event always when the player attempts a smash attack and add a boolean field canSmashAttack, that tells if the player is allowed to perform a smash attack before any modifications, because the event can still be used to to something if there has been a successful smash attack.

lynxplay:
Yea kinda. You'd basically mark the event as "cancelled" by setting `event.setCancelled(!canSmashAttack(player))` before calling it.
That way, plugins can "override" vanillas behaviour by setting cancelled back to`false`.

lynxplay:
Hi, thanks for the changes. Right now tho, every attack is a smash attack because it is just passing the vanilla canSmash value to the event.
Just remove that boolean from the event again, as I said above, you'd just `event.setCancelled(vanillaSmashFlagHere)` before calling.

Chaosdave34:
My idea was that the information about the vanilla behviour persists, even when other plugins "uncancel" the event.
The event is set to cancelled depending on canSmashAttack in the event constructor.

lynxplay:
Fair point, that sounds smart!
Can you still (just for my sanity) move the cancelled setter to the callside instead of the constructor.
That is how all the other "pre-canclled" cases look, avoids surprises down the road.

lynxplay (approved):

lynxplay:
After some more review, I did end up changing this back to result xD
Sorry  for the extra work, thank you for the PR!

Machine-Maker (approved):

PR #12117, opened 2025-02-15
Add Point Of Interest API

Y2Kwastaken:
The goal of this Pull Request is to add a basic level of support for Point of Interest in the API. This more used within entity AI, however, there can be exposed a useful detect Point of interests in the world. I want to flesh this out if possible, however I also want testing to be done on the current API to make sure it works as intended.

I will work more on testing and seeing if this API can be fleshed out more in the coming days/weeks.


paper-server/src/main/java/io/papermc/paper/util/PaperPoiSearchResult.java
```diff
@@ -0,0 +1,24 @@
+package io.papermc.paper.util;
+
+import io.papermc.paper.entity.PoiType;
+import org.bukkit.Location;
+
+public class PaperPoiSearchResult implements PoiSearchResult {
```
lynxplay: record maybe?
Sound easier. Also nullmarked.

paper-api/src/main/java/io/papermc/paper/entity/PoiType.java
```diff
@@ -0,0 +1,147 @@
...
+         * The poi is occupied
+         */
+        Occupancy IS_OCCUPIED = occupancy("IS_OCCUPIED");
+        /**
+         * The poi is either occupied or has space
+         */
+        Occupancy ANY = occupancy("ANY");
+
+        private static Occupancy occupancy(String enumEntryName) {
+            return Bukkit.getUnsafe().createOccupancy(enumEntryName);
```
lynxplay: Use InternalAPIBridge, that is generally the place we throw stuff like this now.

paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -4083,6 +4086,20 @@ default Location locateNearestBiome(@NotNull Location origin, @NotNull Biome bio
...
+     * @param origin where to start looking for a new point of interest at
+     * @param poiType the poi type to find
+     * @param radius the radius
+     * @param occupancy the current required occupancy of the point of interest
+     * @return a PoiSearchResult containing the closest {@link Location},
+     *         {@link PoiType} and {@link PoiType.Occupancy}, or null if no poi
+     *         was found.
+     */
+    @Nullable
+    PoiSearchResult locateNearestPoi(@NotNull Location origin, @NotNull PoiType poiType, int radius, @NotNull PoiType.Occupancy occupancy);
```
lynxplay: Radius should probably be non-negative and enforced via precondition.

paper-server/src/main/java/io/papermc/paper/entity/PaperPoiType.java
```diff
@@ -0,0 +1,84 @@
...
+import net.minecraft.core.registries.Registries;
+import net.minecraft.world.entity.ai.village.poi.PoiManager;
+import org.bukkit.NamespacedKey;
+import org.bukkit.block.data.BlockData;
+import org.bukkit.craftbukkit.CraftRegistry;
+import org.bukkit.craftbukkit.block.data.CraftBlockData;
+import org.bukkit.craftbukkit.util.Handleable;
+import org.jetbrains.annotations.NotNull;
+
+public class PaperPoiType implements PoiType, Handleable<net.minecraft.world.entity.ai.village.poi.PoiType> {
```
lynxplay: NullMarked.

paper-server/src/main/java/org/bukkit/craftbukkit/util/CraftMagicNumbers.java
```diff
@@ -771,6 +774,11 @@ public org.bukkit.inventory.ItemStack createEmptyStack() {
     }
     // Paper end - proxy ItemStack
 
+    @Override
+    public PoiType.Occupancy createOccupancy(final String enumNameEntry) {
```
lynxplay: See other, move to `PaperServerInternalAPIBridge`

lynxplay (changes requested):

paper-server/src/main/java/io/papermc/paper/util/PaperPoiSearchResult.java
Y2Kwastaken: This is standard to null mark implementations, so I know going forward?

Y2Kwastaken: I switched this, but I still think maybe if we go record should we use record notation for methods in PoiSearchResult API e.g. location() instead of getLocation() and poiType() instead of getPoiType()? I like the record pattern more for such objects, but this really isn't the "standard" with the other result APIs

Y2Kwastaken:
Is there any way in specific we should handle this case

I got this comment on spigot a few months ago. Do we either 
1) specify the discrepancy in the API documentation
2) each PoiType internal would run through something like CraftMenus but for PaperPoiTypes etc that will map to some Predicate that can make up for the non straightforward discrepancies. This isn't super straightforward and kind of annoying, but it does "fix" some things internals is a bit uneven about. 

> Just tested this, and will I get the expected result for Villager workstation (specifically farmer).
I do not get the result for beehives.
For beehives it gets only found for occupancy ANY, and returns null for the other two ones. Regardless if it is empty, 1 / 2 Bees or full.

Owen1212055:
I think that documenting the fact that some POITypes are used more as "markers" for the block rather than storing occupancy would be useful. 

Could maybe even add a method onto the poi type class if it has 0 max tickets so people can check themselves.

paper-api/src/main/java/io/papermc/paper/entity/PoiType.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    boolean is(@NotNull final BlockData data);
+
+    /**
+     * Determines whether or not this PoiType acts more of like a marker rather
+     * than acting as a store for occupants
+     *
+     * @return true if this PoiType is acting as a "marker" otherwise false
+     */
+    boolean isMarker();
```
Y2Kwastaken: @Owen1212055 thoughts on this method? I think I might need to tweak the docs a bit more on this one. I also changed the comment on the PoiType interface to go into more depth about what this method entails.

Owen1212055: Maybe instead we could say "hasOcccupants" whether or not this POI can hold occupants. Reuses terminology

Owen1212055 (approved):

paper-server/src/main/java/io/papermc/paper/entity/poi/PaperPoiSearchResult.java
```diff
@@ -0,0 +1,15 @@
+package io.papermc.paper.entity.poi;
+
+import net.minecraft.world.entity.ai.village.poi.PoiRecord;
+import org.bukkit.Location;
+import org.bukkit.World;
+import org.bukkit.craftbukkit.util.CraftLocation;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public record PaperPoiSearchResult(PoiType poiType, Location location) implements PoiSearchResult {
```
lynxplay, line 10: Probably want to clone the loc on access.

paper-server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
```diff
@@ -1910,6 +1915,46 @@ public BiomeSearchResult locateNearestBiome(Location origin, int radius, int hor
         return new CraftBiomeSearchResult(CraftBiome.minecraftHolderToBukkit(found.getSecond()), CraftLocation.toBukkit(found.getFirst(), this));
     }
 
+    @Override
+    public Location locateNearestPoi(@NotNull final Location origin, @NotNull final PoiType poiType, final int radius, final PoiType.@NotNull Occupancy occupancy) {
+        Preconditions.checkArgument(origin != null, "Location cannot be null");
+        Preconditions.checkArgument(origin.getWorld().equals(this), "The provided location must be in the same world");
```
lynxplay: getWorld is nullable, this.equals(orgin.world)

paper-server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java
```diff
@@ -1910,6 +1915,46 @@ public BiomeSearchResult locateNearestBiome(Location origin, int radius, int hor
...
+
+        final Predicate<Holder<net.minecraft.world.entity.ai.village.poi.PoiType>> predicate = type -> poiTypePredicate.test(PaperPoiType.minecraftHolderToBukkit(type));
+        final PoiManager.Occupancy nmsOccupancy = PaperPoiType.PaperOccupancy.bukkitToMinecraft(occupancy);
+        final BlockPos sourcePos = CraftLocation.toBlockPosition(origin);
+
+        return this.getHandle().getPoiManager().getInRange(predicate, sourcePos, radius, nmsOccupancy)
+            .map(record -> PaperPoiSearchResult.from(record, this))
+            .toList();
+        /*
+        with Moonrise:
```
lynxplay: Yea, we need a feature hook for this and then moonrise can patch that over.
I don't *think* we need a moonrise PR for that, moonrise does not maintain the feature hook class in its repo.

lynxplay (dismissed):

PR #12121, opened 2025-02-16
Improvement in /plugins command

Doc94:
This PR handle..
- a TODO for add a description in the `/plugins` command with the hangar mention and also a URL to current docs for plugins.
- style fixes
- warning of missing anotations
- mention docs for add plugins
- add the count of plugins per type in header

### DIFF
--- a/paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
+++ b/paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
@@ -39,7 +39,7 @@ public class PaperPluginsCommand extends BukkitCommand {
         .append(asPlainComponents("""
             Server plugins can add new behavior to your server!
             You can find new plugins on Paper's plugin repository, Hangar.
-                        
+            
             https://hangar.papermc.io/
             """));
 
@@ -53,7 +53,7 @@ public class PaperPluginsCommand extends BukkitCommand {
         .append(asPlainComponents("""
             A legacy plugin is a plugin that was made on
             very old unsupported versions of the game.
-                        
+            
             It is encouraged that you replace this plugin,
             as they might not work in the future and may cause
             performance issues.
--- a/paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
+++ b/paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
@@ -55,6 +55,7 @@ public class PaperPluginsCommand extends BukkitCommand {
 
     private static final Component LEGACY_PLUGIN_STAR = Component.text('*', TextColor.color(255, 212, 42)).hoverEvent(LEGACY_PLUGIN_INFO);
     private static final Component INFO_ICON_START = Component.text("ℹ ", INFO_COLOR);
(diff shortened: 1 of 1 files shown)


paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
```diff
@@ -198,7 +197,7 @@ public boolean execute(@NotNull CommandSender sender, @NotNull String currentAli
         if (!spigotPlugins.isEmpty()) {
             sender.sendMessage(BUKKIT_HEADER);
         }
-        
+
```
Warriorrrr, line 206: Unneeded diff

Warriorrrr (commented):
The todo on line 39 can be removed

paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
Doc94: The space removed? Or that show the ide when edit

Leguan16: I don't see why this diff is a problem. He is just removing the space. It's not in a patch anymore so I think it's fine?

electronicboy: it's extra diff, but, shouldn't really be an issue here; it would be more of a concern if this was touching some area we're expecting 20 PRs for

paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
```diff
@@ -100,13 +98,13 @@ private static <T> List<Component> formatProviders(TreeMap<String, PluginProvide
                 isFirst = false;
             } else {
                 component = PLUGIN_TICK_EMPTY;
-                //formattedSublists.add(Component.empty()); // Add an empty line, the auto chat wrapping and this makes it quite jarring.
+                //formattedSubLists.add(Component.empty()); // Add an empty line, the auto chat wrapping and this makes it quite jarring.
```
Doc94: not sure about this... can be removed?

Doc94:
With the current changes the /plugins command looks like.
image
image

paper-server/src/main/java/io/papermc/paper/command/PaperPluginsCommand.java
```diff
@@ -36,35 +35,28 @@ public class PaperPluginsCommand extends BukkitCommand {
...
             as they might not work in the future and may cause
             performance issues.
             """));
 
     private static final Component LEGACY_PLUGIN_STAR = Component.text('*', TextColor.color(255, 212, 42)).hoverEvent(LEGACY_PLUGIN_INFO);
     private static final Component INFO_ICON_START = Component.text("ℹ ", INFO_COLOR);
-    private static final Component PAPER_HEADER = Component.text("Paper Plugins:", TextColor.color(2, 136, 209));
-    private static final Component BUKKIT_HEADER = Component.text("Bukkit Plugins:", TextColor.color(237, 129, 6));
+    private static final Component PAPER_HEADER = Component.text("Paper Plugins (PLUGIN_SIZE_PLACEHOLDER):", TextColor.color(2, 136, 209));
+    private static final Component BUKKIT_HEADER = Component.text("Bukkit Plugins (PLUGIN_SIZE_PLACEHOLDER):", TextColor.color(237, 129, 6));
```
Doc94: Persona... dont like this but its like the same than move to MiniMessage and apply a tag finally...

lynxplay: If they aren't constants anymore, just make them into private static methods that take the int and return the component. E.g.
```java
private static Component paperHeader(int count) { return text(....) }
```

lynxplay (approved):

PR #12122, opened 2025-02-16
[ci skip] improvement example in javadoc for DatapackRegistrar

Doc94:
This PR based in https://github.com/PaperMC/Paper/pull/12103#issuecomment-2661093797 just mention how this example assume you add the directory need... if you just copy and run this throw a NPE for the `Objects.requireNonNull`


Lulu13022002:
That's not really what i meant, i have updated the pr

Doc94:
> That's not really what i meant, i have updated the pr

Ahhh... How i font notice that if i copy-paste the example... Or what i do now i think ._.

Machine-Maker (approved):

PR #12124, opened 2025-02-16
Fix cancelled Projectile Events still consuming arrows

### LINKED ISSUE #12123: Cancelled EntityShootBowEvents/ProjectileLaunchEvents still consume arrows
### Expected behavior

Cancelling the events causes the arrow(s) to not be consumed

### Observed/Actual behavior

The arrows are consumed anyways

### Steps/models to reproduce

Cancel EntityShootBowEvents or ProjectileLaunchEvents and fire a bow or crossbow and observe your arrow stack count/the crossbow unloading without firing an arrow.

### Plugin and Datapack List

only testing plugin

### Paper version

1.21.4#152

### Other

This happens because the arrow is already removed from the inventory/crossbow before the events are even called

notTamion:
fixes #12123
pr hardfork #9949

draft pr for now. need to come back to this

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/item/ProjectileWeaponItem.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/item/ProjectileWeaponItem.java.patch
@@ -1,6 +1,53 @@
 --- a/net/minecraft/world/item/ProjectileWeaponItem.java
 +++ b/net/minecraft/world/item/ProjectileWeaponItem.java
-@@ -62,12 +_,29 @@
+@@ -40,7 +_,20 @@
+ 
+     public abstract int getDefaultProjectileRange();
+ 
+-    protected void shoot(
++    // Paper start - prevent item consumption for cancelled events
++    protected record UnrealizedDrawResult(
++        List<ItemStack> projectileStacks,
++        @Nullable ItemStack originalInPlayerInventory // Null in case the unrealised draw result is a noop (case of Crossbow)
++    ) {
++        public void consumeProjectilesFromPlayerInventory(int projectStackIndex) {
++            if (projectStackIndex != 0 || originalInPlayerInventory == null) return;
++            if (projectileStacks.isEmpty()) return; // Whatever happened here, nothing
++            final ItemStack nonIntangibleStack = projectileStacks.get(projectStackIndex);
++            originalInPlayerInventory.shrink(nonIntangibleStack.getCount());
++        }
++    }
++    protected boolean shoot(
++    // Paper end - prevent item consumption for cancelled events
+         ServerLevel level,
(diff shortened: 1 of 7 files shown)


notTamion:
should generally be ready for review now. one thing to note is that crossbows now remove items one by one from their charged projectiles, that way plugins can choose to fire only certain projectiles and have the other ones remain charged

davidmayr:
Shouldn't setConsumeItem be updated in the API? It's currently deprecated and marked as "not currently functional"

Also, is the state of this PR stable enough to apply these patches to a server for production use? Or does this still require more work? Thanks

paper-server/patches/sources/net/minecraft/world/item/ProjectileWeaponItem.java.patch
```diff
@@ -29,31 +78,93 @@
...
++                    // moved up to ensure events uncancelled
++                    float shotPitch = crossbow.getShotPitch(shooter.getRandom(), i1);
++                    shooter.level().playSound(null, shooter.getX(), shooter.getY(), shooter.getZ(), net.minecraft.sounds.SoundEvents.CROSSBOW_SHOOT, shooter.getSoundSource(), 1.0F, shotPitch);
++                }
++                if (!event.shouldConsumeItem() && projectile instanceof final AbstractArrow abstractArrow) abstractArrow.pickup = AbstractArrow.Pickup.CREATIVE_ONLY;
++                if (event.shouldConsumeItem()) {
++                    if (weapon.is(net.minecraft.world.item.Items.CROSSBOW)) {
++                        List<ItemStack> newProjectiles = new java.util.ArrayList<>(weapon.get(DataComponents.CHARGED_PROJECTILES).getItems());
++                        newProjectiles.remove(i - (projectileItems.size() - newProjectiles.size()));
++                        weapon.set(DataComponents.CHARGED_PROJECTILES, net.minecraft.world.item.component.ChargedProjectiles.of(newProjectiles));
```
Dreeam-qwq, line 95: This set call of crossbow should change like this
```
weapon = shooter.getItemInHand(hand);
if (weapon.is(Items.CROSSBOW)) {
    ...
    weapon.set(DataComponents.CHARGED_PROJECTILES, net.minecraft.world.item.component.ChargedProjectiles.of(newProjectiles));
```
Since plugin can change player inventory's item during this stage, and set data to `weapon` can not update actual item data properly. So it needs to update item reference to the actual crossbow on hand.

See https://github.com/Winds-Studio/Leaf/compare/fe444c434f9d34ea33d0332763cf7fffdf779d38...802fed0045ad0c85a196738e0be2280d6aa928de for reference.

Dreeam-qwq (changes requested):

PR #12125, opened 2025-02-16
[ci skip] Clarify getChunkAtAsyncUrgently javadocs

Warriorrrr:
Makes it more clear that chunks loaded with this method will have a higher priority over other chunk loads, adds javadocs to a default method that didn't have them before

### DIFF
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -1481,14 +1481,14 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
 
     /**
      * Requests a {@link Chunk} to be loaded at the given coordinates
-     *
+     * <p>
      * This method makes no guarantee on how fast the chunk will load,
      * and will return the chunk to the callback at a later time.
-     *
+     * <p>
      * You should use this method if you need a chunk but do not need it
-     * immediately, and you wish to let the server control the speed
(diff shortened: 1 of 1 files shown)


paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -1559,25 +1559,45 @@ default void getChunkAtAsync(final @NotNull Block block, final boolean gen, fina
...
-     *
+     * immediately, and you wish for it to be prioritised over other
+     * chunk loads in queue.
+     * <p>
      * The future will always be executed synchronously
      * on the main Server Thread.
      *
-     * @param x X Coord
-     * @param z Z Coord
+     * @param x Chunk X-coordinate of the chunk - floor(world coordinate / 16)
```
lynxplay: just chunk x coordinate should be enough.
chunk x and loc x is already confusing enough, I don't think we can explain that in a single line nicely without making it even more confusing.

Warriorrrr: I originally copied it from another method but that pseudocode example actually seems to be wrong for negative coordinates, should I fix that in this PR or a new one?

lynxplay: Follow up PR sounds good :+1:

lynxplay (approved):

Machine-Maker (approved):

PR #12129, opened 2025-02-16
Fix `disable-tripwire-updates` option not cancelling tripwire hook updates

### LINKED ISSUE #12127: `disable-tripwire-updates` option is not cancelling tripwire hook updates
### Expected behavior

`disable-tripwire-updates` should cancel all tripwire block state updates completely

### Observed/Actual behavior

The option correctly cancels updates related to other tripwire blocks placed nearby and entities stepping on it.

It does NOT cancel updates from tripwire hook blocks, which change the `attached` state property of the tripwire — that's a bug.

### Steps/models to reproduce

1. Set up a paper server
2. Change the `block-updates.disable-tripwire-updates` option in `paper-global.yml` to `true`
3. Start the server and join the world
4. Place two solid blocks 3 blocks apart
5. Place tripwire hooks on the sides of the placed blocks facing each other
6. Place a string inbetween them — it will have `attached` state property set to `false` — correct
7. Break one of the tripwire hooks and place it back
8. The string will have the `attached` state property changed to `true` — bug

### Plugin and Datapack List

```
[15:28:45 INFO]: Server Plugins (0):
```

```
[15:28:56 INFO]: There are 3 data pack(s) enabled: [vanilla (built-in)], [file/bukkit (world)], [paper (built-in)]
[15:28:56 INFO]: There are no more data packs available
```

### Paper version

```
[15:29:07 INFO]: Checking version, please wait...
[15:29:07 INFO]: This server is running Paper version 1.21.4-152-main@4041678 (2025-02-15T22:36:39Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
```

## Possible fix
using mojang mappings under current latest paper version, 1.21.4
in class: `net.minecraft.world.level.block.TripWireHookBlock`

line 127, method `calculateState` -> the first `if (optionalValue.isPresent())` condition should be modified to:
```
if (optionalValue.isPresent() && !io.papermc.paper.configuration.GlobalConfiguration.get().blockUpdates.disableTripwireUpdates)
```

Dqu1J:
Fixes #12127 

Currently the `disable-tripwire-updates` has a bug, it does NOT cancel updates from tripwire hook blocks, which change the attached state property of the tripwire.

This PR fixes this by blocking the `calculateState` method inside `TripWireHookBlock` from running.

This is justified as the methods only functionality is:
- Checking nearby blocks to find a correct tripwire line
- Setting `attached` appropriately
- Also managing redstone signals and firing events related to them (which already never happens when `disable-tripwire-updates` is on as the tripwires never get `powered`)

TLDR it only does block updates and nothing else, half of which are already not being ran with `disable-tripwire-updates`, hence I think the best solution is to just cancel the entire method.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/TripWireHookBlock.java.patch
@@ -1,6 +1,11 @@
 --- a/net/minecraft/world/level/block/TripWireHookBlock.java
 +++ b/net/minecraft/world/level/block/TripWireHookBlock.java
-@@ -127,10 +_,10 @@
+@@ -123,14 +_,15 @@
+     public static void calculateState(
+         Level level, BlockPos pos, BlockState hookState, boolean attaching, boolean shouldNotifyNeighbours, int searchRange, @Nullable BlockState state
+     ) {
++        if (io.papermc.paper.configuration.GlobalConfiguration.get().blockUpdates.disableTripwireUpdates) return; // Paper - prevent tripwire from updating
+         Optional<Direction> optionalValue = hookState.getOptionalValue(FACING);
          if (optionalValue.isPresent()) {
              Direction direction = optionalValue.get();
              boolean flag = hookState.getOptionalValue(ATTACHED).orElse(false);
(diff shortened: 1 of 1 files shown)


lynxplay (changes requested):
Welcome to paper :tada: 

Thank you for the PR!
I am a bit unsure on cancelling the *entire* method here.
We are only preventing updates to tripwires, not to the hooks.
Additionally, if something else ever happens in this method in the future, this change would make it very difficult for us to spot that during a version update.

Does moving this check down to the actual update of the tripwire also solve the issue correctly? (I'd imagine so)

lynxplay (approved):
Thanks for the quick response!
I think that is fine, the tripwire hooks are not part of the cancelled update, even if the resulting state looks fine.
Complete nitpick on my end, but can you just move the if statement directly infront of the setBlock line.
E.g.
```java
if (!config.disableUpdatesTripWire) level.setBlock(blockPos1,....) // Paper - rest of comment
```

That way, in future updates, the patch fails to apply if the setBlock line there changes and we can correct it.
Beyond that, LGTM

Issue #12132, opened 2025-02-16
Throw EntityChangeBlockEvent during BrushableBlockEntity#brush


wizjany:
### Is your feature request related to a problem?

Currently, only the initial interaction with a brush item throws an event. Dragging the brush between blocks while holding allows players to continue brushing suspicious blocks without plugin oversight.

### Describe the solution you'd like.

`BrushableBlockEntity` should throw `EntityChangeBlockEvent`s before making changes to the block. This may require some reshuffling of the logic to preserve loot table state. Events should be thrown both for the `dusted` property change as well as the final `brushingCompleted` change.

### Describe alternatives you've considered.

https://github.com/PaperMC/Paper/pull/9898#pullrequestreview-1726114004

lynxplay:
2754d7c3b92e8fe5b04fde42fc62fc44c7737bff

PR #12133, opened 2025-02-17
Add Throw EntityChangeBlockEvent for BrushableBlockEntity#brush

Y2Kwastaken:
This Pull Request adds calls for EntityChangeBlockEvent for BrushableBlockEntities. This allows for the increased tracking described in #12132.

This warrants further testing due to the fact that the point where the loot table is unpacked at *could* cause issues, however, I was unable to notice anything significant during my testing. Input for this aspect is appreciated.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/entity/BrushableBlockEntity.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/entity/BrushableBlockEntity.java.patch
@@ -1,6 +1,70 @@
 --- a/net/minecraft/world/level/block/entity/BrushableBlockEntity.java
 +++ b/net/minecraft/world/level/block/entity/BrushableBlockEntity.java
-@@ -139,7 +_,12 @@
+@@ -64,17 +_,34 @@
+             return false;
+         } else {
+             this.coolDownEndsAtTick = startTick + 10L;
++            // Paper start - EntityChangeBlockEvent
++            // The vanilla logic here is *so* backwards, we'd be moving basically *all* following calls down.
++            // Instead, compute vanilla ourselves up here and just replace the below usages with our computed values for a free diff-on-change.
++            final int currentCompletionStage = this.getCompletionState();
++            final boolean enoughBrushesToBreak = ++this.brushCount >= REQUIRED_BRUSHES_TO_BREAK;
++            final int nextCompletionStage = this.getCompletionState();
++            final boolean differentCompletionStages = currentCompletionStage != nextCompletionStage;
++            final BlockState nextBrokenBlockState = this.getBlockState().setValue(BlockStateProperties.DUSTED, nextCompletionStage);
++            if (enoughBrushesToBreak || differentCompletionStages) {
++                if (org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(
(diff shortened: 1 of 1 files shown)


lynxplay (changes requested):
The event calls are too late right now.
Right now, the block would drop its content even if people cancelled the change event.
It would also unpack its loottable, which, that modifies the block.

We need to move the unpacking down, dropContent calls it so, I think we can just move it right before the block set when a completion state changed (needs testing).
But yea, cancelling these events should maintain the block in the same state (not drop items/unpack loot tables).

lynxplay:
Mhmmm, I did not even think about the brushCount, good catch.
This makes this entire block a lot more difficult, but yea, I think this is the way to go forward (shuffling).
The client should be perfectly fine and we still unpack the loottable on the first brush (the first brush count moves from 0 to 1 in completion stage)

Event is still not fully cancelling the "last" and final brush tho, the game event is still called and the brushCount is not reset either.

paper-server/patches/sources/net/minecraft/world/level/block/entity/BrushableBlockEntity.java.patch
```diff
@@ -1,5 +1,66 @@
 --- a/net/minecraft/world/level/block/entity/BrushableBlockEntity.java
 +++ b/net/minecraft/world/level/block/entity/BrushableBlockEntity.java
+@@ -49,6 +_,7 @@
+     @Nullable
+     public ResourceKey<LootTable> lootTable;
+     public long lootTableSeed;
++    private int oldBrushCount; // PaperMC - store old brush count for EntityChangeBlockEvent reversion
```
Y2Kwastaken: ended up going with a field here, other option was an extra parameter in the brushingComplete, method, and I'm not sure if I like that diff as much, felt it was easier to insert as a field here.

lynxplay:
769de10f62 might be a separate solution, not gonna push over yet because I am too tired.
Will review tomorrow.

paper-server/patches/sources/net/minecraft/world/level/block/entity/BrushableBlockEntity.java.patch
```diff
@@ -1,5 +1,69 @@
...
++            // Paper start - EntityChangeBlockEvent
++            // The vanilla logic here is *so* backwards, we'd be moving basically *all* following calls down.
++            // Instead, compute vanilla ourselves up here and just replace the below usages with our computed values for a free diff-on-change.
++            final int currentCompletionStage = this.getCompletionState();
++            final boolean enoughBrushesToBreak = ++this.brushCount >= REQUIRED_BRUSHES_TO_BREAK;
++            final int nextCompletionStage = this.getCompletionState();
++            final boolean differentCompletionStages = currentCompletionStage != nextCompletionStage;
++            final BlockState nextBrokenBlockState = this.getBlockState().setValue(BlockStateProperties.DUSTED, nextCompletionStage);
++            if (enoughBrushesToBreak || differentCompletionStages) {
++                if (org.bukkit.craftbukkit.event.CraftEventFactory.callEntityChangeBlockEvent(
```
Lulu13022002: This looks inverted

lynxplay (approved):

Owen1212055 (approved):

PR #12140, opened 2025-02-17
Fix getForwards/SidewaysMovement for players

### LINKED ISSUE #12139: Player#getForwardsMovement/getSidewaysMovement is not updated
### Expected behavior

The method should return the player's movement direction from -1 to 1

### Observed/Actual behavior

The method always returns 0 even when player is moving

### Steps/models to reproduce

Check getForwardsMovement value while a player is moving

### Plugin and Datapack List

Single plugin checking getForwardsMovement

### Paper version

This server is running Paper version 1.21.4-164-main@8eb8e44 (2025-02-16T22:13:26Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-151-9b9de82 (MC: 1.21.4)

### Other

_No response_

Warriorrrr:
Closes #12139

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/LivingEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/LivingEntity.java
@@ -1115,9 +1115,11 @@ public interface LivingEntity extends Attributable, Damageable, ProjectileSource
      * Retrieves the sideways movement direction of the entity.
      * <p>
      * The returned value ranges from -1 to 1, where:
-     * - Positive 1 represents movement to the left.
-     * - Negative 1 represents movement to the right.
-     * <p>
+     * <ul>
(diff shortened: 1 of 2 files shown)


lynxplay (approved):
Approved outside of the if branch styling.

Machine-Maker (approved):

Lulu13022002:
Maybe getUpwardsMovement could be overriden too for consistency even if the javadocs say otherwise.
Can you also format the javadocs to make proper list (or at least newlines inbetween) currently the list is flattened.

lynxplay:
Don't think we need to go further than sidewards/forward.
The new way of this is going to be the player input that is exposed.
We can obsolete them on Player in a follow up PR if we want but yea.

PR #12141, opened 2025-02-17
Correctly call BlockFadeEvents

### LINKED ISSUE #11566: Trampling farmland calls BlockFadeEvent
### Expected behavior

Trampling should not call the BlockFadeEvent

### Observed/Actual behavior

It does call the BlockFadeEvent

### Steps/models to reproduce

1. listen for and cancel the block fade event
2. trample a farmland block by jumping on it
3. _doesnt work_

### Plugin and Datapack List

irrelevant

### Paper version

[15:42:27 INFO]: Checking version, please wait...
[15:42:27 INFO]: This server is running Paper version 1.21.1-130-ver/1.21.1@4d2672e (2024-10-30T18:33:50Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are 1 version(s) behind
Previous version: 1.21.1-128-d348cb8 (MC: 1.21.1)

### Other

The problem with this is that if you want to prevent natural block fading for farmland but allow trampling, you would have to implement a hacky solution which involves listening for the `PlayerInteractEvent` to track when a player tramples it 
temporarily store that information for that specific block and let the `BlockFadeEvent` pass if the conditions are met

The-Epic:
Only calls BlockFadeEvents for blocks fading no longer players interacting, resolves #11566

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/DirtPathBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/DirtPathBlock.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/level/block/DirtPathBlock.java
 +++ b/net/minecraft/world/level/block/DirtPathBlock.java
-@@ -60,6 +_,11 @@
+@@ -60,6 +_,16 @@
  
      @Override
      protected void tick(BlockState state, ServerLevel level, BlockPos pos, RandomSource random) {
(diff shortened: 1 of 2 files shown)


Machine-Maker:
I wonder if its maybe better to add a parameter to the method that signals if the event should be called?

Thinking about future uses of that turnToDirt method not being noticed and so the event call not being done there.

Machine-Maker:
Yeah, I'm good with that. no entity = world caused which is what we are going for.

lynxplay (approved):

lynxplay (approved):

Machine-Maker (approved):

Issue #12126, opened 2025-02-16
Add FishHook#(get/set)BobbingFluids API


NamiUni:
### Is your feature request related to a problem?

I am developing a fishing plugin and would like to implement fishing in lava.

However, the `TagKey<Fluid>` referenced by the NMS `FishingHook` class is hardcoded. To allow fishing in lava, I would have to completely reimplement the vanilla water fishing mechanics from scratch. This would require significant effort and could negatively impact server performance.

### Describe the solution you'd like.

- Add a TagKey<Fluid> field to the `FishingHook` class and set its default value to `FluidTags.WATER`.
- Modify references to `FluidTags.WATER` within the `FishingHook` class to use the newly added field instead.
- Add two methods to the Bukkit `FishHook` class:
  - `TagKey<Fluid> getBobbingFluids()`
  - `void setBobbingFluids(TagKey<Fluid> fluids)`
- Implement these methods in `CraftFishHook`. (Would parsing with `PaperRegistries.toNms` be the appropriate implementation?)

I considered adding this as a World Configuration option, but since the demand for this feature is limited and the runtime cost for plugins is low, I believe it is unnecessary.

If this request is accepted, I am willing to submit a PR myself :)

### Describe alternatives you've considered.

- Reimplement vanilla water fishing mechanics for lava fishing within the plugin.
  - This could lead to inconsistencies between versions.
  - It may negatively impact performance.
- Attempted to modify `FluidState` using `ServerTickStartEvent`, but it didn't work.
  - Even if it worked, it would negatively impact performance.
- Use a data pack to modify `fluid/water.json` in fluid tags by adding `flowing_lava` and `lava`. (This is the closest to an ideal implementation.)
  - However, it would affect other game mechanics that rely on fluid tags. [Reference](https://minecraft.fandom.com/wiki/Tag#Fluids)
- Implement a `CustomFishingHook` class.
  - This would involve copying the NMS `FishingHook` class and modifying only a few lines in the tick method. However, publishing this code on GitHub or including it in a JAR could violate Mojang Mappings licensing terms.
  - Additionally, using NMS directly introduces compatibility issues across versions.
- Fork Paper and modify the `FishingHook` class directly.
  - This is stupid solution.

NamiUni:
I forgot that we also need the FluidRegistry modification API. However, since this is off-topic for this issue, we should create a separate issue.

Machine-Maker:
[...]

PR #12147, opened 2025-02-18
Add FishHook#(get/set)BobbingFluids API

NamiUni:
### Implements #12126

This PR allows modifying the `TagKey` referenced by `FishHook` for bobbing.  
If the associated `TagKey` includes `lava` or `flowing_lava`, fish will now be able to bite in those fluids as well.  

### ⚠ Important Notes  
This works correctly on the **server-side**, but there are some strange **client-side desync issues** to be aware of.  

If the `TagKey` includes `lava` or `flowing_lava` **without** including `water` or `flowing_water`, the client perceives the behavior differently:  
- The fishing hook **does not appear to bob** on lava.  
- The fishing hook **appears to bob** on water, even if it shouldn't.  

To make this issue clearer, I’ve attached a video demonstrating the behavior.

https://github.com/user-attachments/assets/befab695-133e-46a3-862b-5a68bf354217

I'm not very familiar with **packet handling**, so I don’t know of a good solution for this—sorry about that.  

### Other Notes  
I’ve read the **CONTRIBUTING.md**, but since this is my first contribution to Paper and my English isn’t very strong, I might have written something strange. Please check the javadoc carefully! 😆  

Additionally, I used **`PaperRegistries`** for the conversion between NMS and Paper `TagKey`, but I’m not entirely sure if that’s the correct approach.  

Thank you for your time and review! 🙇‍♂️

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftFishHook.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftFishHook.java
@@ -1,11 +1,15 @@
 package org.bukkit.craftbukkit.entity;
 
 import com.google.common.base.Preconditions;
+import io.papermc.paper.registry.PaperRegistries;
+import io.papermc.paper.registry.tag.TagKey;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.entity.projectile.FishingHook;
+import org.bukkit.Fluid;
 import org.bukkit.craftbukkit.CraftServer;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.FishHook;
+import org.jetbrains.annotations.NotNull;
 
 public class CraftFishHook extends CraftProjectile implements FishHook {
     private double biteChance = -1;
@@ -232,6 +236,16 @@ public class CraftFishHook extends CraftProjectile implements FishHook {
         final FishingHook hook = this.getHandle();
         hook.resetTimeUntilLured();
         hook.timeUntilHooked = 0; // Reset time until hooked, will be repopulated once lured time is ticked down.
+    }
+
+    @Override
+    public @NotNull TagKey<Fluid> getBobbingFluids() {
+        return PaperRegistries.fromNms(this.getHandle().fluidTagKey);
+    }
+
+    @Override
+    public void setBobbingFluids(@NotNull final TagKey<Fluid> fluids) {
+        this.getHandle().fluidTagKey = PaperRegistries.toNms(fluids);
     }
     // Paper end
 }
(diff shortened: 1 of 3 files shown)


NamiUni:
Maybe, will sending the ClientboundSetEntityMotionPacket solve the desync issue?
I'll give it a try for now.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftFishHook.java
```diff
@@ -233,5 +237,15 @@ public void resetFishingState() {
...
         hook.timeUntilHooked = 0; // Reset time until hooked, will be repopulated once lured time is ticked down.
     }
+
+    @Override
+    public @NotNull TagKey<Fluid> getBobbingFluids() {
+        return PaperRegistries.fromNms(this.getHandle().fluidTagKey);
+    }
+
+    @Override
+    public void setBobbingFluids(@NotNull final TagKey<Fluid> fluids) {
```
notTamion, line 247: Precondition here to make sure fluids isn't null

notTamion (changes requested):
Welcome to Paper!
Just from quickly looking at the video: Do the items instantly get burned when you fish?
We definitely have to figure out an efficient way to have the client keep up with the hooks position as otherwise setBobbingFluids isn't really doing what it says it does.

Owen1212055:
I feel like something like this is very tricky to get right without Vanilla here, especially due to the rather large desync introduced by Vanilla. Probably better to just wait for them to add a fluid tag at one point into this area of code, since the desync is unavoidable without some nasty hacks. Because of course, the client expects water.

lynxplay:
I agree with Owen here.
This seems a bit too hacky imo given the client does desync quite heavily.
Exposing an API like this that in the end barely works feels rough, especially when its a feature mojang does not remotely support.

NamiUni:
> I feel like something like this is very tricky to get right without Vanilla here, especially due to the rather large desync introduced by Vanilla. Probably better to just wait for them to add a fluid tag at one point into this area of code, since the desync is unavoidable without some nasty hacks. Because of course, the client expects water.

I've been trying various approaches over the past few days, but as you said, the only available methods are dirty hacks.
The simplest approach was to send a MotionPacket only when the tag is set to something other than FluidTags.WATER, but it seems that the client ignores the motion packet for the fishing hook, so that didn't work 🥲

NamiUni:
When I added lava to fluid/water.json, the fishing mechanics worked correctly.
I’d like to confirm something: My understanding is that the server sends the water tag and its associated values to the client, and the client reads this tag to determine whether the fishing hook can bob on a fluid. Is this correct?

Implementing this with a hack could cause compatibility issues in the future if Vanilla updates its mechanics. So, I’ve decided to temporarily withdraw this PR and wait for Vanilla to make changes.
If there are no objections, could you kindly close this PR?

In the meantime, I’ll add details about the issue preventing this feature’s implementation to #12126.

NamiUni:
> I agree with Owen here. This seems a bit too hacky imo given the client does desync quite heavily. Exposing an API like this that in the end barely works feels rough, especially when its a feature mojang does not remotely support.

I agree

PR #12150, opened 2025-02-18
 CreativeModeTab API 

### LINKED ISSUE #8770: getCreativeCategory() doesn't work correctly

Warriorrrr:
Closes #8770

Supersedes #8665

Adds ItemStack#getCreativeCategories and ItemType#getCreativeCategories to replace the old creative category api methods and updates the enum (javadocs copied from MM's PR).


Warriorrrr:
Is there any way to get an instance of FeatureFlagSet that works both at runtime & during testing, or would I have to add something similar to what CraftRegistry does with its static registry field

paper-api/src/main/java/org/bukkit/inventory/CreativeCategory.java
```diff
@@ -9,48 +9,92 @@ public enum CreativeCategory implements net.kyori.adventure.translation.Translat
...
-    COMBAT("combat"), // Paper
+    COMBAT("combat"),
     /**
      * All items related to brewing and potions including all types of potions,
      * their variants, and ingredients to brew them.
+     * @deprecated removed in 1.19.3
      */
-    BREWING("brewing"); // Paper
-    // Paper start
+    @Deprecated(forRemoval = true, since = "1.21.4")
```
kennytv: since should be the version that it became unusable with

Machine-Maker:
Since it's been broken for so long, I wonder if its not just better to create a whole new type, and make it an interface and registry-able type since its a registry in vanilla.

paper-server/src/main/java/org/bukkit/craftbukkit/inventory/CraftItemStack.java
```diff
@@ -693,4 +698,23 @@ public boolean matchesWithoutData(final ItemStack item, final java.util.Set<io.p
...
+        for (final Map.Entry<CreativeModeTab, Collection<net.minecraft.world.item.ItemStack>> tab : PaperCreativeCategory.CATEGORY_CONTENTS.entrySet()) {
+            if (tab.getValue().contains(this.handle)) {
+                builder.add(PaperCreativeCategory.fromNms(tab.getKey()));
+            }
+        }
+
+        // Differing data components (i.e. durability) may cause the above contains to fail, return the categories for the item type if empty.
+        final Collection<CreativeCategory> built = builder.build();
+        if (built.isEmpty()) {
+            return CraftItemType.minecraftToBukkitNew(this.handle.getItem()).getCreativeCategories();
```
Lulu13022002: This will throw a NPE for empty stacks

Lulu13022002 (commented):
The method on Material should be deprecated too

Warriorrrr:
Added a new CreativeModeTab API class and corresponding registry, deprecated the old creative category api in favour of this new one.

paper-api/src/main/java/org/bukkit/inventory/CreativeCategory.java
```diff
@@ -1,56 +1,103 @@
 package org.bukkit.inventory;
 
+import io.papermc.paper.inventory.CreativeModeTab;
+
 /**
  * Represents a category in the creative inventory.
+ *
+ * @deprecated Replaced by {@link CreativeModeTab}
  */
+@Deprecated(since = "1.21.11")
```
Doc94: Can be good add a forRemoval based in how the other methods where its used are marked with that too?

paper-api/src/main/java/io/papermc/paper/inventory/CreativeModeTab.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @return {@code true} if the scrollbar is visible on the right while this tab is focused
+     */
+    boolean scrollbarShown();
+
+    /**
+     * @param type the item type
+     * @return {@code true} if this tab contains an item with the specified item type
+     */
+    boolean containsItem(ItemType type);
```
Doc94:
```suggestion
    boolean containsItem(final ItemType type);
```

paper-api/src/main/java/io/papermc/paper/inventory/CreativeModeTab.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * @param stack the item stack
+     * @return {@code true} if this tab contains the specified item
+     */
+    boolean containsItemStack(ItemStack stack);
```
Doc94:
```suggestion
    boolean containsItemStack(final ItemStack stack);
```

paper-api/src/main/java/io/papermc/paper/inventory/CreativeModeTab.java
Lulu13022002: final modifier is not needed for methods without body.

PR #12152, opened 2025-02-19
Hide soul speed particles for vanished players

bonan:
Replaces the call to `sendParticles()` with `sendParticlesSource()` in `SpawnParticlesEffect` in order to attribute the effect to the player


Warriorrrr (approved):
Perhaps it would be nice to modify sendParticlesSource in the future so that it accepts any entity instead of only players (the method itself was added by CB), since entities can be hidden too.

Owen1212055 (approved):
Could be done by the merger, lgtm.

notTamion (approved):

lynxplay (approved):

PR #12162, opened 2025-02-23
Add additional ray trace API

TonytheMacaroni:
Supersedes #11166.

- Adds overloads to `World#rayTrace` and `World#rayTraceBlocks`, as well as the method `PositionedRayTraceConfigurationBuilder#blockCollisionMode`, that take a `BlockCollisionMode` instead of the boolean `ignorePassableBlocks` to determine block collisions.
- ~~Add `FluidCollisionMode.WATER`.~~
- Remove outdated javadocs on `World#rayTrace`, `World#rayTraceBlocks`, `PositionedRayTraceConfigurationBuilder#fluidCollisionMode`, and `PositionedRayTraceConfigurationBuilder#ignorePassableBlocks` that mention caveats with portal blocks and fluids that no longer exist. Portal blocks are always considered 'passable', and the value of `ignorePassableBlocks` has no effect on fluid collisions.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/World.java
+++ b/paper-api/src/main/java/org/bukkit/World.java
@@ -1901,13 +1901,6 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
      * Performs a ray trace that checks for block collisions using the blocks'
      * precise collision shapes.
      * <p>
-     * If collisions with passable blocks are ignored, fluid collisions are
-     * ignored as well regardless of the fluid collision mode.
-     * <p>
-     * Portal blocks are only considered passable if the ray starts within
-     * them. Apart from that collisions with portal blocks will be considered
-     * even if collisions with passable blocks are otherwise ignored.
-     * <p>
      * This may cause loading of chunks! Some implementations may impose
      * artificial restrictions on the maximum distance.
      *
@@ -1924,17 +1917,9 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
         return this.rayTraceBlocks(start, direction, maxDistance, fluidCollisionMode, ignorePassableBlocks, null);
     }
 
-    // Paper start
     /**
      * Performs a ray trace that checks for block collisions using the blocks'
      * precise collision shapes.
-     * <p>
-     * If collisions with passable blocks are ignored, fluid collisions are
-     * ignored as well regardless of the fluid collision mode.
-     * <p>
-     * Portal blocks are only considered passable if the ray starts within
-     * them. Apart from that collisions with portal blocks will be considered
-     * even if collisions with passable blocks are otherwise ignored.
      * <p>
      * This may cause loading of chunks! Some implementations may impose
      * artificial restrictions on the maximum distance.
@@ -1950,7 +1935,6 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
      * @return the ray trace hit result, or <code>null</code> if there is no hit
      */
     @Nullable RayTraceResult rayTraceBlocks(io.papermc.paper.math.@NotNull Position start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, boolean ignorePassableBlocks, @Nullable Predicate<? super Block> canCollide);
-    // Paper end
 
     /**
      * Performs a ray trace that checks for both block and entity collisions.
@@ -1958,13 +1942,6 @@ public interface World extends RegionAccessor, WorldInfo, PluginMessageRecipient
      * Block collisions use the blocks' precise collision shapes. The
      * <code>raySize</code> parameter is only taken into account for entity
      * collision checks.
-     * <p>
-     * If collisions with passable blocks are ignored, fluid collisions are
-     * ignored as well regardless of the fluid collision mode.
-     * <p>
-     * Portal blocks are only considered passable if the ray starts within them.
-     * Apart from that collisions with portal blocks will be considered even if
-     * collisions with passable blocks are otherwise ignored.
      * <p>
      * This may cause loading of chunks! Some implementations may impose
      * artificial restrictions on the maximum distance.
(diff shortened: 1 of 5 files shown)


paper-api/src/main/java/org/bukkit/World.java
```diff
@@ -1891,7 +1876,24 @@ default Iterable<? extends net.kyori.adventure.audience.Audience> audiences() {
...
+     * @param start the start position
+     * @param direction the ray direction
+     * @param maxDistance the maximum distance
+     * @param fluidCollisionMode the fluid collision mode
+     * @param blockCollisionMode the block collision mode
+     * @param canCollide predicate for blocks the ray can potentially collide
+     *     with, or <code>null</code> to consider all blocks
+     * @return the ray trace hit result, or <code>null</code> if there is no hit
+     */
+    @Nullable RayTraceResult rayTraceBlocks(io.papermc.paper.math.@NotNull Position start, @NotNull Vector direction, double maxDistance, @NotNull FluidCollisionMode fluidCollisionMode, @NotNull io.papermc.paper.raytrace.BlockCollisionMode blockCollisionMode, @Nullable Predicate<? super Block> canCollide);
```
lynxplay: The entire point of the builder was that we do not overload this further.
It is already *pain* xD

TonytheMacaroni: The overloads are necessary for the builder to function? Since it delegates to the overly long methods.

lynxplay: No, CraftWorld executes the builder.
It can/should use non public/exposed methods for that.
Overload the methods needed one last time in CraftWorld.
Any additions after that will just use the method you are about to add and add params.

TonytheMacaroni: Alright, removed the method from `World`.

paper-api/src/main/java/org/bukkit/FluidCollisionMode.java
```diff
@@ -13,6 +13,10 @@ public enum FluidCollisionMode {
      * Only collide with source fluid blocks.
      */
     SOURCE_ONLY,
+    /**
+     * Collide only with water.
+     */
+    WATER,
```
notTamion: If we are going to expose water we should add lava to the nms enum and add it here aswell

paper-api/src/main/java/io/papermc/paper/raytrace/BlockCollisionMode.java
```diff
@@ -0,0 +1,25 @@
...
+     */
+    OUTLINE,
+    /**
+     * Use the visual shape.
+     */
+    VISUAL,
+    /**
+     * Use the shape of a full block, but only consider blocks tagged with {@link org.bukkit.Tag#FALL_DAMAGE_RESETTING}.
+     */
+    FALL_DAMAGE_RESETTING
```
notTamion: Feels kinda weird to expose this one, as i think it would feel pretty random to someone using the API. And its only really there for internal Falldamage calculations

Machine-Maker: yeah, you should be able to replicate it yourself with the predicate anyways right?

TonytheMacaroni: Kinda - for some reason, this mode always uses a full block shape, instead of the actual collision shape of the block. I do agree that I don't see too much of a point of keeping it, beyond perhaps making replicating fall damage calculations in API easier to do. I don't see much of a point to *not* include it, though.

notTamion: It makes the enums we have in the API seem rather random and odly specific, same thing for the WATER, LAVA Review i left. If this uses a collision mode (full block shape) we don‘t otherwise have in the api then we should instead expose that one without the fall damage filtering part, that seems a lot more useful.

lynxplay: Generally am also on the side to expose a much as we need, as little as we can.

Overexposing internals for the sake of exposing them just makes your life harder when things change.

paper-api/src/main/java/org/bukkit/FluidCollisionMode.java
TonytheMacaroni: Added.

paper-api/src/main/java/io/papermc/paper/raytrace/BlockCollisionMode.java
TonytheMacaroni: Removed.

paper-api/src/main/java/org/bukkit/FluidCollisionMode.java
```diff
@@ -16,5 +16,13 @@ public enum FluidCollisionMode {
...
-    ALWAYS;
+    ALWAYS,
+    /**
+     * Only collide with lava fluid blocks.
+     */
+    LAVA,
+    /**
+     * Only collide with water fluid blocks.
+     */
+    WATER
```
orang3i: `WATER;`

notTamion (approved):
Looks good! might be nice to mention something in the ignorePassable jdocs (at least on the builder) that its equivalent to setting the block collision mode to the respective value

paper-api/src/main/java/org/bukkit/FluidCollisionMode.java
notTamion: doesn't really mater, afaik bukkit did it a lot? paper not so much

kennytv (approved):

Owen1212055:
The only thing I am not sure about is the whole fluid conditions.
It seems very arbitrary and at least I don't see why we should be adding our own conditions esp when its fairly easy to just check the blockstate directly.

TonytheMacaroni:
Checking using the block filter doesn't really work. As an example:

```java
Location location = ...;
World world = location.getWorld();

// Hits collidable blocks & lava fluid blocks
world.rayTrace(builder -> builder
    .blockCollisionMode(BlockCollisionMode.COLLIDER)
    .fluidCollisionMode(FluidCollisionMode.LAVA)
    .targets(RayTraceTarget.BLOCK)
);

// Hits non-waterlogged collidable blocks & lava fluid blocks
world.rayTrace(builder -> builder
    .blockCollisionMode(BlockCollisionMode.COLLIDER)
    .blockFilter(block -> {
        FluidData data = world.getFluidData(block.getX(), block.getY(), block.getZ());
        Fluid fluid = data.getFluidType();

        return fluid == Fluid.EMPTY || fluid == Fluid.LAVA || fluid == Fluid.FLOWING_LAVA;
    })
    .targets(RayTraceTarget.BLOCK)
);
```
The block filter can't really be used to be distinguish hitting the block part of a waterlogged block, or the fluid part.

TonytheMacaroni:
Rebased, and removed the fluid changes for now.

Not particularly satisfied with adding in the `LAVA` mode (and by extension, exposing `WATER`); feels like there should either be a separate fluid predicate, or modify how the block predicate is done to avoid the distinguishing issue I mentioned earlier. Either way, removed it from this PR.

Owen1212055 (approved):

PR #12165, opened 2025-02-23
Add FishHookStateChangeEvent to detect fishing hook state transitions.

NamiUni:
Currently, `PlayerFishEvent` is not triggered when a `FishHook` changes its state. This means that if you want to track state transitions, you have to monitor them using a scheduler, which is inefficient. To address this, I have implemented `FishHookStateChangeEvent`.  

- This event is useful when you want to send messages, titles, or sounds to players based on fishing hook state transitions.  
- ~~Canceling the transition to `HookState.BOBBING` prevents `Minecraft#FishingHook#catchingFish` from executing.~~
- ~~Be cautious when handling this event, as canceling it will not prevent the state transition from being re-evaluated in the next tick. If the conditions are still met, the event will be instantiated and called repeatedly. (This behavior is documented in the Javadoc.)~~ Reverted due to spamming of events and desync with the client side.

For consistency, this event is mostly copied from `PufferFishStateChangeEvent`. However, I feel that "State Transition" might be a more appropriate term than "State Change."

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/projectile/FishingHook.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/projectile/FishingHook.java.patch
@@ -42,48 +42,34 @@
                      return;
                  }
              } else {
-@@ -169,15 +_,23 @@
+@@ -169,15 +_,19 @@
              boolean flag = f > 0.0F;
              if (this.currentState == FishingHook.FishHookState.FLYING) {
                  if (this.hookedIn != null) {
--                    this.setDeltaMovement(Vec3.ZERO);
--                    this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
--                    return;
-+                    // Paper start - Add FishHookStateChangeEvent. #HOOKED_ENTITY
-+                    if (new io.papermc.paper.event.entity.FishHookStateChangeEvent((org.bukkit.entity.FishHook) getBukkitEntity(), org.bukkit.entity.FishHook.HookState.HOOKED_ENTITY).callEvent()) {
-+                        this.setDeltaMovement(Vec3.ZERO); // before this.setDeltaMovement(Vec3.ZERO);
-+                        this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
-+                        return;
-+                    }
-+                    // Paper end - Add FishHookStateChangeEvent. #HOOKED_ENTITY
++                    if (new io.papermc.paper.event.entity.FishHookStateChangeEvent((org.bukkit.entity.FishHook) getBukkitEntity(), org.bukkit.entity.FishHook.HookState.HOOKED_ENTITY).callEvent()) {// Paper start - Add FishHookStateChangeEvent. #HOOKED_ENTITY
+                     this.setDeltaMovement(Vec3.ZERO);
+                     this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
+                     return;
++                    }// Paper end - Add FishHookStateChangeEvent. #HOOKED_ENTITY
                  }
  
                  if (flag) {
--                    this.setDeltaMovement(this.getDeltaMovement().multiply(0.3, 0.2, 0.3));
--                    this.currentState = FishingHook.FishHookState.BOBBING;
--                    return;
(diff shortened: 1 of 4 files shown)


paper-server/patches/sources/net/minecraft/world/entity/projectile/FishingHook.java.patch
```diff
@@ -42,6 +42,51 @@
...
+                 if (this.hookedIn != null) {
+-                    this.setDeltaMovement(Vec3.ZERO);
+-                    this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
+-                    return;
++                    // Paper start - Add FishHookStateChangeEvent. #HOOKED_ENTITY
++                    if (new io.papermc.paper.event.entity.FishHookStateChangeEvent((org.bukkit.entity.FishHook) getBukkitEntity(), org.bukkit.entity.FishHook.HookState.HOOKED_ENTITY).callEvent()) {
++                        this.setDeltaMovement(Vec3.ZERO); // before this.setDeltaMovement(Vec3.ZERO);
++                        this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
++                        return;
++                    }
```
Owen1212055: Avoid changing the indentation of these lines to help with diff. 
So you would have a paper comment on the if statement, then the bracket 
So
if (...) { // Paper - blah
...
...
..
} // Paper - blah

This way you are not changing the code inside of the if, and it wraps the diff cleanly.

NamiUni: Thanks for the review! That makes a lot of sense!

paper-server/patches/sources/net/minecraft/world/entity/projectile/FishingHook.java.patch
```diff
@@ -42,6 +42,37 @@
...
                  }
              } else {
+@@ -169,15 +_,19 @@
+             boolean flag = f > 0.0F;
+             if (this.currentState == FishingHook.FishHookState.FLYING) {
+                 if (this.hookedIn != null) {
++                    if (new io.papermc.paper.event.entity.FishHookStateChangeEvent((org.bukkit.entity.FishHook) getBukkitEntity(), org.bukkit.entity.FishHook.HookState.HOOKED_ENTITY).callEvent()) {// Paper start - Add FishHookStateChangeEvent. #HOOKED_ENTITY
+                     this.setDeltaMovement(Vec3.ZERO);
+                     this.currentState = FishingHook.FishHookState.HOOKED_IN_ENTITY;
+                     return;
```
Warriorrrr, line 51: If the event is cancelled then the return here and in the block below is also skipped, could that have any unwanted effect?

NamiUni: Ah, I completely overlooked that. The client's interpolation would cause strange desync issues, making it appear as if the hook is still attached from the client’s perspective.

While canceling this event could be useful, it might lead to event spamming or further desync issues. I'm okay with making it non-cancellable, but what do you think?

Warriorrrr: My thoughts while testing was that it being cancellable isn't great due to the mentioned spam and desync issue, it's likely the reason there isn't a setter in the api for the current state

NamiUni: Thank you for testing!
You're absolutely right. Since most of the desired functionality can already be achieved through PlayerFishEvent, I'll remove the cancelable implementation 👍

NamiUni:
After re-testing, I confirmed that everything works perfectly. If nothing has been overlooked, then this should be complete👍

https://github.com/user-attachments/assets/3c847469-9a7e-42a0-a7be-15f39db6b8fb

Warriorrrr (approved):

kennytv (approved):

notTamion (approved):

NamiUni:
Hi notTamion!
Sorry for leaving the pull request unattended for a while. I just noticed that you updated and merged it—thank you so much for taking care of it! I really appreciate your help and the time you took to make those edits.

PR #12169, opened 2025-02-23
Update effect

### LINKED ISSUE #12082: Effect.STEP_SOUND is actually the block break effect
### Expected behavior

levelevent id 2001 should be PARTICLES_DESTROY_BLOCK, which spawns block break particles and makes the noise, but 

### Observed/Actual behavior

in the effect api its 
```java
  /**
     * Sound of a block breaking. Needs {@link org.bukkit.block.data.BlockData} as additional info.
     */
    STEP_SOUND(2001, Type.SOUND, org.bukkit.block.data.BlockData.class, Material.class), // Paper - block data is more correct, but the impl of the mtehods will still work with Material
```

### Steps/models to reproduce

N/A

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.21.4-DEV-main@742968e (1970-01-01T00:00:00Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are 1 version(s) behind

### Other

Used in breakNaturally:
```java
 @Override
    public boolean breakNaturally(ItemStack item, boolean triggerEffect, boolean dropExperience) {
      ...
                    this.world.levelEvent(net.minecraft.world.level.block.LevelEvent.PARTICLES_DESTROY_BLOCK, this.position, net.minecraft.world.level.block.Block.getId(iblockdata));
                }
```

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/12082

And fix more issues:
- RECORD_PLAY should take a jukebox song
- DRAGON_BREATH/EXTINGUISH should take a boolean

This is just a bandaid, later a better api should be done to properly enforce data types

### DIFF
--- a/paper-api/src/main/java/org/bukkit/Color.java
+++ b/paper-api/src/main/java/org/bukkit/Color.java
@@ -201,10 +201,10 @@ public final class Color implements ConfigurationSerializable {
     }
 
     private Color(int alpha, int red, int green, int blue) {
-        Preconditions.checkArgument(alpha >= 0 && alpha <= BIT_MASK, "Alpha[%s] is not between 0-255", alpha);
-        Preconditions.checkArgument(red >= 0 && red <= BIT_MASK, "Red[%s] is not between 0-255", red);
-        Preconditions.checkArgument(green >= 0 && green <= BIT_MASK, "Green[%s] is not between 0-255", green);
-        Preconditions.checkArgument(blue >= 0 && blue <= BIT_MASK, "Blue[%s] is not between 0-255", blue);
+        Preconditions.checkArgument(alpha >= 0 && alpha <= 0xff, "Alpha[%s] is not between 0-255", alpha);
+        Preconditions.checkArgument(red >= 0 && red <= 0xff, "Red[%s] is not between 0-255", red);
+        Preconditions.checkArgument(green >= 0 && green <= 0xff, "Green[%s] is not between 0-255", green);
+        Preconditions.checkArgument(blue >= 0 && blue <= 0xff, "Blue[%s] is not between 0-255", blue);
 
         this.alpha = (byte) alpha;
         this.red = (byte) red;
@@ -401,7 +401,6 @@ public final class Color implements ConfigurationSerializable {
         );
     }
 
-    @SuppressWarnings("javadoc")
     @NotNull
     public static Color deserialize(@NotNull Map<String, Object> map) {
         return fromARGB(
--- a/paper-api/src/test/java/org/bukkit/ColorTest.java
+++ b/paper-api/src/test/java/org/bukkit/ColorTest.java
@@ -1,13 +1,15 @@
 package org.bukkit;
 
-import static org.bukkit.support.MatcherAssert.*;
-import static org.hamcrest.Matchers.*;
-import static org.junit.jupiter.api.Assertions.*;
 import java.util.Locale;
 import org.bukkit.configuration.file.YamlConfiguration;
 import org.junit.jupiter.api.Test;
 
-@SuppressWarnings("javadoc")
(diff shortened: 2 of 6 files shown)


paper-api/src/main/java/org/bukkit/Color.java
```diff
@@ -201,10 +201,10 @@ private Color(int red, int green, int blue) {
     }
 
     private Color(int alpha, int red, int green, int blue) {
-        Preconditions.checkArgument(alpha >= 0 && alpha <= BIT_MASK, "Alpha[%s] is not between 0-255", alpha);
-        Preconditions.checkArgument(red >= 0 && red <= BIT_MASK, "Red[%s] is not between 0-255", red);
-        Preconditions.checkArgument(green >= 0 && green <= BIT_MASK, "Green[%s] is not between 0-255", green);
-        Preconditions.checkArgument(blue >= 0 && blue <= BIT_MASK, "Blue[%s] is not between 0-255", blue);
+        Preconditions.checkArgument(alpha >= 0 && alpha <= 0xff, "Alpha[%s] is not between 0-255", alpha);
```
FireInstall, line 204: What's up with this diff? Why use the Value directly instead of `BIT_MASK`?

paper-api/src/main/java/org/bukkit/Color.java
```diff
@@ -146,7 +146,7 @@ public static Color fromRGB(int red, int green, int blue) throws IllegalArgument
      * @param blue integer from 0-255
      * @param green integer from 0-255
      * @param red integer from 0-255
-     * @return a new Color object for the red, green, blue
+     * @return a new Color object for the blue, green, red
```
FireInstall, line 149: I mean it's fine, but why would you change the listing from rgb to bgr?

lynxplay: because this is the javadoc for the fromBRG method?

paper-api/src/main/java/org/bukkit/Color.java
lynxplay: I am semi inclined to agree here.
I guess the intention is that "comparing a value to a mask makes no sense".

The more correct precondition I guess would be `(alpha & BITMASK) == alpha`.

paper-api/src/test/java/org/bukkit/ColorTest.java
```diff
@@ -32,7 +31,8 @@ static class TestColor {
             this.r = r;
             this.g = g;
             this.b = b;
-            this.name = id + ":" + Integer.toHexString(argb).toUpperCase(Locale.ROOT) + "_" + Integer.toHexString(rgb).toUpperCase(Locale.ROOT) + "_" + Integer.toHexString(bgr).toUpperCase(Locale.ROOT) + "-a" + Integer.toHexString(a).toUpperCase(Locale.ROOT) + "-r" + Integer.toHexString(r).toUpperCase(Locale.ROOT) + "-g" + Integer.toHexString(g).toUpperCase(Locale.ROOT) + "-b" + Integer.toHexString(b).toUpperCase(Locale.ROOT);
+            this.name = id + ":" + Integer.toHexString(argb).toUpperCase(Locale.ROOT) + "_" + Integer.toHexString(bgr).toUpperCase(Locale.ROOT) + "-a" + Integer.toHexString(a).toUpperCase(Locale.ROOT) + "-r" + Integer.toHexString(r).toUpperCase(Locale.ROOT) + "-g" + Integer.toHexString(g).toUpperCase(Locale.ROOT) + "-b" + Integer.toHexString(b).toUpperCase(Locale.ROOT);
```
lynxplay, line 37: If this is a cleanup PR, this should be `id++ + ":" + `

PR #12173, opened 2025-02-24
Added a config option to enable/disable sending transfer packets

Biquaternions:
Transferring a player will require both sender and receiver to have transfers enabled in paper-global.yml and server.properties respectively.

This increases security as server staff with elevated permissions can send players to self-hosted servers to ip-grab them without having console access in the "sender" server.
This also breaks attack vectors where server staff is tricked into transferring players directly via command or clicking a sign.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/network/ServerCommonPacketListenerImpl.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/network/ServerCommonPacketListenerImpl.java.patch
@@ -230,7 +230,7 @@
              }
  
              return false;
-@@ -149,6 +_,13 @@
+@@ -149,6 +_,15 @@
      }
  
      public void send(Packet<?> packet, @Nullable PacketSendListener listener) {
@@ -239,7 +239,9 @@
 +            return;
 +        } else if (packet instanceof net.minecraft.network.protocol.game.ClientboundSetDefaultSpawnPositionPacket defaultSpawnPositionPacket) {
 +            this.player.compassTarget = org.bukkit.craftbukkit.util.CraftLocation.toBukkit(defaultSpawnPositionPacket.getPos(), this.getCraftPlayer().getWorld());
-+        }
++        } else if (packet instanceof net.minecraft.network.protocol.common.ClientboundTransferPacket && !io.papermc.paper.configuration.GlobalConfiguration.get().misc.allowServerTransfers) { // Paper start - check if server allows sending transfer packets
++            return;
++        } // Paper end - check if server allows sending transfer packets
 +        // CraftBukkit end
          if (packet.isTerminal()) {
              this.close();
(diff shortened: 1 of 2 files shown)


lynxplay:
I am not a large fan of this change.
Paper already assigns permission nodes to all vanilla commands.
Servers concerned with this should just be removing the permission for the transfer command `minecraft.command.transfer` from everyone.

Has the same effect without the need for a new server option.

Biquaternions:
I thought of something similar while reviewing the code so I made the implementation on the send packet function to also prevent plugins from doing this using packets.
So... it is not limited to just the `/transfer` command.

lynxplay:
I am confused, why would a plugin transfer players if the server owner is not interested in transfers?

lynxplay:
To follow up on this, the transfer command is not something *normal* players have access to.

A server would have to effectively grant operator permissions to someone for them to be able to use it or explicitly the transfer permissions. That is already incredibly dangerous, transferring seems like the *least* of your worries if someone has such level of permissions.

The angle of a plugin transferring has two parts.
Either the plugin maliciously does it, at which point, you have malware.
Or the plugin exposes a transfer mechanism and somehow a staff member not meant to have that permission gained that? Which is pretty much the same point as above.

I just really don't see an angle where a server might have *accidental* transfers enabled without it being aggressively missconfigured, which I doubt those servers will use a config option to fix that.

I'll throw it at some other members for their input, but from my point of view, I don't think I want to pull this change.

Biquaternions:
Sometimes people upload malicious plugins to official platforms like it happened in modrinth a couple of months ago.
Other times server owners get custom plugins from other people like "friends" or worse, players.

Tbh, that last one is mostly their fault 😅
But still is something that sometimes happens, specially for non experienced server owners, and the ones at risk ends up being the players

Biquaternions:
No problems, you can close it.
Thx for the compliment!

lynxplay:
Yea malware certainly exist :sweat_smile: Issue with that is usually "ah fuck, malware might as well just change the parsed config value before sending this". Fighting malware outside of "do not install it in the first place" is sadly usually a lost cause.

I'll throw the PR at someone else before I close it just to make sure the team agrees with me on this. But yea, thank you already for the PR no matter the outcome!

masmc05:
> Sometimes people upload malicious plugins to official platforms like it happened in modrinth a couple of months ago.

Tbh I can't see a reason why would someone transfer people instead of silently grab the exact same information on your server and send it back to the bad dev. People will quickly start to notice those transfers and complain, exposing the plugin, all this for no extra information that a transferred server could extract.

Biquaternions:
Btw, this is the first PR I make to Paper, so I'm not yet aware of your workflow if usually you close the PRs or I should be closing it.

electronicboy:
This implementation is far too trivial to bypass by anybody remotely familiar with the networking library in use, and wouldn't take much rocket science to see how to bypass the method here. The only way you'd be able to do this is to handle this in the pipeline itself, but, this generally just offers a fake security assurance which nobody can promise when you're running arbitrary code on your server.

PR #12181, opened 2025-02-25
Make minecarts respect frictionState in water

brickmonster:
Water friction is just a type of friction, like air and ground friction.

This more closely matches my expectation of "this shouldn't slow down naturally over time without external forces".

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/vehicle/AbstractMinecart.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/vehicle/AbstractMinecart.java.patch
@@ -96,6 +96,14 @@
          }
      }
  
+@@ -451,6 +_,7 @@
+     }
+ 
+     protected Vec3 applyNaturalSlowdown(Vec3 speed) {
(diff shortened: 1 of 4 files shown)


Lulu13022002:
It seems this config always only applied to air even in the original PR that was mentioned (and in the javadocs of Frictional), if something like that is done it should be the same for all entities if possible and the javadocs changed

brickmonster:
Minecarts don't care about `frictionState` if they're not on a rail, so they're already treated differently than other entities.

I wouldn't bother with this change if they didn't make minecarts work on rails underwater.

Currently: 
* minecarts on rails in air respect `frictionState`,
* minecarts on rails in water don't respect `frictionState`.
* minecarts not on rails don't respect `frictionState`.

I think minecarts should respect `frictionState` on rails regardless of water so map makers can set them off at a specific speed without worrying about them being treated differently in an underwater section of the railway.

It's a unique situation to minecarts. I don't think other entities not ignoring water needs to be considered unless this change evolves into adding a water friction flag.

PR #12182, opened 2025-02-25
Add a method on Registry to get the size

GliczDev:


GliczDev:
I wasn't sure what to do about `ADVANCEMENT`, `BOSS_BARS` and `MEMORY_MODULE_TYPE` registries.
Any ideas, or is my implementation good enough?

paper-api/src/main/java/org/bukkit/Registry.java
```diff
@@ -138,6 +144,11 @@ public Iterator<Advancement> iterator() {
             return Bukkit.getBossBar(key);
         }
 
+        @Override
+        public int size() {
```
lynxplay: This could probably just be a default method.

Like, by default we use Iterables.size and then override this for delegate, map based and internal reg.
It is 100% fine for the deprecated registries to be a bit slower like this.

GliczDev: wouldn't it be better to just make NotARegistry use Iterables.size by default since the issue only affects them?

lynxplay: Or that yea :+1:

kennytv (approved):

PR #12183, opened 2025-02-25
Update custom tags

Lulu13022002:
Slowly migrate custom tags to vanilla tags by deprecating the redundant tag, the other tags should probably be generated using the data generator later


Owen1212055 (dismissed):

Owen1212055:
(Although this review isn't really helpful, at least currently, these look good, although I am a bit worried about the ambiguity of some of these tags... like the wooden fence gates -> fence gates)

paper-api/src/main/java/com/destroystokyo/paper/MaterialSetTag.java
```diff
@@ -93,8 +92,4 @@ public boolean isTagged(@NotNull Block block) {
     public boolean isTagged(@NotNull ItemStack item) {
         return isTagged(item.getType());
     }
-
-    public boolean isTagged(@NotNull Material material) {
-        return this.tagged.contains(material);
-    }
```
Machine-Maker: Why remove this method? It will break ABI for no reason.

lynxplay (approved):
LGTM, will come with the idea that, once we move the rest of the code-defined tags into normal json defined tags in the paper namespace, we guarantee backwards compatibility for our own custom tags and *presumably* also the vanilla tags we deem worthy of keeping.

lynxplay:
(Also, I hope no one is going to miss the wall variants of all the skulls in head equippable but blocks tend to be hard to put on heads so, fair change)

Machine-Maker (approved):

PR #12190, opened 2025-02-26
Fix invulnerability damage and armour

okx-code:
If a plugin changes the damage in `EntityDamageEvent`, it can cause the `INVULNERABILITY_REDUCTION` damage modifier to be set to 0 and render invulnerability damage reduction, added in https://github.com/PaperMC/Paper/pull/11599, ineffective.

This happens because a float (the `lastHurt` value in a Player) is being compared to a double (stored in the modifiers of EntityDamageEvent), and a float can be greater than a double, even if they are the same noiminal value:

```
jshell> 3.2 < (float) 3.2
$2 ==> true
```

Due to lossy floating point precision, 3.2 becomes 3.200000047683716, which indeed would be greater than 3.2.

`float lastHurt` is the result of casting the double from `EntityDamageEvent#getFinalDamage`, which causes this imprecision. The fix is to cast the doubles from `EntityDamageEvent` to floats when comparing them here. This ensures that if they are the same value, then one will not be less than the other, because they have both been cast from the same `double` value. The opposite approach, casting a `double` to a `float` and back to a `double` again, does not work, because is lossy. We cannot really convert `lastHurt` to a `double` without touching a lot of the damage processing code.

I have also taken the liberty of fixing how armour damage is calculated. In the #11599 PR, armour would still take damage even if the damage should have been blocked due to invulnerability. This change will now take that into account. It's also necessary to individual cast each component to a float, rather than adding the doubles like before, otherwise the armourDamage may be marginally greater than 0. I've done this in a similar way to the `LivingEntity#computeAmountFromEntityDamageEvent` function.

The BASE damage and INVULNERABILITY_DAMAGE modifers may be slightly different doubles, but represent the same float. The INVULNERABILITY_DAMAGE represents the BASE modifier, but casted to a float and back to a double. Therefore we must cast BASE and INVULNERABILITY_DAMAGE to floats before adding them.

To test this PR, you can use the following code and run the `/test` command (plugin.yml not included)

```java
public class TestPlugin extends JavaPlugin implements Listener, CommandExecutor {

    @Override
    public void onEnable() {
        getCommand("test").setExecutor(this);
    }

    @Override
    public boolean onCommand(CommandSender sender, Command command, String label, String[] args) {
        ServerPlayer handle = ((CraftPlayer) sender).getHandle();
        handle.hurtServer(handle.serverLevel(), handle.level().damageSources().cactus(), 4);
        handle.hurtServer(handle.serverLevel(), handle.level().damageSources().cactus(), 4);
        return true;
    }

    @EventHandler
    public void on(EntityDamageEvent event) {
        getLogger().info("Invulnerability reduction before: " + event.getDamage(EntityDamageEvent.DamageModifier.INVULNERABILITY_REDUCTION));
        event.setDamage(3.2);
        getLogger().info("Invulnerability reduction after: " + event.getDamage(EntityDamageEvent.DamageModifier.INVULNERABILITY_REDUCTION));
    }
}
```

Without this fix:

```
[TestPlugin] Invulnerability reduction before: 0.0
[TestPlugin] Invulnerability reduction after: 0.0
[TestPlugin] Invulnerability reduction before: -3.200000047683716
[TestPlugin] Invulnerability reduction after: 0.0
```

With this fix:

```
[TestPlugin] Invulnerability reduction before: 0.0
[TestPlugin] Invulnerability reduction after: 0.0
[TestPlugin] Invulnerability reduction before: -3.200000047683716
[TestPlugin] Invulnerability reduction after: -3.200000047683716
```

cc @lynxplay

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
@@ -971,7 +971,7 @@
                  int i = (this.getEffect(MobEffects.DAMAGE_RESISTANCE).getAmplifier() + 1) * 5;
                  int i1 = 25 - i;
                  float f = damageAmount * i1;
-@@ -1768,24 +_,217 @@
+@@ -1768,24 +_,219 @@
          }
      }
  
@@ -984,6 +984,10 @@
 +            if (invulnerabilityRelatedLastDamage == 0) return 0D; // no last damage, no reduction
 +            // last damage existed, this means the reduction *technically* is (new damage - last damage).
 +            // If the event damage was changed to something less than invul damage, hard lock it at 0.
++            //
++            // Cast the passed in double down to a float as double -> float -> double is lossly.
++            // If last damage is a (float) 3.2D (given spigot events use doubles), we cannot compare
++            // the new damage value of this damage instance by upcasting it again to a double as 3.2D != (double) (float) 3.2D.
 +            if (d.floatValue() < invulnerabilityRelatedLastDamage) return 0D;
 +            return (double) -invulnerabilityRelatedLastDamage;
 +        };
@@ -1118,12 +1122,10 @@
 +
 +            // Apply damage to armor
 +            if (!damageSource.is(DamageTypeTags.BYPASSES_ARMOR)) {
-+                // Paper start - fix invulnerability reduction in EntityDamageEvent
 +                float armorDamage = (float) event.getDamage();
 +                armorDamage += (float) event.getDamage(DamageModifier.INVULNERABILITY_REDUCTION);
 +                armorDamage += (float) event.getDamage(DamageModifier.BLOCKING);
 +                armorDamage += (float) event.getDamage(DamageModifier.HARD_HAT);
-+                // Paper end - fix invulnerability reduction in EntityDamageEvent
 +                this.hurtArmor(damageSource, armorDamage);
 +            }
 +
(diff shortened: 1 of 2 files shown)


lynxplay (approved):

paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -984,7 +984,11 @@
 +            if (invulnerabilityRelatedLastDamage == 0) return 0D; // no last damage, no reduction
 +            // last damage existed, this means the reduction *technically* is (new damage - last damage).
 +            // If the event damage was changed to something less than invul damage, hard lock it at 0.
-+            if (d < invulnerabilityRelatedLastDamage) return 0D;
++            //
++            // Cast the passed in double down to a float as double -> float -> double is lossly.
++            // If last damage is a (float) 3.2D (given spigot events use doubles), we cannot compare
```
kennytv:
```suggestion
+            // Cast the passed in double down to a float as double -> float -> double is lossy.
+            // If last damage is a (float) 3.2D (since the events use doubles), we cannot compare
```

kennytv (approved):

paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -1118,7 +1122,10 @@
 +
 +            // Apply damage to armor
 +            if (!damageSource.is(DamageTypeTags.BYPASSES_ARMOR)) {
-+                float armorDamage = (float) (event.getDamage() + event.getDamage(DamageModifier.BLOCKING) + event.getDamage(DamageModifier.HARD_HAT));
++                float armorDamage = (float) event.getDamage();
```
Lulu13022002, line 1125: This computation is flawed anyway since it doesn't take in account the recent damage modifier added by Spigot (freezing) accounted in Vanilla. So the armor generally take less damage in some environment.

For example with a datapack that clear the #bypasses_armor damage type tag and add the player to the #freeze_hurts_extra_types entity type tag.

PR #12193, opened 2025-02-26
Add a method on InventoryView to get the MenuType

GliczDev:
It is possible to get the `InventoryType`, but not `MenuType`.
Since there is a new (better) way to create views for players using `MenuType`, it would be nice to also be able to get it back from `InventoryView` after creating.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
@@ -321,8 +321,8 @@ public interface InventoryView {
     /**
      * Gets the menu type of the inventory view.
      *
-     * @return the menu type of the inventory view
+     * @return the menu type of the inventory view or null if not applicable
      */
     @ApiStatus.Experimental
-    @NotNull MenuType getMenuType();
+    @Nullable MenuType getMenuType();
 }
--- a/paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
@@ -319,7 +319,11 @@ public interface InventoryView {
     public void setTitle(@NotNull String title);
 
     /**
-     * Gets the menu type of the inventory view.
+     * Gets the menu type of the inventory view if applicable.
+     * <p>
+     * Some inventory types do not support a menu type. In such cases, this method
+     * returns null. This typically applies to inventories belonging to entities
+     * like players or animals (e.g., a horse).
      *
      * @return the menu type of the inventory view or null if not applicable
      */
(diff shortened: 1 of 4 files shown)


paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
```diff
@@ -316,4 +317,12 @@ default net.kyori.adventure.text.Component title() {
...
     @Deprecated(since = "1.21.1") // Paper
     public void setTitle(@NotNull String title);
+
+    /**
+     * Gets the menu type of the inventory view.
+     *
+     * @return the menu type of the inventory view
+     */
+    @ApiStatus.Experimental
+    @NotNull MenuType getMenuType();
```
Y2Kwastaken: Given the changes required to make this API work effectively, this should be marked as Nullable.

Y2Kwastaken (changes requested):
In order to ensure the API functions properly, some adjustments are needed. I’ve tested it, and currently, the API throws an error for any menu that doesn’t have a MenuType. This is an issue in scenarios like the player’s inventory or with animals, where a MenuType is not applicable. In these cases, it results in an UnsupportedOperationException. If you’re aiming for this to be merged, I’d recommend modifying the behavior to return null instead of throwing the exception. Otherwise, everything else seems fine. I'm not in love with the CraftContainer impl, but I think its pretty much unavoidable at this point sadly.

GliczDev:
> In these cases, it results in an UnsupportedOperationException.
@Y2Kwastaken does it also throw in `CraftContainer` implementation? It seems that `CraftContainer.getNotchInventoryType` never returns null, so there should be no issues, but I want to verify it.

Y2Kwastaken:
> > In these cases, it results in an UnsupportedOperationException.
> 
> @Y2Kwastaken does it also throw in `CraftContainer` implementation? It seems that `CraftContainer.getNotchInventoryType` never returns null, so there should be no issues, but I want to verify it.

The naive conversion for Inventory's is fine and won't throw that exception to see why its thrown see AbstractContainerMenu#getType

paper-api/src/main/java/org/bukkit/inventory/InventoryView.java
```diff
@@ -321,8 +321,8 @@ default net.kyori.adventure.text.Component title() {
     /**
      * Gets the menu type of the inventory view.
      *
-     * @return the menu type of the inventory view
+     * @return the menu type of the inventory view or null if not applicable
```
Y2Kwastaken, line 328: rather than just stating null if not applicable might be smart to list the cases where null is returned here as far as I'm aware this is the case only player and animal menus, but it might be worth poking around a bit more to make sure that I'm 100% right about this.  Might just be nice to have a little paragraph not in the return statement being more involved on why this method might return null

Y2Kwastaken (approved):
This is a lot better thank you, From my testing it works as you'd expect.

Warriorrrr (approved):

kennytv (approved):

Owen1212055 (approved):

PR #12194, opened 2025-02-26
Fix cancelled PotionSplashEvent putting out fires

### LINKED ISSUE #11981: PotionSplashEvent doesn't prevent putting out fires
### Expected behavior

If you cancel the PotionSplashEvent, it should prevent putting out fires with water potions

### Observed/Actual behavior

Fires are always put out regardless of cancellation

### Steps/models to reproduce

1. Throw water splash potion (or lingering) at a fire
2. Cancel potion splash event
3. Fire goes out

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.21.1-132-ver/1.21.1@b48403b (2024-11-21T10:14:27Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
Previous version: 1.21.1-DEV-164e116 (MC: 1.21.1)

### Other

_No response_

notTamion:
fixes #11981
i think we should prevent this from happening as to me it seems the only reason this wasn't put where all the other water splash potion logic is is that the onHitBlock method was just convenient for mojang

EDIT: this also provides us the opportunity to expand the WaterBottleSplashEvent, as waterbottles putting out fires is obviously a big part of their behavior.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/projectile/ThrownPotion.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/projectile/ThrownPotion.java.patch
@@ -1,15 +1,16 @@
 --- a/net/minecraft/world/entity/projectile/ThrownPotion.java
 +++ b/net/minecraft/world/entity/projectile/ThrownPotion.java
-@@ -59,8 +_,7 @@
+@@ -59,9 +_,7 @@
          return 0.05;
      }
  
 -    @Override
 -    protected void onHitBlock(BlockHitResult result) {
+-        super.onHitBlock(result);
 +    protected void extinguishFires(BlockHitResult result) { // Paper - move after PotionSplashEvent call to prevent putting out fires for cancelled events
-         super.onHitBlock(result);
          if (!this.level().isClientSide) {
              ItemStack item = this.getItem();
+             Direction direction = result.getDirection();
 @@ -82,51 +_,90 @@
      @Override
      protected void onHit(HitResult result) {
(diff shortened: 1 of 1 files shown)


paper-server/patches/sources/net/minecraft/world/entity/projectile/ThrownPotion.java.patch
```diff
@@ -1,6 +1,16 @@
...
 +++ b/net/minecraft/world/entity/projectile/ThrownPotion.java
-@@ -82,51 +_,86 @@
+@@ -59,8 +_,7 @@
+         return 0.05;
+     }
+ 
+-    @Override
+-    protected void onHitBlock(BlockHitResult result) {
++    protected void extinguishFires(BlockHitResult result) { // Paper - move after PotionSplashEvent call to prevent putting out fires for cancelled events
+         super.onHitBlock(result);
```
Lulu13022002: I'm not sure about calling the super method here. It means the super method is called twice in a normal launch so two TargetHitEvent for example.

paper-server/patches/sources/net/minecraft/world/entity/projectile/ThrownPotion.java.patch
```diff
@@ -64,18 +74,27 @@
...
++                }
++            }
++        }
++
 +        // Paper start - Fix potions splash events
 +        io.papermc.paper.event.entity.WaterBottleSplashEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callWaterBottleSplashEvent(
 +            this, result, affected, rehydrate, extinguish
 +        );
 +        if (!event.isCancelled()) {
++            if (result.getType() == HitResult.Type.BLOCK) {
```
Lulu13022002: This will throw a NPE from the api since result is nullable

Lulu13022002:
It also feels weird to only cancel interaction with fire but still handle redstone thing and everything called by Block#onProjectileHit. So maybe the responsibily should be moved to the ProjectileHitEvent.

PR #12197, opened 2025-02-27
Infer block entity data in brigadier blockstate argument

### LINKED ISSUE #11700: BlockState#copy() does not infer block entity
### Expected behavior

```java
BlockState copy = blockState.copy(location);
copy.update(true, true)
```
This should set the block type and copy over all block entity data like inventories, bee data, etc.

### Observed/Actual behavior

Snippet above only properly sets the block type and does not copy over the block entity data. Throws following warning `Skipping BlockEntity with id`.

### Steps/models to reproduce

Run this snippet with any valid input as `blockState` and `location`
```java
BlockState copy = blockState.copy(location);
copy.update(true, true)
```

### Plugin and Datapack List

Issue occurs with only the plugin I'm testing with

Server Plugins (1):
Paper Plugins:
BlockEdit

Datapacks:

### Paper version

This server is running Paper version 1.21.3-77-master@1e035f3 (2024-12-01T15:52:35Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)

### Other

_No response_

The-Epic:
Applies id to blockstate tag when not provided, fixes #11700

### DIFF
--- a/paper-server/src/main/java/io/papermc/paper/command/brigadier/argument/VanillaArgumentProviderImpl.java
+++ b/paper-server/src/main/java/io/papermc/paper/command/brigadier/argument/VanillaArgumentProviderImpl.java
@@ -74,8 +74,6 @@ import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.level.Level;
-import net.minecraft.world.level.block.EntityBlock;
-import net.minecraft.world.level.block.entity.BlockEntity;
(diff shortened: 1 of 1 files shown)


Lulu13022002:
This is absolutely not the right fix and you seem to be confused about the block id with the block entity id (its block entity type key)

Lulu13022002:
This now throw a NPE when the tag is not present for block entities

The-Epic:
Missed that during testing, will resolve soon

Lulu13022002:
The NPE is still here, it's not the block input that can be null but the tag

Lulu13022002 (approved):

PR #12199, opened 2025-02-27
Add client tick end event

### LINKED ISSUE #12198: Add a ClientTickEndEvent

Mojang added a new packet in 1.21.3, `ServerboundClientTickEndPacket`, to let the server know when a client-sided tick ends. Some features, namely display entity interpolation, are wholly dependent on the client's tick loop, not the server's. Currently, it is impossible for a plugin to tell if it is safe to perform transformations on a display entity and having the guarantee that it will be interpolated expectedly.

A `ClientTickEndEvent` that is called when the server receives the aforementioned `ServerboundClientTickEndPacket`.

Delaying applying transformations to a display entity for an arbitrary amount of time (one server tick might not be enough to guarantee the client will interpolate the changes) (yuck), or using internals to listen to the packet being received (yucker).

### Other

I recognise the event would be **very** hot, with potentially every player sending the packet, firing the event at every tick. This shouldn't be a reason to not consider it, as we have a good handful of very hot events already that are fine to listen to when the listener's code takes into consideration such condition. A number one step would be to not fire the event if there are no registered listeners. And given the simplicity of this event (all the state it would need is the `Player`), it could be constructed once and cached into the `ServerPlayer` instance. I also acknowledge that a modified client (or an old client connected through ViaVersion or similar) could/would simply not send the packet, but that is none of the server's concern; VV could in theory emulate the packet by sending it to the server on the server's tick end, and if a client is modified to not send it, that sucks for them. But those are details beyond the scope of this feature request. Owen and pop want this.

Warriorrrr:
Closes #12198

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
@@ -70,7 +70,7 @@
      private double firstGoodX;
      private double firstGoodY;
      private double firstGoodZ;
-@@ -236,22 +_,39 @@
+@@ -236,22 +_,41 @@
      private int receivedMovePacketCount;
      private int knownMovePacketCount;
      private boolean receivedMovementThisTick;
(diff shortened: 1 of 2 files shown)


kennytv (approved):

NoahvdAa (approved):

paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -2620,6 +2622,11 @@
  
      @Override
      public void handleClientTickEnd(ServerboundClientTickEndPacket packet) {
+         PacketUtils.ensureRunningOnSameThread(packet, this, this.player.serverLevel());
++        this.tickEndEvent.callEvent(); // Paper - add client tick end event
```
Owen1212055, line 2626: Please check that the handler list is > 0 before executing this. 
Just cause of how often this will be running.

Warriorrrr: I thought about that, but since we're re-using the event instance checking the handler list size wouldn't really avoid anything, one thread check at most

Machine-Maker (approved):

electronicboy:
PME is mutable, this is not

PR #12204, opened 2025-02-28
Add some missing annotations and an incorrect one

brickmonster:
Specifically used jspecify for arrays since paper is moving torwards it and either javax or jetbrains handle arrays differently.

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
@@ -42,7 +42,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
+import net.kyori.adventure.util.TriState;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.minecraft.advancements.AdvancementProgress;
 import net.minecraft.commands.Commands;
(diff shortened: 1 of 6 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -309,7 +309,7 @@ public interface TransferCookieConnection {
         void kickPlayer(Component reason, org.bukkit.event.player.PlayerKickEvent.Cause cause); // Paper - kick event causes
     }
 
-    public record CookieFuture(ResourceLocation key, CompletableFuture<byte[]> future) {
+    public record CookieFuture(ResourceLocation key, CompletableFuture<byte @org.jspecify.annotations.Nullable []> future) {
```
Warriorrrr: Switching out the import for jspecify should be fine

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftSign.java
```diff
@@ -231,16 +231,4 @@ public static Component[] sanitizeLines(String[] lines) {
...
-        String[] lines = new String[components.length];
-        for (int i = 0; i < lines.length; i++) {
-            lines[i] = CraftSign.revertComponent(components[i]);
-        }
-        return lines;
-    }
-
-    private static String revertComponent(Component component) {
-        return CraftChatMessage.fromComponent(component);
-    }
```
Doc94: ?

brickmonster: Didn't think the dead code was worth another PR.

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
brickmonster: You're right, the others in CraftPlayer are all "boring" annotations.

Warriorrrr (approved):

kennytv (approved):

Owen1212055 (approved):

PR #12205, opened 2025-02-28
Skip alpha channel for custom model data component

Lulu13022002:


kennytv (approved):

Warriorrrr (approved):

Owen1212055 (approved):
In the future, I think some kind of utility method should be made because there seems to be a lot of alpha related stripping everywhere.

PR #12206, opened 2025-02-28
Call BlockRedstoneEvent for more interactions

Lulu13022002:
Handle more case that would change the powered/power property once a block is placed:
- skull, copper bulb, bell, target block, note block
- wind charge on button/lever

Fixes some other problems:
- lighting strike on lightning rod only called the event for the transition 0 -> 15 and not 15 -> 0
- command block called the event even for non redstone change (for example when placed/already powered)
- detector rail called the event with wrong arguments (oldCurrent always equals to newCurrent)
- trapdoor didn't call the event when already opened then powered
- door and trapdoor had the real power received instead of its two states (0/15) unlike similar blocks (lever/fence gate)
- iron/gold pressure plates power change was not accounted properly

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/block/TrapDoorBlock.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/block/TrapDoorBlock.java.patch
@@ -1,21 +1,15 @@
 --- a/net/minecraft/world/level/block/TrapDoorBlock.java
 +++ b/net/minecraft/world/level/block/TrapDoorBlock.java
-@@ -133,7 +_,37 @@
+@@ -133,7 +_,31 @@
          if (!level.isClientSide()) {
              boolean signal = level.hasNeighborSignal(pos);
              if (signal != state.getValue(POWERED)) {
 -                if (state.getValue(OPEN) != signal) {
-+                // CraftBukkit start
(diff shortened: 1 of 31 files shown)


paper-server/patches/sources/net/minecraft/world/level/block/TrapDoorBlock.java.patch
```diff
@@ -1,27 +1,18 @@
 --- a/net/minecraft/world/level/block/TrapDoorBlock.java
 +++ b/net/minecraft/world/level/block/TrapDoorBlock.java
-@@ -146,7 +_,40 @@
+@@ -146,7 +_,31 @@
          if (!level.isClientSide) {
              boolean hasNeighborSignal = level.hasNeighborSignal(pos);
              if (hasNeighborSignal != state.getValue(POWERED)) {
 -                if (state.getValue(OPEN) != hasNeighborSignal) {
-+                // if (state.getValue(OPEN) != hasNeighborSignal) {
 +                // CraftBukkit start
```
Warriorrrr: Can be swapped out for paper, since all the cb stuff has been replaced

Warriorrrr (approved):

kennytv (approved):

Owen1212055 (approved):

Owen1212055 (approved):

PR #12209, opened 2025-03-01
Remove ItemFactory#enchantWithLevels range check for vanilla parity

### LINKED ISSUE #12208: ItemStack#enchantWithLevels is arbitrarily capped at 30

ItemStack#enchantWithLevels should allow using any level to enchant with (though any level lower than 0 would effectively do nothing). The vanilla logic already allows this, there's just a level cap at 30 for some reason. Vanilla loot tables already use level 39, as seen in the end city treasure table. 

ItemStack#enchantWithLevels and the ItemFactory equivalents should accept levels up to max int, as there is really no vanilla restriction to the method. The minimum value of 1 still makes sense, as calling it with 0 or lower would do nothing. 

ItemFactory#enchantItem doesn't have either a minimum or a maximum level cap and seems to fulfill pretty much the same purpose, but is deprecated. 

### Other

_No response_

Esophose:
Resolves #12208 by removing the level range check on ItemFactory#enchantWithLevels. The old ItemFactory#enchantItem did not have a level range, and the vanilla method does not have a level range either. There isn't really any reason to have this. If desired, it could be capped with a lower bound of 1, but it doesn't seem necessary.

Vanilla has at least one loot table that uses a level above 30, see the end city treasure table.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/inventory/ItemFactory.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/ItemFactory.java
@@ -327,6 +327,10 @@ public interface ItemFactory {
      *
      * <p>If the provided ItemStack is already enchanted, the existing enchants will be removed before enchanting.</p>
      *
+     * <p>Enchantment tables use levels in the range {@code [1, 30]}.</p>
+     *
+     * <p>Non-positive or too high values of {@code levels} will result in no enchantments being applied.</p>
+     *
      * @param itemStack ItemStack to enchant
      * @param levels levels to use for enchanting
      * @param allowTreasure whether to allow enchantments where {@link org.bukkit.enchantments.Enchantment#isTreasure()} returns true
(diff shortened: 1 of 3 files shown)


Doc94:
I dont check but this was not based in the enchantment table? Based in the range of 1-30 also this PR now allow you pass 0 or negative levels (not sure how that works)

Esophose:
It was based on the enchantment table, but arbitrarily so, there is no such lower bound or upper bound on the actual method being called. 0 or negative levels gracefully do nothing.

TonytheMacaroni:
It's also useful to note the internal method `ItemFactory#enchantWithLevels` relies on (`EnchantmentHelper#enchantItem`) is also used for the `minecraft:enchant_with_levels` item modifier, which doesn't have a level restriction. I.e. it using a range of `[20, 39]` levels for the `minecraft:chests/end_city_treasure` loot table, as mentioned in the PR's body.

Warriorrrr (commented):
I think that it's worth noting that you can only use levels 1-30 in a normal enchanting table, and that at higher levels (>= 100) you will not be able to get any enchants in vanilla.

Warriorrrr (approved):

Owen1212055 (approved):

Machine-Maker (approved):

PR #12215, opened 2025-03-02
Fix cancelling PlayerInteractEvent at (0, 0, 0)

### LINKED ISSUE #12196: Cancelling PlayerInteractEvent does not prevent breaking blocks at 0, 0, 0
### Expected behavior

blocks at 0 0 0 will not break

### Observed/Actual behavior

blocks at 0 0 0 break, they do not drop item

### Steps/models to reproduce

1. place a block at 0 0 0
2. cancell PlayerInteractEvent using plugin: `@EventHandler
    public void PlayerInteractEvent(PlayerInteractEvent event) {
            event.setCancelled(true);
    }`
3. go in survival mode and try to break block

### Plugin and Datapack List

testplugin (what I used for cancelling block event)

### Paper version

1.21.4-185-main@0a6e743 (2025-02-26T13:10:43Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

NoahvdAa:
When the PlayerInteractEvent gets canceled, the destroyPos field won't be updated, meaning the block won't get broken because the position doesn't match in the check at ServerPlayerGamemode line 261. However, the default value for destroyPos is BlockPos.ZERO.

This leads to the condition returning to true if the player has not mined any blocks before, even if the interactevent was denied. The issue also happens if someone is allowed to break a block the first time, but is denied the second time, since the destroyPos will stay the same

 This commit fixes #12196 by making the destroyPos nullable instead of using BlockPos.ZERO as a null value.

### DIFF
--- a/paper-server/patches/features/0004-Anti-Xray.patch
+++ b/paper-server/patches/features/0004-Anti-Xray.patch
@@ -145,7 +145,7 @@ index 3a384175f8e7f204234bbaf3081bdc20c47a0d4b..5699bc15eba92e22433a20cb8326b59f
  
      private ClientboundLevelChunkWithLightPacket(RegistryFriendlyByteBuf buffer) {
 diff --git a/net/minecraft/server/level/ServerLevel.java b/net/minecraft/server/level/ServerLevel.java
-index 40b799fd90b0db13bdaa8834c021f5ca8f25ce10..400b56657414177cd76a7b94c426dc7c886aa957 100644
+index da8848e2a3e3745949eb2356a049451d30f763a7..192977dd661ee795ada13db895db770293e9b402 100644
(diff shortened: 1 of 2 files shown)


NoahvdAa:
I'm not super familiar with how the breaking system works, so I'm not sure if this has an unintended side effect somewhere down the chain (did check all reads of destroyPos ofc). I did a fair bit of testing though and couldn't get it to break...

Owen1212055 (changes requested):
Couldn't we instead check if this.destroyPos != BlockPos.ZERO instead?

Owen1212055 (approved):

Machine-Maker (approved):

PR #12216, opened 2025-03-02
Fix ItemStack amount issues with Chat Components

### LINKED ISSUE #12213: Failed to encode Minecraft Component on displayName()
### Stack trace

```
java.lang.RuntimeException: Failed to encode Minecraft Component: translation{key='chat.square_brackets', args=[empty[style={italic}, siblings=[io.papermc.paper.adventure.AdventureComponent@d7c3825c]]]}[style={color=aqua,hoverEvent=TypedHoverEvent[action=<action show_item>, value=net.minecraft.network.chat.HoverEvent$ItemStackInfo@9f802d36]}]; Value must be within range [1;99]: 169
	at io.papermc.paper.adventure.WrapperAwareSerializer.lambda$deserialize$0(WrapperAwareSerializer.java:28) ~[paper-1.21.4.jar:1.21.4-DEV-1d5e5a5]
	at com.mojang.serialization.DataResult$Error.getOrThrow(DataResult.java:287) ~[datafixerupper-8.0.16.jar:?]
	at io.papermc.paper.adventure.WrapperAwareSerializer.deserialize(WrapperAwareSerializer.java:28) ~[paper-1.21.4.jar:1.21.4-DEV-1d5e5a5]
	at io.papermc.paper.adventure.WrapperAwareSerializer.deserialize(WrapperAwareSerializer.java:13) ~[paper-1.21.4.jar:1.21.4-DEV-1d5e5a5]
	at io.papermc.paper.adventure.PaperAdventure.asAdventure(PaperAdventure.java:192) ~[paper-1.21.4.jar:1.21.4-DEV-1d5e5a5]
	at org.bukkit.craftbukkit.inventory.CraftItemFactory.displayName(CraftItemFactory.java:229) ~[paper-1.21.4.jar:1.21.4-DEV-1d5e5a5]
	at org.bukkit.inventory.ItemStack.displayName(ItemStack.java:728) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
```

### Plugin and Datapack List

My plugin

Create itemstack with more than 99 amount, undefined behaviour thereafter

### Paper version

This server is running Paper version 1.21.4-DEV-main@1d5e5a5 (1970-01-01T00:00:00Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-DEV-f63dbea (MC: 1.21.4)

### Other

_No response_

Doc94:
Close #12213 by handle two cases where this can happen.
- You try pass the displayname to a component for chat
```java
ItemStack itemStack = new ItemStack(Material.STICK, 200);
player.sendMessage(Component.text("Item1").appendSpace().append(itemStack.displayName()));
```
- You try to pass an item for the hover event
```java
ItemStack itemStack = new ItemStack(Material.STICK, 200);
player.sendMessage(Component.text("Item2").hoverEvent(itemStack));
```

for the first case the only "easy" way is just mention in javadocs about how use this for chat messages can cause issues because mess with precontitions breaks others "normal" behaviours, the second case is more easy add a Precondition and the mention in javadocs about this.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/inventory/ItemFactory.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/ItemFactory.java
@@ -210,12 +210,12 @@ public interface ItemFactory {
     @Deprecated(since = "1.19.3") // Paper
     ItemStack enchantItem(@NotNull final ItemStack item, final int level, final boolean allowTreasures);
 
-    // Paper start - Adventure
     /**
      * Creates a hover event for the given item.
      *
      * @param item The item
      * @return A hover event
+     * @throws IllegalArgumentException if the {@link ItemStack#getAmount()} is not between 1 and 99
      */
     @NotNull
     net.kyori.adventure.text.event.HoverEvent<net.kyori.adventure.text.event.HoverEvent.ShowItem> asHoverEvent(final @NotNull ItemStack item, final @NotNull java.util.function.UnaryOperator<net.kyori.adventure.text.event.HoverEvent.ShowItem> op);
@@ -223,12 +223,13 @@ public interface ItemFactory {
     /**
      * Get the formatted display name of the {@link ItemStack}.
      *
+     * @apiNote this component include a {@link net.kyori.adventure.text.event.HoverEvent item hover event}.
+     * When used in chat, make sure to follow the ItemStack rules regarding amount, type, and other properties.
      * @param itemStack the {@link ItemStack}
      * @return display name of the {@link ItemStack}
      */
     @NotNull
     net.kyori.adventure.text.Component displayName(@NotNull ItemStack itemStack);
-    // Paper end - Adventure
 
     // Paper start - add getI18NDisplayName
     /**
--- a/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
@@ -7,6 +7,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.function.Consumer;
 import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.event.HoverEvent;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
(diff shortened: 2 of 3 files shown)


Owen1212055 (approved):
Generally I think this is fine.
Really the only way to get itemstacks in a bad state like this anyways is with the whole item amount nonsense, which we may need to properly revisit in the future

Lulu13022002:
I'd prefer document that broken behaviour in the constructor of ItemStack and setAmount (and the related methods to it), and then mention the few methods that support those oversized stack in the first placed. Then some replacement to those methods can be added and at the end the precondition after some times.
There's no need to be overly focused on those methods.

Doc94:
> I'd prefer document that broken behaviour in the constructor of ItemStack and setAmount (and the related methods to it), and then mention the few methods that support those oversized stack in the first place. Then some replacement to those methods can be added and at the end the precondition after some times. There's no need to be overly focused on those methods.

You mean also add more mentions not?

Malfrador:
In my opinion just preventing creation of such itemstacks in the first place would be better instead of dealing with individual cases like this.

The codec being unable to encode stacks with invalid amounts leads to inventories not saving, so data loss. That is pretty catastrophic, especially as that will affect the entire inventory and not just the oversized item itself.
The client is also unable to decode the stack, so previous uses of invalid stacks like GUI displays etc. are no longer possible. I really don't see why this still has to be allowed in the API, at all. 

So if we continue this down route instead of disabling oversized stacks outright, but still want to prevent plugins using the API from accidentally causing data loss or client errors, we would need to add checks to every other API method that causes items to be sent to the client, or to be placed/modified in an inventory.

The only method where it somewhat makes sense is creating ItemStacks and then using Inventory#addItem, as that deals with splitting the item stack into individual, valid, stacks. But that seems like a pretty niche usecase to begin with.

Doc94:
sounds stale... i wait for more comments or a Paper definition for this.. i can move this to just block create oversised things and well 1.21.5 the breaking change for "dont create items with infinite number..."

kennytv (approved):
Unfortunately the oversized behavior is used enough that we can't break it without a longer discussion and warnings, so for now this is okay

Owen1212055:
Merging this for now to hopefully prevent these items from being used.

But this is pending future conversation to stop/nag the creation of these bad item stacks.

EnderKill98:
Why only allow **2** to 99 though (`Preconditions.checkArgument(item.getAmount() > 1 && ...`)? 🤔

Did I miss some context reading through this or would this indeed fail single items now?

Doc94:
> Why only allow **2** to 99 though? 🤔
> 
> ```diff
> + Preconditions.checkArgument(item.getAmount() > 1 && item.getAmount() <= 99, "ItemStack amount must be between 1 and 99 but was %s", item.getAmount());
> ```
> 
> Did I miss some context reading through this or would this indeed fail single items now?

ehhh... true i missing an `=` because i just test the big numbers and not the normal ItemStack with 1 unit...
PR for fix that... https://github.com/PaperMC/Paper/pull/12505

PR #12218, opened 2025-03-02
add null check to level ref in Entity constructor

### LINKED ISSUE #12210: Entity::new null level constructor broken by configurable entity despawn time
### Expected behavior

Calling Entity::new with a null Level within a custom implementation of Entity will create an Entity instance without requiring a costly Level object, useable for constructing packets which take an Entity as a parameter without huge overhead. The constructor will succeed even with a null Level.

### Observed/Actual behavior

As of 2f58805, calling Entity::new in the described scenario throws a NullPointerException due to no null check being present for the Level in the recently merged custom despawn time system.

### Steps/models to reproduce

Create a custom implementation of Entity, which passes a null Level to the super constructor. Create a new instance of the custom entity.

### Plugin and Datapack List

None

### Paper version

This server is running Paper version 1.21.4-187-main@1d5e5a5 (2025-02-28T19:53:30Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21-130-b1b5d4c (MC: 1.21)

### Other

https://github.com/PaperMC/Paper/pull/11454

This issued was introduced in the aforementioned custom despawn time system.

Image

Paper already modifies the Entity constructor heavily, and performs a null check when accessing attributes of the Level in the activation range system. The lack of null check in the custom despawn system is not consistent with Paper's preexisting patches in Entity. The Entity constructor with a null Level still works as of 1.21.4 outside of Paper servers.

Image

PolyEdge:
Closes https://github.com/PaperMC/Paper/issues/12210

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -192,7 +192,7 @@
          this.entityData = builder.build();
          this.setPos(0.0, 0.0, 0.0);
          this.eyeHeight = this.dimensions.eyeHeight();
-+        this.despawnTime = type == EntityType.PLAYER ? -1 : level.paperConfig().entities.spawning.despawnTime.getOrDefault(type, io.papermc.paper.configuration.type.number.IntOr.Disabled.DISABLED).or(-1); // Paper - entity despawn time limit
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

kennytv (approved):
Thanks, though please note that we don't make any promises of allowing the passed level to be null in the future, especially given the alternative of passing a proper level is trivial

PR #11954, opened 2025-01-12
Don't auto-create any brig redirects

Machine-Maker:
Supersedes https://github.com/PaperMC/Paper/pull/11794
Supersedes https://github.com/PaperMC/Paper/pull/11953
Fixes https://github.com/PaperMC/Paper/issues/10827
Fixes https://github.com/PaperMC/Paper/issues/11647
Fixes https://github.com/PaperMC/Paper/issues/11384

----

Don't automatically create any brig-style redirects. For alises and `namespace:command`, just copy over the requirement, command, redirect and children to a new literal. Also do this for the vanilla namespaced commands. This should solve all redirect issues (at least the ones not inherently part of brig itself like nested redirects).

PR #12219, opened 2025-03-03
Fix children copying when filling missing command redirects

booky10:
The redirect flattening previously copied over children nodes without checking node requirements (e.g. permissions)

Before https://github.com/PaperMC/Paper/pull/11954, this caused node requirements to be ignored for second-level command nodes created using the brigadier api (if namespaced commands were prevented from being sent using the spigot config option), see https://github.com/PaperMC/Paper/pull/11953

Since https://github.com/PaperMC/Paper/pull/11954 is merged now, it's much less of an issue, but as https://github.com/PaperMC/Paper/pull/12136 has been merged, it still is an issue

### DIFF
--- a/paper-server/patches/sources/net/minecraft/commands/Commands.java.patch
+++ b/paper-server/patches/sources/net/minecraft/commands/Commands.java.patch
@@ -172,7 +172,7 @@
              }
  
              return null;
-@@ -359,26 +_,121 @@
+@@ -359,26 +_,131 @@
      }
  
      public void sendCommands(ServerPlayer player) {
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):

Owen1212055:
After further reviewing this logic there are significant things that need to be redone with the flattening logic. We will be revisiting this in hopes of better resolving issues like these. In this current implementation, (not inheritably your changes) there are circular redirects/forks that occur that cause improving this logic to fail. 

This requires more significant refactors that we will be reserving for the dev team.

Thank you so much for your PR, and sorry for the wait. ❤️

PR #12220, opened 2025-03-03
ServerTickManager#requestGameToSprint - Silence command like feedback

ShaneBeee:
This PR aims to silence the feedback from server sprinting via the API.

When using the API to request the server to sprint:
<https://jd.papermc.io/paper/1.21.4/org/bukkit/ServerTickManager.html#requestGameToSprint(int)>

The tick manager still provides feedback as if the vanilla `/tick sprint` command were run:
Screenshot 2025-03-02 at 10 45 31 PM

When using the API to do this task, the server shouldn't be responding as if a command were run.

This has driven me nuts since Bukkit added this, I just kept forgetting to report it and/or actually do something about it.
But nows my chance.

One thing I did think was someone out there using this API might actually like this feedback?!?!?
So I was thinking, should I make another `ServerTickManager#requestGameToSprint` method with an option to silence?
Looking for feedback!

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/ServerTickRateManager.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/ServerTickRateManager.java.patch
@@ -17,7 +17,7 @@
 +    }
 +
 +    public boolean requestGameToSprint(int sprintTime, boolean silent) {
-+        this.silent = silent;
++        if (!isSprinting()) this.silent = silent;
 +        // Paper end - silence feedback when API requests sprint
          boolean flag = this.remainingSprintTicks > 0L;
          this.sprintTimeSpend = 0L;
--- a/paper-server/patches/sources/net/minecraft/server/ServerTickRateManager.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/ServerTickRateManager.java.patch
@@ -1,0 +1,34 @@
+--- a/net/minecraft/server/ServerTickRateManager.java
++++ b/net/minecraft/server/ServerTickRateManager.java
+@@ -14,6 +_,7 @@
+     private long scheduledCurrentSprintTicks = 0L;
+     private boolean previousIsFrozen = false;
+     private final MinecraftServer server;
++    private boolean silent; // Paper - silence feedback when API requests sprint
+ 
+     public ServerTickRateManager(MinecraftServer server) {
+         this.server = server;
+@@ -67,6 +_,13 @@
+     }
+ 
+     public boolean requestGameToSprint(int sprintTime) {
++        // Paper start - silence feedback when API requests sprint
++        return requestGameToSprint(sprintTime, false);
++    }
++
++    public boolean requestGameToSprint(int sprintTime, boolean silent) {
++        this.silent = silent;
++        // Paper end - silence feedback when API requests sprint
+         boolean flag = this.remainingSprintTicks > 0L;
+         this.sprintTimeSpend = 0L;
+         this.scheduledCurrentSprintTicks = sprintTime;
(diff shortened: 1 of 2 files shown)


paper-server/patches/sources/net/minecraft/server/ServerTickRateManager.java.patch
```diff
@@ -0,0 +1,36 @@
+--- a/net/minecraft/server/ServerTickRateManager.java
++++ b/net/minecraft/server/ServerTickRateManager.java
+@@ -14,6 +_,7 @@
+     private long scheduledCurrentSprintTicks = 0L;
+     private boolean previousIsFrozen = false;
+     private final MinecraftServer server;
++    private boolean silent; // Paper - silence feedback when API requests sprint
```
Leguan16, line 7: Why did you make this a field when it's only used once in the method?

ShaneBeee: because its carried across to another method

The diff makes it not look that way, but it is split:
Screenshot 2025-03-03 at 12 23 33 AM
Screenshot 2025-03-03 at 12 24 36 AM

Leguan16: Ahh ok that makes sense.

NonSwag: Wouldn't this mean that if the method is invoked multiple times, the feedback might not be printed correctly?
For example, suppose Player 1 uses the /tick sprint command to make the game sprint forward by 10 seconds. During this sprint, a plugin unexpectedly decides to silently sprint the game forward by an entire day. As a result, Player 1's message is not printed immediately, but instead appears only after a full day has passed.

ShaneBeee: You're right. ... I made a change!
API will now no longer override if the server is already sprinting.

In a new test, I sprinted 1 day with the command, then 1 second with the API.
As expected the sprint was overridden by the API (as Minecraft intends) ... and the message was sent.

NonSwag: in theory the problems still persists 
1. start sprinting for 1 minute
2. start sprinting silently for 2 minutes
3. after the game sprinted 1 minute forward -> message
4. after 2 minutes -> message

ShaneBeee: 3 wont happen, there aren't 2 sprints going at the same time.
2 will override 1
4 will send the message because we wont silence the sprint from the command if the sprint is already running.

NonSwag: oh yeah you are right nvm then

Lulu13022002: The other way around is more tricky to handle but i don't think it's worth (when a plugin sprint and the stop command is executed)

ShaneBeee: > The other way around is more tricky to handle but i don't think it's worth (when a plugin sprint and the stop command is executed)

I agree. 
Sprinting isn't something servers are going to be doing on a regular basis outside of testing/debugging/stress-testing.
I cant imagine a plugin randomly issuing a sprint without the owner/admins knowledge, to the point they'd ALSO use the `/tick sprint` command.

Machine-Maker (approved):

Owen1212055 (approved):

PR #12221, opened 2025-03-03
Expand PlayerDeathEvent API

dawon:
Replaces #8913.

In the end I have deprecated the old .deathMessage() methods in the event and replaced it with deathChatMessage and deathScreenMessage to allow changing those two independently. I tried to make it so the behavior does not change at all when using the old methods.

I have not yet tested it, will do later, but can be reviewed....

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
@@ -19,7 +19,9 @@ public class PlayerDeathEvent extends EntityDeathEvent {
     private int newExp = 0;
     private int newLevel = 0;
     private int newTotalExp = 0;
+    private boolean showDeathMessages;
     private Component deathMessage;
+    private Component deathScreenMessageOverride = null;
     private boolean doExpDrop;
     private boolean keepLevel = false;
     private boolean keepInventory = false;
@@ -27,27 +29,28 @@ public class PlayerDeathEvent extends EntityDeathEvent {
     private final List<ItemStack> itemsToKeep = new ArrayList<>();
 
     @ApiStatus.Internal
-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final @Nullable Component deathMessage) {
-        this(player, damageSource, drops, droppedExp, 0, deathMessage);
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final @Nullable Component deathMessage, final boolean showDeathMessages) {
+        this(player, damageSource, drops, droppedExp, 0, deathMessage, showDeathMessages);
     }
 
     @ApiStatus.Internal
-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final @Nullable Component deathMessage) {
-        this(player, damageSource, drops, droppedExp, newExp, 0, 0, deathMessage);
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final @Nullable Component deathMessage, final boolean showDeathMessages) {
+        this(player, damageSource, drops, droppedExp, newExp, 0, 0, deathMessage, showDeathMessages);
     }
 
     @ApiStatus.Internal
-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable Component deathMessage) {
-        this(player, damageSource, drops, droppedExp, newExp, newTotalExp, newLevel, deathMessage, true);
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable Component deathMessage, final boolean showDeathMessages) {
+        this(player, damageSource, drops, droppedExp, newExp, newTotalExp, newLevel, deathMessage, showDeathMessages, true);
     }
 
     @ApiStatus.Internal
-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable Component deathMessage, final boolean doExpDrop) {
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final int newTotalExp, final int newLevel, final @Nullable Component deathMessage, final boolean showDeathMessages, final boolean doExpDrop) {
         super(player, damageSource, drops, droppedExp);
         this.newExp = newExp;
         this.newTotalExp = newTotalExp;
         this.newLevel = newLevel;
         this.deathMessage = deathMessage;
+        this.showDeathMessages = showDeathMessages;
         this.doExpDrop = doExpDrop;
     }
 
@@ -76,6 +79,7 @@ public class PlayerDeathEvent extends EntityDeathEvent {
         this.newExp = newExp;
         this.newTotalExp = newTotalExp;
         this.newLevel = newLevel;
+        this.showDeathMessages = true;
         this.deathMessage = LegacyComponentSerializer.legacySection().deserializeOrNull(deathMessage);
         this.doExpDrop = doExpDrop;
     }
@@ -84,6 +88,30 @@ public class PlayerDeathEvent extends EntityDeathEvent {
     @Override
     public Player getEntity() {
         return (Player) this.entity;
+    }
+
+    /**
+     * Get whether the death message should be shown.
+     * By default, this is determined by {@link org.bukkit.GameRule#SHOW_DEATH_MESSAGES}.
+     *
+     * @return whether the death message should be shown
+     * @see #deathMessage()
+     * @see #deathScreenMessageOverride()
+     */
+    public boolean getShowDeathMessages() {
+        return showDeathMessages;
+    }
+
+    /**
+     * Set whether the death message should be shown.
+     * By default, this is determined by {@link org.bukkit.GameRule#SHOW_DEATH_MESSAGES}.
+     *
+     * @param displayDeathMessage whether the death message should be shown
+     * @see #deathMessage()
+     * @see #deathScreenMessageOverride()
+     */
+    public void setShowDeathMessages(boolean displayDeathMessage) {
+        this.showDeathMessages = displayDeathMessage;
     }
 
     /**
@@ -177,7 +205,7 @@ public class PlayerDeathEvent extends EntityDeathEvent {
     /**
      * Set the death message that will appear to everyone on the server.
      *
-     * @param deathMessage Message to appear to other players on the server.
+     * @param deathMessage message to appear to other players on the server.
      * @deprecated in favour of {@link #deathMessage(Component)}
      */
     @Deprecated
@@ -195,6 +223,32 @@ public class PlayerDeathEvent extends EntityDeathEvent {
     @Deprecated
     public String getDeathMessage() {
         return LegacyComponentSerializer.legacySection().serializeOrNull(this.deathMessage);
+    }
+
+    /**
+     * Overrides the death message that will appear on the death screen of the dying player.
+     * By default, this is null.
+     * <p>
+     * If set to null, death screen message will be same as {@code deathMessage()}.
+     * <p>
+     * If the message exceeds 256 characters it will be truncated.
+     *
+     * @param deathScreenMessageOverride Message to appear on the death screen to the dying player.
+     */
+    public void deathScreenMessageOverride(@Nullable Component deathScreenMessageOverride) {
+        this.deathScreenMessageOverride = deathScreenMessageOverride;
+    }
+
+    /**
+     * Get the death message override that will appear on the death screen of the dying player.
+     * By default, this is null.
+     * <p>
+     * If set to null, death screen message will be same as {@code deathMessage()}.
+     * <p>
+     * @return Message to appear on the death screen to the dying player.
+     */
+    public @Nullable Component deathScreenMessageOverride() {
+        return this.deathScreenMessageOverride;
     }
 
     /**
--- a/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
@@ -1,5 +1,13 @@
 --- a/net/minecraft/server/level/ServerPlayer.java
 +++ b/net/minecraft/server/level/ServerPlayer.java
+@@ -65,7 +_,6 @@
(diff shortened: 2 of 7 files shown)


paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +155,118 @@ public Player getEntity() {
...
+
+    /**
+     * Set whether the death message should be shown.
+     * By default, this is determined by {@link org.bukkit.GameRule#SHOW_DEATH_MESSAGES}.
+     *
+     * @param displayDeathMessage whether the death message should be shown
+     * @see #deathChatMessage()
+     * @see #deathScreenMessage()
+     */
+    public void setShouldDisplayDeathMessage(boolean displayDeathMessage) {
```
novitpw: plz, rename to shouldDisplayDeathMessage

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -12,39 +14,47 @@
...
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final @Nullable Component deathMessage) {
+        this(player, damageSource, drops, droppedExp, 0, true, deathMessage);
     }
 
     @org.jetbrains.annotations.ApiStatus.Internal
-    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final int newExp, final @Nullable net.kyori.adventure.text.Component deathMessage) {
-        this(player, damageSource, drops, droppedExp, newExp, 0, 0, deathMessage);
+    public PlayerDeathEvent(final @NotNull Player player, final @NotNull DamageSource damageSource, final @NotNull List<ItemStack> drops, final int droppedExp, final boolean displayDeathMessage, final @Nullable Component deathMessage) {
+        this(player, damageSource, drops, droppedExp, 0, displayDeathMessage, deathMessage);
+    }
```
Warriorrrr: Unused constructors marked with internal can be removed

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -363,13 +364,54 @@
 +        }
 +    }
 +    // Paper end - PlayerDeathEvent#getItemsToKeep
- 
++
++    // Paper start - Expand PlayerDeathEvent API
++    private void sendClientboundPlayerCombatKillPacket(boolean displayMessage, Component deathScreenMessage) {
```
Warriorrrr: Why is this method needed, can't you just replace it with the death screen message where it was originally?

Warriorrrr (dismissed):

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +155,118 @@ public Player getEntity() {
...
+
     /**
      * Set the death message that will appear to everyone on the server.
      *
      * @param deathMessage Component message to appear to other players on the server.
+     * @deprecated in favour of {@link #deathChatMessage(Component)} and {@link #deathScreenMessage(Component)}
      */
-    public void deathMessage(final net.kyori.adventure.text.@Nullable Component deathMessage) {
-        this.deathMessage = deathMessage;
+    @Deprecated
```
DerEchtePilz: Why does this need to be deprecated? Seems like a good shortcut if you want to set both messages to the same value instead of calling the two new methods.

dawon: I have undeprecated the setter, I have left the getter deprecated though...

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
dawon: I could, but I would need to duplicate this code over to branch when you set deathChatMessage to null but set deathScreenMessage to something.. This seemed somewhat cleaner when updated during version updates....

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -69,7 +69,9 @@ public PlayerDeathEvent(@NotNull final Player player, final @NotNull DamageSourc
         this.newExp = newExp;
         this.newTotalExp = newTotalExp;
         this.newLevel = newLevel;
-        this.deathMessage = net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserializeOrNull(deathMessage); // Paper
+        this.displayDeathMessage = true;
+        this.deathChatMessage = LegacyComponentSerializer.legacySection().deserializeOrNull(deathMessage); // Paper
```
notTamion: no need for paper comments here

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +145,116 @@ public Player getEntity() {
...
+    /**
+     * Set whether the death message should be shown.
+     * By default, this is determined by {@link org.bukkit.GameRule#SHOW_DEATH_MESSAGES}.
+     *
+     * @param displayDeathMessage whether the death message should be shown
+     * @see #deathChatMessage()
+     * @see #deathScreenMessage()
+     */
+    public void setShouldDisplayDeathMessage(boolean displayDeathMessage) {
+        this.displayDeathMessage = displayDeathMessage;
```
notTamion: Do we really need this? Unless i am missing something event#deathMessage(null) will have the same effect right? Besides the shouldDisplayDeathMessage getter will return true even if both chat and screen message are set to null. Considering we already specify that setting these to null will lead to no message being sent

EDIT: could just call the event with a null deathChatMessage if the gamerule was changed

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +145,116 @@ public Player getEntity() {
...
+    }
+
+    /**
+     * Set the death message that will appear in chat and on death screen to everyone on the server.
```
notTamion: Sounds a bit weird imo.

```suggestion
     * Set the death message that will be broadcasted and appear on the players death screen.
```

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +145,116 @@ public Player getEntity() {
...
-        return this.deathMessage;
+    @Deprecated
+    public @Nullable Component deathMessage() {
+        return this.deathChatMessage;
     }
-    // Paper end - adventure
 
     /**
-     * Set the death message that will appear to everyone on the server.
+     * Set the death message that will appear in chat and on death screen to everyone on the server.
```
notTamion: here too

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -143,49 +145,116 @@ public Player getEntity() {
...
+        return LegacyComponentSerializer.legacySection().serializeOrNull(this.deathChatMessage);
     }
-    // Paper end
+
+    /**
+     * Set the death message that will appear in the chat.
+     * <p>
+     * If set to null, no message will be sent to chat.
+     *
+     * @param deathChatMessage Message to appear in the chat when the player dies.
```
notTamion:
```suggestion
     * @param deathChatMessage message to appear in the chat when the player dies.
```
below aswell

notTamion (dismissed):

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
dawon: Should I then remove all the Paper comments from the event class? I am not sure what is the policy after softspoon regarding existing Paper comments outside of the NMS...

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
dawon: The idea was to expose the gamerule to the event, you are right about the workaround using null argument, but I would argue that explicit boolean is more transparent. But I am open to discussion and can remove it if it is deemed not worth it...

notTamion: yeah nvm the boolean might be a better way to represent that. though i think we should then probably rename it to something like allowDisplayingDeathMessage or something similar.

Owen1212055 (commented):
I am a little unsure with this boolean AND the fact that deathChatMessage can be set to null to control if the message is set. Like, I do understand that the boolean kinda serves a different purpose but I feel like its odd to have that functionality covered in both places.

Unless others think its fine, just wondering tho.

notTamion:
I had left a review about that, though i don't think there is a way for us to get rid of the boolean without breaking existing plugins.  Since Plugins currently overriding this message probably don't check for the gamerule this could potentially lead to players information being exposed (on a pvp enabled server for example other players on the server now know you are standing around at your bed/spawn without any stuff). Not sure if we really need to break api here. We could mention the allowDisplayingDeathMessage method in the deathMessage(Component) methods jdocs to make it more clear to plugin devs.

dawon:
For me, this solution is more transparent while preserving original behavior. Also, there was a TODO to allow overriding the gamerule, so it made sense for me to add it as well. But I do really care about the split of death screen and death chat message controls, so if you agree to remove it, I have no problem removing it.

dawon:
Rebased on master...

kennytv (approved):

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
```diff
@@ -86,6 +92,30 @@ public Player getEntity() {
...
 
+    /**
+     * Get whether the death message should be shown.
+     * By default, this is determined by {@link org.bukkit.GameRule#SHOW_DEATH_MESSAGES}.
+     *
+     * @return whether the death message should be shown
+     * @see #deathChatMessage()
+     * @see #deathScreenMessage()
+     */
+    public boolean allowDisplayingDeathMessage() {
```
lynxplay: I am a bit confused on the usecase for this.

In case the gamerule is true, plugins can already null it out to avoid sending the messages.
In case it is false (the non default) it feels pretty weird for a plugin to change this here?

lynxplay (changes requested):
Generally lgtm on the idea of this.
During PR Party, the concern that came up was a very niche change like the screen message requiring a rename to the death message.
Would you be fine with moving screen death component to an "override" kinda system?

`null` would indicate falling back to the deathMessage()
Anything else would override.

dawon:
> Generally lgtm on the idea of this. During PR Party, the concern that came up was a very niche change like the screen message requiring a rename to the death message. Would you be fine with moving screen death component to an "override" kinda system?
> 
> `null` would indicate falling back to the deathMessage() Anything else would override.

You would then lose the ability to remove the screen message... Also it being different model from deathChatMessage might confuse devs. I think. Also also I left the original deathMessage() to allow changing both screen&chat values... 

If you insist on this, I can change it, but I think this model would be more consistent...

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
dawon: I added this based on a TODO which was in the code and also to expose the gameRule... We already disccussed this with NotTamion above (https://github.com/PaperMC/Paper/pull/12221#discussion_r1983046406) But I can remove it, if you want...

lynxplay:
Fair points, tho I will say that you do not loose the ability to hide the death message.
Component.empty() would work perfectly fine there to "hide it". The server has to send the packet and has to provide a component, it is just "empty". `null` would be the only value for the death message component that would trigger a fallback to the chat/deathMessage.

Whether or not the override system is confusing is I guess a good discussion to be had.
I'd just really like to avoid a rename/deprecation change to #deathMessage for a feature that is as niche as the screen death message being different from the one in chat, so the override layout was the thing we came up with.
I'll grumble about it for a bit longer, maybe I can come up with a solution outside of override but yea, such a system certainly would not prevent setting the screen message to empty()

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
lynxplay: Ah, the discussion was marked as resolved, I did not see it :sweat_smile: mb
I guess the magic boolean has always been there, so exposing it is a general good change.
I'll chat with owen on the naming of this, I think I'd prefer this being named *very* closely after the gamerule and/or include the fact that  this is a gamerule in the method name but, that is pending further chats with other team members.

dawon:
> Component.empty() would work perfectly fine there to "hide it". The server has to send the packet and has to provide a component, it is just "empty". `null` would be the only value for the death message component that would trigger a fallback to the chat/deathMessage.

Fair point...

> I'd just really like to avoid a rename/deprecation change to #deathMessage for a feature that is as niche as the screen death

AFAIR I only deprecated the getter, setter is still there nondeprecated as a utility method to rename both. I could leave the getter undeprecated as well, but then it might be confusing...


I would also argue it is not as niche. I think it is quite useful to allow displaying different values to the player dying and other players and currently there is no api for that (I do it with packets currently)...

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
dawon: I have no problem renaming it :-)

dawon: So: 

getShowDeathMessagesGameRule / setShowDeathMessagesGameRule would be better?

but then again, you are not actually changing the gamerule here... so maybe with Override suffix?

lynxplay:
> AFAIR I only deprecated the getter, setter is still there nondeprecated as a utility method to rename both. I could leave the getter undeprecated as well, but then it might be confusing...

Yea, the setter isn't deprecated, tho I am unsure *how* consistent that is with a full separation of chat and screen death message when it is just named "deathMessage".

> I would also argue it is not as niche. I think it is quite useful to allow displaying different values to the player dying and other players and currently there is no api for that (I do it with packets currently)...

It certainly is quite useful, I just have never heard a feature request for it on github or the discord (outside of your just now), which is why I am considering this niche. It certainly isn't the main behaviour of the event to split these two.
Having a single source of "true" death message is pretty useful when plugins attempt to interop. Promoting the single-player packet only screen death message to the same level as the "death message" or now "chat death message" means choosing the "correct" one is now pretty much up to the plugins.

paper-api/src/main/java/org/bukkit/event/entity/PlayerDeathEvent.java
lynxplay: mhm maybe, tho those names are quite verbose. Might be okay to not include the gamerule suffix, I'll think a bit about it throughout the day. Worse case we can spam the javadocs a bit about how this isn't really a boolean added by us but is direct control over the gamerule for this specific death.

dawon:
> It certainly is quite useful, I just have never heard a feature request for it on github or the discord (outside of your just now), which is why I am considering this niche. It certainly isn't the main behaviour of the event to split these two.

I blame it to the fact users either didn't need that or did not know that was possible :D

> Having a single source of "true" death message is pretty useful when plugins attempt to interop. Promoting the single-player packet only screen death message to the same level as the "death message" or now "chat death message" means choosing the "correct" one is now pretty much up to the plugins.

Ok that is a strong point, let's go with the override model... I will update the PR later today...

dawon:
I've rebased on (almost) latest master and changed both name of the boolean and the screen message to Override...

lynxplay (approved):

Owen1212055 (approved):

PR #12223, opened 2025-03-03
Cancel PlayerLaunchProjectileEvent properly for enderpearls

### LINKED ISSUE #12217: Cancellation of PlayerLaunchProjectileEvent for Ender pearls still teleports players after relog
### Expected behavior

When `PlayerLaunchProjectileEvent` is cancelled any relog for players should be natural without teleportation.

### Observed/Actual behavior

When `PlayerLaunchProjectileEvent` is cancelled for ender pearl throw and player relogs after that, they will be teleported as if the pearl was thrown.

### Steps/models to reproduce

1. Listen to `PlayerLaunchProjectileEvent` and specifically ender pearls.
2. Cancell all events.
3. Throw an ender pearl and see it does not get thrown.
4. Leave and join the server.
5. Notice player gets teleported for all throws.

How it works for my plugin with a cooldown on the "special" ender pearl:

https://github.com/user-attachments/assets/214bf92a-a281-44ab-a576-47bc0434292e

### Plugin and Datapack List

[15:51:06 INFO]: Paper Plugins: (0):
[15:51:06 INFO]: Bukkit Plugins: (2):
[15:51:06 INFO]:  - LuckPerms, MilkyItems

The portion for the code  in my plugin
```java
    @Override
    public @NotNull Listener getListener() {
        return new Listener() {
            @EventHandler
            public void onPearlUse(PlayerLaunchProjectileEvent event) {
                if (!(event.getProjectile() instanceof EnderPearl pearl)) {
                    return;
                }
                if (!(pearl.getShooter() instanceof Player player)) {
                    return;
                }

                var item = event.getItemStack();

                if (!item.getType().equals(Material.ENDER_PEARL)) {
                    return;
                }

                if (!isSpecialItem(item)) {
                    playersUsingPearl.remove(player.getUniqueId());
                    return;
                }

                event.setShouldConsume(false);

                var cooldown = checkCooldownAndNotify(player, item, false);

                if (cooldown > 0) {
                    event.setCancelled(true);
                    return;
                }

                playersUsingPearl.add(player.getUniqueId());
                setCooldown(item);

                new BukkitRunnable() {
                    private int maxTicks = 200;
                    @Override
                    public void run() {
                        if (!pearl.isValid() || maxTicks-- <= 0) {
                            this.cancel();
                            return;
                        }
                        pearl.getWorld().spawnParticle(Particle.HEART, pearl.getLocation(), 1, 0.05, 0.05, 0.05, 0.05);
                    }
                }.runTaskTimer(MilkyItems.getInstance(), 0L, 1L);
            }
```

### Paper version

> paper version
[15:53:18 INFO]: Checking version, please wait...
[15:53:18 INFO]: This server is running Paper version 1.21.4-187-main@1d5e5a5 (2025-02-28T19:53:30Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-150-46f4fda (MC: 1.21.4)

### Other

_No response_

Lulu13022002:
Closes https://github.com/PaperMC/Paper/issues/12217 and resync the cooldown too like the wind charge

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/item/EnderpearlItem.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/item/EnderpearlItem.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/item/EnderpearlItem.java
 +++ b/net/minecraft/world/item/EnderpearlItem.java
-@@ -21,22 +_,38 @@
+@@ -21,22 +_,42 @@
      @Override
      public InteractionResult use(Level level, Player player, InteractionHand hand) {
          ItemStack itemInHand = player.getItemInHand(hand);
@@ -39,7 +39,11 @@
 +                );
 +                player.awardStat(Stats.ITEM_USED.get(this));
 +            } else {
-+            // Paper end - PlayerLaunchProjectileEvent
++                if (player instanceof net.minecraft.server.level.ServerPlayer serverPlayer) {
++                    serverPlayer.deregisterEnderPearl(thrownEnderpearl.projectile());
++                    serverPlayer.connection.send(new net.minecraft.network.protocol.game.ClientboundCooldownPacket(player.getCooldowns().getCooldownGroup(itemInHand), 0)); // prevent visual desync of cooldown on the slot
++                }
++                // Paper end - PlayerLaunchProjectileEvent
 +                player.containerMenu.sendAllDataToRemote();
 +                return InteractionResult.FAIL;
 +            }


paper-server/patches/sources/net/minecraft/world/item/EnderpearlItem.java.patch
```diff
@@ -39,7 +39,11 @@
 +                );
 +                player.awardStat(Stats.ITEM_USED.get(this));
 +            } else {
-+            // Paper end - PlayerLaunchProjectileEvent
++                if (player instanceof net.minecraft.server.level.ServerPlayer serverPlayer) {
++                    serverPlayer.deregisterEnderPearl(thrownEnderpearl.projectile());
++                    serverPlayer.connection.send(new net.minecraft.network.protocol.game.ClientboundCooldownPacket(player.getCooldowns().getCooldownGroup(itemInHand), 0)); // prevent visual desync of cooldown on the slot
```
Doc94, line 44: just for curious this is the same than `serverPlayer.getCooldowns().addCooldown(player.getCooldowns().getCooldownGroup(itemInHand), 0, false);` ? well just the map thing.

Warriorrrr (approved):

paper-server/patches/sources/net/minecraft/world/item/EnderpearlItem.java.patch
Owen1212055: Not wanting to store it in the map is defo intentional, basically we want to revert the client state without messing with the server so we want to avoid using that.

Owen1212055 (approved):

Machine-Maker (approved):

PR #12224, opened 2025-03-04
Add ItemStack#copyDataFrom

TonytheMacaroni:
Adds a method to copy data components from a source `ItemStack`, with an optional filter for which components to copy.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
+++ b/paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
@@ -1307,6 +1307,43 @@ public class ItemStack implements Cloneable, ConfigurationSerializable, Translat
     }
 
     /**
+     * Copies the values of components from the provided ItemStack. Will only copy the value of components from the
+     * source ItemStack that are set or overridden.
+     *
+     * @param source the item stack to copy from
+     */
+    @org.jetbrains.annotations.ApiStatus.Experimental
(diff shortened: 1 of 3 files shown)


Machine-Maker:
Maybe copyDataFrom to make it more clear where stuff is being copied from, to.

paper-api/src/main/java/org/bukkit/inventory/ItemStack.java
```diff
@@ -1306,6 +1306,29 @@ public void resetData(final io.papermc.paper.datacomponent.@NotNull DataComponen
...
+    @org.jetbrains.annotations.ApiStatus.Experimental
+    public void copyDataFrom(final @NotNull ItemStack source) {
+        this.copyDataFrom(source, null);
+    }
+
+    /**
+     * Copies the values of components from the provided ItemStack. Will only copy the value of components from the
+     * source ItemStack that are set or overridden.
+     *
+     * @param source the item stack to copy from
```
Machine-Maker, line 1327: Maybe a code example here would be nice, something that shows you can just create a constant list of types, and then use `THAT_LIST::contains` as the parameter in order to nicely copy data from a specific list

TonytheMacaroni: Added an example.

Owen1212055 (approved):

Machine-Maker (approved):

Warriorrrr (approved):

PR #12231, opened 2025-03-05
Improve scheduler

sunmisc:
Originally, I wanted to fix a small memory leak when parsePending occurs
tail equals head = last task, the last task is saved and lives until new tasks arrive. To count ticks, it is not necessary to have a counter for each scheduler (for asynchronous and synchronous)

For the asynchronous scheduler, I can't really understand why synchronization was used yet. There fixed the continuing loop after finding a task

Pain.
I wanted to handle the angular moment when tasks are added in the main thread by adding them to a local non-blocking list, that's what `temp` does, for example, without overloading the priority queue while traversing the actual tasks. But the scheduler api has a lot of unnecessary methods, and no necessary ones

For example, canceling a task by id is probably a bad idea, both in terms of performance and security.
The isQueued / isCurrentlyRunning methods also relate to this.
This functionality should be (if it is needed at all) in BukkitTask
At the same time, there are no really useful methods like pool closing (at least for asynchronous)


Owen1212055:
This is a rather extensive refactor for something that otherwise has been left for a very long time. It may be nice to first have a PR to fix the parsePending issue you have described.

The issue is that this is a pretty extensive looking factor for code that otherwise has not been touched in ages, so making sure that the behavior here is the same is tricky realistically.

Owen1212055:
Closing as per the previous comment, thank you for your contribution. Feel free to reopen with the smaller aformentioned changes. 😄

PR #12232, opened 2025-03-06
Don't resync all attributes when updating scaled health

### LINKED ISSUE #11997: Taking damage after hitting an entity slows sprinting speed
### Expected behavior

Taking damage after hitting an entity should not alter sprinting speed. This is the behaviour seen on the vanilla server.

### Observed/Actual behavior

The FOV and sprinting speed are reduced when taking damage after hitting an entity during uninterrupted sprinting.

### Steps/models to reproduce

1. Hold down your sprint key (or, enable toggle sprint)
2. Begin running by holding W
3. Attack an entity
4. Take any form of damage (e.g. fire, fall)
5. Observe that FOV and running speed are reduced

Demonstration video: https://youtu.be/MytXWu4hQGc

### Plugin and Datapack List

None

### Paper version

This server is running Paper version 1.21.4-118-main@73f7736 (2025-01-18T13:37:35Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

Enabling paper world config `misc.disable-sprint-interruption-on-attack` is a workaround to this issue but has other side-effects such as every consecutive hit being registered as a sprint hit (preventing being able to crit while sprinting).

notTamion:
fixes #11997 
When a player is damaged cb needs to send max health to the player to ensure the client has the correct health data incase scaled health is being used. Instead they are sending all attributes including the movement speed to the client which causes it to slow down.

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
@@ -2828,15 +2828,10 @@ public class CraftPlayer extends CraftHumanEntity implements Player {
 
     public void updateScaledHealth(boolean sendHealth) {
         AttributeMap attributemapserver = this.getHandle().getAttributes();
-        Collection<AttributeInstance> set = new HashSet<>(1);
-
-        this.injectScaledMaxHealth(set, true);
 
         // SPIGOT-3813: Attributes before health
         if (this.getHandle().connection != null) {
-            if (!set.isEmpty()) {
-                this.getHandle().connection.send(new ClientboundUpdateAttributesPacket(this.getHandle().getId(), set));
-            }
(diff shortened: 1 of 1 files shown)


Owen1212055 (approved):
Nice catch.

_although this api is weird idk if anyone actually uses it._

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -2828,13 +2828,15 @@ public void updateScaledHealth() {
 
     public void updateScaledHealth(boolean sendHealth) {
         AttributeMap attributemapserver = this.getHandle().getAttributes();
-        Collection<AttributeInstance> set = attributemapserver.getSyncableAttributes();
+        Collection<AttributeInstance> set = new HashSet<>(1);
```
kennytv: This would cause an instant resize once you add a single entry

Machine-Maker (approved):

PR #12241, opened 2025-03-08
Force update attributes

0-x-2-2:
Fixes attribute swap exploit: https://www.youtube.com/watch?v=jL98SsqfdvU
TLDR: You can swap an item right before using it and the new item will contain the same attributes as the first item: attack speed, attack damage, dig speed, etc.

This exploit allows you to use the attributes from the previous weapon/tool right before swapping. Let me know if I missed any easily abusable actions other than attack/interact? Let me know if anyone also finds issues with this patch. It should work perfectly with my limited testing, but it would be nice to know if this causes new exploits.


0-x-2-2:
There are many different ways to move items, such as pick block, window click, offhand swap, and switch item. The client runs all of those on render, so it is very easy to do all of them out of sync with the server quickly. I am guessing all those packets would be a good starting point.

0-x-2-2:
Not sure if it is correct yet or not still wip

Spottedleaf (approved):

twentyfived2:
hey guyss

PR #12243, opened 2025-03-08
Player - Expose player score

ShaneBeee:
This PR aims to expose NMS `Player#getScore()` and its derivatives.

This score is show in the death screen of the player.   
The name was suggested/discussed with @Owen1212055 on discord.
Chose to use `get/setDeathScreenScore` to prevent confusion/collision with Scoreboard/Objective `getScore()`

Credit goes to Owen for the idea.
(Could have sworn this was already in the API, but nope)

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -3902,4 +3902,28 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      * @return the result of this method, holding leftovers and spawned items.
      */
     PlayerGiveResult give(Collection<ItemStack> items, boolean dropIfFull);
+
+    /**
+     * Get the score that shows in the death screen of the player.
+     * <p>This amount is added to when the player gains experience.</p>
+     *
+     * @return Death screen score of player
+     */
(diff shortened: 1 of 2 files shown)


Owen1212055:
I don't think we need increase. Get/set seems fine since this really won't be used frequently enough. LGTM

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3902,4 +3902,28 @@ default PlayerGiveResult give(final Collection<ItemStack> items) {
...
+    /**
+     * Get the score that shows in the death screen of the player.
+     * <p>This amount is added to when the player gains experience.</p>
+     *
+     * @return Death screen score of player
+     */
+    int getDeathScreenScore();
+
+    /**
+     * Get the score that shows in the death screen of the player.
```
Owen1212055: Copy pasta

ShaneBeee: haha, yep, my bad ... already fixed.

Ill remove increase tho.

Owen1212055 (approved):

kennytv (approved):

Machine-Maker (approved):

PR #12247, opened 2025-03-08
Allows servers to prevent chickens from laying eggs by setting EggLayTime to -1

brickmonster:
I was surprised that NoAI:1 still causes eggs to be laid, but is a more flexible solution for servers that want both as -1 is never normally reached. 

-1 prevents ticking.
0 works as before for instant egg lay.

Try with `/data merge entity @n[type=chicken] {EggLayTime:-1}`

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/animal/Chicken.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/animal/Chicken.java.patch
@@ -1,9 +1,11 @@
 --- a/net/minecraft/world/entity/animal/Chicken.java
 +++ b/net/minecraft/world/entity/animal/Chicken.java
-@@ -91,10 +_,12 @@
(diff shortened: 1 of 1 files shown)


Machine-Maker:
I'm not sure I like changing the vanilla behavior of that value. Server's *can* just set it to max int, which is like 3.5 years of uptime or whatever, of if we *really* want a thing to prevent it, I'd rather it be a separate field that we add.

Owen1212055:
Agreed with above, just honestly set it to a big number.

brickmonster:
It could pause at max int instead, making the setter's intent match the behaviour. Otherwise that's fine by me, feel free to close.

Owen1212055:
Yeah, gonna close this. Just in general with alot of MC related fields like this, setting it to a big number is just fine.

PR #12248, opened 2025-03-08
feat: add event to wind charge explode

mja00:
Looks like this was missed when wind charges were introduced.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/projectile/windcharge/BreezeWindCharge.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/projectile/windcharge/BreezeWindCharge.java.patch
@@ -13,3 +13,14 @@
          this.level()
              .explode(
                  this,
+@@ -28,8 +_,8 @@
+                 pos.x(),
+                 pos.y(),
+                 pos.z(),
+-                3.0F,
+-                false,
++                event.getRadius(), // Paper
++                event.getFire(), // Paper
+                 Level.ExplosionInteraction.TRIGGER,
+                 ParticleTypes.GUST_EMITTER_SMALL,
+                 ParticleTypes.GUST_EMITTER_LARGE,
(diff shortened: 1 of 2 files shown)


Lulu13022002:
The parameters of the event seems wrong and should take effect here (radius and fire)

mja00:
> The parameters of the event seems wrong and should take effect here (radius and fire)

Ah didn't even notice there was another event method.

Lulu13022002:
A bit weird how explosive values are only relevant for some entities that extends them.
But the current values are still frozen and cannot be changed via the setters on the event.
(fire will still be broken cause Spigot broke that a while ago)

mja00:
Yeah realized I didn't implement the setters right after lol

paper-server/patches/sources/net/minecraft/world/entity/projectile/windcharge/BreezeWindCharge.java.patch
```diff
@@ -0,0 +1,26 @@
+--- a/net/minecraft/world/entity/projectile/windcharge/BreezeWindCharge.java
++++ b/net/minecraft/world/entity/projectile/windcharge/BreezeWindCharge.java
+@@ -20,6 +_,12 @@
+ 
+     @Override
+     public void explode(Vec3 pos) {
++        // Paper start - Breeze Wind Charge event
++        org.bukkit.event.entity.ExplosionPrimeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callExplosionPrimeEvent(this, 3.0F, false);
```
Lulu13022002: You can use the constant RADIUS, same for the other class

mja00: man I really need to read. I was looking for something like that

kennytv (approved):

Warriorrrr (approved):

PR #12257, opened 2025-03-10
Call EntityPushedByEntityAttackEvent for Mace AoE

### LINKED ISSUE #12253: No knockback event fired for mace smash attack AoE
### Expected behavior

`EntityKnockbackByEntityEvent` getting called for a player that gets knocked back as a result of a mace smash attack AoE.

**\*** "**Area of effect** (often referred to as **AoE**) refers to mechanics that apply to an area of the ground rather than targeting a particular entity."

### Observed/Actual behavior

A knockback event is called only for the main target of the attack, not for the additional player affected by the AoE.

### Steps/models to reproduce

Have a way of logging events or listen to the `EntityKnockbackByEntityEvent` event directly.
```java
public class KnockbackListener implements Listener {
    @EventHandler
    public void onKnockback(EntityPushedByEntityAttackEvent e) {
        Bukkit.getLogger().info("Knockback on " + e.getEntityType());
    }
}
```
Log: `[18:01:20 INFO]: Knockback on SHEEP`

Do a smash attack with a mace on a mob standing next to another player. This requires falling for at least 1.5 blocks and hitting the mob.

### Plugin and Datapack List

EventLogger (simply logs on knockback event)

### Paper version

This server is running Paper version 1.21.4-204-main@3d13b11 (2025-03-09T09:36:05Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

I noticed [1.21 Tracker](https://github.com/PaperMC/Paper/issues/10876) has "Mace Knockback logic for respective events" included. Not sure if this would be a part of it and I should just wait.

Due to the event not being fired protection plugins like WorldGuard cannot prevent players from pushing others outside of PvE regions in another way than disabling maces altogether.

Opened an [issue on Spigot](https://hub.spigotmc.org/jira/browse/SPIGOT-8016) as well.

Doc94:
Closes #12253

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/item/MaceItem.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/item/MaceItem.java.patch
@@ -15,3 +15,12 @@
              ServerLevel serverLevel = (ServerLevel)attacker.level();
              attacker.setDeltaMovement(attacker.getDeltaMovement().with(Direction.Axis.Y, 0.01F));
              if (attacker instanceof ServerPlayer serverPlayer) {
+@@ -139,7 +_,7 @@
+             double knockbackPower = getKnockbackPower(attacker, livingEntity, vec3);
+             Vec3 vec31 = vec3.normalize().scale(knockbackPower);
+             if (knockbackPower > 0.0) {
+-                livingEntity.push(vec31.x, 0.7F, vec31.z);
++                livingEntity.push(vec31.x, 0.7F, vec31.z, attacker); // Paper - Add EntityKnockbackByEntityEvent and EntityPushedByEntityAttackEvent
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

Wertik:
For what it's worth I tested the PR and it works. The event now gets fired for all targets of the AoE.

```java
@EventHandler
public void onKnockback(EntityKnockbackEvent e) {
    Bukkit.getLogger().info(e.getClass().getSimpleName() + " (" + e.getEntityType() + ")");
}
```

When hitting a sheep next to another player:
```
[21:59:54 INFO]: EntityKnockbackByEntityEvent (SHEEP)
[21:59:54 INFO]: EntityPushedByEntityAttackEvent (PLAYER)
```

kennytv (approved):

PR #12260, opened 2025-03-10
Fix dropped items looking like 1 item with item sanitation

### LINKED ISSUE #12252: Item obfuscation affects item entities' quantity + firework rocket explosions
### Expected behavior

Quantity data from dropped item entities should probably not be obfuscated, as it makes all item entities appear as though they have a quantity of 1 (i.e. no clumping), even if they have a quantity of 64 for example.

### Observed/Actual behavior

Already described in "Expected behaviour" section.

### Steps/models to reproduce

1. Enable item obfuscation.
2. Drop an item stack of 64 on the ground.
3. Observe as it appears to be an item stack of 1.

### Plugin and Datapack List

N/A

### Paper version

This server is running Paper version 1.21.4-188-main@8de7e35 (2025-03-03T16:59:40Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

Warriorrrr:
Closes #12252

This seems like the most readable approach to this, rather than cooking up some expression to do this in one line

### DIFF
--- a/paper-server/src/main/java/io/papermc/paper/util/ItemComponentSanitizer.java
+++ b/paper-server/src/main/java/io/papermc/paper/util/ItemComponentSanitizer.java
@@ -66,10 +66,20 @@ public final class ItemComponentSanitizer {
     public static int sanitizeCount(final ItemObfuscationSession obfuscationSession, final ItemStack itemStack, final int count) {
         if (obfuscationSession.obfuscationLevel() != ItemObfuscationSession.ObfuscationLevel.ALL) return count; // Ignore if we are not obfuscating
 
-        if (GlobalConfiguration.get().anticheat.obfuscation.items.binding.getAssetObfuscation(itemStack).sanitizeCount()) {
-            return 1;
-        } else {
+        if (!GlobalConfiguration.get().anticheat.obfuscation.items.binding.getAssetObfuscation(itemStack).sanitizeCount() || count <= 1) {
             return count;
+        }
+
+        // See https://minecraft.wiki/w/Item_(entity)#Appearance for reference, this returns the lowest possible count
+        // for the item to keep it looking the same.
+        if (count <= 16) { // 2 - 16
+            return 2;
+        } else if (count <= 32) { // 17 - 32
+            return 17;
+        } else if (count <= 48) { // 33 - 48
+            return 33;
+        } else { // 49+
+            return 49;
         }
     }
 


Owen1212055:
Yeah so, this technically works. Has the consequence of causing ALL items to no longer have count serialized, however.
Imo a more proper solution would be to add something to the syncher to denote that it shouldn't be sanitized.

paper-server/src/main/java/io/papermc/paper/util/ItemComponentSanitizer.java
```diff
@@ -66,11 +66,21 @@ private static ItemEnchantments dummyEnchantments() {
     public static int sanitizeCount(final ItemObfuscationSession obfuscationSession, final ItemStack itemStack, final int count) {
         if (obfuscationSession.obfuscationLevel() != ItemObfuscationSession.ObfuscationLevel.ALL) return count; // Ignore if we are not obfuscating
 
-        if (GlobalConfiguration.get().anticheat.obfuscation.items.binding.getAssetObfuscation(itemStack).sanitizeCount()) {
-            return 1;
-        } else {
+        if (!GlobalConfiguration.get().anticheat.obfuscation.items.binding.getAssetObfuscation(itemStack).sanitizeCount() || count <= 1) {
```
masmc05, line 69: Maybe first check count? Seems like a free performance gain for most equipment

lynxplay:
I wonder if it wouldn't be smarted to replace the codecs used by the entity data accessors that we *actually* want to sanitize and do the session starting there.

paper-server/src/main/java/io/papermc/paper/util/ItemComponentSanitizer.java
0-x-2-2: In reality, it most likely does not matter that much and honestly, branch prediction taking the wrong path might be worse.

masmc05: It's indeed not worth big discussion as the gain is minimal, so won't continue past this message explaining my thoughts. If someone could mark this as resolved would be glad 

Since the check is already present, it makes sense to make the easy check first that already has all the data on stack instead of the pointer chasing through all the ram, especially since in current implementation the majority of items passing here will be from entiy equipment, which is mostly either 1 for unstackable items (armor) or 0 for empty slots, so since it's true it's not gonna even need to check the config

TheosisOfficial:
I added a comment with additional info to #12252 (https://github.com/PaperMC/Paper/issues/12252), so I don't think this entirely closes it anymore.

Warriorrrr:
That probably should've been in a separate issue, but I'll close this for now anyways

TheosisOfficial:
> That probably should've been in a separate issue, but I'll close this for now anyways

Do you want me create a separate issue for it now?

PR #12264, opened 2025-03-11
Properly grow plants with growth modifier

### LINKED ISSUE #8544: Modifying cactus growth modifier causes it to not break after growing

TrollsterCooleg:
Post-softspoon version of https://github.com/PaperMC/Paper/pull/8564
Resolves: https://github.com/PaperMC/Paper/issues/8544

Spigots implementation of the growth modifier boils down to a plain chance for a plant to either grow immediately or to grow a single age value towards its fully grown age.

This behaviour is specifically troubling for plants like the cactus or the sugar cane, as they change an unrelated block when growing up (the free air block above their tip). This blocks validity is checked during neighbour updated which are triggered by resetting the previous tip of the plant back to [age=0].

With spigots modifier implementation, blocks at the age of zero are able to grow a new block above them however, causing the block change of the previous tip to change a block of age zero to a block of age zero, which does not cause a neighbour update as nothing actually changed. This means that the about block is not updated and hence may remain in an otherwise invalid state. (E.g. a cactus block next to a solid block).

This commit rewrites the existing spigot growth modifier implementation specifically for the plants that grow new blocks like the cactus or the sugar can block.

The modifier provided for the plants is separated into two distinct values. One being the remains after a floored division by 100, e.g. the original modifier without its 10^0 and 10^1 digits, the other the modifier modulo 100.
The modifier 475 would, for example, be split into a base growth of 4 and a percentage chance of an additional age progression of 75%. This way, on average, the plant would grow 4.75 times as fast as a normal plant, which would only grow a single age per random tick.

### DIFF
--- a/paper-server/src/main/java/org/spigotmc/SpigotWorldConfig.java
+++ b/paper-server/src/main/java/org/spigotmc/SpigotWorldConfig.java
@@ -133,6 +133,14 @@ public class SpigotWorldConfig {
         this.pitcherPlantModifier = this.getAndValidateGrowth("PitcherPlant"); // Paper
     }
 
+    public int computeBlockGrowingAgeProgression(final net.minecraft.util.RandomSource random, final int modifier) {
+        final double modifierScaled = modifier / 100.0D;
+        final int baseGrowth = (int) modifierScaled;
+        final double chanceOfDoubleGrowth = modifierScaled - baseGrowth;
+
+        return baseGrowth + (random.nextFloat() < chanceOfDoubleGrowth ? 1 : 0);
+    }
+
     public double itemMerge;
     private void itemMerge() {
         this.itemMerge = this.getDouble("merge-radius.item", 0.5);
--- a/paper-server/src/main/java/org/spigotmc/SpigotWorldConfig.java
+++ b/paper-server/src/main/java/org/spigotmc/SpigotWorldConfig.java
@@ -83,7 +83,7 @@ public class SpigotWorldConfig {
     public int beetrootModifier;
     public int carrotModifier;
     public int potatoModifier;
-    public int torchFlowerModifier; // Paper
+    public int torchFlowerModifier;
     public int wheatModifier;
     public int wartModifier;
     public int vineModifier;
@@ -94,8 +94,8 @@ public class SpigotWorldConfig {
     public int twistingVinesModifier;
     public int weepingVinesModifier;
     public int caveVinesModifier;
-    public int glowBerryModifier; // Paper
-    public int pitcherPlantModifier; // Paper
+    public int glowBerryModifier;
+    public int pitcherPlantModifier;
 
     private int getAndValidateGrowth(String crop) {
         int modifier = this.getInt("growth." + crop.toLowerCase(java.util.Locale.ENGLISH) + "-modifier", 100);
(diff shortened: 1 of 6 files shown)


lynxplay:
> Sorry for the late reply
I don't think we need to overthink this and imo this impl should be straightforward it's just spigot that has made a unwise decision in the first place. But i really like your solution and think that just a note on the paperdoc about what we change for theses config is enough and say that even if we fixed spigot it can still break niche plugin/command/datapack/resource pack for map maker purpose i guess when the percent overflow 100 percent since it skip some natural step.
There potentially some similar issue for the beetroot and crops where the percent change the moisture rate and so is not independent like the other. The beetroot/torchflower has even a bigger issue since its class override the parent method with a 1/3 chance.

> There's also another issue when the cactus reach its maximum age (15) it normally creates a new block only in the next random tick however with this pr the new block is created instantly without waiting a new randomtick. That means the cactus grow faster than normal even with the default growth modifier. Another issue is that the last stage (15) is always skipped since it is almost immediately reset to age=0 when a new block grow and so this stage will be unseen in the server.

From the old PR.

paper-server/src/main/java/org/spigotmc/SpigotWorldConfig.java
```diff
@@ -133,6 +133,16 @@ private void growthModifiers() {
         this.pitcherPlantModifier = this.getAndValidateGrowth("PitcherPlant"); // Paper
     }
 
+    // Paper start - Properly grow plants with growth modifier
```
Doc94: This comments are not necessary for the non NMS files.

TrollsterCooleg:
I'll work on the other comments later, all I did there is remove the SpigotWorldConfig comments.

TrollsterCooleg:
This feels janky, and it may need restructuring, but I believe that the issues raised above are now solved. I honestly couldn't think of a better way to handle the beetroot and fireflower 2/3 chance, but I'd love to hear better ideas, or better ways to denote it in the class for the crops other than leaving the method there where all it does is call the super and a comment.

Lulu13022002:
That's fair feel free to split the PR, i only noticed that while testing the old pr. In that case just rename the title to be more specific. To sum up what i said previously i'm fine with this approach but would like to document a bit the stepping behavior that happens here which is not consistent with the other modifiers.

TrollsterCooleg:
With the incoming addition in 1.21.5 of the cactus flower, I'd be interested in hearing how we want this PR to integrate with the cactus flower.

As lynxplay mentioned in the old commit, with a system like this, its entirely possible for some crop ages to be skipped. With the cactus flower only being able to grow with a crop age of 8, skipping an age would entirely disable the possibility of cactus flower growth, which would obviously be problematic if not intended.

We could make it so that if the modifier for cactus or a hypothetical modifier for cactus flower is modified than it changes the condition from requiring that the age be 8 to being a 1/16 chance instead, but that would feel a bit odd for this pr to replace one system of breaking how vanilla grows crops to a different alternative when modifiers are applied.

Alternatively we could just scale down the pr to simply modify the age to 15 on a cactus before applying the old instant growth logic and not improve the modifier's functionality in any way besides fixing the bug caused by lack of block updates.

If we didn't want to do either, we could of course simply document that changing these modifiers could cause undesired behavior or impacts on the growth of the other.

Both of the first two would be causing modifier edits to branch off of vanilla growth in more than just probability or speed, and I'm not sure if either option would really be better. I think I'd personally prefer the first option as that would also make a cactus flower growth modifier more useful and dynamic as you could then make it 100% likely for a cactus flower to grow on any random tick as an example. Of course now that its no longer a relatively low-compromise change and any option would diverge off vanilla in more signifigant wats, I think it makes sense to ask for more input regarding how to handle this.

PR #12272, opened 2025-03-11
Update PlayerAdvancementDoneEvent.java

BrihtaKai:
PlayerAdvancementDoneEvent can be cancelled.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/player/PlayerAdvancementDoneEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/player/PlayerAdvancementDoneEvent.java
@@ -2,6 +2,7 @@ package org.bukkit.event.player;
 
 import org.bukkit.advancement.Advancement;
 import org.bukkit.entity.Player;
+import org.bukkit.event.Cancellable;
 import org.bukkit.event.HandlerList;
 import org.jetbrains.annotations.NotNull;
 // Paper start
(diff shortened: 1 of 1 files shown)


Leguan16:
Hey,

For this to actually have any effect you would also need to change the server code.

notTamion:
Is there any specific reason you would like for this event to be cancellable? PlayerAdvancementCriterionGrantEvent, which is fired right before the PlayerAdvancementDoneEvent already allows cancellation.

BrihtaKai:
> Is there any specific reason you would like for this event to be cancellable? PlayerAdvancementCriterionGrantEvent, which is fired right before the PlayerAdvancementDoneEvent already allows cancellation.

For more flexible advancement management.

notTamion:
I mean i am not 100 % against making it cancellable. Just saying there is already a way to do it (which is more flexible than the PlayerAdvancementDoneEvent since its called not only for done advancements)

BrihtaKai:
> Hey,
> 
> For this to actually have any effect you would also need to change the server code.

Where?

notTamion:
> Where?

please use our discord for help regarding contributing. You have to find where PlayerAdvancementDoneEvent is currently being called and find a way to nicely implement it

electronicboy:
I don't think that such an event being cancellable makes sense; You've already got all the criteria and are considered by everything to have "won" the advancement, this just seems like it would moreso be just disabling the side-effect triggers of that?

notTamion:
Given whats been said above, is there anything you would like to add to that or can we close this PR?

BrihtaKai:
> Given whats been said above, is there anything you would like to add to that or can we close this PR?

No. You can close.

PR #12273, opened 2025-03-11
Add configuration interface to expose certain config values

NonSwag:
Previously, people (I) have been using the spigot and paper file configurations to determine whether the server is configured for proxy online mode
Now the spigot interface got deprecated for good reasons, but there is no alternative
  
image
  
This commit introduces methods to check the online mode and enablement status of Velocity and BungeeCord. These changes allow plugins to easily query proxy-related server configurations.

I would like to have some input on the method names since Bukkit uses **get**OnlineMode 
Also, I wasn't really sure on whether to return bungee/velocity online mode true only if the server is actually using bungee or velocity


kennytv (approved):

Machine-Maker:
I wouldn't be immediately opposed to a new interface where **select** values from the configuration can be exposed more directly. I wouldn't want even more methods on Bukkit/Server for this though. I think there are probably these and several more values which we can consider stable enough to expose in the API.

NonSwag:
> I wouldn't be immediately opposed to a new interface where **select** values from the configuration can be exposed more directly. I wouldn't want even more methods on Bukkit/Server for this though. I think there are probably these and several more values which we can consider stable enough to expose in the API.

I like that idea, let me cook something up

NonSwag:
I've added a new interface `ServerConfiguration`

kennytv (approved):

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
+package io.papermc.paper.configuration;
+
+public interface ServerConfiguration {
```
Warriorrrr, line 9: Probably wanna note here that this class doesn't aim to expose every single config option in existance, only select ones.

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Gets whether the Server is behind a proxy that uses online mode.
+     *
+     * @return true if the server is in proxied online mode, false otherwise
+     */
+    boolean isProxyOnlineMode();
+
+    /**
+     * Gets whether the Server is configured for a Velocity proxy
```
mbax: Period at end of this line

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return true if the server has Velocity enabled; false otherwise.
```
mbax: Remove period at end of this line. Also, semicolon? Are we that fancy?

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    boolean isVelocityEnabled();
+    /**
+     * Gets whether the Server is configured for BungeeCord online mode
```
mbax: Add period at end of this line

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return true if the server is in BungeeCord online mode; false otherwise.
```
mbax: Remove period at end of this line. Also, semicolon? Are we that fancy?

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    boolean isBungeeCordEnabled();
+
+    /**
+     * Gets whether the Server is configured for Velocity online mode
```
mbax: Add period at end of this line

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return true if the server is in Velocity online mode; false otherwise.
```
mbax: Remove period at end of this line. Also, semicolon? Are we that fancy?

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    boolean isVelocityOnlineMode();
+
+    /**
+     * Gets whether the Server is configured for a BungeeCord proxy
```
mbax: Add period at end of this line

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return true if the server has BungeeCord enabled; false otherwise.
```
mbax: Remove period at end of this line. Also, semicolon? Are we that fancy?

paper-server/src/main/java/io/papermc/paper/configuration/PaperServerConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public boolean isBungeeCordEnabled() {
+        return SpigotConfig.bungee;
+    }
+
+    @Override
+    public boolean isVelocityOnlineMode() {
+        return GlobalConfiguration.get().proxies.velocity.onlineMode;
```
mbax: Is this actually only true if the proxy stuff is overall enabled? The new method javadocs appear to conflict with this code. The statement "is in Velocity online mode" implies, to me, that the server has velocity enabled *and* has the online-mode setting true in that section. Same for the bungee one.

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
mbax: Also this one says "Server" while most other things say "server"

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
mbax: Also this one says "Server" while most other things say "server"

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
mbax: Also this one says "Server" while most other things say "server"

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
mbax: Also this one says "Server" while most other things say "server"

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,38 @@
+package io.papermc.paper.configuration;
+
+public interface ServerConfiguration {
+
+    /**
+     * Gets whether the Server is behind a proxy that uses online mode.
```
mbax: Also this one says "Server" while most other things say "server"

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return true if the server has Velocity enabled, false otherwise
+     */
+    boolean isVelocityEnabled();
+
+    /**
+     * Gets whether the server is configured for BungeeCord online mode.
+     *
+     * @return true if the server is in BungeeCord online mode, false otherwise
+     */
+    boolean isBungeeCordEnabled();
```
Warriorrrr: This javadoc and the one for bungee online mode are swapped.

Warriorrrr (approved):

electronicboy:
While I approve of the general concept of this, I'm not entirely fond about having highly specific stuff exposed if the context starts relating into stuff like security aspects, etc; i.e. what happens if we add a brand new forwarding mechanism that has its own option, does that get merged into the existing one or do we have an entire open hole? Do we really need this level of accessibility?

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
Andre601: I feel like "selected critical settings." still makes it rather vague. Like why exactly are those critical? Are they critical for the server? The plugin dev? Both?

Maybe mentioning why they are considered critical would be beneficial here.

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+ * This interface doesn't aim to cover every possible server configuration
+ * option but focuses on selected critical settings.
+ */
+public interface ServerConfiguration {
+
+    /**
+     * Gets whether the server is in {@link org.bukkit.Server#getOnlineMode() online mode}
+     * or behind a proxy that uses online mode.
+     *
+     * @return true if the server is in proxied online mode, false otherwise
```
Andre601: The paragraph and return statement contradict each other.

Like which is it? Is it returning true if server is in online mode or behind a online mode proxy, OR is it only returning true if behind a online mode proxy?

paper-api/src/main/java/io/papermc/paper/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return true if the server is in Velocity online mode, false otherwise
+     */
+    boolean isVelocityOnlineMode();
+
+    /**
+     * Gets whether the server is configured for BungeeCord online mode.
+     *
+     * @return true if the server is in BungeeCord online mode, false otherwise
+     */
+    boolean isBungeeCordOnlineMode();
```
Andre601: I feel like it would be worth to add a note that these also check if the proxy mode is enabled to begin with, given both return only true if their proxy mode is enabled and in online mode.

Andre601 (commented):
Some comments regarding the comments.

NonSwag:
> While I approve of the general concept of this, I'm not entirely fond about having highly specific stuff exposed if the context starts relating into stuff like security aspects, etc; i.e. what happens if we add a brand new forwarding mechanism that has its own option, does that get merged into the existing one or do we have an entire open hole? Do we really need this level of accessibility?

As written in the ServerConfiguration docs, it only aims to provide the most important settings
In my opinion there is nothing else that is really important to be exposed here as well
but I think this is better discussed internally 

> "This interface doesn't aim to cover every possible server configuration option but focuses on selected critical settings."

electronicboy:
these settings are important why? What is the merit of exposing them individually rather than exposing the underlying concepts behind them?

NonSwag:
I don't think I can follow. What do you mean by "the underlying concepts"?
The problem is that it isn't possible to check whether the server is configured for online mode and changing the Server#getOnlineMode impl seems iffy
So exposing them here/at all is kinda _critical_

Andre601 (approved):

electronicboy:
Yes, so, isProxyOnline mode makes sense, individual bungee/velocity forwarding mode ones, I'm not sure about

NonSwag:
You are right, thinking about it, I don't see a real upside in exposing them.
I think I am going to remove them

NonSwag:
I removed the forwarding mode related methods
I would suggest _for the distant future,_ moving stuff like getPort and getIp and other methods from the Bukkit interface to this class to unclutter that mess of a class

Andre601:
> Yes, so, isProxyOnline mode makes sense, individual bungee/velocity forwarding mode ones, I'm not sure about

It *could* be useful in case there would be Velocity-only features to use through Paper in the future.
Tho, that could also be checked via some message channel stuff I suppose.

lynxplay:
If there are such features in the future, we can expand this type :wink:

electronicboy:
I mean, the two major things I'd imagine plugins might want to see on this front are "Are we in online mode?" and "Are we behind a proxy?" disambiguation is complex, I generally wouldn't want us to be in a state where we add support for some new proxy mode and now every plugin relying on this API to check that is broken

lynxplay (approved):

Owen1212055 (approved):

NonSwag:
I just realized there actually is a real use case for having a method to check whether the server is configured for a proxy
I have a plugin which provides a lobby command only if the server is configured for a proxy
At the moment I use the deprecated config methods and having a method for that as well would solve that in a more fashionable way
What do you guys think? What about a generic "isProxyEnabled" method, not tied to bungee or velocity?

PR #12276, opened 2025-03-12
[ci skip] Make compilation logs actually readable

kennytv:


Machine-Maker:
I mean ideally we should just actually fix all the warnings, and suppress the ones that should be suppressed... but maybe that's too much effort idk.

kennytv:
There's A LOT... though most we could get rid of if we disabled the necessity for every enum/public static field to have javadoc comments, which I definitely feel like aren't always needed anyways

Machine-Maker:
those sound like different warnings. I thought this was compile warnings. Not javadoc warnings

kennytv:
Ah whoops I mixed that up. Instead adding `options.compilerArgs.addAll(listOf("-Xlint:-deprecation", "-Xlint:-removal"))` would be good enough for the compilation warnings

Machine-Maker (approved):

Warriorrrr (approved):

PR #12277, opened 2025-03-12
Add ArmadilloStateChangeEvent

### LINKED ISSUE #12174: Add event for armadillo state changes

At present, it's not possible to detect when an armadillo changes its state and cancel it. This is problematic for vanish plugins, as they can't cancel the state change caused by a vanished player.

I suggest introducing a new cancellable `ArmadilloStateChangeEvent`, similar to the `WardenAngerChangeEvent`.

#12026 seems to be a good basis for this feature request, as it adds API features for armadillos (such as an enum for the armadillo state).

N/A

### Other

_No response_

orang3i:
closes #12174 
depends on #12031

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
@@ -1,0 +1,60 @@
+package io.papermc.paper.event.entity;
+
+import org.bukkit.entity.Armadillo;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.entity.EntityEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Called when an armadillo's state changes.
+ * <p>
+ * If the event is cancelled, the armadillo's state will not change.
+ */
+@NullMarked
+public class ArmadilloStateChangeEvent extends EntityEvent {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Armadillo.State oldState;
+    private final Armadillo.State newState;
+
+    @ApiStatus.Internal
+    public ArmadilloStateChangeEvent(final Armadillo armadillo, final Armadillo.State oldState, final Armadillo.State newState) {
+        super(armadillo);
+        this.oldState = oldState;
+        this.newState = newState;
+    }
+
+    /**
+     * Gets the old state of the armadillo.
+     *
+     * @return the previous armadillo state
+     */
+    public Armadillo.State getOldState() {
+        return this.oldState;
+    }
+
+    /**
+     * Gets the new state that the armadillo is transitioning to.
+     *
+     * @return the new armadillo state
+     */
+    public Armadillo.State getNewState() {
+        return this.newState;
+    }
+
+    @Override
+    public Armadillo getEntity() {
+        return (Armadillo) super.getEntity();
+    }
+
+    @Override
+    public HandlerList getHandlers() {
+        return HANDLER_LIST;
+    }
+
+    public static HandlerList getHandlerList() {
+        return HANDLER_LIST;
+    }
+}
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -16,7 +16,7 @@ public interface Armadillo extends Animals {
     State getState();
 
     /**
-     * Attempt to roll up if it in {@link State#IDLE}
+     * Attempt to roll up if the armadillo is {@link State#IDLE}
      */
     void rollUp();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -35,7 +35,7 @@ public interface Armadillo extends Animals {
     void rollOut();
 
     /**
-     * Represents the current state of the Armadillo.
+     * Represents the current state of the armadillo.
      */
     enum State {
         IDLE,
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -21,7 +21,7 @@ public interface Armadillo extends Animals {
     void rollUp();
 
     /**
-     * Attempt to roll out if it's not in {@link State#IDLE}
+     * Attempt to roll out if the armadillo is not {@link State#IDLE}
      */
     void rollOut();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -1,8 +1,47 @@
 package org.bukkit.entity;
 
+import org.jspecify.annotations.NullMarked;
+
 /**
  * Represents an Armadillo.
  */
+@NullMarked
 public interface Armadillo extends Animals {
+
+    /**
+     * Get the current state of the armadillo.
+     *
+     * @return the state of the armadillo
+     */
+    public State getState();
+
+    /**
+     * Set a new state for the armadillo.
+     * <br>
+     * This will also make the armadillo make the transition to the new state.
+     *
+     * @param state the new state
+     */
+    public void setState(final State state);
+
+    /**
+     * Attempt to rollUp if it in {@link State#IDLE}
+     */
+    public void rollUp();
+
+    /**
+     * Attempt to rollUp if it's not in {@link State#IDLE}
+     */
+    public void rollOut();
+
+    /**
+     * Represents the current state of the Armadillo.
+     */
+    public enum State {
+        IDLE,
+        ROLLING,
+        SCARED,
+        UNROLLING;
+    }
 
 }
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -25,12 +25,12 @@ public interface Armadillo extends Animals {
     public void setState(final State state);
 
     /**
-     * Attempt to rollUp if it in {@link State#IDLE}
+     * Attempt to roll up if it in {@link State#IDLE}
      */
     public void rollUp();
 
     /**
-     * Attempt to rollUp if it's not in {@link State#IDLE}
+     * Attempt to roll out if it's not in {@link State#IDLE}
      */
     public void rollOut();
 
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -16,13 +16,14 @@ public interface Armadillo extends Animals {
     State getState();
 
     /**
-     * Set a new state for the armadillo.
-     * <br>
-     * This will also make the armadillo make the transition to the new state.
-     *
-     * @param state the new state
+     * Attempt to roll up if it in {@link State#IDLE}
      */
-    void setState(final State state);
+    void rollUp();
+
+    /**
+     * Attempt to roll out if it's not in {@link State#IDLE}
+     */
+    void rollOut();
 
     /**
      * Represents the current state of the armadillo.
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -25,16 +25,6 @@ public interface Armadillo extends Animals {
     void setState(final State state);
 
     /**
-     * Attempt to roll up if it in {@link State#IDLE}
-     */
-    void rollUp();
-
-    /**
-     * Attempt to roll out if it's not in {@link State#IDLE}
-     */
-    void rollOut();
-
-    /**
      * Represents the current state of the armadillo.
      */
     enum State {
--- a/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Armadillo.java
@@ -1,8 +1,57 @@
 package org.bukkit.entity;
 
+import org.jspecify.annotations.NullMarked;
+
 /**
  * Represents an Armadillo.
  */
+@NullMarked
 public interface Armadillo extends Animals {
+
+    /**
+     * Get the current state of the armadillo.
+     *
+     * @return the state of the armadillo
+     */
+    State getState();
+
+    /**
+     * Attempt to roll up if the armadillo is {@link State#IDLE}
+     * If the armadillo's state is currently frozen, this operation will unfreeze it.
+     */
+    void rollUp();
+
+    /**
+     * Attempt to roll out if the armadillo is not {@link State#IDLE}
+     * If the armadillo's state is currently frozen, this operation will unfreeze it.
+     */
+    void rollOut();
+
+    /**
+     * Freeze the current state of the armadillo and prevents the AI from changing the states.
+     * {@link #rollUp()} and {@link #rollOut()} can still change the state.
+     */
+    void freezeState();
+
+    /**
+     * Unfreezes the armadillo, allowing the AI to change states.
+     */
+    void unFreezeState();
+
+    /**
+     * Checks if the armadillo is frozen.
+     * @return {@code true} if the state is frozen; {@code false} otherwise.
+     */
+    boolean isFrozen();
+
+    /**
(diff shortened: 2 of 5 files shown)


NonSwag:
I am not sure if the issues with the other PR have already been resolved
There was a bug with armadillos not keeping their state 
Also personally I would prefer a generale event that includes all entities _but that's probably not easily possible since states are an awful mess_

paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
```diff
@@ -0,0 +1,85 @@
...
+public class ArmadilloStateChangeEvent extends EntityEvent implements Cancellable {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private final Armadillo.State oldState;
+    private Armadillo.State newState;
+    private boolean cancelled;
+
+    @ApiStatus.Internal
+    public ArmadilloStateChangeEvent(final Armadillo armadillo, final Armadillo.State oldState, final Armadillo.State newState) {
```
NonSwag, line 23: Can't we get the old state from the armadillo itself instead of passing it to the constructor?

Doc94:
> I am not sure if the issues with the other PR have already been resolved There was a bug with armadillos not keeping their state Also personally I would prefer a generale event that includes all entities _but that's probably not easily possible since states are an awful mess_

if not wrong Lulu fix the issues in my PR for the roll methods and has Kenny approve then i think is all resolved in that xd

paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
```diff
@@ -0,0 +1,85 @@
...
+        this.newState = newState;
+    }
+
+    /**
+     * Gets the old state of the armadillo.
+     *
+     * @return the previous armadillo state
+     */
+    public Armadillo.State getOldState() {
+        return this.oldState;
```
NonSwag, line 35: I would suggest removing the old state from the event completely and only provide the new state
If you want to get the previous/current state you can just access the armadillo

paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Sets the new state for the armadillo.
+     * <p>
+     * You can modify the outcome of the state change by setting a different state.
+     *
+     * @param newState the new state to set, must not be null
+     */
+    public void setNewState(final Armadillo.State newState) {
+        Preconditions.checkNotNull(newState, "newState cannot be null");
```
Doc94:
```suggestion
        Preconditions.checkArgument(newState != null, "newState cannot be null");
```
Because checkNotNull throws NPE and need throw [IllegalArgumentException](http://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html?is-external=true) (based in another methods like this)

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/Armadillo.java.patch
```diff
@@ -1,5 +1,26 @@
 --- a/net/minecraft/world/entity/animal/armadillo/Armadillo.java
 +++ b/net/minecraft/world/entity/animal/armadillo/Armadillo.java
+@@ -109,7 +_,19 @@
+     }
+ 
+     public void switchToState(Armadillo.ArmadilloState state) {
+-        this.entityData.set(ARMADILLO_STATE, state);
++         //Paper start - add ArmadilloStateChangeEvent
```
Doc94:
```suggestion
+         // Paper start - add ArmadilloStateChangeEvent
```

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/Armadillo.java.patch
```diff
@@ -1,5 +1,26 @@
...
++             return;
++         }
++        io.papermc.paper.event.entity.ArmadilloStateChangeEvent event = new io.papermc.paper.event.entity.ArmadilloStateChangeEvent((org.bukkit.entity.Armadillo) this.getBukkitEntity(), org.bukkit.craftbukkit.entity.CraftArmadillo.stateToBukkit(oldState),org.bukkit.craftbukkit.entity.CraftArmadillo.stateToBukkit(state));
++        org.bukkit.Bukkit.getPluginManager().callEvent(event);
++        if (event.isCancelled()) {
++            return;
++        }
++        Armadillo.ArmadilloState finalState = org.bukkit.craftbukkit.entity.CraftArmadillo.stateToNMS(event.getNewState());
++        this.entityData.set(ARMADILLO_STATE, finalState);
++        //Paper end
```
Doc94:
```suggestion
+        // Paper end
```

paper-api/src/main/java/io/papermc/paper/event/entity/ArmadilloStateChangeEvent.java
orang3i: It is designed in the same way WardenAngerChangeEvent is designed. you can either get anger from the event or directly from warden.

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/Armadillo.java.patch
```diff
@@ -1,5 +1,26 @@
...
+     public void switchToState(Armadillo.ArmadilloState state) {
+-        this.entityData.set(ARMADILLO_STATE, state);
++         //Paper start - add ArmadilloStateChangeEvent
++         ArmadilloState oldState = this.getState();
++         if (oldState == state) {
++             return;
++         }
++        io.papermc.paper.event.entity.ArmadilloStateChangeEvent event = new io.papermc.paper.event.entity.ArmadilloStateChangeEvent((org.bukkit.entity.Armadillo) this.getBukkitEntity(), org.bukkit.craftbukkit.entity.CraftArmadillo.stateToBukkit(oldState),org.bukkit.craftbukkit.entity.CraftArmadillo.stateToBukkit(state));
++        org.bukkit.Bukkit.getPluginManager().callEvent(event);
++        if (event.isCancelled()) {
```
Doc94:
```suggestion  
+        if (!event.callEvent()) {
```
make with GH please check this xd

orang3i: Thanks! replaced that and fixed other formatting stufs

Lulu13022002:
The indent is mangled and i'm not sure if the event should allow mutation here, cause it really spam when cancelled/state is set to another. The main issue is that we lack a brain api i think.

orang3i:
> The indent is mangled and i'm not sure if the event should allow mutation here, cause it really spam when cancelled/state is set to another. The main issue is that we lack a brain api i think.

for now let me remove mutation!

orang3i:
> > The indent is mangled and i'm not sure if the event should allow mutation here, cause it really spam when cancelled/state is set to another. The main issue is that we lack a brain api i think.
> 
> for now let me remove mutation and cancellation.

let me also try the fix used in Armadillo#rollOut() methods

orang3i:
Instead of using memory, it just uses a variable now, so it is non persistant. Also fixed sound spamming

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java.patch
```diff
@@ -0,0 +1,72 @@
...
++                            this.nextPeekTimer = this.pickNextPeekTimer(owner);
++                        }
++
++                        if (timeUntilExpiry < Armadillo.ArmadilloState.UNROLLING.animationDuration()) {
++                            owner.playSound(SoundEvents.ARMADILLO_UNROLL_START);
++                            owner.switchToState(Armadillo.ArmadilloState.UNROLLING);
++                        }
++                    } else if (state == Armadillo.ArmadilloState.UNROLLING && timeUntilExpiry > Armadillo.ArmadilloState.UNROLLING.animationDuration()) {
++                        owner.switchToState(Armadillo.ArmadilloState.SCARED);
++                    }
```
notTamion: use guard clauses here and in the other applicable methods.

notTamion (changes requested):
I think i would prefer having a boolean which freezes the state for the ai instead of having it be an entirely different state. And then we allow mutation of the state even when frozen using the normal api methods

orang3i:
> I think i would prefer having a boolean which freezes the state for the ai instead of having it be an entirely different state. And then we allow mutation of the state even when frozen using the normal api methods

Yeah, changed to boolean now, and added methods to freeze and unfreeze using normal api.

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java.patch
```diff
@@ -0,0 +1,24 @@
...
++
+             super.tick(level, owner, gameTime);
+             if (this.nextPeekTimer > 0) {
+                 this.nextPeekTimer--;
+@@ -196,7 +_,7 @@
+                 this.dangerWasAround = flag;
+                 if (state == Armadillo.ArmadilloState.SCARED) {
+                     if (this.nextPeekTimer == 0 && owner.onGround() && flag) {
+-                        level.broadcastEntityEvent(owner, (byte)64);
++                        level.broadcastEntityEvent(owner, (byte) 64);
```
NonSwag: Extra space

orang3i: image
that is weird.... in my ide, it is correct and i don't think i touched it. it probably has smth to do with building patches.

orang3i: @NonSwag nvm, i think its because i pressed ctrl+alt+L to format it. Leme fix. Thx!

notTamion:
I was more of thinking of only having the boolean only affect the ai and letting the normal rollUp and other methods continue to work as usual no need to add extra methods for it.

orang3i:
> I was more of thinking of only having the boolean only affect the ai and letting the normal rollUp and other methods continue to work as usual no need to add extra methods for it.

Then there would be no way to force a state no?

if the state is frozen and rollUp() is used normally, then the state will be unfrozen.

notTamion:
no we just add an overload to the nms rollup/rollout with a boolean ignoreFrozen and that one is used by the api to still change state if the armadillo state is frozen

> I think it would be best if there are seperate methods for frozen states

why though? it seems very unnecessary to me to have multiple methods which essentially do the same thing.

besides wouldn't this freezing api be better implemented in the original PR from doc?

orang3i:
> no we just add an overload to the nms rollup/rollout with a boolean ignoreFrozen and that one is used by the api to still change state if the armadillo state is frozen

Okay that makes sense, i will change that.

orang3i:
> no we just add an overload to the nms rollup/rollout with a boolean ignoreFrozen and that one is used by the api to still change state if the armadillo state is frozen
> 
> > I think it would be best if there are seperate methods for frozen states
> 
> why though? it seems very unnecessary to me to have multiple methods which essentially do the same thing.
> 
> besides wouldn't this freezing api be better implemented in the original PR from doc?

Well, now i overloaded the rollUp() and rollOut() methods with boolean ignoreFrozen instead of freezeRollUp(0 and freezeRollOut()

orang3i:
@notTamion  Let me know, if we want to totally remove the overload and make so that rollUp() and rollOut() bypasses freeze and then if they want to keep it frozen they just run freezeState() again.

notTamion:
Yeah well the current implementation isn't exactly what i had in mind either but yeah. Make api rollUp and rollOut just unfreeze the state.

orang3i:
> EDIT: still think the freezing api should just go into @Doc94 PR

@notTamion this just adds on to doc's pr. it doesn't remove that.

orang3i:
@notTamion  Now, the rollUp() and rollOut() unfreezes the armadillo.

If you want, once things are approved i can pr all of this into doc's fork

paper-api/src/main/java/org/bukkit/entity/Armadillo.java
```diff
@@ -1,8 +1,54 @@
...
 
+    /**
+     * Get the current state of the armadillo.
+     *
+     * @return the state of the armadillo
+     */
+    State getState();
+
+    /**
+     * Attempt to roll up if the armadillo is {@link State#IDLE}
```
notTamion, line 19: should mention this will unfreeze state if frozen. same for rollUp

paper-api/src/main/java/org/bukkit/entity/Armadillo.java
```diff
@@ -1,8 +1,54 @@
...
+     */
+    void rollUp();
+
+    /**
+     * Attempt to roll out if the armadillo is not {@link State#IDLE}
+     */
+    void rollOut();
+
+    /**
+     * Freeze the current state of the armadillo.
```
notTamion: Mention that it only happens for Ai changes like you do in the unfreeze method

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java
++++ b/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java
+@@ -175,6 +_,12 @@
+ 
+         @Override
+         protected void tick(ServerLevel level, Armadillo owner, long gameTime) {
++            // Paper start - checks if the armadillo is frozen.
++            if (owner.isFrozen()) {
++                return;
```
notTamion: i think having an this early return here will actually disable peeking which ig is something we don't really want to disable as its not exactly an state change

notTamion (changes requested):
yea just leave it on this PR for now we can discuss that later

notTamion (changes requested):
yea just leave it on this PR for now we can discuss that later

EDIT: ignore github sent it twice for some reason

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java.patch
orang3i: Fixed it, now peaking works. Just added the early returns just before state change instead of it being at the head of #tick()

paper-server/patches/sources/net/minecraft/world/entity/animal/armadillo/ArmadilloAi.java.patch
```diff
@@ -0,0 +1,23 @@
...
+@@ -201,10 +_,12 @@
+                     }
+ 
+                     if (timeUntilExpiry < Armadillo.ArmadilloState.UNROLLING.animationDuration()) {
++                        if (owner.isFrozen()) { return; } // Paper - checks if the armadillo is frozen.
+                         owner.playSound(SoundEvents.ARMADILLO_UNROLL_START);
+                         owner.switchToState(Armadillo.ArmadilloState.UNROLLING);
+                     }
+                 } else if (state == Armadillo.ArmadilloState.UNROLLING && timeUntilExpiry > Armadillo.ArmadilloState.UNROLLING.animationDuration()) {
++                    if (owner.isFrozen()) { return; } // Paper - checks if the armadillo is frozen.
```
NonSwag: Why `{ return; }`?
Instead of 
```java
{
   return;
}
```
Or
`if (...) return;`

orang3i: Well i just prefer `if ( .. ) { return; }` over `if ( .. ) return;`

and

```
{
   return;
}
```
 
 would mean addition of so many extra lines.
 
 but if it is not upto paper standards im fine with changing to whatever is requested.

NonSwag: It's not really best practice and very uncommon to use `{return;}`
I am pretty sure that when the checkstyle PR is merged paper will follow the standard of using line breaks while also allowing no braces

orang3i: Okay, thanks will change now.

orang3i:
@notTamion The requested changes are done. lmk how we proceed with this PR ( with respect to @Doc94 's #12031 )

orang3i:
Rebased to 1.21.5 🥳

PR #12288, opened 2025-03-14
Restore ArmorStand items if EntityDeathEvent is cancelled

### LINKED ISSUE #12286: Cancelling EntityDeathEvent for an Armorstand Loses Items
### Expected behavior

When in survival mode, I expect the armorstand to not lose their items when killing the armorstand with the event cancelled.

### Observed/Actual behavior

The items get deleted.

### Steps/models to reproduce

 ```java
@EventHandler
    public void killEvent(EntityDeathEvent event) {
        event.setCancelled(true);
    }
```

### Plugin and Datapack List

N/A

### Paper version

1.21.4

### Other

_No response_

Doc94:
Closes #12286 
This PR make a copy of the items removed and added to drop (hand/armor) and check if the event was cancelled for restore, also double method called maybe for the hardfork thing.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/decoration/ArmorStand.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/decoration/ArmorStand.java.patch
@@ -198,7 +198,7 @@
                  }
  
                  return true;
-@@ -400,31 +_,34 @@
+@@ -400,31 +_,42 @@
          float health = this.getHealth();
          health -= damageAmount;
          if (health <= 0.5F) {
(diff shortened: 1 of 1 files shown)


notTamion:
I think it would be better if we moved the removal of the items past the dropAllDeathLoot call so we don't remove those items at all before we know the event was cancelled. this way. this would also restore parity with other entities which also have the event called with their equipment/items still on

Doc94:
> I think it would be better if we moved the removal of the items past the dropAllDeathLoot call so we don't remove those items at all before we know the event was cancelled. this way. this would also restore parity with other entities which also have the event called with their equipment/items still on

that not require a littlle more refactor for another behaviours rather than just the armorstand?

Doc94:
> I was thinking about something like this: https://pastes.dev/3ZkbtqLipa

hmm yeah that works too.. i wanna add the check for valid slots BUT... API allow set saddle/body but item replace command not but that is another thing.

notTamion (approved):

kennytv (approved):

Owen1212055 (approved):

lynxplay (approved):

PR #12296, opened 2025-03-14
Fix unnecessary map data saves (#12295)

### LINKED ISSUE #12295: MapCanvas#drawImage unnecessarily causes map data file to save
### Expected behavior

.dat file isn't marked dirty since `drawImage` drawing the image only modifies the buffer inside the craft canvas and not the underlying colors array the game saves

### Observed/Actual behavior

.dat file is marked dirty, causing it to save unnecessarily 

### Steps/models to reproduce

- Use MapCanvas#drawImage on a map with a custom MapRenderer

### Plugin and Datapack List

none

### Paper version

latest. minecraft 1.21.4, paper build 211

### Other

This can cause large lag spikes (of more than 10s in some cases) on servers largely utilizing imageonmap-like plugins

Dqu1J:
Fixes #12295 

Currently, MapCanvas API unnecessarily causes map data file to save.

The API calls `this.mapView.worldMap.setColorsDirty()` to mark colors dirty for players, however this has the side-effect of saving the map .dat files as well:
```java
public void setColorsDirty(int x, int z) {
    this.setDirty(); // This saves the data file! 

    for (MapItemSavedData.HoldingPlayer holdingPlayer : this.carriedBy) {
        holdingPlayer.markColorsDirty(x, z); // This is what the API wants to do!
    }
}
```

This causes unnecessary lag during world saves, which scales with the amount of maps handled with API since last save. On servers that heavily rely on imageonmap-like plugins, it can cause lag spikes of dozens of seconds.

This PR changes this method to add a boolean argument to the method that determines whether the file is saved or not, which is used by the API. An overload is added for compatibility, and for vanilla.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java.patch
@@ -96,6 +96,20 @@
      }
  
      public static void addTargetDecoration(ItemStack stack, BlockPos pos, String type, Holder<MapDecorationType> mapDecorationType) {
+@@ -354,7 +_,12 @@
+     }
+ 
+     public void setColorsDirty(int x, int z) {
+-        this.setDirty();
++    // Paper start - Fix unnecessary map data saves
++        this.setColorsDirty(x, z, true);
++    }
++    public void setColorsDirty(int x, int z, boolean markFileDirty) {
++        if (markFileDirty) this.setDirty();
++    // Paper end - Fix unnecessary map data saves
+ 
+         for (MapItemSavedData.HoldingPlayer holdingPlayer : this.carriedBy) {
+             holdingPlayer.markColorsDirty(x, z);
 @@ -395,7 +_,7 @@
                  return true;
              }
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
@@ -64,7 +64,7 @@ public class CraftMapCanvas implements MapCanvas {
             return;
         if (this.buffer[y * 128 + x] != color) {
             this.buffer[y * 128 + x] = color;
-            this.mapView.worldMap.setColorsDirty(x, y);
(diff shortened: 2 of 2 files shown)


paper-server/patches/sources/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java.patch
```diff
@@ -141,6 +141,31 @@
...
++        if (markFileDirty) {
++            this.setDirty();
++        }
+ 
+         for (MapItemSavedData.HoldingPlayer holdingPlayer : this.carriedBy) {
+             holdingPlayer.markColorsDirty(x, z);
+         }
+     }
+ 
++    public void setColorsDirty(int x, int z) {
```
electronicboy: put this method just above the modified method

paper-server/src/main/java/org/bukkit/craftbukkit/map/CraftMapCanvas.java
```diff
@@ -64,7 +64,7 @@ public void setPixel(int x, int y, byte color) {
             return;
         if (this.buffer[y * 128 + x] != color) {
             this.buffer[y * 128 + x] = color;
-            this.mapView.worldMap.setColorsDirty(x, y);
+            this.mapView.worldMap.setColorsDirty(x, y, false); // Paper - Fix unnecessary map data saves
```
FireInstall, line 67: Is it really safe to never save the data after any changes where done here?
I mean wouldn't that break the expectations of current plugins?
And if yes, maybe pulling up a "doNotSave" optional parameter into the API?

Dqu1J: @FireInstall custom renderers never end up saving data to .dat files afaik
quoting what Warriorrrr told me on discord, "drawing the image only modifies the buffer inside the craft canvas and not the underlying colors array the game saves"

FireInstall: alright, if not only drawing but also chaning pixels should not save, then thanks!
In that case this can get marked as resolved. Have a nice day ~

Warriorrrr (approved):

kennytv (approved):

Owen1212055 (approved):

PR #12301, opened 2025-03-15
Configuration API

Owen1212055:
Please provide feedback on naming, and general event consensis. 

But my goal is to have the following:
- Full cookie getting/setting support in the login/config/play stage. This means you'll be able to disconnect players during LOGIN depending on sent cookies. (Bukkit only allows play stage, and PARTLY config stage) (`CookieConnection`)
- Allowing to boot players into the config stage/ out of it
- Allow to add custom configuration tasks, so for example being able to load player data. (`AsyncPlayerConnectionConfigurateEvent`)
- Allowing to load resource packs during config stage better
- Transfer the player in config/player
- send custom report details (shown in client debug log)


paper-api/src/main/java/io/papermc/paper/connection/CookieConnection.java
```diff
@@ -0,0 +1,29 @@
+package io.papermc.paper.connection;
+
+import org.bukkit.NamespacedKey;
+import org.jspecify.annotations.NullMarked;
+import java.util.concurrent.CompletableFuture;
+
+@NullMarked
+public interface CookieConnection extends PlayerConnection {
```
lynxplay: PlayerCookieConnection

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
```diff
@@ -0,0 +1,58 @@
...
+     * @return profile
+     */
+    PlayerProfile getProfile();
+
+    /**
+     * Gets the player IP address.
+     *
+     * @return The IP address
+     */
+    InetAddress getAddress();
```
lynxplay: This is kinda doubled between this and the login connection and then it isn't present on the game connection because player presumably has it.

While the profile may not be available, address information (Inet or Socket (hi emily)) feels very very basic?

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Gets the raw address of the player logging in
```
lynxplay: Dots at the end of jd bodies.
Goes for the rest of this PR too.

paper-api/src/main/java/io/papermc/paper/connection/CookieConnection.java
lynxplay: Experimental the living shit out of *all* of these types.

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Clears the players chat history and their local chat.
+     */
+    void clearChat();
+
+    /**
+     * Completes the configuration for this player, which will cause this player to reenter the game.
+     * <p>
+     * Note, this should be only be called if you are reconfiguring the player.
```
lynxplay, line 33: This, similar to #enterConfiguration poses the interesting issue of "ah, you just requested what is basically a destruction of this connection state, what happens to *this* instance".

Calling #clearChat after this is illegal.
Getters *might* be fine. But we should look into a proper way to signal, "you destroyed this connection instance".

paper-api/src/main/java/io/papermc/paper/connection/PlayerConnection.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Disconnects the player connection.
+     * <p>
+     * Note that calling this during connection related events may caused undefined behavior.
+     * @param component disconnect reason
+     */
+    void disconnect(Component component);
+
+    /**
+     * Gets if this connection originated from a transferred connection.
```
lynxplay, line 20: Formatting, obligatory note that this is completely controlled by the client and to be taken with a grain of salt.

paper-api/src/main/java/io/papermc/paper/connection/PlayerLoginConnection.java
```diff
@@ -0,0 +1,58 @@
...
+import org.bukkit.NamespacedKey;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.NullMarked;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+
+@NullMarked
+public interface PlayerLoginConnection extends CookieConnection  {
```
lynxplay: Extra useless space.

paper-api/src/main/java/io/papermc/paper/connection/PlayerLoginConnection.java
```diff
@@ -0,0 +1,58 @@
...
+    @Nullable
+    PlayerProfile getUnsafeProfile();
+
+    /**
+     * Gets the player IP address.
+     *
+     * @return The IP address
+     */
+    @NotNull
+    InetAddress getAddress();
```
lynxplay: hi emily

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/AsyncPlayerConnectionConfigureEvent.java
```diff
@@ -0,0 +1,33 @@
+package io.papermc.paper.event.connection.configuration;
+
+import io.papermc.paper.connection.PlayerConfigurationConnection;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A task that allows you to configurate the player.
```
lynxplay: This is an event, not a task.

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/AsyncPlayerConnectionConfigureEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * This is async and allows you to run configuration code on the player.
+ * Once this event has finished execution, the player connection will continue.
+ * <p>
+ * This occurs after configuration, but before the player has entered the world.
+ */
+public class AsyncPlayerConnectionConfigureEvent extends PlayerConfigurationConnectionEvent {
```
lynxplay: Naming here indeed hits the wall a bit.
A "AsyncPlayerConnectionConfigureEvent", representing when a player is near-finished with their configuration phase, feels rough, the parent name "PlayerConfigurationConectionEvent" makes this even worse, either it is Configuration or Configure, pick one xD
The fact they are swapped also hurts a bit.

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConnectionInitialConfigureEvent.java
```diff
@@ -0,0 +1,27 @@
+package io.papermc.paper.event.connection.configuration;
+
+import io.papermc.paper.connection.PlayerConfigurationConnection;
+import org.bukkit.Bukkit;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+public class PlayerConnectionInitialConfigureEvent extends PlayerConfigurationConnectionEvent {
+    private static final HandlerList handlers = new HandlerList();
```
lynxplay: Here and everywhere else, HANDLERS, or HANDLER_LIST.

paper-server/patches/sources/net/minecraft/network/DisconnectionDetails.java.patch
```diff
@@ -0,0 +1,17 @@
+--- a/net/minecraft/network/DisconnectionDetails.java
++++ b/net/minecraft/network/DisconnectionDetails.java
+@@ -5,7 +_,13 @@
+ import java.util.Optional;
+ import net.minecraft.network.chat.Component;
+ 
+-public record DisconnectionDetails(Component reason, Optional<Path> report, Optional<URI> bugReportLink) {
++// Paper start
```
lynxplay: identifying comments. What API is this for.

paper-server/patches/sources/net/minecraft/server/network/ServerCommonPacketListenerImpl.java.patch
```diff
@@ -1,11 +1,11 @@
 --- a/net/minecraft/server/network/ServerCommonPacketListenerImpl.java
 +++ b/net/minecraft/server/network/ServerCommonPacketListenerImpl.java
-@@ -27,30 +_,67 @@
+@@ -27,22 +_,30 @@
  import net.minecraft.util.profiling.Profiler;
  import org.slf4j.Logger;
  
 -public abstract class ServerCommonPacketListenerImpl implements ServerCommonPacketListener {
-+public abstract class ServerCommonPacketListenerImpl implements ServerCommonPacketListener, org.bukkit.craftbukkit.entity.CraftPlayer.TransferCookieConnection { // CraftBukkit
++public abstract class ServerCommonPacketListenerImpl implements ServerCommonPacketListener{
```
lynxplay: Useless diff.

lynxplay (dismissed):

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConfigurationConnectionEvent.java
```diff
@@ -0,0 +1,24 @@
+package io.papermc.paper.event.connection.configuration;
+
+import io.papermc.paper.connection.PlayerConfigurationConnection;
+import io.papermc.paper.connection.PlayerLoginConnection;
+import org.bukkit.event.Event;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+abstract class PlayerConfigurationConnectionEvent extends Event {
```
Machine-Maker: We don't need this abstract event. Just pull the one field up to the parent events.

paper-server/patches/sources/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java.patch
```diff
@@ -46,19 +70,38 @@
...
              }
  
 -            ServerPlayer playerForLogin = playerList.getPlayerForLogin(this.gameProfile, this.clientInformation);
-+            ServerPlayer playerForLogin = playerList.getPlayerForLogin(this.gameProfile, this.clientInformation, this.player); // CraftBukkit
++            // Paper start
++            ServerPlayer playerForLogin;
++            if (this.connection.savedPlayerForLoginEventLegacy != null) {
++                playerForLogin = this.connection.savedPlayerForLoginEventLegacy;
++                // WE have to do this because the player isnt updated properly
++                playerForLogin.updateOptions(this.clientInformation);
```
Machine-Maker: calling updateOptions will fire the optionschange event. Which makes that event now fire sometimes and not other times depending on if anyone is listening to PlayerLoginEvent

use updateOptionsNoEvents I think.

paper-server/patches/sources/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java.patch
```diff
@@ -70,3 +113,44 @@
              this.connection.send(new ClientboundDisconnectPacket(DISCONNECT_REASON_INVALID_DATA));
              this.connection.disconnect(DISCONNECT_REASON_INVALID_DATA);
          }
+@@ -171,7 +_,7 @@
+         }
+     }
+ 
+-    private void finishCurrentTask(ConfigurationTask.Type taskType) {
++    public void finishCurrentTask(ConfigurationTask.Type taskType) { // Paper
```
Machine-Maker: at

paper-server/patches/sources/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java.patch
```diff
@@ -1,18 +1,29 @@
 --- a/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java
 +++ b/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java
-@@ -48,8 +_,10 @@
+@@ -40,18 +_,20 @@
+ public class ServerConfigurationPacketListenerImpl extends ServerCommonPacketListenerImpl implements ServerConfigurationPacketListener, TickablePacketListener {
+     private static final Logger LOGGER = LogUtils.getLogger();
+     private static final Component DISCONNECT_REASON_INVALID_DATA = Component.translatable("multiplayer.disconnect.invalid_player_data");
+-    private final GameProfile gameProfile;
++    public final GameProfile gameProfile; // Paper
```
Machine-Maker: at

paper-server/patches/sources/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java.patch
```diff
@@ -1,18 +1,29 @@
...
+     private final Queue<ConfigurationTask> configurationTasks = new ConcurrentLinkedQueue<>();
+     @Nullable
+-    private ConfigurationTask currentTask;
+-    private ClientInformation clientInformation;
++    public ConfigurationTask currentTask; // Paper
++    public ClientInformation clientInformation; // Paper
```
Machine-Maker: at

paper-api/src/main/java/io/papermc/paper/event/player/PlayerServerFullCheckEvent.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Fires when a server is currently considered "full", and is executed for each player
+ * trying to join the server while it is in this state.
+ */
+@NullMarked
+public class PlayerServerFullCheckEvent extends Event {
```
Machine-Maker, line 38: Wouldn't this event be more useful if functioned as a possible "override" for the entire fullnest check instead of only firing when the server was full? We can use the event tristate for it, and have a method that returns whether vanlla logic would allow or deny.

paper-api/src/main/java/io/papermc/paper/connection/PlayerConnection.java
```diff
@@ -0,0 +1,20 @@
+package io.papermc.paper.connection;
+
+import net.kyori.adventure.text.Component;
+
+public interface PlayerConnection {
```
Machine-Maker, line 8: It would be really good if the hierarchy of types here was adjusted so this class could be sealed and permit the two API interfaces. Would make switching on it nicer, especially in events where you only have the PlayerConnection, and to know what type it is, you do a switch pattern match. It's not a super simple change, because of the impl CommonCookeConnection abstract class, but things could def be moved around.

paper-api/src/main/java/io/papermc/paper/connection/PlayerCommonConnection.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+import java.util.UUID;
+import net.kyori.adventure.resource.ResourcePackRequest;
+import org.bukkit.ServerLinks;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Represents a connection that has properties shared between the GAME and CONFIG stage.
+ */
+public interface PlayerCommonConnection extends CookieConnection {
```
Machine-Maker: Same situation here regarding sealed interfaces being nicer for switch pattern matching (doesn't require default case)

paper-api/src/main/java/io/papermc/paper/event/connection/common/PlayerFinishConnectionPhaseEvent.java
```diff
@@ -0,0 +1,80 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Called when a player's connection phase is effectively "finished", meaning
+ * that they are progressing to another connection stage.
+ * <p>
+ * Currently fires during the login and configuration protocol.
+ */
+public class PlayerFinishConnectionPhaseEvent extends Event {
```
Machine-Maker: Isn't this event a little redundant? It seems to just fire right after the PlayerConnectionValidateLoginEvent. Any way these can be condensed into one? If not, the javadoc comments probably need to talk about the difference between the two.

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConnectionReconfigurateEvent.java
```diff
@@ -0,0 +1,28 @@
...
+import io.papermc.paper.connection.PlayerConfigurationConnection;
+import org.bukkit.Bukkit;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Indicates that this player is being reconfigured, meaning that this connection will be held in the configuration
+ * stage unless kicked out through {@link PlayerConfigurationConnection#completeConfiguration()}
+ */
+public class PlayerConnectionReconfigurateEvent extends PlayerConfigurationConnectionEvent {
```
Machine-Maker: I feel like this event, and PlayerConnectionInitialConfigureEvent can be compiled with just an enum "cause" or something. They are just both fired right before switching to configuration phase, from either login or from game. Just a PlayerEnterConfigurationPhaseEvent or something with a cause?

Machine-Maker (dismissed):

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConnectionReconfigurateEvent.java
Owen1212055: The reason why I made this separate is because both are treated very differently. 
Basically I felt like differenciating that this player is going to be held forever in config phase vs they are going through the config phase as normal was an important enough disctinction.

paper-api/src/main/java/io/papermc/paper/event/connection/common/PlayerFinishConnectionPhaseEvent.java
Owen1212055: Yep, I think you are correct.

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConnectionReconfigurateEvent.java
Machine-Maker: But in both cases, doesn't the player just automatically go through the tasks in the configuration phase?

In plain vanilla, if you call that function to kick the player back to config phase, won't they just come right back in as if they had just logged in? Or do you have to somehow trigger something to *let* them back in?

Owen1212055: No, in reconfiguration the tasks are not run. This causes the player to just be stuck in the stage until released.
startConfiguration is only called when login packet is received, so there are no "tasks" that are being executed.

You have to "somehow" trigger to let them back in.

paper-api/src/main/java/io/papermc/paper/connection/PlayerConnection.java
Owen1212055: After playing around with this a bit, I am not sure about it really. This does require alot of moving around and possible duplication that I am not fully sure is worth it.

Additionally, its a bit of an odd situation, as I am not sure in what cases you'd be casting like this in the first place.

paper-api/src/main/java/io/papermc/paper/connection/PlayerCommonConnection.java
Owen1212055: see above

paper-api/src/main/java/io/papermc/paper/event/connection/common/PlayerFinishConnectionPhaseEvent.java
Owen1212055: See https://discord.com/channels/289587909051416579/925530366192779286/1385780483815637042

paper-api/src/main/java/io/papermc/paper/event/connection/configuration/PlayerConnectionReconfigurateEvent.java
Owen1212055: See https://discord.com/channels/289587909051416579/925530366192779286/1385790639345631333

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
Owen1212055: I agree here. But, I am not sure as the current standard has been to silently fail in many places.
image
So we could very much add some validation, but it all ends up nooping.

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
Owen1212055: Abstracted out. Thanks for this

paper-api/src/main/java/io/papermc/paper/connection/PlayerLoginConnection.java
Owen1212055: idk what this means? ;-;

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    void completeConfiguration();
```
kennytv: If it's just for reconfiguration, call it completeReconfiguration

paper-api/src/main/java/io/papermc/paper/connection/PlayerGameConnection.java
```diff
@@ -0,0 +1,21 @@
...
+
+public interface PlayerGameConnection extends PlayerCommonConnection {
+
+    /**
+     * Bumps the player to the configuration stage.
+     * <p>
+     * This will, by default, cause the player to stay until their connection is released by
+     * {@link PlayerConfigurationConnection#completeConfiguration()}
+     */
+    void enterConfiguration();
```
kennytv: reenterConfiguration?

paper-api/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java
```diff
@@ -301,6 +304,15 @@ public boolean isTransferred() {
         return this.transferred;
     }
 
+    /**
+     * Gets the connection for the player logging in.
+     * @return connection
+     */
+    @NotNull
+    public PlayerLoginConnection getPlayerLoginConnection() {
```
kennytv: Just getConnection to be in line with the other connection getters in and outside of events

paper-server/src/main/java/io/papermc/paper/connection/HorriblePlayerLoginEventHack.java
```diff
@@ -0,0 +1,90 @@
...
+                
+                This event forces an alternative player loading path that is
+                deprecated and will be REMOVED in a future release.
+                
+                This also does not properly represent the fact that this validation logic is ran
+                twice in vanilla.
+                
+                Please notify the following plugin developers: %s
+
+                Continued usage may cause instability in future builds.
```
kennytv: This should be shortened/reworded a bit, maybe use a link to the javadocs instead of that "See ..." line:

```suggestion
                WARNING: Legacy PlayerLoginEvent usage detected!
                
                This event forces an alternative player loading path that is
                deprecated and will be removed in a future release.
                See the PlayerLoginEvent javadocs for more information.
                
                Please notify the following plugin developers: %s
```

paper-api/src/main/java/org/bukkit/event/player/PlayerLoginEvent.java
```diff
@@ -14,7 +16,12 @@
  * Note that this event is called <i>early</i> in the player initialization
  * process. It is recommended that most options involving the Player
  * <i>entity</i> be postponed to the {@link PlayerJoinEvent} instead.
+ * @deprecated Use {@link PlayerConnectionValidateLoginEvent} to handle pre-login logic
+ * (e.g., authentication or ban checks), or {@link io.papermc.paper.event.player.PlayerServerFullCheckEvent} to allow
+ * players to bypass the server's maximum player limit.
```
kennytv: This should also include info on why it's deprecated

kennytv (changes requested):
also ctrl+f and remove at least half of the mentions of the word "currently" (especially in event headers)💩

kennytv (approved):

paper-api/src/main/java/io/papermc/paper/connection/PlayerConfigurationConnection.java
```diff
@@ -0,0 +1,48 @@
...
+     * To access identifying information about this player, use {@link net.kyori.adventure.pointer.Pointered}
+     * with the following pointers:
+     * <ul>
+     *   <li>{@link net.kyori.adventure.identity.Identity#UUID}</li>
+     *   <li>{@link net.kyori.adventure.identity.Identity#NAME}</li>
+     * </ul>
+     *
+     * @return the configuring player audience
+     */
+    Audience getConfiguringPlayer();
```
Machine-Maker: If its just an "Audience", how to I, in a callback where I only have the audience, get access to the methods I can use on the connection?

PR #12302, opened 2025-03-16
Skull - get/set customName

ShaneBeee:
This PR aims to add a getter/setter for the custom name of a Skull (tile entity) block.
I believe Minecraft added this fairly recently?

We were talking about this over at [Skript](https://github.com/SkriptLang/Skript/issues/6826) recently, and realized Paper didn't have API for this. 

I've never touched BlockState stuff with CraftBukkit before, so I'm hoping I did this correctly.
I was basically going off how the profile was done.... 🤞

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftSkull.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftSkull.java
@@ -2,6 +2,8 @@ package org.bukkit.craftbukkit.block;
 
 import com.google.common.base.Preconditions;
 import com.mojang.authlib.GameProfile;
+import io.papermc.paper.adventure.PaperAdventure;
+import net.kyori.adventure.text.Component;
 import net.minecraft.Util;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.MinecraftServer;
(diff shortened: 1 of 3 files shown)


Warriorrrr (commented):
I think it's worth noting that while every skull block entity can have a custom name, it will only be carried over to the item for player heads when broken in vanilla.

ShaneBeee:
> I think it's worth noting that while every skull block entity can have a custom name, it will only be carried over to the item for player heads when broken in vanilla.

I sent a message on Discord, since GitHub's video limit is so small.
<https://discord.com/channels/289587909051416579/555462289851940864/1350869621372358656>

Unless I am missing something, I believe this also works in Paper (as seen in the video)

Warriorrrr (approved):

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftSkull.java
```diff
@@ -45,6 +48,7 @@ public void load(SkullBlockEntity skull) {
         if (owner != null) {
             this.profile = owner;
         }
+        this.customName = PaperAdventure.asAdventure(skull.customName);
```
Lulu13022002: This will return an empty component when the field is null, i don't think that's what you want if the methods are nullable

Lulu13022002 (commented):
I think this should get/set the custom name from the snapshot directly like the note block sound so you don't need to call update() on live view state. The profile needs an update in all case so it doesn't matter that much here.

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftSkull.java
ShaneBeee: Ok, I fixed that issue, as well changed to the snapshot. I hope I did it correctly.

kennytv (approved):

PR #12303, opened 2025-03-16
Update visual fire handling with TriState support

### LINKED ISSUE #12084: Inversed visual fire & lock fire ticks

Currently, we have the ability to set an entity visually on fire without it being actually burning
but the way visual fire works, it is not possible to disable visual fire if the entity is really burning

Internally, the visualFire field could use a nullable Boolean object and default to null and make the entity burn if true
not burn if false
and send the actual fire state when null
I understand this will cause problems with existing plugins that utilize this method, other suggestions are more than welcome

-

### Other

Also, I would like to suggest adding a `lockFireTicks` method to be able to make entities burn or not start burning as long as the ticks are locked

NonSwag:
Replaced the Boolean-based visual fire system with TriState for improved clarity and flexibility, enabling three distinct states: TRUE, FALSE, and NOT_SET. Deprecated older methods in favor of new ones and updated internal handling to reflect these changes. Adjusted serialization and deserialization logic to accommodate the new TriState implementation.

Resolves #12084


GliczDev:
I don't think this should be a feature patch, also Paper comments are missing

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java
```diff
@@ -359,13 +360,25 @@ public void setFireTicks(int ticks) {
     }
 
     @Override
+    @Deprecated
     public void setVisualFire(boolean fire) {
-        this.getHandle().hasVisualFire = fire;
+        setVisualFire(TriState.byBoolean(fire));
```
GliczDev: shouldn't it be `UNSET` for `false` to reflect old behavior?

NonSwag: true missed that

NonSwag:
> I don't think this should be a feature patch, also Paper comments are missing

Just took another look at the code and I thought that visualFire was introduced by paper 
that's why I thought it is already in between paper start / stop comments and didn't add them

GliczDev:
> Just took another look at the code and I thought that visualFire was introduced by paper 
> that's why I thought it is already in between paper start / stop comments and didn't add them

Even if the changes are in between Paper start/end comments, new ones should also be added to indicate what the new changes are doing

paper-server/patches/features/0032-Update-visual-fire-handling-with-TriState-support.patch
```diff
@@ -0,0 +1,53 @@
...
+     }
+ 
+     public void checkBelowWorld() {
+@@ -2465,8 +2465,10 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+                 compound.putInt("TicksFrozen", this.getTicksFrozen());
+             }
+ 
+-            if (this.hasVisualFire) {
+-                compound.putBoolean("HasVisualFire", this.hasVisualFire);
++            if (!this.visualFire.equals(net.kyori.adventure.util.TriState.NOT_SET)) {
```
masmc05: I don't think interfering with vanilla serialization output is a smart choice. Looking at other serializations, I think for such cases `HasVisualFire` would take `this.visualFire.toBooleanOrElse(false)` and a `Paper.HasVisualFire` would take the TriState, the paper one taking priority when deserializing

NonSwag: The getBoolean method always returns false if there is nothing defined 
but I agree, even though I don't like adding more values to the file

NonSwag: I think I can actually leave it how I did it
the created compound is always empty, so vanilla actually only stores the visual Fire if it is enabled
I can just save it if it is not not_set

paper-server/patches/features/0032-Improve-visual-fire-API-in-Entity-class.patch
```diff
@@ -0,0 +1,55 @@
...
+             if (!this.tags.isEmpty()) {
+                 ListTag listTag = new ListTag();
+@@ -2590,7 +2592,11 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+                 this.setNoGravity(compound.getBoolean("NoGravity"));
+                 this.setGlowingTag(compound.getBoolean("Glowing"));
+                 this.setTicksFrozen(compound.getInt("TicksFrozen"));
+-                this.hasVisualFire = compound.getBoolean("HasVisualFire");
++                 // Paper start - improve visual fire API
++                this.visualFire = compound.contains("HasVisualFire")
++                    ? net.kyori.adventure.util.TriState.byBoolean(compound.getBoolean("HasVisualFire"))
```
GliczDev: this breaks old behavior, e.g. when loading a vanilla world or a world from an older build.
a custom Paper tag for visual fire `TriState` would be better here, like @masmc05 stated earlier

NonSwag: why would this break old worlds?
it still uses the getBoolean while just using NOT_SET if the valua is not present

GliczDev: isn't it possible for the value to be set as false in the older builds here?

NonSwag: Technically no, it only writes the value if it is true

Owen1212055 (changes requested):
Yeah as mentioned above, this is technically breaking.

Because if I supply `HasVisualFire:false` on vanilla vs paper will have different behavior.

So, a separate value should probably be added, potentially independent from has visual fire. Maybe fire override, or something.

And then we can direct people trying to use visual fire to instead use this api for more control

NonSwag:
I added an independent FireOverride tag and if that tag is present the vanilla one will be ignored

lynxplay (approved):

Owen1212055 (approved):

PR #12310, opened 2025-03-17
FishHook - add method to retrieve

ShaneBeee:
This PR aims to add a method to retrieve the current fish hook casted by a player.
I've seen this asked for many times, so I think its my time to add this.

Looking for some feedback on a few things:
1) Javadocs = I think I described things well? But always looking for feedback. Especially with the return value
2) Using `CraftItemStack.asNMSCopy()` = I chose this method to prevent actually damaging the original ItemStack, allowing the dev to do the damage themselves, they may or may not want this. 
Opinions here would be appreciated. Should it be a copy? Should it be the original?
Another option could be nixing the ItemStack all together, and internally handle it (grab based on slot chosen), maybe a boolean like `shouldDamage`, where could then decide to use copy/original.
Actually nevermind nixing the itemstack, maybe the dev wants to pass thru some super powerful fishing rod. Ohhh the possibilities.
This could go so many different ways, so feedback would be nice :)

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/FishHook.java
+++ b/paper-api/src/main/java/org/bukkit/entity/FishHook.java
@@ -377,5 +377,5 @@ public interface FishHook extends Projectile {
      * @param itemStack The fishing rod item currently cast out
      * @return The amount of damage which would be applied to the itemstack
      */
-    int retrieve(EquipmentSlot slot, ItemStack itemStack);
+    int retrieve(@NotNull EquipmentSlot slot, @NotNull ItemStack itemStack);
 }
--- a/paper-api/src/main/java/org/bukkit/entity/FishHook.java
+++ b/paper-api/src/main/java/org/bukkit/entity/FishHook.java
@@ -1,5 +1,7 @@
 package org.bukkit.entity;
 
+import org.bukkit.inventory.EquipmentSlot;
+import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -323,7 +325,6 @@ public interface FishHook extends Projectile {
         BOBBING;
     }
 
-    // Paper start - More FishHook API
     /**
      * Get the number of ticks the hook needs to wait for a fish to bite.
      *
@@ -367,5 +368,14 @@ public interface FishHook extends Projectile {
      * enchantment.
      */
     void resetFishingState();
-    // Paper end
+
+    /**
+     * Retrieve this fishhook back to the casting player.
+     * <p>This will act the same as if the player manually reels in their fishhook.</p>
+     *
+     * @param slot Slot holding the fishing rod (must be HAND/OFF_HAND)
+     * @param itemStack The fishing rod item currently cast out
+     * @return The amount of damage which would be applied to the itemstack
+     */
+    int retrieve(EquipmentSlot slot, ItemStack itemStack);
 }
--- a/paper-api/src/main/java/org/bukkit/entity/FishHook.java
+++ b/paper-api/src/main/java/org/bukkit/entity/FishHook.java
@@ -1,7 +1,6 @@
 package org.bukkit.entity;
 
 import org.bukkit.inventory.EquipmentSlot;
-import org.bukkit.inventory.ItemStack;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
--- a/paper-api/src/main/java/org/bukkit/entity/FishHook.java
+++ b/paper-api/src/main/java/org/bukkit/entity/FishHook.java
@@ -371,7 +371,7 @@ public interface FishHook extends Projectile {
     /**
      * Retrieve this fishhook back to the casting player.
      * <p>
-     * This method will trigger all normal bukkit events, which are subject to cancellation.
+     * This method will trigger and respect API events, which may be subject to cancellation.
      * Plugins listening to {@link org.bukkit.event.player.PlayerFishEvent} might for example cancel this action.
      *
      * @param slot Slot holding the fishing rod (must be HAND/OFF_HAND)
--- a/paper-api/src/main/java/org/bukkit/entity/FishHook.java
+++ b/paper-api/src/main/java/org/bukkit/entity/FishHook.java
@@ -370,9 +370,15 @@ public interface FishHook extends Projectile {
 
     /**
      * Retrieve this fishhook back to the casting player.
+     * <p>
+     * This method will trigger all normal bukkit events, which are subject to cancellation.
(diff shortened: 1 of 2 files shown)


paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftFishHook.java
```diff
@@ -233,5 +238,11 @@ public void resetFishingState() {
         hook.resetTimeUntilLured();
         hook.timeUntilHooked = 0; // Reset time until hooked, will be repopulated once lured time is ticked down.
     }
-    // Paper end
+
+    @Override
+    public int retrieve(EquipmentSlot slot, ItemStack itemStack) {
+        Preconditions.checkNotNull(slot, "Equipment slot cannot be null");
+        Preconditions.checkNotNull(itemStack, "ItemStack cannot be null");
```
Doc94:
```suggestion
        Preconditions.checkArgument(slot != null, "Equipment slot cannot be null");
        Preconditions.checkArgument(itemStack, != null "ItemStack cannot be null");
```

ShaneBeee: Curious ... why?
Internally they do about the same thing.
Is there a benefit to this method?
A message along with the suggestion change would have been appreciated!

Doc94: Hi... well if remember the idea is throw IllegalArgumentException (many methods from Preconditions throw that) but checkNotNull throws NPE.
and now i remember was mentioned in [#paper-contrib](https://canary.discord.com/channels/289587909051416579/925530366192779286/1336746860186173523)

ShaneBeee: Duly noted, thank you!

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftFishHook.java
```diff
@@ -233,5 +238,11 @@ public void resetFishingState() {
         hook.resetTimeUntilLured();
         hook.timeUntilHooked = 0; // Reset time until hooked, will be repopulated once lured time is ticked down.
     }
-    // Paper end
+
+    @Override
+    public int retrieve(EquipmentSlot slot, ItemStack itemStack) {
```
Doc94: a few things now i see this...
- This accept any item (dont throw issues) maybe can be good include a precondition for that?
- This accept any EquipmentSlot causing later the CraftEquipmentSlot.getHand throws a issue then is good add the same check here for avoid that

ShaneBeee: item: Sorry I dont understand what you mean

slot: good idea, I'll add a check for those 2 specific slots

Doc94: oh i mean you can pass any ItemStack not just the Fishrot, if NMS have more data can be good just a retrive() but i think the slot/itemstack can just get with tricks...

ShaneBeee: I tested with passing in a Diamond (that was enchanted) as well as a fishing rod.
Both "work", but they don't really seem to do much.

The luck appears to be based on the fishing rod that cast the fish hook, so I'm not even sure what the itemstack is used for. (I see it's used in the LootTable thingy, but I dont know what it's actually providing).

That said, my apologies, but I dont understand what you're asking me to change?

Doc94: i ask for check the item type because i dont se a reason for allow pass random items. but maybe its not necesary make that check.

ShaneBeee: I personally don't see a reason to restrict that.
I have limited knowledge of LootTables so Im not sure what the LootTable is actually looking for, but I feel other people are most likely more versed on that topic and may find it useful to pass in other random items.

Minecraft doesn't appear to restrict it and/or get mad about it. So I feel it should stay as is.
(Also, with all the new components Mojang keeps adding, who knows what they'll bring next. Leaving this open makes it more future safe)

notTamion (changes requested):
We shouldn't allow plugins to pass in an itemstack at all. instead we should get the itemstack thats in the provided hand and ensure its a fishing rod. i am not sure whether calling this method with a stack other than a rod will cause any internal problems however i think it might cause some issues for plugins expecting the PlayerFishEvent to always be called with a rod

ShaneBeee:
> We shouldn't allow plugins to pass in an itemstack at all. instead we should get the itemstack thats in the provided hand and ensure its a fishing rod. i am not sure whether calling this method with a stack other than a rod will cause any internal problems however i think it might cause some issues for plugins expecting the PlayerFishEvent to always be called with a rod

I can definitely change it, since it seems to be consensus.

for the record:
> i am not sure whether calling this method with a stack other than a rod will cause any internal problems

The passed ItemStack appears to be only used for criteria triggers and for the loot table. (Although not sure what it does exactly in the loot table)

> however i think it might cause some issues for plugins expecting the PlayerFishEvent to always be called with a rod

The event doesn't use this itemstack, it appears to grab straight from the slot of the player.

I guess at the end of the day its pretty useless to allow passing in another itemstack, ill change it.

ShaneBeee:
Update, I have removed the ItemStack from the method, and now retrieving it from the player based on the provided slot.
Also added a check to make sure the player is actually holding a fishing rod

kennytv (approved):

paper-api/src/main/java/org/bukkit/entity/FishHook.java
```diff
@@ -367,5 +368,13 @@ public enum HookState {
      * enchantment.
      */
     void resetFishingState();
-    // Paper end
+
+    /**
+     * Retrieve this fishhook back to the casting player.
+     * <p>This will act the same as if the player manually reels in their fishhook.</p>
```
notTamion: It will not. There will be no sound played nor will sculk sensors be alerted.

ShaneBeee: You're right, my bad.
Line has been removed.

notTamion (approved):
Merger should look into whether we should instead call FishingRodItem#use as that also triggers sculk sensors and makes the sound. Not sure whats best here

lynxplay (approved):
I am fine with this calling / respecting API events, given the resulting items might change drastically.
Given we do not apply damage to the fishing rod item either, I think it is fine to not trigger stack related actions here.

PR #12311, opened 2025-03-17
ChunkGenerator - add getHeight to ChunkData

### LINKED ISSUE #12048: ChunkGenerator - Implement a way to get noise terrain height in surface generation

My problem is I can't get the vanilla terrain heigh in a generator.

I hope I can explain this well.
Let's say I have created my own [ChunkGenerator](https://jd.papermc.io/paper/1.21.4/org/bukkit/generator/ChunkGenerator.html)

I have implemented the [generateSurface](https://jd.papermc.io/paper/1.21.4/org/bukkit/generator/ChunkGenerator.html#generateSurface(org.bukkit.generator.WorldInfo,java.util.Random,int,int,org.bukkit.generator.ChunkGenerator.ChunkData)) method

I have also implemented [shouldGenerateNoise](https://jd.papermc.io/paper/1.21.4/org/bukkit/generator/ChunkGenerator.html#shouldGenerateNoise()) ... telling Minecraft that it will generate the base noise terrain.

My plan is to just generate the surface.
Issue is there is (from what I can find) no way to get the vanilla noise terrain heigh (basically what vanilla has already generated in the chunk by the time it gets to my surface generator)

I tried [getting highest Y](https://jd.papermc.io/paper/1.21.4/org/bukkit/RegionAccessor.html#getHighestBlockYAt(int,int)) but unfortunately this froze/crashed the server (I'm assuming this just causes an endless loop)

I can't seem to find a way to get the vanilla noise terrain height.

This might be something that could be included in [ChunkData](https://jd.papermc.io/paper/1.21.4/org/bukkit/generator/ChunkGenerator.ChunkData.html)

I can't think of any alternatives.

### Other

_No response_

ShaneBeee:
This PR aims to add an option to ChunkGenerator.ChunkData to get the current height.
This allows to get the height generated in a previous step.
For example, let's say I have a generator where I allow Minecraft to generate the noise, but I want to generate the surface.
I can get the height that was previously generated in the noise step, and plop my surface block on top of that in the surface step.

Closes #12048

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/generator/CraftChunkData.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/generator/CraftChunkData.java
@@ -8,7 +8,6 @@ import net.minecraft.world.level.block.EntityBlock;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.chunk.ChunkAccess;
-import net.minecraft.world.level.levelgen.Heightmap;
 import org.bukkit.HeightMap;
 import org.bukkit.Material;
 import org.bukkit.World;
@@ -16,7 +15,6 @@ import org.bukkit.block.Biome;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.craftbukkit.CraftHeightMap;
 import org.bukkit.craftbukkit.block.CraftBiome;
-import org.bukkit.craftbukkit.block.CraftBlockType;
 import org.bukkit.craftbukkit.block.data.CraftBlockData;
 import org.bukkit.craftbukkit.util.CraftMagicNumbers;
 import org.bukkit.generator.ChunkGenerator;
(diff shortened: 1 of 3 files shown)


kennytv (approved):

Lulu13022002:
A check for out of bound position might be nice like the other methods and then return 0 (or throw) to avoid unexpected value. And the first line of javadocs seems incomplete.

ShaneBeee:
> A check for out of bound position might be nice like the other methods and then return 0 (or throw) to avoid unexpected value. 

Thats a good idea, thank you, good catch.

> And the first line of javadocs seems incomplete.

Well that is embarrassing 😬

kennytv:
I'd just throw if it's OOB

paper-server/src/main/java/org/bukkit/craftbukkit/generator/CraftChunkData.java
```diff
@@ -180,4 +181,13 @@ private void setBlock(int x, int y, int z, BlockState type) {
             access.removeBlockEntity(blockPosition);
         }
     }
+
+    @Override
+    public int getHeight(HeightMap heightMap, final int x, final int z) {
+        Preconditions.checkArgument(heightMap != null, "HeightMap cannot be null");
+        if (x < 0 || x > 15 || z < 0 || z > 15) {
+            throw new IllegalArgumentException("Cannot get height outside of a chunks bounds");
```
electronicboy: maybe print the coords, just so it's easier to see what has happened when debugging, etc; People mistaking block/chunk bound coords is always kinda fun

ShaneBeee: Ah yeah, fair point... will change now!

electronicboy (approved):

Warriorrrr (approved):

PR #12316, opened 2025-03-18
[ci skip] CONTRIBUTING.md use latest Java 21 version for Docker

Doc94:
Based in discord a few mentions about early versions of Java 21 this PR just replace the version used in the Docker example for just use the latest of Java 21.


Doc94:
okay i read in a folia issue the mention for 21.0.5 but maybe i just misread that. closed then.

PR #12320, opened 2025-03-20
Fix cancellation of PlayerTrackEntityEvent

### LINKED ISSUE #14280: Entity#getTrackedBy includes players whose tracking was cancelled
### Expected behavior

When cancelling `PlayerTrackEntityEvent`, subsequent calls to `Entity#getTrackedBy` on the entity and player involved would not include the player, i.e. `Entity#isTrackedBy` would return false.

### Observed/Actual behavior

When cancelling `PlayerTrackEntityEvent`, calls to `Entity#getTrackedBy` while the entity is still in range of the player include the player in the result, and `Entity#isTrackedBy` also returns true for the same pair.

### Steps/models to reproduce

* Register a listener for `PlayerTrackEntityEvent` that cancels the event
* Subsequently, check `Entity#isTrackedBy` on the same entity and player.

Note that the event works correctly in that the player can't see the entity after it's cancelled.

### Plugin and Datapack List

None

### Paper version

```
This server is running Paper version 26.2-121-main@a2a42c5 (2026-08-29T11:32:25Z) (Implementing API version 26.2.build.121-stable)
You are 2 version(s) behind
Previous version: 26.2-DEV-acd0ad4 (MC: 26.2)
```

### Other

This occurs because the player is still in the `seenBy` set in `ChunkMap.TrackedEntity`, presumably so that the event doesn't fire multiple times? However, that is also the set checked by the entity methods.

Gerrygames:
Cancelling the PlayerTrackEntityEvent did not remove the player from the ``seenBy`` set which leads to updates being broadcast despite the entity not being present on the client and more crucially the entity can only be tracked again by first failing to meet any condition to be tracked.

### DIFF
--- a/paper-server/patches/features/0016-Moonrise-optimisation-patches.patch
+++ b/paper-server/patches/features/0016-Moonrise-optimisation-patches.patch
@@ -24270,7 +24270,7 @@ index e823b8aac00158892538083bc877ccf99895909a..7d871318065f19540748363809de8265
      private static final ChunkStep FULL_CHUNK_STEP = ChunkPyramid.GENERATION_PYRAMID.getStepTo(ChunkStatus.FULL);
      public static final int RADIUS_AROUND_FULL_CHUNK = FULL_CHUNK_STEP.accumulatedDependencies().getRadius();
 diff --git a/net/minecraft/server/level/ChunkMap.java b/net/minecraft/server/level/ChunkMap.java
-index ad665c7535c615d2b03a3e7864be435f933235dd..3dff97f13586be3b52bbe786852c185f6753a019 100644
+index a1a0624228f46abe2534b13b59e1fc1a429f409b..c62e1465fa95b077d60f7f3ca97604e029f5a77c 100644
 --- a/net/minecraft/server/level/ChunkMap.java
 +++ b/net/minecraft/server/level/ChunkMap.java
 @@ -96,7 +96,7 @@ import net.minecraft.world.level.storage.LevelStorageSource;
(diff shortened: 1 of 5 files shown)


lynxplay:
This reads like it'll start spamming this event every time, attempting to track the player on every tick.

Gerrygames:
Ah yes this is correct. I will add a check if the entity is currently tracked before calling the event.

Gerrygames:
Hm which will still spam the event when the entity is not tracked. But I think this is how this event is supposed to work.

lynxplay:
Eh, I think the proper solution is the one we wanted for a while, converting seenBy into a Map that attaches some tracking config.

That way, the entity knows vanilla would track it, but we can prevent e.g. packets being sent. This would also be a massive performance boost to the API canSee/hide/show entity stuff.

electronicboy:
Yea, as then we can mark it as "tracked" but pretend it isn't, making it easy to allow API to retrack and being able to inline a bunch of stuff much easier rather than several layers of indirection in order to check if we can send to who and what

Gerrygames:
I do agree that this system in general can be improved. The ``Entity#canSee`` check can also be quite heavy when entities are hidden from many players. But I  think this is outside the scope of this PR. My goal was to fix that cancelling the ``PlayerTrackEntityEvent`` leads to a inconsistent state. If this approach is considered too hard of a performance hit then I am happy to wait for a broader rework of the tracking system. However I don't think this is too much of a problem with my updated implementation. The event will only be called upon initial tracking or for entities for which tracking was cancelled by this event. So for servers that aren't cancelling this event at all there should be no performance impact at all.

roccodev:
I stumbled upon this shortly after I submitted my own version of it (#14282), my implementation was slightly different so maybe it fits better in the current design?

PR #12321, opened 2025-03-20
Check if BUNDLE_CONTENTS is present in InventoryClickEvent

notTamion:


kennytv:
Shouldn't it fully replace the BundleItem instance check?

notTamion:
you can put the BUNDLE_CONTENTS component on a non bundle item (which will have no actual affect but i am not sure if plugins might use it to store data for some reason)

kennytv (approved):

Warriorrrr (approved):

PR #12323, opened 2025-03-21
Fix PlayerLoginEvent firing multiple times.

### LINKED ISSUE #8676: PlayerLoginEvent fires twice
### Expected behavior

When a player connects, the event should only fire once

### Observed/Actual behavior

The event fires twice

### Steps/models to reproduce

Add the below event listener to your plugin (This event listener can be replaced by turning on the whitelist and looking at the "Disconnecting" logs. The addition of code just makes it a bit easier to see)

```java
@EventHandler
public void playerLoginEvent (PlayerLoginEvent event) {
     System.out.println("Testing");
}
```

Join on an account and "Testing" will be logged twice

### Plugin and Datapack List

[22:58:48 INFO]: Plugins (3): Campfire, LuckPerms, spark

### Paper version

[22:58:00 INFO]: This server is running Paper version git-Paper-333 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: eec64a4)

### Other

_No response_

orang3i:
fixes #8676

### DIFF
--- a/paper-server/patches/sources/net/minecraft/network/Connection.java.patch
+++ b/paper-server/patches/sources/net/minecraft/network/Connection.java.patch
@@ -17,7 +17,7 @@
      );
      private static final ProtocolInfo<ServerHandshakePacketListener> INITIAL_PROTOCOL = HandshakeProtocols.SERVERBOUND;
      private final PacketFlow receiving;
-@@ -88,6 +_,11 @@
+@@ -88,6 +_,12 @@
      private final Queue<Consumer<Connection>> pendingActions = Queues.newConcurrentLinkedQueue();
      public Channel channel;
      public SocketAddress address;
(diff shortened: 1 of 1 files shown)


paper-server/patches/sources/net/minecraft/network/Connection.java.patch
```diff
@@ -232,21 +233,38 @@
...
 +        this.preparing = false; // Spigot
          if (this.channel == null) {
              this.delayedDisconnect = disconnectionDetails;
-         }
+-        }
++            return; // Paper - Defer disconnect until channel is available.
++        }
++
++        // Paper start - Fix PlayerLoginEvent firing multiple times.
++        if (!this.disconnecting.compareAndSet(false, true)) {
```
kennytv, line 248: I haven't tested it, but it looks like setting this *before* the isConnected call below means the actual disconnect logic is skipped?

Owen1212055:
"Resolved" by https://github.com/PaperMC/Paper/pull/12301

However, the underlying issue is still there.

PR #12326, opened 2025-03-22
Deprecate getPotentialRespawnLocation and add getRespawnBlock

NonSwag:
Introduced `getRespawnBlock` to retrieve the player's respawn block directly and deprecated `getPotentialRespawnLocation` due to its misleading name.
Updated documentation to clarify usage.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -7,6 +7,7 @@ import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
+import org.bukkit.block.Block;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
@@ -440,7 +441,6 @@ public interface HumanEntity extends LivingEntity, AnimalTamer, InventoryHolder 
     public int getSleepTicks();
 
 
-    // Paper start - Potential bed api
     /**
      * Gets the Location of the player's bed, null if they have not slept
      * in one. This method will not attempt to validate if the current bed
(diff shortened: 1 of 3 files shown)


paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -463,10 +463,26 @@ default Location getPotentialBedLocation() {
...
+    @Deprecated(since = "1.21.4")
+    default Location getPotentialRespawnLocation() {
+        Block block = this.getRespawnBlock();
+        return block != null ? block.getLocation() : null;
+    }
+
+    /**
+     * Retrieves the block located at the player's designated respawn point.
+     * This method will not attempt to validate if the current respawn point is still valid.
+     *
```
notTamion, line 480: these jdocs are still kinda misleading/unclear. Perhaps an example for a bed and for a respawn anchor could be included. Besides i think we should stick to returning a Location as returning an air block when the player had their spawn location changed in another way feels weird because the block itself isn't directly linked to the player having their respawn location there. getRespawnBlockLocation might be a good name then (also makes it clear that this doesn't just return the block at the location of #getRespawnLocation).

notTamion (changes requested):
kinda sucks that we are deprecating a really new method here, but i definitely agree that the change in #11950 were unclear.

notTamion:
ahh just remembered #11960 actually already handles this

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
NonSwag: Yeah I think you are right, I will change that later

Lulu13022002:
I think there's some confusion, the reason this method should be deprecated is because it doesn't handle the respawn angle (that is available even without loading the position). Now this method return the block position because that's how the game works, it save the block position and then recompute the exact location when needed.

NonSwag:
i thing we should have a getter for the actual block position regardless

notTamion:
Yeah i agree. Having plugins rely on calling this method with load = false to get that information and not providing its own method for it is bad imo

Owen1212055:
Agreed, this block position being masked under this pontential respawn location method is very confusing.

Owen1212055:
The api around this has been reworked and it has been decided that this is more of a documentation issue, and will be updated as such a bit later once the final wording has been decided.

Thank you for your contribution!

PR #12327, opened 2025-03-22
RideCommand - config to remove player as vehicle restriction

ShaneBeee:
This PR aims to remove the player as vehicle restriction in the vanilla `/ride` command.
Its well know that the API allows mounting any entity to the player as a passenger, so it seems fitting to allow this in the ride command as well.
Unfortunately Mojang imposed a restriction here.
Some people may often rely on this command for funsies, or possibly even for debugging/testing when writing code.

Doc PR: https://github.com/PaperMC/docs/pull/552

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/commands/RideCommand.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/commands/RideCommand.java.patch
@@ -5,7 +5,7 @@
          if (vehicle1 != null) {
              throw ERROR_ALREADY_RIDING.create(target.getDisplayName(), vehicle1.getDisplayName());
 -        } else if (vehicle.getType() == EntityType.PLAYER) {
-+        } else if (vehicle.getType() == EntityType.PLAYER && !io.papermc.paper.configuration.GlobalConfiguration.get().commands.rideCommandSkipPlayerAsVehicleRestriction) { // Paper - allow player as vehicle
++        } else if (vehicle.getType() == EntityType.PLAYER && !io.papermc.paper.configuration.GlobalConfiguration.get().commands.rideCommandAllowPlayerAsVehicle) { // Paper - allow player as vehicle
              throw ERROR_MOUNTING_PLAYER.create();
          } else if (target.getSelfAndPassengers().anyMatch(passenger -> passenger == vehicle)) {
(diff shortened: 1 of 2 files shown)


Owen1212055:
Not really a fan of changing vanilla behavior of commands like this.

NonSwag:
Plugins can't really hook into this to allow riding players, so some way to allow it anyway would be nice
MAYBE a config option?

ShaneBeee:
well dang.

kennytv:
Owen is cringe that's fine, but yeah a config option would be good

ShaneBeee:
ok cool, I'll attempt a config option, and report back when I get that figured out.

ShaneBeee:
Ok, that was a heck of a lot easier than I expected.
Id love some feedback on my very wordy config option. 
Too wordy? Just right? Not wordy enough?

NonSwag:
what about skipPlayerRideRestrictionCheck?

ShaneBeee:
> what about `skipPlayerRideRestrictionCheck`? Still very long but opposed to `rideCommandSkipPlayerAsVehicleRestriction`

I worry that that does not explain which command tho.

NonSwag:
i mean, you have the comment... sooo...

ShaneBeee:
> i mean, you have the comment... sooo...

true, but that doesn't show in the actual config, ie:
Screenshot 2025-03-22 at 12 58 03 PM

NonSwag:
_yaml is weird_
i think your name is alright 
the average key length in that file makes it appear shorter

ShaneBeee:
> _yaml is weird_ i think your name is alright the average key length in that file makes it appear shorter

thank you "suggest-player-names-when-null-tab-completions" for making me look fine 😆

lynxplay:
`rideCommandAllowPlayerAsVehicle` seems shorter and does the same.
Additionally, please open a pr at papermc/docs for this feaature :+1:

ShaneBeee:
> `rideCommandAllowPlayerAsVehicle` seems shorter and does the same. Additionally, please open a pr at papermc/docs for this feaature 👍

Done and done: 
https://github.com/PaperMC/docs/pull/552
:)

kennytv (approved):

Warriorrrr (approved):

PR #12330, opened 2025-03-23
Add flush parameter to World#save

### LINKED ISSUE #12329: Expand world saving API

When dynamically creating worlds with the goal of an in-game map editor, saving a world does, by default, not flush the level data save. (`World#save()` calls `ServerLevel#save(null, false, false)`. The first `false` declaring the `flush` variable). This makes subsequent file copy operations on the world folder unreliable, as not all chunks have been saved yet.

I would like to expose another function, `World#save(boolean flush)`, which allows an API user to directly set the `flush` parameter without relying on internals.

Currently, the only way to achieve this is with the use of internals. More specifically:
```java
final World world = ...;
((CraftWorld) world).getHandle().save(null, true, false);
```

### Other

None 🚀

Strokkur424:
Resolves #12329.

Adds a new method, `World#save(boolean flush)`, in order to expose the internal functionality. See the mentioned issue for detail.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/RegionAccessor.java
+++ b/paper-api/src/main/java/org/bukkit/RegionAccessor.java
@@ -11,13 +11,14 @@ import org.bukkit.block.data.BlockData;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.LivingEntity;
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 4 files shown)


Strokkur424:
Would it be okay for me if I would clean up World and CraftWorld, migrating them to jspecify and removing the redundant `public` modifiers in the World interface? I am touching those files already anyways, so might as well

kennytv (approved):

Warriorrrr (approved):

PR #12333, opened 2025-03-24
Call Command#getPermission instead of directly referencing the object

Leguan16:
Changes the `Command#testPermissionSiltent` method to call `Command#getPermission` instead of directly referencing the permission object. 

This way it is also possible to override the `Command#getPermission` method if someone chooses to do so.


Lulu13022002:
What is the use case here, why you can't use the setter?

Leguan16:
Someone on discord complained that you can't override the getter. I thought it makes sense to do either.

Owen1212055:
Not sure if this is worth testing, as this API is generally "obsolete" with our new API and I am not sure if introducing changes is the best.

lynxplay:
I do not believe it is healthy to encourage a disconnect between getPermission and setPermission, which seems to be the only usecase for this to exist over people using the setter.

Thank you for the PR anyway tho!

PR #12346, opened 2025-03-27
Fix EntityTeleportEvent triggering twice

### LINKED ISSUE #12317: EntityTeleportEvent fires twice when entity teleports
### Expected behavior

The EntityTeleportEvent event fires once when an entity teleports, with for example the /minecraft:tp command.

### Observed/Actual behavior

The EntityTeleportEvent event fires twice when an entity teleports.

### Steps/models to reproduce

I used a plugin called ExecutableEvents that utilizes this event. It allows you to fire commands when one of these events happen. According to the plugin developer, this does indeed use the EntityTeleportEvent event and fires twice.

### Plugin and Datapack List

Image

Image

### Paper version

[10:31:55 INFO]: Checking version, please wait...
[10:31:55 INFO]: This server is running Paper version 1.21.4-211-main@6ea4202 (2025-03-13T11:49:31Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are 1 version(s) behind
Previous version: 1.21.4-187-1d5e5a5 (MC: 1.21.4)

### Other

Plugin utilized to find this out: ExecutableEvents by Ssomar

One of the plugin devs shared these screenshots with me:

Image

Image

orang3i:
fixes #12317 

fix vanilla tp command firing the EntityTeleportEvent twice and CraftEntity#teleport firing the EntityTeleportEvent twice for cross world teleports.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/commands/TeleportCommand.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/commands/TeleportCommand.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/server/commands/TeleportCommand.java
 +++ b/net/minecraft/server/commands/TeleportCommand.java
-@@ -290,7 +_,31 @@
+@@ -290,7 +_,16 @@
              float f1 = relatives.contains(Relative.X_ROT) ? xRot - target.getXRot() : xRot;
              float f2 = Mth.wrapDegrees(f);
              float f3 = Mth.wrapDegrees(f1);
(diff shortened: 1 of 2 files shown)


orang3i:
- for CraftEntity#teleport the event is called twice only for cross world teleports. This was fixed by adding a check before the event call in CraftEntity#teleport to check if the destination and current location are of the same worlds and if they are, we don't call the event and the event is fired from vanilla teleport.

- for vanilla tp class, the event call is entirely removed and the event will be called from vanilla teleport.

orang3i:
This pr only fixes the vanilla tp firing the EntityTeleportEvent twice.

Warriorrrr:
Thank you for your contribution and apologies for the delay, we're currently planning on going ahead with the changes #13181 soon which is a larger rework of the teleport implementation, which also happens to fix both of these issues, so I think it's best if that PR supersedes this one.

PR #12347, opened 2025-03-27
Simplify custom payload handling

kennytv:


electronicboy (approved):

lynxplay (approved):

WouterGritter:
Is it papers goal to maintain a persistent/reliable api for internal minecraft classes like these? My team is working on a plugin that calls the data() method on this record and it broke after this commit. The fix is trivial with the use of reflection, but it's still not great having a plugin fail after a seemingly small and trivial change.

lynxplay:
No. Internals change whenever we see fit. There is 0 guarantee internal fields remain the same at all at any point in time.
The API for this behaviour is the plugin messaging API, which remained stable.
Internals are not API and do not have stability guarantees between versions or even builds.

PR #12354, opened 2025-03-28
Clone exit location passed to teleport event

### LINKED ISSUE #12340: PlayerTeleportEvent::setTo(Location) does not function correctly when passed modified object returned by getTo()
### Expected behavior

This is a somewhat recent "breakage", this behavior changed sometime after 1.21.3 (unsure when in 1.21.4)

Calling PlayerTeleportEvent::getTo(), modifying that object and calling setTo() should modify the teleport destination.

_Frankly simply calling getTo() and modifying it should also modify the teleport destination given it is returned by ref, however my plugin always called setTo anyway._

### Observed/Actual behavior

Unless the result of getTo() is cloned before passing it to setTo() the teleport destination is unchanged.

### Steps/models to reproduce

```java
public void onEvent(PlayerTeleportEvent event) {
        Location dest = event.getTo();
        if (dest == null) {
            return;
        }
        dest = event.getTo();  // cloning this object here restores correct behavior
        dest.add(0, 8, 0);
        event.setTo(dest);
    }
```

### Plugin and Datapack List

```
[22:19:37 INFO]: ℹ Server Plugins (2):
[22:19:37 INFO]: Bukkit Plugins:
[22:19:37 INFO]:  - EternalSurvival, LuckPerms
```

### Paper version

Paper version 1.21.4-221-main@c467df9 (2025-03-24T00:18:41Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Other

_No response_

emilyy-dev:
The exit location passed to PlayerTeleportEvent/PlayerTeleportEndGatewayEvent in ServerPlayer#teleport(TeleportTransition) needs to be cloned, as it is later compared in case it was changed. Not cloning it results in the ability for plugins to mutate it but such changes will be ignored, as the instance of the Location being the same results in a successful equality check.

It is not necessary to clone the location in other instantiations of the event (or subclasses) as those are not compared afterwards to change the outcome.

Fixes #12340

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
@@ -620,10 +620,10 @@
 +            org.bukkit.Location enter = this.getBukkitEntity().getLocation();
 +            PositionMoveRotation absolutePosition = PositionMoveRotation.calculateAbsolute(PositionMoveRotation.of(this), PositionMoveRotation.of(teleportTransition), teleportTransition.relatives());
 +            org.bukkit.Location exit = org.bukkit.craftbukkit.util.CraftLocation.toBukkit(absolutePosition.position(), level.getWorld(), absolutePosition.yRot(), absolutePosition.xRot());
-+            org.bukkit.event.player.PlayerTeleportEvent tpEvent = new org.bukkit.event.player.PlayerTeleportEvent(this.getBukkitEntity(), enter, exit, teleportTransition.cause());
++            org.bukkit.event.player.PlayerTeleportEvent tpEvent = new org.bukkit.event.player.PlayerTeleportEvent(this.getBukkitEntity(), enter, exit.clone(), teleportTransition.cause());
 +            // Paper start - gateway-specific teleport event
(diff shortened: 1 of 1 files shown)


emilyy-dev:
lynx can cherry-pick this commit for the 1.21.5 branch

rymiel (approved):

brbavar:
I see how changes to the instance fields `x`, `y`, and `z` of `exit` would be ignored were `newExit.equals(exit)` to return `true`, since the condition `!newExit.equals(exit)` would consequently evaluate to `false`. They would be ignored in the sense that they would have no effect on the teleport transition resulting from `tpEvent`; `teleportTransition` would not be reassigned, so its value would remain what it was when `teleportTransition` was passed to `ServerPlayer#teleport(TeleportTransition)`.

But if I'm not mistaken, `newExit.equals(exit)` does not necessarily return `true` if `newExit` and `exit` refer to the same instance of `Location`. `Location#equals(Object)` contains `if` statements that check via bitwise comparison if this location's `x`, `y`, and `z` fields have the same numeric values as the argument's `x`, `y`, and `z` fields, respectively. Thus, as long as a plugin successfully mutates one of those fields of `exit` before the invocation of `equals` in the condition, `newExit.equals(exit)` evaluates to `false` and the condition to `true`.

I'm new to this codebase, so it's entirely possible that I'm missing something.

Warriorrrr (approved):

kennytv (approved):

PR #12358, opened 2025-03-29
fix structures not generating when decorations are disabled

### LINKED ISSUE #12353: ChunkGenerator#shouldGenerateStructures() doesn't fully work when #shouldGenerateDecorations() is false
### Expected behavior

I expect structures to generate when I return true for `shouldGenerateStructures()`

### Observed/Actual behavior

I noticed when doing some generator stuff, that structures aren't placed into the world when `ChunkGenerator#shouldGenerateDecorations()` return false.

I did some digging and came to the realization why this happens.
Structures are broken up into 2 parts:
1) StructureStarts = This is an early stage in the generation steps. Minecraft decides which types of structures will go where.
When using `shouldGenerateStructures` returned as true... this stage decides it will place a structure there.

2) StructurePlacements = Later on in the chunk generation steps, in the same stage that decorations (features) are placed, the structure pieces are also placed.
After seeing this in the code I realized if `shouldGenerateDecorations` returns false (thus preventing features from being placed), the structure piece placements are also skipped.

When this happens, the game will assume a structure is placed in a specific spot, this can be seen using the locate command (see steps below), where it'll find a structure, but nothing actually got placed.

### Steps/models to reproduce

Simple code to replicate:
```java
public void testGenerator() {
    WorldCreator worldCreator = new WorldCreator("world_test_for_paper");
    worldCreator.generator(new ChunkGenerator() {

        @Override
        public boolean shouldGenerateNoise() {
            return true; // Vanilla noise
        }

        @Override
        public boolean shouldGenerateSurface() {
            return true; // Vanilla surface
        }

        @Override
        public boolean shouldGenerateStructures() {
            return true; // Yes we want structures
        }

        @Override
        public boolean shouldGenerateDecorations() {
            return false; // No we dont want your trees
        }

        @Override
        public Location getFixedSpawnLocation(@NotNull World world, @NotNull Random random) {
            return new Location(world,1,1,1); // Speed up the process
        }
    });
    World testWorld = worldCreator.createWorld();
}
```

- start up server
- join server and teleport to new world
- run command `/locate structure minecraft:village_plains`
- Once the game finds the village, click the TP thing to TP to it.
- Behold, no village.

### Plugin and Datapack List
```
[14:29:19 INFO]: ℹ Server Plugins (1):
[14:29:19 INFO]: Paper Plugins (1):
[14:29:19 INFO]:  - CorePlugin #(my plugin for testing)
```

### Paper version

Currently using the 1.21.5 update branch:
```
[14:24:04 INFO]: This server is running Paper version 1.21.5-DEV-update/1.21.5@ce91a8c (1970-01-01T00:00:00Z) (Implementing API version 1.21.5-R0.1-SNAPSHOT)
Previous version: 1.21.5-DEV-b6d70a9 (MC: 1.21.5)
```

I believe this would also happen on previous versions. 
I'm pretty sure this happened to me in 1.21.4 and just forgot to report it.

### Other

I'm open to doing a PR for this. 
Looking thru the code I feel like this could potentially be a simple fix.
I wanted to wait til after 1.21.5 was released, as I'm sure you guys have too much on your plate as it is.

orang3i:
fixes #12353 

Please feel free to let me know if any other simpler fix is possible

(tested on 1.21.4)

### DIFF
--- a/paper-server/patches/features/0016-Moonrise-optimisation-patches.patch
+++ b/paper-server/patches/features/0016-Moonrise-optimisation-patches.patch
@@ -31553,7 +31553,7 @@ index d63d745a0220963a297cfedf1e8983aeb471a045..6d565b52552534ce9cacfc35ad1bf4ad
  
      public record PackedTicks(List<SavedTick<Block>> blocks, List<SavedTick<Fluid>> fluids) {
 diff --git a/net/minecraft/world/level/chunk/ChunkGenerator.java b/net/minecraft/world/level/chunk/ChunkGenerator.java
-index b320519578f7e2c52b2116afb247e071b322696e..6ed51cf42b5864194d671b5b56f5b9bdf0291dc0 100644
+index 378e8ddef41c68884f9b29b3b07e23b454bab0c7..b202ba800bd3317abc38bf1c8a4df36ecfe0128c 100644
 --- a/net/minecraft/world/level/chunk/ChunkGenerator.java
 +++ b/net/minecraft/world/level/chunk/ChunkGenerator.java
 @@ -116,7 +116,7 @@ public abstract class ChunkGenerator {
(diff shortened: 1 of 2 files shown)


lynxplay:
So, the boolean check there might want to respect the bukkit ChunkGenerator#shouldGenerateStructures method.
Similar to how the applyBiomeDecoration method in CustomChunkGenerator respects shouldGenerateDecorations

orang3i:
@lynxplay  Bukkit ChunkGenerator#shouldGenerateStructures is now respected 🫡

PR #12359, opened 2025-03-29
Make fishing bobbers respect NoGravity

brickmonster:
Normally NoGravity is checked in `getGravity` but fishing bobbers don't call or override it with an appropriate value.


brickmonster:
Nevermind, the client prediction makes this too glitchy to be useful.

Issue #12350, opened 2025-03-27
API for adding Mob Variants to the Registry


BertTowne:
### Is your feature request related to a problem?

Like Enchantments and, more similarly, Painting Variants, certain mob variants are now data-driven as of Minecraft 1.21.5. Those being:
- Wolf
- Cat
- Frog
- Chicken
- Cow
- Pig

With this list ever-expanding, exposing this functionality will give plugin developers the opportunity to easily add new forms of these entities with the help of a resource pack - something that has previously been possible only through mods, and is only now possible with data packs.

### Describe the solution you'd like.

A new `RegistryEntry` and `RegistryEventProvider` for each of the new mob variants to allow dynamic registration of new mob variants during the bootstrap phase.

### Describe alternatives you've considered.

This is currently easily possible through data packs, however the ability to add "new" mobs in a way is something that plugin developers have dreamed of since the early days of server-side modding.

BertTowne:
Hey! I have implemented this locally for chickens so far, and I'd be happy to make a PR for it. That way you can give feedback on what I have, I can improve it, and implement the other mob variants.

Machine-Maker:
Yeah, you can open a PR. The biggest things to decide is representing spawn conditions as well as the modelandtype. Cause some of those types probably have use outside the registry API.

BertTowne:
Done! #12364 I agree with you, especially on the SpawnConditions. I was a little stumped on the best way to go about adding that myself, so they have been intentionally left out while we brainstorm ideas.

lynxplay:
Closed by #12417

PR #12364, opened 2025-03-30
Registry API for supported Mob Variants

BertTowne:
#12350 Currently only implemented Chicken variants as a way to gather feedback before adding the rest. I'm especially wanting opinions on the best way to implement SpawnConditions to the API, as those have intentionally been left out of this, but right now custom mob variants are given `SpawnPrioritySelectors.EMPTY`.

`ModelAndTexture` has been separated into Model and Texture for API purposes, as the vanilla-defined models are not entirely the same across each mob and are unfortunately stored as enums. `assetId` is required, and `model` will default to NORMAL.


Lulu13022002:
Changes between generated comments will be lost. Registration is now done in RegistryEntries in the generator module. I have some plan later to move that over a json file but for now that should do it. The registry don't need to be delayed too.

BertTowne:
Thank you, I have made the proposed changes and went forward with implementing the rest of the data-driven mob variants in a similar way to that of the Chicken Variants from before.

kennytv:
Oop, please reopen a PR targeting main again

Issue #12369, opened 2025-03-31
Calling getCustomModelData() if the ItemStack hasCustomModelData() throws "No numeric custom model data"


kangarko:
### Expected behavior

Hello,

I'd like to report what I believe is a bug in Paper where an exception is raised by a simple method call.

### Observed/Actual behavior

```
[19:41:24 ERROR]: Could not pass event InventoryOpenEvent to Protect v2.4.6
java.lang.IllegalStateException: No numeric custom model data
at com.google.common.base.Preconditions.checkState(Preconditions.java:513) ~[guava-33.3.1-jre.jar:?]
at org.bukkit.craftbukkit.inventory.CraftMetaItem.getCustomModelData(CraftMetaItem.java:1417) ~[paper-1.21.4.jar:1.21.4-138-5395ae3]
at Protect-2.4.6.jar/org.mineacademy.protect.operator.Rule$RuleCheck.check0(Rule.java:418) ~[Protect-2.4.6.jar:?]
at Protect-2.4.6.jar/org.mineacademy.protect.operator.Rule$RuleCheck.checkPlayerInventory(Rule.java:260) ~[Protect-2.4.6.jar:?]
at Protect-2.4.6.jar/org.mineacademy.protect.operator.Rule.filter(Rule.java:200) ~[Protect-2.4.6.jar:?]
at Protect-2.4.6.jar/org.mineacademy.protect.operator.Rule.filterOpenContainer(Rule.java:148) ~[Protect-2.4.6.jar:?]
at Protect-2.4.6.jar/org.mineacademy.protect.listener.ScanListener.onInventoryOpen(ScanListener.java:121) ~[Protect-2.4.6.jar:?]
at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:80) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:70) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
at io.papermc.paper.plugin.manager.PaperEventManager.callEvent(PaperEventManager.java:54) ~[paper-1.21.4.jar:1.21.4-138-5395ae3]
at io.papermc.paper.plugin.manager.PaperPluginManagerImpl.callEvent(PaperPluginManagerImpl.java:131) ~[paper-1.21.4.jar:1.21.4-138-5395ae3]
[...]

PR #12370, opened 2025-03-31
Deprecate ItemMeta#get/setCustomModelData

electronicboy:
The API is no longer able to properly reflect the internal state here

References #12369


lynxplay:
We should cherry pick https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/47480cd07c0957a94b220f3087b851594b063e54

montlikadani:
Also it would be good to remove `Experimental` ApiStatus from CustomModelDataComponent and its derivatives as well

lynxplay:
Superseded by: #12414

PR #12373, opened 2025-04-01
Fix portal create event block list

gecko10000:
The current logic for the portal gets the portal blocks incorrectly. It retrieves blocks in a diagonal line, as seen below (the two lime wool are part of this incorrect line). This PR changes the variable used to correctly align the blocks retrieved.
image


Lulu13022002:
That's fair, but generally this event call is weird because:
1) the blocks around the portal are not really created at this point (i get why it's exposed but that will conflict with 2)
2) plugins can't mutate the new block states here unlike the other call of the events

gecko10000:
My testing was for lighting an already-created (player-built) portal, meaning the obsidian already exists in that context. Looking at the code for exit portal generation, I can't seem to find where the obsidian is actually set in the level. In my case, though, I'm able to modify the obsidian because it's not modified by the portal's creation.

kennytv (approved):

lynxplay (approved):

lynxplay:
Agree with lulu, the event is in a bit of a weird place there but yea, w/e, not part of this PR.

PR #12374, opened 2025-04-01
Fix floating warning log when the kick event is canceled

pedroagrs:
Currently, the floating kick log is sent before the `PlayerKickEvent` is fired, resulting in incorrect and potentially spammy logging when the event is properly cancelled. This PR fixes the issue by moving the logging to be sent after the event logic.


paper-server/patches/features/0032-Fix-floating-disconnect-logging.patch
```diff
@@ -0,0 +1,56 @@
...
+
+diff --git a/net/minecraft/server/network/ServerCommonPacketListenerImpl.java b/net/minecraft/server/network/ServerCommonPacketListenerImpl.java
+index af75f1ca60dc01045f27cb550ac9e3deacf1a1cf..fcc597697795d48424f2912ea5cd70fce510548f 100644
+--- a/net/minecraft/server/network/ServerCommonPacketListenerImpl.java
++++ b/net/minecraft/server/network/ServerCommonPacketListenerImpl.java
+@@ -25,6 +25,7 @@ import net.minecraft.server.MinecraftServer;
+ import net.minecraft.server.level.ClientInformation;
+ import net.minecraft.util.VisibleForDebug;
+ import net.minecraft.util.profiling.Profiler;
++import org.bukkit.event.player.PlayerKickEvent.Cause;
```
dawon: You should use fully qualified names to avoid imports in NMS code...

pedroagrs: Auto import issues... Fixed!

notTamion:
I think we should additionally reset aboveGroundTickCount when the server attempts to disconnect a player to prevent the event from being spammed

lynxplay:
> I think we should additionally reset aboveGroundTickCount when the server attempts to disconnect a player to prevent the event from being spammed

Possibly, tho in a separate PR :+1:

lynxplay (approved):

PR #12384, opened 2025-04-05
Prevent keep inventory deaths from dropping open container/cursor items

NewwindServer:
When a player has an open inventory—such as with a merchant or anvil—and
places items in it, or has an item on their cursor, closing the inventory typically 
returns those items to the player's main inventory. However, if the player dies with
the inventory still open, the items are dropped onto the ground.

This usually isn't a problem, as their entire inventory would be dropped anyway.
But when keep inventory is enabled, these items should return directly to the player's inventory
to prevent valuable items from being accidentally dropped—especially
in spawn areas where keep inventory is active.

Rebased version of: #11533

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
@@ -894,6 +894,7 @@ public class CraftEventFactory {
         // Paper end
 
         victim.keepLevel = event.getKeepLevel();
+        victim.keepInventory = event.getKeepInventory();
         victim.newLevel = event.getNewLevel();
         victim.newTotalExp = event.getNewTotalExp();
         victim.expToDrop = event.getDroppedExp();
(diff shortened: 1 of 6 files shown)


lynxplay:
I don't think this is worth a feature patch.

Can you please merge this into the per-file patches, thanks!
Generally, I like the concept tho, a config option with vanilla behaviour is also 100% something I am favour of as implemented in this PR :+1:

paper-server/src/main/java/org/bukkit/craftbukkit/event/CraftEventFactory.java
```diff
@@ -1034,6 +1034,7 @@ public static PlayerDeathEvent callPlayerDeathEvent(ServerPlayer victim, DamageS
         // Paper end
 
         victim.keepLevel = event.getKeepLevel();
+        victim.keepInventory = event.getKeepInventory(); // Paper - keep inventory tracking
```
NonSwag: paper comments not really needed anymore

notTamion (approved):

lynxplay:
After some more thought, this can literally be implemented via
```java
@EventHandler
public void on(final PlayerDropItemEvent event) {
    if (event.getPlayer().isDead()) event.setCancelled(true);
}
```

Passing the "was keep inventory" around is also simple on a plugin perspective with the death event being called prior to the dropping.

In the name of not polluting our configuration with things plugins could be implementing, I'll be closing the PR.
Thank you so much for it anyway, as per usual, really well crafted and documented <3

PR #12388, opened 2025-04-06
Add PlayerAttackEntityEvent

TransparentWhit:
Improved from [#12319](https://github.com/PaperMC/Paper/pull/12319)

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/player/PlayerAttackEntityEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/player/PlayerAttackEntityEvent.java
@@ -39,7 +39,7 @@ public class PlayerAttackEntityEvent extends PlayerEvent implements Cancellable 
      *
      * @return entity that was attacked
      */
-    public Entity getAttacked() {
+    public @NotNull Entity getAttacked() {
         return this.attacked;
     }
 
@@ -70,7 +70,7 @@ public class PlayerAttackEntityEvent extends PlayerEvent implements Cancellable 
      *
      * @return type of the attack
      */
-    public PlayerAttackType getType() {
+    public @NotNull PlayerAttackType getType() {
         return this.type;
     }
 
@@ -79,7 +79,7 @@ public class PlayerAttackEntityEvent extends PlayerEvent implements Cancellable 
      *
      * @param type type of the attack
      */
-    public void setType(final PlayerAttackType type) {
+    public void setType(final @NotNull PlayerAttackType type) {
         this.type = type;
     }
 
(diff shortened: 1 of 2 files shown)


dawon:
I am not sure whether the knockback should be part of the enum. IMO you can crit & knockback and sweep & knockback... Maybe a separate boolean?

Maxsh001:
> I am not sure whether the knockback should be part of the enum. IMO you can crit & knockback and sweep & knockback... Maybe a separate boolean?

In vanilla, a kb attack requires the player to be sprinting, and crit/sweep requires that the player is not sprinting. However, when you do a kb attack, unlike crit/sweep, the normal attack sound effect is also played so I'm not really sure what vanilla is doing here (or what it's intending to do) but it kinda suggests the possibility. Also kb attacks don't involve extra particles or anything so it wouldn't be hard to implement.

(I'm busy these days. Maybe this weekend)

dawon:
Just tested this and it seems to be working quite well.

paper-api/src/main/java/io/papermc/paper/event/player/PlayerAttackEntityEvent.java
```diff
@@ -0,0 +1,108 @@
...
+
+
+/**
+ * Called when the player attacks an entity.
+ * <p>
+ * This event is not called if {@link PrePlayerAttackEntityEvent} is cancelled.
+ * <p>
+ * Cancelling this event results in an attack with no damage, like when hitting an invulnerable entity. To cancel the damage completely, use {@link PrePlayerAttackEntityEvent}.
+ */
+public class PlayerAttackEntityEvent extends PlayerEvent implements Cancellable {
```
dawon, line 20: Class should be @NullMarked, then you can remove all NotNull annotations

paper-server/patches/sources/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -266,6 +315,10 @@
                      } else {
                          this.playServerSideSound(SoundEvents.PLAYER_ATTACK_NODAMAGE);
                      }
++                    // Paper end - PlayerAttackEntityEvent
```
dawon, line 318: I think this can be moved above `this.setLastHurtMob(target);` as no changes below these lines...

dawon (commented):
Couple of minor points, also probably when updating to 26.1, there will be a need to rename/fix the diff to match proper local variable names as they unobfuscated MC code...

paper-server/patches/sources/net/minecraft/world/entity/player/Player.java.patch
TransparentWhit: But `this.setLastHurtMob(target);` is in an `if` branch

dawon: AFAIK the aim is to have the shortest diff possible while making it easy to maintain and quite often that means the start and end comments might be on different levels/scopes... But I might be wrong, really this is just a nitpick, final word has the paper team...

Issue #7369, opened 2022-01-18
op permission level glitch


smartowl08:
### Expected behavior

when I set op permission level in server.properties to 0~3, ops can't use /stop command

### Observed/Actual behavior

when I set op permission level in server.properties to 0~3, ops can use /stop and other commands freely even 0. 

### Steps/models to reproduce

set op-permission-level in server.properties to 0~3
deop yourself and op again through console
test /stop

### Plugin and Datapack List

none

### Paper version

1.18.1 -160

e-im:
`/stop` requires permission level 4, this is intended behaviour. Use a permission plugin like luckperms if you would like more granularity.

e-im:
Sorry, read that a bit too quickly

zml2008:
bukkit takes over permissions handling, and it's permissions model predates the numbered op levels. and fix would require re-considering those design decisions. probably not worth fixing this when the more flexible node system already exists.

Chew:
Sulu's message still stands to an extent, op permission levels are generally unsupported and would be a pain to update as stated above. We're not stopping anyone from attempting, but this is something low priority, if any at all, imo.

People-11:
Same issue

PR #12389, opened 2025-04-07
Add ability to disable Bukkit permissions handling

luisch444:
# PR Details

## Information

This PR intends to add the ability for server to disable Bukkit permissions checking for commands. With this will be able to fix some problems with permission levels checking in Bukkit.

## Technical comments

The `PermissibleBase#hasPermission(@NotNull String inName)` is a Bukkit method that checks the permissions of a property, from the comment of @zml2008 `permissions model predates the numbered op levels` we can assume it should be thought as a refactor in the future. Because the Bukkit implementation does not allow levels of permissions.

Note: The modifications are made in the less affecting way possible, we know this might induce into some configurations changes to unintended behavior. With this I want to ask: I should enhance the hole system or only change the minimum to fix the issues

## Why do not approach the full solution?

IMHO; this should be thought, when the project has fulfill all the major goals and all the important discussions has being fixed, by the contributors for having a nice permission handling without affecting vanilla implementation while containing all the features needed, Bukkit's and Paper's, 

## Why still in draft?

This PR is in draft because, me luisch444, have not completely tested this feature and neither added any test. Nevertheless, we would like to hear back by the contributors of project if any adjustment is needed.

As well, this modifications (with the config value active) might introduces some breaks into some plugins only supported in Bukkit.

## Fixed issues (Please update/comment on need)

- Workaround to #7369

### DIFF
--- a/paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
+++ b/paper-server/patches/sources/net/minecraft/commands/CommandSourceStack.java.patch
@@ -74,7 +74,7 @@
 +    // CraftBukkit start
 +    public boolean hasPermission(int i, String bukkitPermission) {
 +        // Paper start - Fix permission levels for command blocks
-+        final java.util.function.BooleanSupplier hasBukkitPerm = () -> this.source == CommandSource.NULL /*treat NULL as having all bukkit perms*/ || this.getBukkitSender().hasPermission(bukkitPermission); // lazily check bukkit perms to the benefit of custom permission setups
++        final java.util.function.BooleanSupplier hasBukkitPerm = () -> !this.getServer().server.ignoreBukkitPermissions && (this.source == CommandSource.NULL /*treat NULL as having all bukkit perms*/ || this.getBukkitSender().hasPermission(bukkitPermission)); // lazily check bukkit perms to the benefit of custom permission setups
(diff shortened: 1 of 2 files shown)


electronicboy:
I mean, just disabling the perm system for a select set of commands, IMHO, is just a non starter; The actual fix here would probably be to make op levels mapped over to perm nodes in order to let perm plugins deal with this stuff more fluently/properly, I just don't know if there is much care to put up the engineering effort to implement and maintain such a mechanism which basically hasn't been used for well over a decade due to much better solutions existing

lynxplay:
While I am a fan of small PRs, introducing a config option to bypass a system that (presumably) is properly fixed later just for the config option to die then is not really something I am interested in.
This issue is **not problematic enough** to require a "hotfix" config option like this when simply using permissions plugins over operator levels has been common practice in the community for a near decade.

As cat stated, the "proper" solution for this would be a *lot* more work that we may or may not even be interested in maintaining (tho we do maintain CommandPermissions already so, it might be *fine*?) In general, fine grained permissions are a objectively better approach than vanillas operator system so I am not sure if a proper solution is worth maintaining.

I appreciate your work on this but I do not believe this approach is something I am willing to merge.
If you want to look into a proper mapping of permission levels to command permissions, feel free, tho that would be a larger change.
Thanks again for the work and the nicely explained PR and discussion!

PR #12390, opened 2025-04-07
Expand PotionMeta Api to allow getting effective potion colour and effects

456dev:


paper-api/src/main/java/org/bukkit/inventory/meta/PotionMeta.java
```diff
@@ -74,6 +74,16 @@ public interface PotionMeta extends ItemMeta {
...
 
+    /**
+     * All effects that this item applies.
+     * <p>
+     * This is a Combination of the base potion type and any custom effects.
+     *
+     * @return an unmodifiable list of all effects.
+     */
+    @NotNull
+    List<PotionEffect> getEffects();
```
kennytv: Maybe "all" here too, to better differentiate it from custom effects?

kennytv (approved):

paper-api/src/main/java/io/papermc/paper/datacomponent/item/PotionContents.java
```diff
@@ -60,6 +60,26 @@ static PotionContents.Builder potionContents() {
...
+
+    /**
+     * The resulting colour of the potion.
+     * <p>
+     * It may or may not have an alpha, used for tipped arrows.
+     *
+     * @return the visible colour
+     */
+    @Contract(pure = true)
+    Color resultingColor();
```
lynxplay: Any reason for this to not be named the same as "effective colour" in the PotionMeta?

456dev: I don't believe so.
I think it came from PotionMeta having just `getColor` vs the data component having them as `customColor`, where there might be some confusion.

but one or the other is fine

lynxplay (approved):

Owen1212055 (approved):

PR #12398, opened 2025-04-09
Structure - Create PlacementOptions

### LINKED ISSUE #12381: Structure#place - add option for strict placement

Not really. Just new API that would be nice.
I guess if I were to think of a problem, when I place structures, block updates happen when I may not want them to.
A good example is flowing water, waterlogging (weird bug in MC), fences updates, etc.

Minecraft 1.21.5 added a new feature for "strict" placement of structures using the structure block and `/place template` command.
Strict placement allows for the structure to be placed without block updates, ex: water not flowing.

Looking into how the command/structure block do it, it would be as easy as changing this line:
https://github.com/PaperMC/Paper/blob/9b1798d6438107fdf0d5939b79a8cf71f4d16e2c/paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java#L100
changing the flags from `2` to `2 | (strict ? 816 : 0)` and adding in a method parameter for `strict`
For reference the 816 flag is `UPDATE_SKIP_ALL_SIDEEFFECTS`

Nothing I can think of!

### Other

Visual Example:

Image

ShaneBeee:
This PR originally aimed to add an option to place structures without doing block updates.
This was something Minecraft added in 1.21.5 to Structure Blocks as well as the `/place template` command.

After a great suggestion from Kashike, this PR changed to add a new PlacementOptions class.
This will allow for deciding which options you want to add, or fallback to defaults.
This will also be great in the future if Mojang adds more options for placement, won't need to create more methods in the structure class.

While I understand the team is not wanting PRs for 1.21.5 yet, I just wanted to get this done as early as possible.
I'm ok with the PR sitting here til the team accepts 1.21.5 PRs.

This feature (the earlier mentioned strict option) has been requested by users of my plugin for over a year now, so Im super excited to have this.
I did test with my plugin locally and it works like a charm :) 

Side note (re: Strict placement):
I did a test on 1.21.4 for this, it technically works as well, but only like 50/50.
For example:
- Fences didn't connect as expected (yay)
- Water still flowed (boo)
- I didn't test other blocks but it felt a bit unstable, so I didn't want to do the PR for 1.21.4 due to instability.

Closes #12381

### DIFF
--- a/paper-api/src/main/java/org/bukkit/structure/Structure.java
+++ b/paper-api/src/main/java/org/bukkit/structure/Structure.java
@@ -107,54 +107,12 @@ public interface Structure extends PersistentDataHolder {
      * randomly skipping blocks to place. This value can range from 0 to 1. With
      * 0 removing all blocks and 1 spawning the structure in pristine condition.
      * @param random The randomizer used for setting the structure's
-     * @param strict Whether to prevent block updates (ie: prevent flowing water or fence connections)
-     * {@link org.bukkit.loot.LootTable}s and integrity.
-     */
-    @ApiStatus.Experimental
-    void place(@NotNull Location location, boolean includeEntities, @NotNull StructureRotation structureRotation, @NotNull Mirror mirror, int palette, float integrity, @NotNull Random random, boolean strict);
-
-    /**
-     * Place a structure in the world.
-     *
-     * @param location The location to place the structure at.
-     * @param includeEntities If the entities present in the structure should be
-     * spawned.
-     * @param structureRotation The rotation of the structure.
-     * @param mirror The mirror settings of the structure.
-     * @param palette The palette index of the structure to use, starting at
-     * {@code 0}, or {@code -1} to pick a random palette.
-     * @param integrity Determines how damaged the building should look by
-     * randomly skipping blocks to place. This value can range from 0 to 1. With
-     * 0 removing all blocks and 1 spawning the structure in pristine condition.
-     * @param random The randomizer used for setting the structure's
      * {@link org.bukkit.loot.LootTable}s and integrity.
      * @param blockTransformers A collection of {@link BlockTransformer}s to apply to the structure.
      * @param entityTransformers A collection of {@link EntityTransformer}s to apply to the structure.
      */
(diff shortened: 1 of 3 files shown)


paper-api/src/main/java/org/bukkit/structure/Structure.java
```diff
@@ -93,6 +93,26 @@ public interface Structure extends PersistentDataHolder {
...
+     * {@code 0}, or {@code -1} to pick a random palette.
+     * @param integrity Determines how damaged the building should look by
+     * randomly skipping blocks to place. This value can range from 0 to 1. With
+     * 0 removing all blocks and 1 spawning the structure in pristine condition.
+     * @param random The randomizer used for setting the structure's
+     * @param strict Whether to prevent block updates (ie: prevent flowing water or fence connections)
+     * {@link org.bukkit.loot.LootTable}s and integrity.
+     */
+    @ApiStatus.Experimental
+    void place(@NotNull Location location, boolean includeEntities, @NotNull StructureRotation structureRotation, @NotNull Mirror mirror, int palette, float integrity, @NotNull Random random, boolean strict);
```
kashike: Is it worth having some sort of `PlacementOptions` object instead of having multiple different methods with ~10 arguments each?

ShaneBeee: Who doesn't love 10 arguments 😬 ... jk, but yeah that might be a better idea.

ShaneBeee: Update: 
Ok, I made a placement class. 
Looking for feedback.

notTamion: can we also move the transformers onto the PlacementOptions? Or we could do this similarly to PositionedRayTraceConfigurationBuilder

ShaneBeee: At first I didn't think that would make sense, but yeah... it does make sense.
I have moved them 👍🏻

kennytv:
Oop, please reopen a PR targeting main again

PR #12407, opened 2025-04-11
CraftBlock - fix applyBoneMeal false result

ShaneBeee:
This PR aims to fix the `Block#applyBoneMeal` method always returning false.

I'm assuming the NMS method used here was probably changed at one point from CONSUME to SUCCESS

For context:
This is the method used in `CraftBlock#applyBoneMeal` for getting the `result`
Screenshot 2025-04-11 at 2 20 11 PM


electronicboy (approved):
`24w33a is the first snapshot for Java Edition 1.21.2` is probably when this got borked

ShaneBeee:
> `24w33a is the first snapshot for Java Edition 1.21.2` is probably when this got borked

Duly noted. Not too long so thats good!

PR #12411, opened 2025-04-12
Structure - add placement options

### LINKED ISSUE #12381: Structure#place - add option for strict placement

ShaneBeee:
Redo of https://github.com/PaperMC/Paper/pull/12398

This PR originally aimed to add an option to place structures without doing block updates.
This was something Minecraft added in 1.21.5 to Structure Blocks as well as the `/place template` command.

After a great suggestion from Kashike, this PR changed to add a new PlacementOptions class.
This will allow for deciding which options you want to add, or fallback to defaults.
This will also be great in the future if Mojang adds more options for placement, won't need to create more methods in the structure class.

While I understand the team is not wanting PRs for 1.21.5 yet, I just wanted to get this done as early as possible.
I'm ok with the PR sitting here til the team accepts 1.21.5 PRs.

This feature (the earlier mentioned strict option) has been requested by users of my plugin for over a year now, so Im super excited to have this.
I did test with my plugin locally and it works like a charm :) 

Side note (re: Strict placement):
I did a test on 1.21.4 for this, it technically works as well, but only like 50/50.
For example:
- Fences didn't connect as expected (yay)
- Water still flowed (boo)
- I didn't test other blocks but it felt a bit unstable, so I didn't want to do the PR for 1.21.4 due to instability.

Closes #12381


paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java
```diff
@@ -97,7 +114,7 @@ public void place(RegionAccessor regionAccessor, BlockVector location, boolean i
         access.setDelegate(handle);
         access.setStructureTransformer(new CraftStructureTransformer(handle, new ChunkPos(pos), blockTransformers, entityTransformers));
 
-        this.structure.placeInWorld(access, pos, pos, definedstructureinfo, randomSource, 2);
+        this.structure.placeInWorld(access, pos, pos, definedstructureinfo, randomSource, 2 | (placementOptions.isStrict() ? 816 : 0));
```
kennytv: This should use the actual Block constants and/or reference the placeStructure method it's taken from

kennytv (approved):
Not tested, but generally looks good!

kennytv:
Maybe also on the names, but I don't think we have proper guidelines for that, stuff like `isIncludeEntities` is a bit off, I think we usually do `shouldIncludeEntities` or `includesEntities`

paper-api/src/main/java/org/bukkit/structure/Structure.java
```diff
@@ -155,6 +155,27 @@ public interface Structure extends PersistentDataHolder {
...
+    /**
+     * Place a structure in the world.
+     *
+     * @param location The location to place the structure at.
+     * @param placementOptions Options used to determine different settings applied to the structure.
+     * @param blockTransformers A collection of {@link BlockTransformer}s to apply to the structure.
+     * @param entityTransformers A collection of {@link EntityTransformer}s to apply to the structure.
+     */
+    @ApiStatus.Experimental
+    void place(@NotNull Location location, @NotNull PlacementOptions placementOptions, @NotNull Collection<BlockTransformer> blockTransformers, @NotNull Collection<EntityTransformer> entityTransformers);
```
lynxplay: What is the point of this method, I am a bit confused here.
If the placement options already track the transformers for blocks and entities, this method seems problematic.

ShaneBeee: You're right... that was a mistaken on my part.
I originally didnt have the collections as part of the placement Options, and was suggested to include in, then silly me forgot to remove from that method. 
Oops! Will fix shortly.

paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java
ShaneBeee: Fixed!

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+        this.structureRotation = structureRotation;
+        return this;
+    }
+
+    /**
+     * The rotation a structure will place with.
+     *
+     * @return Rotation to place with.
+     */
+    public @NotNull StructureRotation getStructureRotation() {
```
notTamion:
```suggestion
    public @NotNull StructureRotation structureRotation() {
```

use adventure style naming, same for all other getters

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Set the rotation a structure will place with.
+     * <p>Will default to {@link StructureRotation#NONE} if not set.</p>
+     *
+     * @param structureRotation Rotation to place with.
+     * @return This PlacementOptions.
+     */
+    public @NotNull PlacementOptions structureRotation(@NotNull StructureRotation structureRotation) {
```
notTamion:
```suggestion
     * @return This PlacementOptions.
     */
    @Contract(value = "_ -> this", mutates = "this")
    public @NotNull PlacementOptions structureRotation(@NotNull StructureRotation structureRotation) {
```

add contract annotations, same for all setters

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+
+import com.google.common.base.Preconditions;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Random;
+import org.bukkit.block.structure.Mirror;
+import org.bukkit.block.structure.StructureRotation;
+import org.bukkit.util.BlockTransformer;
+import org.bukkit.util.EntityTransformer;
+import org.jetbrains.annotations.NotNull;
```
notTamion: use jspecify @ NullMarked on the class and only mark anything that is Nullable

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
+package org.bukkit.structure;
```
notTamion: new api goes in io.papermc.paper.*

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * Set the integrity used for placing a structure.
+     * <p>Will default to 1 if not set</p>
+     *
+     * @param integrity Determines how damaged the building should look by
+     *                  randomly skipping blocks to place. This value can range from 0 to 1. With
+     *                  0 removing all blocks and 1 spawning the structure in pristine condition.
+     * @return This PlacementOptions
+     */
+    public @NotNull PlacementOptions integrity(float integrity) {
```
notTamion: Add a Range annotation here

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Add a {@link BlockTransformer} to apply to the structure.
+     * <p>Will default to an empty list if not included.</p>
+     *
+     * @param blockTransformer BlockTransformer to apply.
+     * @return This PlacementOptions.
+     */
+    public @NotNull PlacementOptions addBlockTransformer(@NotNull BlockTransformer blockTransformer) {
+        this.blockTransformers.add(blockTransformer);
```
notTamion: This will fail if the Collection provided by the plugin is immutable. I think the normal setter which sets the entire collection should suffice.

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Set the collection of {@link BlockTransformer BlockTransformers} to apply to the structure.
+     *
+     * @param blockTransformers Collection of BlockTransformers.
+     * @return This PlacementOptions.
+     */
+    public @NotNull PlacementOptions blockTransformers(@NotNull Collection<BlockTransformer> blockTransformers) {
+        this.blockTransformers = blockTransformers;
```
notTamion: clone so plugins cannot modify outside of setter

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+        return this;
+    }
+
+    /**
+     * Collection of {@link BlockTransformer BlockTransformers} to apply to the structure.
+     *
+     * @return BlockTransformers to apply.
+     */
+    public @NotNull Collection<BlockTransformer> getBlockTransformers() {
+        return blockTransformers;
```
notTamion: clone and add UnmodifiableView annotation

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Set the collection of {@link EntityTransformer EntityTransformers} to apply to the structure.
+     * <p>Will default to an empty list if not included.</p>
+     *
+     * @param entityTransformers EntityTransformers to apply.
+     * @return This PlacementOptions.
+     */
+    public @NotNull PlacementOptions entityTransformers(@NotNull Collection<EntityTransformer> entityTransformers) {
```
notTamion: same for entity transformer methods

notTamion (changes requested):

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
ShaneBeee: Call me crazy, but the Jetbrains' Range annotation is for ints, I don't see any thing available for floats

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
ShaneBeee: I'll admit, I have never used contracts before, and frankly I have no clue how they work 😬 
I have taken what you have suggested and applied to setters, I hope I did this right, please confirm.

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
NonSwag: The annotation specifies a range _inclusive to exclusive_, and can also be applied to floats
It is mainly for users

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
notTamion: yup that looks good, This exact contract just means:
For any argument passed ( _ ) this method will return ( -> ) the current object ( this )

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
notTamion: interesting, i thought there was one. nvm then

ShaneBeee: ~~Ahh gotcha, thank you, I'll apply now.~~ 
Edit: ok, I wont add one.

paper-api/src/main/java/org/bukkit/structure/PlacementOptions.java
Lulu13022002: you forgot includeEntities

ShaneBeee: It appears I did! Oops. Fixed!

ShaneBeee:
@notTamion when you have a chance could you please re-review this. 
Would love to get this out of "changes requested"
Thank you 😊 

Would love to get this into a plugin soon.

paper-api/src/main/java/io/papermc/paper/structure/PlacementOptions.java
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Create a new set of placement options for placing a {@link Structure}.
+     *
+     * @param random The randomizer used for setting the structure's
+     *               {@link org.bukkit.loot.LootTable LootTables} and integrity.
+     */
+    public PlacementOptions(Random random) {
+        this.random = random;
+    }
```
Owen1212055: Would rather this be a static method to create that takes a random.

paper-api/src/main/java/io/papermc/paper/structure/PlacementOptions.java
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Collection of {@link EntityTransformer EntityTransformers} to apply to the structure.
+     *
+     * @return EntityTransformers to apply.
+     */
+    public @UnmodifiableView Collection<EntityTransformer> entityTransformers() {
+        return new ArrayList<>(entityTransformers);
+    }
+
```
Owen1212055: These should be immutable views.

However.... I am just SO not sure about exposing these methods they technically can be package private). Yes there is already a method that does, but in general that api is *god awful* and it's still experimental so I am not sure if our official type should expose this yet.

paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java
```diff
@@ -73,22 +72,42 @@ public void place(RegionAccessor regionAccessor, BlockVector location, boolean i
...
+        this.place(world, blockVector, placementOptions);
+    }
 
-        Preconditions.checkArgument(integrity >= 0F && integrity <= 1F, "Integrity value (%S) must be between 0 and 1 inclusive", integrity);
+    @Override
+    public void place(RegionAccessor regionAccessor, BlockVector location, PlacementOptions placementOptions) {
+        Preconditions.checkArgument(location != null, "Location cannot be null");
+        Preconditions.checkArgument(regionAccessor != null, "RegionAccessor cannot be null");
+        Collection<BlockTransformer> blockTransformers = placementOptions.blockTransformers();
+        Preconditions.checkState(blockTransformers != null, "BlockTransformers cannot be null");
```
Owen1212055: Nitpick but call all the preconditions first

paper-api/src/main/java/io/papermc/paper/structure/PlacementOptions.java
Owen1212055: Consider the bottom half of the comment not really relevant until some further conversation.

Owen1212055 (changes requested):

paper-api/src/main/java/io/papermc/paper/structure/PlacementOptions.java
```diff
@@ -219,7 +224,7 @@ public PlacementOptions blockTransformers(Collection<BlockTransformer> blockTran
      * @return BlockTransformers to apply.
      */
     public @UnmodifiableView Collection<BlockTransformer> blockTransformers() {
-        return new ArrayList<>(blockTransformers);
+        return List.copyOf(blockTransformers);
```
NonSwag, line 227: `List#copyOf` returns an independent copy so instead of `@UnmodifiableView` it should be `@Unmodifiable` since changes to the original aren't reflected within the copy

paper-api/src/main/java/io/papermc/paper/structure/PlacementOptions.java
```diff
@@ -241,7 +247,7 @@ public PlacementOptions entityTransformers(Collection<EntityTransformer> entityT
      * @return EntityTransformers to apply.
      */
     public @UnmodifiableView Collection<EntityTransformer> entityTransformers() {
-        return new ArrayList<>(entityTransformers);
+        return List.copyOf(entityTransformers);
```
NonSwag, line 250: same here

NonSwag (changes requested):

NonSwag:
Just a general thought 
It would be more consistent to make the PlacementOptions class an interface
That way the random instance doesn't have to be passed or exposed and could be moved to the internals unless that is intentionally expised

Soniflare:
Hi, I don't know if this is still being worked on or not, but I would still love to see this added, as I'm at a really big roadblock in my soon-to-be-public server, because I am not able to (as far as I'm aware) use strict placement to paste my custom structure files.  This is a really big issue for me.  I hope this gets approved soon.

paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java
```diff
@@ -73,31 +72,48 @@ public void place(RegionAccessor regionAccessor, BlockVector location, boolean i
...
         BlockPos pos = CraftBlockVector.toBlockPosition(location);
         WorldGenLevel handle = ((CraftRegionAccessor) regionAccessor).getHandle();
 
         TransformerGeneratorAccess access = new TransformerGeneratorAccess();
         access.setDelegate(handle);
-        access.setStructureTransformer(new CraftStructureTransformer(handle, new ChunkPos(pos), blockTransformers, entityTransformers));
+        access.setStructureTransformer(new CraftStructureTransformer(handle, new ChunkPos(pos), placementOptions.blockTransformers(), placementOptions.entityTransformers()));
 
-        this.structure.placeInWorld(access, pos, pos, definedstructureinfo, randomSource, 2);
+        // Strict placement option copied from net.minecraft.server.commands.PlaceCommand#placeTemplate
```
notTamion, line 115: Add a diff on change comment to where you got it from.

paper-server/src/main/java/org/bukkit/craftbukkit/structure/CraftStructure.java
```diff
@@ -73,31 +72,48 @@ public void place(RegionAccessor regionAccessor, BlockVector location, boolean i
...
-                .setMirror(net.minecraft.world.level.block.Mirror.valueOf(mirror.name()))
-                .setRotation(Rotation.valueOf(structureRotation.name()))
-                .setIgnoreEntities(!includeEntities)
-                .addProcessor(new BlockRotProcessor(integrity))
+                .setMirror(net.minecraft.world.level.block.Mirror.valueOf(placementOptions.mirror().name()))
+                .setRotation(Rotation.valueOf(placementOptions.structureRotation().name()))
+                .setIgnoreEntities(!placementOptions.includeEntities())
+                .addProcessor(new BlockRotProcessor(placementOptions.integrity()))
+                .setKnownShape(placementOptions.strict())
+                .setLiquidSettings(placementOptions.applyWaterlogging() ? LiquidSettings.APPLY_WATERLOGGING : LiquidSettings.IGNORE_WATERLOGGING)
```
notTamion, line 104: might be worth adding a LiquidSettings enum to the api so we won't have to do so later incase mojang adds some new option (like in #12162). Can't think of anything they would add here though just in case

notTamion (changes requested):

Ninjamrcool:
I would also very much like this for my server, would optimize things a ton

EysseW:
I am still waiting for this :)

PR #11925, opened 2025-01-06
Add registry builder for Instrument

Doc94:
This PR aims to provide more fields to the MusicInstrument currently this class with many deprecated methods for the instrument but not provide any method for the fields in instrument https://minecraft.wiki/w/Data_component_format#instrument like the description or duration.

This depends on https://github.com/PaperMC/Paper/pull/11918 for items with instruments custom

PR #12415, opened 2025-04-12
Expose more data for MusicInstrument

Doc94:
This PR take the base of https://github.com/PaperMC/Paper/pull/11925 for expose the data in MusicInstrument for keep the old PR just for create new ones when support exists.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/MusicInstrument.java
+++ b/paper-api/src/main/java/org/bukkit/MusicInstrument.java
@@ -60,9 +60,9 @@ public abstract class MusicInstrument implements Keyed, net.kyori.adventure.tran
     }
 
     /**
-     * Gets for how long the use duration is for the instrument.
+     * Gets the use duration of this music instrument.
      *
-     * @return the duration.
+     * @return the duration expressed in seconds.
      */
     public abstract float getDuration();
 
@@ -83,7 +83,7 @@ public abstract class MusicInstrument implements Keyed, net.kyori.adventure.tran
     /**
      * Gets the sound for this instrument.
      *
-     * @return a sound
+     * @return the sound
      */
     public abstract Sound getSound();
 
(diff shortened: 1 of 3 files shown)


Athar42:
Thanks @Doc94 to still looking into this, still waiting for it to get into Paper so I can finally get those custom goat horn working like I want them 🤣

paper-api/src/main/java/org/bukkit/MusicInstrument.java
```diff
@@ -48,25 +50,50 @@ public static MusicInstrument getByKey(@NotNull NamespacedKey namespacedKey) {
...
+     * @return the description component.
+     */
+    public abstract Component description();
+
+    /**
+     * Gets the sound/sound-event for this instrument.
+     *
+     * @return a sound
+     */
+    public abstract Sound getSoundEvent();
```
kennytv: I'd just call it getSound instead of SoundEvent

Doc94: Hmm just the method not?
I set this name for avoid confusing for what this get and make consist with the create instrument PR.

Owen1212055 (approved):
LGTM Otherwise please change the method name of sound.
Consistency with the instrument PR is not needed in this case since those represent the concrete schema types.

Doc94:
> LGTM Otherwise please change the method name of sound. Consistency with the instrument PR is not needed in this case since those represent the concrete schema types.

okay this and with the Kenny comment i rename the method

paper-server/src/main/java/org/bukkit/craftbukkit/CraftMusicInstrument.java
```diff
@@ -63,8 +65,28 @@ public CraftMusicInstrument(Holder<Instrument> holder) {
...
+        return this.getHandle().useDuration();
+    }
+
+    @Override
+    public float getRange() {
+        return this.getHandle().range();
+    }
+
+    @Override
+    public Component description() {
```
kennytv, line 83: Just do getDescription to keep it consistent with the rest, we don't have to be scared of upstream conflicts anymore

lynxplay (approved):

PR #12417, opened 2025-04-12
Registry API for supported Mob Variants

BertTowne:
Re-upload of #12364


BertTowne:
This should build/test successfully now. Internal mob variants should not have had `SpawnPrioritySelectors.EMPTY`

paper-api/src/main/java/org/bukkit/entity/Chicken.java
```diff
@@ -59,6 +62,11 @@ public interface Chicken extends Animals {
      */
     interface Variant extends Keyed {
 
+        enum Model {
```
lynxplay: I am a bit iffy on having the model be defined here.
Given those are very *very* version specific, I think I'd prefer them on the registry entries.

In general, I am unsure how useful exposing the assetId and model of a variant truely is to the *general* server API. I don't *really* have a usecase in mind for it, if you have one please share :)
Otherwise, I think we'd be best of just not exposing these in the stable API.

paper-api/src/main/java/org/bukkit/entity/Cow.java
```diff
@@ -40,6 +49,20 @@ interface Variant extends Keyed {
...
+         * @return the asset id
+         */
+        @NotNull Key assetId();
+
+        /**
+         * Get the cow variant's {@link Cow.Variant.Model}. Defaults to {@link Cow.Variant.Model#NORMAL}.
+         *
+         * @return the model
+         */
+        @Nullable Model getModel();
```
lynxplay: Same here as on chicken.

paper-api/src/main/java/org/bukkit/entity/Frog.java
```diff
@@ -61,6 +62,13 @@ interface Variant extends OldEnum<Variant>, Keyed {
         Variant WARM = getVariant("warm");
         // End generate - FrogVariant
 
+        /**
+         * Get the frog variant's asset id
+         *
+         * @return the asset id
+         */
+        @NotNull Key assetId();
```
lynxplay: Same here as on chicken.

paper-api/src/main/java/org/bukkit/entity/Pig.java
```diff
@@ -31,6 +34,11 @@ public interface Pig extends Steerable, Vehicle {
      */
     interface Variant extends Keyed {
 
+        enum Model {
```
lynxplay: Same here as on chicken.

paper-api/src/main/java/org/bukkit/entity/Wolf.java
```diff
@@ -131,6 +132,27 @@ interface Variant extends Keyed {
         Variant WOODS = getVariant("woods");
         // End generate - WolfVariant
 
+        /**
+         * Get the wolf variant's asset id used when the wolf is angry
+         *
+         * @return the asset id
+         */
+        @NotNull Key assetIdAngry();
```
lynxplay: Same here as on chicken.

lynxplay (changes requested):
Glanced a bit more, this is not correctly tracking the fact that ClientAsset can have a different texture path from the id.
That either needs to be tracked internally or exposed to the API.

lynxplay:
FWIW, I have diff fixing a lot of this locally. I'll push that tomorrow before we duplicate work here.

BertTowne:
Sounds good, thank you! I didn't see the comments until I got the notif for this, but I do no have any use in mind for exposing that additional variant info to the general API - I more so added that for completionism. Leaving it in the registry entries makes sense to me.

lynxplay:
Pushed over my changes, will review down the line. Feel free to also review my changes, hope I didn't miss anything.

image

paper-api/src/main/java/io/papermc/paper/registry/data/client/ClientAsset.java
```diff
@@ -0,0 +1,66 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
+
+/**
+ * A data-centric version-specific representation of a client asset, composed of an identifier and a path.
+ * Follows the same, version-specific, compatibility guarantees as the RegistryEntry types it is used in.
+ */
+@ApiStatus.Experimental
+@ApiStatus.NonExtendable
+public interface ClientAsset {
```
masmc05: I feel like something like `TextureAsset` would be a better name, as the api already uses other client assets like item models, which aren't represented by/with this.

lynxplay: Hmmm, yea *maybe*. ClientTextureAsset?
I agree it is a bit weird that some things are only represented via a Key, others by this new type, but that is kinda how internals go and the registry data API is pretty much an API mirror of those.

masmc05: ClientTextureAsset sounds good
> but that is kinda how internals go

Yeah, ik, still I consider copy pasting non-datapack nms names that are confusing in context of the api isn't that good

lynxplay: Yea, I guess the method already includes "texture" in there.
If things change down the line, we can rename it anyway as this isn't stable API.

BertTowne:
I like the changes. `ClientTextureAsset` is much cleaner. I do feel kinda bad about not touching SpawnConditions at all, but I would have no idea how to elegantly implement that in a way that can not only be used by these custom mob variants, but possibly other mobs through a custom SpawnCondition api.

lynxplay:
Generally, it is perfectly fine like this and if/once a usecase for spawn conditions is raised by someone, they can PR it :+1:

lynxplay (approved):

Owen1212055 (approved):

PR #12422, opened 2025-04-13
Added `missingRespawnBlock` to PlayerRespawnEvent

NonSwag:
Introduced a new property to check if the respawn block is missing.


lynxplay:
The ctors are internal so we can change them :sweat_smile: 
No point in overloading more.

NonSwag:
I would like to add a setter for this as well but i am not sure what to do about `PlayerList#respawn L929`
That line clears the respawn location if the respawn block is missing _fix for SPIGOT-5988 and calling the PlayerSetSpawnEvent_

NonSwag:
> The ctors are internal so we can change them 😅 No point in overloading more.

Alright, did that just because the other constructors are internal as well but marked as removal since idk how long.

NonSwag:
also is there a chance this could be backported to 1.21.4?

kennytv (approved):

kennytv:
Yeah, you can make a follow-up PR after this is merged, or we'll backport it later

NonSwag:
I clarified the docs for the `isMissingRespawnBlock` method lmk if it is fine like that

notTamion (approved):

PR #12423, opened 2025-04-13
Fix SignChangeEvent NPE (& external list mutability)

SoSeDiK:
- Updated SignChangeEvent class
	- Fix NPE on `getLine()` / `getLines()` / legacy String[] constructor
	- Updated the annotations to JSpecify, removed Paper comments and fully qualified imports as it's quite a small class & change
	- `getLine()` / `line()` / constructor now correctly show nullability
		- To clarify, this PR does not break the null contract, but simply adds a missing nullability annotations for those
	- Change lines handling to not allow external modifications (otherwise iffy things are possible like passing an immutable list, changing the amount of elements) and mandate 4 elements. Adds a method to set all the lines at once as an alternative.
		- That's changing behavior (updating the list won't affect event anymore), but I personally wasn't expecting this in the first place (it's not mentioned in javadoc, and differs from other similar places within API), and it's not a major break. The legacy getter didn't allow mutability already.

---

~~I'm sorry this grew a liiiiiiiiiitle bigger than initially planned as changes just snowballed 🫠
Can be easily split if any of the changes are debatable.~~
Edit: split into https://github.com/PaperMC/Paper/pull/12462

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/block/SignChangeEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/block/SignChangeEvent.java
@@ -59,7 +59,7 @@ public class SignChangeEvent extends BlockEvent implements Cancellable {
         this.player = thePlayer;
         this.adventure$lines = new ArrayList<>(4);
         for (String theLine : theLines) {
-            this.adventure$lines.add(theLine == null ? null : LegacyComponentSerializer.legacySection().deserialize(theLine));
+            this.adventure$lines.add(LegacyComponentSerializer.legacySection().deserializeOrNull(theLine));
         }
         this.side = side;
     }
@@ -128,7 +128,7 @@ public class SignChangeEvent extends BlockEvent implements Cancellable {
      */
     @Deprecated
     public @Nullable String[] getLines() {
-        return this.adventure$lines.stream().map(line -> line == null ? null : LegacyComponentSerializer.legacySection().serialize(line)).toArray(String[]::new);
+        return this.adventure$lines.stream().map(LegacyComponentSerializer.legacySection()::serializeOrNull).toArray(String[]::new);
     }
 
     /**
@@ -144,8 +144,7 @@ public class SignChangeEvent extends BlockEvent implements Cancellable {
     @Nullable
     @Deprecated
     public String getLine(int index) throws IndexOutOfBoundsException {
-        Component line = line(index);
-        return line == null ? null : LegacyComponentSerializer.legacySection().serialize(line);
+        return LegacyComponentSerializer.legacySection().serializeOrNull(this.adventure$lines.get(index));
     }
 
     /**
@@ -159,7 +158,7 @@ public class SignChangeEvent extends BlockEvent implements Cancellable {
      */
     @Deprecated
     public void setLine(int index, @Nullable String line) throws IndexOutOfBoundsException {
-        this.adventure$lines.set(index, line != null ? LegacyComponentSerializer.legacySection().deserialize(line) : null);
+        this.adventure$lines.set(index, LegacyComponentSerializer.legacySection().deserializeOrNull(line));
     }
 
     /**
--- a/paper-api/src/main/java/org/bukkit/event/block/SignChangeEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/block/SignChangeEvent.java
@@ -1,5 +1,6 @@
 package org.bukkit.event.block;
 
+import com.google.common.base.Preconditions;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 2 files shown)


AnttiMK:
MM did a pull request on this way back which was never merged (https://github.com/PaperMC/Paper/pull/9468). That PR used empty components instead of explicit nulls in the lists -  would it be easier to work with the lines if those were used instead?

SoSeDiK:
> MM did a pull request on this way back which was never merged (#9468). That PR used empty components and a non-null transforming list instead of explicit nulls in the lists - would it be easier to work with the lines if those were used instead?

Oh, haven't seen that PR, thanks for the info! Missed `(de)serializeOrNull` being a thing, updated.

I do agree that getting rid of nulls would be better, but this will break backwards compatibility, and sadly the event existed in this state for years now.

1. By changing nulls, suddenly old "empty/untouched line" checks based on null will no longer pass.

2. Similarly, changing lines getter to immutable will break plugins relying on its mutability.
	- There could be an API break for this one, since the mutability was never an API contract and likely wasn't intended, but suddenly breaking it isn't perfect either imo.

lynxplay:
We definitely cannot break the non null contract on the lines getters/setters.
We might be able to get away with null in the internal list but it would need to be transformed to empty().
If a legacy plugin calls setLine(0, null) and a follow up plugin calls `line(0, line(0))`, I think it is *fine* to remove the null value there. They boil down to the same thing once consumed by the server.

But yea, "breaking" (eventho yea, it has been technically broken) newer API in favour of older ones, *especially when this has been the case since 1.16.5 and hence been part of the API for  more than 4 years now is not a change I'd be happy with.

lynxplay:
Any update on this? Given that it bundles the isInRaitOrWater check, it is a bit too on my radar.

SoSeDiK:
> Any update on this? Given that it bundles the isInRaitOrWater check, it is a bit too on my radar.

I've moved the first two commits into a separate PR (https://github.com/PaperMC/Paper/pull/12462) given that the sign changes seem stuck.

> We definitely cannot break the non null contract on the lines getters/setters.

To clarify, the changes in this PR do not break the null contract. `getLine`/ `line` were always nullable and `getLines`/`lines` were (and currently are) returning nulls as well, just that the javadoc didn't cover that.

> We might be able to get away with null in the internal list

If we remove the `null` internally, there are two things to consider:
1. Are we fine breaking existing `null` checks in plugins? empty() will pass those
2. `lines` currently returns the internal (array) list directly. We either need to break mutability or return a list that would reject/convert nulls on changes to the list (event constructor also need to convert nulls then)

@Machine-Maker, thoughts? With your prior changes in https://github.com/PaperMC/Paper/pull/9468, would be nice to hear opinions

lynxplay:
> Ive moved the first two commits into a separate PR (https://github.com/PaperMC/Paper/pull/12462) given that the sign changes seem stuck.

thank you!
> To clarify, the changes in this PR do not break the null contract. getLine/ line were always nullable and getLines/lines were (and currently are) returning nulls as well, just that the javadoc didn't cover that.

I do think MM's original PR is nicer in this regard yea, tho yea you are also right, I missed the fact that the component ones are nullable already. Presumably *only* due to the fact that legacy strings have that nasty habit.
> Are we fine breaking existing null checks in plugins? empty() will pass those

I'd be at least. Something going from nullable to non null is generally *fine* as the non null version could have already been in there. Especially for components, for legacy strings yea, we might be able to find some hacky way to yield back null.
> lines currently returns the internal (array) list directly. We either need to break mutability or return a list that would reject/convert nulls on changes to the list (event constructor also need to convert nulls then)

Ah yes, the bane of everyones existence, mutable lists. I hate them with a passion tho yea, that is the bit that I'd be most worried about breaking. MM's PR introduced a nag for this, which sounds like a solid solution for this issue.
Before investing a lot of time into this, lemme throw this PR at the rest of the team tho, so not all of this is based on my personal opinion :sweat_smile:

Warriorrrr:
I don't think breaking mutability in this case would be too bad, I could only find [two public plugins](https://github.com/search?q=%28%22.lines%28%29.add%22+OR+%22.lines%28%29.remove%22+OR+%22.lines%28%29.set%22+OR+%22.lines%28%29.clear%22+OR+%22.lines.addAll%22%29+%22org.bukkit.event.block.SignChangeEvent%22+language%3AJava+NOT+is%3Afork&type=code) on github who would be affected by the change, the majority are already correctly using the line(int, Component) method to set a component.

Owen1212055:
I agree, breaking is not too bad here imo. But we should do this most likely on an update rather than mid cycle.

PR #12425, opened 2025-04-13
Expose block, entity and flag context to PlayerPickItemEvent

NonSwag:
Enhanced event to include associated block, entity, and data flag.


kennytv:
Maybe this should just be separate events from an abstract class

NonSwag:
Initially I had the same idea, but I wasn't sure how backwards compat would work
Would PlayerPickItemEvent listeners still be called if the exact event isn't called anymore?

NonSwag:
> Would PlayerPickItemEvent listeners still be called if the exact event isn't called anymore?

To answer my own question, yes the event is still called

Owen1212055 (approved):

lynxplay (approved):
A little on the fence regarding the naming for PlayerPickEntityEvent, as it does not really convey that the player is picking an item by clicking the entity.
It instead sounds a bit like the player is picking an entity itself (somehow). Open for naming changes but, if team is fine with this, also down with merging this this weekend.

Vrganj:
This now requires a BlockPos/Entity be passed to the tryPickItem method and gets rid of the old signature. Could this be reviewed as I and others have relied on the method without a relevant BlockPos or Entity?

PR #12427, opened 2025-04-13
Add BlockType/ItemType getters and setters

ImTheLion:
Adds getters/setters in some classes, making BlockType and ItemType API even more useable without converting from/to material.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/block/Block.java
+++ b/paper-api/src/main/java/org/bukkit/block/Block.java
@@ -102,6 +102,14 @@ public interface Block extends Metadatable, Translatable, net.kyori.adventure.tr
     Material getType();
 
     /**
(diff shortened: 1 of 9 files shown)


lynxplay:
Hi, thank you for the PR and welcome to paper :partying_face: !
I am semi certain we are not planning on doing it this way as Item/BlockType will simply *replace* the existing Material parameters and return types when we retire Material.

I'll check in with the team on that tho.

electronicboy:
I believe that this kinda thing was intended as a core team sorta task given that we also just wanted to start purging material at the same time and have bytecode translations for it

lynxplay:
Okay, yea that is indeed the plan.
Thank you for the PR but we have different plans for the integration of those type :)

PR #12429, opened 2025-04-13
Make GameRule a FeatureDependant

DerEchtePilz:
So I was playing around with game rules a bit and after running into an issue that `GameRule#values()` does contain even game rules that are disabled because the experimental data pack isn't enabled, I was wondering if there was a way to check if a game rule is enabled or not.

What I've found is that `World#getGameRules()` should contain only enabled game rules, however that wasn't applicable for my usecase since I didn't have a world available.

This PR adds an `isEnabled()` method to the `GameRule` class to make it possible to check if the game rule is enabled without needing access to a world.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
@@ -247,14 +247,6 @@
              GameRules.VisitorCaller<T> visitorCaller,
              Class<T> valueClass,
              FeatureFlagSet requiredFeatures
-@@ -608,6 +_,7 @@
-             this.visitorCaller.call(visitor, key, this);
-         }
- 
-+        @Override
-         public FeatureFlagSet requiredFeatures() {
-             return this.requiredFeatures;
-         }
 @@ -620,16 +_,16 @@
              this.type = type;
          }
--- a/paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
@@ -14,7 +14,7 @@
 +
      public static final int DEFAULT_RANDOM_TICK_SPEED = 3;
      static final Logger LOGGER = LogUtils.getLogger();
-     private static final Map<GameRules.Key<?>, GameRules.Type<?>> GAME_RULE_TYPES = Maps.newTreeMap(Comparator.comparing(entry -> entry.id));
+     public static final Map<GameRules.Key<?>, GameRules.Type<?>> GAME_RULE_TYPES = Maps.newTreeMap(Comparator.comparing(entry -> entry.id));
 @@ -86,10 +_,10 @@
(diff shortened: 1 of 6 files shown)


lynxplay:
We already have a system for that around FeatureDependant and FeatureFlagSetHolder.
GameRule should implement FeatureDependant

DerEchtePilz:
That doesn't seem to be the case.

lynxplay:
Yea it doesn't rn xD SOrry I wasnt clear, the PR should do that instead of adding a new system.

DerEchtePilz:
Okay, I'll do that then 👍

lynxplay (approved):
Generally LGTM!
I do wonder however if it would be smart to patch the GameRules internals to also maintain a Map<String, GameRule.Type<?>> for faster access.
The fact that this method is going to be pretty spam-heavy for checks would mean such optimisations would have a hopefully measurable effect.

paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
```diff
@@ -243,6 +247,14 @@
              GameRules.VisitorCaller<T> visitorCaller,
              Class<T> valueClass,
              FeatureFlagSet requiredFeatures
+@@ -608,6 +_,7 @@
+             this.visitorCaller.call(visitor, key, this);
+         }
+ 
++        @Override
```
lynxplay: Paper comments :+1:

paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
```diff
@@ -14,7 +14,7 @@
 +
      public static final int DEFAULT_RANDOM_TICK_SPEED = 3;
      static final Logger LOGGER = LogUtils.getLogger();
-     private static final Map<GameRules.Key<?>, GameRules.Type<?>> GAME_RULE_TYPES = Maps.newTreeMap(Comparator.comparing(entry -> entry.id));
+     public static final Map<GameRules.Key<?>, GameRules.Type<?>> GAME_RULE_TYPES = Maps.newTreeMap(Comparator.comparing(entry -> entry.id));
```
masmc05, line 17: Maybe https://github.com/PaperMC/Paper/blob/main/CONTRIBUTING.md#access-transformers would be better here?

paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
masmc05: Since the method is already here and this isn't an abstract class, maybe it would be better to just not include the optional Override annotation to reduce diff?

paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
lynxplay: That is an AT.

paper-server/patches/sources/net/minecraft/world/level/GameRules.java.patch
lynxplay: Yea, probably a good call.

PR #12435, opened 2025-04-14
Expand cooldown API

dawon:
Currently, if a plugin wants to set a Cooldown group on cooldown, they need either use NMS or they need to construct an ItemStack, which should not be necessary...

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
+++ b/paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
@@ -3,6 +3,8 @@ package org.bukkit.entity;
 import java.util.Collection;
 import java.util.Set;
 import java.util.function.Consumer;
+import io.papermc.paper.datacomponent.item.UseCooldown;
+import net.kyori.adventure.key.Key;
 import org.bukkit.GameMode;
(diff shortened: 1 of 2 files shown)


kennytv (approved):

paper-api/src/main/java/org/bukkit/entity/HumanEntity.java
```diff
@@ -428,6 +429,31 @@ default void closeInventory() {
...
+     * shields to prevent them from being used repeatedly.
+     * <p>
+     * Note that cooldowns will not by themselves stop an item from being used
+     * for attacking.
+     *
+     * @param key cooldown group to set the cooldown for
+     * @param ticks the amount of ticks to set or 0 to remove
+     * @see UseCooldown#cooldownGroup()
+     */
+    public void setCooldown(NamespacedKey key, int ticks);
```
masmc05: Can this be an adventure key instead? (the conversation is then in PaperAdventure instead of CraftNamespacedKey)

dawon: Good point, thank you, done!

lynxplay (approved):

PR #12445, opened 2025-04-16
Add isSuffocating for Block and BlockState

Doc94:
This PR add methods for Block and BlockState for check if can Suffocate, this behaviour cannot be check using isSolid (ex: Leaves are consider solid but cannot suffocate)

image
image

### DIFF
--- a/paper-api/src/main/java/org/bukkit/block/Block.java
+++ b/paper-api/src/main/java/org/bukkit/block/Block.java
@@ -821,7 +821,7 @@ public interface Block extends Metadatable, Translatable, net.kyori.adventure.tr
     /**
      * Checks if the block can suffocate.
      *
-     * @return <code>true</code> if the block can suffocate
+     * @return {@code true} if the block can suffocate
      */
     boolean isSuffocating();
 }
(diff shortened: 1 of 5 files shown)


kennytv (approved):

paper-api/src/main/java/org/bukkit/block/Block.java
```diff
@@ -817,4 +817,11 @@ default float getDestroySpeed(@NotNull ItemStack itemStack, boolean considerEnch
         return this.getBlockData().getDestroySpeed(itemStack, considerEnchants);
     }
     // Paper end - destroy speed API
+
+    /**
+     * Checks if the block can suffocate.
+     *
+     * @return <code>true</code> if the block can suffocate
```
Lulu13022002: You can use the `@code` tag instead, same for the other method

paper-server/src/main/java/org/bukkit/craftbukkit/block/CraftBlockState.java
```diff
@@ -75,6 +75,7 @@ public void setWorldHandle(LevelAccessor generatorAccess) {
 
     // Returns null if weakWorld is not available and the BlockState is not placed.
     // If this returns a World instead of only a GeneratorAccess, this implies that this BlockState is placed.
+    @org.jspecify.annotations.Nullable
```
Lulu13022002: Don't mix nullable annotations, keep using javax if you don't want to clean things

PR #12448, opened 2025-04-16
Effective Locales

lukashert:
Adds Player#setEffectiveLocale(Locale) / Player#getEffectiveLocale() to set and get a player-specific locale, which is used for server-side translations. 

Closes #10570

### DIFF
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -7,6 +7,7 @@ import java.time.Instant;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -3282,14 +3283,15 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      *
      * @return the locale.
      */
-    java.util.Locale getEffectiveLocale();
+    Locale getEffectiveLocale();
 
     /**
      * Change the locale the server will be translating its messages to.
+     * If set, the effective locale overrides the client locale for server-side translations.
      *
      * @param locale the locale to translate to
      */
-    void setEffectiveLocale(java.util.@Nullable Locale locale);
+    void setEffectiveLocale(@Nullable Locale locale);
 
     /**
      * Gets the player's estimated ping in milliseconds.
--- a/paper-api/src/main/java/org/bukkit/entity/Player.java
+++ b/paper-api/src/main/java/org/bukkit/entity/Player.java
@@ -3275,6 +3275,22 @@ public interface Player extends HumanEntity, Conversable, OfflinePlayer, PluginM
      */
     java.util.Locale locale();
     // Paper end
+
+    /**
+     * Returns the locale the server will use to send messages translated via the Adventure global translator.
+     * By default, the value of {@link Player#locale()} is used.
+     *
+     * @return the locale.
+     */
+    java.util.Locale getEffectiveLocale();
+
+    /**
+     * Change the locale the server will be translating its messages to.
+     *
+     * @param locale the locale to translate to
+     */
+    void setEffectiveLocale(java.util.@Nullable Locale locale);
+
     /**
      * Gets the player's estimated ping in milliseconds.
      *
--- a/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
@@ -1,12 +1,13 @@
 --- a/net/minecraft/server/level/ServerPlayer.java
 +++ b/net/minecraft/server/level/ServerPlayer.java
-@@ -235,7 +_,8 @@
+@@ -235,7 +_,9 @@
      private int levitationStartTime;
      private boolean disconnected;
      private int requestedViewDistance = 2;
 -    public String language = "en_us";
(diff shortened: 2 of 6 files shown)


kennytv (approved):

paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java
```diff
@@ -2996,6 +2996,22 @@ public java.util.Locale locale() {
...
+        if (getHandle().effectiveLocale == null) {
+            return locale();
+        }
+        return getHandle().effectiveLocale;
+    }
+
+    @Override
+    public void setEffectiveLocale(final java.util.Locale locale) {
+        getHandle().effectiveLocale = locale;
+        getHandle().connection.connection.channel.attr(io.papermc.paper.adventure.PaperAdventure.LOCALE_ATTRIBUTE).set(getEffectiveLocale());
```
Lulu13022002: This will throw a NPE, if called in the PlayerLoginEvent for example

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3275,6 +3275,22 @@ default boolean hasResourcePack() {
...
     java.util.Locale locale();
     // Paper end
+
+    /**
+     * Returns the locale the server will use to send messages translated via the Adventure global translator.
+     * By default, the value of {@link Player#locale()} is used.
+     *
+     * @return the locale.
+     */
+    java.util.Locale getEffectiveLocale();
```
lynxplay: We have hardforked and can now import things in non-patched sources like the API :partying_face:

paper-api/src/main/java/org/bukkit/entity/Player.java
```diff
@@ -3275,6 +3275,22 @@ default boolean hasResourcePack() {
...
+
+    /**
+     * Change the locale the server will be translating its messages to.
```
lynxplay, line 3289: I'd maybe also mention that the effective locale will override any changes from the client as well.
E.g. if a plugin reads a players locale as en_us, sets their effective locale to en_gb, the player changing their locale to de_de will not have an affect on the effective local.

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -1,12 +1,13 @@
...
 +++ b/net/minecraft/server/level/ServerPlayer.java
-@@ -235,7 +_,8 @@
+@@ -235,7 +_,9 @@
      private int levitationStartTime;
      private boolean disconnected;
      private int requestedViewDistance = 2;
 -    public String language = "en_us";
 +    public String language = null; // Paper - default to null
 +    public java.util.Locale adventure$locale = java.util.Locale.US; // Paper
++    public java.util.Locale effectiveLocale = null; // Paper
```
lynxplay, line 11: Ought to be annotated with `@Nullable`. I believe nms uses the javax annotation set for this type.

lynxplay (changes requested):

lynxplay:
I am semi uncertain on the naming "effective locale".
I know this mirrors velocity's API, however in the context of velocity, it being "effective" makes sense, given that all of velocity's messages are using the global translator to be resolved.

This differs drastically on backend servers like paper, where basically *none* of the messages sent are actually translated on the server side beyond plugin messages. Command feedback, server system messages, etc are all not translated on the server but instead resolved by the client itself.
Given I don't have an immediate better idea for the naming, nor  am I sure this change in naming is 100% needed, I'll throw this at the rest of team/contributors.

davidmayr:
> Given I don't have an immediate better idea for the naming, nor am I sure this change in naming is 100% needed, I'll throw this at the rest of team/contributors.

How about just mentioning what that locale actually does in the javadoc?

lynxplay:
Well, javadocs are a nice tool. But they are generally a second resort and do not save us from poorly named methods.
A user looking for a way to force a locale will still run into Player#setEffectiveLocale and just *not* read those docs until already hammering their head into a wall as to why it isn't working.

paper-server/patches/sources/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -1,12 +1,14 @@
...
 +++ b/net/minecraft/server/level/ServerPlayer.java
-@@ -235,7 +_,8 @@
+@@ -235,7 +_,10 @@
      private int levitationStartTime;
      private boolean disconnected;
      private int requestedViewDistance = 2;
 -    public String language = "en_us";
 +    public String language = null; // Paper - default to null
 +    public java.util.Locale adventure$locale = java.util.Locale.US; // Paper
++    @Nullable
```
Doc94, line 10: try put this in the same line because the comment only apply to the field and not this.

lynxplay:
Yea certainly a good point to expand the javadocs here regardless :+1:

kezz:
Seeing as Velocity also has this and Adventure has a lot of requests of users wanting to override the locale used for server-side translating, I've gone ahead and opened a PR to implement this in Adventure directly, which also solves the issue this PR has about confusing developers. https://github.com/KyoriPowered/adventure/pull/1221

Brokkonaut:
How about Player.get/setServerSideLocale() ?

lynxplay:
Yea, setServerSideLocale is a nice name, however the adventure PR might be nicer.

Owen1212055:
Seeing above, I am going to go ahead and close this PR.

Thank you so much for your contribution, but this will be much better supported natively by the adventure PR listed above.

PR #12454, opened 2025-04-18
Support DamagePerBlock for Virtual WorldBorder

### LINKED ISSUE #12372: Fix virtual world borders
### Expected behavior

Virtual world borders set to specific player would act as normal world border, taking damage when outside of the world border and not glitching when global world border is smaller then player's one.

### Observed/Actual behavior

When player has virtual worldborder, server doesn't check if player is inside or outside of the worldborder which causes, that player doesn't take damage when outside but still takes damage if outside the global world border.

### Steps/models to reproduce

Using this code, I'm able to set player's world border to be visible only to him with specific center and size (virtual world border).

Checking that position is always in relation to global world border, which creates weird invisible border, if the global one is smaller than player's one that he can see.

I didn't found any other way how to set virtual worldborder to certain player with allowing to check its position and dealing damage afterwards. (If it is possible, than correct me <3)

```java
CraftWorldBorder worldBorder = new CraftWorldBorder(new net.minecraft.world.level.border.WorldBorder());
worldBorder.setCenter(x + 0.5d, z + 0.5d);
worldBorder.setSize(borderSize);
worldBorder.setWarningTime(0);
worldBorder.setWarningDistance(0);
worldBorder.setDamageAmount(5);
worldBorder.setDamageBuffer(0);
bukkitPlayer.setWorldBorder(worldBorder);
```

### Plugin and Datapack List

ℹ Server Plugins (1):
Paper Plugins:
 - PaperTestPlugin      <--- Custom plugin, that virtual world border to player (code mentioned above)

### Paper version

This server is running Paper version 1.21.4-222-main@9b1798d (2025-03-27T13:35:40Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
Previous version: 1.21.4-212-bb1beda (MC: 1.21.4)

### Other

_No response_

Doc94:
This PR close https://github.com/PaperMC/Paper/issues/12372 by adding the support for server to consider the damage, this make the break change where the WorldBorder created by Server/Bukkit set the damage in zero.

### DIFF
--- a/paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
@@ -60,6 +60,36 @@
              }
          }
  
+@@ -394,10 +_,10 @@
+             boolean flag = this instanceof Player;
+             if (this.isInWall()) {
+                 this.hurtServer(serverLevel1, this.damageSources().inWall(), 1.0F);
+-            } else if (flag && !serverLevel1.getWorldBorder().isWithinBounds(this.getBoundingBox())) {
+-                double d = serverLevel1.getWorldBorder().getDistanceToBorder(this) + serverLevel1.getWorldBorder().getDamageSafeZone();
++            } else if (flag && !this.getWorldBorder().isWithinBounds(this.getBoundingBox())) { // Paper - use the first WorldBorder available
++                double d = this.getWorldBorder().getDistanceToBorder(this) + serverLevel1.getWorldBorder().getDamageSafeZone(); // Paper - use the first WorldBorder available
+                 if (d < 0.0) {
+-                    double damagePerBlock = serverLevel1.getWorldBorder().getDamagePerBlock();
++                    double damagePerBlock = this.getWorldBorder().getDamagePerBlock(); // Paper - use the first WorldBorder available
+                     if (damagePerBlock > 0.0) {
+                         this.hurtServer(serverLevel1, this.damageSources().outOfBorder(), Math.max(1, Mth.floor(-d * damagePerBlock)));
+                     }
+@@ -478,6 +_,15 @@
+         return Mth.lerp((float)this.getAttributeValue(Attributes.MOVEMENT_EFFICIENCY), super.getBlockSpeedFactor(), 1.0F);
+     }
+ 
++    // Paper start - helper for WorldBorder
++    private net.minecraft.world.level.border.WorldBorder getWorldBorder() {
++        if (this instanceof ServerPlayer serverPlayer && serverPlayer.getBukkitEntity().hasClientWorldBorder()) {
++            return ((org.bukkit.craftbukkit.CraftWorldBorder)serverPlayer.getBukkitEntity().getWorldBorder()).getHandle();
++        }
++        return this.level().getWorldBorder();
++    }
++    // Paper end
++
+     public float getLuck() {
+         return 0.0F;
+     }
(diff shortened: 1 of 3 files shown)


Doc94:
I need a few feedback for the follow case:

If create a WorldBorder and set damage all fine, but if i change the World WorldBorder that change override in client the WorldBorder make you can not notice the virtual WorldBorder and get damage by the virtual... then not sure what is the best aproach for this:
- Notice this in JavaDoc for make users handle this in plugins
- Modify the BorderChangeListener in the Level class for add a filter and avoid send changes to players who has virtual worldborder and make Player#setWorldBorder refresh the data

456dev:
2 makes more sense to me, and makes the API easier to use

Esp if other parts of the server still think the virtual world border is still active / it mixes settings with the update packets

Doc94:
> 2 makes more sense to me, and makes the API easier to use
> 
> Esp if other parts of the server still think the virtual world border is still active / it mixes settings with the update packets

yeah i found two places where the listener are created with the send packets... for PlayerList and Level for that i wanna wait more feedback and then touch that

paper-server/patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -60,6 +60,36 @@
...
          }
  
+@@ -394,10 +_,10 @@
+             boolean flag = this instanceof Player;
+             if (this.isInWall()) {
+                 this.hurtServer(serverLevel1, this.damageSources().inWall(), 1.0F);
+-            } else if (flag && !serverLevel1.getWorldBorder().isWithinBounds(this.getBoundingBox())) {
+-                double d = serverLevel1.getWorldBorder().getDistanceToBorder(this) + serverLevel1.getWorldBorder().getDamageSafeZone();
++            } else if (flag && !this.getWorldBorder().isWithinBounds(this.getBoundingBox())) { // Paper - use the first WorldBorder available
++                double d = this.getWorldBorder().getDistanceToBorder(this) + serverLevel1.getWorldBorder().getDamageSafeZone(); // Paper - use the first WorldBorder available
```
456dev, line 70: Very nitpicky
But I don't believe "use first world border available" really describes the changes

Maybe smth like "use players worldborder, if present"

davidmayr:
I personally would like an API to fully disable this. I use the virtual world boarder to simulate the world boarder not being there (as I don't like its visual presence), and I would personally like it to remain this way. 

I'd like an API that makes the player receives damage if the player leaves the "real" world boarder, even if his personal world boarder is larger. (Basically make it fully the same as it is now) And I personally think my suggestion should be the default (as otherwise it is sort of a breaking change). 

Setting the damage to 0 by default is a good start, but the real world boarder is still not active anymore

Owen1212055:
Ngl, it’s so easy to just check it yourself and damage the player urself. I also don’t really see why this is super useful especially as this is described as a sort of “virtual” border.

electronicboy:
as I said, the only way I could support this would be if it was an optional override to the main scoreboard, changing behavior here is just bleh, people use the client one for visual stuff and expect the global border to still be active and do stuff, even if not visual to players

electronicboy:
> Scoreboard?

long day ;-;

Owen1212055:
Yeah I am not quite sure about this. Its called a virtual world border meaning that I would think that this doesnt affect players (with damage, etc) so, I think this is better for plugins to add themselves if they want.

Doc94:
Based in comments close this in favor of https://github.com/PaperMC/Paper/pull/13338

PR #12456, opened 2025-04-18
Add GhastShootFireballEvent

BlueNightFury46:
I've added a new event that fires when a Ghast fires a fireball, I've implemented some basic setExplosionPower and getExplosionPower methods to mimic those provided by net.minecraft.world.entity.monster.Ghast in the future (such as a method to modify the speed, etc). I opted to set the values directly, because I didn't want to use any weird casts within the event listener's method to get a bukkit ghast.

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/entity/GhastShootFireballEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/entity/GhastShootFireballEvent.java
@@ -1,0 +1,41 @@
+package io.papermc.paper.event.entity;
+
+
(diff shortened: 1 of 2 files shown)


lynxplay:
Thank you for your first PR.

In what way does this event expose any benefits one could not already implement via the ProjectileLaunchEvent?
```java
@EventHandler
public void on(final ProjectileLaunchEvent event) {
    if (!(event.getEntity() instanceof final Fireball fireball && fireball.getShooter() instanceof final Ghast ghast)) return;
    
    ......
}
```
does (at least in my mind) the exact same already.

PR #12461, opened 2025-04-20
Improved Attribute Handling with New Attributable Interface Methods

xcutiboo:
This pull request enhances the Attributable interface in Attributable.java by introducing a set of useful new methods for managing and interacting with attributes more effectively. These additions make it easier to check if an attribute exists, access all registered attributes, and work with both their base and final values.

**Here's what's new:**

**Check if an attribute exists:**
The new hasAttribute(@NotNull Attribute attribute) method lets you verify whether a specific attribute is currently registered.

**Access all registered attributes:**
The getAttributes() method returns a collection of all attributes currently registered to the object.

**Manage base values of attributes:**

getBaseValue(@NotNull Attribute attribute) retrieves the base value of a given attribute. If the attribute isn't registered, it returns 0 by default.

setBaseValue(@NotNull Attribute attribute, double value) sets the base value of a given attribute, and throws an exception if the attribute isn't registered.

**Get final computed value of an attribute:**

The getValue(@NotNull Attribute attribute) method calculates the final value of an attribute, taking all modifiers into account. If the attribute isn't registered, it also returns 0 by default

### DIFF
--- a/paper-api/src/main/java/com/destroystokyo/paper/ParticleBuilder.java
+++ b/paper-api/src/main/java/com/destroystokyo/paper/ParticleBuilder.java
@@ -3,21 +3,62 @@ package com.destroystokyo.paper;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


masmc05:
I think you forgot to add the implementation of the added methods.

Also did you want to include the ParticleBuilder changes? You didn't include those in your pr description.

lynxplay:
The PR seems to not add anything useful that isn't already present on the AttributeInstance.
Additionally you seem to have random changes to the particle builder.
We will not be pulling this, thank you for your PR!

PR #12462, opened 2025-04-22
Undeprecate Entity#isInWaterOrRain & correct some javadocs

SoSeDiK:
- Javadoc corrections
	- `Tool.Rule#speed()` had a description for `Tool.Rule#correctForDrops()`
	- `AbstractArrow#setItem` now points to `#setItemStack` instead of `#getItemStack`
- Undeprecate `Entity#isInWaterOrRain`
	- I've asked in Discord about this, there weren't any strong opinions for/against it.
	- Deprecating `BubbleColumn` overload in 1.21.5 makes sense, but the rain one still exists internally and is a neat way to check "is this entity wet?". I don't mind leaving it deprecated, but also I've used this method 34 times in just one codebase and my lazy kicked in.
		- <details>
			  <summary>Image</summary>
			  
			image
			  
			</details>

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/datacomponent/item/Tool.java
+++ b/paper-api/src/main/java/io/papermc/paper/datacomponent/item/Tool.java
@@ -1,6 +1,7 @@
 package io.papermc.paper.datacomponent.item;
 
 import io.papermc.paper.datacomponent.DataComponentBuilder;
+import io.papermc.paper.datacomponent.DataComponentTypes;
 import io.papermc.paper.registry.set.RegistryKeySet;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


lynxplay (approved):

Lulu13022002:
Kinda against this, what prevent you from having this method in your plugin if you have a lot of use for it, being lazy is not an argument for me. The method existing internally doesn't matter, api goal is not to map every methods found in the source but rather providing powerful tools.

lynxplay:
General idea imo is the concept of "is the entity wet".
E.g. might distinguish fire.
Tho, urgh, yea *technically* that ought to be a separate method instead of this syntax sugar.

lynxplay:
Given the change was on the "no one has strong opinions, we know it isn't great but *sure*" basis, and we have found an opinion on it, I dropped the deprecation removal :+1:

SoSeDiK:
Sure, that's fine by me. I've updated the deprecated javadocs then instead

PR #12463, opened 2025-04-22
Fix CCE in LingeringPotionSplashEvent

wizjany:
Regression in 1.21.5 update ([ref](https://github.com/PaperMC/Paper/commit/f00727c57e564f3a8cb875183a54142feb693db7#diff-bd1c99844d4a4fbb8a38c4e3519004b029ab994dc325fbb513ccb62313ee7a0eR41), if it loads).

Might ultimately make more sense to rename the AoE cloud instead of having two things named `entity` in "scope" but that's a separate issue.


electronicboy (approved):

electronicboy:
Welcome to paper! Thanks!

PR #12466, opened 2025-04-23
Remove redundant distance check.

ScriptLineStudios:
Taking a look at the logic for block breaking, it seems an unnecessary, duplicate distance/interaction check is being performed. 

This distance check feels a bit redundant, as immediately after, a more comprehensive sanity check is performed in handleBlockBreakAction.

```java
public void handleBlockBreakAction(BlockPos pos, ServerboundPlayerActionPacket.Action action, Direction face, int maxBuildHeight, int sequence) {
        if (!this.player.canInteractWithBlock(pos, 1.0)) {
            if (true) return; // Paper - Don't allow digging into unloaded chunks; Don't notify if unreasonably far away
            this.debugLogging(pos, false, sequence, "too far");
        } else if (pos.getY() > maxBuildHeight) {
            this.player.connection.send(new ClientboundBlockUpdatePacket(pos, this.level.getBlockState(pos)));
            this.debugLogging(pos, false, sequence, "too high");
        } else {
            ...
        }
}
```

### DIFF
--- a/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
+++ b/paper-server/patches/sources/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
@@ -1139,7 +1139,7 @@
                  case ABORT_DESTROY_BLOCK:
                  case STOP_DESTROY_BLOCK:
 +                    // Paper start - Don't allow digging into unloaded chunks
-+                    if (this.player.level().getChunkIfLoadedImmediately(pos.getX() >> 4, pos.getZ() >> 4) == null || !this.player.canInteractWithBlock(pos, 1.0)) {
++                    if (this.player.level().getChunkIfLoadedImmediately(pos.getX() >> 4, pos.getZ() >> 4) == null) {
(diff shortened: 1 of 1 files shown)


Owen1212055:
Closing this as in general it's not really certain what you are trying to fix here. Feel free to respond with a more detailed description of your changes, but it's not really understood what kind of "unnecessary  distance/interaction check" this is resolving here.

ScriptLineStudios:
> Closing this as in general it's not really certain what you are trying to fix here. Feel free to respond with a more detailed description of your changes, but it's not really understood what kind of "unnecessary distance/interaction check" this is resolving here.

It's not trying to "fix" anything, it's just observing that the same check is being performed here twice

electronicboy:
There is a difference in the final outcome of those events, it's not exactly redundant

ScriptLineStudios:
> There is a difference in the final outcome of those events, it's not exactly redundant

is there? both branches end up with the attempt being acknowledged, sorry I can't tell if I'm missing something else here.

Owen1212055:
There's a bit of logic that will be run when removing this check. So no, not redundant.

PR #12475, opened 2025-04-25
Fix cursor item being lost when PlayerDropItemEvent is cancelled and …

florianreuth:
…the player inventory is full

See this video for a tutorial on how to reproduce this issue: https://youtu.be/xonJe2S7jQ8. Note that you also need to have the PlayerDropItemEvent cancelled in the test plugin.

### DIFF
--- a/paper-server/patches/features/0003-Entity-Activation-Range-2.0.patch
+++ b/paper-server/patches/features/0003-Entity-Activation-Range-2.0.patch
@@ -527,10 +527,10 @@ index fd0cc1bf9e4425a0924ed77854907deec1a7348e..1c9e5f61d182cf60caa885135abddc87
              movement = this.maybeBackOffFromEdge(movement, type);
              Vec3 vec3 = this.collide(movement);
 diff --git a/net/minecraft/world/entity/LivingEntity.java b/net/minecraft/world/entity/LivingEntity.java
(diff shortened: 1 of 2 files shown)


B1ue-Bandit:
Listen to InventoryClickEvent instead and cancel it if the click type is DROP

FlorianMichael:
What? The event is cancellable for a reason and this is clearly a wrong implementation, why would this be a proper fix?

B1ue-Bandit:
No no, you're right. Your fix would be a welcome change, I was just suggesting a workaround for the time being but I probably should've said that

FlorianMichael:
Oh I see, thanks for the suggestion, I have it already fixed in our paper
fork and decided to upstream the change :)

Blue Bandit ***@***.***> schrieb am Fr., 25. Apr. 2025, 05:22:

lynxplay:
This only resolves the issue when dropping the entire stack on cursor.
Dropping a single stack from the cursor does still loose items (specifically, the items left in the cursor being overridden by the dropped stack)

notTamion (approved):

Clexus:
A question, what if player quit the game with full inventory and items on cursor? will them drop or just disappear

FlorianMichael:
@Clexus The item will be lost, I also noticed something similar happening when using the bundle item.

aerulion:
For reference, there has already been a discussion about this issue (#7726 and #7727). Although I still like the idea of having the overflow consumer described in the linked pull request.

PR #12478, opened 2025-04-25
removal: unused warnings

SpigotRCE:


NonSwag:
I think kicking the player is a bad idea and the logic should be untouched
Players with bad ping or laggy servers might cause inventory desync so this would kick legit players for nothing

kennytv (approved):

PR #12488, opened 2025-04-28
Add missing exception docs to Player#listPlayer

NonSwag:
The exception was not documented


NonSwag:
internals as reference
image

Lulu13022002:
No need to repeat the word throws here.

NonSwag:
fixed

lynxplay (approved):

PR #12489, opened 2025-04-28
Better javadoc of getHotbarButton (off-hand swaps)

semenishchev:
I was working on inventory plugin and I was quite confused by the documentation and off-hand swaps between inventories (when you click F while aimed at a slot in a different inventory), so in this change I've cleared up the confusion.

### DIFF
--- a/paper-api/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
+++ b/paper-api/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
@@ -203,10 +203,13 @@ public class InventoryClickEvent extends InventoryInteractEvent {
 
     /**
      * If the ClickType is NUMBER_KEY, this method will return the index of
-     * the pressed key (0-8).
+     * the pressed key (0-8) and -1 if player swapped with off-hand (or the action is not NUMBER_KEY).
+     * ClickType -> NUMBER_KEY, getHotbarButton() -> -1 -> off-hand swap
+     * ClickType -> NUMBER_KEY, getHotbarButton() -> 0-8 -> hotbar slots swap
+     * ClickType -> other than NUMBER_KEY, getHotbarButton() -> -1 -> something else
      *
-     * @return the number on the key minus 1 (range 0-8); or -1 if not
-     *     a NUMBER_KEY action
+     * @return the number on the key minus 1 (range 0-8);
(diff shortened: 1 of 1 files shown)


paper-api/src/main/java/org/bukkit/event/inventory/InventoryClickEvent.java
```diff
@@ -203,10 +203,13 @@ public int getRawSlot() {
 
     /**
      * If the ClickType is NUMBER_KEY, this method will return the index of
-     * the pressed key (0-8).
+     * the pressed key (0-8) and -1 if player swapped with off-hand (or the action is not NUMBER_KEY).
+     * ClickType -> NUMBER_KEY, getHotbarButton() -> -1 -> off-hand swap
```
lynxplay: What is the point of these three additional lines?
They describe nothing the text above didn't already describe.

semenishchev: to absolutely clear up any confusion. you can suggest how to reword it, but I think it's very important to mention off-hand swap. I've spent like 30 minutes before realising that functionality

lynxplay: Yea I am fine with the addition on the first sentence.
I am just confused on the 
```diff
+     * ClickType -> NUMBER_KEY, getHotbarButton() -> -1 -> off-hand swap
+     * ClickType -> NUMBER_KEY, getHotbarButton() -> 0-8 -> hotbar slots swap
+     * ClickType -> other than NUMBER_KEY, getHotbarButton() -> -1 -> something else
```
addition. Feels pretty unfinished.

semenishchev: it specifies each case of the event state. Basically if ClickType is NUMBER_KEY, hotbar button is -1, then its an offhand swap, if click type is number key, hotbar button is between 0-8, then it's a hotbar swap, otherwise its another action

lynxplay: But doesn't the sentence right above also describe this? This just feels like doubled info in an objectively pretty random format.

semenishchev: fair

lynxplay (approved):

PR #12490, opened 2025-04-28
Allow Server#getDefaultGameMode before worlds are initialized

NonSwag:
Instead of accessing the level directly we return the configured value from properties 

Without fix
image

With fix
image

### DIFF
--- a/paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
+++ b/paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
@@ -41,6 +41,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.imageio.ImageIO;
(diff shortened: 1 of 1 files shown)


lynxplay (approved):

lynxplay:
There isn't much we can do in that case tho. We cannot modify the server properties state.
We could maybe fail nicer?

Doc94:
> There isn't much we can do in that case tho. We cannot modify the server properties state. We could maybe fail nicer?

yeah i was thinking about that.. a Precondition with a state exception but that can be another PR not? or in this for avoid a +1 xd

lynxplay:
Well, rn it doesn't fail at all, it just doesn't do anything.
Maybe we nag? I don't know how useful the setter even is but yea, definitely can just be a new PR.

Lulu13022002:
getMap has a similar issue if you want to look, also .overworld() exists too

PR #12492, opened 2025-04-28
Fix NPE in Server#getMap before worlds are loaded

NonSwag:


paper-server/src/main/java/org/bukkit/craftbukkit/CraftServer.java
```diff
@@ -1945,10 +1945,10 @@ public BiomeProvider getBiomeProvider(String world) {
     @Override
     @Deprecated
```
Doc94:
```suggestion
```

lynxplay (approved):

PR #12471, opened 2025-04-24
 Add WorldDifficultyChangeEvent

NonSwag:
I added a WorldDifficultyChangeEvent to allow cancellation and tracking of difficulty changes
My specific usecase for this is to set the difficulty of specific other worlds to the same when changed via command/API

PR #12501, opened 2025-04-30
Added DefaultGameModeChangeEvent and Server#isForceGameMode

NonSwag:
javadocs are coming

### DIFF
--- a/paper-api/src/main/java/io/papermc/paper/event/server/DefaultGameModeChangeEvent.java
+++ b/paper-api/src/main/java/io/papermc/paper/event/server/DefaultGameModeChangeEvent.java
@@ -8,6 +8,9 @@ import org.jetbrains.annotations.ApiStatus;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


Owen1212055:
What exactly is the point of this? I feel like this can just be an infinite game of wack a mole with "x change event" values that are only set by commands/api anyways.

NonSwag:
Personally I need such an event to allow per world default game modes and to be able to update the game mode of just specific players instead of all

lynxplay:
I pretty much agree with owen here. This game of "only a command or the API can change something, I need an event for this" feels like quite the endless.
*Especially* around defaultgamemode, which does two very different things between the API call (which sets it on every loaded world) and the command, which only sets it on the primary world data.

Same can be said about #12471 to *some* degree, tho at least there it is per world and difficult *might* be an actually useful concept compared to default game mode :skull:

Owen1212055:
Closing this on the general consensus above, thank you for your contribution however this seems a bit too niche for the api.. especially for a behavior which is so easy to be replicated anyways.

